PR #3761, opened 2019-09-03
Forward declare util::Status when possible

wilhuff:
This is an experiment to see if there's any savings to be had by forward declaring when possible. I figured `util::Status` is widely used and included so it might be a good test case for whether or not there was overall savings to be had.

With this change the build times improve:
  * Full build, just Firestore: -2.75 s (-3.5% of 79.0838 s)
  * Incremental build, just Firestore: 1 s (-2.2% of 44.840)

Unfortunately, a full build including everything only shows a very minor improvement (-1.8% 145.496 s).

You can reproduce with this [benchmark script](https://gist.github.com/wilhuff/8e4635f14dff5cc7e2732c693448875c) I threw together.


Firestore/core/src/firebase/firestore/util/status_fwd.h
```diff
@@ -14,22 +14,28 @@
...
+
 template <typename T>
 using StatusOrCallback = std::function<void(util::StatusOr<T>)>;
 
 }  // namespace util
 }  // namespace firestore
 }  // namespace firebase
 
-#endif  // FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_UTIL_STATUSOR_CALLBACK_H_
+#endif  // FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_UTIL_STATUS_H_
```
var-const: Seems like the comment is incorrect (missing `FWD` part).

Issue #3728, opened 2019-08-30
Crashed at FIRComponentContainer on 6.7.0


pot8os:
### [REQUIRED] Step 2: Describe your environment

  * Firebase SDK version: 6.7.0
  * Firebase Component: Below is from my `Podfile`.
```
    pod 'Firebase/Core', '~> 6.0'
    pod 'Firebase/AdMob'
    pod 'Firebase/Auth'
    pod 'Google-Mobile-Ads-SDK'
    pod 'Firebase/Messaging'
    pod 'Firebase/Functions'
    pod 'Firebase/RemoteConfig'
    pod 'Firebase/Performance'
```

#### Steps to reproduce:
Not sure how to reproduce so far, I just found some users of my app got this crash due to this issue. Seems it accesses non-existent key against cached instances.
https://github.com/firebase/firebase-ios-sdk/blob/bb563c8497c0a8fb4d237604062535701dd7ba25/Firebase/Core/FIRComponentContainer.m#L150
```
Crashed: com.google.perf.FPREventsQueue
0  libobjc.A.dylib              0x223548d68 objc_msgSend + 8
1  CoreFoundation               0x22424bb60 -[__NSDictionaryM objectForKeyedSubscript:] + 212
2  MyApp                        0x104e26d78 -[FIRComponentContainer instanceForProtocol:] + 150 (FIRComponentContainer.m:150)
3  MyApp                        0x104e2710c +[FIRComponentType instanceForProtocol:inContainer:] + 25 (FIRComponentType.m:25)
4  MyApp                        0x104ef7eac -[FIRRemoteConfigComponent remoteConfigForNamespace:] + 4380917420
5  MyApp                        0x104ef4438 +[FIRRemoteConfig remoteConfigWithFIRNamespace:app:] + 4380902456
6  MyApp                        0x104e714c0 __38+[FPRRemoteConfigFlags sharedInstance]_block_invoke + 4380366016
7  ???                          0x223d9a484 (Missing)
8  ???                          0x223d6a234 (Missing)
9  MyApp                        0x104e71468 +[FPRRemoteConfigFlags sharedInstance] + 4380365928
10 MyApp                        0x104e702cc -[FPRConfigurations setupRemoteConfigFlags] + 4380361420
11 MyApp                        0x104e7020c -[FPRConfigurations update] + 4380361228
12 MyApp                        0x104e73500 __42-[FPRClient startWithConfiguration:error:]_block_invoke + 4380374272
[...]

PR #3766, opened 2019-09-03
Gate access to cached instances.

ryanwilson:
This is causing an issue in #3728. Hopefully this will fix it, but no
repro is available yet to test. There's still an underlying race
condition where a `creationBlock` could be called twice for a component
that is supposed to cache the instance returned, but that should be
addressed in a follow up PR.

### DIFF
--- a/Firebase/Core/FIRComponentContainer.m
+++ b/Firebase/Core/FIRComponentContainer.m
@@ -23,7 +23,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface FIRComponentContainer ()
+@interface FIRComponentContainer () {
+  dispatch_queue_t _containerQueue;
+}
 
 /// The dictionary of components that are registered for a particular app. The key is an NSString
 /// of the protocol.
@@ -67,6 +69,8 @@ static NSMutableSet<Class> *sFIRComponentRegistrants;
     _app = app;
     _cachedInstances = [NSMutableDictionary<NSString *, id> dictionary];
     _components = [NSMutableDictionary<NSString *, FIRComponentCreationBlock> dictionary];
+    _containerQueue =
+        dispatch_queue_create("com.google.FirebaseComponentContainer", DISPATCH_QUEUE_SERIAL);
 
     [self populateComponentsFromRegisteredClasses:allRegistrants forApp:app];
   }
@@ -92,7 +96,7 @@ static NSMutableSet<Class> *sFIRComponentRegistrants;
       // Store the creation block for later usage.
       self.components[protocolName] = component.creationBlock;
 
-      // Instantiate the
+      // Instantiate the instance if it has requested to be instantiated.
       BOOL shouldInstantiateEager =
           (component.instantiationTiming == FIRInstantiationTimingAlwaysEager);
       BOOL shouldInstantiateDefaultEager =
@@ -136,7 +140,9 @@ static NSMutableSet<Class> *sFIRComponentRegistrants;
 
   // The instance is ready to be returned, but check if it should be cached first before returning.
   if (shouldCache) {
-    self.cachedInstances[protocolName] = instance;
+    dispatch_sync(_containerQueue, ^{
+      self.cachedInstances[protocolName] = instance;
+    });
   }
 
   return instance;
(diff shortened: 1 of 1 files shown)


Firebase/Core/FIRComponentContainer.m
```diff
@@ -161,14 +171,16 @@ - (nullable id)instanceForProtocol:(Protocol *)protocol {
...
   // Loop through the cache and notify each instance that is a maintainer to clean up after itself.
-  for (id instance in self.cachedInstances.allValues) {
-    if ([instance conformsToProtocol:@protocol(FIRComponentLifecycleMaintainer)] &&
-        [instance respondsToSelector:@selector(appWillBeDeleted:)]) {
-      [instance appWillBeDeleted:self.app];
+  dispatch_sync(_containerQueue, ^{
+    for (id instance in self.cachedInstances.allValues) {
+      if ([instance conformsToProtocol:@protocol(FIRComponentLifecycleMaintainer)] &&
+          [instance respondsToSelector:@selector(appWillBeDeleted:)]) {
+        [instance appWillBeDeleted:self.app];
```
maksymmalyhin: Nit: Do you think it may somehow call protected by `_containerQueue`? If yes, then we may use separate `dispatch_sync` to copy `self.cachedInstances` and for `[self.cachedInstances removeAllObjects];`

Firebase/Core/FIRComponentContainer.m
```diff
@@ -147,7 +153,11 @@ - (nullable id)instantiateInstanceForProtocol:(Protocol *)protocol
 - (nullable id)instanceForProtocol:(Protocol *)protocol {
   // Check if there is a cached instance, and return it if so.
   NSString *protocolName = NSStringFromProtocol(protocol);
-  id cachedInstance = self.cachedInstances[protocolName];
+  __block id cachedInstance = nil;
```
paulb777: nit: nil initialization is dead code before dispatch_sync

Firebase/Core/FIRComponentContainer.m
```diff
@@ -161,14 +171,16 @@ - (nullable id)instanceForProtocol:(Protocol *)protocol {
 
 - (void)removeAllCachedInstances {
   // Loop through the cache and notify each instance that is a maintainer to clean up after itself.
-  for (id instance in self.cachedInstances.allValues) {
-    if ([instance conformsToProtocol:@protocol(FIRComponentLifecycleMaintainer)] &&
-        [instance respondsToSelector:@selector(appWillBeDeleted:)]) {
-      [instance appWillBeDeleted:self.app];
+  dispatch_sync(_containerQueue, ^{
+    for (id instance in self.cachedInstances.allValues) {
```
paulb777: This mutating while looping would have been an issue before the queue was added.

Firebase/Core/FIRComponentContainer.m
```diff
@@ -161,14 +171,16 @@ - (nullable id)instanceForProtocol:(Protocol *)protocol {
...
+      if ([instance conformsToProtocol:@protocol(FIRComponentLifecycleMaintainer)] &&
+          [instance respondsToSelector:@selector(appWillBeDeleted:)]) {
```
paulb777: Isn't the selector check redundant in a mandatory protocol?

ryanwilson: I think it's still good practice - I can't remember if it'd result in a compiler error or just a warning but if it's a warning, I'd rather skip cleanup than crash with unknown selector.

Firebase/Core/FIRComponentContainer.m
ryanwilson: I don't think so but it's probably a good idea regardless. Will look at doing so.

Firebase/Core/FIRComponentContainer.m
paulb777: I meant if another thread changed it during the loop, there would be a __NSFastEnumerationMutationHandler crash.

ryanwilson: Ahhh got it, yes that would have been an issue for sure. I don't think it's super common though since it only happens on `FirebaseApp.delete()`, but definitely an issue.

Firebase/Core/FIRComponentContainer.m
ryanwilson: Done.

PR #3768, opened 2019-09-03
Transition from NSAssert to GDTAssert in all GDT libraries

### LINKED ISSUE #3747: [GoogleDataTransport][I-GDT001006]
* Xcode version: 10.3
  * Firebase SDK version: 6.7.0
  * Firebase Component: Core, Messaging, DynamicLinks
  * Component version: FirebaseCore = 6.2.1 ; FirebaseAnalytics = 6.1.1 ; FirebaseDynamicLinks 4.0.3 ; FirebaseMessaging = 4.1.3 ; FirebaseAnalytics = 6.1.1

I recently updated my app with the latest firebase version. After launching the app and leaving it in idle for about 20 seconds, the app crashes and this is what I read from the console:

`[GoogleDataTansport][I-GDT001006] There was an error reading extension bytes from disk: Error Domain=NSCocoaErrorDomain Code=260 "Impossible open file “event-2009730023690886256” because there is no file of that type"`

mikehaney24:
Fixes #3747 

- Moves GDTAssert to be a public API
- Always logs to console
- Differentiates between fatal and non-fatal assertions
- Messaging of assertions encourages creating an issue
- Tests assertion use with NS_BLOCK_ASSERTIONS defined and undefined


GoogleDataTransport/GDTLibrary/GDTAssert.m
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-#import "GDTLibrary/Private/GDTAssert.h"
+#import "GDTLibrary/Public/GDTAssert.h"
```
paulb777: Why Public?

GoogleDataTransport/GDTLibrary/GDTStorage.m
```diff
@@ -86,6 +86,10 @@ - (void)storeEvent:(GDTEvent *)event {
     }];
   }
 
+  if (event == nil) {
```
paulb777: Move to top of function?

GoogleDataTransport/GDTLibrary/Public/GDTConsoleLogger.h
```diff
@@ -56,7 +56,12 @@ typedef NS_ENUM(NSInteger, GDTMessageCode) {
   GDTMCETransportBytesError = 1005,
 
   /** For general purpose error messages in a dependency. */
-  GDTMCEGeneralError = 1006
+  GDTMCEGeneralError = 1006,
+
+  /** For fatal errors. Please go to https://github.com/firebase/firebase-ios-sdk/issues and open
+   * an issue if you encounter an error with this code.
+   */
+  GDTMCEFatalAssertion = 1007
```
paulb777: I'd prefer something like "Serious" over "Fatal".  From an app developer perspective, It won't be fatal if GDT fails while doing CoreDiagnostics. 

I think it should be left to the clients to determine if something is fatal or not.

GoogleDataTransport/GDTLibrary/GDTAssert.m
mikehaney24: Because it needs to be visible to GDTCCT and I've been trying to avoid module imports in implementation files.

GoogleDataTransport/GDTLibrary/Public/GDTConsoleLogger.h
mikehaney24: This has to do with the operation of GDT, not CoreDiagnostics. Maybe while FIRCD is the only consumer this is true, but soon SDK service data will transit through GDT.

paulb777: Since the comment seems to be intended for anyone who sees a crash, will you clarify in the comment that a fatal error in GDT may not be fatal for the app depending upon what the GDT client is using it for.

paulb777:
Still unresolved comment, but just a nit.

GoogleDataTransport/GDTLibrary/GDTStorage.m
mikehaney24: Done.

PR #3769, opened 2019-09-04
Fix erroneous move

var-const:
Here, the error actually has to be copied because it is passed to more than one listener. Currently, this doesn't make a difference because `util::Status` has no move constructor.

Also, make the relevant method take the status by const reference.


var-const:
@wu-hui Introducing this is my oversight, apologies.

PR #3773, opened 2019-09-04
Add move operations to Status.

wu-hui:
Note the introduction of `is_moved_from_` in `Status`, this is needed to make moved `Status`' `ok() == false`. Otherwise we see errors around destructing moved `StatusOr`.


Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -132,6 +137,10 @@ class ABSL_MUST_USE_RESULT Status {
   // a `State` structure containing the error code and message(s)
   std::unique_ptr<State> state_;
 
+  // Whether this instance has been moved. Without this `ok()` will be true
+  // for moved instances.
+  bool is_moved_from_ = false;
```
wilhuff: Accessing a moved-from instance is not something we especially need to care about because you're not supposed to do it. It's perfectly reasonable (and normal) for the moved-from state to be equivalent to the default-initialized state. Indeed some in the community advocate that this is the way to do it: https://foonathan.net/blog/2016/08/24/move-default-ctor.html.

We should avoid this flag. StatusOr's destructor is making incorrect assumptions about the behavior of Status after move. The assertion is wrong and should be fixed/removed.

wu-hui: As discussed offline, reusing the `State*` member to tag if it is moved.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -161,6 +221,10 @@ inline Status::State::State(const State& s)
                                                    : s.platform_error->Copy()) {
 }
 
+inline Status::State::State(Error code, std::string msg)
+    : code(code), msg(msg) {
```
var-const: Move `msg`. `Status` is used widely, so these copies would add up.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
+  // OK status has a `nullptr` state_, and `IsMoved() == false`.  Otherwise,
+  // either `state_` points to a `State` structure containing the error code
+  // and message(s), or this instance has been moved.
+  StatePtr state_;
+
+  // Whether this instance has been moved. Without this `ok()` will be true
+  // for moved instances.
+  bool IsMoved() const;
+  // Tags this instance as `moved`.
+  void SetMoved();
```
var-const: Nit: I think `SetMovedFrom` would be slightly clearer.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
+  }
+
+  // OK status has a `nullptr` state_, and `IsMoved() == false`.  Otherwise,
+  // either `state_` points to a `State` structure containing the error code
+  // and message(s), or this instance has been moved.
+  StatePtr state_;
+
+  // Whether this instance has been moved. Without this `ok()` will be true
+  // for moved instances.
+  bool IsMoved() const;
```
var-const: After implementing the rest of the feedback, I think this function is no longer used.

Firestore/core/src/firebase/firestore/util/status.cc
```diff
@@ -28,9 +28,8 @@ namespace util {
 
 Status::Status(Error code, absl::string_view msg) {
   HARD_ASSERT(code != Error::Ok);
-  state_ = absl::make_unique<State>();
-  state_->code = code;
-  state_->msg = static_cast<std::string>(msg);
+  state_ = MakeStatePtr(
+      Status::State(std::move(code), static_cast<std::string>(msg)));
```
var-const: Using the variadic overload, this can be just
```cpp
state_ = MakeStatePtr(code, msg);
```
I.e., just forward the arguments to `State` constructor, without calling the constructor yourself.

Firestore/core/src/firebase/firestore/util/status.cc
```diff
@@ -69,11 +68,24 @@ Status& Status::WithPlatformError(std::unique_ptr<PlatformError> error) {
   return *this;
 }
 
+bool Status::IsMoved() const {
+  return IsStatePtrMoved(state_);
+}
+
+void Status::SetMoved() {
+  if (IsMoved()) {
```
var-const: I don't think this check buys us anything. Constructing a moved-from pointer is very cheap, probably cheaper than having conditional logic. From the correctness perspective, I can't see any issue with setting a moved-from pointer to the same special value again.

Firestore/core/src/firebase/firestore/util/status.cc
```diff
@@ -69,11 +68,24 @@ Status& Status::WithPlatformError(std::unique_ptr<PlatformError> error) {
...
+    return;
+  }
+
+  // Set pointer value to `0x1` as the pointer is no longer useful.
+  state_ = MakeStatePtrWithMovedIndicator();
```
var-const: I'd refactor this slightly differently:
* create a function that returns the magic value:
```cpp
State* MovedFromIndicator() {
  return reinterpret_cast<State*>(0x01);
}
```
* use the `unique_ptr` constructor here directly:
```cpp
state_ = StatePtr{MovedFromIndicator()};
```
* when checking, do it like this:
```cpp
return state.get() == MovedFromIndicator();
```

Firestore/core/src/firebase/firestore/util/status.cc
```diff
@@ -69,11 +68,24 @@ Status& Status::WithPlatformError(std::unique_ptr<PlatformError> error) {
...
+}
+
 void Status::SlowCopyFrom(const State* src) {
   if (src == nullptr) {
-    state_ = nullptr;
+    state_ = MakeStatePtrWithNull();
```
var-const: Can be just `state_ = nullptr`;

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
+
+  static StatePtr MakeStatePtrWithNull() {
+    return StatePtr(nullptr, &Status::CustomStateDelete);
+  }
+
+  static StatePtr MakeStatePtrWithMovedIndicator() {
+    return StatePtr(reinterpret_cast<State*>(0x1), &Status::CustomStateDelete);
+  }
+
+  static bool IsStatePtrMoved(const StatePtr& ptr) {
```
var-const: After applying all other suggestions, this function is probably unnecessary.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
-  // a `State` structure containing the error code and message(s)
-  std::unique_ptr<State> state_;
+
+  // A `unique_ptr` with a custom deleter. If the pointer's value has been set
+  // to a special value (0x1) to indicate it is moved, the custom deleter will
+  // stop `unique_ptr` trying to destrcut from an invalid address.
+  using StatePtr = std::unique_ptr<State, std::function<void(State*)>>;
+
+  static void CustomStateDelete(State* s) {
+    if (reinterpret_cast<uintptr_t>(s) != 0x1) {
```
var-const: Following a suggestion above, this can probably be:
```cpp
if (state == MovedFromIndicator()) {
```

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
     std::unique_ptr<PlatformError> platform_error;
   };
-  // OK status has a `nullptr` state_.  Otherwise, `state_` points to
-  // a `State` structure containing the error code and message(s)
-  std::unique_ptr<State> state_;
+
+  // A `unique_ptr` with a custom deleter. If the pointer's value has been set
+  // to a special value (0x1) to indicate it is moved, the custom deleter will
+  // stop `unique_ptr` trying to destrcut from an invalid address.
+  using StatePtr = std::unique_ptr<State, std::function<void(State*)>>;
```
var-const: Question: now that you've added a constructor taking two parameters to `State`, it seems like `State` can be immutable? Can you try making this `std::unique_ptr<const State, Deleter>`? (i.e., a pointer to const `State`). If it turns out to be hard to do, just ignore.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
     // NSError* to Status and back to NSError* losslessly.
     std::unique_ptr<PlatformError> platform_error;
   };
-  // OK status has a `nullptr` state_.  Otherwise, `state_` points to
-  // a `State` structure containing the error code and message(s)
-  std::unique_ptr<State> state_;
+
+  // A `unique_ptr` with a custom deleter. If the pointer's value has been set
+  // to a special value (0x1) to indicate it is moved, the custom deleter will
+  // stop `unique_ptr` trying to destrcut from an invalid address.
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +144,49 @@ class ABSL_MUST_USE_RESULT Status {
...
+  using StatePtr = std::unique_ptr<State, std::function<void(State*)>>;
+
+  static void CustomStateDelete(State* s) {
```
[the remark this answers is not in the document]
wu-hui: But now the `Deleter` itself needs `MovedFromIndicator`, which means the field declaration needs it too.

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: There are couple of assignment to `state_->platform_error` prevent this.

Firestore/core/test/firebase/firestore/util/status_test.cc
```diff
@@ -55,13 +55,35 @@ TEST(Status, Copy) {
...
   Status b;
   b = a;
   ASSERT_EQ(a.ToString(), b.ToString());
 }
 
+TEST(Status, CanAssignToMovedFromStatus) {
+  Status a(Error::InvalidArgument, "Invalid");
+  Status b = std::move(a);
+
+  a = Status(Error::Internal, "Internal");
```
var-const: Suggestion: wrap this line in `EXPECT_NO_THROW` to make the intention a little clearer:
```cpp
EXPECT_NO_THROW({
  a = Status(Error::Internal, "Internal");
});

File: Firestore/core/src/firebase/firestore/util/status.h [hunk 15]
```diff
@@ -161,6 +198,10 @@ inline Status::State::State(const State& s)
                                                    : s.platform_error->Copy()) {
 }
 
+inline Status::State::State(Error code, std::string msg)
+    : code(std::move(code)), msg(std::move(msg)) {
```
var-const: There's no need to move `code`, moving built-ins is the same as copying them.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -161,6 +198,10 @@ inline Status::State::State(const State& s)
                                                    : s.platform_error->Copy()) {
 }
 
+inline Status::State::State(Error code, std::string msg)
```
var-const: Nit: can you make `msg` an rvalue reference (`std::string&& msg`)? `Status` is used widely enough that it _might_ be worthwhile for performance (avoids one extra move).

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +134,40 @@ class ABSL_MUST_USE_RESULT Status {
...
+  static State* MovedFromIndicator() {
+    return reinterpret_cast<State*>(0x01);
+  }
+
+  static StatePtr MakeStatePtrWithMovedIndicator() {
+    return StatePtr(MovedFromIndicator());
+  }
+
+  template <typename... Args>
+  static StatePtr MakeStatePtr(Args... args) {
```
var-const: For full compliance with `make_unique`, the arguments should be [forwarding references](https://en.cppreference.com/w/cpp/language/reference#Forwarding_references):
```cpp
static StatePtr MakeStatePtr(Args&&... args) {
```
Then you can forward them to the constructor:
```cpp
return StatePtr(new State(std::forward(args)...));
```
Forwarding makes sure the constructor receives the arguments _exactly_ as they were passed (otherwise, rvalue references easily degrade to lvalue references).

Firestore/core/src/firebase/firestore/util/status.cc
var-const: Err, #2 and #3 are still not implemented. Let me know if there's any trouble with these suggestions.

Firestore/core/src/firebase/firestore/util/status.h
var-const: How about moving this function to the `State` class itself? Then it could be called just `MakePtr` (and be invoked like `State::MakePtr`).

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +134,40 @@ class ABSL_MUST_USE_RESULT Status {
...
+    return StatePtr(new State(args...));
+  }
+
+  // OK status has a `nullptr` state_, and `IsMoved() == false`.  Otherwise,
```
var-const: Nit: remove the mention of `IsMoved` from the comment.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +134,40 @@ class ABSL_MUST_USE_RESULT Status {
...
+  // A `unique_ptr` with a custom deleter. If the pointer's value has been set
+  // to a special value (0x1) to indicate it is moved, invoking the custom
+  // deleter will be a no-op.
+  using StatePtr = std::unique_ptr<State, Deleter>;
+
+  static State* MovedFromIndicator() {
+    return reinterpret_cast<State*>(0x01);
+  }
+
+  static StatePtr MakeStatePtrWithMovedIndicator() {
```
var-const: I hope this function can go away (see one of the comments to the `.cc` file).

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -128,9 +134,40 @@ class ABSL_MUST_USE_RESULT Status {
...
+  static StatePtr MakeStatePtr(Args... args) {
+    return StatePtr(new State(args...));
+  }
+
+  // OK status has a `nullptr` state_, and `IsMoved() == false`.  Otherwise,
+  // either `state_` points to a `State` structure containing the error code
+  // and message(s), or this instance has been moved.
+  StatePtr state_;
+
+  // Tags this instance as `moved_from`.
```
var-const: Nit: `moved_from` looks like it's the name of a variable or a function, whereas I think it's actually just a concept. Replace with "moved-from" with no backticks?

wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: Done.

Firestore/core/test/firebase/firestore/util/status_test.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: Done.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -120,10 +120,28 @@ class ABSL_MUST_USE_RESULT Status {
...
+    // deleter will be a no-op.
+    using StatePtr = std::unique_ptr<State, Deleter>;
+
+    static State* MovedFromIndicator() {
+      return reinterpret_cast<State*>(0x01);
+    }
+
+    template <typename... Args>
+    static StatePtr MakePtr(Args&&... args) {
+      return StatePtr(new State(std::move(args)...));
```
var-const: This should be `std::forward<Args>`, not `std::move`.

The idea behind forwarding is to make sure the function to which the arguments are forwarded receives them _exactly_ as they were in the function that does forwarding. The more correct technical description is "preserving [value category](https://en.cppreference.com/w/cpp/language/value_category)", where "value category" basically means "is this an lvalue or an rvalue". Colloquially, it can be said that rvalues are temporaries and lvalues are non-temporaries.

In simpler terms, imagine we have two overloads of a function `Foo`, which only differ on whether they take the argument by const lvalue reference or an rvalue reference:
```cpp
void Foo(const std::string& str) { std::cout << "Lvalue overload\n"; }
void Foo(std::string&& str) { std::cout << "Rvalue overload\n"; }
```
Now we want to write a forwarding template function such that it forwards its argument to `Foo` _and_ invokes the correct overload:
```cpp
template <typename T>
void Fwd(T&& arg) {
  Foo(/*how do we pass arg to Foo here?*/);
}

std::string s;
Fwd(s); // Should invoke the lvalue overload, because `s` is not a temporary
Fwd(std::string{}); // Should invoke the rvalue overload, because the string is a temporary
```
(note: for simplicity, I make this a non-variadic template function. Whether the function is variadic or not is orthogonal to forwarding)

If we simply pass `arg` to `Foo`:
```cpp
template <typename T>
void Fwd(T&& arg) {
  Foo(arg);
}
```
the result would be that it _always_ chooses the lvalue overload, even when invoked with a temporary! Let's see how that happens. When invoked with a temporary, the type of `arg` within `Fwd` will be `std::string&&`, that is, it will be an rvalue reference -- which is correct. However, somewhat counterintuitively, an rvalue reference is by itself an lvalue -- that is, it _refers_ to an rvalue, but the reference itself is an lvalue. That is, when we use `arg` within the body of `Fwd`, it's actually an lvalue.

Unfortunately, applying `std::move` is also not a perfect solution:
```cpp
template <typename T>
void Fwd(T&& arg) {
  Foo(std::move(arg));
}
```
The problem is that `std::move` is unconditional. If previously `arg` was never moved, now the problem is that it is _always_ moved. Any arguments passed to `Fwd` will now invoke the rvalue overload:
```cpp
std::string s;
Fwd(s); // Invokes the rvalue overload
Fwd(std::string{}); // Invokes the rvalue overload
```
This is actually worse than before, because it means `s` may be moved from inside the implementation of `Foo`, even though the calling code never explicitly requested a move.

So we have one solution that never moves and one solution that always moves. This is where `std::forward` comes into play.

One way of looking at `std::forward` is that it conditionally moves. Another way to describe it is that it preserves the type of reference exactly as it was. If `arg` was an rvalue reference, `std::forward`ing `arg` will also produce an rvalue. If `arg` was an lvalue reference, `std::forward` will preserve its "lvalueness". This is exactly the behavior we want:
```cpp
template <typename T>
void Fwd(T&& arg) {
  Foo(std::forward<T>(arg));
}

std::string s;
Fwd(s); // Invokes the lvalue overload, correctly
Fwd(std::string{}); // Invokes the rvalue overload, correctly
```

See [live example](https://wandbox.org/permlink/T196w2N21ZPBtAYM). I'm happy to follow up on this if you have any questions (it's a complicated topic).

Firestore/core/src/firebase/firestore/util/status.h
var-const: Can you please replace the backticks with quotes?

var-const:
@wilhuff I'll defer the final decision to you. I think the implementation is sound at this point, so it mostly comes down to whether you consider the overall approach to be reasonable.

One potential issue is that accessing a moved-from `Status` will now result in a likely crash. OTOH, it's never valid to access a moved-from instance, so that might not be particularly important.

Firestore/core/test/firebase/firestore/util/status_test.cc
```diff
@@ -55,13 +55,35 @@ TEST(Status, Copy) {
   ASSERT_EQ(a.ToString(), b.ToString());
 }
 
+TEST(Status, Move) {
+  Status s(Status(Error::InvalidArgument, "Invalid"));
```
var-const: Can you add some more tests here?
* Check that invoking move assignment operator is okay (the current test only exercises the move constructor), with both error and non-error statuses:
```cpp
{
  Status ok;
  Status reassigned1{Error::InvalidArgument, "Foo"};
  reassigned = std::move(ok);
   // check the state of reassigned...
}
{
  Status bad{Error::InvalidArgument, "Foo"};
  Status reassigned1{Error::ResourceExhausted, "Bar"};
  reassigned = std::move(bad);
   // check the state of reassigned...
}
```
* check the pathological case where a status is move-assigned to itself:
```cpp
Status ok;
ok = std::move(ok);
Status bad{Error::InvalidArgument, "Foo"};
bad = std::move(bad);
```

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -115,9 +120,28 @@ class ABSL_MUST_USE_RESULT Status {
...
+
+    struct Deleter {
+      void operator()(const State* ptr) const;
+    };
+    // A `unique_ptr` with a custom deleter. If the pointer's value has been set
+    // to a special value (0x01) to indicate it is moved, invoking the custom
+    // deleter will be a no-op.
+    using StatePtr = std::unique_ptr<State, Deleter>;
+
+    static State* MovedFromIndicator() {
```
wilhuff: `MovedFromIndicator` is unused in the header so this can be moved into `status.cc` as a member of the anonymous namespace.

In general try to favor outright hiding over having private methods declared in classes.

Firestore/core/src/firebase/firestore/util/status.h
```diff
@@ -169,6 +203,19 @@ inline void Status::operator=(const Status& s) {
...
 
+inline Status::Status(Status&& s) noexcept : state_(std::move(s.state_)) {
+  s.SetMovedFrom();
+}
+
+inline void Status::operator=(Status&& s) noexcept {
+  state_ = std::move(s.state_);
+
+  // Mark `s` as `MovedFrom` unless it is moving to itself.
+  if (this != &s) {
```
wilhuff: Since you're already having to guard against self assignment, it would be clearer if the whole method were guarded.

Firestore/core/test/firebase/firestore/util/status_test.cc
```diff
@@ -55,13 +55,46 @@ TEST(Status, Copy) {
...
   Status b;
   b = a;
   ASSERT_EQ(a.ToString(), b.ToString());
 }
 
+TEST(Status, MoveAssign) {
+  Status ok;
+  Status reassigned{Error::InvalidArgument, "Foo"};
+  reassigned = std::move(ok);
+  ASSERT_EQ(reassigned, Status::OK());
```
wilhuff: Adding

```c++
  ASSERT_EQ(ok, Status::OK());
```
right after the line above results in a crash. That doesn't seem right.

Testing `s.ok()` should be equivalent to `s == Status::OK()`, right?

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: It is now used by `AssertStatePtr` in header. Also, it returns `State*` which is not accessible from anonymous namespace right?

Firestore/core/src/firebase/firestore/util/status.h
wu-hui: Done.

Firestore/core/test/firebase/firestore/util/status_test.cc
wu-hui: Added assert before accessing pointer, this would still crash, but with better message.

Firestore/core/src/firebase/firestore/util/status.cc
```diff
@@ -47,23 +47,24 @@ Status& Status::CausedBy(const Status& cause) {
     return *this;
   }
 
-  absl::StrAppend(&state_->msg, ": ", cause.error_message());
+  absl::StrAppend(&AssertStatePtr()->msg, ": ", cause.error_message());
```
wilhuff: `util::Status` in google3 does not fail here. In general I think we should avoid failing. I've sent you notes privately about comparisons with google3 Status.

wu-hui: Changed such that access after moved will not crash.

PR #3774, opened 2019-09-04
Add ComponentContainer tests for dependencies.

ryanwilson:
This ensures that no deadlocking occurs when fetching from the
container, and when a delete occurs.


ryanwilson:
I haven't verified that yet, but wrapping the entire `removeAllCachedInstances` method in a `dispatch_sync` block does indeed crash, so @maksymmalyhin's suggestion was warranted.

PR #3775, opened 2019-09-04
Create a GDT test app for generating events

mikehaney24:
Allows generating events that are saved to disk.


GoogleDataTransport.podspec
```diff
@@ -61,4 +61,10 @@ Shared library for iOS SDK data transport needs.
     test_spec.pod_target_xcconfig = header_search_paths
     test_spec.dependency 'GCDWebServer'
   end
+
+  # Test app specs
+  s.app_spec 'TestApp' do |app_spec|
```
paulb777: Awesome!!!

mikehaney24: Actually, hold off. I want to do something else that I think might be useful/cool.

mikehaney24: I've added support for macOS and tvOS to the test app (because it was necessary for this to work). I've had to write a somewhat x-platform delegate since macOS and iOS/tvOS have different lifecycles. Each platform also required its own storyboard, unfortunately.

PTAnotherL

PR #3781, opened 2019-09-05
Update FDL CHANGELOG

ryanwilson:
cc @dmandar


Firebase/DynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.0.4
+- [fixed] Removed references to UIWebView to comply with App Store Submission warning. (#3723)
```
paulb777: Should be 3722 for the issue.

Thanks!

ryanwilson: Whoops, good call, will follow up.

PR #3784, opened 2019-09-05
Add optional value to ExplicitNull and ServerTimestamp

wu-hui:


Firestore/Swift/Source/Codable/ExplicitNull.swift
```diff
@@ -39,7 +39,7 @@ public enum ExplicitNull<Wrapped> {
   }
 
   /// Get the `Optional` representation of `ExplicitNull`.
```
wilhuff: Comments should be phrased with an implicit "This method ..."

What do you think of rephasing this as

> Returns this value as an `Optional<Wrapped>`?

Firestore/Swift/Source/Codable/FieldValue+Encodable.swift
```diff
@@ -49,6 +49,16 @@ public enum ServerTimestamp: Codable, Equatable {
   /// value to `stamp`.
   case resolved(Timestamp)
 
+  /// Get the `Optional` representation of `ServerTimestamp`.
```
wilhuff: Similar concern about phrasing here. What do you think of this?

> Returns this value as an `Optional<Timestamp>`.
> 
> If the server timestamp is still pending, the returned optional will be `.none`. Once resolved, the returned optional will be `.some` with the resolved timestamp.

Firestore/Swift/Source/Codable/ExplicitNull.swift
wu-hui: Better, done.

Firestore/Swift/Source/Codable/FieldValue+Encodable.swift
wu-hui: Done.

PR #3786, opened 2019-09-05
Fix SIGSEGV in callback after block is nil'd

### LINKED ISSUE #3750: EXC_BAD_ACCESS (SIGSEGV) with Storage and Carthage

  * Xcode version: 11.0 Beta 6
  * Firebase SDK version: 6.7.0
  * Firebase Component: Storage
  * Component version: 6.7.0

Upon completion of uploadTasks, in about 1/3 of cases the app crashes with EXC_BAD_ACCESS (SIGSEGV)

#### Steps to reproduce:

1. Initialize a simple app with 1 table view controller
2. Sign in with Firebase Auth
3. kick off an upload task (either from a file or Data, 1MB) on a cell press
4. see it crashing once upload is complete.
NOTE: it doesn't crash on every single upload, crashes about 1/3 of the time.

#### Relevant Code:
```swift

func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
        // TODO: Expand cell, add sharing / uploading
        guard let user = Auth.auth().currentUser, let email = user.email else {
            // handle sign in experience
            return
        }

        let item = items[indexPath.row]
        let storageRef = Storage.storage().reference()
        let itemsRef = storageRef.child("items")
        let userBucketRef = itemsRef.child(email)

        let fileName = item.fileURL.lastPathComponent
        let fileRef = userBucketRef.child(fileName)


        let uploadTask = fileRef.putFile(from: item.fileURL, metadata: nil)
        { [weak self] metadata, error in
            guard error == nil else {
                let alert = UIAlertController(title: "Error", message: error?.localizedDescription,
                                              preferredStyle: .alert)
                alert.addAction(UIAlertAction(title: "OK", style: .default))
                self?.show(alert, sender: nil)
                return
            }

            // Happy path
            log(.debug, "Upload done!")
        }
        uploadTask.enqueue()
    }


```
Gist with full crash log: https://gist.github.com/supermarin/b9fe575cd95efeb905e3b771f0fd883c

paulb777:
Fix #3750

### DIFF
--- a/Firebase/Storage/FIRStorageUploadTask.m
+++ b/Firebase/Storage/FIRStorageUploadTask.m
@@ -141,8 +141,6 @@
     // Process fetches
     strongSelf.state = FIRStorageTaskStateRunning;
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Warc-retain-cycles"
     strongSelf->_fetcherCompletion = ^(NSData *_Nullable data, NSError *_Nullable error) {
       // Fire last progress updates
       [self fireHandlersForStatus:FIRStorageTaskStatusProgress snapshot:self.snapshot];
@@ -172,11 +170,12 @@
 
       [self finishTaskWithStatus:FIRStorageTaskStatusSuccess snapshot:self.snapshot];
     };
-#pragma clang diagnostic pop
 
     [strongSelf->_uploadFetcher
         beginFetchWithCompletionHandler:^(NSData *_Nullable data, NSError *_Nullable error) {
-          weakSelf.fetcherCompletion(data, error);
+          if (weakSelf.fetcherCompletion != nil) {
+            weakSelf.fetcherCompletion(data, error);
+          }
         }];
   }];
 }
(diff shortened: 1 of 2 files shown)


Firebase/Storage/FIRStorageUploadTask.m
```diff
@@ -172,11 +170,12 @@ - (void)enqueue {
...
       [self finishTaskWithStatus:FIRStorageTaskStatusSuccess snapshot:self.snapshot];
     };
-#pragma clang diagnostic pop
 
     [strongSelf->_uploadFetcher
         beginFetchWithCompletionHandler:^(NSData *_Nullable data, NSError *_Nullable error) {
-          weakSelf.fetcherCompletion(data, error);
+          if (weakSelf.fetcherCompletion != nil) {
+            weakSelf.fetcherCompletion(data, error);
+          }
```
schmidt-sebastian: I think we need to solve the same problem in line 131.  Can you check?

paulb777: I'm not sure what you mean.  Line 131 is  `weakSelf.state = FIRStorageTaskStateProgress;` I don't see any block invocations around there.

schmidt-sebastian: Yes, there are no block invocations. I assume that assigning properties on a nil object is also a no-op for Carthage. LGTM.

PR #3789, opened 2019-09-06
Fix symbol collision failure introduced by 6.8.0 release

paulb777:
The cron tests failed last night. See https://travis-ci.org/firebase/firebase-ios-sdk/jobs/581517301

With the Firebase 6.8.0 release, GoogleMobileVision is now incorporated into FirebaseML and the old GMV pod is no longer compatible.


paulb777:
cc: @temrich

PR #3790, opened 2019-09-06
Release 6.8.0 cherry pick: conflicting Dynamic Links changelog changes.

maksymmalyhin:
FDL changelog was modified in both release and master branches. The version from master was used to generate changelog for Firebase site:
- cherry pick #3781 
- cherry pick #3785


maksymmalyhin:
@paulb777 That was my initial plan, but in this case the release branch will not fully reflect what we released (the changelog will be different from the published one). WDYT?

paulb777:
I don't care a lot about keeping CHANGELOG's on the right branch, since their version info is inside the file.

In which case I'm fine either way.

PR #3793, opened 2019-09-06
Allow overriding of the default failure handler

rsgowman:


Firestore/core/src/firebase/firestore/util/hard_assert.cc
```diff
@@ -44,6 +47,33 @@ void Fail(const char* file,
 #endif
 }
 
+namespace {
+
+#if __APPLE__
```
rsgowman: This (and also below) "pollute" this file and make it now somewhat platform dependent. (This has implications on the cmake tests too.) Some alternatives:
* separate failure_handler_callback into it's own .cc file.
* Let StdioFailureHandlerCallback be the unconditional default, and have iOS override that. But we'd need to ensure that happens prior to any firebase calls. The FIRFirestore ctor is the obvious place, but there's a couple factory methods that could be called prior to that too (and who knows what will be added in the future.) We could add the set call to the ctor of a global object... but ordering with other global objects that are also statically initialized isn't defined, so it would still be *possible* to end up calling the default.

rsgowman: > +[FIRFirestore initialize]

*This* is what I was missing. Done.

Firestore/core/src/firebase/firestore/util/hard_assert.cc
```diff
@@ -21,17 +21,20 @@
...
 namespace firebase {
 namespace firestore {
 namespace util {
-namespace internal {
 
-void Fail(const char* file,
-          const char* func,
-          const int line,
-          const std::string& message) {
+ABSL_ATTRIBUTE_NORETURN void StdioFailureHandlerCallback(
```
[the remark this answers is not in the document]
rsgowman: Done.

Firestore/core/src/firebase/firestore/util/hard_assert.h
```diff
@@ -101,6 +101,43 @@
 namespace firebase {
 namespace firestore {
 namespace util {
+
+typedef void (*FailureHandlerCallback)(const char* file,
```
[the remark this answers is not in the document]
rsgowman: Both done.

Firestore/core/src/firebase/firestore/util/hard_assert.h
```diff
@@ -101,6 +101,43 @@
...
+ * be called directly, though could be called as an implementation detail of a
+ * custom failure handler. eg a failure handler that added logging might look
+ * like:
+ *
+ *   SetFailureHandler([](...) {
+ *     Log("There was an error");
+ *     DefaultFailureHandlerCallback(...);
+ *   });
+ */
+ABSL_ATTRIBUTE_NORETURN void StdioFailureHandlerCallback(
```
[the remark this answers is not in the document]
rsgowman: Comment pruned.

I've left it as a function declaration though. Not necessarily for a good reason: I left it as is mostly to avoid figuring out two separate names for the same thing (one for the function definition, and another for the pointer to it that would be declared here.) In the unlikely event someone *does* use it directly, they'll figure it out pretty quickly. :)  Feel free to push back, in which case, I'll add it in.

wilhuff:
Needs formatting. Note that the `clang-format` versions have [changed recently](https://github.com/firebase/firebase-ios-sdk/blob/master/README.md#code-formatting).

Firestore/core/src/firebase/firestore/util/hard_assert.h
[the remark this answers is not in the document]
wilhuff: The various failure handler implementations are still FailureHandlerCallback. Could you take a pass through and s/FailureHandlerCallback/FailureHandler/g?

Firestore/core/src/firebase/firestore/util/hard_assert.h
```diff
@@ -119,20 +116,13 @@ typedef void (*FailureHandlerCallback)(const char* file,
...
- * Default failure handler for non-apple platforms. This should *typically* not
- * be called directly, though could be called as an implementation detail of a
- * custom failure handler. eg a failure handler that added logging might look
- * like:
- *
- *   SetFailureHandler([](...) {
- *     Log("There was an error");
- *     DefaultFailureHandlerCallback(...);
- *   });
+ * Default failure handler for non-apple platforms. This should typically not
```
wilhuff: This is just the default failure handler. It's even the default no Apple platforms when using just C++.

Firestore/core/src/firebase/firestore/util/hard_assert.h
```diff
@@ -131,6 +131,8 @@ ABSL_ATTRIBUTE_NORETURN void DefaultFailureHandler(const char* file,
 
 namespace internal {
 
+extern FailureHandler failure_handler_callback;
```
wilhuff: Rather than exposing this global, you can make `SetFailureHandler` return the old value.

Firestore/core/test/firebase/firestore/util/hard_assert_test.cc
```diff
@@ -48,6 +48,21 @@ TEST(HardAssertTest, WithMessage) {
 }
 
 TEST(HardAssertTest, NonDefaultFailureHandler) {
+  // Used to ensure the original failure handler is restored.
+  class FailureHandlerRestorer {
```
wilhuff: I like this idea. It's probably more broadly applicable than just this one test but rolling this out to other tests is something we can do later.

Firestore/core/src/firebase/firestore/util/hard_assert.cc
```diff
@@ -46,18 +43,14 @@ ABSL_ATTRIBUTE_NORETURN void DefaultFailureHandler(const char* file,
...
 
-namespace {
-
-FailureHandler failure_handler_callback = DefaultFailureHandler;
-
-}  // namespace
-
 void SetFailureHandler(FailureHandler callback) {
-  failure_handler_callback = callback;
+  internal::failure_handler_callback = callback;
```
wilhuff: s/failure_handler_callback/failure_handler/g ? :-).

PR #3794, opened 2019-09-06
Pristine copy of JSONEncoder.swift

wilhuff:
From
https://github.com/apple/swift/commit/3391b746d9fb36d7bbf428bdc3f5f69e2bb9ac6f
pulled with the tag swift-DEVELOPMENT-SNAPSHOT-2019-09-04-a.

This is an initial import that we intend to modify heavily to implement an entirely different kind of serialization.

Note that this does not build yet and is not part of any podspec that would otherwise deliver to an end user so this can't conflict with the standard library in its current form.


jktomer:
LGTM for third-party.

PR #3797, opened 2019-09-07
Remove specialization of std::hash for Timestamp

var-const:


var-const:
My understanding is that the Unity implementation of `Timestamp` is independent and doesn't use the C++ class at all.

PR #3723, opened 2019-08-29
Remove references to UIWebView from DynamicLinks to comply with Apple…

dloic:
Linked to https://github.com/firebase/firebase-ios-sdk/issues/3722

Issue #3722, opened 2019-08-29
UIWebView usage and ITMS-90809: Deprecated API Usage


dloic:
As of Thursday, August 29, 2019, AppStoreConnect is providing a warning for uploaded applications that use UIWebViews APIs:

> Dear Developer,
> 
> We identified one or more issues with a recent delivery for your app, "XXX Dev" A.B.C (X.Y.Z). Your delivery was successful, but you may wish to correct the following issues in your next delivery:
> 
> ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.
> 
> After you’ve corrected the issues, you can use Xcode or Application Loader to upload a new binary to App Store Connect.
> 
> Best regards,
> The App Store Team

I did have a look at Firebase and Dynamic Links component is still making use of UIWebView for iOS version 7.X. I'm not sure what is the lowest iOS version Firebase wants to support, but I'm pretty sure the AppStoreConnect warning will turn soon into an error.

This will avoid people to upload apps that are making use of DynamicLinks.

  * Firebase SDK version: All
  * Firebase Component: Dynamic Links

paulb777:
Fixed by #3723

PR #3799, opened 2019-09-08
Remove UIWebViewDelegate references

paulb777:
Follow up to #3722 and #3723 for UIWebViewDelegate in addition to UIWebView

### DIFF
--- a/Firebase/DynamicLinks/CHANGELOG.md
+++ b/Firebase/DynamicLinks/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v4.0.5
+- [fixed] Removed references to UIWebViewDelegate to comply with App Store Submission warning. (#3722)
+
 # v4.0.4
 - [fixed] Removed references to UIWebView to comply with App Store Submission warning. (#3722)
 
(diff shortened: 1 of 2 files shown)


dloic:
I did search for each ref, but obviously missed this one. Sorry for this, will have a better look next time.

yadits:
Thanks!! when can we expect a release? I'm using CocoaPods

paulb777:
We'll see if we can release this sometime this week. In the meantime, the workaround would be deleting the UIWebKitDelegate references after the CocoaPods install.

dmandar:
LGTM

eleomilagrosa:
Hi I am still getting an error from the app store.
Can anyone confirm this fix works?

"ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information."

paulb777:
@eleomilagrosa This fix is only for FirebaseDynamicLinks which does not seem to be in your app.  Use `grep` to search source libraries for UIWebView and `nm` to search binary libraries to isolate the problem library.

PR #3819, opened 2019-09-10
Add RC test app.

dmandar:


FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -22,13 +22,13 @@
 #import <FirebaseCore/FIRLogger.h>
 #import <FirebaseCore/FIROptionsInternal.h>
 #import "FirebaseRemoteConfig/Sources/FIRRemoteConfigComponent.h"
-#import "FirebaseRemoteConfig/Sources/FIRRemoteConfig_Internal.h"
+#import "FirebaseRemoteConfig/Sources/Private/FIRRemoteConfig_Internal.h"
```
paulb777: Would be nice to rename to FIRRemoteConfig_Private.h

FirebaseRemoteConfig/Tests/Sample/Podfile
```diff
@@ -0,0 +1,18 @@
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
+
+target 'RemoteConfigSampleApp' do
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
+  pod 'FirebaseAnalytics'
+  pod 'FirebaseCore', :path => '../../../'
```
visumickey: Would it better to qualify the podspec instead of an open path here?

Here and below!

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/FRCLog.m
```diff
@@ -0,0 +1,76 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "FRCLog.h"
+#import <UIKit/UIKit.h>
+
+@implementation FRCLog
+
+__weak UITextView* _logView;
```
visumickey: Creating weak properties might be a better way of handling this (when compared to iVars?

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/ViewController.m
```diff
@@ -0,0 +1,501 @@
...
+#import <FirebaseRemoteConfig/FIRRemoteConfig_Private.h>
+#import <FirebaseRemoteConfig/FirebaseRemoteConfig.h>
+#import "FRCLog.h"
+
+static NSString *const FIRPerfNamespace = @"fireperf";
+static NSString *const FIRDefaultFIRAppName = @"__FIRAPP_DEFAULT";
+static NSString *const FIRSecondFIRAppName = @"secondFIRApp";
+
+@interface ViewController ()
+@property(nonatomic, strong) IBOutlet UIButton *fetchButton;
```
visumickey: nit: IBOutlets are generally weak since the view already holds a strong reference to the outlet. For eg. fetchButton is already attached with the view strongly. So fetchButton could as well be weak. Does not make a whole lot of difference. But checking if I'm missing any scenario here.

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/ViewController.m
```diff
@@ -0,0 +1,501 @@
...
+    [[FRCLog sharedInstance]
+        logToConsole:[NSString stringWithFormat:@"New %@ added %@ : %@\n", showText,
+                                                self.keyLabel.text, self.valueLabel.text]];
+
+    self.keyLabel.text = @"";
+    self.valueLabel.text = @"";
+  }
+}
+
+- (void)apply {
```
visumickey: Can we qualify the method name to something like applyConfigs or activateConfigs?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
dmandar: Done.

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/ViewController.m
dmandar: Ack. Am migrating the existing (internal) app for this CL. Will have a cleanup CL and use this input then.

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/ViewController.m
dmandar: Ack. Am migrating the existing (internal) app for this CL. Will have a cleanup CL and use this input then.

FirebaseRemoteConfig/Tests/Sample/RemoteConfigSampleApp/FRCLog.m
chliangGoogle: Also shouldn't the ivar be defined like this?
@interface FRCLog (
  UITextView* _logView;
  NSString* _textUntilTextViewSet;

)

PR #3836, opened 2019-09-12
Enable `-Werror` for Firestore XCode project

wu-hui:


wilhuff:
I've checked this out and verified that we also build cleanly under Xcode 11.

PR #3847, opened 2019-09-13
Change the order of checks in applyRemoteEvent()

schmidt-sebastian:
If there is a chance that a limbo resolution fails for a document that no longer exists, we could get into a state where we write a document into the RemoteDocumentCache with a snapshot version of zero. By re-ordering the statements, we make sure this does not happen

Port of firebase/firebase-android-sdk#802


schmidt-sebastian:
@wu-hui FYI

PR #3850, opened 2019-09-13
Adding onSnapshotsInSync (not public)

thebrianchen:
porting from [web](https://github.com/firebase/firebase-js-sdk/pull/2100) and [android](https://github.com/firebase/firebase-android-sdk/pull/778/files#diff-bf71aeb1e3faa0fe99fed19201160e24R500)


Firestore/core/src/firebase/firestore/api/firestore.mm
```diff
@@ -176,6 +179,16 @@
...
 }
 
+std::unique_ptr<ListenerRegistration> Firestore::AddSnapshotsInSyncListener(
+    std::unique_ptr<core::EventListener<Empty>> listener) {
+  EnsureClientConfigured();
+  auto async_listener = AsyncEventListener<Empty>::Create(
+      client_->user_executor(), std::move(listener));
+  client_->AddSnapshotsInSyncListener(std::move(async_listener));
+  return absl::make_unique<SnapshotsInSyncListenerRegistration>(
+      std::move(client()), std::move(async_listener));
```
wilhuff: `std::move(client())` is problematic:

  * The return type of `client()` is a const reference so `std::move()` has no effect and is misleading
  * You're mixing usage of `client_` (the field) and `client()` (the getter for that field) in the same method
  * Of course changing this to `std::move(client_)` would be disastrous (because it would give away the `client_` member, leaving the next operation in here to fail.

This should be just `client_` because we should be merely adding another reference to the client.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include <memory>
+#include <utility>
+
+#include "Firestore/core/src/firebase/firestore/api/listener_registration.h"
+#include "Firestore/core/src/firebase/firestore/core/event_listener.h"
+#include "Firestore/core/src/firebase/firestore/util/empty.h"
+#include "Firestore/core/src/firebase/firestore/util/nullability.h"
+
+NS_ASSUME_NONNULL_BEGIN
```
wilhuff: There's no Objective-C here or nullability attributes, please remove `NS_ASSUME_NONNULL_*` and the `nullability.h`.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace firebase {
+namespace firestore {
+namespace core {
+class FirestoreClient;
+}
```
wilhuff: Add `  // namespace core` after closing braces for namespaces and newlines between the opening and closing braces (so that this looks like `api`, below).

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace api {
+
+/**
+ * An internal handle that encapsulates a user's ability to request that we
```
wilhuff: Copypasta error: this isn't about stopping listening to a query.

More generally, copying the parent class's comment to the child isn't a great start. The interface specifies the contract you're describing here, so really this should discuss how this particular class implements that interface or what special behaviors it has beyond the interface.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
```diff
@@ -0,0 +1,82 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_API_SNAPSHOTS_IN_SYNC_LISTENER_REGISTRATION_H_
+#define FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_API_SNAPSHOTS_IN_SYNC_LISTENER_REGISTRATION_H_
+
+#include <memory>
+#include <utility>
```
wilhuff: `<utility>` is unused.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.mm
```diff
@@ -0,0 +1,44 @@
...
+
+void SnapshotsInSyncListenerRegistration::Remove() {
+  auto async_listener = async_listener_.lock();
+  if (async_listener) {
+    async_listener->Mute();
+    async_listener_.reset();
+  }
+
+  client_.reset();
+  client_->RemoveSnapshotsInSyncListener(async_listener);
```
wilhuff: This invocation is in the wrong order: `reset()` nulls out the pointer and then you're invoking a method on a null pointer.

Firestore/core/src/firebase/firestore/core/event_manager.h
```diff
@@ -100,6 +109,13 @@ class EventManager : public SyncEngineCallback {
   QueryEventSource* query_event_source_ = nullptr;
   model::OnlineState online_state_ = model::OnlineState::Unknown;
   std::unordered_map<core::Query, QueryListenersInfo> queries_;
+  std::unordered_set<std::shared_ptr<EventListener<util::Empty>>>
+      snapshots_in_sync_listeners_;
+
+  /**
+   * Call all global snapshot listeners that have been set.
+   */
+  void RaiseSnapshotsInSyncEvent();
```
wilhuff: Member functions should be declared before fields.

Firestore/core/src/firebase/firestore/core/event_manager.mm
```diff
@@ -67,30 +75,60 @@
...
+      }
     }
   }
+  if (raised_event) {
+    RaiseSnapshotsInSyncEvent();
+  }
+}
+
+void EventManager::RaiseSnapshotsInSyncEvent() {
+  for (const std::shared_ptr<EventListener<Empty>>& listener :
```
wilhuff: nit: since this is such a huge type, consider making this `const auto&` instead.

Firestore/core/src/firebase/firestore/core/event_manager.mm
```diff
@@ -67,30 +75,60 @@
...
+       snapshots_in_sync_listeners_) {
+    listener->OnEvent(Empty());
```
wilhuff: Though it's an almost negligible cost, it's good practice to initialize things once if possible. That is, you can hoist the construction of the instance of `Empty` out of the loop by declaring a variable like `Empty empty{};` at the method level and then passing that for each call here.

Firestore/core/src/firebase/firestore/core/event_manager.mm
```diff
@@ -109,7 +147,6 @@
   // stopListening] after an error.
   queries_.erase(found_iter);
 }
-
```
wilhuff: Please restore this newline: namespaces should be separated from the code they contain.

Firestore/core/src/firebase/firestore/core/firestore_client.mm
```diff
@@ -454,6 +456,19 @@ QuerySnapshot result(query.firestore(), query.query(), std::move(snapshot),
...
+  auto shared_this = shared_from_this();
+  worker_queue()->Enqueue([shared_this, user_listener] {
+    shared_this->event_manager_->AddSnapshotsInSyncListener(
+        std::move(user_listener));
+  });
+}
+
+void FirestoreClient::RemoveSnapshotsInSyncListener(
+    const std::shared_ptr<EventListener<Empty>>& user_listener) {
+  // do stuff
```
wilhuff: Indeed, please do :-).

Firestore/core/src/firebase/firestore/util/empty.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#ifndef FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_UTIL_EMPTY_H_
+#define FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_UTIL_EMPTY_H_
+
+namespace firebase {
+namespace firestore {
+namespace util {
+
+// An empty value.
```
wilhuff: Use a doc comment (now that this isn't a trivial implementation detail).

Firestore/core/src/firebase/firestore/core/query_listener.h
```diff
@@ -83,9 +83,9 @@ class QueryListener {
     return snapshot_;
   }
 
-  virtual void OnViewSnapshot(ViewSnapshot snapshot);
+  virtual bool OnViewSnapshot(ViewSnapshot snapshot);
```
wilhuff: `bool` and other primitive return values or argument types pretty much always need a comment that explains what they main. Please add comments to these declarations.

Firestore/Source/API/FIRFirestore+Internal.h
```diff
@@ -68,6 +69,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * snapshots are in sync with the server. Use SnapshotMetadata in the
+ * individual listeners to determine if a snapshot is from the cache or
+ * the server.
+ *
+ * @param listener A callback to be called every time all snapshot
+ * listeners are in sync with each other.
+ * @return A FIRListenerRegistration object that can be used to remove the
+ * listener.
+ */
+- (id<FIRListenerRegistration>)addSnapshotsInSyncListener:(void (^)())listener
```
wilhuff: The type of the parameter should be `void (^)(void)`.

This declaration is intended to be part of the public Objective-C API, and that means that the type must conform to C's rules, not C++.

There are subtle differences between them: C treats the empty parameter list (i.e. `()`) as parameters unspecified. In C, to specify that the function actually takes no arguments you need to declare it as `foo(void)`. C++ has strict function prototypes and does not allow you to declare a function without declaring its parameters so in C++, `()` means no parameters.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
```diff
@@ -247,6 +252,38 @@ - (void)drainQueue {
   return _currentUser;
 }
 
+//- (const std::vector<std::shared_ptr<EventListener<Empty>>>&)snapshotsInSyncListeners {
```
wilhuff: Commented out code.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
```diff
@@ -311,6 +313,28 @@ typedef std::unordered_map<firebase::firestore::auth::User,
...
+ * The number of snapshots-in-sync calls that have been run.
+ */
+@property(nonatomic, readonly) int snapshotsInSyncEvents;
+
+- (void)incrementSnapshotsInSyncEvents;
+
+- (void)resetSnapshotsInSyncEvents;
+
+/**
+ * Adds a snapshots-in-sync listener from the event manager the spec test vector.
```
wilhuff: "from" should probably be "to".

Also, this sentence is missing words. Maybe you meant "event manager *and* the spec test vector"? However, consider dropping "spec test vector" instead: I have no idea what that even is.

If you do want to keep it, describe what you're doing using higher-level terminology: aim at someone who hasn't implemented the feature describing the behavior in standalone terms. Only describe specific changes to the implementation when it's important or subtle.

If you were talking about this adding to `_snapshotsInSyncListeners`, then I'd rephrase as:

> Adds a snapshots-in-sync listener to the event manager and keeps track of it so that it can be easily removed later.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
```diff
@@ -311,6 +313,28 @@ typedef std::unordered_map<firebase::firestore::auth::User,
...
+ */
+- (void)addSnapshotsInSyncListener;
+
+/**
+ * Removes a snapshots-in-sync listener from the event manager the spec test vector.
```
wilhuff: Similar concerns to above.

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
```diff
@@ -531,6 +531,44 @@ - (void)testAddingToACollectionYieldsTheCorrectDocumentReference {
...
+    if ([events count] == 3) {
+      // We should have an initial snapshots-in-sync event, then a snapshot event
+      // for set(), then another event to indicate we're in sync again.
+      NSArray<NSString *> *expected = @[ @"snapshots-in-sync", @"doc", @"snapshots-in-sync" ];
+      XCTAssertEqualObjects(events, expected);
+      [done fulfill];
+    }
+  }];
+
+  [ref setData:@{@"foo" : @3}
```
wilhuff: This is `-writeDocumentRef:data:`.

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
```diff
@@ -531,6 +531,44 @@ - (void)testAddingToACollectionYieldsTheCorrectDocumentReference {
...
 
+- (void)testSnapshotsInSyncListenerFiresAfterListenersInSync {
+  FIRCollectionReference *coll = [self.db collectionWithPath:@"collection"];
+  FIRDocumentReference *ref = [coll addDocumentWithData:@{@"foo" : @1}];
+  NSMutableArray<NSString *> *events = [NSMutableArray array];
+
+  XCTestExpectation *gotInitialSnapshot = [self expectationWithDescription:@"gotInitialSnapshot"];
+  __block bool setupComplete = false;
+
+  [ref addSnapshotListener:^(FIRDocumentSnapshot *snapshot, NSError *error) {
```
wilhuff: Why do we need this initial snapshot listener?

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
```diff
@@ -247,6 +252,38 @@ - (void)drainQueue {
...
+  _snapshotsInSyncEvents += 1;
+}
+
+- (void)resetSnapshotsInSyncEvents {
+  _snapshotsInSyncEvents = 0;
+}
+
+- (void)addSnapshotsInSyncListener {
+  std::shared_ptr<EventListener<Empty>> eventListener = EventListener<Empty>::Create(
+      [self](const StatusOr<Empty> &v) { [self incrementSnapshotsInSyncEvents]; });
```
wilhuff: Since `v` is unused, don't name the argument.

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
thebrianchen: We want to use the snapshot listener to check that the snapshots-in-sync fires after the snapshot listener fires.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
thebrianchen: moved to the .mm file for std::move.

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
wilhuff: You're not listening to both concurrently though so this test only verifies that the snapshot listener fires and then when you start a snapshots-in-sync listener that also fires but these are effectively unrelated because we weren't listening to both at the same time.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
wilhuff: You've changed the type from `const StatusOr<Empty> &` to `const StatusOr<Empty>`. Instead of merely ignoring the argument, this is now copying and discarding the argument. Please only remove the `v` without altering the type.

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
wilhuff: > and newlines between the opening and closing braces (so that this looks like api, below)

Still not done. It should look like:

```c++
namespace core {

class FirestoreClient;

}  // namespace core
```

Firestore/core/src/firebase/firestore/api/snapshots_in_sync_listener_registration.h
```diff
@@ -18,40 +18,24 @@
...
- *
- * Getting shutdown code right is tricky so SnapshotsInSyncListenerRegistration
- * is very forgiving. It will tolerate:
- *
- *   * Multiple calls to Remove(),
- *   * calls to Remove() after we send an error,
- *   * calls to Remove() even after deleting the App in which the listener was
- *     started.
+ * stop listening to the snapshots-in-sync listener. When a user calls Remove(),
+ * ListenerRegistration will synchronously mute the listener and then send a
```
wilhuff: `ListenerRegistration` here should be `SnapshotsInSyncListenerRegistration`.

thebrianchen:
Adding Michael to check for port accuracy.

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
mikelehen: I don't understand @wilhuff's feedback or @thebrianchen's fix.

I think "You're not listening to both concurrently" was incorrect.  We _are_ listening to both concurrently.  We just wait for the first event from the snapshot listener before adding the snapshots-in-sync listener to avoid race conditions between the initial server snapshot (a nonexistent doc snapshot) and the local write.  But they both end up active concurrently.

By removing the gotInitialSnapshot expectation and waiting for it, the test is now race-y since we may or may not get the initial server snapshot before our local write is processed.  If we _don't_ get the server snapshot first, then we'll only end up getting 3 events ["snapshots-in-sync", "doc", "snapshots-in-sync"]

So I think the prior state of this test was correct and the latest change should be undone. Unless I'm missing something.

wilhuff: To clarify without arguing any specific point:

  * My initial feedback was that I didn't understand why this was necessary.
  * I was essentially prompting for a comment; a comment like the description you just gave would satisfy my concern.
  * During discussion with Brian he didn't articulate that point of view.
  * So I suggested that if this is what it takes to get a reliable sequence of events we can't expect our users to do this and to the extent we can do so, we should make the tests mirror interactions are users might make

If disabling the synchronization makes the test flaky then this isn't the fix we want.

Secondarily though I don't entirely understand why this test would be flaky. If we have one document listener and it chooses not to raise a snapshot initially, why would we get a snapshots-in-sync event?

mikelehen: The race is because:
1. We start a listener which sends a request to the backend for the current data. When that data arrives, we'll raise an event.
2. We perform a local write, which will generate a latency-compensated event (that overrides any server data).

If the backend's response to #1 gets into the AsyncQueue before the local write operation does, then we get 2 events: the backend response, then the local write.  If it doesn't, then we'll only get 1: the local write.

I would expect that as written this test will fail against prod, since the local write will win the race and so we'd only get one event, but (I guess) against the emulator, the backend response wins and we get 2 events, so the test passes.

The "fix" is to add the listener and then immediately wait for the initial snapshot (from the backend) before we proceed with the rest of the test and do our local write.  That way we know we will get one (additional) event.

We have lots of existing tests that do this (add a listener and then immediately wait for the initial empty event), e.g.: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Tests/Integration/API/FIRWriteBatchTests.mm#L208

But adding a comment to make it clearer isn't a bad idea. Something like:

> Wait for the initial event from the backend so that we know we'll get exactly one snapshot event for our local write below.

thebrianchen: Ah sorry about that Gil. I was struggling to explain the concept behind Michael's comment in my [web pr](https://github.com/firebase/firebase-js-sdk/pull/2100#discussion_r319221356). Surprisingly, I tried running the integration test with synchronization disabled 100 times in a loop several times, and I didn't get any flakes. 

> Secondarily though I don't entirely understand why this test would be flaky. If we have one document listener and it chooses not to raise a snapshot initially, why would we get a snapshots-in-sync event? 

In what situations do we not get a snapshot initially? I thought the snapshot listener will always fire once after it has been added. The snapshots-in-sync listener is scheduled to be fired immediately after it has been added for the first time. On Android, I ran into the following situation (different that what Michael posted):
1. snapshots-in-sync listener fires after it has been added.
2. snapshot listener fires its initial event, causing the snapshots-in-sync listener to fire again.
3. As a result, we get ["snapshots-in-sync", "doc", snapshots-in-sync"] before we write to the doc again.

Given this, is the right approach to leave the synchronization check in place to prevent flakiness?

wilhuff: At this point I think you and @mikelehen have worked out how this should work and I'm convinced that the nature of this test could cause flakes (even if we haven't observed them).

My initial comment remains: it's not obvious to an uninitiated reader why the test has to wait for the first document snapshot. Please add a comment that describes the race you're avoiding.

PR #3867, opened 2019-09-16
Add support for the FLL endpoint

mikehaney24:
All data format is with regard to CCT, so renaming wasn't very necessary except in tests.

- Ensures that test event generation creates files, not uses existing ones
- Implements prioritizer and uploader for FLL server
  - Main diff from CCT is needing gzipped data and api key
- With the sqlite refactor, this will be more consolidated


mikehaney24:
Largely, this a copy-paste from the CCT implementation since the data format, prioritization, and uploading are largely identical. The exceptions are noted in the PR description.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLPrioritizer.m
```diff
@@ -0,0 +1,188 @@
...
+  dispatch_sync(_queue, ^{
+    NSSet<GDTCORStoredEvent *> *logEventsThatWillBeSent;
+    // A high priority event effectively flushes all events to be sent.
+    if ((conditions & GDTCORUploadConditionHighPriority) == GDTCORUploadConditionHighPriority) {
+      package.events = self.events;
+      return;
+    }
+
+    // If on wifi, upload logs that are ok to send on wifi.
+    if ((conditions & GDTCORUploadConditionWifiData) == GDTCORUploadConditionWifiData) {
```
paulb777: Why bitwise-and tests, but no bitwise-or assigns?

mikehaney24: The upload conditions are computed and assigned in GDT. This is just the same thing that GDTCCT's prioritizer does.

mikehaney24:
Also, sorry for the immense size of this PR--if it's easier, I could close this, break this up by file or something, and send it back out.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLPrioritizer.m
paulb777: I don't see bitwise or's in GDT either. Am I missing where the conditions are merged together?

mikehaney24: No you're right. Right now, upload conditions are mutually exclusive--I just don't think that'll always be the case. There used to be a condition called "High Priority" that was ORed in when a high priority log was waiting to be sent, and that may be the case again soon with the sqlite refactor.

https://github.com/firebase/firebase-ios-sdk/blob/6be7fc0d7dd047dc26dce1937e55bac2d9050e4d/GoogleDataTransport/GDTCORLibrary/GDTCORUploadCoordinator.m#L134

paulb777: OK thanks. SGTM

PR #3880, opened 2019-09-18
ZipBuilder: move ManifestReader to a separate package to be reusable.

maksymmalyhin:
The `ManifestReader` is going to be used by other release tools.

Swift proto files updated with: `protoc Release.proto --swift_opt=Visibility=Public  --swift_out=./ `


ZipBuilder/Package.swift
```diff
@@ -21,12 +21,20 @@ import PackageDescription
 
 let package = Package(
   name: "ZipBuilder",
+  products: [
+      .library(name: "ManifestReader", targets: ["ManifestReader"]),
```
ryanwilson: I don't think we need to vend a library here - this should be an `executable` as well.

maksymmalyhin: Removed `products` section

maksymmalyhin:
Updated README.

PR #3882, opened 2019-09-18
Fix GULMutableDictionary keyed subscript methods

paulb777:
b/141259138

### DIFF
--- a/GoogleUtilities/Network/GULMutableDictionary.m
+++ b/GoogleUtilities/Network/GULMutableDictionary.m
@@ -77,17 +77,11 @@
 }
 
 - (void)setObject:(id)obj forKeyedSubscript:(id<NSCopying>)key {
-  dispatch_async(_queue, ^{
-    self->_objects[key] = obj;
-  });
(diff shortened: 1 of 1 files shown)


GoogleUtilities/Network/GULMutableDictionary.m
```diff
@@ -76,14 +76,18 @@ - (NSUInteger)count {
   return count;
 }
 
-- (id)objectForKeyedSubscript:(id<NSCopying>)key {
-  // The method this calls is already synchronized.
-  return [self objectForKey:key];
+- (void)setObject:(id)obj forKeyedSubscript:(id<NSCopying>)key {
+  dispatch_async(_queue, ^{
```
maksymmalyhin: I would prefer to keep old implementation here that reuses the method implementation rather than copying it. Despite that the implementation is minimal now, modifying it in the future will be easier.

paulb777: After discussion agreed to keep separate implementations because of the different requirements for nil keys.

PR #3889, opened 2019-09-19
Add a product to the SwiftPM manifest.

ryanwilson:
It's named `ReleasePackager` instead of `ZipBuilder` since we should change the name now that it packages Zip and Carthage builds.


ryanwilson:
> Should the directory name change as well?

The target is still called `ZipBuilder` for now, and that's what the directory structure is based off of. I do plan on doing a large restructure to name things better, and that directory name would change along with that.

PR #3897, opened 2019-09-20
Add cross-platform support for creating Executors

wilhuff:
... and use it consistently in tests and production code that doesn't
have to potentially wrap a user-supplied dispatch_queue_t.


Firestore/core/test/firebase/firestore/testutil/time_testing.h
```diff
@@ -27,6 +27,10 @@ namespace testutil {
 
 using time_point = std::chrono::system_clock::time_point;
 
+/** Returns the current time, in milliseconds. */
+std::chrono::time_point<std::chrono::steady_clock, std::chrono::milliseconds>
```
var-const: I think this was previously defined in `async_tests_util.h` (called `now`), should the previous version of this function now be deleted?

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -0,0 +1,41 @@
...
+
+using util::AsyncQueue;
+using util::Executor;
+
+std::unique_ptr<util::Executor> ExecutorForTesting(const char* name) {
+  std::string label = absl::StrCat("firestore.testing.", name);
+  return Executor::CreateSerial(label.c_str());
+}
+
+std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
```
var-const: Question: previously, tests would name the executor differently. I presume that removing this ability is intentional, but just checking to be sure.

Firestore/core/test/firebase/firestore/testutil/time_testing.h
wilhuff: A subsequent PR removes async_tests_util.h altogether in the interest of making these utilities more broadly available that just the AsyncQueue and Executor tests.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
wilhuff: Yes. The names previously didn't matter. The only cases where the names matter are when there's an AsyncQueue and another executor simulating a user notification queue. All that matters then is that the names aren't the same.

Removing this removes significant cruft from the tests.

PR #3899, opened 2019-09-20
Add Expectations and Async/Await for testing

wilhuff:
Also rework existing tests to take advantage of them.

I've carved this out of a port of existing Objective-C tests to C++. Those tests typically have more than one expectation and so the existing `TestWithTimeoutMixin` does not work. Manually working with std::promise in all these cases seemed unattractive since it's verbose and error prone.

I believe this also fixes a core problem with the prior approach of using `std::async` to create a background task in the test, wherein the destructor for the future would hang if the underlying task hangs. This now uses a packaged_task along with a detached thread instead. This is arguably less efficient, but it decouples the lifetime of the future from the backing task, which makes it possible for a test to timeout without hanging the whole suite.

`Async` and `Await` could arguably be free functions, but I've chosen to make them members of the `AsyncTest` mixin to force users of these primitives to actually use the mixin, which automatically adds a googletest `ScopedTrace` without any macro magic.

In any case async tasks and expectations can uniformly be Awaited, making it straightforward to mix and match these pieces as required to make a test work.


Firestore/core/test/firebase/firestore/testutil/async_testing.h
```diff
@@ -46,6 +52,84 @@ std::unique_ptr<util::Executor> ExecutorForTesting(const char* name);
...
+constexpr auto kTimeout = std::chrono::seconds(5);
+
+/**
+ * An expected outcome of an asynchronous test.
+ */
+class Expectation {
+ public:
+  Expectation();
+
+  /** Marks this expectation as fulfilled. */
```
var-const: One thing that seems worthwhile to add to the comment is what happens if `Fulfill` is called more than once.

Firestore/core/test/firebase/firestore/testutil/async_testing.h
```diff
@@ -46,6 +52,84 @@ std::unique_ptr<util::Executor> ExecutorForTesting(const char* name);
...
+  /**
+   * Returns a callback function, that when invoked, fullfills the expectation.
+   *
+   * The returned function has a lifetime that's independent of the Expectation
+   * that created it.
+   */
+  std::function<void()> AsCallback() const;
+
+  /**
+   * Returns a `future` that represents the completion of this Expectation.
```
var-const: Can you add to the comment whether this function can be called more than once?

Firestore/core/test/firebase/firestore/testutil/async_testing.h
```diff
@@ -46,6 +52,84 @@ std::unique_ptr<util::Executor> ExecutorForTesting(const char* name);
...
+  std::shared_ptr<std::promise<void>> promise_;
+};
+
+/**
+ * A mixin that supplies utilities for safely writing asynchronous tests.
+ */
+class AsyncTest {
+ public:
+  AsyncTest()
+      : trace_("Test case name",
```
var-const: Very optional: if you'd like, you can probably move this into an in-class initializer and just have `= default` here.

Firestore/core/test/firebase/firestore/testutil/async_testing.h
```diff
@@ -46,6 +52,84 @@ std::unique_ptr<util::Executor> ExecutorForTesting(const char* name);
...
+  void Await(const std::future<void>& future,
+             std::chrono::milliseconds timeout = kTimeout);
+
+  /**
+   * Waits for the promise to become ready. The future associated with the
+   * promise is consumed via `promise.get_future()`.
+   *
+   * Fails the current test if the timeout occurs.
+   */
+  void Await(std::promise<void>& promise,  // NOLINT(runtime/references)
```
var-const: This is, strictly speaking, against style guide, though I'm certainly not complaining -- feel free to leave as is.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
   return std::make_shared<AsyncQueue>(ExecutorForTesting("worker"));
 }
 
+// MARK: - Expectation
+
+Expectation::Expectation() : promise_(std::make_shared<std::promise<void>>()) {
```
var-const: Note: you could move this into an in-class initializer and have an `= default` constructor.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
...
+
+std::function<void()> Expectation::AsCallback() const {
+  // Additional copy required because putting `promise_` in the capture list
+  // is not allowed. Generalized capture in C++14 would allow direct
+  // initialization.
+  auto promise = promise_;
+  return [promise] { promise->set_value(); };
+}
+
+std::future<void> Expectation::get_future() const {
```
var-const: This is a destructive operation, so it probably shouldn't be `const` (even though the language rules allow modifying a value through a pointer inside a const function). `std::promise::get_future` isn't `const`.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
...
+  return promise_->get_future();
+}
+
+// MARK: - AsyncTest
+
+std::future<void> AsyncTest::Async(std::function<void()> action) {
```
var-const: Optional: it seems like this function and all functions below can be `const`.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
...
+  std::packaged_task<void()> task(std::move(action));
+  auto future = task.get_future();
+
+  std::thread thread(std::move(task));
+  thread.detach();
+  return future;
+}
+
+void AsyncTest::Await(const std::future<void>& future,
+                      const std::chrono::milliseconds timeout) {
```
var-const: Optional: remove `const` from `timeout` (I'm all for having local variables, but I think we have previously agreed to not have them, so this is inconsistent with the rest of the codebase). Applies below as well.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
...
+  std::future_status result = future.wait_for(timeout);
+  if (result == std::future_status::ready) {
+    return;
+  }
+
+  ADD_FAILURE();
```
var-const: Question: is it possible/worthwhile to add some message here? Or is the fact that the failure is within `Await` already clear enough?

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
```diff
@@ -36,6 +40,62 @@ std::shared_ptr<util::AsyncQueue> AsyncQueueForTesting() {
...
+}
+
+// MARK: - AsyncTest
+
+std::future<void> AsyncTest::Async(std::function<void()> action) {
+  std::packaged_task<void()> task(std::move(action));
+  auto future = task.get_future();
+
+  std::thread thread(std::move(task));
+  thread.detach();
```
var-const: A detached thread has its own problems, though I agree it's the lesser of two evils in this case.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
wilhuff: I've changed this around to just use `std::shared_future` instead of `std::future` on the grounds that efficiency doesn't matter here and the move-only nature of `std::future` harms the design considerably. With `std::shared_future` this becomes just an accessor declared in the header and no longer has any consequences as far as reuse goes.

Firestore/core/test/firebase/firestore/testutil/async_testing.cc
var-const: Note: this is unchanged. Once again, it's a purely optional suggestion -- just making sure this is intentional.

Issue #2665, opened 2019-03-28
Non-portable path warnings from Protobuf pod


OkiRules:
Please update for Swift 5 support.

Thanks!

maksymmalyhin:
@OkiRules Thank you for your request. For most of the components Swift 5 should be supported now. Would you be able to provide examples of the issues you have with Swift 5, please?

OkiRules:
<img width="266" alt="Screen Shot 2019-03-29 at 8 23 03 pm" src="https://user-images.githubusercontent.com/18746541/55222616-77b1f480-5260-11e9-966f-d5009e154574.png">

paulb777:
This is an Apple issue tracked at https://openradar.appspot.com/radar?id=6091887509241856. More discussion at https://github.com/protocolbuffers/protobuf/issues/3218.

We can investigate if we can find any better workarounds. In the meantime, it's safe to ignore the warnings.

paulb777:
I'm going to change the title since the issue is not related to Swift 5.

paulb777:
One workaround is adding `pod 'Protobuf', :inhibit_warnings => true` to the Podfile.

paulb777:
Sent https://github.com/protocolbuffers/protobuf/pull/6190

iatman:
It worked, it was enough to change the letter of the folder: Pods/Protobuf/objectivec/google/protobuf  (protobuf -> Protobuf)

paulb777:
protocolbuffers/protobuf#6464 in progress to fix

devkokodev:
I'm trying to install 'Firebase/Messaging' it install protobuf etc. I'm using swift 5, deployment ios 11 and i have this issues:
![Screen Shot 2019-08-08 at 10 44 49 AM](https://user-images.githubusercontent.com/53897025/62688744-beabb580-b9c9-11e9-857d-f881d86c1fe4.png)
![Screen Shot 2019-08-08 at 10 45 15 AM](https://user-images.githubusercontent.com/53897025/62688745-beabb580-b9c9-11e9-98c8-8709e84812cd.png)
![Screen Shot 2019-08-08 at 10 49 55 AM](https://user-images.githubusercontent.com/53897025/62689028-485b8300-b9ca-11e9-882c-f5ee776e80fa.png)


I've tried (cmd + shift + k) delete derived data, pod install, pod update, nothing solved this issue.
Any help?

paulb777:
[...]

PR #3916, opened 2019-09-24
Use at least Protobuf 3.9.2

paulb777:
To ensure that we get the fix in Protobuf 3.9.2 for #3851

### DIFF
--- a/FirebaseABTesting.podspec
+++ b/FirebaseABTesting.podspec
@@ -42,7 +42,7 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
   }
   s.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
   s.dependency 'FirebaseCore', '~> 6.1'
-  s.dependency 'Protobuf', '~> 3.9'
+  s.dependency 'Protobuf', '~> 3.9', '>= 3.9.2'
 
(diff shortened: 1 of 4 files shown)


wilhuff:
Why not specify at least 3.9.2?

paulb777:
Because we want to pick up 3.10.0 automatically when it comes out - which should fix #2665

wilhuff:
In that case, we want multiple version specs:

```
  s.dependency 'Protobuf', '~> 3.9', '>= 3.9.2'
```

I had version 3.9.0 installed in my project and got this this error on pod install after adding that to Firestore:

```
[!] CocoaPods could not find compatible versions for pod "Protobuf":
  In snapshot (Podfile.lock):
    Protobuf (= 3.9.0, ~> 3.0, ~> 3.1)

  In Podfile:
    FirebaseFirestore (from `../../`) was resolved to 1.5.0, which depends on
      Protobuf (>= 3.9.2, ~> 3.9)

It seems like you've changed the constraints of dependency `Protobuf` inside your development pod `FirebaseFirestore`.
You should run `pod update Protobuf` to apply changes you've made.
```

cruisediary:
I'm waiting for next SDK release include this PR :D

Issue #3524, opened 2019-08-07
Support iOS 13's new lifecycle paradigm


mikehaney24:
iOS 13 introduces a new lifecycle paradigm with iOS 13.

According to @DavidPhillipOster, a full iOS 13 app will just get the UIScene lifecycle notifications, running on either iPhone or iPad. Before iOS 13, it will just get the UIApplication lifecycle notifications, but if it has a SceneDelegate, but does not have the dictionary in its Info.plist it will get both sets of notifications on iOS 13.

In GDT, getting multiple background/foregrounding calls is ok, so I'm solving it with #3520

Other resources:
https://stackoverflow.com/a/56508769/361918

Reach out to me (haneym@) for additional context and example fixes if needed.

maksymmalyhin:
TODO:
- [ ] background/foreground notifications: I think it is actually a good case where we should use `UI Scene...` notifications, so a new message can be presented when a new window is opened. @christibbs WDYT?
- [ ] background/foreground notifications: update notifications to the scene specific in view controllers like `FIDBaseRenderingViewController`
- [ ] consider changing URL handling and posting to scene specific vs app specific
- [ ]

PR #3927, opened 2019-09-25
Update FIAM for iOS 13 Scene lifecycle paradigm

christibbs:
Two key fixes:
- Throw both `UIScene` and `UIApplication` lifecycle notifications
- Update `UIWindow` creation logic to initialize based on a foregrounded scene (if iOS 13 is available)

Fixes https://github.com/firebase/firebase-ios-sdk/issues/3524 with the exception around concerns with opening URL from a scene. From testing, though, current URL opening logic still works as intended.

### DIFF
--- a/Firebase/InAppMessaging/Analytics/FIRIAMClearcutLogStorage.m
+++ b/Firebase/InAppMessaging/Analytics/FIRIAMClearcutLogStorage.m
@@ -92,6 +92,14 @@ static NSString *const kEventExtensionJson = @"extension_js";
                                              selector:@selector(appWillBecomeInactive)
                                                  name:UIApplicationWillResignActiveNotification
                                                object:nil];
+    
+    if (@available(iOS 13.0, *)) {
+      [[NSNotificationCenter defaultCenter] addObserver:self
+                                               selector:@selector(appWillBecomeInactive)
+                                                   name:UISceneWillDeactivateNotification
+                                                 object:nil];
+    }
+    
     @try {
       [self loadFromCachePath:cachePath];
     } @catch (NSException *exception) {
--- a/Firebase/InAppMessaging/Analytics/FIRIAMClearcutLogStorage.m
+++ b/Firebase/InAppMessaging/Analytics/FIRIAMClearcutLogStorage.m
@@ -92,13 +92,14 @@ static NSString *const kEventExtensionJson = @"extension_js";
                                              selector:@selector(appWillBecomeInactive)
                                                  name:UIApplicationWillResignActiveNotification
                                                object:nil];
-
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
     if (@available(iOS 13.0, *)) {
       [[NSNotificationCenter defaultCenter] addObserver:self
                                                selector:@selector(appWillBecomeInactive)
                                                    name:UISceneWillDeactivateNotification
                                                  object:nil];
     }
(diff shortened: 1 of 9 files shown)


Firebase/InAppMessaging/Flows/FIRIAMActivityLogger.m
```diff
@@ -106,7 +106,14 @@ - (instancetype)initWithMaxCountBeforeReduce:(NSInteger)maxBeforeReduce
                                              selector:@selector(appWillBecomeInactive)
                                                  name:UIApplicationWillResignActiveNotification
                                                object:nil];
-
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    if (@available(iOS 13.0, *)) {
+      [[NSNotificationCenter defaultCenter] addObserver:self
```
ryanwilson: Are there situations where `appWillBecomeInactive` could be called twice, i.e. could `UIApplicationWillResignActiveNotification` and `UISceneWillDeactivateNotification` both be fired for the same app close?

Firebase/InAppMessagingDisplay/FIDRenderingWindowHelper.m
```diff
@@ -24,7 +24,16 @@ + (UIWindow *)UIWindowForModalView {
   static dispatch_once_t onceToken;
 
   dispatch_once(&onceToken, ^{
-    UIWindowForModal = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    if (@available(iOS 13.0, *)) {
+      UIWindowScene *foregroundedScene = [[self class] foregroundedScene];
```
ryanwilson: Could this ever be `nil` on iOS 13?

christibbs: Yes, could be `nil` if this is called there are no foregrounded scenes. In that case it should be a no-op. Though, now that you mention it I should force the nil case to make sure nothing weird happens.

Firebase/InAppMessaging/Analytics/FIRIAMClearcutLogStorage.m
```diff
@@ -92,6 +92,15 @@ - (instancetype)initWithExpireAfterInSeconds:(NSInteger)expireInSeconds
                                              selector:@selector(appWillBecomeInactive)
                                                  name:UIApplicationWillResignActiveNotification
                                                object:nil];
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    if (@available(iOS 13.0, *)) {
+      [[NSNotificationCenter defaultCenter] addObserver:self
+                                               selector:@selector(appWillBecomeInactive)
```
mikehaney24: ideally, nsnotification observers have a single argument, the notification.

Firebase/InAppMessaging/Flows/FIRIAMActivityLogger.m
mikehaney24: Same thing for becoming active.

Firebase/InAppMessagingDisplay/FIDRenderingWindowHelper.m
christibbs: Tested this, it's a no-op as expected.

PR #3933, opened 2019-09-26
Fix unit testing crash with localization on

### LINKED ISSUE #3928: Crash in GULLogBasic when running test plan with Localization Screenshots option

    file a Github issue.
    * If this is a __feature request__ make sure the issue title starts with "FR:".
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

  * Xcode version: 11.0 (11A420a)
  * Firebase SDK version: 6.8.0
  * Firebase Component: GoogleUtilities
  * Component version: 6.2.5

When an Xcode 11 test plan has the "Localization Screenshots" option enabled, Google Utilities will crash the app shortly after launch in attempting to log that analytics have started.  This does not occur if the test plan changes the "Localization Screenshots" option to off.

```
(lldb) bt
* thread #5, queue = 'APMMonitorQueue', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
    frame #0: 0x00000001298c7f22 UIAccessibility`-[NSPlaceholderStringAccessibility initWithFormat:locale:arguments:] + 55
  * frame #1: 0x0000000108805821 GoogleUtilities`GULLogBasic(level=GULLoggerLevelNotice, service=@"[Firebase/Analytics]", forceLog=NO, messageCode=@"I-ACS023007", message=@"Analytics v.60101000 started", args_ptr=0x0000000000000000) at GULLogger.m:164:22
    frame #2: 0x0000000107a5dbc6 TestPlanFail`-[APMASLLogger logMessage:logTag:messageCode:withLogLevel:] + 208
    frame #3: 0x0000000107a358b4 TestPlanFail`__44-[APMMonitor logToConsole:message:logLevel:]_block_invoke + 91
    frame #4: 0x0000000109511888 libdispatch.dylib`_dispatch_call_block_and_release + 12
    frame #5: 0x00000001095127f9 libdispatch.dylib`_dispatch_client_callout + 8
    frame #6: 0x0000000109518566 libdispatch.dylib`_dispatch_lane_serial_drain + 707
    frame #7: 0x0000000109518f9c libdispatch.dylib`_dispatch_lane_invoke + 388
    frame #8: 0x000000010952306c libdispatch.dylib`_dispatch_workloop_worker_thread + 626
    frame #9: 0x0000000109989763 libsystem_pthread.dylib`_pthread_wqthread + 290
    frame #10: 0x00000001099895c3 libsystem_pthread.dylib`start_wqthread + 15
```

Screen Shot 2019-09-25 at 3 38 57 PM
Screen Shot 2019-09-25 at 3 42 56 PM

#### Steps to reproduce:

1) Create an app that uses Firebase & give it a test plan
2) In the test plan, modify the "Localization Screenshots" option to On
3) Attempt to run your UI tests and note how the app will crash out on GULLogBasic

This can be worked around by raising the logging level of GoogleUtilities logging so it doesn't log at all.

#### Relevant Code:

Attached is a sample project that shows this off.  The GoogleService-Info.plist has been removed - please update the bundle ID, run "pod install", and add your own GoogleService-Info.plist to run this.

[TestPlanFail.zip](https://github.com/firebase/firebase-ios-sdk/files/3655120/TestPlanFail.zip)

If you switch to the test tab and try to run the tests, you'll see the crash.

ryanwilson:
For whatever reason, `GULLogBasic` crashes when running a test plan with Localization Screenshots enabled. I can't figure out exactly why this throws but even assigning `logMsg = message;` triggers the same crash, leading me to believe that `message` is invalid at that point (even though the debugger can print the value).

Fixes #3928.  Thanks again @abotkin-cpi for the sample project, made it easy to track down!

I avoided copying in the regular path since we don't copy to begin with and it's working as expected.

### DIFF
--- a/GoogleUtilities/Logger/GULLogger.m
+++ b/GoogleUtilities/Logger/GULLogger.m
@@ -161,7 +161,12 @@ void GULLogBasic(GULLoggerLevel level,
                                                                     range:messageCodeRange];
   NSCAssert(numberOfMatches == 1, @"Incorrect message code format.");
 #endif
-  NSString *logMsg = [[NSString alloc] initWithFormat:message arguments:args_ptr];
+  NSString *logMsg;
+  if (args_ptr == NULL) {
+    logMsg = [message copy];
+  } else {
+    logMsg = [[NSString alloc] initWithFormat:message arguments:args_ptr];
+  }
   logMsg = [NSString stringWithFormat:@"%s - %@[%@] %@", sVersion, service, messageCode, logMsg];
   dispatch_async(sGULClientQueue, ^{
     asl_log(sGULLoggerClient, NULL, (int)level, "%s", logMsg.UTF8String);
--- a/GoogleUtilities/Logger/GULLogger.m
+++ b/GoogleUtilities/Logger/GULLogger.m
@@ -163,7 +163,7 @@ void GULLogBasic(GULLoggerLevel level,
 #endif
   NSString *logMsg;
   if (args_ptr == NULL) {
-    logMsg = [message copy];
+    logMsg = message;
   } else {
     logMsg = [[NSString alloc] initWithFormat:message arguments:args_ptr];
   }
--- a/GoogleUtilities/CHANGELOG.md
+++ b/GoogleUtilities/CHANGELOG.md
@@ -5,6 +5,7 @@
 # 6.3.0
 - GULSecureCoding introduced. (#3707)
 - Mark unused variables. (#3854)
+- Fix crash in GULLogBasic. (#3928)
 
 # 6.2.5
 - Remove test-only method and update tests to include Catalyst. (#3544)
--- a/GoogleUtilities/CHANGELOG.md
+++ b/GoogleUtilities/CHANGELOG.md
@@ -1,11 +1,11 @@
 # 6.3.1
 - Fix GULMutableDictionary keyed subscript methods. (#3882)
 - Update Networking to receive data for POST requests. (#3940)
+- Fix crash in GULLogBasic. (#3928)
 
 # 6.3.0
 - GULSecureCoding introduced. (#3707)
 - Mark unused variables. (#3854)
-- Fix crash in GULLogBasic. (#3928)
 
 # 6.2.5
 - Remove test-only method and update tests to include Catalyst. (#3544)


GoogleUtilities/Logger/GULLogger.m
```diff
@@ -161,7 +161,12 @@ void GULLogBasic(GULLoggerLevel level,
                                                                     range:messageCodeRange];
   NSCAssert(numberOfMatches == 1, @"Incorrect message code format.");
 #endif
-  NSString *logMsg = [[NSString alloc] initWithFormat:message arguments:args_ptr];
+  NSString *logMsg;
+  if (args_ptr == NULL) {
```
wilhuff: `args_ptr` is not guaranteed to be a pointer type, which makes this comparison invalid when it's not. We even have a messy declaration in the header that accounts for this:

```c
// On 64-bit simulators, va_list is not a pointer, so cannot be marked nullable
// See: http://stackoverflow.com/q/29095469
#if __LP64__ && TARGET_OS_SIMULATOR || TARGET_OS_OSX
                        va_list args_ptr
#else
                        va_list _Nullable args_ptr
#endif
```

At a high level it's semantically broken to treat this value as nullable--`va_list` is an opaque type. We're going to have no end of messy workarounds in here if we don't just fix the usage. Callers should be required to pass this value.

ryanwilson: Ahh thanks for pointing that out.

The strange thing is, even calling 
```
NSString *messageCopy = [message copy];
NSString *logMsg = [[NSString alloc] initWithFormat:messageCopy arguments:args_ptr];
```

fails at `[message copy]` without this check. I'm wondering if it's a coincidence, but seems strange.

Any thoughts on what else it could be?

The reason `NULL` is sometimes passed in is some SDKs call `FIRLogBasic` directly instead of using the macros `GULLog{Error,Warning,Notice,Info,Debug}`, which is another problem.

wilhuff: It's not hard to call `FIRLogBasic` properly. Firestore does so here: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/util/log_apple.mm#L56.

Anyone passing `NULL` is just doing it wrong.

Looking through our code the only caller passing null is in `GULNetwork_logWithLevel:messageCode:message:contexts:`. We should just fix the caller.

If you're crashing with that method not in the stack trace we have a different problem.

ryanwilson: Makes sense, thank you for the recommendation! I'll take a look again at other closed source usages and try to make a decision. I think as we switch over to `os_log` soon enough it should warrant a re-thinking of how we're doing logging anyways, so maybe the least-effort solution here would be best.

wilhuff: If you have a repro and this successfully fixes it on a simulator, and 32- and 64-bit hardware then it's fine for now. It's the kind of fix that we can't just implicitly trust because it's exploiting unspecified behavior.

PR #3849, opened 2019-09-13
Replay FirestoreEncoder changes

wilhuff:
Replay the changes made in the old non-conforming location and migrate
to using the FirestoreEncoder from the new location.

Changes:

f14979510 (Marek Gilbert, 2 hours ago)
   Migrate to encoders in third_party/FirestoreEncoder

330ef74f2 (wu-hui, 7 days ago)
   Add @DocumentId equivalent to Swift Codable (#3762)

   * Commit with failed optional test case.

   * Integration Test and cleanup

   * swiftformat

   * better doc formatting

   * addressing comments

   * rename to SelfDocumentID

   * Addressing Comments

   * fix comment

e39eb4013 (Hui-Wu, 3 months ago)
   Allow firestore api to work with Codable objects + ExplicitNull +
   Integration tests (#3261)

   * Convinience api for codables

   * Integration Tests

   * ExplicitNull implementation

14c8d1bec (Morgan Chen, 3 months ago)
   Fix token fetch not invoking callback for unchanged token (#3251)

78b18282e (Hui-Wu, 3 months ago)
   FieldValue and DocumentReference codable support. (#3231)

   * working serverTimestamp

   * working FieldValue. Still need: document ref and ability to Eq of
   FieldValue

   * DocumentReference implemented and testing works

   * clean up header search path

   * improving FieldValue enum comments

   * run through formatting

   * addressing comments

   * formatted with swiftformat 0.35.7

   * add missing bool->BOOL change

   * undo unintended swift version change

   * remove unwanted proj file change

8ecec2b82 (Hui-Wu, 3 months ago)
   Firestore swift Codable support (#3198)

   FirestoreEncoder/Decoder and Timestamp/GeoPoint codable support.

30e493730 (Marek Gilbert, 2 hours ago)
   Rename JSONEncoder; prep for split

8285f2015 (Marek Gilbert, 7 days ago)
   Pristine copy of JSONEncoder.swift

   From 
   https://github.com/apple/swift/commit/3391b746d9fb36d7bbf428bdc3f5f69e2bb9ac6f 
   pulled with the tag swift-DEVELOPMENT-SNAPSHOT-2019-09-04-a.

PR #3935, opened 2019-09-26
Merge Firestore Codable changes to master

wilhuff:
This takes the changes in #3794 and #3849 and merges them (finally) to master.


wilhuff:
As in #3849, the other commits are from @wu-hui's old github handle.

PR #3943, opened 2019-09-28
Update spec tests

schmidt-sebastian:
Update iOS spec tests, up to and including https://github.com/firebase/firebase-js-sdk/pull/2205


schmidt-sebastian:
This won't work until we update the Spec test runner to understand the new format.

Issue #3948, opened 2019-09-30
Fix the FIAM Display unit test failure for Xcode 11


paulb777:
- Fix the Xcode 11 unit test failure
- Remove CI workarounds added in #3944 to .travis.yml and scripts/build.sh

See https://travis-ci.org/firebase/firebase-ios-sdk/jobs/591176709

    ✗ testModalViewWithNarrowImage, Failed to synthesize event: Failed to scroll to visible (by AX action) Button, identifier: 'message-action-button', label: 'Take action', error: Error kAXErrorCannotComplete performing AXAction 2003 on element AX element pid: 4222, elementOrHash.elementID: 140654379669232.67

PR #3952, opened 2019-09-30
Make FIAM tests more robust in Travis

paulb777:
- Split FIAM from FIAM display to separate stages
- Split iOS and iPad to separate stages
- Stop running static library tests from xcodebuild. `pod lib lint` tests are sufficient for that
- Require FIAM tests to pass for overall run to pass
- Xcode 11 build issue remains. See #3948

There's now enough time in each stage for travis_retry to eliminate flakes.


paulb777:
Investigating why the new FIAMDisplay travis tests succeeds running on the branch but fails as a PR ....

PR #3955, opened 2019-09-30
testing only

paulb777:


paulb777:
Lesson learned - missing cases in if_changed.sh give very cryptic failures when run on travis that show up in PRs, but not branches.

PR #3966, opened 2019-10-01
Add the ability to define an initialization priority when registering a library

jasonhu-g:
There are useful cases where it will be good to define some ordering to the initialization of registered Firebase libraries. 

An example is to have Crashlytics startup before other Firebase SDKs (not implemented yet) so it can collect crashes from other SDK's initializations as well as just being able to collect crashes earlier in app initialization.

### DIFF
--- a/Example/Core/Tests/FIRAppInitializationPriorityTests.m
+++ b/Example/Core/Tests/FIRAppInitializationPriorityTests.m
@@ -1,0 +1,118 @@
+// Copyright 2019 Google
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


wilhuff:
There's already a mechanism for ordering between SDKs: dependencies. The topological sort should already handle this. It sounds to me like there's just missing (optional) dependencies in the graph from SDK authors that want crash reports.

As it stands this does not handle outlier cases very well, where for example component A is Low priority and B is high priority, with a dependency on A. In your PR, priority wins and the initialization order is wrong. You either need to handle this by donating B's higher priority to A (by traversing dependencies and bumping priority to at least the minimum of the root in the chain) or by declaring this to be equivalent to a cycle in the graph and treat it as an error.

jasonhu-g:
Abandoning this change in favor of using eager instantiation if we need a bump in ordering.

PR #3974, opened 2019-10-02
Fix main thread crashing issue, reenable tests.

### LINKED ISSUE #3973: Re-enable GDT Background Tests
As part of #3896 the background flag was changed, and subsequently harder to test. Two tests were temporarily disabled, but should be re-enabled ASAP.

ryanwilson:
Checking `applicationState` while not on the  main thread results in a crash. Revert to using notifications and a flag to track this, but with a single source of truth.

Fixes #3973.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -570,4 +570,3 @@ jobs:
 branches:
   only:
     - master
-    - rw-gdt-background-mainthread
--- a/.travis.yml
+++ b/.travis.yml
@@ -570,3 +570,4 @@ jobs:
 branches:
   only:
     - master
+    - rw-gdt-background-mainthread
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -37,6 +37,13 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
 #endif  // TARGET_OS_IOS
 }
 
+@interface GDTCORApplication () {
+  /** Private flag to match the existing `readonly` public flag. */
+  BOOL _isRunningInBackground;
+}
+
+@end
+
 @implementation GDTCORApplication
 
 + (void)load {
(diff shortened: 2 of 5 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -37,6 +37,13 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
 #endif  // TARGET_OS_IOS
 }
 
+@interface GDTCORApplication () {
+  /** Private flag to match the existing `readonly` public flag. */
+  BOOL _isRunningInBackground;
```
maksymmalyhin: We don't write it from different threads (only read), but it may make sense to add `volatile` just in case.

maksymmalyhin: or declare it as an atomic property (which is a bit more common way to declare thread-safe variables)

GoogleDataTransport/GDTCORTests/Lifecycle/GDTCORLifecycleTest.m
```diff
@@ -105,10 +108,8 @@ - (void)DISABLED_testBackgrounding {
       },
       5.0);
 
-  // TODO(#3973): This notification no longer triggers the `isRunningInBackground` flag. Find
-  // another way to test it.
   NSNotificationCenter *notifCenter = [NSNotificationCenter defaultCenter];
-  [notifCenter postNotificationName:kGDTCORApplicationDidEnterBackgroundNotification object:nil];
+  [notifCenter postNotificationName:UIApplicationDidEnterBackgroundNotification object:nil];
```
mikehaney24: Why change from the abstracted notification to an iOS-specific one?

ryanwilson: We previously used the abstract one because it triggered each classes' `foreground/background` flags, but now since the flag moved it's set before the abstracted notification is fired. Meaning: the abstract notification no longer triggers the flag to change, only the iOS specific one does.

.travis.yml
```diff
@@ -570,3 +570,4 @@ jobs:
 branches:
   only:
     - master
+    - rw-gdt-background-mainthread
```
paulb777: Remove

ryanwilson: Thanks, done with the latest commit.

PR #3977, opened 2019-10-02
Update nanopb from 0.3.9.2 to 0.3.9.3

mikehaney24:
No appreciable change.


wilhuff:
Doesn't this require a corresponding change to the nanopb CocoaPod?

mikehaney24:
> Doesn't this require a corresponding change to the nanopb CocoaPod?

It shouldn't since the generated sources are identical. I've been using a newer nanopb to generate these sources from the beginning (on accident).

paulb777:
The [podspec is still at 0.3.901](https://github.com/CocoaPods/Specs/blob/master/Specs/6/1/e/nanopb/0.3.901/nanopb.podspec.json#L3).

Is there a reason for updating or going out of sync?

mikehaney24:
I just set a periodic reminder to update nanopb. I'm actually ahead of the cocoapod and have been using 0.3.9.2 to generate since the creation of GDT. Since the actual code didn't change at all, we can consider this to just future-proof GDT for a while.

paulb777:
That seems reasonable as long as there are also no differences with what gets generated by 0.3.9.1.  Have you verified that?

mikehaney24:
Yeah, this PR is the difference is generated sources. The only thing it did was change the nanopb version number in the file comment.

wilhuff:
I’m not sure I agree with the rationale of this change. While the generated sources may not have changed, the runtime has and the change gives the appearance that you’ve tested with 0.3.9.3.

Given that there’s no effective change and you haven’t tested I’d advocate for deferring this. I don’t feel strongly though, so I won’t block the PR.

mikehaney24:
I'll hold off on updating in the future to make sure the cocoapod matches
the generation version.

PR #3978, opened 2019-10-02
FirebaseInstallations FID generation: don't cut UUID

maksymmalyhin:
- copy last 4 bits of UUID to the FID prefix (they were cut in the previous implementation).

### DIFF
--- a/FirebaseInstallations/Source/Library/FIRInstallationsItem.m
+++ b/FirebaseInstallations/Source/Library/FIRInstallationsItem.m
@@ -72,22 +72,21 @@
 
   NSData *uuidData = [NSData dataWithBytes:uuidBytes length:16];
 
-  uint8_t prefix = 0b01110000;
-  NSMutableData *fidData = [NSMutableData dataWithBytes:&prefix length:1];
+  uint8_t uuidLast4Bits = uuidBytes[15] & 0b00001111;
+
+  // FID first 4 bits must be `0111`. The last 4 UUID bits will be cut later to form a proper FID. To keep 16 random bytes we copy these last 4 UUID to the FID 1st byte.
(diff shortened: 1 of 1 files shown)


maksymmalyhin:
cc @andirayo, @ankitaj224

FirebaseInstallations/Source/Library/FIRInstallationsItem.m
```diff
@@ -72,22 +72,22 @@ + (NSString *)generateFID {
 
   NSData *uuidData = [NSData dataWithBytes:uuidBytes length:16];
```
paulb777: Make a define for the 16 - UUID_LENGTH perhaps?

FirebaseInstallations/Source/Library/FIRInstallationsItem.m
```diff
@@ -72,22 +72,22 @@ + (NSString *)generateFID {
...
 
-  uint8_t prefix = 0b01110000;
-  NSMutableData *fidData = [NSMutableData dataWithBytes:&prefix length:1];
+  uint8_t uuidLast4Bits = uuidBytes[15] & 0b00001111;
```
paulb777: UUID_LENGTH-1

chliangGoogle: acronym should be capitalized. same as below

maksymmalyhin: I defined a local var `NSUInteger UUIDLength = sizeof(uuid_t);`, so the UUIDBytes[UUIDLength - 1] would not crash even if actual `UUID` length changes.

PR #3980, opened 2019-10-02
Fix GDT monkey test path

mikehaney24:
Also fix unwrapping an optional in the test app


mikehaney24:
@ryanwilson What would you think of me enabling these monkey tests for travis cronjobs?

ryanwilson:
> @ryanwilson What would you think of me enabling these monkey tests for travis cronjobs?

SGTM as long as they aren't flaky. You may need to update master to get Travis to pass here as well.

PR #3982, opened 2019-10-02
Allow for multiple uploads during CCT Test.

ryanwilson:
Hopefully this should reduce flakiness.


GoogleDataTransportCCTSupport/GDTCCTTests/Integration/GDTCCTIntegrationTest.m
```diff
@@ -79,13 +83,22 @@ - (void)setUp {
   }
 }
 
+- (void)tearDown {
+  if (self.uploadObserver) {
+    [[NSNotificationCenter defaultCenter] removeObserver:self.uploadObserver];
```
mikehaney24: should you also nil-out the observer?

ryanwilson: Nice catch, meant to do that! Thanks.

Issue #3778, opened 2019-09-05
Failed to open DB: Internal: Failed to open LevelDB database [FSTFirestoreClient initializeWithUser:settings:] 


avinashkumarmishra:
### Describe your environment
Xcode version: Version 10.3(10G8)
Firebase Component: Firestore
Using Firebase (6.5.0)
Using FirebaseAnalytics (6.0.4)
Using FirebaseAuthInterop (1.0.0)
Using FirebaseCore (6.1.0)
Using FirebaseFirestore (1.4.2)
Using FirebaseInstanceID (4.2.2)

### Describe the problem
An app on a certain device was crashed by HARD_ASSERT in this code repeatedly. I couldn't find a way to handle or recover this error except uninstalling the app once to delete persistent data.

<img width="997" alt="Screen Shot 2019-09-05 at 12 01 53 PM" src="https://user-images.githubusercontent.com/13217186/64317365-193e2e80-cfd5-11e9-86bb-71d8ac5876c2.png">

### [READ] Step 1: Are you in the right place?

```
  Fatal Exception: NSInternalInconsistencyException
0  (Missing)                      0x1b992698c (Missing)
1  (Missing)                      0x1b8aff9f8 (Missing)
2  (Missing)                      0x1b9830bc0 (Missing)
3  Fernsnpetals                   0x102a60180 -[FSTFirestoreClient initializeWithUser:settings:] + 1406 (string:1406)
4  Fernsnpetals                   0x102a6303c std::__1::__function::__func<-[FSTFirestoreClient initWithDatabaseInfo:settings:credentialsProvider:userExecutor:workerQueue:]::$_2, std::__1::allocator<-[FSTFirestoreClient initWithDatabaseInfo:settings:credentialsProvider:userExecutor:workerQueue:]::$_2>, void ()>::operator()() + 1406 (string:1406)
5  Fernsnpetals                   0x102a29448 firebase::firestore::util::AsyncQueue::ExecuteBlocking(std::__1::function<void ()> const&) + 921 (atomic:921)
6  Fernsnpetals                   0x102a3884c firebase::firestore::util::internal::DispatchAsync(NSObject<OS_dispatch_queue>*, std::__1::function<void ()>&&)::$_0::__invoke(void*) + 1862 (functional:1862)
7  (Missing)                      0x1b93657d4 (Missing)
8  (Missing)                      0x1b9340dec (Missing)
9  (Missing)                      0x1b934192c (Missing)
[...]

PR #3984, opened 2019-10-02
Fix race condition in FIRComponentContainer instance creation

### LINKED ISSUE #3967: Failed to open DB: LOCK: already held by process
Full error:
```
Fatal Exception: NSInternalInconsistencyException
Failed to open DB: Internal: Failed to open LevelDB database at 
/var/mobile/Containers/Data/Application/D5EA17E8-B506-4E84-84C2-
10A92D7A0AA4/Documents/firestore/__FIRAPP_DEFAULT/xxx/main: LevelDB
 error: IO error: lock /var/mobile/Containers/Data/Application/D5EA17E8-B506-4E84-84C2-
10A92D7A0AA4/Documents/firestore/__FIRAPP_DEFAULT/xxx/main/LOCK: 
already held by process
```

Xcode version: 10.3
Using Firebase (6.6.0)
Using FirebaseAnalytics (6.1.0)
Using FirebaseAnalyticsInterop (1.3.0)
Using FirebaseAuth (6.2.2)
Using FirebaseAuthInterop (1.0.0)
Using FirebaseCore (6.2.0)
Using FirebaseCoreDiagnostics (1.0.1)
Using FirebaseCoreDiagnosticsInterop (1.0.0)
Using FirebaseDatabase (6.0.0)
Using FirebaseDynamicLinks (4.0.2)
Using FirebaseFirestore (1.4.3)
Using FirebaseInstanceID (4.2.3)
Using FirebaseMessaging (4.1.2)
Using FirebaseStorage (3.4.0)
Using FirebaseUI (8.0.4)

Can't reproduce this problem. I am not deleting FIRApp instances, I'm only accessing the database via `Firestore.firestore()` or `[[FIRFirestore firestore]` which should provide the same instance back. The only thing I can think of is `[FIRApp configure];` getting called twice with thread race, but that is supposed to be thread-safe as well.

Found this in another issue #3778 from Will. @wilhuff Is there any other scenarios which might cause this error?

> @ChristianVinterly your case is different. Somehow you've managed to open Firestore more than once in a single process and the second one failed to open its database because the first one still had the database locked.
> 
> Normally, the Firestore SDK keeps a map of active instances so that [FIRFirestore firestore] will return the same instance. The only way I can think of that you might have encountered this is if you deleted FIRApp and then managed to create a new one and a new Firestore instance before the old one shut down.
> 
> If you're at all able to reproduce this, please create a new issue with the details of what you're doing--this is unrelated to the "Operation Not Permitted" issue going on in this discussion.

wilhuff:
Reported in #3967. Users sometimes seeing a crash when multiple threads
concurrently try to create Firestore instances. By inspection I found
that there's a race here where two threads can check for presence in the
cache, see nothing, proceed to creation, both store a value in the cache,
and both return their copies. This would lead to two Firestores being
created for a single App.

This replaces `dispatch_sync` on a serial queue with good old
`@synchronized` blocks. `dispatch_sync` will self-deadlock if called
from a thread already on the queue while `@synchronized` allows
recursion.

No new tests are added because several tests like
`testDependencyDoesntBlock` already cover this case.

Fixes #3967. Could possibly address #3778, though the error doesn't seem like it should arise from this.

### DIFF
--- a/Firebase/Core/FIRComponentContainer.m
+++ b/Firebase/Core/FIRComponentContainer.m
@@ -99,8 +99,9 @@ static NSMutableSet<Class> *sFIRComponentRegistrants;
           (component.instantiationTiming == FIRInstantiationTimingEagerInDefaultApp &&
            [app isDefaultApp]);
       if (shouldInstantiateEager || shouldInstantiateDefaultEager) {
-        @synchronized (self) {
-          [self instantiateInstanceForProtocol:component.protocol withBlock:component.creationBlock];
+        @synchronized(self) {
+          [self instantiateInstanceForProtocol:component.protocol
+                                     withBlock:component.creationBlock];
         }
       }
     }
@@ -153,7 +154,7 @@ static NSMutableSet<Class> *sFIRComponentRegistrants;
   NSString *protocolName = NSStringFromProtocol(protocol);
 
   id cachedInstance;
-  @synchronized (self) {
+  @synchronized(self) {
     cachedInstance = self.cachedInstances[protocolName];
     if (!cachedInstance) {
       // Use the creation block to instantiate an instance and return it.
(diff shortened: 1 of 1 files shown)


paulb777:
Any objection to cherry-picking to M57 (6.10.0)?

ryanwilson:
> Any objection to cherry-picking to M57 (6.10.0)?

Merging and cherry-picking, feel free to raise concerns if you have any!

Issue #3924, opened 2019-09-25
didReceiveRegistrationToken not called on first app run


willk37:
### Describe your environment

  * Xcode version: 11.0 (11A420a)
  * Firebase SDK version: 6.8.1
  * Firebase Component: Messaging
  * Component version: 4.1.4

### [REQUIRED] Step 3: Describe the problem

The MessaginDelegate function `messaging(_ messaging: Messaging, didReceiveRegistrationToken fcmToken: String)` is not called on the first run a fresh installed build. I've also confirmed this is happening in Firebase SDK version 6.9.0. 

This was working fine on Firebase SDK version 6.2.0/FirebaseMessaging version 4.0.2 but recently updated to 6.8.1/4.1.4.

morganchen12:
Is this reproducible in a sample project like the [Messaging quickstart](https://github.com/firebase/quickstart-ios/tree/master/messaging)?

ryanwilson:
Another piece of valuable information: does this happen on iOS 13 only, or older devices as well?

willk37:
@morganchen12 After modifying the linked sample project to replicate the conditions in our app (we disable swizzling) I was not able to reproduce the issue there. But I was still able to reproduce it in our app after completely stripping it down to just setting up push notifications. I also noticed some relevant console messages from that:

```
2019-09-26 10:53:11.855378-0500 Tasty[7040:1958829] 6.8.1 - [Firebase/InstanceID][I-IID003009] Failed to fetch default token Error Domain=NSURLErrorDomain Code=-999 "cancelled" UserInfo={NSErrorFailingURLStringKey=https://device-provisioning.googleapis.com/checkin, NSLocalizedDescription=cancelled, NSErrorFailingURLKey=https://device-provisioning.googleapis.com/checkin}

2019-09-26 10:53:11.855877-0500 Tasty[7040:1958829] 6.8.1 - [Firebase/InstanceID][I-IID003007] Failed to retrieve the default FCM token after 0 retries
```

@ryanwilson I've also been able to reproduce this on iOS 12.4.1/Xcode 10.3 and our QA team reproduced it on iOS 12.1

morganchen12:
Can you share your stripped-down project? I'll try to repro and identify the root cause.

willk37:
[...]

PR #3988, opened 2019-10-03
Cherrypick for #3984 along with CHANGELOG update.

ryanwilson:


Firebase/Core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v6.3.1 -- M57
+- [fixed] Fixed race condition in component container. (#3967)
```
paulb777: Do we think this also fixes #3924?

ryanwilson: Tentatively, yes - good call. Will add it

PR #3989, opened 2019-10-03
Simplify update-versions script with --use-json

paulb777:
This should work, but the M58 release lead should confirm.


paulb777:
Adding `--sources` would be a good follow up. We need to verify it does what's expected.

PR #3991, opened 2019-10-03
Add QueryData.with() helpers

schmidt-sebastian:
Port of https://github.com/firebase/firebase-js-sdk/pull/2148

### DIFF
--- a/Firestore/core/src/firebase/firestore/local/query_data.cc
+++ b/Firestore/core/src/firebase/firestore/local/query_data.cc
@@ -85,10 +85,16 @@ QueryData QueryData::Invalid() {
                    SnapshotVersion(SnapshotVersion::None()), {});
 }
 
-QueryData QueryData::Copy(SnapshotVersion snapshot_version,
-                          ByteString resume_token,
-                          model::ListenSequenceNumber sequence_number) const {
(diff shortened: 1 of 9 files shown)


Firestore/core/src/firebase/firestore/local/query_data.cc
```diff
@@ -85,10 +85,16 @@ QueryData QueryData::Invalid() {
...
-                          ByteString resume_token,
-                          model::ListenSequenceNumber sequence_number) const {
+QueryData QueryData::WithSequenceNumber(
+    model::ListenSequenceNumber sequence_number) const {
   return QueryData(Query(query_), target_id_, sequence_number, purpose_,
+                   snapshot_version_, resume_token_);
+}
+
+QueryData QueryData::WithResumeToken(
+    nanopb::ByteString resume_token,
```
wilhuff: There are already `using` statements for `ByteString` and `SnapshotVersion`. You can use these unqualified in the .cc file (but not in the header).

Also consider adding a `using` statement for `ListenSequenceNumber` and `TargetId` and shortening those too.

schmidt-sebastian: Done (added using statements for all the types you mentioned).

PR #3992, opened 2019-10-03
Port remaining serializer methods

wu-hui:


Firestore/Protos/protos/google/firestore/v1/firestore.options
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# In proto3 mode, Nanopb doesn't allow distinguishing between unset fields and
+# fields having default values, even for non-primitive types. Using the
+# workaround suggested in
+# https://github.com/nanopb/nanopb/issues/255#issuecomment-291842903
+
+# cause is not set if everything is OK, serializer need to be able to tell
```
var-const: Nit: s/need/needs/

Firestore/core/src/firebase/firestore/remote/serializer.h
```diff
@@ -346,6 +347,9 @@ class Serializer {
   std::unique_ptr<remote::WatchChange> DecodeDocumentChange(
       nanopb::Reader* reader,
       const google_firestore_v1_DocumentChange& change) const;
+  std::vector<model::TargetId> DecodeTargetIdArray(nanopb::Reader* reader,
```
var-const: Nit: can you move this function down below this group and make it its own group (so that it is preceded and followed by an empty line)?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
-  // TODO(varconst): implement.
-  // - (MutationResult)decodedMutationResult:(GCFSWriteResult *)mutation
-  //     commitVersion:(const SnapshotVersion &)commitVersion;
-  return MutationResult{{}, {}};
+  // NOTE: Deletes don't have an update_time, use commit_version instead.
+  SnapshotVersion version =
+      write_result.has_update_time
+          ? DecodeSnapshotVersion(reader, write_result.update_time)
+          : commit_version;
+  absl::optional<std::vector<FieldValue>> transform_results;
```
var-const: Nit: can you add an empty line before this line? (I know it was a wall of text in the original)

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  if (write_result.transform_results_count > 0) {
+    transform_results = std::vector<FieldValue>{};
+    for (pb_size_t i = 0; i < write_result.transform_results_count; i++) {
+      transform_results->push_back(
+          DecodeFieldValue(reader, write_result.transform_results[i]));
+    }
+  }
+  return MutationResult(std::move(version), std::move(transform_results));
```
var-const: Nit: can you add an empty line before this line?

var-const: Nit: moving `SnapshotVersion` is no more efficient than copying it, I'd suggest removing the `move` from `version`.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
 
 std::unordered_map<std::string, std::string>
 Serializer::EncodeListenRequestLabels(const QueryData& query_data) const {
-  // TODO(varconst): implement.
-  // - (nullable NSMutableDictionary<NSString *, NSString *> *)
-  //     encodedListenRequestLabelsForQueryData: (const QueryData &)queryData;
-  return {};
+  auto value = EncodeLabel(query_data.purpose());
+
+  std::unordered_map<std::string, std::string> result(1);
```
var-const: Note: this will always allocate a single entry in the map, even if `value` is empty. How about:
```cpp
auto value = EncodeLabel(query_data.purpose());
if (value.empty()) {
  return {};
}

// ...Like before, but without checking for `empty` again...
```

Also, the Objective-C code preallocates capacity, whereas the current implementation creates a map with a single "blank" element. Calling `reserve` would be more similar, but I don't think it matters at all here, so I'd suggest just default-constructing the map.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  if (!value.empty()) {
+    result["goog-listen-tags"] = value;
```
var-const: Nit: you can `move` `value` here.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
-  // - (nullable NSString *)encodedLabelForPurpose:(QueryPurpose)purpose;
-  return {};
+  switch (purpose) {
+    case QueryPurpose::Listen:
+      return "";
+    case QueryPurpose::ExistenceFilterMismatch:
+      return "existence-filter-mismatch";
+    case QueryPurpose::LimboResolution:
+      return "limbo-document";
+    default:
```
var-const: Nit: when a switch is exhaustive, rather than adding a `default` case, it's better to add the `UNREACHABLE` macro _after_ the switch (see examples in this file). The advantage is that the compiler will issue a warning if there are unhandled cases in the switch, but that warning is canceled out by a `default` case.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+    case google_firestore_v1_ListenResponse_document_delete_tag:
+      return DecodeDocumentDelete(reader, watch_change.document_delete);
+
+    case google_firestore_v1_ListenResponse_document_remove_tag:
+      return DecodeDocumentRemove(reader, watch_change.document_remove);
+
+    case google_firestore_v1_ListenResponse_filter_tag:
+      return DecodeExistenceFilterWatchChange(reader, watch_change.filter);
+
+    default:
```
var-const: Same nit here: it's better to use `UNREACHABLE`.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  // is a read_time set and it applies to all targets (i.e. the list of targets
+  // is empty). The backend is guaranteed to send such responses.
+  if (listen_response.which_response_type !=
+      google_firestore_v1_ListenResponse_target_change_tag) {
+    return SnapshotVersion::None();
+  }
+  if (listen_response.target_change.target_ids_count != 0) {
+    return SnapshotVersion::None();
+  }
+  return DecodeSnapshotVersion(reader, listen_response.target_change.read_time);
```
var-const: Nit: can you add an empty line before this line?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
 std::unique_ptr<WatchChange> Serializer::DecodeTargetChange(
     nanopb::Reader* reader,
     const google_firestore_v1_TargetChange& change) const {
-  // TODO(varconst): implement.
-  // - (std::unique_ptr<WatchChange>)decodedTargetChangeFromWatchChange:
-  //     (GCFSTargetChange *)change;
-  return {};
+  WatchTargetChangeState state =
+      DecodeTargetChangeState(reader, change.target_change_type);
+  auto target_ids =
```
var-const: Nit: in C++, if you use the `vector` constructor that takes two iterators, then this helper function becomes pretty unnecessary:
```cpp
std::vector<TargetId> target_ids(change.target_ids, change.target_ids + change.target_ids_count);
```

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+      DecodeTargetIdArray(reader, change.target_ids, change.target_ids_count);
+  ByteString resume_token = ByteString(change.resume_token);
```
var-const: Nit: you don't need copy initialization here:
```cpp
ByteString resume_token(change.resume_token);
```

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+      return WatchTargetChangeState::NoChange;
+    case google_firestore_v1_TargetChange_TargetChangeType_ADD:
+      return WatchTargetChangeState::Added;
+    case google_firestore_v1_TargetChange_TargetChangeType_REMOVE:
+      return WatchTargetChangeState::Removed;
+    case google_firestore_v1_TargetChange_TargetChangeType_CURRENT:
+      return WatchTargetChangeState::Current;
+    case google_firestore_v1_TargetChange_TargetChangeType_RESET:
+      return WatchTargetChangeState::Reset;
+    default:
```
var-const: Same nit about `UNREACHABLE`.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  SnapshotVersion version =
+      DecodeSnapshotVersion(reader, change.document.update_time);
+  HARD_ASSERT(version != SnapshotVersion::None(),
+              "Got a document change with no snapshot version");
+  // The document may soon be re-serialized back to protos in order to store it
+  // in local persistence. Memoize the encoded form to avoid encoding it again.
+  Document document(std::move(value), key, version, DocumentState::kSynced,
+                    change.document);
+
+  auto updated_target_ids =
```
var-const: Same comment here: personally, I'd remove `DecodeTargetIdArray` and just use the vector constructor.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -117,6 +120,25 @@ QueryData CreateQueryData(absl::string_view str) {
...
+    case WatchChange::Type::Document:
+      return static_cast<const DocumentWatchChange&>(lhs) ==
+             static_cast<const DocumentWatchChange&>(rhs);
+    case WatchChange::Type::TargetChange:
+      return static_cast<const WatchTargetChange&>(lhs) ==
+             static_cast<const WatchTargetChange&>(rhs);
+    case WatchChange::Type::ExistenceFilter:
+      return static_cast<const ExistenceFilterWatchChange&>(lhs) ==
+             static_cast<const ExistenceFilterWatchChange&>(rhs);
+    default:
```
var-const: Same nit about `UNREACHABLE`.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -117,6 +120,25 @@ QueryData CreateQueryData(absl::string_view str) {
   return CreateQueryData(Query(str));
 }
 
+bool operator==(const WatchChange& lhs, const WatchChange& rhs) {
+  if (lhs.type() != rhs.type()) {
+    return false;
+  }
+  switch (lhs.type()) {
```
var-const: There's already a helper function like this [here](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Tests/Remote/FSTSerializerBetaTests.mm#L116-L136). I'd suggest just copying it over.

var-const: Please add an empty line before this line.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
 std::unique_ptr<WatchChange> Serializer::DecodeDocumentDelete(
     nanopb::Reader* reader,
     const google_firestore_v1_DocumentDelete& change) const {
-  // TODO(varconst): implement.
-  // -(std::unique_ptr<WatchChange>)decodedDocumentDelete:
-  //     (GCFSDocumentDelete*)change;
-  return {};
+  DocumentKey key = DecodeKey(reader, change.document);
+  // Note that version might be unset in which case we use
+  // SnapshotVersion::None()
```
var-const: Ultranit: please add a full stop at the end of the sentence.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  SnapshotVersion version =
+      change.has_read_time ? DecodeSnapshotVersion(reader, change.read_time)
+                           : SnapshotVersion::None();
+  NoDocument document(key, version, /* has_committed_mutations= */ false);
+
+  std::vector<TargetId> removed_target_ids = DecodeTargetIdArray(
```
var-const: Same comment about `vector` constructor here.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
-  // TODO(varconst): implement.
-  // -(std::unique_ptr<WatchChange>)decodedDocumentChange:
-  //     (GCFSDocumentChange*)change;
-  return {};
+  ObjectValue value = ObjectValue::FromMap(DecodeFields(
+      reader, change.document.fields_count, change.document.fields));
+  DocumentKey key = DecodeKey(reader, change.document.name);
+  SnapshotVersion version =
+      DecodeSnapshotVersion(reader, change.document.update_time);
+  HARD_ASSERT(version != SnapshotVersion::None(),
```
var-const: Hmm, perhaps you should also do a check like `change.document.has_update_time` here? Can you please raise this in the chat?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1492,101 +1493,202 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+              "Got a document change with no snapshot version");
+  // The document may soon be re-serialized back to protos in order to store it
+  // in local persistence. Memoize the encoded form to avoid encoding it again.
+  Document document(std::move(value), key, version, DocumentState::kSynced,
+                    change.document);
```
var-const: This is pretty dangerous.

The constructor of `Document` [takes its last argument](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/model/document.h#L69) as `absl::any`, which in many regards is similar to `void*`. Then the value of `any` will be unconditionally cast to an Objective-C type ([here](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Local/FSTLocalSerializer.mm#L90)). `any` should throw if it's being cast to a different type than what it currently contains -- I'm surprised tests didn't catch this.

I'd suggest using the overload of the constructor without the last argument and adding a TODO to revisit this while porting the local serializer (where it's used).

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1424,6 +1460,115 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
   ExpectRoundTrip(model, proto);
 }
 
+TEST_F(SerializerTest, EncodesListenRequestLabels) {
+  core::Query q = Query("docs");
+
+  std::vector<
+      std::pair<QueryPurpose, std::unordered_map<std::string, std::string>>>
+      purposeToLabel = {
```
var-const: Nit: snake_case.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1424,6 +1460,115 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+          {QueryPurpose::Listen, {}},
+          {QueryPurpose::LimboResolution,
+           {{"goog-listen-tags", "limbo-document"}}},
+          {QueryPurpose::ExistenceFilterMismatch,
+           {{"goog-listen-tags", "existence-filter-mismatch"}}},
+      };
+
+  for (const auto& p : purposeToLabel) {
+    QueryData model(q, 1, 0, p.first, SnapshotVersion::None(), Bytes(1, 2, 3));
```
var-const: Note: there's an overload of `QueryData` constructor that takes fewer arguments, you can omit everything after `p.first`.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1424,6 +1460,115 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
   ExpectRoundTrip(model, proto);
 }
 
+TEST_F(SerializerTest, EncodesListenRequestLabels) {
+  core::Query q = Query("docs");
+
+  std::vector<
```
var-const: Can you use `std::map<QueryPurpose, std::unordered_map<...>>`? (I'm using `map` just because it's shorter and efficiency doesn't matter here)

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -465,6 +492,15 @@ class SerializerTest : public ::testing::Test {
     EXPECT_EQ(model.query(), actual_model);
   }
 
+  void ExpectDeserializationRoundTrip(const WatchChange& model,
+                                      const v1::ListenResponse& proto) {
+    auto actual_model = Decode<google_firestore_v1_ListenResponse>(
+        google_firestore_v1_ListenResponse_fields,
+        std::mem_fn(&Serializer::DecodeWatchChange), proto);
+
+    ASSERT_TRUE(model == *actual_model);
```
var-const: 1. We generally prefer `EXPECT` to `ASSERT`.
2. How about `EXPECT_EQ(model, *actual_model)` instead of doing the comparison manually?

Firestore/core/src/firebase/firestore/remote/serializer.cc
wu-hui: I think you are right. The contract here in all platforms are if there is not `update_time`, fail the SDK.

The annoying thing is `Document` is also used for encoding in `LocalSerializer`, and `has_update_time` now has to be on top of `update_time` there, I guess we will just have to live with that.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -156,6 +181,11 @@ class SerializerTest : public ::testing::Test {
     ExpectDeserializationRoundTrip(key, absl::nullopt, read_time, proto);
   }
 
+  void ExpectWatchChangeDeserializationRoundTrip(
```
var-const: I don't think this function adds any value. I'd remove it and just rely on the new overload of `ExpectDeserializationRoundTrip`. I think uniformity makes tests easier to read.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1429,6 +1468,114 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+  for (const auto& p : purpose_to_label) {
+    QueryData model(q, 1, 0, p.first);
+
+    auto result = serializer.EncodeListenRequestLabels(model);
+    EXPECT_EQ(result, p.second);
+  }
+}
+
+TEST_F(SerializerTest, DecodesListenResponseWithTargetChange) {
+  WatchTargetChange model(WatchTargetChangeState::Added, {1, 2},
```
var-const: There are [3 tests](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Tests/Remote/FSTSerializerBetaTests.mm#L870-L908) involving `WatchTargetChange` in Objective-C, can you please port them all?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1429,6 +1468,114 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+  document_change->add_target_ids(1);
+  document_change->add_target_ids(3);
+  document_change->add_removed_target_ids(2);
+  document_change->add_removed_target_ids(4);
+
+  SCOPED_TRACE("DecodesListenResponseWithDocumentChange");
+  ExpectWatchChangeDeserializationRoundTrip(model, proto);
+}
+
+TEST_F(SerializerTest, DecodesListenResponseWithDocumentDelete) {
```
var-const: Similarly, there are [4 tests](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Tests/Remote/FSTSerializerBetaTests.mm#L910-L966) involving `DocumentWatchChange` in Objective-C, can you please port them all?

In general, it's okay to add extra tests, but I think all existing tests should be ported.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1429,6 +1468,114 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+      "projects/p/databases/d/documents/one/two/three/four");
+
+  document_remove->add_removed_target_ids(1);
+  document_remove->add_removed_target_ids(2);
+
+  SCOPED_TRACE("DecodesListenResponseWithDocumentRemove");
+  ExpectWatchChangeDeserializationRoundTrip(model, proto);
+}
+
+TEST_F(SerializerTest, DecodesListenResponseWithExistenceFilter) {
```
var-const: Good job adding a test for this!

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1429,6 +1468,114 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+
+  v1::ListenResponse proto;
+
+  proto.mutable_filter()->set_count(2);
+  proto.mutable_filter()->set_target_id(100);
+
+  SCOPED_TRACE("DecodesListenResponseWithExistenceFilter");
+  ExpectWatchChangeDeserializationRoundTrip(model, proto);
+}
+
```
var-const: Can you also port [tests](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Tests/Remote/FSTSerializerBetaTests.mm#L490-L513) that exercise `DecodeMutationResult`?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1486,101 +1487,196 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
 }
 
 SnapshotVersion Serializer::DecodeVersion(
     nanopb::Reader* reader,
     const google_firestore_v1_ListenResponse& listen_response) const {
-  // TODO(varconst): implement.
-  // - (SnapshotVersion)versionFromListenResponse:
-  //     (GCFSListenResponse *)watchChange;
-  return {};
+  // We have only reached a consistent snapshot for the entire stream if there
```
var-const: Can you add a new test for this function?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
var-const: This is unfortunate -- `EXPECT_EQ` can provide more information upon a failure. Anonymous namespace shouldn't create any problems here. Can you show me the error?

var-const: Ok, it's because `EXPECT_EQ` expands into a template, and when the dependent name lookup is being performed during template instantiation, it only considers functions with external linkage. So you're right that the error is due to using an anonymous namespace.

I don't think there's any harm in moving the declaration `operator==` to `watch_change.h`, and defining `operator==` and helper function `Equals` in `watch_change.cc`, so I'd suggest doing that.

wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -512,15 +505,17 @@ class SerializerTest : public ::testing::Test {
...
   }
 
-  template <typename T, typename F, typename P>
-  auto Decode(const pb_field_t* fields, F decode_func, const P& proto) ->
-      typename F::result_type {
+  template <typename T, typename F, typename P, typename... Args>
+  auto Decode(const pb_field_t* fields,
+              F decode_func,
+              const P& proto,
+              const Args&... args) -> typename F::result_type {
```
var-const: Nice!

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1488,7 +1483,49 @@ TEST_F(SerializerTest, EncodesListenRequestLabels) {
   }
 }
 
-TEST_F(SerializerTest, DecodesListenResponseWithTargetChange) {
+TEST_F(SerializerTest, DecodesMutationResult) {
+  std::vector<FieldValue> transformations({FieldValue::FromBoolean(true),
+                                           FieldValue::FromInteger(1234),
+                                           FieldValue::FromString("string")});
+  MutationResult model(Version(123456789), std::move(transformations));
```
var-const: This version is repeated several times throughout the test, can you create a local variable for it?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1488,7 +1483,49 @@ TEST_F(SerializerTest, EncodesListenRequestLabels) {
...
+  v1::WriteResult proto;
+
+  proto.mutable_update_time()->set_seconds(
+      Version(123456789).timestamp().seconds());
+  proto.mutable_update_time()->set_nanos(
+      Version(123456789).timestamp().nanoseconds());
+  proto.mutable_transform_results()->Add();
+  proto.mutable_transform_results()->Add();
+  proto.mutable_transform_results()->Add();
+  (*proto.mutable_transform_results())[0] = ValueProto(true);
```
var-const: Nit: I'd move each assignment to be right after the `Add`.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1488,7 +1483,49 @@ TEST_F(SerializerTest, EncodesListenRequestLabels) {
...
+  v1::WriteResult proto;
+
+  proto.mutable_transform_results()->Add();
+  proto.mutable_transform_results()->Add();
+  proto.mutable_transform_results()->Add();
+  (*proto.mutable_transform_results())[0] = ValueProto(true);
+  (*proto.mutable_transform_results())[1] = ValueProto(1234);
+  (*proto.mutable_transform_results())[2] = ValueProto("string");
+
+  SCOPED_TRACE("DecodesMutationResult");
```
var-const: Nit: `"DecodesMutationResultWithNoUpdateTime"`.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1488,7 +1483,49 @@ TEST_F(SerializerTest, EncodesListenRequestLabels) {
...
+
+TEST_F(SerializerTest, DecodesMutationResultWithNoUpdateTime) {
+  std::vector<FieldValue> transformations({FieldValue::FromBoolean(true),
+                                           FieldValue::FromInteger(1234),
+                                           FieldValue::FromString("string")});
+  MutationResult model(Version(10000000), std::move(transformations));
+
+  v1::WriteResult proto;
+
+  proto.mutable_transform_results()->Add();
```
var-const: Objective-C doesn't add any transformations in this case, which implicitly tests for the case when the array is empty. Perhaps do the same, or reduce the number of transformations to 1?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1500,8 +1537,41 @@ TEST_F(SerializerTest, DecodesListenResponseWithTargetChange) {
...
   proto.mutable_target_change()->set_resume_token("resume_token");
 
-  SCOPED_TRACE("DecodesListenResponseWithTargetChange");
-  ExpectWatchChangeDeserializationRoundTrip(model, proto);
+  SCOPED_TRACE("DecodesListenResponseWithAddedTargetChange");
+  ExpectDeserializationRoundTrip(model, proto);
+}
+
+TEST_F(SerializerTest, DecodesListenResponseWithRemovedTargetChange) {
+  WatchTargetChange model(WatchTargetChangeState::Removed, {1, 2},
```
var-const: The Objective-C version also checks the error case by setting a failed `Status`, can this piece be ported, too?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1573,7 +1643,39 @@ TEST_F(SerializerTest, DecodesListenResponseWithExistenceFilter) {
   proto.mutable_filter()->set_target_id(100);
 
   SCOPED_TRACE("DecodesListenResponseWithExistenceFilter");
-  ExpectWatchChangeDeserializationRoundTrip(model, proto);
+  ExpectDeserializationRoundTrip(model, proto);
+}
+
+TEST_F(SerializerTest, DecodesVersion) {
+  SnapshotVersion model(Version(123456789).timestamp());
```
var-const: I'd create a local variable for the version.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1573,7 +1643,39 @@ TEST_F(SerializerTest, DecodesListenResponseWithExistenceFilter) {
...
+  auto model = SnapshotVersion::None();
+
+  v1::ListenResponse proto;
+
+  SCOPED_TRACE("DecodesVersionWithNoReadTime");
+  ExpectDeserializationRoundTrip(model, proto);
+}
+
+TEST_F(SerializerTest, DecodesVersionWithTargets) {
+  auto model = SnapshotVersion::None();
```
var-const: Hmm, the test doesn't seem to add a target ID to the model object?

var-const:
@wilhuff Can you please also take a look?

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Model is `SnapshotVersion`, the point of this test is to verify it will always decode to `None()` if `target_ids` is not empty in the proto.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -142,7 +142,7 @@ google_firestore_v1_Document LocalSerializer::EncodeDocument(
   }
 
   result.update_time = rpc_serializer_.EncodeVersion(doc.version());
-
+  result.has_update_time = true;
```
wilhuff: Do we have any convention on how to pair these up? Should we adopt a convention like set the `has_` field just before/after assigning to the field?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1486,101 +1487,196 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
+  return MutationResult(version, std::move(transform_results));
 }
 
 std::unordered_map<std::string, std::string>
 Serializer::EncodeListenRequestLabels(const QueryData& query_data) const {
-  // TODO(varconst): implement.
-  // - (nullable NSMutableDictionary<NSString *, NSString *> *)
-  //     encodedListenRequestLabelsForQueryData: (const QueryData &)queryData;
-  return {};
+  auto value = EncodeLabel(query_data.purpose());
```
wilhuff: This return type is wrong: the serializer needs to produce an array of `google_firestore_v1_ListenRequest_LabelsEntry`, not a `std::unordered_map<std::string, std::string>`.

Firestore/core/src/firebase/firestore/remote/serializer.cc
wilhuff: Note: Android calls this `decodeVersionFromListenResponse` and calls the thing that decodes the read time `decodeVersion`. Could you make line up more closely?

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1486,101 +1487,196 @@ google_firestore_v1_MapValue Serializer::EncodeMapValue(
...
 std::unique_ptr<WatchChange> Serializer::DecodeWatchChange(
     nanopb::Reader* reader,
     const google_firestore_v1_ListenResponse& watch_change) const {
-  // TODO(varconst): implement.
-  // - (std::unique_ptr<WatchChange>)decodedWatchChange:
-  //     (GCFSListenResponse *)watchChange;
-  return {};
+  switch (watch_change.which_response_type) {
+    case google_firestore_v1_ListenResponse_target_change_tag:
+      return DecodeTargetChange(reader, watch_change.target_change);
```
wilhuff: I like this split you've made here. Could you apply this to the other platforms? For example, `decodeWatchChange` on Android is awful.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wilhuff: `Add` returns a pointer to the thing that was added. You could avoid manual indexing with something like this:

```c++
auto transform_results = proto.mutable_transform_results();
*transform_results->Add() = ValueProto(true);
*transform_results->Add() = ValueProto(1234);
*transform_results->Add() = ValueProto("string");
```

(Note this works because member access (`->` or `.`) has higher precedence than indirection (`*`).

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -1429,6 +1463,222 @@ TEST_F(SerializerTest, EncodesResumeTokens) {
...
+}
+
+TEST_F(SerializerTest, DecodesListenResponseWithRemovedTargetChange) {
+  WatchTargetChange model(WatchTargetChangeState::Removed, {1, 2},
+                          ByteString("resume_token"),
+                          Status{Error::PermissionDenied, "Error message"});
+
+  v1::ListenResponse proto;
+
+  proto.mutable_target_change()->set_target_change_type(
```
wilhuff: Unless you've been given advice to the contrary, you might consider adding temporaries to make it clearer what each line is doing:

```c++
auto change = proto.mutable_target_change();
change->add_target_ids(1);
// ...
```

Firestore/core/src/firebase/firestore/remote/serializer.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/remote/serializer.cc
wu-hui: b/142331692

Firestore/core/src/firebase/firestore/remote/serializer.cc
wu-hui: Done.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done, thanks!

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
wu-hui: I am not aware of any conventions, and no strong preference, costa thinks before is better, just so it's harder to forget to do so.

PR #3993, opened 2019-10-03
The instanceID sync API is deprecated. Implement the async instanceID API

dmandar:


FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -215,14 +215,27 @@ - (void)refreshInstanceIDTokenAndFetchCheckInInfoWithCompletionHandler:
...
+        __weak RCNConfigFetch *instanceIDSelf = strongSelf;
+        [instanceID getIDWithHandler:^(NSString *_Nullable identity, NSError *_Nullable error) {
+          RCNConfigFetch *instanceIDStrongSelf = instanceIDSelf;
+          instanceIDStrongSelf->_settings.configInstanceID = identity;
+
+          if (identity && !error) {
+            FIRLogInfo(kFIRLoggerRemoteConfig, @"I-RCN000022", @"Success to get iid : %@.",
+                       instanceIDStrongSelf->_settings.configInstanceID);
+          } else {
+            FIRLogWarning(kFIRLoggerRemoteConfig, @"I-RCN000055", @"Error getting iid : %@.",
```
paulb777: Is it possible for !error && !identity? If so, this will crash.

FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -215,14 +215,27 @@ - (void)refreshInstanceIDTokenAndFetchCheckInInfoWithCompletionHandler:
...
       if (token) {
-        NSError *IIDError;
         strongSelf->_settings.configInstanceIDToken = [token copy];
-        strongSelf->_settings.configInstanceID = [instanceID appInstanceID:&IIDError];
-        FIRLogInfo(kFIRLoggerRemoteConfig, @"I-RCN000022", @"Success to get iid : %@.",
-                   strongSelf->_settings.configInstanceID);
+        __weak RCNConfigFetch *instanceIDSelf = strongSelf;
+        [instanceID getIDWithHandler:^(NSString *_Nullable identity, NSError *_Nullable error) {
+          RCNConfigFetch *instanceIDStrongSelf = instanceIDSelf;
+          instanceIDStrongSelf->_settings.configInstanceID = identity;
```
maksymmalyhin: Should `_settings` be modified from the `_lockQueue`? `[instanceID getIDWithHandler:]` is called on the main queue. I wonder if the code will be simpler if we call `[instanceID getIDWithHandler:]` first (before line 209) and move the code with `dispatch_async(instanceIDHandlerSelf->_lockQueue` inside the `getIDWithHandler:` handler?

FirebaseRemoteConfig/Sources/RCNFetch.m
dmandar: yes it is..not sure where we would crash though..

FirebaseRemoteConfig/Sources/RCNFetch.m
dmandar: Good pt. Done.

PR #3995, opened 2019-10-03
Implement test apps and monkey tests for CCT and FLL

mikehaney24:


mikehaney24:
Please ignore .storyboards, .proto, and .pb.swift files.

GoogleDataTransportCCTSupport/GDTCCTTestApp/app.swift
```diff
@@ -0,0 +1,85 @@
...
+          theTransport = backendSwitch?.selectedSegmentIndex == 0 ? cctTransport : fllTransport
+        }
+      } else {
+        theTransport = backendSwitch?.selectedSegmentIndex == 0 ? cctTransport : fllTransport
+      }
+      return theTransport
+    }
+  }
+
+// macOS specifics.
```
ryanwilson: No Catalyst? Or does `#if os(iOS)` capture that as well?

mikehaney24:
There's currently no catalyst. I should eventually add that, and perhaps
another test app with an iOS extension.

PR #3997, opened 2019-10-03
Add RemoteDocumentReadTimeKey to LevelDb

schmidt-sebastian:
This PR adds encode/decode functionality to store keys for the RemoteDocumentReadTime index in LevelDB.

The proposed format of the entries is:

Key: Collection Path, Read Time
Value: List of Document Ids

### DIFF
--- a/Firestore/core/src/firebase/firestore/local/leveldb_key.cc
+++ b/Firestore/core/src/firebase/firestore/local/leveldb_key.cc
@@ -46,6 +46,7 @@ const char* kTargetDocumentsTable = "target_document";
 const char* kDocumentTargetsTable = "document_target";
 const char* kRemoteDocumentsTable = "remote_document";
 const char* kCollectionParentsTable = "collection_parent";
+const char* kRemoteDocumentReadTimeTable = "remote_document_read_time";
 
 /**
  * Labels for the components of keys. These serve to make keys self-describing.
@@ -95,6 +96,9 @@ enum ComponentLabel {
    * collection_parent table, but not for collection IDs within paths).
    */
   CollectionId = 14,
+
+  /** A component containing a snapshot version. */
+  SnapshotVersion = 15,
 
   /**
    * A path segment describes just a single segment in a resource path. Path
@@ -169,6 +173,12 @@ class Reader {
   std::string ReadCollectionId() {
     return ReadLabeledString(ComponentLabel::CollectionId);
   }
+
+  /**
+   * Reads a snapshot version, encoded as a component label and a pair of seconds
+   * (int64) and nanoseconds (int32).
+   */
+  model::SnapshotVersion ReadSnapshotVersion();
 
   /**
    * Reads component labels and strings from the key until it finds a component
@@ -323,6 +333,26 @@ class Reader {
   }
 
   /**
+   * Reads a signed number from the key.
+   *
+   * If the read is unsuccessful, returns 0 and fails the Reader.
+   *
+   * Otherwise, returns the number and advances the Reader to the next unread
+   * byte.
+   */
+  int32_t ReadInt64() {
+    if (ok_) {
+      int64_t raw_result = ReadSignedNumIncreasing();
+      if (ok_) {
+        return raw_result;
+      }
+    }
+
+    Fail();
+    return 0;
+  }
+
+  /**
    * Reads a component label and a string from the key verifies that the label
    * matches the expected_label.
    *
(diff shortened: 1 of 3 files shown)


Firestore/core/src/firebase/firestore/local/leveldb_key.cc
```diff
@@ -170,6 +174,12 @@ class Reader {
     return ReadLabeledString(ComponentLabel::CollectionId);
   }
 
+  /**
+   * Reads a snapshot version, encoded as a component label and a pair of seconds
+   * (int64) and nanoseconds (int32).
+   */
+  model::SnapshotVersion ReadSnapshotVersion();
```
schmidt-sebastian: Note: I can't add `using model::SnapshotVersion` since the SnapshotVersion in the component labels would clash with it.

schmidt-sebastian:
Unassigning for now since I have to add the document ID to the key path, which will allow multiple entries per read time and avoids adding a DocumentChangeBuffer.

schmidt-sebastian:
Added the DocumentId to the key.

Firestore/core/src/firebase/firestore/local/leveldb_key.cc
```diff
@@ -322,6 +342,26 @@ class Reader {
...
 
+  /**
+   * Reads a signed number from the key.
+   *
+   * If the read is unsuccessful, returns 0 and fails the Reader.
+   *
+   * Otherwise, returns the number and advances the Reader to the next unread
+   * byte.
+   */
+  int32_t ReadInt64() {
```
wilhuff: The return type should be `int64_t`?

Firestore/core/src/firebase/firestore/local/leveldb_key.cc
```diff
@@ -322,6 +342,26 @@ class Reader {
...
+    if (ok_) {
```
wilhuff: `ReadSignedNumIncreasing` already does this dance so we don't need these additional checks. This could really just be `return ReadSignedNumIncreasing();`. `ReadInt32` does additional work (a checked cast) so it can't just be a pass-through.

Alternatively, you could just call `ReadSignedNumIncreasing` directly in the caller and omit this.

Firestore/core/src/firebase/firestore/local/leveldb_key.h
```diff
@@ -83,6 +83,11 @@ namespace local {
 //   - table_name: string = "collection_parent"
 //   - collectionId: string
 //   - parent: ResourcePath
+//
+// remote_document_read_time:
+//   - table_name: string = "remote_document_read_time"
+//   - collection: ResourcePath
+//   - read_time: SnapshotVersion
```
wilhuff: This is missing the document id

Firestore/core/test/firebase/firestore/local/leveldb_key_test.cc
```diff
@@ -356,6 +371,59 @@ TEST(RemoteDocumentKeyTest, Description) {
...
+
+TEST(RemoteDocumentReadTimeKeyTest, EncodeDecodeCycle) {
+  LevelDbRemoteDocumentReadTimeKey key;
+
+  std::vector<std::string> collection_paths{"foo", "foo/doc/bar",
+                                            "foo/doc/bar/doc/baz"};
+  std::vector<int64_t> versions{1, 1000000, 1000001};
+  std::vector<std::string> document_ids{"docA", "docB"};
+
+  for (auto&& collection_path : collection_paths) {
```
wilhuff: `auto&&` is "full magic", but we don't really need that here. `const auto&` will work just as well here and more clearly communicates that we're not mutating the values.

Firestore/core/src/firebase/firestore/local/leveldb_key.cc
schmidt-sebastian: Yes. Fixed.

Firestore/core/src/firebase/firestore/local/leveldb_key.cc
schmidt-sebastian: I left the method to make the callsites more obvious, but removed the additional ok_ checks.

Firestore/core/src/firebase/firestore/local/leveldb_key.h
schmidt-sebastian: Fixed.

Firestore/core/test/firebase/firestore/local/leveldb_key_test.cc
schmidt-sebastian: Fixed.

PR #3999, opened 2019-10-04
Add SimpleQueryEngine

schmidt-sebastian:
This PR just adds SimpleQueryEngine, but leaves it unused for now.

Port of https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/simple_query_engine.ts and https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/query_engine.ts

### DIFF
--- a/Firestore/core/src/firebase/firestore/local/query_engine.h
+++ b/Firestore/core/src/firebase/firestore/local/query_engine.h
@@ -59,7 +59,7 @@ class QueryEngine {
   virtual model::DocumentMap GetDocumentsMatchingQuery(
       const core::Query& query,
       const model::SnapshotVersion& last_limbo_free_snapshot_version,
-      model::DocumentKeySet remote_keys) const = 0;
+      const model::DocumentKeySet& remote_keys) const = 0;
 };
 
 }  // namespace local
--- a/Firestore/core/src/firebase/firestore/local/query_engine.h
+++ b/Firestore/core/src/firebase/firestore/local/query_engine.h
@@ -17,32 +17,48 @@
 #ifndef FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_LOCAL_QUERY_ENGINE_H_
 #define FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_LOCAL_QUERY_ENGINE_H_
 
-#include "Firestore/core/src/firebase/firestore/core/query.h"
-#include "Firestore/core/src/firebase/firestore/local/local_documents_view.h"
 #include "Firestore/core/src/firebase/firestore/model/document_key_set.h"
-#include "Firestore/core/src/firebase/firestore/model/document_map.h"
-#include "Firestore/core/src/firebase/firestore/model/snapshot_version.h"
 
 namespace firebase {
 namespace firestore {
+
+namespace model {
+
+class DocumentMap;
+class SnapshotVersion;
+
+}  // namespace model
+
+namespace core {
+
+class Query;
+
+}  // namespace core
+
 namespace local {
+
+class LocalDocumentsView;
 
 /**
  * Represents a query engine capable of performing queries over the local
- * document cache. You must call setLocalDocumentsView() before using.
+ * document cache. You must call `SetLocalDocumentsView()` before using.
  */
 class QueryEngine {
  public:
-  virtual ~QueryEngine() {
-  }
+  virtual ~QueryEngine() = default;
 
-  /** Sets the document view to query against. */
+  /**
+   * Sets the document view to query against.
+   *
+   * The caller owns the LocalDocumentView and must ensure that it outlives the
+   * QueryEngine.
+   */
   virtual void SetLocalDocumentsView(LocalDocumentsView* local_documents) = 0;
 
   /** Returns all local documents matching the specified query. */
   virtual model::DocumentMap GetDocumentsMatchingQuery(
-      core::Query query,
-      model::SnapshotVersion last_limbo_free_snapshot_version,
+      const core::Query& query,
+      const model::SnapshotVersion& last_limbo_free_snapshot_version,
       model::DocumentKeySet remote_keys) const = 0;
 };
 
--- a/Firestore/core/src/firebase/firestore/local/query_engine.h
+++ b/Firestore/core/src/firebase/firestore/local/query_engine.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 Google
+ *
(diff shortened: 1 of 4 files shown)


Firestore/core/src/firebase/firestore/local/query_engine.h
```diff
@@ -0,0 +1,53 @@
...
+#include "Firestore/core/src/firebase/firestore/model/document_map.h"
+#include "Firestore/core/src/firebase/firestore/model/snapshot_version.h"
+
+namespace firebase {
+namespace firestore {
+namespace local {
+
+/**
+ * Represents a query engine capable of performing queries over the local
+ * document cache. You must call setLocalDocumentsView() before using.
```
wilhuff: Capitalize SetLocalDocumentsView? Also maybe put in backticks?

Firestore/core/src/firebase/firestore/local/simple_query_engine.h
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  model::DocumentMap GetDocumentsMatchingQuery(
+      core::Query query,
+      model::SnapshotVersion last_limbo_free_snapshot_version,
+      model::DocumentKeySet remote_keys) const override;
+
+ private:
+  LocalDocumentsView* local_documents_view_;
+};
```
wilhuff: Add a newline

Firestore/core/src/firebase/firestore/local/query_engine.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+class QueryEngine {
+ public:
+  virtual ~QueryEngine() {
+  }
+
+  /** Sets the document view to query against. */
```
wilhuff: Add a note on ownership and lifetime. Something like:

The caller owns the LocalDocumentView and must ensure that it outlives the QueryEngine.

Firestore/core/src/firebase/firestore/local/query_engine.h
```diff
@@ -0,0 +1,53 @@
...
+namespace firestore {
+namespace local {
+
+/**
+ * Represents a query engine capable of performing queries over the local
+ * document cache. You must call setLocalDocumentsView() before using.
+ */
+class QueryEngine {
+ public:
+  virtual ~QueryEngine() {
```
wilhuff: Prefer default declarations to empty ones: `virtual ~QueryEngine() = default;`

Firestore/core/src/firebase/firestore/local/query_engine.h
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_LOCAL_QUERY_ENGINE_H_
+#define FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_LOCAL_QUERY_ENGINE_H_
+
+#include "Firestore/core/src/firebase/firestore/core/query.h"
```
wilhuff: We've seen nontrivial compilation speedups by using forward declarations in headers. To the extent that this is possible, we should do so. Not all existing code has been migrated to this style yet.

In general the compiler needs to know the full declaration of any non-pointer members of classes or any base classes. Otherwise concrete types can be forward declared. Note that type aliases generally can't be forward declared and have to be included.

In this case you can forward declare LocalDocumentsView, Query, and SnapshotVersion. DocumentKeySet can't be forward declared because it's a type alias.

See [`sizer.h`](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/local/sizer.h) for an example of what this looks like with differing namespaces.

Firestore/core/src/firebase/firestore/local/query_engine.h
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  virtual ~QueryEngine() {
+  }
+
+  /** Sets the document view to query against. */
+  virtual void SetLocalDocumentsView(LocalDocumentsView* local_documents) = 0;
+
+  /** Returns all local documents matching the specified query. */
+  virtual model::DocumentMap GetDocumentsMatchingQuery(
+      core::Query query,
```
wilhuff: Passing arguments like this makes copies. These likely should be const references (`const core::Query&`).

Firestore/core/src/firebase/firestore/local/query_engine.h
schmidt-sebastian: I thought it was bad style to use forward declarations and that it is better to include the relevant headers. The build time argument makes sense to me and seems like a worthwhile tradeoff.

I added the forward declarations to query_engine.h, but not to simple_query_enigine.h. Is it better to have all dependencies spelled out in each individual header file?

Firestore/core/src/firebase/firestore/local/query_engine.h
schmidt-sebastian: Done and done.

Firestore/core/src/firebase/firestore/local/query_engine.h
schmidt-sebastian: Done.

Firestore/core/src/firebase/firestore/local/query_engine.h
wilhuff: Passing by const reference is even cheaper; much cheaper.

`DocumentKeySet` and similar types are documented as "cheap" to copy in that they're O(1) to copy when the expectation would be O(n) for a deep copy. Internally they keep a shared_ptrs to their immutable representation and copies are really just reference count increments.

However, this is still an atomic integer operation which has a nontrivial cost that can easily be avoided.

As a general rule:
  * Pass by value
    * Primitives (e.g. ints, doubles)
    * Very small, trivial structs/classes (e.g. `string_view` or `Timestamp`)
    * Move-only types (e.g. `std::unique_ptr`)
    * When you're going to make a copy anyway. This gives the caller the opportunity to give away their copy with std::move (e.g. constructor arguments that will be assigned to members)
  * Otherwise, pass by const reference

schmidt-sebastian: Updated the PR to pass by const reference. Thanks for providing the guidelines, this should take some (maybe even all) of the guesswork out when it comes to picking the right argument type.

PR #4002, opened 2019-10-04
Fix formatting that was somehow ignored

wilhuff:
Also, minor tweaks to style.sh reflecting an updated source layout.

  * Firestore/Port was removed some time ago
  * The encoder moved from Firestore/Swift/Source/third_party to
    Firestore/third_party/FirestoreEncoder

### DIFF
--- a/scripts/style.sh
+++ b/scripts/style.sh
@@ -140,13 +140,12 @@ s%^./%%
 # pod gen output
 \%^gen/% d
 
-# Firestore/Swift contains code that is under 'third_party' but should be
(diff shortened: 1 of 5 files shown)


scripts/style.sh
```diff
@@ -140,13 +140,12 @@ s%^./%%
 # pod gen output
 \%^gen/% d
 
-# Firestore/Swift contains code that is under 'third_party' but should be
+# FirestoreEnoder is under 'third_party' for licensing reasons but should be
```
paulb777: Encoder

PR #4003, opened 2019-10-04
Rewrite ServerTimestamp as a property wrapper

wilhuff:
Add Swift4ServerTimestamp that behaves as before.


Firestore/Swift/Source/Codable/ServerTimestamp.swift
```diff
@@ -0,0 +1,161 @@
...
+#if swift(>=5.1)
+
+  /// A marker protocol that indicates the type can be annotated with an
+  /// `@ServerTimestamp` property wrapper. Only `Timestamp`, `Date`, and `NSDate`
+  /// can be wrapped with `@ServerTimestamp`.
+  public protocol ServerTimestampWrappable {}
+
+  extension Date: ServerTimestampWrappable {}
+  extension NSDate: ServerTimestampWrappable {}
+  extension Timestamp: ServerTimestampWrappable {}
```
wilhuff: @ryanwilson I'm mostly interested in your take on this kind of type constraint. The rest of this is essentially just like ExplicitNull so feel free to ignore.

ryanwilson: I think it's reasonable, and looks like a good solution. One thing that may be helpful is to actually encode what we're trying to get out of the type and implement it in the extension, making the decode a bit cleaner. Potentially just having two properties: seconds and nanoseconds. What do you think?

wilhuff: That's a great idea. It makes the mechanism completely extensible. I've added a test showing how @ServerTimestamp apply to arbitrary types in 4121604.

ryanwilson: Looks great!

PR #4012, opened 2019-10-07
Fix small issues in C++ serializer and port remaining tests

var-const:
These were discovered by running integration tests using the new `Serializer`.

Also, port all the missing tests from `FSTSerializerBetaTests.mm` (that I could find).


Firestore/core/src/firebase/firestore/remote/serializer.h
```diff
@@ -360,7 +366,6 @@ class Serializer {
       const google_firestore_v1_ExistenceFilter& filter) const;
 
   model::DatabaseId database_id_;
-  const std::string database_name_;
```
var-const: Objective-C code [always reencodes](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Remote/FSTSerializerBeta.mm#L242) database ID, while Android [does it once](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java#L114) and [caches the value](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java#L241).

I removed this member variable because it was unused. Alternatively, I can implement caching similar to Android.

Firestore/Protos/protos/google/firestore/v1/write.options
```diff
@@ -11,14 +11,16 @@
...
 # limitations under the License.
-#
 
 # In proto3 mode, Nanopb doesn't allow distinguishing between unset fields and
 # fields having default values, even for non-primitive types. Using the
 # workaround suggested in
 # https://github.com/nanopb/nanopb/issues/255#issuecomment-291842903
 
-# update_time is not set for delete, serializer needs to know this in order to
+google.firestore.v1.DocumentDelete.read_time proto3:false
```
var-const: I sorted the fields alphabetically.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -612,7 +615,13 @@ google_firestore_v1_Write Serializer::EncodeMutation(
       result.which_operation = google_firestore_v1_Write_update_tag;
       auto patch_mutation = static_cast<const PatchMutation&>(mutation);
       result.update = EncodeDocument(mutation.key(), patch_mutation.value());
-      result.update_mask = EncodeFieldMask(patch_mutation.mask());
+      // Note: the fact that this field is set (even if the mask is empty) is
```
var-const: At least, that's my understanding based on our conversation -- let me know if I got the details wrong.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -633,6 +642,12 @@ google_firestore_v1_Write Serializer::EncodeMutation(
             EncodeFieldTransform(field_transform);
         i++;
       }
+
+      // NOTE: We set a precondition of exists: true as a safety-check, since we
```
var-const: Note: I think you originally omitted this when implementing mutations in `Serializer`. Integration tests work without it, but perhaps it should be restored for consistency with [Objective-C](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Remote/FSTSerializerBeta.mm#L536-L539)?

Firestore/Protos/protos/google/firestore/v1/write.options
```diff
@@ -11,14 +11,16 @@
...
+# Whether `update_mask` field is set is crucial for the backend to distinguish
+# between "set" and "patch" mutations.
+google.firestore.v1.Write.update_mask proto3:false
+# `update_time` is not set for delete, serializer needs to know this in order to
```
wilhuff: These seem like they'd be more readable with a blank line before the comment.

Firestore/core/src/firebase/firestore/remote/serializer.cc
wilhuff: Great catch! This omission was a serious oversight.

The integration tests work probably because there isn't a test that verifies this is done, but it's very important because it prevents us from violating the promise that patches won't create partial documents. We should add a test that verifies that patching a document that doesn't exist 

However, one thing this shows is that nanopb makes us very sloppy: `current_document` should only be set if we have a precondition to send, but we're sending it unconditionally. Also, don't we have to set `which_condition_type`? We definitely need a test that this works.

Firestore/core/src/firebase/firestore/remote/serializer.h
```diff
@@ -124,6 +124,12 @@ class Serializer {
    */
   static pb_bytes_array_t* EncodeBytes(const std::vector<uint8_t>& bytes);
 
+  /**
+   * Returns the database ID, such as
+   * `projects/{project_id}/databases/{database_id}`.
+   */
+  pb_bytes_array_t* EncodeDatabaseId() const;
```
wilhuff: In Firestore's protocol/naming scheme "IDs" are unqualified things and "names" are qualified (see [proto comments](https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L568)). In this case, for a project ID of "p" and a database ID of "d", the database name is "projects/p/databases/d".

This should be `EncodeDatabaseName`.

Firestore/core/src/firebase/firestore/remote/serializer.h
wilhuff: It's hard to cache in a useful way on iOS because of the raw pointers in use in nanopb. Not caching is probably preferable until we can assign values to nanopb messages without giving ownership.

Firestore/core/src/firebase/firestore/remote/serializer.cc
var-const: Based on our offline discussion, the error case seems impossible to trigger, so we decided against adding the test.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -647,8 +664,10 @@ google_firestore_v1_Write Serializer::EncodeMutation(
 
 Mutation Serializer::DecodeMutation(
     nanopb::Reader* reader, const google_firestore_v1_Write& mutation) const {
-  Precondition precondition =
-      DecodePrecondition(reader, mutation.current_document);
+  auto precondition = Precondition::None();
+  if (mutation.has_current_document) {
```
var-const: Note: `DecodePrecondition` actually guards against the case where `current_document` is not set, but I think having an explicit check here is better -- it's more explicit, and also the way `DecodePrecondition` does the check is a little hacky (it checks whether the fields are zeroed).

wilhuff: SGTM. This seems like the best we can do given the way nanopb separates the has-marker from the actual value.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -121,6 +134,15 @@ QueryData CreateQueryData(absl::string_view str) {
   return CreateQueryData(Query(str));
 }
 
+// Returns the full key path, including the database name, as a string.
+std::string KeyString(const std::string& key) {
```
wilhuff: `KeyString` is pretty generic and doesn't indicate that it's the encoded form.

In the protocol, this is the "resource name". Maybe rename to something related to that? Alternatively, EncodedKey?

var-const: Changed to `ResourceName`.

I deliberately avoided `Encode` in the name because in the serializer, `Encode` functions return Nanopb types, and I didn't want to disrupt that pattern. Also, since this is a helper function, the name has to be short for it to actually help with reducing verbosity. Let me know if you have other suggestions.

PR #4013, opened 2019-10-07
Use C++ serializer for communicating with the backend

var-const:
Replace all uses of `FSTSerializerBeta` in `remote_objc_bridge.{h,mm}` with the C++ `Serializer`. Also, make most of `remote/` sources pure C++ and remove no-longer-used methods in `FSTSerializerBeta`. All integration tests pass, under Address Sanitizer as well.

Unfortunately, pretty-printing is still delegated to Objective-C, so that `remote_objc_bridge.mm` still has to be an Objective-C++ source (and which prevents inlining `remote_objc_bridge` into the streams).

#no-changelog Will add a changelog entry when local serializer is ported, too.


Firestore/core/test/firebase/firestore/nanopb/message_test.cc
```diff
@@ -0,0 +1,89 @@
...
+  grpc::ByteBuffer BadProto() const {
+    return {};
+  }
+
+ private:
+  grpc::GrpcLibraryCodegen grpc_initializer_;
+  Serializer serializer_{DatabaseId{"p", "d"}};
+};
+
+TEST_F(MessageTest, Move) {
```
var-const: I don't have too many ideas for test cases for this class, let me know if you'd like to implement any additional  tests.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -0,0 +1,177 @@
...
+}  // namespace bridge
+}  // namespace remote
+
+namespace nanopb {
+
+template <typename T>
+class Message;
+
+template <typename T>
+using MaybeMessage = util::StatusOr<Message<T>>;
```
var-const: After some thinking, I have decided against wrapping the `StatusOr` inside `Message`, because this imposes checks even in cases where the creation of `Message` cannot fail.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  Message& operator=(Message&& other) noexcept {
+    fields_ = other.fields_;
+    proto_ = other.proto_;
+    other.fields_ = nullptr;
+  }
+
+  const T& proto() const {
+    HARD_ASSERT(owns_proto(),
```
var-const: Let me know if you'd prefer to remove this assertion.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Serializes this message into a byte buffer.
+   *
+   * The lifetime of the return value is entirely independent of this message.
+   */
+  grpc::ByteBuffer CreateByteBuffer() const;
+
+ private:
+  // For access to `mutable_proto` and the explicit constructor. Most code
```
var-const: This made the class's public interface nicer. Let me know if you have any objections or suggestions.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.h
```diff
@@ -17,158 +17,125 @@
...
-/**
- * A C++ bridge to `FSTSerializerBeta` that allows creating
- * `GCFSListenRequest`s and parsing `GCFSListenResponse`s.
- */
 class WatchStreamSerializer {
  public:
-  explicit WatchStreamSerializer(FSTSerializerBeta* serializer)
-      : serializer_{serializer} {
-  }
+  explicit WatchStreamSerializer(Serializer serializer);
```
var-const: `Serializer` is actually cheap to copy, so it seemed more straightforward to just pass values around.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -127,170 +112,233 @@ bool IsLoggingEnabled() {
...
+  Message<google_firestore_v1_ListenRequest> result{
+      google_firestore_v1_ListenRequest_fields};
+  auto& request = result.mutable_proto();
+
+  request.database = serializer_.EncodeDatabaseId();
+  request.which_target_change =
+      google_firestore_v1_ListenRequest_add_target_tag;
+  request.add_target = serializer_.EncodeTarget(query);
+
+  auto labels = serializer_.EncodeListenRequestLabels(query);
```
var-const: Ideally, `Serializer` should return a Nanopb type, but in this case, I'd postpone it until I implement `ArrayRef` (so that serializer can return both the array and its size in a single value).

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -127,170 +112,233 @@ bool IsLoggingEnabled() {
...
+  return Message<google_firestore_v1_WriteResponse>::Parse(
+      google_firestore_v1_WriteResponse_fields, message);
 }
 
 model::SnapshotVersion WriteStreamSerializer::ToCommitVersion(
-    GCFSWriteResponse* proto) const {
-  return [serializer_ decodedVersion:proto.commitTime];
+    const google_firestore_v1_WriteResponse& proto) const {
+  nanopb::Reader reader;
+  auto result = serializer_.DecodeSnapshotVersion(&reader, proto.commit_time);
```
var-const: Theoretically, `reader` may have an error, though the function signature currently has no way to report that. Do you have any thoughts on how to approach error handling here?

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.h
```diff
@@ -17,158 +17,125 @@
...
- * This file contains operations in remote/ folder that are still delegated to
- * Objective-C: proto parsing and delegates.
- *
- * The principle is that the C++ implementation can only take Objective-C
- * objects as parameters or return them, but never instantiate them or call any
- * methods on them -- if that is necessary, it's delegated to one of the bridge
- * classes. This allows easily identifying which parts of remote/ still rely on
- * not-yet-ported code.
- */
+// TODO(varconst): remove this file?
```
var-const: Honestly, I kind of like the current code structure, even though it diverges from other platforms. Let me know what you think.

Firestore/Source/Remote/FSTSerializerBeta.h
```diff
@@ -84,9 +84,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (GPBTimestamp *)encodedVersion:(const model::SnapshotVersion &)version;
 - (model::SnapshotVersion)decodedVersion:(GPBTimestamp *)version;
 
-/** Returns the database ID, such as `projects/{project id}/databases/{database_id}`. */
```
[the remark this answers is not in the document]
wilhuff: Removing these now makes me vaguely uneasy because it will make it harder to perform additional audits of the serializer at the point when we're trying to cut over.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -90,33 +84,26 @@
   }
 }
 
-grpc::ByteBuffer ConvertToByteBuffer(NSData* data) {
-  grpc::Slice slice{[data bytes], [data length]};
+template <typename T>
+grpc::ByteBuffer ConvertToByteBuffer(const pb_field_t* fields,
```
wilhuff: This seems to duplicate code in `Message`.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -90,33 +84,26 @@
...
-      StringFormat("Unable to parse response from the server.\n"
-                   "Underlying error: %s\n"
-                   "Expected class: %s\n"
-                   "Received value: %s\n",
-                   error, [Proto class], ToHexString(message));
-
-  *out_status = {Error::Internal, error_description};
-  return nil;
+template <typename T, typename U>
+std::string DescribeRequest(const pb_field_t* fields, const U& request) {
```
wilhuff: Could this take `const Message<U> &`? In addition to making it possible to reuse `CreateByteBuffer` it would make it possible to get the fields out of the message too.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -127,170 +114,232 @@ bool IsLoggingEnabled() {
...
+  return Message<google_firestore_v1_ListenResponse>::Parse(
+      google_firestore_v1_ListenResponse_fields, message);
 }
 
 std::unique_ptr<WatchChange> WatchStreamSerializer::ToWatchChange(
-    GCFSListenResponse* proto) const {
-  return [serializer_ decodedWatchChange:proto];
+    const google_firestore_v1_ListenResponse& response) const {
+  nanopb::Reader reader;
+  return serializer_.DecodeWatchChange(&reader, response);
```
wilhuff: This ignores the failure that might happen while decoding.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
wilhuff: Ultimately we need to decide if we're going to propagate this error out to somewhere such that this response can be handled as broken or we have to assert.

To do the former it seems like this has to return `StatusOr<SnapshotVersion>`, no?

If we're going to do that everywhere I suggest a method like the following on reader:

```c++
template <typename T>
StatusOr<T> ToStatusOr(T&& value) {
  if (ok()) {
    return std::move(value);
  } else {
    return status();
  }
}
```

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -127,170 +114,232 @@ bool IsLoggingEnabled() {
...
-      return {};
+    auto maybe_proto =
+        Message<google_firestore_v1_BatchGetDocumentsResponse>::Parse(
+            google_firestore_v1_BatchGetDocumentsResponse_fields, response);
+    if (!maybe_proto.ok()) {
+      return maybe_proto.status();
     }
-    MaybeDocument doc = [serializer_ decodedMaybeDocumentFromBatch:proto];
+
+    const auto& proto = maybe_proto.ValueOrDie().proto();
```
wilhuff: This pattern makes it such that `DecodeMaybeDocument` can't steal the document message instance--it'll have to make a copy. It's worth thinking about how that could work. Changing this is likely work for a later PR when we actually make that happen, but something to think about until then.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
var-const: Let's discuss this. The only two alternatives I can think of are either asserting that reader is `ok`, or else returning `StatusOr` pretty much everywhere. I presume the former is closer to what we had before (Objective-C proto library would probably throw upon failure). It sounds like you would prefer the latter, using `StatusOr` throughout?

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
var-const: Postponing for now; let's discuss error handling in the thread below.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
wilhuff: Even as we have it today, moving is definitely more efficient than copying because moves are shallow, while copies are deep. Most of the interesting protos have arrays of sub-values making moves of the parent quite cheap because all arrays are dynamically allocated.

Firestore/core/src/firebase/firestore/remote/watch_stream.cc
```diff
@@ -82,13 +82,14 @@
...
-    return status;
+  MaybeMessage<google_firestore_v1_ListenResponse> maybe_response =
+      serializer_bridge_.ParseResponse(message);
+  if (!maybe_response.ok()) {
+    return maybe_response.status();
   }
 
+  const auto& response = maybe_response.ValueOrDie().proto();
+
   if (bridge::IsLoggingEnabled()) {
```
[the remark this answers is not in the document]
wilhuff: `LOG_DEBUG` already handles this for you: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/util/log.h#L46.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -235,8 +237,8 @@ class SerializerTest : public ::testing::Test {
     reader.ReadNanopbMessage(
         google_firestore_v1_BatchGetDocumentsResponse_fields, &nanopb_proto);
     serializer.DecodeMaybeDocument(&reader, nanopb_proto);
-    reader.FreeNanopbMessage(
-        google_firestore_v1_BatchGetDocumentsResponse_fields, &nanopb_proto);
+    FreeNanopbMessage(google_firestore_v1_BatchGetDocumentsResponse_fields,
```
wilhuff: Not for this PR, but consider making the constructor for `Message` public so that we can write these tests in terms of it.

Firestore/core/src/firebase/firestore/remote/write_stream.cc
```diff
@@ -136,9 +139,17 @@
     // write itself might be causing an error we want to back off from.
     backoff_.Reset();
 
-    callback_->OnWriteStreamMutationResult(
-        serializer_bridge_.ToCommitVersion(response),
-        serializer_bridge_.ToMutationResults(response));
+    auto maybe_version = write_serializer_.ToCommitVersion(*response);
+    if (!maybe_version.ok()) {
```
wilhuff: This kind of clutter was exactly what I had hoped to avoid by introducing the `Reader` previously. 

Also, shouldn't serializer operations be `Encode` and `Decode` to clearly indicate which direction the transformation is going?

```c++
Reader reader;
auto version = write_serializer_.DecodeCommitVersion(&reader, *response);
auto results = write_serializer_.DecodeMutationResults(&reader, *response);
if (!reader.ok()) return reader.status();
callback_->OnWriteStreamMutationResult(version, results);
```

wilhuff: Parse for bytes to proto seems reasonable.

Issue #269, opened 2017-09-14
FR: watchOS Support


mknippen:
With the release of watchOS 4 and an LTE Apple Watch, there’s no better time to plan support for Apple Watch. The Series 3 Watch has significantly more power, and applications will want to connect to Firebase.

salqadri:
Hi Matthew! Yes WatchOS4 looks awesome and we are keeping a close watch (pun intended) on this platform. We will continue to evaluate officially supporting the platform as it grows in apps and users, but until we do, we welcome any community contributions in this area.

boblepepeur:
I'm In !

coolioxlr:
+1

robpeach:
+1

dyah10:
+1

wipnick:
+1

ugenlik:
+1

Lawris:
+1

polok:
+1

thattyson:
+1

AsteaDevelopment:
+1

GioPalusa:
Watch OS 5 is here, and an addition with Firebase to WatchOS would be greatly appreciated! +1

zhgchgli0718:
+1

mikemike396:
Any update on this?

mariohahn:
+1

Supertecnoboff:
2019 and still no watchOS support 😢

sotrosh:
just watched firebase 2019. Added wearOS support. No watchOS support still.

pepejeria:
I guess that this won't be an issue anymore with watchOS 6 since apps are independent?

mknippen:
I think this will be a much bigger issue. Hopefully with all of the streaming data and network framework, we can make this available ASAP!

Supertecnoboff:
> I think this will be a much bigger issue. Hopefully with all of the streaming data and network framework, we can make this available ASAP!
> […](#)

Surely you can add watchOS support because presumably watchOS 6 has the CFNetwork framework?

robbiet480:
My _assumption_ is that watchOS won't be supported until Google Analytics is supported which is gonna be a long time since we don't have UIKit for watchOS and at this point it's unlikely we ever will with the introduction of SwiftUI.

ryanwilson:
[...]

PR #4016, opened 2019-10-07
Enables Firebase Messaging push notification function on watch only and independent watch app

chliangGoogle:
This PR enables FCM push notification function on watch only app, and you can also send images/videos on notifications that deliver to watch only app or independent watch app.  This is different than the watch app with an iOS companion app, where the watch app notification is delivered by apple system. For the independent watch app and watch only app, developers can request push notification independently inside the watch app, so it has its own apns token.

This is also part of the effort #269.

We can conditionally disable some of the FireLog and Reachability functionality if detected watchOS. Any feedback is welcome.


paulb777:
It looks like there are code format issues.

What is the unit testing strategy before @morganchen12's OCMock PR lands?

chliangGoogle:
This probably will break travis as I comment out systemConfiguration from podspec file, I've left the discussion in the description box.

paulb777:
Here is an example of platform specific Apple framework dependencies - https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage.podspec#L29

chliangGoogle:
Nice Thanks Paul!

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -1157,13 +1161,7 @@ - (BOOL)isProductionApp {
                              @"most likely a Dev profile.");
   }
 
-#if TARGET_OS_IOS || TARGET_OS_TV
   NSString *apsEnvironment = [plistMap valueForKeyPath:kEntitlementsAPSEnvironmentKey];
-#elif TARGET_OS_OSX
-  NSDictionary *entitlements = [plistMap valueForKey:kEntitlementsKeyForMac];
```
ryanwilson: Just wanted to confirm that `plistMap` does contain a value for `valueForKeyPath:kEntitlementsAPSEnvironmentKey` for macOS still.

Firebase/Messaging/FIRMessagingUtilities.m
```diff
@@ -149,8 +149,24 @@ int64_t FIRMessagingCurrentTimestampInMilliseconds(void) {
   return version;
 }
 
+
+NSString * FIRMessagingBundleIdentifierByRemovingLastPartFrom(NSString *bundleIdentifier) {
```
ryanwilson: Nit: `s/bundleIdentifier/bundleID/`

Firebase/InstanceID/FIRInstanceIDUtilities.m
```diff
@@ -78,6 +78,16 @@ int64_t FIRInstanceIDCurrentTimestampInMilliseconds() {
   return version;
 }
 
+NSString *FIRInstanceIDBundleIdentifierByRemovingLastPartFrom(NSString *bundleIdentifier) {
```
ryanwilson: Nit: `s/bundleIdentifier/bundleID/`

Firebase/Messaging/FIRMessagingUtilities.m
```diff
@@ -149,8 +149,24 @@ int64_t FIRMessagingCurrentTimestampInMilliseconds(void) {
...
+  NSMutableArray<NSString *> *bundleIDComponents =
+      [[bundleIdentifier componentsSeparatedByString:bundleIDComponentsSeparator] mutableCopy];
+  [bundleIDComponents removeLastObject];
+
+  return [bundleIDComponents componentsJoinedByString:bundleIDComponentsSeparator];
+}
+
 NSString *FIRMessagingAppIdentifier(void) {
-  return [[NSBundle mainBundle] bundleIdentifier];
+  NSString *bundleIdentifier = [[NSBundle mainBundle] bundleIdentifier];
```
ryanwilson: Nit: `s/bundleIdentifier/bundleID/` for the variable name.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -53,7 +53,7 @@ typedef void (*GULRealDidFailToRegisterForRemoteNotificationsIMP)(id,
 
 typedef void (*GULRealDidReceiveRemoteNotificationIMP)(id, SEL, GULApplication *, NSDictionary *);
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000 && !TARGET_OS_WATCH
```
ryanwilson: We can actually remove the `__IPHONE_OS_VERSION_MAX_ALLOWED >= 70000` now and just leave it as `!TARGET_OS_WATCH` :)

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -518,7 +518,7 @@ + (void)proxyRemoteNotificationsMethodsWithAppDelegateSubClass:(Class)appDelegat
        storeDestinationImplementationTo:realImplementationsBySelector];
 
   // For application:didReceiveRemoteNotification:fetchCompletionHandler:
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000 && !TARGET_OS_WATCH
```
ryanwilson: Same here

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -880,7 +884,7 @@ - (void)application:(GULApplication *)application
   }
 }
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000 && !TARGET_OS_WATCH
```
ryanwilson: and here

GoogleUtilities/Reachability/GULReachabilityChecker.m
```diff
@@ -22,7 +22,7 @@
 #import <GoogleUtilities/GULReachabilityChecker.h>
 
 static GULLoggerService kGULLoggerReachability = @"[GULReachability]";
-
+#if !TARGET_OS_WATCH
```
ryanwilson: Looks like we can replace this the functionality with https://developer.apple.com/documentation/watchconnectivity/wcsession/1615683-reachable?language=objc

Firebase/InstanceID/FIRInstanceID.m
chliangGoogle: Yeah, we redefined this key at line 60 for macOS

#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH
static NSString *const kEntitlementsAPSEnvironmentKey = @"Entitlements.aps-environment";	
#else	
static NSString *const kEntitlementsAPSEnvironmentKey =
    @"Entitlements.com.apple.developer.aps-environment";
#endif

Firebase/InstanceID/FIRInstanceIDUtilities.m
chliangGoogle: Done.

Firebase/Messaging/FIRMessagingUtilities.m
chliangGoogle: Done

Firebase/Messaging/FIRMessagingUtilities.m
chliangGoogle: Done.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
chliangGoogle: nice catch!

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
chliangGoogle: Done.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
chliangGoogle: Done

GoogleUtilities/Reachability/GULReachabilityChecker.m
chliangGoogle: is the reachability class to test an iPhone is connected to internet? The WCSession seems to test if a watch is connected to a companion phone and i don't know if this is 1:1 matching functionality. I wonder if I should work with the reachability owner to see how we can incorporate watch API in it. 

This is one of my concern in GoogleUtilities and FireLog pod, where I just comment out code because for messaging we don't need it. But I guess to make each pod function in watchOS, it's going to take more work.

chliangGoogle:
This PR is mainly blocked by OCMock release, the owner responded that they are likely to release this weekend (yay). We normally add unit test coverage when we add new platform support, but hopefully this CL won't get too huge since now we just have to specify them in podspec. @paulb777 @morganchen12 

If you prefer to start now, I can also merge from your changes to master later. You can also wait until this one is submitted. Up to you.

morganchen12:
IMO tests should be done in a separate PR. You can make the test PR target this branch and we can keep both pending, or merge this one early. Whether or not to merge this first I'll leave to Core to decide.

paulb777:
The PR should not merge to master until CI is green and I believe that the `pod lib lint` tests will fail until the watchOS unit tests pass.

Perhaps, there should be a watchOS branch in the meantime?

chliangGoogle:
@paulb777 Looks like XCTest doesn't support watchOS https://stackoverflow.com/questions/33756093/watchos-unit-testing

Should we unblock this PR since most of the tests are cross platform tested?

paulb777:
To go forward, we should update relevant test_specs with all platforms except watchos like https://github.com/CocoaPods/CocoaPods/issues/8283#issuecomment-439503006

chliangGoogle:
We have to specify iOS 7 because pod lib lint failed at UIBackgroundFetchResult which is API_AVAILABLE(ios(7.0)). For some reason, this failed and not recognize UIBackgroundFetchResult if not specify iOS 7 above.

chliangGoogle:
Pod lib lint GoogleUtilities.podspec succeed locally but failed on travis :
     - ERROR | [watchOS] unknown: Encountered an unknown error (Simulator for watchOS 6.0 is not available.) during validation.
Not sure why this only happens at GoogleUtilities pod.

paulb777:
Might be because the other libraries run platform specific pod lib lint runs. 

That would probably be a better fix to update .travis.yml similarly for GoogleUtilities.

.travis.yml
```diff
@@ -370,7 +370,9 @@ jobs:
       env:
         - PROJECT=GoogleUtilities METHOD=pod-lib-lint
       script:
```
paulb777: Looks good. Please make the same change for the GoogleUtilitiesCron job below.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -940,8 +943,7 @@ - (void)application:(GULApplication *)application
   }
 }
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000
-// This is needed to for the library to be warning free on iOS versions < 7.
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000 && !TARGET_OS_WATCH
```
paulb777: No longer needed now?

chliangGoogle: This is needed because of UIBackgroundFetchResult which is API_AVAILABLE(ios(7.0)). Somehow pod lib lint will fail if we ignore this check. I think something to do with the syntax API_AVAILABLE that checks below iOS 7 and it thinks UIBackgroundFetchResult is not recognized.

chliangGoogle: Considering this is already there before my change, I can also add a todo for the owner to double check on this.

.travis.yml
chliangGoogle: Done. There are two cmds under GoogleUtilitiesCron so I split each of them to 3 platforms.

chliangGoogle:
I also tested again on iOS and OSX to make sure the certificates are routed correctly. Will test tvOS once the lab is setup in new office again.

PR #4017, opened 2019-10-07
Use project number instead of project ID in Remote Config fetches

clp93:


morganchen12:
Looks like the style linter is failing your PR. Run `./scripts/style.sh` and commit the result.

FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -486,13 +486,13 @@ - (NSString *)constructServerURL {
   NSString *serverURLStr = [[NSString alloc] initWithString:kServerURLDomain];
   serverURLStr = [serverURLStr stringByAppendingString:kServerURLVersion];
 
-  if (_options.projectID) {
+  if (_options.gcmSenderID) {
```
ryanwilson: Although `GCMSenderID` is required, this could be a breaking change causing developer's fetches to fail if they don't have the appropriate `GCMSenderID` in their `FIROptions`.

Also, this should be `GCMSenderID` in Objective-C (and likely isn't compiling right now).

clp93: Thanks Ryan. I'll hold off on this PR for now then.

clp93: I'll rework this PR such that the SDK defaults to `GCMSenderID`, but then uses `ProjectID` if the `GCMSenderID` doesnt exist. Is that acceptable?

PR #4024, opened 2019-10-08
FirebaseInstallations: CreateInstallation API request error handling

maksymmalyhin:
- avoid sending CreateInstallation API requests when Firebase configuration is invalid (see go/fis-ios-error-handling for details).


FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
```diff
@@ -201,6 +218,14 @@ - (instancetype)initWithGoogleAppID:(NSString *)appID
   });
 }
 
+- (BOOL)areInstallationRegistrationParametersEqualToCurrent:
+    (FIRInstallationsStoredRegistrationParameters *)parameters {
+  NSString *APIKey = self.APIService.APIKey;
+  NSString *projectID = self.APIService.projectID;
+  return (parameters.APIKey == APIKey || [parameters.APIKey isEqual:APIKey]) &&
+         (parameters.projectID == projectID || [parameters.projectID isEqual:projectID]);
```
chliangGoogle: parameters.APIKey == APIKey
Is this for comparing the address to be the same as well?

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
```diff
@@ -234,6 +263,55 @@ - (instancetype)initWithGoogleAppID:(NSString *)appID
...
+                @"Firebase Installation registration failed for app with name: %@, error: %@",
+                self.appName, error);
+
+    FIRInstallationsItem *failedInstallation = [installation copy];
+    [failedInstallation updateWithRegistrationError:error
+                             registrationParameters:[self currentRegistrationParameters]];
+
+    // Save the error and then fail with the API error.
+    return
+        [self.installationsStore saveInstallation:failedInstallation].then(^NSError *(id result) {
```
chliangGoogle: If registration returns with error, why do we still save such installation item?

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
maksymmalyhin: I use `parameters.APIKey == APIKey` to cover the case when they are both `nil`. `[parameters.APIKey isEqual:APIKey]` returns `NO` in this case.  It is the easiest way to cover all equality cases I came up with, but I am open for suggestions.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
maksymmalyhin: I've added a design doc link explaining the details (sorry I should've added it before). TL;DR: we would like to store the registration parameters and the error received, so we can return the error without sending the API request. We will send the API request again only if the parameters changed.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
chliangGoogle: nice! I didn't know that.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
maksymmalyhin: In this case `[self doesRegistrationErrorRequireConfigChange:error]` at line #269 will return `NO` so the error will not be saved and the installation status will stay "unregistered".

chliangGoogle: Got it thanks!

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
maksymmalyhin: yeah, I didn't pay attention to it until the integration tests failures. Basically if e.g. `parameters.APIKey == nil` then any message to it will return `nil` which is casted to `NO`.

PR #3147, opened 2019-06-05
 Deprecated old Core configuration method

ryanwilson:
This shouldn't have been added - the component registration system can
provide proper configuration instead of relying on a static
`configureWithApp:` call. Many of the SDKs relied on singletons being
available there, which wasn't necessary. Instead, this should happen
in the component creation block and use the `instantiationTiming`
parameter on `registerInternalLibrary`

PR #4030, opened 2019-10-09
Fix container instantiation timing, IID startup.

ryanwilson:
The container instantiation could happen at the wrong time, before all
components have been added to the container. This fixes it, and also
moves IID to use the proper instantiation timing instead of relying on
the `configureWithApp:` call. This is part continuation of #3147, where
I'll be fixing the rest of the SDKs in follow up PRs.

Edit: Further work...

This moves the instantiation of components to after the FirebaseApp is
completely configured and assigned, as it should be. This also exposed
a recursive issue with IID calling the Messaging singleton, which in
turn called IID and instantiated multiple instances.

A change was made to break the cycle: the Messaging `isAutoInitEnabled`
call was changed to a static method vs an instance method, allowing IID
to call it during initialization without instantiating the Messaging
instance (since it doesn't have a direct dependency on it).

### DIFF
--- a/Example/Messaging/Tests/FIRMessagingTest.m
+++ b/Example/Messaging/Tests/FIRMessagingTest.m
@@ -22,7 +22,6 @@
 #import <FirebaseInstanceID/FirebaseInstanceID.h>
 #import <FirebaseAnalyticsInterop/FIRAnalyticsInterop.h>
 #import <FirebaseMessaging/FIRMessaging.h>
-#import <GoogleUtilities/GULUserDefaults.h>
 
 #import "Example/Messaging/Tests/FIRMessagingTestUtilities.h"
 #import "Firebase/Messaging/FIRMessaging_Private.h"
--- a/Example/Messaging/Tests/FIRMessagingTest.m
+++ b/Example/Messaging/Tests/FIRMessagingTest.m
@@ -22,6 +22,7 @@
 #import <FirebaseInstanceID/FirebaseInstanceID.h>
 #import <FirebaseAnalyticsInterop/FIRAnalyticsInterop.h>
 #import <FirebaseMessaging/FIRMessaging.h>
+#import <GoogleUtilities/GULUserDefaults.h>
 
 #import "Example/Messaging/Tests/FIRMessagingTestUtilities.h"
 #import "Firebase/Messaging/FIRMessaging_Private.h"
@@ -36,6 +37,9 @@ static NSString *const kFIRMessagingDefaultsTestDomain = @"com.messaging.tests";
 @property(nonatomic, readwrite, strong) NSString *defaultFcmToken;
 @property(nonatomic, readwrite, strong) NSData *apnsTokenData;
 @property(nonatomic, readwrite, strong) FIRInstanceID *instanceID;
+
+// Expose autoInitEnabled static method for IID.
++ (BOOL)isAutoInitEnabledWithUserDefaults:(NSUserDefaults *)userDefaults;
 
 // Direct Channel Methods
 - (void)updateAutomaticClientConnection;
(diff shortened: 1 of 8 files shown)


Example/Messaging/Tests/FIRMessagingTest.m
```diff
@@ -22,6 +22,7 @@
 #import <FirebaseInstanceID/FirebaseInstanceID.h>
 #import <FirebaseAnalyticsInterop/FIRAnalyticsInterop.h>
 #import <FirebaseMessaging/FIRMessaging.h>
+#import <GoogleUtilities/GULUserDefaults.h>
```
jasonhu-g: This import doesn't seem to be needed since the class uses NSUserDefaults directly

Firebase/Messaging/FIRMessaging.m
```diff
@@ -541,9 +541,20 @@ - (void)setAPNSToken:(NSData *)apnsToken type:(FIRMessagingAPNSTokenType)type {
...
+  // Defer to the class method since we're just reading from regular userDefaults and we need to
+  // read this from IID without instantiating the Messaging singleton.
+  return [[self class] isAutoInitEnabledWithUserDefaults:_messagingUserDefaults];
+}
+
+/// Checks if Messaging auto-init is enabled in the user defaults instance passed in. This is
+/// exposed as a class property for IID to fetch the property without instantiating an instance of
+/// Messaging. Since Messaging can only be used with the default FIRApp, we can have one point of
+/// entry without context of which FIRApp instance is being used.
+/// ** THIS METHOD IS DEPENDED ON INTERNALLY BY IID. PLEASE DO NOT CHANGE THE SIGNATURE. **
```
paulb777: From my read it looks like `isAutoInitEnabled` is depended upon by Messaging. `isAutoInitEnabledWithUserDefaults` is only depended upon by the tests.

paulb777: nvm, I missed the call.

ryanwilson:
Discussed offline with @paulb777 - we'll look at fixing the IID dependency on Messaging in a follow up PR by making Messaging instantiate first and initialize IID appropriately.

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -738,29 +717,20 @@ - (BOOL)isFCMAutoInitEnabled {
...
+  if (![messagingClass respondsToSelector:autoInitSelector]) {
     return YES;
   }
 
-  // Get autoInitEnabled method.
-  IMP isAutoInitEnabledIMP = [messagingInstance methodForSelector:autoInitSelector];
-  BOOL (*isAutoInitEnabled)(id, SEL) = (BOOL(*)(id, SEL))isAutoInitEnabledIMP;
+  // Get the autoInitEnabled class method.
+  IMP isAutoInitEnabledIMP = [messagingClass methodForSelector:autoInitSelector];
+  BOOL(*isAutoInitEnabled)
```
wilhuff: I believe you can cast through `(void *)` to avoid repeating yourself on the right hand side of this statement.

Firebase/Core/FIRComponentContainer.m
```diff
@@ -92,24 +98,39 @@ - (void)populateComponentsFromRegisteredClasses:(NSSet<Class> *)classes forApp:(
...
+  // instantiation.
+  @synchronized(self) {
+    for (Protocol *protocol in self.eagerProtocolsToInstantiate) {
+      // Get an instance for the protocol, which will instantiate it since it couldn't have been
+      // cached yet. Ignore the instance coming back since we don't need it.
+      __unused id unusedInstance = [self instanceForProtocol:protocol];
+    }
+
+    // All eager instantiation is complete, empty and clear the stored property now.
+    [self.eagerProtocolsToInstantiate removeAllObjects];
```
wilhuff: Assigning `nil` is sufficient to release all the contents of the array.

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -639,41 +638,41 @@ + (void)load {
 + (nonnull NSArray<FIRComponent *> *)componentsToRegister {
   FIRComponentCreationBlock creationBlock =
       ^id _Nullable(FIRComponentContainer *container, BOOL *isCacheable) {
+    // InstanceID only works with the default app.
+    if (!container.app.isDefaultApp) {
+      // Only configure for the default FIRApp.
+      FIRInstanceIDLoggerDebug(kFIRInstanceIDMessageCodeFIRApp002,
```
wilhuff: Shouldn't this be a warning or even an error?

Feel free to ignore--I see that this is just copypasta from the prior state.

Firebase/Messaging/FIRMessaging.m
wilhuff: This comment doesn't strongly enough indicate that the call is made reflectively so it won't necessarily show up in find usages.

Firebase/InstanceID/FIRInstanceID.m
ryanwilson: This is also a "this code should never be run" so maybe it should be an error, but I'll leave it as is since it was copied from before.

Firebase/Core/FIRComponentContainer.m
ryanwilson: Changed, thanks.

PR #4037, opened 2019-10-10
Move FIRInstanceIDInstanceProvider into the private header file for components to look up an instance of FIRInstanceID

jasonhu-g:
Move FIRInstanceIDInstanceProvider into the private header file for components to look up an instance of FIRInstanceID. This is needed when a components loads with componentsToRegister and depends on instanceID when initializing.

### DIFF
--- a/Firebase/InstanceID/Private/FIRInstanceID_Private.h
+++ b/Firebase/InstanceID/Private/FIRInstanceID_Private.h
@@ -23,7 +23,6 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol FIRInstanceIDInstanceProvider
 @end
 
-@class FIRInstanceIDCheckinPreferences;
 /**
  * Private API used by other Firebase SDKs.
  */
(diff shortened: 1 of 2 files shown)


Firebase/InstanceID/Private/FIRInstanceID_Private.h
```diff
@@ -41,13 +45,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable NSString *)token;
 
-/**
- *  Verify if valid checkin preferences have been loaded in memory.
- *
- *  @return YES if valid checkin preferences exist in memory else NO.
- */
-- (BOOL)hasValidCheckinInfo;
```
ryanwilson: Was this intentional?

jasonhu-g: Yep, so Travis CI failed because XCode generated a warning where FIRInstanceID does not have an implementation for hasValidCheckinInfo. I think this is old code since there's no references that method in FIRInstanceID. It looks like the code got refactored and that method no lives in FIRInstanceIDCheckinPreferences.

jasonhu-g:
This change is no longer needed with a fix Core initialization (https://github.com/firebase/firebase-ios-sdk/pull/4030)

PR #4043, opened 2019-10-10
Fix sanitizer issues

### LINKED ISSUE #4027: GoogleDataTransportCCTSupport error after pod update
On iOS 13, errors inside GoogleDataTransport 

Xcode version: 11.0
Firebase SDK version: 6.10.0
Firebase Component: Core
Component version: latest
I am getting this error 

**Pods/GoogleDataTransportCCTSupport/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTUploader.m:119:65: runtime error: null pointer passed as argument 1, which is declared to never be null**

**Pods/GoogleDataTransportCCTSupport/GoogleDataTransportCCTSupport/GDTCCTLibrary/Private/GDTCCTNanopbHelpers.h:110:52: note: _Nonnull type annotation specified here**

Any solution ?

### LINKED ISSUE #4039: GDTCORStorage Null pointer exception
On iOS 13, errors inside GDTCORStorage

Xcode version: 11.0
Firebase SDK version: 6.10.0
Firebase Component: Core
Component version: latest
I am getting this error

Getting Null pointer passed as argument 2, which is declared to never be null **on** 

`[NSKeyedUnarchiver unarchivedObjectOfClass:[GDTCORStorage class] fromData:data error:&error]; in line 220`

Screen Shot 2019-10-10 at 10 41 45 AM

mikehaney24:
Fixes #4039
Fixes #4027

Also fixes an issue with an implicit `self` retain by making it explicit.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
@@ -217,7 +217,9 @@ static NSString *GDTCORStoragePath() {
   if (@available(macOS 10.13, iOS 11.0, tvOS 11.0, *)) {
     NSError *error;
     NSData *data = [NSData dataWithContentsOfFile:[GDTCORStorage archivePath]];
-    [NSKeyedUnarchiver unarchivedObjectOfClass:[GDTCORStorage class] fromData:data error:&error];
+    if (data) {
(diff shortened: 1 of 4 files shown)


mikehaney24:
I still believe there is a stale pointer issue in the SDK, I just can't figure out the conditions that cause it.

PR #4045, opened 2019-10-10
Remove the class FIRMessagingRmq2PersistentStore 

chliangGoogle:
The file FIRMessagingRmqManager is essentially calling the same functions of FIRMessagingRmq2PersistentStore, that becomes an unnecessary class layer.
Merging FIRMessagingRmq2PersistentStore into FIRMessagingRmqManager and remove the file FIRMessagingRmq2PersistentStore.

This is a clean up before fixing the issue: b/142357385
We need to move database operations off main thread and by reducing the unnecessary class layer helps us adding async call in the next PR.

minor: remove removeRmqMessagesWithRmqId: method as it's only used in test.


Firebase/Messaging/FIRMessagingDefines.h
```diff
@@ -45,3 +45,14 @@ _Pragma("clang diagnostic pop")
     return nil; \
   } while (0)
 #endif
+
+#if defined(DEBUG)
+#define FIRMesssaging_MUST_NOT_BE_MAIN_THREAD()                                                 \
```
morganchen12: nit: `Messsaging` -> `Messaging`

morganchen12: Looks like this macro may be unused?

Firebase/Messaging/FIRMessagingDefines.h
```diff
@@ -45,3 +45,14 @@ _Pragma("clang diagnostic pop")
...
+  do {                                                                                \
+    NSAssert(![NSThread isMainThread], @"Must not be executing on the main thread."); \
```
morganchen12: Though it's debug-only, this assert may trigger false-positives if the main GCD queue is suspended. See http://blog.benjamin-encz.de/post/main-queue-vs-main-thread/

Firebase/Messaging/FIRMessagingDefines.h
chliangGoogle: Yeah I plan to use it in the next PR when I'm moving the db operation off main thread. I can remove from this PR if you like.

Firebase/Messaging/FIRMessagingDefines.h
chliangGoogle: Removed it for now.

Firebase/Messaging/FIRMessagingDefines.h
morganchen12: I think it's ok to leave in, but the macro in the `#else` statement should have the same name or it'll cause unrecognized identifier errors.

PR #4046, opened 2019-10-10
LocalSerializer feature complete

wu-hui:
* Added missing features to LocalSerializer
* Added corresponding tests


Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -67,30 +70,29 @@ firestore_client_MaybeDocument LocalSerializer::EncodeMaybeDocument(
...
-      // TODO(rsgowman): heldwriteacks:
-      // result.has_committed_mutations = no_doc.HasCommittedMutations();
+      result.has_committed_mutations =
+          static_cast<const NoDocument&>(maybe_doc).has_committed_mutations();
       return result;
 
     case MaybeDocument::Type::UnknownDocument:
       result.which_document_type =
           firestore_client_MaybeDocument_unknown_document_tag;
       result.unknown_document =
```
var-const: Please double-check with Gil, but I think that the intended way of downcasting documents is to just construct the derived type by passing over the base type to the constructor, like Objective-C does it:
```cpp
    UnknownDocument unknownDocument(document);
```
Note that it's cheap to construct. This will also avoid having to repeat the cast twice (in the cases above).

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -67,30 +70,29 @@ firestore_client_MaybeDocument LocalSerializer::EncodeMaybeDocument(
     case MaybeDocument::Type::Document:
       result.which_document_type = firestore_client_MaybeDocument_document_tag;
       result.document = EncodeDocument(static_cast<const Document&>(maybe_doc));
```
var-const: Please add a TODO to port [this logic](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Local/FSTLocalSerializer.mm#L88-L93) (checking whether the document contains a memoized proto).

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -148,6 +152,23 @@ google_firestore_v1_Document LocalSerializer::EncodeDocument(
...
+    bool has_committed_mutations) const {
+  FieldValue::Map fields_internal =
+      rpc_serializer_.DecodeFields(reader, proto.fields_count, proto.fields);
+  SnapshotVersion version =
+      rpc_serializer_.DecodeVersion(reader, proto.update_time);
+
+  DocumentState state = has_committed_mutations
+                            ? DocumentState::kCommittedMutations
+                            : DocumentState::kSynced;
+  return Document(ObjectValue::FromMap(std::move(fields_internal)),
```
var-const: Hmm, perhaps `DecodeFields` should return an `ObjectValue` instead, like `FSTSerializerBeta` does?

Firestore/core/test/firebase/firestore/local/local_serializer_test.cc
```diff
@@ -353,6 +369,43 @@ TEST_F(LocalSerializerTest, EncodesQueryData) {
...
+  ByteStringWriter writer;
+  google_firestore_v1_Target_DocumentsTarget proto =
+      remote_serializer.EncodeDocumentsTarget(query_data.query());
+  writer.WriteNanopbMessage(google_firestore_v1_Target_DocumentsTarget_fields,
+                            &proto);
+  remote_serializer.FreeNanopbMessage(
+      google_firestore_v1_Target_DocumentsTarget_fields, &proto);
+
+  ByteString query_target_bytes = writer.Release();
+  auto documents_target_proto =
```
var-const: I think using our `Serializer` to create the representation here defeats the purpose of the check a little bit. Can you set up a `::firestore::client::Target::DocumentsTarget` manually instead and add it to `documents_proto`?

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -67,30 +70,29 @@ firestore_client_MaybeDocument LocalSerializer::EncodeMaybeDocument(
     case MaybeDocument::Type::Document:
       result.which_document_type = firestore_client_MaybeDocument_document_tag;
       result.document = EncodeDocument(static_cast<const Document&>(maybe_doc));
-      // TODO(rsgowman): heldwriteacks:
-      // result.has_committed_mutations = existing_doc.HasCommittedMutations();
+      result.has_committed_mutations =
```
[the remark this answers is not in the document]
wu-hui: Done, though need to add scope to make it work.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -67,30 +70,29 @@ firestore_client_MaybeDocument LocalSerializer::EncodeMaybeDocument(
...
       result.unknown_document =
           EncodeUnknownDocument(static_cast<const UnknownDocument&>(maybe_doc));
-      // TODO(rsgowman): heldwriteacks:
-      // result.has_committed_mutations = true;
+      result.has_committed_mutations = true;
       return result;
 
     case MaybeDocument::Type::Invalid:
-      // TODO(rsgowman): Error handling
+      // TODO(wuandy): Remove Invalid from the enum.
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
wu-hui: Added.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -67,33 +67,35 @@ firestore_client_MaybeDocument LocalSerializer::EncodeMaybeDocument(
...
 
   switch (maybe_doc.type()) {
-    case MaybeDocument::Type::Document:
+    case MaybeDocument::Type::Document: {
       result.which_document_type = firestore_client_MaybeDocument_document_tag;
-      result.document = EncodeDocument(static_cast<const Document&>(maybe_doc));
-      result.has_committed_mutations =
-          static_cast<const Document&>(maybe_doc).has_committed_mutations();
+      Document doc(maybe_doc);
+      // TODO(wuandy): Check of `doc` already has a proto and use that if yes.
```
var-const: Nit: s/of `doc`/if `doc`/

Firestore/core/src/firebase/firestore/local/local_serializer.cc
wilhuff: Nit: The braces around this case now seem unnecessary.

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -103,10 +107,12 @@ MaybeDocument LocalSerializer::DecodeMaybeDocument(
 
   switch (proto.which_document_type) {
     case firestore_client_MaybeDocument_document_tag:
-      return rpc_serializer_.DecodeDocument(reader, proto.document);
+      return DecodeDocument(reader, proto.document,
+                            proto.has_committed_mutations);
```
wilhuff: This is subject to the [same bug](https://jpa.kapsi.fi/nanopb/docs/migration.html#fix-undefined-behavior-related-to-bool-fields) we work around in the [remote serializer](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/remote/serializer.cc#L395).

We probably need a helper function that encapsulates that fix and use it anytime we read a `bool`-valued field. Note that the `has_` accessors seem unaffected based on discussion in [the fix](https://github.com/nanopb/nanopb/commit/eab98ebbd68489987b8399d37cf5517792353fc7).

Note that upgrading to nanopb 0.3.9.4 is also an option for addressing this, but that needs to be done in google3 before we can depend on it in open source. It's likely easier to add the workaround for now.

We can punt resolving this into a separate PR, which is likely worthwhile because we really should audit `bool` fields in our protos and ensure that we're handling all of them uniformly.

wu-hui: Done.

Firestore/core/src/firebase/firestore/nanopb/nanopb_util.h
```diff
@@ -88,6 +90,20 @@ inline std::vector<uint8_t> MakeVector(const ByteString& str) {
...
+ * Due to the nanopb implementation, nanopb_boolean could be an integer
+ * other than 0 or 1, (such as 2). This leads to undefined behaviour when
+ * it's read as a boolean. eg. on at least gcc, the value is treated as
+ * both true *and* false. So we'll instead memcpy to an integer (via
+ * absl::bit_cast) and compare with 0.
+ *
+ * Note that it is necessary to pass-by-reference here to get the original
+ * value of `nanopb_boolean`.
+ */
+inline bool ReadBoolean(const bool& nanopb_boolean) {
```
wilhuff: `bool` is a primitive type so you can just pass it by value (i.e. just `bool` instead of `const bool&`).

Also consider naming this something like `SafeReadBoolean` to help indicate that this is defensive.

wu-hui: Renamed.

It has to be a `const bool&` though..pass-by-value here is subject to the same issue that this might be both true and false. `TEST_F(SerializerTest, BadBoolValueInterpretedAsTrue)` fails with pass-by-value here, because in `SafeReadBoolean`, the passed in value is `false` already.

wilhuff: Makes sense. Thanks for clarifying.

PR #4052, opened 2019-10-11
FIRInstallationsAPIService: Retry API request on HTTP 500.

maksymmalyhin:
- logic to retry API request on response HTTP 500 was moved to the API service and applied to all requests.


maksymmalyhin:
cc @ankitaj224

PR #4053, opened 2019-10-11
move database operation off main thread

chliangGoogle:
b/142357385 People find the app stuck on RMQ database write operation as they are on main thread.
Moving all the write operations off main thread.

Some of the sync methods used to return a value, it is used to tell the operation succeed and print error if not. We will remove those return value and print error inside the method.
There are other return values are not used at all by the parent class. Remove those and make them async method.

Also refactor the method scanWithRmqMessageHandler: which is very interesting. It has two callbacks as parameters, each is used on different occasions. This function reads data from database, instead of return the results at once, it triggers callback per row of database read. This used to work because database operation is blocking, so client just blocking read each row of database and return a callback to continue the next step. Unified both callback to one and trigger the callback after database read is complete and return with the database results.

### DIFF
--- a/Firebase/Messaging/FIRMessagingClient.m
+++ b/Firebase/Messaging/FIRMessagingClient.m
@@ -393,7 +393,7 @@ static NSUInteger FIRMessagingServerPort() {
   }
 }
 
-- (int)connectionDidReceiveAckForRmqIds:(NSArray *)rmqIds {
+- (void)connectionDidReceiveAckForRmqIds:(NSArray *)rmqIds {
   NSSet *rmqIDSet = [NSSet setWithArray:rmqIds];
   NSMutableArray *messagesSent = [NSMutableArray arrayWithCapacity:rmqIds.count];
   [self.rmq2Manager scanWithRmqMessageHandler:nil
@@ -406,7 +406,7 @@ static NSUInteger FIRMessagingServerPort() {
   for (GtalkDataMessageStanza *message in messagesSent) {
     [self.dataMessageManager didSendDataMessageStanza:message];
   }
-  return [self.rmq2Manager removeRmqMessagesWithRmqIds:rmqIds];
+  [self.rmq2Manager removeRmqMessagesWithRmqIds:rmqIds];
 }
 
 #pragma mark - Private
--- a/Firebase/Messaging/FIRMessagingClient.m
+++ b/Firebase/Messaging/FIRMessagingClient.m
@@ -16,6 +16,8 @@
 
 #import "Firebase/Messaging/FIRMessagingClient.h"
 
+#import "Firebase/Messaging/Protos/GtalkCore.pbobjc.h"
+
 #import <FirebaseInstanceID/FIRInstanceID_Private.h>
 #import <FirebaseMessaging/FIRMessaging.h>
 #import <GoogleUtilities/GULReachabilityChecker.h>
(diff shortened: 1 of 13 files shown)


chliangGoogle:
This is not going to the incoming release. So no time urgency.

Firebase/Messaging/FIRMessagingRmqManager.m
```diff
@@ -717,88 +739,93 @@ - (BOOL)saveMessageWithRmqId:(int64_t)rmqId
   return YES;
 }
 
-- (int)deleteMessagesFromTable:(NSString *)tableName
+-(void)deleteMessagesFromTable:(NSString *)tableName
```
morganchen12: nit: space between `-` and `(`

Firebase/Messaging/FIRMessagingRmqManager.m
```diff
@@ -106,6 +119,8 @@
 
 @interface FIRMessagingRmqManager () {
   sqlite3 *_database;
```
morganchen12: Though it's not part of this change, make sure the database is opened with `SQLITE_OPEN_FULLMUTEX`.

chliangGoogle: We put all database operations in serial queue already.

Firebase/Messaging/FIRMessagingClient.m
```diff
@@ -393,20 +395,22 @@ - (void)connectionDidRecieveMessage:(GtalkDataMessageStanza *)message {
   }
 }
 
-- (int)connectionDidReceiveAckForRmqIds:(NSArray *)rmqIds {
+- (void)connectionDidReceiveAckForRmqIds:(NSArray *)rmqIds {
```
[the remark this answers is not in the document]
chliangGoogle: The parent class (callers) is not expecting response on success or failure though, and considering this whole part of MCS is about to be deprecated, I lean not to add more logs on it than it already has.

Firebase/Messaging/FIRMessagingRmqManager.m
chliangGoogle: Updated.

chliangGoogle:
breakage at the podspec which is strange cuz I didn't update the podspec file.

paulb777:
It's probably a build or test failure that fails to log. Run `./scripts/pod_lib_lint.rb --verbose FirebaseMessaging.podspec` to diagnose.

Firebase/Messaging/FIRMessagingClient.m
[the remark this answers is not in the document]
morganchen12: SGTM

Firebase/Messaging/FIRMessagingRmqManager.m
morganchen12: SGTM, I had overlooked that bit

chliangGoogle:
Looks like more stall is detected at b/143297145, the travis is green now and please take a look at.

Firebase/Messaging/FIRMessagingRmqManager.m
```diff
@@ -285,60 +315,34 @@ - (int64_t)queryLastRmqId {
...
-    _FIRMessagingRmqLogAndExit(statement, @[]);
-  }
-  NSMutableArray *rmqIDArray = [NSMutableArray array];
-  while (sqlite3_step(statement) == SQLITE_ROW) {
-    const char *rmqID = (char *)sqlite3_column_text(statement, 0);
-    [rmqIDArray addObject:[NSString stringWithUTF8String:rmqID]];
-  }
-  sqlite3_finalize(statement);
+  __block NSMutableArray *rmqIDArray = [NSMutableArray array];
+  dispatch_sync(_databaseOperationQueue, ^{
```
maksymmalyhin: Ideally, we should not use  `dispatch_sync` unless it is really necessary because it is pretty easy to loose track of all consequent calls and call e.g. `unackedS2dRmqIds` from `_databaseOperationQueue` which will lead to a deadlock.

chliangGoogle: We def don't call a dispatch_sync call inside the queue, unackedS2dRmqIds is only called from outside. This class structure is quite simple that most of the writes are async and read is sync and can be ensure it's not called from a queue either.
Some of the read operations of database can not be async for now because either the public API is a sync call that calling this or we need to load data from database after app start right away because some other operation needs to access the data right away and a delay could result it thinks there's no such data. 

The most painful reason is that the original structure of this SDK is poorly designed for everything to be blocking call, so it will take a huge refactor to eventually move everything off main thread.
At this point it's still better to first move some of the operations off main thread, than everything on main thread.

maksymmalyhin: Yaeh, I agree and completely understand. Just would like to share our experience because we had issue with `dispatch_sync` just recently in GDT

PR #4054, opened 2019-10-11
Firestore podspec: avoid adding Firestore headers to sources

### LINKED ISSUE #4035: Firebase/Firestore can not work with SVGKit. it wil build error.
* Xcode version:  11
  * Firebase SDK version: Firebase/Firestore 6.7.0
  * Firebase Component: 

```
  pod 'Firebase/Firestore'
```
 * pod version:  1.8.1
 * macos version: 10.14.6 

If I add the `Firestore` edit to the newly created project, an error will occur. If you remove the `Firestore` edit, it will be fine.

如果我在新创建的工程中添加`Firestore`编辑就会出现错误。如果移除`Firestore`编辑就正常了。

#### Steps to reproduce:

1 use xcode 11.0(11A420a) create new ios project. 
2 use podfile. 
3 run pod install. 
4 run command + b.

#### Relevant Code:

podfile content:

```
platform :ios, '10.0'
inhibit_all_warnings!

target 'TestFirebase' do
  use_frameworks!
  pod 'SVGKit', '3.0.0-beta1'
  pod 'Firebase/Firestore'
end
```

var-const:
See #4035: a certain combination of pods can lead to a broken build. The underlying issue is that CocoaPods generate a header map that includes headers from _all_ the pods in a project. A header map maps short file names to full paths and is unable to handle collisions; thus, in a CocoaPods-based build, if one pod contains a header with a common name, and the other pod tries to import a header with the same short name by relying on header maps (without specifying a longer path), the header from one pod may end up getting erroneously picked up by the other pod.

In that particular case, `SVGKit` is relying on header maps and trying to import `Document.h`, which, for reasons described above, happens to pick up `document.h` from Firestore. However, since header maps cannot handle collisions, this could manifest with other pods as well.

The workaround is to avoid adding Firestore header files (except the public ones) to sources.

See https://github.com/facebook/react-native/issues/14326, which describes the same problem.

Fixes #4035.

### DIFF
--- a/FirebaseFirestore.podspec
+++ b/FirebaseFirestore.podspec
@@ -25,6 +25,7 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   s.prefix_header_file = false
 
   s.source_files = [
+    'Firestore/Source/Public/*.h',
     'Firestore/Source/**/*.{m,mm}',
     'Firestore/Protos/nanopb/**/*.cc',
     'Firestore/Protos/objc/**/*.m',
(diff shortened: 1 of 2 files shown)


var-const:
Thanks for the quick review!

paulb777:
Based on initial discussion at https://github.com/CocoaPods/CocoaPods/issues/9248, there doesn't seem to be any near-term CocoaPods solution.

What would be the feasibility of namespacing the Firestore headers with a prefix?

Issue #3521, opened 2019-08-07
Support iOS 13's new lifecycle paradigm


mikehaney24:
iOS 13 introduces a new lifecycle paradigm with iOS 13.

According to @DavidPhillipOster, a full iOS 13 app will just get the UIScene lifecycle notifications, running on either iPhone or iPad. Before iOS 13, it will just get the UIApplication lifecycle notifications,  but if it has a SceneDelegate, but does not have the dictionary in its Info.plist it will get both sets of notifications on iOS 13.

In GDT, getting multiple background/foregrounding calls is ok, so I'm solving it with #3520 

Other resources:
https://stackoverflow.com/a/56508769/361918

Reach out to me (haneym@) for additional context and example fixes if needed.

maksymmalyhin:
NOTES:
- adopting of background/foreground calls for auth should be pretty straightforward
- auth will have to adopt new way of handling open URL actions using [UISceneDelegate](https://developer.apple.com/documentation/uikit/uiscenedelegate/3238059-scene?language=objc) which implies implementation of UISceneDelegate swizzling.

PR #4066, opened 2019-10-14
Use UIApplication.connectedScenes on iOS 13

renkelvin:
Solve item #3 for https://github.com/firebase/firebase-ios-sdk/issues/3521

### DIFF
--- a/Firebase/Auth/Source/Utilities/FIRAuthDefaultUIDelegate.m
+++ b/Firebase/Auth/Source/Utilities/FIRAuthDefaultUIDelegate.m
@@ -72,7 +72,7 @@ NS_ASSUME_NONNULL_BEGIN
   }
 
   UIViewController *topViewController;
-  if (@available(iOS 13.0, *)) {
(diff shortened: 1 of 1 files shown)


Firebase/Auth/Source/Utilities/FIRAuthDefaultUIDelegate.m
```diff
@@ -70,8 +70,26 @@ - (void)dismissViewControllerAnimated:(BOOL)flag completion:(nullable void (^)(v
...
+          if (window.isKeyWindow) {
+            topViewController = window.rootViewController;
+          }
+        }
+      }
+    }
+  } else {
+    UIApplication *application = [applicationClass sharedApplication];
+    topViewController = application.keyWindow.rootViewController;
+  }
```
maksymmalyhin: Nit: moving this code out to a separate method may improve readability of the code.

renkelvin: I tried while it seems the code blocks depend on each other and hard to be separated.

PR #4051, opened 2019-10-11
Add OIDC nonce support

renkelvin:

PR #4071, opened 2019-10-14
Add OIDC nonce support

renkelvin:


renkelvin:
Recommit https://github.com/firebase/firebase-ios-sdk/pull/4051 (removed sample app part)

PR #4081, opened 2019-10-15
FIS: travis - split jobs by platform speed up builds.

maksymmalyhin:
Often FIS job is the longest to finish (e.g. https://travis-ci.org/firebase/firebase-ios-sdk/builds/598321771). I would like try to split the tests so be able to be run in parallel.


maksymmalyhin:
Doing `use-libraries` and `use-modular-headers` only on the cron job will be implemented as a separate PR if you don't mind.

PR #4082, opened 2019-10-15
Revert "Add OIDC nonce support"

renkelvin:
Reverts firebase/firebase-ios-sdk#4071


renkelvin:
We decide to postpone it to next milestone

PR #4084, opened 2019-10-16
Fix broken Xcode 8.3 build

paulb777:
The Xcode 8.3 build broke with #4066. This PR fixes it.

PR #4085, opened 2019-10-16
Bump Firestore conformance test to 9.4

paulb777:
Based on discussion in #4084 

I would like to get one of these two PRs approved ASAP to get a clean travis run before the 6.11.0 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -399,9 +399,9 @@ jobs:
 
     # Test Firestore on Xcode 8 to use old llvm to ensure C++ portability.
     - stage: test
-      osx_image: xcode8.3
+      osx_image: xcode9.4
       env:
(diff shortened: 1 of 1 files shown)


paulb777:
nvm for now. GDT does not build with Xcode 9.   cc: @mikehaney24 

https://travis-ci.org/firebase/firebase-ios-sdk/jobs/598466887

wilhuff:
Reviewing my notes on this we were using Xcode 8.3 to catch issues we weren't seeing in Xcode 9 so keeping an Xcode 9 build around doesn't really help. I think at this point this build is dead. We'll resurrect this in some alternative form that doesn't require GDT to build with unsupported versions of Xcode. Likely we'll do this once we start to try to get CI running for Visual Studio 2015.

PR #4091, opened 2019-10-16
Fix potential race during client initialization

schmidt-sebastian:
Ports the fix from https://github.com/firebase/firebase-android-sdk/commit/0213e1e81bf9bc61806496cb07ba1797c2628acf, but using the logic of the Web client (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/firestore_client.ts#L184)

### DIFF
--- a/Firestore/core/src/firebase/firestore/core/firestore_client.mm
+++ b/Firestore/core/src/firebase/firestore/core/firestore_client.mm
@@ -94,18 +94,20 @@ std::shared_ptr<FirestoreClient> FirestoreClient::Create(
       new FirestoreClient(database_info, std::move(credentials_provider),
                           std::move(user_executor), std::move(worker_queue)));
 
-  auto user_promise = std::make_shared<std::promise<User>>();
   bool credentials_initialized = false;
 
   std::weak_ptr<FirestoreClient> weak_client(shared_client);
-  auto credential_change_listener = [credentials_initialized, user_promise,
-                                     weak_client](User user) mutable {
+  auto credential_change_listener = [credentials_initialized, weak_client,
+                                     settings](User user) mutable {
     auto shared_client = weak_client.lock();
     if (!shared_client) return;
 
     if (!credentials_initialized) {
       credentials_initialized = true;
-      user_promise->set_value(user);
+
+      shared_client->worker_queue()->Enqueue([shared_client, user, settings] {
+        shared_client->Initialize(user, settings);
+      });
     } else {
       shared_client->worker_queue()->Enqueue([shared_client, user] {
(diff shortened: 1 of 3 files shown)


Firestore/core/src/firebase/firestore/core/firestore_client.mm
```diff
@@ -94,18 +94,20 @@
...
+  auto credential_change_listener = [credentials_initialized, weak_client,
+                                     settings](User user) mutable {
     auto shared_client = weak_client.lock();
     if (!shared_client) return;
 
     if (!credentials_initialized) {
       credentials_initialized = true;
-      user_promise->set_value(user);
+
+      shared_client->worker_queue()->Enqueue([shared_client, user, settings] {
```
wilhuff: This doesn't work.

The whole point of enqueueing outside the credential change listener is that we want to prevent any API call from getting on the async queue before we have the current identity. 

The `user_promise->get_future().get()` blocks in the body of that callback essentially preventing the async queue from doing anything until we've gotten the callback.

With this change, any API call can now proceed before Auth has called us back and will hit the internals before we're initialized.

schmidt-sebastian: As discussed offline, this is actually safe since `SetCredentialChangeListener` invokes the callback synchronously on the calling thread when it is first registered. Thus, the client initialization continues to be the first item enqueue on the worker queue.

I added a comment and an assert that makes this more obvious. As part of this, I also had to change credentials_initialized tp be captured by reference.

Firestore/core/src/firebase/firestore/core/firestore_client.mm
```diff
@@ -119,15 +125,9 @@ new FirestoreClient(database_info, std::move(credentials_provider),
...
-  // credential_change_listener. This is guaranteed to be synchronously
-  // dispatched onto our worker queue, so we will be initialized before any
-  // subsequently queued work runs.
-  shared_client->worker_queue()->Enqueue(
-      [shared_client, user_promise, settings] {
-        User user = user_promise->get_future().get();
-        shared_client->Initialize(user, settings);
-      });
+  HARD_ASSERT(
+      credentials_initialized,
```
wilhuff: Now that `credentials_changed` is captured by reference it's no longer value to use in the `credential_change_listener` once it escapes the current stack frame. Let's move `credentials_listener` to be a member of `FirestoreClient` to avoid these lifetime issues.

schmidt-sebastian: Fun. This is fixed.

PR #4095, opened 2019-10-16
Built for xcode 10

renkelvin:


Firebase/Auth/Source/Utilities/FIRAuthDefaultUIDelegate.m
```diff
@@ -72,23 +72,36 @@ - (void)dismissViewControllerAnimated:(BOOL)flag completion:(nullable void (^)(v
   }
 
   UIViewController *topViewController;
-  if (@available(iOS 13.0, tvOS 13.0, *)) {
```
paulb777: the if @available should still be here for Xcode 11 building for iOS 12 and under.

renkelvin: `#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000` should work for Xcode 11 no?

paulb777: That is a compile time checks that is always true on Xcode 11.

There needs to be a run time check (@available) to determine if the same built app is running on iOS 12 or 13.

renkelvin: Thanks, added!

PR #4096, opened 2019-10-16
tokenWithAuthorizedEntity: method should also check if token is fresh

chliangGoogle:
This API is used to fetch token directly from server. Right now if APNS info is not changed, the client will simply use cached token.
We should also check token freshness, which checks if IID and token is consistent or if app version has changed, etc. This helps us migrate to FIS. (b/140506963)

### DIFF
--- a/Example/InstanceID/Tests/FIRInstanceIDTokenInfoTest.m
+++ b/Example/InstanceID/Tests/FIRInstanceIDTokenInfoTest.m
@@ -26,7 +26,9 @@
 
 static NSString *const kAuthorizedEntity = @"authorizedEntity";
 static NSString *const kScope = @"scope";
-static NSString *const kToken = @"eMP633ZkDYA:APA91bGfnlnbinRVE7nUwJSr_k6cuSTKectOlt66dKv1r_-9Qvhy9XljAI62QPw307rgA0MaFHPnrU5sFxGZvsncRnkfuciwTUeyRpPNDZMFhNXt7h1BKq9Wb2A0LAANpQefrPHVUp4p";
+static NSString *const kToken = @"eMP633ZkDYA:APA91bGfnlnbinRVE7nUwJSr_k6cuSTKectOlt66dKv1r_-"
+                                @"9Qvhy9XljAI62QPw307rgA0MaFHPnrU5sFxGZvsncRnkfuciwTUeyRpPNDZMFhNXt"
+                                @"7h1BKq9Wb2A0LAANpQefrPHVUp4p";
 static NSString *const kFirebaseAppID = @"firebaseAppID";
 static NSString *const kIID = @"eMP633ZkDYA";
 static BOOL const kAPNSSandbox = NO;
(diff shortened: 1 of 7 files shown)


chliangGoogle:
@maksymmalyhin This change needs to be production as well because we don't want APNS info block any token update. I can check in to your branch after this PR is submitted.

Firebase/InstanceID/FIRInstanceIDTokenManager.m
```diff
@@ -281,12 +281,12 @@ - (BOOL)checkTokenRefreshPolicyWithIID:(NSString *)IID {
   NSMutableArray<FIRInstanceIDTokenInfo *> *tokenInfosToDelete =
       [NSMutableArray arrayWithCapacity:tokenInfos.count];
   for (FIRInstanceIDTokenInfo *tokenInfo in tokenInfos) {
-    BOOL isTokenFresh = [tokenInfo isFresh];
-    if (isTokenFresh && [tokenInfo.token hasPrefix:IID]) {
+    BOOL isTokenFresh = [tokenInfo isFreshWithIID:IID];
+    if (isTokenFresh) {
```
maksymmalyhin: nit: Doesn't look like the variable is used anywhere else. If so, can we just call the method directly in `if`?

Example/InstanceID/Tests/FIRInstanceIDTokenInfoTest.m
```diff
@@ -126,55 +127,67 @@ - (void)testTokenInfoEncodingAndDecodingWithMissingFields {
...
       [[FIRInstanceIDTokenInfo alloc] initWithAuthorizedEntity:kAuthorizedEntity
                                                          scope:kScope
                                                          token:kToken
                                                     appVersion:@"1.1"
                                                  firebaseAppID:FIRInstanceIDFirebaseAppID()];
-  XCTAssertFalse([self.validTokenInfo isFresh]);
+  XCTAssertFalse([self.validTokenInfo isFreshWithIID:kIID]);
+}
+
+-(void)testTokenInconsistentWithToken {
```
maksymmalyhin: Did you mean `testTokenInconsistentWithIID`?

Firebase/InstanceID/FIRInstanceIDTokenInfo.h
```diff
@@ -74,8 +74,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * 5. token is consistent with the current IID.
+ * 6. APNS info has changed.
+ * @param IID  The app identifiier that is used to check if token is prefixed with.
+ * @return If token is fresh.
+ *
+ */
+- (BOOL)isFreshWithIID:(NSString *)IID;
+
+/*
+ * Check wehther the token is default token.
```
morganchen12: nit: `whether` typo

chliangGoogle: Done! Thanks!

maksymmalyhin:
@chliangGoogle As for the functional test - I think we may avoid using different versions of IID for this specific case. I was thinking of mocking the token manager (or even putting the storage into the state) so InstanceID is not consistent with the token, then call the method and make sure all required methods called and eventual result is correct.

chliangGoogle:
Yep, that's exactly what new test I added is doing.

chliangGoogle:
This should fix https://github.com/firebase/firebase-ios-sdk/issues/4167

PR #4097, opened 2019-10-16
Integrate with Abseil via CocoaPods

wilhuff:
Also switch the CMake build to use Abseil as an external dependency as well.


wilhuff:
This PR unblocks adoption of Abseil within gRPC.

PR #4098, opened 2019-10-16
add heartbeat feature for ios sdks

VinayGuthal:
The heartbeat feature enables to tag the request with a code which will indicate as to whether its the first time in a given day the sdk is making a request.


GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
```diff
@@ -35,14 +35,46 @@ - (instancetype)initWithFileURL:(NSURL *)fileURL {
...
+  NSMutableDictionary *json = [NSJSONSerialization JSONObjectWithData:objectData
+                                                              options:NSJSONReadingMutableContainers
+                                                                error:&jsonError];
+  if (!json) {
+    NSLog(@"Got an error: %@", jsonError);
+  }
+  return json;
+}
+
+- (BOOL)writeDictionary:(NSMutableDictionary *)dictionary error:(NSError **)outError {
```
maksymmalyhin: I can see the following issues with the current storage API:
1. All its clients should be aware of the dictionary structure (what keys and values types and meaning are)
2. One client with a bug can break entire storage. 
3. Any dictionary related logic update should be taken into account by all clients

I would propose the following API to hide more implementation details form clients:
`- (nullable NSDate *)heartbeatDateForTag:(NSString *)tag`
`- (BOOL)setHearbeatDate:(NSDate *)date forTag:(NSString *)tag`
In this case we can change the way we store data without affecting the class clients

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
```diff
@@ -35,14 +35,46 @@ - (instancetype)initWithFileURL:(NSURL *)fileURL {
...
+- (nullable NSMutableDictionary *)getDictionary {
+  NSString *jsonString = [NSString stringWithContentsOfURL:self.fileURL
+                                                  encoding:NSUTF8StringEncoding
+                                                     error:nil];
+  NSError *jsonError;
+  if (jsonString == nil) {
+    return [NSMutableDictionary dictionary];
+  }
+  NSData *objectData = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+  NSMutableDictionary *json = [NSJSONSerialization JSONObjectWithData:objectData
```
maksymmalyhin: [NSFileCoordinator](https://developer.apple.com/documentation/foundation/nsfilecoordinator?language=objc) looks like the best way to synchronize access to the file to me. It will cover both access from different threads of the same app as well as from different processes (in case of iOS app extensions). An example can be found [here](https://www.atomicbird.com/blog/sharing-with-app-extensions).

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
```diff
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-#import "FIRCDLibrary/FIRCoreDiagnosticsDateFileStorage.h"
+#import "GULStorageHeartbeat.h"
 
-@interface FIRCoreDiagnosticsDateFileStorage ()
+@interface GULStorageHeartbeat ()
```
maksymmalyhin: nit: `GULHeartbeatStoreage` sounds a bit more natural to me.

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
VinayGuthal: as discussed offline this is not needed.

Firebase/Core/FIRHeartbeatInfo.m
```diff
@@ -0,0 +1,81 @@
...
++ (NSURL *)filePathURLWithName:(NSString *)fileName {
+  @synchronized(self) {
+    NSArray<NSString *> *paths =
+        NSSearchPathForDirectoriesInDomains(NSApplicationSupportDirectory, NSUserDomainMask, YES);
+    NSArray<NSString *> *components = @[ paths.lastObject, @"Google/FIRApp" ];
+    NSString *directoryString = [NSString pathWithComponents:components];
+    NSURL *directoryURL = [NSURL fileURLWithPath:directoryString];
+
+    NSError *error;
+    if (![directoryURL checkResourceIsReachableAndReturnError:&error]) {
```
maksymmalyhin: Creating the directory should be moved to the heartbeat storage.

Firebase/Core/Private/FIRHeartbeatInfo.h
```diff
@@ -0,0 +1,27 @@
...
+//  Created by Vinay Guthal on 10/17/19.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FIRHeartbeatInfo : NSObject
+
+typedef NS_ENUM(NSInteger, Heartbeat) {
```
maksymmalyhin: Here is a [good example](https://github.com/firebase/firebase-ios-sdk/blob/fb441a0376b0343d35ff36c18ee246f40f5be2b3/Firebase/Core/Public/FIRLoggerLevel.h#L23) of defining an enum. So the enum name should be like `FIRHeartbeatInfoCode` with the cases named like `FIRHeartbeatInfoCodeNone`, etc. 
`NS_SWIFT_NAME` is not necessary (not a public API) but encouraged in this case.

Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
```diff
@@ -153,7 +152,7 @@ @interface FIRCoreDiagnostics : NSObject <FIRCoreDiagnosticsInterop>
 @property(nonatomic, readonly) GDTCORTransport *transport;
 
 /** The storage to store the date of the last sent heartbeat. */
-@property(nonatomic, readonly) FIRCoreDiagnosticsDateFileStorage *heartbeatDateStorage;
+@property(nonatomic, readonly) GULStorageHeartbeat *heartbeatDateStorage;
```
maksymmalyhin: The variable name looks like a perfect name for the class itself - `GULHeartbeatDateStorage`, WDYT?

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
```diff
@@ -761,11 +761,10 @@ - (void)loadExperimentWithCompletionHandler:(RCNDBCompletion)handler {
 
     if (handler) {
       dispatch_async(dispatch_get_main_queue(), ^{
-        handler(
-            YES, @{
-              @RCNExperimentTableKeyPayload : [experimentPayloads copy],
-              @RCNExperimentTableKeyMetadata : [experimentMetadata copy]
-            });
+        handler(YES, @{
```
maksymmalyhin: Why it was changed?

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
maksymmalyhin: Or `GULHeartbeatDateStorage` ?

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
```diff
@@ -35,20 +36,61 @@ - (instancetype)initWithFileURL:(NSURL *)fileURL {
   return self;
 }
 
+- (nullable NSMutableDictionary *)getDictionary {
```
maksymmalyhin: nit: A method name like `heartbeatDictionary` will be more Objective-C friendly.

GoogleUtilities/FileDataStorage/GULStorageHeartbeat.m
```diff
@@ -35,20 +36,61 @@ - (instancetype)initWithFileURL:(NSURL *)fileURL {
...
+  NSError *error;
+  NSData *objectData = [NSData dataWithContentsOfURL:self.fileURL options:0 error:&error];
```
maksymmalyhin: Just a reminder, we still need to synchronize this, `setHearbeatDate: forTag:` and "check and create directory"(that is to be moved here) methods  with [NSFileCoordinator](https://developer.apple.com/documentation/foundation/nsfilecoordinator?language=objc)

Example/Core/Tests/FIRHeartbeatInfoTest.m
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017 Google
```
paulb777: 2019

Firebase/Core/FIRHeartbeatInfo.m
```diff
@@ -0,0 +1,60 @@
+// Copyright 2017 Google
```
paulb777: 2019

GoogleUtilities/FileDataStorage/GULHeartbeatDateStorage.h
```diff
@@ -18,17 +18,31 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/// Stores a date to a specified file.
-@interface FIRCoreDiagnosticsDateFileStorage : NSObject
+/// Stores either a date/dictionary to a specified file.
```
paulb777: or a dictionary?

GoogleUtilities/FileDataStorage/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,172 @@
...
+    NSTimeInterval timestamp = [date timeIntervalSinceReferenceDate];
+    stringToSave = [NSString stringWithFormat:@"%f", timestamp];
+  }
+  return [stringToSave writeToURL:self.fileURL
+                       atomically:YES
+                         encoding:NSUTF8StringEncoding
+                            error:outError];
+}
+
+// TODO(vguthal): Deprecate this and use setHeartbeatDate
```
paulb777: Why can't this be done now?

VinayGuthal: done with this change now

GoogleUtilities/FileDataStorage/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,146 @@
...
+}
+
+- (nullable NSDate *)heartbeatDateForTag:(NSString *)tag {
+  NSMutableDictionary *dictionary = [self heartbeatDictionary];
+  return dictionary[tag];
+}
+
+- (BOOL)setHearbeatDate:(NSDate *)date forTag:(NSString *)tag {
+  NSMutableDictionary *dictionary = [self heartbeatDictionary];
+  dictionary[tag] = date;
```
maksymmalyhin: I am afraid we may have a race condition here which can lead to lost updates:
- two thread read dict
- both modify their own dict copy
- the one which writes last overwrites changes from the one which writes first.

Looks like [this method](https://developer.apple.com/documentation/foundation/nsfilecoordinator/1413385-coordinatereadingitematurl?language=objc) can help us to solve it. So we will need to have a single coordinated operation for reading, modification and writing.

GoogleUtilities/FileDataStorage/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,146 @@
...
+  NSError *error;
+  BOOL isSuccess = [self writeDictionary:dictionary error:&error];
+  if (isSuccess == false) {
+    NSLog(@"Error writing dictionary data %@", error);
```
maksymmalyhin: We need to replace `NSLog` with `GULLog` here and at other places.

GoogleUtilities.podspec
```diff
@@ -37,6 +37,15 @@ other Google CocoaPods. They're not intended for direct public usage.
     ls.dependency 'GoogleUtilities/Environment'
   end
 
+  s.subspec 'FileDataStorage' do |ls|
```
maksymmalyhin: Let's add the class to `Environment` subspec to reduce amount of subspecs.
cc @paulb777 @ryanwilson

Firebase/Core/FIRHeartbeatInfo.m
```diff
@@ -0,0 +1,60 @@
...
++ (BOOL)updateIfNeededHeartbeatDateForTag:(NSString *)heartbeatTag {
+  @synchronized(self) {
+    NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
+    GULHeartbeatDateStorage *dataStorage =
+        [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
+    NSDate *heartbeatTime = [dataStorage heartbeatDateForTag:heartbeatTag];
+    NSDate *currentDate = [NSDate date];
+    if (heartbeatTime != nil) {
+      NSTimeInterval secondsBetween = [currentDate timeIntervalSinceDate:heartbeatTime];
+      if (secondsBetween < 84000) {
```
maksymmalyhin: nit: define a constant for `84000 `

GoogleUtilities/Environment/third_party/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,154 @@
...
+                                                   attributes:nil
+                                                        error:&error]) {
+                            GULLogWarning(kGULHeartbeatDateStorage, YES, @"I-COR100001",
+                                          @"Unable to create internal state storage: %@", error);
+                          }
+                        }
+                      }];
+}
+
+- (nullable NSMutableDictionary *)heartbeatDictionary:(NSURL *)readingFileURL {
```
maksymmalyhin: nit: more Objective-C name would be `heartbeatDictionaryWithFileURL:`

GoogleUtilities/Environment/third_party/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,154 @@
...
+  NSError *error;
+  NSMutableDictionary *dict;
+  NSData *objectData = [NSData dataWithContentsOfURL:readingFileURL options:0 error:&error];
+  if (objectData == nil || error != nil) {
+    dict = [NSMutableDictionary dictionary];
+  } else {
+    dict =
+        [GULSecureCoding unarchivedObjectOfClass:NSObject.class fromData:objectData error:&error];
```
maksymmalyhin: `NSObject.class` -> `NSMutableDictionary.class`

GoogleUtilities/Environment/third_party/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,154 @@
...
+- (BOOL)setHearbeatDate:(NSDate *)date forTag:(NSString *)tag {
+  NSError *error;
+  __block BOOL isSuccess = false;
+  [self.fileCoordinator
+      coordinateReadingItemAtURL:self.fileURL
+                         options:0
+                writingItemAtURL:self.fileURL
+                         options:0
+                           error:&error
+                      byAccessor:^(NSURL *readingURL, NSURL *writingURL) {
```
maksymmalyhin: 👍

GoogleUtilities/Environment/third_party/GULHeartbeatDateStorage.h
```diff
@@ -18,17 +18,31 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/// Stores a date to a specified file.
-@interface FIRCoreDiagnosticsDateFileStorage : NSObject
+// Stores either a date or a dictionary to a specified file.
```
maksymmalyhin: nit: The comment looks outdated. Also an Apple Docs single line comment should start from `///`. I think it can be fixed in a followup PR.

GoogleUtilities/Environment/third_party/GULHeartbeatDateStorage.m
```diff
@@ -0,0 +1,156 @@
...
+  NSURL *directoryURL = [NSURL fileURLWithPath:directoryString];
+  return directoryURL;
+}
+
+/** Checks and creates a directory for the directory specified by the
+ * directory url
+ * @param directoryPathURL The path to the directory which needs to be created.
+ * @param fileCoordinator The fileCoordinator object to coordinate writes to the directory.
+ */
++ (void)checkAndCreateDirectory:(NSURL *)directoryPathURL:(NSFileCoordinator *)fileCoordinator {
```
maksymmalyhin: nit: It should be `+ (void)checkAndCreateDirectory:(NSURL *)directoryPathURL fileCoordinator:(NSFileCoordinator *)fileCoordinator`. Can be in a followup PR as well.

maksymmalyhin:
@paulb777 Either @VinayGuthal or I will move `GULSecureCoding` into `Environments` subspec in a separate PR if you don't mind.

Firebase/Core/FIRHeartbeatInfo.m
```diff
@@ -16,6 +16,7 @@
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
 #import <GoogleUtilities/GULLogger.h>
 
+static long secondsInDay = 84000;
```
morganchen12: nit: This variable should be declared `const`. Also, there are 86400 seconds in one day, not 84000.

VinayGuthal: Thanks a lot!

VinayGuthal:
cc @wilhuff how do we solve the firestore failures?

paulb777:
Added #4310 for the Firestore CI issue.

GoogleUtilities.podspec
```diff
@@ -4,7 +4,7 @@ Pod::Spec.new do |s|
   s.summary          = 'Google Utilities for iOS (plus community support for macOS and tvOS)'
```
wilhuff: GoogleUtilities should get a minor version bump for this change and you should change FirebaseCore to require at least that version.

maksymmalyhin:
@wilhuff We have Firestore failures like [this](https://travis-ci.org/firebase/firebase-ios-sdk/jobs/611565230?utm_medium=notification&utm_source=github_status) that happens because a public version of GoogleUtilities is used for Firestore tests. What would be the best way to handle it? For Firebase Example project we specify source code pod dependency that's why it passes tests e.g. [here](https://travis-ci.org/firebase/firebase-ios-sdk/jobs/611565179?utm_medium=notification&utm_source=github_status)

wilhuff:
@maksymmalyhin LIke I said; the way to fix the build failures here is to make the change to the `Podfile` that I specified.

PR #4101, opened 2019-10-16
Move all ReferenceDelegate handling to .cc file

schmidt-sebastian:
This removes the requirement to include reference_delegate.h in consuming tests.

#no-changelog FTW


schmidt-sebastian:
Sending this to you @wilhuff since you typed it.

PR #4103, opened 2019-10-16
Fix issue where danger filepath match wasn't matching anything

morganchen12:
Fixes PRs not being auto-tagged and changelog checking always warning even if a changelog was modified.

Also fixes an issue where danger would error on PRs from forks.

#no-changelog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ stages:
   - test
 
 before_script:
-  - bundle exec danger
+  - '[ ! -z $DANGER_GITHUB_API_TOKEN ] && bundle exec danger || echo "Skipping Danger for External Contributor"'
 
 jobs:
   include:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,7 +10,7 @@ stages:
   - test
 
 before_script:
-  - bundle exec danger
+  - '[ ! -z $DANGER_GITHUB_API_TOKEN ] && bundle exec danger || echo "Skipping Danger for External Contributor"'
```
wilhuff: Would something like this be reasonable to put in our Dangerfile instead? That way if you try to run it locally you'd get some indication that you need to set `DANGER_GITHUB_API_TOKEN` to proceed.

morganchen12: I had considered this as well, but afaik danger will error before it even gets to the Dangerfile if this variable is not set.

PR #4104, opened 2019-10-16
token request should pass through if token is not fresh

chliangGoogle:


Example/InstanceID/Tests/FIRInstanceIDTest.m
```diff
@@ -190,6 +190,45 @@ - (void)testFCMAutoInitEnabled {
                  @"When FCM is not available, FCM Auto Init Enabled should be NO.");
 }
 
+- (void)testTokenShouldBeRefreshedIfIIDAndTokenAreNotConsistent {
```
maksymmalyhin: Great, thank you for adding the test!

Example/InstanceID/Tests/FIRInstanceIDTest.m
```diff
@@ -190,6 +190,45 @@ - (void)testFCMAutoInitEnabled {
...
+  NSDictionary *tokenOptions = @{
+    APNSKey : fakeAPNSDeviceToken,
+    serverKey : @(isSandbox),
+  };
+  FIRInstanceIDAPNSInfo *optionsAPNSInfo =
+      [[FIRInstanceIDAPNSInfo alloc] initWithTokenOptionsDictionary:tokenOptions];
+  sTokenInfo.APNSInfo = optionsAPNSInfo;
+  [[[self.mockTokenManager stub] andReturn:sTokenInfo]
+      cachedTokenInfoWithAuthorizedEntity:[OCMArg any]
+                                    scope:[OCMArg any]];
```
maksymmalyhin: nit: IMO using OCMock macros is a bit more readable comparing to the `stub` method. So it may be expressed like:
```
OCMStub([self.mockTokenManager cachedTokenInfoWithAuthorizedEntity:[OCMArg any] scope:[OCMArg any]).andReturn(sTokenInfo);
```

chliangGoogle: Yeah I also prefer OCMStub, but I use stub for this is to keep consistency inside the class. Since this class might be deprecated in the future, I would leave like this for now.

PR #4107, opened 2019-10-17
Rationalize input validation exceptions and assertion failures

wilhuff:
... under a single ThrowHandler concept. This makes it possible to set
an alternative `ThrowHandler` in our C# bindings that properly translates
input validation exceptions to the correct .NET exception types.

Existing behavior for Objective-C/Swift users should remain the same, including the fact that assertions are raised via `NSAssertionHandler`.

/cc @rsgowman in case you have any thoughts. Feel free to decline.

#no-changelog YOLO


Firestore/core/src/firebase/firestore/model/field_path.cc
```diff
@@ -183,6 +183,22 @@ std::string FieldPath::CanonicalString() const {
   return absl::StrJoin(begin(), end(), ".", JoinEscaped());
 }
 
+/* static */
```
var-const: FWIW, while it's not forbidden by the style guide, using `/* static */` comments is discouraged. Feel free to ignore.

Firestore/core/src/firebase/firestore/util/exception_apple.mm
```diff
@@ -14,34 +14,58 @@
...
+                                              int line,
+                                              const std::string& message) {
+  if (type == Exception::AssertionFailure) {
+    [[NSAssertionHandler currentHandler]
+        handleFailureInFunction:MakeNSString(func)
+                           file:MakeNSString(file)
+                     lineNumber:line
+                    description:@"%@: %s", ExceptionName(type),
+                                message.c_str()];
+    abort();
```
var-const: Nit: include `<cstdlib>`? Also, what's the rationale to prefer `abort` to `std::terminate`?

Firestore/core/src/firebase/firestore/util/exception.h
```diff
@@ -32,19 +32,63 @@
 // For internal programming errors, including internal argument checking, use
 // HARD_ASSERT or HARD_FAIL().
 
+#include <stdexcept>
```
var-const: Should this be wrapped in `ABSL_HAVE_EXCEPTIONS` as well?

Firestore/core/src/firebase/firestore/util/exception.h
```diff
@@ -32,19 +32,63 @@
...
+  using std::logic_error::logic_error;
+};
+#endif  // ABSL_HAVE_EXCEPTIONS
+
+/**
+ * An enumeration of logical exception types mapping to common user visible
+ * exceptions we might throw in response do some invalid action in an
+ * interaction with the Firestore API.
+ */
+enum class Exception {
```
var-const: The name seems to imply it's a class, maybe something like `ExceptionType`?

Firestore/core/src/firebase/firestore/util/exception.cc
```diff
@@ -0,0 +1,94 @@
...
+ABSL_ATTRIBUTE_NORETURN void Throw(Exception exception,
+                                   const char* file,
+                                   const char* func,
+                                   int line,
+                                   const std::string& message) {
+  throw_handler(exception, file, func, line, message);
+
+  // It's expected that the throw handler above does not return. If it does,
+  // just terminate.
+  std::terminate();
```
var-const: Should we manually terminate after throwing a C++ exception? Would that play nice with Unity?

Firestore/core/src/firebase/firestore/util/exception.h
wilhuff: Renamed to `ExceptionType`.

Firestore/core/src/firebase/firestore/util/exception_apple.mm
wilhuff: No good reason. I think it's just that when I wrote this it was `abort` and Rich wrote the `FailureHandler` using `std::terminate`. I've made them both `std::terminate` (and included `<exception>`, which defines it).

Issue #4108, opened 2019-10-17
RTDB persistence layer triggers a bug with number parsing


mortenbekditlevsen:
**UPDATED** After digging around a bit more, I think that I have gotten closer to the actual issue. My findings are added in the bottom.

### [READ] Step 1: Are you in the right place?
Yup, most definitely.

### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 10.3
  * Firebase SDK version: 6.2.0 (also present in 6.10.0)
  * Firebase Component: Database
  * Component version: 6.2.0

### [REQUIRED] Step 3: Describe the problem

From the user perspective we wrote the value 1000 to the RTDB and when we read it back in we got -844.

What actually happened is that we wrote the value 999.9999999999999487.
Reading back this value through the persistence layer triggers an issue with NSDecimalNumber conversion.

When disabling persistence and reading/writing directly to the RTDB, this problem does not occur.

I started a discussion about the issue on the Swift forums:
https://forums.swift.org/t/nsdecimalnumber-conversion-error/29862/3

And I was directed to previous descriptions of the issue. Apparently it has been around for a long time.
http://www.openradar.me/radar?id=5007005597040640

#### Steps to reproduce:

Use Firebase RTDB with persistence enabled. 
Write an NSDecimalNumber with many decimals of precision.
Read it back and check that the returned value is completely bonkers.
Turn off persistence and repeat to verify that it only happens with persistence enabled.

Add the following code to a Swift project to test.
Run twice to trigger the error. Then disable persistence and run again to see that the error is gone:

```
import UIKit
import FirebaseDatabase
import Firebase

class ViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()

        let db = Database.database()
        db.isPersistenceEnabled = true
        let ref = db.reference(withPath: "a")

        let decimal = Decimal(_exponent: -16, _length: 4, _isNegative: 0, _isCompact: 0, _reserved: 0, _mantissa: (65023, 35303, 8964, 35527, 0, 0, 0, 0))

        _ = ref.observeSingleEvent(of: .value) { snap in
            print("SNAP", snap)
        }

        ref.setValue(decimal as NSDecimalNumber)
    }
}
```

#### Relevant Code:

The code that triggers the issue is in FLevelDBStorageEngine.m:
```
- (id)fixDoubleParsing:(id)value
    __attribute__((no_sanitize("float-cast-overflow"))) {
[...]

PR #4109, opened 2019-10-17
Workaround for NSDecimalNumber longLongValue issue.

mortenbekditlevsen:
### Discussion

The discussion of the issue can be found here:
https://github.com/firebase/firebase-ios-sdk/issues/4108

This PR only provides a workaround for a bug with [NSDecimalNumber longLongValue] on values with high precision.

It does not deal with the fact that the RTDB server appears to store a double wrapped in an NSNumber differently than the same value wrapped in an NSDecimalNumber.
As an NSDecimalNumber the higher precision is stored, but the persistence layer has no knowledge about whether we are in one case or the other.

### Testing

The added logic changes the CFNumber type of one of the existing tests. Otherwise all existing tests passed.

### API Changes

No API changes in the PR.


paulb777:
@mortenbekditlevsen  Thanks for the PR!  It looks like our new Danger integration is breaking CI for external PRs.  cc: @morganchen12  We'll investigate that and get this reviewed.

morganchen12:
@mortenbekditlevsen The danger issue has been fixed in master, if you merge/rebase it should go away. Sorry for the inconvenience!

Firebase/Database/Persistence/FLevelDBStorageEngine.m
```diff
@@ -878,7 +878,17 @@ - (id)fixDoubleParsing:(id)value
     // values, including 2.47. Because we use the exact bytes for hashing on the
     // server this will lead to hash mismatches. The parser of NSNumber seems to
     // be more in line with what the server expects, so we use that here
-    if ([value isKindOfClass:[NSNumber class]]) {
+    if ([value isKindOfClass:[NSDecimalNumber class]]) {
+        NSDecimal d = [(NSDecimalNumber*)value decimalValue];
```
schmidt-sebastian: What do you think of renaming "d" to "original" and "result" to "rounded"?

Firebase/Database/Persistence/FLevelDBStorageEngine.m
```diff
@@ -878,7 +878,17 @@ - (id)fixDoubleParsing:(id)value
     // values, including 2.47. Because we use the exact bytes for hashing on the
     // server this will lead to hash mismatches. The parser of NSNumber seems to
     // be more in line with what the server expects, so we use that here
```
schmidt-sebastian: Can you move this existing comment into the `else if if ([value isKindOfClass:[NSNumber class]])` block?

Firebase/Database/Persistence/FLevelDBStorageEngine.m
schmidt-sebastian: Please add a comment with a link to the underlying Radar here.

Firebase/Database/Persistence/FLevelDBStorageEngine.m
mortenbekditlevsen: Certainly, but the essential idea of what is going on inside both the NSDecimalNumber case and the NSNumber case is the same, so does this comment not apply to both?

I am updating the PR with the moved comment, but I think that perhaps it belongs to the entire method.

mortenbekditlevsen:
Thank you very much for the review. I have updated the PR with requested changes and added additional comments.

Have you had a chance to consider the other issue described in: #4108?
Namely that it appears that the RTDB server can represent the 'same' value in different versions based on whether it is wrapped as an NSDecimalNumber or not.
This is all fine and good, but it seems that the persistence layer can not know which representation is desired, and this could cause hashing/sync issues - and also minor precision issues based on whether you are enabling persistence or not.

PR #4113, opened 2019-10-17
Copy component container to GoogleUtilitiesComponents

ryanwilson:
This will allow non-Firebase SDKs to take advantage of the component
system. Note: Firebase Core will eventually move over to use this, but
more investigation is needed about how this can be done with the least
amount of churn for other SDKs.


GoogleUtilities.podspec
```diff
@@ -76,6 +76,13 @@ other Google CocoaPods. They're not intended for direct public usage.
     adss.dependency 'GoogleUtilities/Environment'
   end
 
+  s.subspec 'ComponentContainer' do |cc|
+    cc.source_files = 'GoogleUtilities/ComponentContainer/**/*.[mh]', 'GoogleUtilities/Common/*.h'
+    cc.public_header_files = 'GoogleUtilities/ComponentContainer/Private/*.h'
```
wilhuff: Did you mean Public here? This seems like a mistake.

The headers below are also added in a Private directory which seems off. Just by their location it seems like they shouldn't be used (even though you're actually marking them part of public_header_files).

paulb777: @wilhuff This is correct.  From the non-intuitive podspec docs:

private_header_filesmulti-platform
A list of file patterns that should be used to mark private headers.

These patterns are matched against the public headers (or all the headers if no public headers have been specified) to exclude those headers which should not be exposed to the user project and which should not be used to generate the documentation. When the library is built, these headers will appear in the build directory.

wilhuff: If the point of this is to make it such that other projects can use the container wouldn't these have to be public headers though?

paulb777: Generally the policy has been to use private headers for APIs that are intended for other Firebase pods that are not intended to be used outside of Firebase.

This is necessary for pods that have both public and internal dependencies.  We could revisit that for pods that are intended purely for internal use and instead use public headers for them.

That brings up another question. I'm not sure we should add another subspec to GoogleUtilities versus making a separate GoogleUtiilitiesComponents pod.

wilhuff: First sentence of description: "This will allow non-Firebase SDKs to take advantage of the component system". Private headers won't work in that circumstance.

ryanwilson: As part of moving this externally to its own CocoaPod, I've made the necessary headers public.

.travis.yml
```diff
@@ -414,6 +414,28 @@ jobs:
...
+        - PROJECT=GoogleUtilitiesComponents METHOD=pod-lib-lint
+      script:
+        - travis_retry ./scripts/if_changed.sh ./scripts/pod_lib_lint.rb GoogleUtilitiesComponents.podspec
+
+    - stage: test
+      if: type = cron
+      env:
+        - PROJECT=GoogleUtilitiesComponentsCron METHOD=pod-lib-lint
+      script:
+
```
paulb777: nit: delete blank line

GoogleUtilities.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'GoogleUtilities'
-  s.version          = '6.3.1'
+  s.version          = '6.4.0'
```
paulb777: Undo all changes in this file

GoogleUtilities/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 6.4.0
+- ComponentContainer introduced. (#4113)
```
paulb777: Undo

GoogleUtilitiesComponents.podspec
```diff
@@ -0,0 +1,39 @@
+Pod::Spec.new do |s|
+  s.name             = 'GoogleUtilitiesComponents'
+  s.version          = '1.0.0'
+  s.summary          = 'Google Utilities Component Container for Apple platforms.'
+
+  s.description      = <<-DESC
+An internal Google utility that is a dependency injection system for SDKs to depend on other SDKs in
```
paulb777: Is "library" more accurate than "SDK"?

ryanwilson: Makes sense!

GoogleUtilities/CHANGELOG.md
ryanwilson: Done.

.travis.yml
ryanwilson: Done.

GoogleUtilities.podspec
ryanwilson: Good catch, thanks.

GoogleUtilitiesComponents.podspec
```diff
@@ -0,0 +1,39 @@
...
+    :git => 'https://github.com/firebase/firebase-ios-sdk.git',
+    :tag => 'UtilitiesComponents-' + s.version.to_s
+  }
+
+  s.ios.deployment_target = '8.0'
+  s.osx.deployment_target = '10.11'
+  s.tvos.deployment_target = '10.0'
+
+  s.cocoapods_version = '>= 1.4.0'
+  s.prefix_header_file = false
```
paulb777: We might want `s.static_frameworks = true` at least until CocoaPods 1.9.x releases and gives fine grain control to Podfiles for controlling static vs dynamic.

GoogleUtilitiesComponents.podspec
```diff
@@ -0,0 +1,39 @@
...
+  s.version          = '1.0.0'
+  s.summary          = 'Google Utilities Component Container for Apple platforms.'
+
+  s.description      = <<-DESC
+An internal Google utility that is a dependency injection system for libraries to depend on other
+libraries in a type safe and potentially weak manner.
+Not intended for direct public usage.
+                       DESC
+
+  s.homepage         = 'https://github.com/firebase/firebase-ios-sdk/tree/master/GoogleUtilitiesComponents'
```
paulb777: Needs to be reachable to pass pod lib lint

GoogleUtilitiesComponents.podspec
ryanwilson: Good call, will add.

GoogleUtilitiesComponents.podspec
ryanwilson: Changed, thanks.

PR #4114, opened 2019-10-17
FirebaseInstallation: send IID checkin data on IID migration.

maksymmalyhin:
- send IID checkin data in a header of CreateInstallation API HTTP request if IID is migrated (b/142821608)

#no-changelog


FirebaseInstallations/Source/Library/FIRInstallationsItem.h
```diff
@@ -47,6 +48,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property(nonatomic, nullable) FIRInstallationsStoredRegistrationError *registrationError;
 
+/// Instance ID checkin data imported from IID checkin store as a part of IID migration.
+@property(nonatomic, nullable) FIRInstallationsStoredIIDCheckin *IIDCheckin;
```
chliangGoogle: My impression is that FIS should not be aware of any checkin data. Especially we are on the plan of removing checkin.

chliangGoogle: talked offline and looks like this is just to read from storage for authentication.

FirebaseInstallations/Source/Library/SecureStorage/FIRInstallationsKeychainUtils.m
```diff
@@ -0,0 +1,107 @@
...
+  } else {
+    if (outError) {
+      *outError = [FIRInstallationsErrorUtil keychainErrorWithFunction:@"SecItemCopyMatching"
+                                                                status:status];
+    }
+  }
+  return nil;
+}
+
++ (BOOL)setItem:(NSData *)item
```
chliangGoogle: Are the write the operations off mian thread?  Might consider doing that later as we got issue before when keychain operation stuck when it's running on main thread.

maksymmalyhin: This method is just a sync helper, so it is up to caller what thread to perform the operation on. This particular method is called from [here](https://github.com/firebase/firebase-ios-sdk/pull/4114/files#diff-af0497ab8d81ffca80c9d5cb3c98cc8dR99) from a keychain dedicated queue.

PR #4115, opened 2019-10-17
[danger] Don't check changelogs on non-source changes

morganchen12:
Reduce bot spam where it's unhelpful.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -4,7 +4,8 @@
 # Taken from samdmarshall/danger
 def didModify(files_array)
   files_array.each do |file_name|
-    if git.modified_files.include?(file_name) || git.deleted_files.include?(file_name)
+    if git.modified_files.include?(file_name) ||
+       git.deleted_files.include?(file_name)
       return true
     end
   end
@@ -94,11 +95,14 @@ has_license_changes = didModify(["LICENSE"])
 @has_zipbuilder_changes = hasChangesIn("ZipBuilder/")
 
 # A FileList containing ObjC, ObjC++ or C++ changes.
-sdk_changes = (git.modified_files + git.added_files + git.deleted_files).select do |line|
+sdk_changes = (git.modified_files + 
+               git.added_files + 
+               git.deleted_files).select do |line|
   line.end_with?(".h") ||
     line.end_with?(".m") ||
     line.end_with?(".mm") ||
-    line.end_with?(".cc")
+    line.end_with?(".cc") ||
+    line.end_with?(".swift")
 end
 
 # Whether or not the PR has modified SDK source files.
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -93,11 +93,20 @@ has_license_changes = didModify(["LICENSE"])
 @has_googleutilities_changes = hasChangesIn("GoogleUtilities/")
 @has_zipbuilder_changes = hasChangesIn("ZipBuilder/")
 
+sdk_changes = (git.modified_files + git.added_files + git.deleted_files).select do |line|
+  line.end_with?(".h") || line.end_with?(".m") || line.end_with?(".mm")
```
paulb777: .cc also?

morganchen12: Good catch, added that to the check.

wilhuff: There's a bunch of others:
.swift
.py
.rb
.sh
.cmake
CMakeLists.txt  <-- not an extension, actual filename

We have a list in https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check_copyright.sh#L25

morganchen12: For the sake of checking SDK changelogs, though, are `.py`, `.rb`, `cmake`, and `.sh` files necessary? I had thought about Swift too but iirc our only Swift library is the Firestore extensions library, and that one doesn't have an associated changelog.

Part of the goal of this PR is to not generate changelog warning checks for scripts and podspecs because most of the time those don't warrant changelog entries (unless I'm mistaken).

paulb777: Agreed, changes to files with those extensions are infrastructure that typically won't require a Changelog update.

wilhuff: I see. FirebaseFirestoreSwift will soon have a changelog so I think it's reasonable to nag .swift changes.

Dangerfile
```diff
@@ -93,11 +93,25 @@ has_license_changes = didModify(["LICENSE"])
...
 ### Actions
 
-# Warn if a changelog is left out on a non-trivial PR
-if !has_changelog_changes && !declared_trivial
-  warn("Did you forget to add a changelog entry? (Add `#no-changelog` to the PR description to silence this warning.)")
+# Warn if a changelog is left out on a non-trivial PR that has modified
+# SDK source files (podspec, markdown, etc changes are excluded).
+if has_sdk_changes
+  if !has_changelog_changes && !declared_trivial
+    warn("Did you forget to add a changelog entry? (Add `#no-changelog` to the PR description to silence this warning.)")
```
wilhuff: Line seems really long. go/ruby-style suggests 80 column line length.

Dangerfile
```diff
@@ -93,11 +93,25 @@ has_license_changes = didModify(["LICENSE"])
 @has_googleutilities_changes = hasChangesIn("GoogleUtilities/")
 @has_zipbuilder_changes = hasChangesIn("ZipBuilder/")
 
+# A FileList containing ObjC, ObjC++ or C++ changes.
+sdk_changes = (git.modified_files + git.added_files + git.deleted_files).select do |line|
+  line.end_with?(".h") ||
+    line.end_with?(".m") ||
+    line.end_with?(".mm") ||
+    line.end_with?(".cc")
```
wilhuff: .swift too

Dangerfile
morganchen12: Done. Fixed a couple other line length violations too.

Dangerfile
morganchen12: Added.

PR #4116, opened 2019-10-18
C++ migration: use C++ serializer for communicating with the local store

var-const:
After this PR is merged, the Objective-C serializers can be completely removed from the project. Also, this PR changes the `Message` class quite a bit; it's probably in its final form now.

I did some performance testing locally and came to the conclusion that replacing Objective-C serializers with their C++ versions brings no performance downgrade; actually, there seems to be a very slight improvement. I tested writing large batches and reading large collections:
* both are network-dominated; wall time is about twice longer than CPU time;
* serialization represents a pretty small percentage of the total CPU time;
* for writing, C++ serialization is a little more efficient, taking ~6% of total CPU time, compared to ~8% in the Objective-C version. I've seen similar results for reading.

All integration tests pass, under Address Sanitizer as well.


Firestore/core/src/firebase/firestore/nanopb/fields_map.h
```diff
@@ -0,0 +1,117 @@
...
+#include "Firestore/Protos/nanopb/firestore/local/target.nanopb.h"
+#include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
+#include "Firestore/Protos/nanopb/google/firestore/v1/firestore.nanopb.h"
+#include "Firestore/Protos/nanopb/google/type/latlng.nanopb.h"
+
+namespace firebase {
+namespace firestore {
+namespace nanopb {
+
+/**
```
var-const: Feel free to suggest a different name for this function/file.

Firestore/core/src/firebase/firestore/nanopb/fields_map.h
```diff
@@ -0,0 +1,117 @@
...
+ * Returns a pointer to the Nanopb-generated array that describes the fields
+ * of the Nanopb proto; the array is required to call most Nanopb functions.
+ *
+ * There is always a one-to-one correspondence between a Nanopb-generated
+ * message type and its fields descriptor; essentially, the fields descriptor is
+ * a property of the type.
+ */
+// The non-specialized version of this function is deleted to make sure that
```
var-const: One of the reasons I chose specialized template over a set of overloaded functions is that I think the error message becomes slightly clearer (otherwise, it will be diluted by the compiler listing all of the overloads that don't match). Also, this approach is more similar to a traits class, which it essentially is.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -130,48 +155,63 @@ class Message {
...
+  ByteString ToByteString() const;
 
-  explicit Message(const pb_field_t* fields) : fields_{fields} {
+ private:
+  // Returns a pointer to the Nanopb-generated array that describes the fields
+  // of the Nanopb proto; the array is required to call most Nanopb functions.
+  //
+  // Note that this is essentially a property of the type, but cannot be made
+  // a template parameter for various technical reasons.
+  static const pb_field_t* fields() {
```
var-const: Working on this PR, I came to the conclusion that the fields descriptor is a property of the type, not an object, so it makes more sense to not store it. It also makes `Message` safer because accidentally providing the wrong descriptor could lead to catastrophic results -- now, all the mappings between Nanopb types and corresponding descriptors are in one place that is much easier to check for correctness.

Ideally, this would be `constexpr`, but making Nanopb generate `constexpr` arrays under the C++11 constraints seems hard.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -55,68 +47,101 @@ using MaybeMessage = util::StatusOr<Message<T>>;
...
  *
  * Note that moving *isn't* a particularly cheap operation in the general case.
- * Even without doing deep copies, Nanopb-generated messages may contain *a lot*
- * of member variables.
+ * Even without doing deep copies, Nanopb protos contain *a lot* of member
+ * variables (at the time of writing, the largest `sizeof` of a Nanopb proto was
+ * 248).
  */
 template <typename T>
 class Message {
```
var-const: Design of `Message` has changed quite a bit. It no longer even tries to pretend to be immutable.

Firestore/core/test/firebase/firestore/local/local_serializer_test.cc
```diff
@@ -138,12 +138,7 @@ class LocalSerializerTest : public ::testing::Test {
 
   ByteString EncodeMaybeDocument(local::LocalSerializer* serializer,
                                  const MaybeDocument& maybe_doc) {
-    ByteStringWriter writer;
-    firestore_client_MaybeDocument proto =
-        serializer->EncodeMaybeDocument(maybe_doc);
-    writer.WriteNanopbMessage(firestore_client_MaybeDocument_fields, &proto);
-    FreeNanopbMessage(firestore_client_MaybeDocument_fields, &proto);
-    return writer.Release();
+    return serializer->EncodeMaybeDocument(maybe_doc).ToByteString();
```
wilhuff: Nice!

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -130,63 +147,82 @@ class Message {
...
+}
 
-  return MaybeMessage<T>{std::move(message)};
+/**
+ * Serializes the given `message` into a `grpc::ByteBuffer`.
+ *
+ * The lifetime of the return value is entirely independent of the `message`.
+ */
+template <typename T>
+grpc::ByteBuffer ToByteBuffer(const Message<T>& message) {
```
var-const: Would it make more sense as a template?
```cpp
template <typename T, typename U>
T ConvertTo(const Message<U>& message) {
  Writer<T> writer;
  writer.WriteNanopbMessage(message.fields(), message.get());
  return writer.Release();
}

auto grpc_bytes = ConvertTo<grpc::ByteBuffer>(message);
auto nanopb_bytes = ConvertTo<ByteString>(message);
auto str = ConvertTo<std::string>(message);
auto bad = ConvertTo<Foo>(message); // Fails -- no Writer<Foo> defined
```
This would require templatizing `Writer`.

Firestore/core/src/firebase/firestore/nanopb/writer.cc
```diff
@@ -125,6 +123,30 @@ std::string StringWriter::Release() {
...
+}  // namespace
+
+GrpcByteBufferWriter::GrpcByteBufferWriter() {
+  stream_.callback = AppendToGrpcBuffer;
+  stream_.state = &buffer_;
+  stream_.max_size = SIZE_MAX;
+}
+
+grpc::ByteBuffer GrpcByteBufferWriter::Release() {
+  return grpc::ByteBuffer{buffer_.data(), buffer_.size()};
```
var-const: By virtue of implementation, this doesn't let go of ownership. Would it make sense to clear `buffer_` for consistency with the "default" `Writer`?

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
```diff
@@ -306,22 +298,21 @@
...
         Message<google_firestore_v1_BatchGetDocumentsResponse>::TryParse(
-            google_firestore_v1_BatchGetDocumentsResponse_fields, response);
-    if (!maybe_proto.ok()) {
-      return maybe_proto.status();
-    }
+            &grpc_reader);
 
-    const auto& proto = maybe_proto.ValueOrDie();
-    MaybeDocument doc = serializer_.DecodeMaybeDocument(&reader, *proto);
+    nanopb::Reader reader;
```
var-const: I suppose I could turn it into something like `nanopb::Reader reader = grpc_reader.ReleaseReader();`, which would preserve a failed status.

Firestore/core/src/firebase/firestore/local/leveldb_query_cache.mm
```diff
@@ -41,47 +42,49 @@
...
-  NSData* data = [[NSData alloc] initWithBytesNoCopy:(void*)value.data()
-                                              length:value.size()
-                                        freeWhenDone:NO];
-
-  NSError* error;
-  FSTPBTargetGlobal* proto = [FSTPBTargetGlobal parseFromData:data
-                                                        error:&error];
-  if (!proto) {
-    HARD_FAIL("FSTPBTargetGlobal failed to parse: %s", error);
+  if (!status.ok()) {
```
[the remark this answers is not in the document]
var-const: Reverted the change and moved handling of `NotFound` into `ReadMetadata`. There was some trouble with existing migration tests relying on `ReadMetadata` returning nil; to work around that, I've added a test-only `TryReadMetadata` function.

Firestore/core/src/firebase/firestore/nanopb/reader.h
```diff
@@ -134,6 +139,31 @@ class Reader {
   pb_istream_t stream_{};
 };
 
+class GrpcByteBufferReader {
```
var-const: This class isn't particularly elegant, though it gets the job done. Will be happy to hear any suggestions. Some issues I have with it:
* the main one is where it should live. I don't see a clear argument for either gRPC layer or Nanopb layer, since it mixes both;
* it has to partially duplicate `Reader` (I somewhat mitigated this by using composition). The most straightforward way to avoid this is to make `ReadNanopbMessage` virtual. I'm a little hesitant since it would be a performance-unfriendly change, though.

Firestore/core/src/firebase/firestore/nanopb/reader.cc
```diff
@@ -47,6 +53,31 @@ void Reader::ReadNanopbMessage(const pb_field_t fields[], void* dest_struct) {
   }
 }
 
+// GrpcByteBufferReader
+
+GrpcByteBufferReader::GrpcByteBufferReader(const grpc::ByteBuffer& buffer) {
+  std::vector<grpc::Slice> slices;
```
var-const: Unfortunately, gRPC doesn't allow to avoid copying here (at least, I'm not aware of a way to achieve that). Since a copy has to be made, it seemed to make sense to have this reader own it rather than burdening the caller with the conversion.

Firestore/core/src/firebase/firestore/local/leveldb_transaction.h
```diff
@@ -175,6 +178,15 @@ class LevelDbTransaction {
    */
   void Put(std::string key, std::string value);
 
+  /**
```
var-const: Note: I think the Objective-C version of this function (above) had a bug: it's not delegating to the "regular" overload of `Put` and it forgets to remove the `key` from its `deletions_` step. I don't know if it could lead to actual incorrect behavior, though -- can you think of an example?

Firestore/core/src/firebase/firestore/local/leveldb_transaction.cc
```diff
@@ -164,7 +161,7 @@ const WriteOptions& LevelDbTransaction::DefaultWriteOptions() {
 
 void LevelDbTransaction::Put(std::string key, std::string value) {
   deletions_.erase(key);
-  mutations_.emplace(std::make_pair(std::move(key), std::move(value)));
+  mutations_[std::move(key)] = std::move(value);
```
var-const: So I think this was a bug that didn't manifest because the Objective-C overload of this function stores the value in the map in a different way, using `operator[]`; I noticed it because the `nanopb::Message` overload of `Put` that I added delegates to this implementation.

The issue is that `operator[]` overwrites an existing key, while `emplace` doesn't (instead, it returns a value indicating whether an element was added). 8 tests in our suite started failing once this version of `Put` started being used exclusively (either directly or through the `Message` overload I added).

Firestore/core/src/firebase/firestore/nanopb/reader.h
wilhuff: It's a single virtual function call so I wouldn't worry about it. We won't be able to measure the difference.

As far as layering goes, I think there's a dependency argument in favor of putting this in `remote`. `nanopb` is a pretty fundamental building block in the system so adding gRPC as a dependency here makes everything depend on gRPC.

As far as class structure goes: `Reader` wasn't made virtual initially because there wasn't any difference in the implementations (like there are for `Writer`). Making it parallel `Writer` in this regard seems like the way to go. In particular:
  * Make `Reader` a base class, leaving the error handling/context stuff in it
  * Make `StringReader`? for reading from strings/bytes
  * Make `ByteBufferReader` also derive from `Reader`.

In a follow up PR we can actually implement `pb_istream_t` in terms of the vector of slices, which would avoid the copy into `ByteString`.

Firestore/core/src/firebase/firestore/nanopb/writer.cc
wilhuff: `grpc_raw_byte_buffer_create` merely bumps the reference count of all the slices, so `Release` really should clear the underlying buffer in order to make the caller's use of the `ByteBuffer` deterministic. Otherwise the buffer's backing slices won't be destroyed until the writer is destroyed.

Firestore/core/src/firebase/firestore/nanopb/writer.h
```diff
@@ -144,6 +145,17 @@ class StringWriter : public Writer {
   std::string buffer_;
 };
 
+/** Creates a `Writer` that writes into a `grpc::ByteBuffer`. */
+class GrpcByteBufferWriter : public Writer {
```
wilhuff: As with discussion of `Reader`, I think this could likely go in the `remote` directory, alongside all the rest of our grpc-related stuff.

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.h
```diff
@@ -71,8 +71,8 @@ class WatchStreamSerializer {
   nanopb::Message<google_firestore_v1_ListenRequest> EncodeUnwatchRequest(
       model::TargetId target_id) const;
 
-  nanopb::MaybeMessage<google_firestore_v1_ListenResponse> ParseResponse(
-      const grpc::ByteBuffer& buffer) const;
+  nanopb::Message<google_firestore_v1_ListenResponse> ParseResponse(
```
wilhuff: Ideally the read method that's called on this would be virtual such that any `Reader*` could be supplied here, no?

Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.mm
wilhuff: If `GrpcByteBufferReader` is-a `Reader` it would be usable for this purpose here too.

Firestore/core/src/firebase/firestore/remote/watch_stream.cc
```diff
@@ -84,14 +84,12 @@ void WatchStream::NotifyStreamOpen() {
 }
 
 Status WatchStream::NotifyStreamResponse(const grpc::ByteBuffer& message) {
-  MaybeMessage<google_firestore_v1_ListenResponse> maybe_response =
-      watch_serializer_.ParseResponse(message);
-  if (!maybe_response.ok()) {
-    return maybe_response.status();
+  GrpcByteBufferReader grpc_reader{message};
```
wilhuff: Optional: it seems unlikely we'll be mixing string readers and gRPC readers in the same method. Consider naming locals like these just `reader` to save clutter.

Firestore/core/src/firebase/firestore/nanopb/message.h
wilhuff: Elsewhere for conversion style functions we've just used "Make" + return type (e.g. `MakeNSString`, `MakeByteBuffer`). Overload resolution on the argument type disambiguates conversion from different sources.

Making this a template seems sort of overkill-ish. Can you think of any code that would want to be generic on the kind of representation we want for the rendered message? If not simple classes are better.

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -130,63 +147,82 @@ class Message {
...
-  nanopb::Reader reader{maybe_bytes.ValueOrDie()};
-  reader.ReadNanopbMessage(fields, message.get());
-  if (!reader.ok()) {
-    return reader.status();
+template <typename U>
+Message<T> Message<T>::TryParse(U* reader) {
+  Message<T> result;
+  reader->ReadNanopbMessage(result.fields(), result.get());
+  if (!reader->ok()) {
+    // Guarantee that a partially-filled message is never returned.
```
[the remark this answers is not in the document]
wilhuff: I believe nanopb starts to fill the structure in and and then bails if it encounters an error. This means that you could end up with a partially filled message.

Note that if `pb_decode` fails, it calls `pb_release` on the message. Should attempt to avoid doing so again?

Firestore/core/src/firebase/firestore/local/leveldb_transaction.h
```diff
@@ -175,6 +177,14 @@ class LevelDbTransaction {
...
+   * Schedules the row identified by `key` to be set to `value` when this
+   * transaction commits.
+   */
+  void Put(std::string key, const nanopb::ByteString& value) {
+    Put(std::move(key), nanopb::MakeString(value.get()));
+  }
+
```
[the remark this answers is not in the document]
wilhuff: > Excellent point on avoiding conversion, thanks!
> 
> I have added a `GrpcByteBufferWriter` class, though in the current state of the code, it doesn't really help with the tight coupling problem:
> 
> * `GrpcByteBufferWriter` is defined in `writer.{h,cc}` files, similar to other writers;
> * I have made `ToByteBuffer`, etc. free functions rather than member functions on `Message`, but they are still defined in the same file (I think helpers like these _should_ exist somewhere; dealing with writers and readers directly is annoying).

For Writers especially it's much more common to do a one shot conversion. I agree that it's reasonable to make functions for this.

> What are your thoughts re. where this translation point between gRPC and Nanopb should end up?

I think this should end up alongside our other gRPC machinery (i.e. as a peer of `grpc_completion.{h,cc}`.

> Incidentally, `Write/ReadNanopbMessage` function names seem like they need an update: `Message` is now confusing due to the newly-added class, and `Nanopb` seems redundant. Maybe just `Write/ReadProto`?

SG. These could even just be `Read` and `Write`.

Firestore/core/src/firebase/firestore/local/leveldb_transaction.h
wilhuff: The only way this would be problematic would be if we deleted/added a row in the same transaction, but I can't think of case where we would do that. Fixing the bug is definitely worthwhile.

Firestore/core/src/firebase/firestore/local/leveldb_transaction.cc
wilhuff: Good catch!

Firestore/core/src/firebase/firestore/local/leveldb_mutation_queue.mm
```diff
@@ -456,50 +445,44 @@ MutationBatch batch(batch_id, local_write_time, std::move(base_mutations),
...
-                                                            error:&error];
-    if (!proto) {
-      HARD_FAIL("FSTPBMutationQueue failed to parse: %s", error);
-    }
-    return proto;
-  } else if (status.IsNotFound()) {
-    return nil;
+
+  Reader reader{value};
+  reader.set_status(ConvertStatus(status));
```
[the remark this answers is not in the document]
wilhuff: This seems fine.

Firestore/core/src/firebase/firestore/local/leveldb_query_cache.mm
```diff
@@ -41,47 +42,50 @@
...
+                maybe_metadata.status().ToString());
+    }
+  }
+
+  metadata_ = std::move(maybe_metadata).ValueOrDie();
+
+  Reader reader;
+  last_remote_snapshot_version_ = serializer_->DecodeVersion(
+      &reader, metadata_->last_remote_snapshot_version);
+  // FIXME
```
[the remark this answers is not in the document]
wilhuff: Yeah. For now I don't think `Start` failing is recoverable.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -219,42 +220,40 @@ class SerializerTest : public ::testing::Test {
...
   void ExpectFailedStatusDuringFieldValueDecode(
       Status status, const std::vector<uint8_t>& bytes) {
-    Reader reader(bytes);
+    StringReader reader(bytes);
 
-    google_firestore_v1_Value nanopb_proto{};
-    reader.ReadNanopbMessage(google_firestore_v1_Value_fields, &nanopb_proto);
-    serializer.DecodeFieldValue(&reader, nanopb_proto);
-    FreeNanopbMessage(google_firestore_v1_Value_fields, &nanopb_proto);
+    auto message = Message<google_firestore_v1_Value>::TryParse(&reader);
```
var-const: Some refactoring.

Firestore/core/src/firebase/firestore/remote/grpc_nanopb_conversions.h
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2019 Google
```
var-const: 1. I don't see a particularly strong reason to split reader and writer into their own files, what do you think?
2. Feel free to suggest a different name for this file.

wilhuff:
Regarding force check I'm not hugely enthusiastic about it.

When writing code that checks the reader, I don't want to have to do much more than check `reader.ok()`. However, there are plenty of places within the body of the serializer where we make such checks already so it's not enough to do that. Using something like `GetStatusAndClear` requires actually extracting the status, but if you call `reader.GetStatusAndClear().ok()` that destroys the status within the reader so you actually have to do it in two steps, which I don't like.

If we used `reader.ToStatusOr<T>()` extensively, and doing so cleared then I think I'd be for this, but it seems like mostly we have if checks that don't map that cleanly onto `ToStatusOr`.

Firestore/core/src/firebase/firestore/remote/grpc_nanopb_conversions.h
var-const: I went with `grpc_nanopb.h`. My reasoning is that "util" is somewhat vague and is better used as a fallback name.

var-const:
Re. `Reader` -- I was mostly concerned about the whole idea of failure in destructor. It's certainly something to leave for later. If we ever start using C++17, structured bindings could provide one way to make this pattern usable:
```cpp
  auto [ok, status] = reader.ReleaseStatus();
  if (!ok) {
    // Do something with status...
  }
```

PR #4121, opened 2019-10-18
Test nanopb 0.3.9.4

paulb777:
#no-changelog


paulb777:
Tests ran as expected. All tests passed except those where I didn't update the Podfile which failed correctly.

PR #4122, opened 2019-10-19
Migrate FSTPersistenceTestHelpers to C++

wilhuff:
#no-changelog


Firestore/core/test/firebase/firestore/local/persistence_testing.mm
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright 2017 Google
```
wu-hui: 2019

PR #4123, opened 2019-10-19
Optimize `DocumentKey` comparisons

var-const:
This came up while I was investigating the effect of new C++ serializers on performance. When writing large batches, one of the hottest functions turned out to be `firebase::firestore::util::operator==(const DocumentKey& lhs, const DocumentKey& rhs)`, accounting for 15-20% of total CPU time. There are two reasons why comparing keys is currently not very efficient:
* `BasePath::CompareTo`, to which this comparison eventually resolves, mistakenly uses the general-purpose `util::Compare` function rather than the more efficient `util::CompareContainer`;
* the default implementation of `Comparable::operator==` results in two comparisons per element ("less-than" followed by "greater-than"), even though one would be sufficient.

Fixing these issues results in 8-10% CPU time improvement when writing large batches.

To work around the latter, I defined a new mixin `ComparableWithEqual` that is exactly the same as `Comparable`, except its `operator==` delegates to `T::IsEqualTo`, avoiding the extra comparison.

Some measurements (all done on the emulator in release mode with no logging) follow; the test case was to write 10 batches of 500 very small documents. Note that I found wall clock to fluctuate too much to be reliable and instead use the measurements from XCode's Time Profiler, which (presumably) shows CPU time. Each measurement shows total CPU time spent on `firebase::firestore::util::operator==` (all instantiations, not just for `DocumentKey`), followed by the percentage of time spent in this function compared to total runtime:
* no optimization:
    * 275ms/18.8%;
    * 245ms/16.9%;
    * 252ms/18.5%.
* `CompareContainer` optimization:
    * 124ms/9.7%;
    * 110ms/8.5%;
    * 128ms/10.0%.
* `CompareContainer` + `ComparableWithEqual` optimizations:
    * 95ms/7.3%;
    * 100ms/8.0%;
    * 112ms/9.2%.

### DIFF
--- a/Firestore/core/src/firebase/firestore/util/comparison.h
+++ b/Firestore/core/src/firebase/firestore/util/comparison.h
@@ -353,6 +353,35 @@ class Comparable {
   }
 };
 
+/**
+ * Same as `Comparable`, but `operator==` is delegated to a function called
+ * `IsEqual`. This is an optimization that avoids doing an extra comparison when
+ * comparing for equality (`Comparable` would have to check for both "less-than"
+ * and "greater-than" to determine whether two values are equal).
+ */
+template <typename T>
+class ComparableWithEqual {
+ public:
+  friend bool operator==(const T& lhs, const T& rhs) {
+    return lhs.IsEqual(rhs);
+  }
+  friend bool operator!=(const T& lhs, const T& rhs) {
+    return !(lhs == rhs);
+  }
+  friend bool operator<(const T& lhs, const T& rhs) {
+    return Ascending(lhs.CompareTo(rhs));
+  }
+  friend bool operator>(const T& lhs, const T& rhs) {
+    return Descending(lhs.CompareTo(rhs));
+  }
+  friend bool operator<=(const T& lhs, const T& rhs) {
+    return !(rhs < lhs);
+  }
+  friend bool operator>=(const T& lhs, const T& rhs) {
+    return !(lhs < rhs);
+  }
+};
+
 }  // namespace util
 }  // namespace firestore
 }  // namespace firebase
--- a/Firestore/core/src/firebase/firestore/util/comparison.h
+++ b/Firestore/core/src/firebase/firestore/util/comparison.h
@@ -355,15 +355,15 @@ class Comparable {
 
 /**
  * Same as `Comparable`, but `operator==` is delegated to a function called
- * `IsEqual`. This is an optimization that avoids doing an extra comparison when
- * comparing for equality (`Comparable` would have to check for both "less-than"
- * and "greater-than" to determine whether two values are equal).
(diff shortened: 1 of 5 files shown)


Firestore/core/src/firebase/firestore/util/comparison.h
```diff
@@ -353,6 +353,35 @@ class Comparable {
...
 };
 
+/**
+ * Same as `Comparable`, but `operator==` is delegated to a function called
+ * `IsEqualTo`. This is an optimization that avoids doing an extra comparison
+ * when comparing for equality (`Comparable` would have to check for both
+ * "less-than" and "greater-than" to determine whether two values are equal).
+ */
+template <typename T>
+class ComparableWithEqual {
```
var-const: One alternative is to only have `Comparable`, use `enable_if` to switch based on whether `T` defines `IsEqualTo`. Seems overkill (and more error prone, because forgetting to define `IsEqualTo` will not be an error).

Firestore/core/src/firebase/firestore/util/comparison.h
```diff
@@ -353,6 +353,35 @@ class Comparable {
   }
 };
 
+/**
```
var-const: Unlike `CompareContainer`, which is a slam dunk, this optimization is less certain. Let me know if you think it's justified by the numbers I measured.

Firestore/core/src/firebase/firestore/util/comparison.h
wilhuff: That doesn’t seem warranted for just the handful of cases we have here. `std::enable_if` should be an absolute last resort.

Firestore/core/src/firebase/firestore/util/comparison.h
wilhuff: I think the principal here is reasonable, but rather than delegate to a new, nonstandard method we might define a `Comparable` variant that just adds inequality operators without defining `operator==` at all. Then types can define it for themselves. Maybe `InequalityComparable`? WDYT?

var-const: Done. I'm not sure how much I like `InequalityComparable` as a name, but I cannot think of a better suggestion either.

var-const:
@wilhuff Do you think this warrants a changelog entry? On one hand, up to 10% CPU time reduction is non-negligible; on the other hand, wall time is dominated by network, so it's unlikely to be very noticeable to the end users, and I don't want to "oversell" the effects of this change.

wilhuff:
Wall time is what matters and since the change isn’t significant I’d lean toward skipping the changelog entry.

Firestore/core/src/firebase/firestore/util/comparison.h
wilhuff: Yeah I'm in the same boat--I couldn't think of anything better either.

PR #4126, opened 2019-10-21
Add checks for DO NOT SUBMIT

morganchen12:


Dangerfile
```diff
@@ -67,6 +81,18 @@ has_changelog_changes = hasChangesIn(["CHANGELOG"])
 # Whether or not the LICENSE file has been modified or deleted.
 has_license_changes = didModify(["LICENSE"])
 
+# Common permutations of DO NOT SUBMIT.
+do_not_submit_variations = [
```
wilhuff: Ideally anything that would cause a PR to be rejected would be checked locally when you run `scripts/check.sh`. For example we're already disallowing `@import` in [check_no_module_imports.hs](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check_no_module_imports.sh).

Is it possible to run danger pre-push?

morganchen12: Danger can be [run locally](https://danger.systems/guides/troubleshooting.html#i-want-to-work-locally-on-my-dangerfile), but not before pushing to github.

Should we keep this as a script instead so contributors don't have to push to GitHub to run the check?

wilhuff: Hrm. I guess the check is checking the PR description itself, which can't be done before pushing.

Also, splitting file contents checking from the PR description seems like more trouble than it's worth.

Upon reflection this seems like it was a bad suggestion.

Dangerfile
```diff
@@ -128,3 +154,7 @@ suggested_labels = labelsForModifiedFiles()
 if !suggested_labels.empty?
   addLabels(suggested_labels)
 end
+
+# Warn if the PR contains DO NOT SUBMIT. Don't fail here, since it's still nice
+# to run travis on WIP pull requests.
+warn("PR contents contain 'do not submit'.") if do_not_submit
```
wilhuff: This is going to become a mess. Could we group the issuing of the warning with the code that does the checking?

Could we make these different checks functions and put things like `do_not_submit` into locals within a function?

morganchen12: I'll look into ways to make this file less spaghetti.

ryanwilson:
@morganchen12 is this still relevant or should it be closed?

morganchen12:
We can close it, it's a bit of a non-feature.

PR #4130, opened 2019-10-21
Prepare FirebaseFirestoreSwift for initial release

wilhuff:


wilhuff:
As far as where to point the tag, I think we can point it at the last commit on release-6.11.0, just like all the other pods. There haven't been any changes to `Firestore/Swift` since that branch was cut.

paulb777:
LGTM. 

Please retarget to the release-6.11.0 branch.  I'll merge, tag, and publish from there after publishing the rest of 6.11.0.

wilhuff:
Does this need an update to `Releases/Manifests/6.11.0.json` or any similar updates?

FirebaseFirestoreSwift.podspec
```diff
@@ -36,5 +36,5 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
     'Firestore/third_party/FirestoreEncoder/*.swift',
   ]
 
-  s.dependency 'FirebaseFirestore', ">= 0.10.0"
+  s.dependency 'FirebaseFirestore', ">= 1.6.1"
```
paulb777: Typical Firebase practice is to allow minor version updates of dependencies. Are you sure about only allowing patches?

FirebaseFirestoreSwift.podspec
```diff
@@ -36,5 +36,5 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
     'Firestore/third_party/FirestoreEncoder/*.swift',
   ]
 
-  s.dependency 'FirebaseFirestore', ">= 1.6.1"
+  s.dependency 'FirebaseFirestore', "~> 1.6.1"
```
paulb777: Should be `~> 1.6` to float minor version

wilhuff: Done. My primary concern was to disallow interoperation with 1.6.0, which is untested.

PR #4131, opened 2019-10-22
remove NSException from firestore client

wu-hui:


wu-hui:
> This is fine as an intermediate state, but really we should be putting ourselves into a failed state and erroring operations that attempt to use persistence.

Which is basically everything except transactions?

Alternatively we could fall back to memory persistence..

PR #4135, opened 2019-10-22
Update changelogs for M58

morganchen12:
#no-changelog

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # v1.6.1
 - [fixed] Fix a race condition that could cause a segmentation fault during
-  client initialization. (#4091)
+  client initialization.
 
 # v1.6.0
 - [feature] Added an `addSnapshotsInSyncListener()` method to
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # v1.6.1
 - [fixed] Fix a race condition that could cause a segmentation fault during
-  client initialization.
+  client initialization. (#4091)
 
 # v1.6.0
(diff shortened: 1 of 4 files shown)


Firestore/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 # v1.6.1
 - [fixed] Fix a race condition that could cause a segmentation fault during
-  client initialization.
+  client initialization. (#4091)
```
wilhuff: I'm leaning against including this link (which is why I omitted it in the first place).

The way I think of these is that they're links for individuals to learn more about the underlying issue to see if it's similar to something they might have encountered, not necessarily just a pointer to any artifact related to the change.

In this case, there was no user reported issue for this in the iOS repo and we generally haven't been linking to PRs, especially ones without any discussion. The actual issue was reported in https://github.com/firebase/firebase-android-sdk/issues/875, but that's not too helpful for iOS users. We made this fix by analogy rather than because anyone had actually reported it.

In short this link is a rabbit hole without any real payoff which I estimate does more harm than good.

I'm open to changing this position, especially if there's some kind of policy in effect.

morganchen12: There's no official policy here, I was just adding changes because I thought they were missing. I'll exclude this one. The release notes page is the [largest source of traffic to this repo from devsite](https://github.com/firebase/firebase-ios-sdk/graphs/traffic?referrer=firebase.google.com#top-domains), so my approach was to include a GitHub link whenever one is available.

There's been some talk among the tech writers of having a more formal release notes policy, so this might change in the future. In general, I think it's ok for GitHub links in release notes to be rabbit holes for curious users to disambiguate changes (for example, this release's initialization race condition fix is listed in three changelogs despite being a single change) and in rare cases look through the code changes themselves.

PR #4136, opened 2019-10-22
Port FSTRemoteDocumentCacheTests to C++

wilhuff:


Firestore/core/test/firebase/firestore/local/leveldb_remote_document_cache_test.mm
```diff
@@ -0,0 +1,81 @@
...
+// document keys.
+const char* kDummy = "1";
+
+/**
+ * Writes a dummy row that looks like a remote document key but is different
+ * enough that it shouldn't be picked up in scans of the table.
+ */
+void WriteDummyRow(LevelDbPersistence* db,
+                   std::initializer_list<std::string> segments) {
+  // TODO(wilhuff): Find some way to share local::(anonymous)::Writer
```
schmidt-sebastian: The dummy row produced by this function is different from `writeDummyRowWithSegments`. I am sure there is a valid reason, but it is not obvious to me. Do you mind explaining?

Firestore/core/test/firebase/firestore/local/leveldb_remote_document_cache_test.mm
```diff
@@ -0,0 +1,81 @@
...
+  OrderedCode::WriteSignedNumIncreasing(&key, 0);  // Terminator
+
+  db->ptr()->Put(WriteOptions(), key, kDummy);
+}
+
+std::unique_ptr<Persistence> PersistenceFactory() {
+  auto persistence = LevelDbPersistenceForTesting();
+
+  WriteDummyRow(persistence.get(), {"remote_documentr", "foo", "bar"});
+  WriteDummyRow(persistence.get(), {"remote_documentsa", "foo", "bar"});
```
schmidt-sebastian: Nit: These looked like typos to me at first glance. Could we make it remote_document_r and remote_document_sa?

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace local {
+
+class Persistence;
+class RemoteDocumentCache;
+
+using FactoryFunc = std::unique_ptr<Persistence> (*)();
+
+/**
+ * These are tests for any implementation of the FSTRemoteDocumentCache
```
schmidt-sebastian: Drop "FST" here and below (and replace "protocol").

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.h
```diff
@@ -0,0 +1,69 @@
...
+ * protocol.
+ *
+ * To test a specific implementation of FSTRemoteDocumentCache:
+ *
+ * + Subclass RemoteDocumentCacheTest
+ * + override -set_up, assigning to remote_document_cache and persistence
+ * + override -tear_down, cleaning up remote_document_cache and persistence
```
schmidt-sebastian: These steps don't seem to apply anymore.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -0,0 +1,222 @@
...
+using testutil::Query;
+
+const char* kDocPath = "a/b";
+const char* kLongDocPath = "a/b/c/d/e/f";
+const int kVersion = 42;
+FieldValue::Map kDocData;
+
+template <typename MapType>
+std::vector<MaybeDocument> ActualDocuments(const MapType& docs) {
+  std::vector<MaybeDocument> result;
```
schmidt-sebastian: Neither the name nor the function signature convey a whole lot about what this method does. Would it be possible to not use a template here (the function always seems to convert from an OptionalMaybeDocumentMap to `std::vector<MaybeDocument>`) and to use a more descriptive name (`ExtractDocuments` or `ExtractExistingDocuments`)?

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -0,0 +1,222 @@
...
+        };
+        OptionalMaybeDocumentMap read = cache_->GetAll(DocumentKeySet{
+            testutil::Key(kDocPath),
+            testutil::Key(kLongDocPath),
+            testutil::Key("foo/nonexistent"),
+        });
+        EXPECT_THAT(read, HasAtLeastDocs(written));
+        auto found = read.find(DocumentKey::FromPathString("foo/nonexistent"));
+        ASSERT_TRUE(found != read.end());
+        ASSERT_EQ(absl::nullopt, found->second);
```
schmidt-sebastian: This seems to match what `HasExactlyDocs` does. Since it is also the only caller of HasAtLeastDocs we could consider using HasExactlyDocs and dropping HasAtLeastDocs.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -0,0 +1,222 @@
...
+    SetTestDocument("c/1");
+
+    core::Query query = Query("b");
+    DocumentMap results = cache_->GetMatching(query);
+    std::vector<Document> docs = {
+        Doc("b/1", kVersion, kDocData),
+        Doc("b/2", kVersion, kDocData),
+    };
+    EXPECT_THAT(results.underlying_map(), HasExactlyDocs(docs));
+    //    ExpectMapHasDocs(results.underlying_map(), docs, /* exactly= */ true);
```
schmidt-sebastian: Please remove.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -0,0 +1,222 @@
...
+Document RemoteDocumentCacheTest::SetTestDocument(
+    const absl::string_view path) {
+  Document doc = Doc(path, kVersion, kDocData);
+  cache_->Add(doc);
+  return doc;
+}
+
+void RemoteDocumentCacheTest::SetAndReadTestDocument(
+    const absl::string_view path) {
+  persistence_->Run("SetAndReadTestDocument", [&] {
```
schmidt-sebastian: This method is only used once and could potentially be inlined.

Firestore/core/test/firebase/firestore/local/leveldb_remote_document_cache_test.mm
wilhuff: `writeDummyRowWithSegments` predates the change to self-describing leveldb keys and I missed this change then.

The idea was to add keys that would bound the table to ensure that we weren't reading before the start or past the end.

I've added comments to this effect in the code.

Firestore/core/test/firebase/firestore/local/leveldb_remote_document_cache_test.mm
wilhuff: The goal of the first one was to be before the first possible remote document key. Changed it "remote_document" and "remote_document_a" along with comments indicating what's going on here.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.h
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.h
wilhuff: Updated.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
wilhuff: Somehow I had convinced myself the template was required but clearly it isn't. Changed to `OptionalMaybeDocumentMap`.

Good idea re the name.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
wilhuff: I don't think this is true.

This is checking that `read` includes a document entry for `"foo/nonexistent"` and that its value is `optional::none`. This is something that `HasExactlyDocs` and `HasAtLeastDocs` ignore.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
wilhuff: It's used by both `SetAndReadADocument` and `SetAndReadADocumentAtDeepPath`.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
schmidt-sebastian: Got it, I misread `ASSERT_TRUE(found != read.end())` and thought it indicated the opposite.

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
schmidt-sebastian: The update here is causing a build break for me. I temporarily reverted it out in my branch.

wilhuff: Yeah, it turns out that the template was required because we do this for both `OptionalMaybeDocumentMap` and `MaybeDocumentMap`.

The reason this code works in both cases is that `optional<MaybeDocument>` will implicitly convert from `MaybeDocument`. This means that in the case of the `MaybeDocumentMap` the assignment to the optional creates a temporary that has a guaranteed value and the if statement is always true.

Another way to resolve this would be to provide overloads for both map types, but this seems annoying.

PR #4137, opened 2019-10-22
Fix component startup time.

ryanwilson:
The `configureWithApp:` call shouldn't have been added - the
component registration system can provide proper configuration
instead of relying on a static `configureWithApp:` call. Many of
the SDKs relied on singletons being available there, which wasn't
necessary. Instead, this should happen in the component creation
block and use the instantiationTiming parameter on
registerInternalLibrary.

This replaces #3147 since other Core and IID changes were made since then.


Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -1884,19 +1884,12 @@ - (BOOL)getUser:(FIRUser *_Nullable *)outUser
     return [[FIRAuth alloc] initWithApp:container.app];
   };
   FIRComponent *authInterop = [FIRComponent componentWithProtocol:@protocol(FIRAuthInterop)
+                                              instantiationTiming:FIRInstantiationTimingAlwaysEager
```
paulb777: Why AlwaysEager?

Firebase/Messaging/FIRMessaging.m
```diff
@@ -233,28 +240,19 @@ + (void)load {
                                  withInstanceID:[FIRInstanceID instanceID]
                                withUserDefaults:[GULUserDefaults standardUserDefaults]];
     [messaging start];
+    [messaging configureNotificationSwizzlingIfEnabled];
     return messaging;
   };
   FIRComponent *messagingProvider =
       [FIRComponent componentWithProtocol:@protocol(FIRMessagingInstanceProvider)
-                      instantiationTiming:FIRInstantiationTimingLazy
+                      instantiationTiming:FIRInstantiationTimingEagerInDefaultApp
```
paulb777: Why this change from lazy to eager?

Firebase/Auth/Source/Auth/FIRAuth.m
ryanwilson: It mimics the previous behaviour removed on the previous line 1897:

 https://github.com/firebase/firebase-ios-sdk/pull/4137/files#diff-f33c561b2a3430b947e6da076758c6d9L1897

every time `[FIRApp configure]` is called, we called `[FIRAuth authWithApp:]` which instantiated the instance. 

The reason it's `AlwaysEager` and not `EagerInDefaultApp` is because `FIRAuth` works with multiple `FIRApp`s.

Firebase/Messaging/FIRMessaging.m
ryanwilson: It mimics the previous behaviour of `configureWithApp:`, where Messaging would instantiate itself as soon as `configureWithApp:` is called.

Firebase/Messaging/FIRMessaging.m
```diff
@@ -233,28 +240,19 @@ + (void)load {
...
                              dependencies:@[ analyticsDep ]
                            creationBlock:creationBlock];
 
   return @[ messagingProvider ];
 }
 
-+ (void)configureWithApp:(FIRApp *)app {
```
chliangGoogle: do we need to fix the same for other SDK as well, such as InstanceID, remote config etc?

Firebase/Auth/Source/Auth/FIRAuth.m
renkelvin: Yeah, `FIRAuth` works with multiple `FIRApp` s.

Firebase/Messaging/FIRMessaging.m
ryanwilson: Apologies for the delay - IID was fixed in #4030, will be addressing the other SDKs in a separate PR.

ryanwilson:
Merged laster to re-run travis and merge this shortly.

ryanwilson:
Merging this now since everything is green - will fix anything if there was a conflict since last merge.

PR #4140, opened 2019-10-22
style.sh generated changes

wilhuff:
These didn't get flagged with your last PR.

### DIFF
--- a/ZipBuilder/Sources/UpdateFirebasePod/FirebasePod.swift
+++ b/ZipBuilder/Sources/UpdateFirebasePod/FirebasePod.swift
@@ -20,7 +20,6 @@ import ManifestReader
 
 /// Updates the Firebase Pod with a release's version set.
 struct FirebasePod {
-
   /// Relevant paths in the filesystem to build the release directory.
   struct FilesystemPaths {
     // MARK: - Required Paths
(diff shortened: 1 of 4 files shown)


wilhuff:
I'm not sure. I wonder if we've broken the exit status of `scripts/style.sh`?

paulb777:
The output of swiftformat has changed.  This line - https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/style.sh#L176 - no longer works since swiftformat returns a string like 
`0/1 files would have been formatted.` or `1/1 files would have been formatted.`  

It should instead check for a non-zero value at the beginning of that string.

wilhuff:
OK, I've verified that `check.sh` would have rejected your earlier changes.

```sh
git checkout wilhuff/swift-format
git revert 43df59489181a7ec946aa231450d138847eaa55e  # the style fix in this PR
scripts/check.sh --test-only ac333d1822e5108ab10fe2a0d2f338affaf7d4b7^  # just prior to the PR adding the formatting issue
```

Yields the expected result:

```
ZipBuilder/Sources/UpdateFirebasePod/FirebasePod.swift needs formatting.
ZipBuilder/Sources/UpdateFirebasePod/LaunchArgs.swift needs formatting.
ZipBuilder/Sources/UpdateFirebasePod/main.swift needs formatting.
Proposed commit is not style compliant.
Run scripts/style.sh and git add the result.
```

PR #4142, opened 2019-10-23
Unit test fixes

schmidt-sebastian:
The tests are all just broken in the branch.

### DIFF
--- a/Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
+++ b/Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
@@ -63,8 +63,8 @@ FieldValue::Map kDocData;
 /**
  * Extracts all the actual MaybeDocument instances from the given document map
  */
-std::vector<MaybeDocument> ExtractDocuments(
-    const OptionalMaybeDocumentMap& docs) {
+template <typename MapType>
+std::vector<MaybeDocument> ExtractDocuments(const MapType& docs) {
   std::vector<MaybeDocument> result;
   for (const auto& kv : docs) {
     const absl::optional<MaybeDocument>& doc = kv.second;
@@ -147,7 +147,7 @@ TEST_P(RemoteDocumentCacheTest, SetAndReadADocumentAtDeepPath) {
 TEST_P(RemoteDocumentCacheTest, SetAndReadDeletedDocument) {
   persistence_->Run("test_set_and_read_deleted_document", [&] {
     absl::optional<MaybeDocument> deleted_doc = DeletedDoc(kDocPath, kVersion);
-    cache_->Add(*deleted_doc);
+    cache_->Add(*deleted_doc, deleted_doc->version());
 
     ASSERT_EQ(cache_->Get(testutil::Key(kDocPath)), deleted_doc);
   });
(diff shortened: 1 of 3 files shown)


Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -63,8 +63,8 @@
 /**
  * Extracts all the actual MaybeDocument instances from the given document map
  */
-std::vector<MaybeDocument> ExtractDocuments(
-    const OptionalMaybeDocumentMap& docs) {
+template <typename MapType>
+std::vector<MaybeDocument> ExtractDocuments(const MapType& docs) {
```
schmidt-sebastian: This is a fix for https://github.com/firebase/firebase-ios-sdk/pull/4136/files/043cdc45b02b9d365aeadbe9b0ff9dffdb03aa55#diff-c4c9d440fb2b6a0d5966b70d73695c31 (a PR I merged into indexfree-master before it was merged to master)

Firestore/core/test/firebase/firestore/local/remote_document_cache_test.mm
```diff
@@ -147,7 +147,7 @@
 TEST_P(RemoteDocumentCacheTest, SetAndReadDeletedDocument) {
   persistence_->Run("test_set_and_read_deleted_document", [&] {
     absl::optional<MaybeDocument> deleted_doc = DeletedDoc(kDocPath, kVersion);
-    cache_->Add(*deleted_doc);
+    cache_->Add(*deleted_doc, deleted_doc->version());
```
schmidt-sebastian: Changes here and below are to make the new RemoteDocumentCache unit tests compile in the branch.

Firestore/core/test/firebase/firestore/testutil/testutil.h
```diff
@@ -381,8 +381,8 @@ inline core::FieldFilter Filter(absl::string_view key,
 
 inline core::FieldFilter Filter(absl::string_view key,
                                 absl::string_view op,
-                                const std::string& value) {
-  return Filter(key, op, model::FieldValue::FromString(value));
+                                const char* value) {
```
schmidt-sebastian: This fixes my test failures since string literals ended up creating boolean filters with the new boolean overload just below.

Firestore/core/test/firebase/firestore/testutil/testutil.h
```diff
@@ -381,8 +381,8 @@ inline core::FieldFilter Filter(absl::string_view key,
...
+  return Filter(key, op, model::FieldValue::FromString(std::string(value)));
 }
 
 inline core::FieldFilter Filter(absl::string_view key,
```
wilhuff: This is actually a problematic definition. We shouldn't allow conversion from pointers to bool. The way to ensure it is to disable the overload when the type of the argument isn't an explicit bool.

The way to do that would be:

```c++
template <typename T,
          typename = typename std::enable_if<std::is_same<bool, T>{}>::type>
inline core::FieldFilter Filter(absl::string_view key,
                                absl::string_view op,
                                T bool_value) {
  return Filter(key, op, model::FieldValue::FromBoolean(bool_value));
}
```

Firestore/core/test/firebase/firestore/testutil/testutil.h
```diff
@@ -381,8 +381,8 @@ inline core::FieldFilter Filter(absl::string_view key,
...
                                 absl::string_view op,
-                                const std::string& value) {
-  return Filter(key, op, model::FieldValue::FromString(value));
+                                const char* value) {
+  return Filter(key, op, model::FieldValue::FromString(std::string(value)));
```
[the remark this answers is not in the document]
schmidt-sebastian: Removed the explicit `std::string` constructor.

Firestore/core/test/firebase/firestore/testutil/testutil.h
schmidt-sebastian: This is definitely something I would never have thought of, but now I get credit for it via `git blame`. Wheee.

PR #4145, opened 2019-10-23
Update nanopb in GDT and FIRCD to 0.3.9.4

mikehaney24:


paulb777:
Let's resolve #4154 before landing this.

PR #4147, opened 2019-10-23
Consolidate CoreDiagnostics pods into Core

paulb777:
Follow on PRs will add a runtime option to disable diagnostics and update the zipbuilder


paulb777:
After discussion with @wilhuff, going to abandon this for now to avoid breaking the use case for binary distribution users of being able to avoid linking the CoreDiagnostics and GDT frameworks.

PR #4149, opened 2019-10-23
Stop using statics in header

paulb777:
So they're not added to the size of every implementation file that includes them.

Update if_changed.sh to be more accurate for CoreDiagnostics changes and to make changes to itself testable.

### DIFF
--- a/Interop/CoreDiagnostics/Public/FIRCoreDiagnosticsData.h
+++ b/Interop/CoreDiagnostics/Public/FIRCoreDiagnosticsData.h
@@ -19,36 +19,34 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /** If present, is a BOOL wrapped in an NSNumber. */
-static NSString *const kFIRCDIsDataCollectionDefaultEnabledKey =
-    @"FIRCDIsDataCollectionDefaultEnabledKey";
+#define kFIRCDIsDataCollectionDefaultEnabledKey @"FIRCDIsDataCollectionDefaultEnabledKey"
 
 /** If present, is an int32_t wrapped in an NSNumber. */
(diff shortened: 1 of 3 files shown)


Interop/CoreDiagnostics/Public/FIRCoreDiagnosticsData.h
```diff
@@ -19,36 +19,34 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /** If present, is a BOOL wrapped in an NSNumber. */
-static NSString *const kFIRCDIsDataCollectionDefaultEnabledKey =
-    @"FIRCDIsDataCollectionDefaultEnabledKey";
+#define kFIRCDIsDataCollectionDefaultEnabledKey @"FIRCDIsDataCollectionDefaultEnabledKey"
 
 /** If present, is an int32_t wrapped in an NSNumber. */
-static NSString *const kFIRCDConfigurationTypeKey = @"FIRCDConfigurationTypeKey";
+#define kFIRCDConfigurationTypeKey @"FIRCDConfigurationTypeKey"
```
wilhuff: Are you able to measure any actual reduction in size using this mechanism? This also results in the value being inlined into every translation, but just as a string literal rather than a static variable.

The way to do this within the language is to declare these here, like so:

```objc
extern NSString *const kFIRCDConfigurationTypeKey;
```
Then in the associated .m file:
```objc
NSString *const kFIRCDConfigurationTypeKey = @"FIRCDConfigurationTypeKey";
```

paulb777: The linker should consolidate the string usages. However, in this case there's only one usage in each library, so no need. This change keeps the strings from being added to all the other files including where the strings aren't being used.  The extern methodology doesn't work across optional interop pod dependencies.

PR #4151, opened 2019-10-23
Remove Objective-C serializers and make local/ and remote/ (almost) fully C++

var-const:
There are more sources that can be `.cc` now (e.g., in API); these are left for a follow-up PR.

The dependency on Objective-C proto library cannot be removed just yet due to current lack of pretty-printing support for Nanopb.


var-const:
@wilhuff Apologies if this clashes with your CMake PR in progress.

Firestore/core/src/firebase/firestore/local/leveldb_mutation_queue.cc
```diff
@@ -438,9 +438,10 @@ MutationBatch batch(batch_id, local_write_time, std::move(base_mutations),
     mutation_iterator->Seek(mutation_key);
     if (!mutation_iterator->Valid() ||
         mutation_iterator->key() != mutation_key) {
-      HARD_FAIL("Dangling document-mutation reference found: "
-                "Missing batch %s; seeking there found %s",
-                DescribeKey(mutation_key), DescribeKey(mutation_iterator));
+      HARD_FAIL(
+          "Dangling document-mutation reference found: "
```
wilhuff: Optional: These seem like they were poorly re-wrapped. If you rejoin the string literal and rerun `clang-format` does it result in a more natural wrapping?

var-const: Done.

PR #4158, opened 2019-10-24
Add support for in-memory and other types of stores

mikehaney24:


paulb777:
Fix travis

mikehaney24:
> Fix travis

Filed #4173, because this looks like a problem with check.sh running on travis. check.sh succeeds locally, and each individual check succeeds locally. Both origin and local have the same histories.

Should I comment out the majority of check.sh so that I can proceed?

image

mikehaney24:
I've removed the breaking parts of check.sh. PTAL

GoogleDataTransport/GDTCORLibrary/GDTCORSqlite.m
```diff
@@ -36,15 +36,15 @@ static BOOL BindCheck(SQLITE_API int (^bindBlock)(void)) {
...
+BOOL GDTCORSQLOpenDB(sqlite3 **db, NSString *path) {
+  if (path == nil || path.length == 0) {
     GDTCORLogError(GDTCORMCEDatabaseError, @"%@", @"A filename for the sqlite db must not be nil");
     return NO;
   }
 
   int flags = SQLITE_OPEN_CREATE | SQLITE_OPEN_READWRITE | SQLITE_OPEN_FILEPROTECTION_COMPLETE |
               SQLITE_OPEN_FULLMUTEX;
-  return sqlite3_open_v2([fileURL path].UTF8String, db, flags, NULL) == SQLITE_OK;
+  return sqlite3_open_v2(path.UTF8String, db, flags, NULL) == SQLITE_OK;
```
chliangGoogle: Also are database are operated on main thread? Normally I kick a dispatch queue to move most of the sqlite operations off main thread to a serial queue. And sometimes because a blocking call is needed when reading from database, I will make a compromise to make all write operations async and read operations sync. You can check remote config and InstanceID for reference.

mikehaney24: The database will run on its own queue, this helper class is not thread-safe. The database PR will come immediately after this one, I'm just breaking down the changes to be easier to review.

PR #4161, opened 2019-10-24
Add --analyze to pod lib lint tests

### LINKED ISSUE #1744: Add Analyze checking to travis
To avoid recurrences of #1735 and #1743

paulb777:
Fix #1744

Now that we're using CocoaPods 1.8.x, we can take advantage of the new `--analyze` option to run the Xcode analyzer on presubmits.

This PR adds it, along with disabling for FIAM(#4163), Functions(#4164) and Firestore(#4165).

#no-changelog

### DIFF
--- a/scripts/pod_lib_lint.rb
+++ b/scripts/pod_lib_lint.rb
@@ -31,7 +31,7 @@ def usage()
   options can be any options for pod spec lint
 
   script options:
-    --skip-analyze: don't run Xcod analyze on this podspec
+    --skip-analyze: don't run Xcode analyze on this podspec
     --ignore-local-podspecs: list of podspecs that should not be added to
       "--include-podspecs" list. If not specified, then all podspec
       dependencies will be passed to "--include-podspecs".
--- a/scripts/pod_lib_lint.rb
+++ b/scripts/pod_lib_lint.rb
@@ -47,7 +47,7 @@ def main(args)
 
   STDOUT.sync = true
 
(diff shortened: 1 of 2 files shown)


jasonhu-g:
🙌 👏 👍- Good thing hack week did not start yet cause I was gonna look into integrate these analysis checks with Travis.  I am going to disable the analysis runs on my XCode Server machine once this is merged.

scripts/pod_lib_lint.rb
```diff
@@ -52,9 +53,13 @@ def main(args)
   # to the pod command.
   pod_args = []
   ignore_local_podspecs = []
+  skip_analyze = false
```
wilhuff: Optional: since the point of this is to compute the `--analyze` flag we're passing down to `pod lib lint`, you might make this `analyze = true` to avoid the double negative below (the "not skip").

scripts/pod_lib_lint.rb
```diff
@@ -31,6 +31,7 @@ def usage()
   options can be any options for pod spec lint
 
   script options:
+    --skip-analyze: don't run Xcode analyze on this podspec
```
wilhuff: Optional: conventionally inverted boolean argument flags are spelled with a no prefix (i.e. for `--analyze` the inverted form would conventionally be `--no-analyze`). Consider following this pattern here.

ryanwilson: Done, great idea.

scripts/pod_lib_lint.rb
ryanwilson: Done, thanks.

Issue #4167, opened 2019-10-25
Crash at [FIRInstanceIDTokenManager checkTokenRefreshPolicyWithIID:] function


phaninderkumar:
**Environment**:

  * Xcode version: 11.0
  * Firebase SDK version: 6.11.0
  * Firebase Component: Core, Messaging
  * Component version: 6.3.2, ~> 4.1.7

**Problem**:
My project keeps crashing at [FIRInstanceIDTokenManager checkTokenRefreshPolicyWithIID:] function.
My app collects location in suspended state if it exits a region. My assumption is this crashes happens at that time. Not always though. 

**Steps to reproduce:**

Configure the firebase project and make sure the app is launched from suspended state multiple times

Attaching my crash log here
[crashReport.txt](https://github.com/firebase/firebase-ios-sdk/files/3770695/crashReport.txt)

phaninderkumar:
Here is the list of Firebase pods being used:

  - Firebase/Core (6.11.0):
    - Firebase/CoreOnly
    - FirebaseAnalytics (= 6.1.3)
  - Firebase/CoreOnly (6.11.0):
    - FirebaseCore (= 6.3.2)
  - Firebase/Messaging (6.11.0):
    - Firebase/CoreOnly
    - FirebaseMessaging (~> 4.1.7)
  - FirebaseAnalytics (6.1.3):
    - FirebaseCore (~> 6.3)
    - FirebaseInstanceID (~> 4.2)
    - GoogleAppMeasurement (= 6.1.3)
    - GoogleUtilities/AppDelegateSwizzler (~> 6.0)
    - GoogleUtilities/MethodSwizzler (~> 6.0)
    - GoogleUtilities/Network (~> 6.0)
    - "GoogleUtilities/NSData+zlib (~> 6.0)"
    - nanopb (~> 0.3.901)
  - FirebaseAnalyticsInterop (1.4.0)
  - FirebaseCore (6.3.2):
    - FirebaseCoreDiagnostics (~> 1.0)
    - FirebaseCoreDiagnosticsInterop (~> 1.0)
    - GoogleUtilities/Environment (~> 6.2)
    - GoogleUtilities/Logger (~> 6.2)
  - FirebaseCoreDiagnostics (1.1.1):
    - FirebaseCoreDiagnosticsInterop (~> 1.0)
    - GoogleDataTransportCCTSupport (~> 1.0)
    - GoogleUtilities/Environment (~> 6.2)
[...]

PR #4172, opened 2019-10-25
update keychain access level

chliangGoogle:
#4167 user reports getting IID becomes nil when a token exists. We suspect this is due to a security issue while keychain access level is set to kSecAttrAccessibleAlwaysThisDeviceOnly that is deprecated in iOS 13. Neitherless we should update it to kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly since app only works in foreground so this is the proper access.

Also remove a parameter to set a optional access level for keychain cuz this parameter was never used.

### DIFF
--- a/Firebase/InstanceID/FIRInstanceIDAuthKeyChain.h
+++ b/Firebase/InstanceID/FIRInstanceIDAuthKeyChain.h
@@ -73,8 +73,8 @@ NS_ASSUME_NONNULL_BEGIN
                            handler:(nullable void (^)(NSError *error))handler;
 
 /**
- *  Set the data for a given service and account with a specific accessibility. If
- *  accessibility is NULL we use `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` which
+ *  Set the data for a given service and account.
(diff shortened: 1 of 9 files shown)


Firebase/InstanceID/FIRInstanceIDAuthKeyChain.h
```diff
@@ -74,22 +74,19 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  *  Set the data for a given service and account with a specific accessibility. If
- *  accessibility is NULL we use `kSecAttrAccessibleAlwaysThisDeviceOnly` which
+ *  accessibility is NULL we use `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` which
```
maksymmalyhin: Looks like the comment should be updated after removing `accessibility` parameter.

chliangGoogle: good catch!

chliangGoogle:
Yeah, I've tested with an existing keychain store and seems working well with settings updated.

We've updated keychain access in the past and haven't seen any regression issues.

PR #4174, opened 2019-10-25
Make IN queries publicly available

thebrianchen:
Also removed clang-format off/on comments b/c we upgraded clang-format and it no longer mangles NS_SWIFT_NAME annotations with more than one colon.


Firestore/Source/Public/FIRQuery.h
```diff
@@ -273,6 +273,70 @@ addSnapshotListenerWithIncludeMetadataChanges:(BOOL)includeMetadataChanges
...
+ *
+ * A query can have only one arrayContains filter and it cannot be combined with arrayContains or
+ * in.
+ *
+ * @param field The name of the field containing an array to search.
+ * @param value The value that contains the values to match.
+ *
+ * @return The created `FIRQuery`.
+ */
+// clang-format off
```
wilhuff: Does clang actually mangle this? If not please remove.

Firestore/Source/Public/FIRQuery.h
```diff
@@ -273,6 +273,70 @@ addSnapshotListenerWithIncludeMetadataChanges:(BOOL)includeMetadataChanges
...
+- (FIRQuery *)queryWhereField:(NSString *)field arrayContainsAny:(id)value;
```
wilhuff: This is missing an `NS_SWIFT_NAME` annotation so I suspect this API does not look right in swift. Look at how `-queryWhereField:arrayContains:` is annotated to get an idea for how this should look.

Also I can't find an entry in `BasicCompileTests.swift` that verifies this looks right in Swift. All public Objective-C APIs need this because we're relying on the Swift compiler to do this translation for us, but the default isn't always right.

Firestore/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 # Unreleased
+- [feature] Added `whereField(In:)` and `whereField(arrayContainsAny:)` query
```
wilhuff: These signatures are invalid (and unfortunately some of the existing signatures in this changelog are invalid too.

In a declaration like this:

```objc
- (FIRQuery *)queryWhereField:(NSString *)field
                arrayContains:(id)value NS_SWIFT_NAME(whereField(_:arrayContains:));
```

The swift signature in there is `whereField(_:arrayContains:)`. Please use this form here. Also please fix the entry in 0.13.0 that's incorrect.

Also "In" in `whereField(In:)` should not be capitalized (and should also be fixed as described above).

Firestore/Source/Public/FIRQuery.h
```diff
@@ -268,10 +242,69 @@ addSnapshotListenerWithIncludeMetadataChanges:(BOOL)includeMetadataChanges
...
 - (FIRQuery *)queryWhereFieldPath:(FIRFieldPath *)path
                     arrayContains:(id)value NS_SWIFT_NAME(whereField(_:arrayContains:));
-// clang-format on
+
+/**
+ * Creates and returns a new `FIRQuery` with the additional filter that documents must contain
+ * the specified field, the value must be an array, and that array must contain at least one value
+ * from the provided array.
+ *
+ * A query can have only one arrayContains filter and it cannot be combined with arrayContains or
```
wilhuff: "in" within this statement is unquoted and when bare it looks weird. We should probably refer to this as an "in filter". How about putting `arrayContains` and `in` within backticks?

Oh actually, this is an `arrayContainsAny` filter anyway, not `arrayContains`. Please update.

Also this statement is slightly redundant--it says you can only have one arrayContains and that you can't combine it with arrayContains. If this applies to `arrayContainsAny` (which I think is true), how about:

> A query can have only one `arrayContainsAny` or `in` filter.

PR #4178, opened 2019-10-28
Fix documentation comments for the Container.

ryanwilson:


ryanwilson:
Documentation only change, so I'm submitting before travis finishes.

PR #4180, opened 2019-10-28
Initial implementation of GDTCORDatabase

mikehaney24:


GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,312 @@
...
+  if (!db) {
+    GDTCORLogError(GDTCORMCEDatabaseError, @"%@", @"The database is closed");
+    return;
+  }
+  int userVersion = GetUserVersion(db);
+  __block int newUserVersion = userVersion;
+  [migrations enumerateKeysAndObjectsUsingBlock:^(NSNumber *_Nonnull key, NSString *_Nonnull obj,
+                                                  BOOL *_Nonnull stop) {
+    if (key.intValue > newUserVersion) {
+      RunNonQuery(db, nil, obj, stmtCache, NO);
```
dmandar: Do you need to check the return value?

mikehaney24: Yes I do, good catch. Done.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,316 @@
...
+      return nil;
+    }
+    _dbQueue = dispatch_queue_create("com.google.GDTCORDatabase", DISPATCH_QUEUE_SERIAL);
+    _fileURL = dbFileURL;
+    _stmtCache =
+        CFDictionaryCreateMutable(CFAllocatorGetDefault(), 0, &kCFTypeDictionaryKeyCallBacks, NULL);
+    dispatch_async(_dbQueue, ^{
+      RunMigrations(self->_db, [migrationStatements copy], self -> _stmtCache);
+    });
+    dispatch_barrier_sync(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
```
maksymmalyhin: I could not find an explicit statement that `dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0)` returns the same queue all the times for any iOS. Would it be safer to create a local concurrent queue together with creating `instancesToDBFiles` ?

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,316 @@
...
+  return self;
+}
+
+- (void)dealloc {
+  [self close];
+}
+
+- (BOOL)open {
+  if (_db == NULL) {
+    return GDTCORSQLOpenDB(&_db, _fileURL ? _fileURL.path : @":memory:");
```
maksymmalyhin: Should it be run on `_dbQueue ` as well?

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,316 @@
...
+  });
+  return returnStatus;
+}
+
+- (BOOL)runQuery:(NSString *)sql
+        bindings:(NSDictionary<NSNumber *, NSString *> *)bindings
+         eachRow:(GDTCORSqliteRowResultBlock)eachRow
+       cacheStmt:(BOOL)cacheStmt {
+  __block BOOL returnStatus = NO;
+  dispatch_sync(_dbQueue, ^{
```
maksymmalyhin: As far as I understand, the class is intended to be a thread-safe but it will be a client responsibility to make sure the main thread is not blocked. It is correct?

GoogleDataTransport/GDTCORLibrary/Public/GDTCORDatabase.h
```diff
@@ -0,0 +1,84 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+/** This class creates and operates sqlite3 databases. */
+@interface GDTCORDatabase : NSObject
+
+/** The user_version PRAGMA of the db. */
+@property(atomic) int userVersion;
+
+/** The serial queue on which the db operates. */
+@property(nonatomic, readonly) dispatch_queue_t dbQueue;
```
maksymmalyhin: nit: Exposing `dbQueue` creates an impression that the instance methods can be safe to run on it, but current implementation will deadlock if we call `runQuery: ...` from `dbQueue`. Would you like to remove it from the class interface of make a comment?

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,316 @@
...
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    instancesToDBFiles = [[NSMutableDictionary alloc] init];
+  });
+  __block GDTCORDatabase *extantDB;
+  dispatch_barrier_sync(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
+    extantDB = instancesToDBFiles[dbFileURL.path];
+  });
+  if (extantDB) {
+    return nil;
```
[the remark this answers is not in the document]
mikehaney24: No, I'd rather it return nil so that you know there's already a db instance working on that file. I don't want the same db instance being operated from two different places.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,316 @@
...
+                 migrationStatements:
+                     (nonnull NSDictionary<NSNumber *, NSString *> *)migrationStatements {
+  // Only 1 instance per file is allowed. Return nil if one is already open.
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    instancesToDBFiles = [[NSMutableDictionary alloc] init];
+  });
+  __block GDTCORDatabase *extantDB;
+  dispatch_barrier_sync(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
+    extantDB = instancesToDBFiles[dbFileURL.path];
```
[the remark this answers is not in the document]
mikehaney24: I'm not sure what you mean by separate out the cache management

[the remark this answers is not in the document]
mikehaney24: Oh, I get what you're saying now.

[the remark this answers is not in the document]
maksymmalyhin: Sorry if it is not clear described. I mean having 2 methods instead of this one initializer:
1. The initializer which will create an instance each time not paying attention to the cache (`instancesToDBFiles`). The initializer should be private since we don't want it to be used from outside (except in tests maybe)
2. A method like `+ (instancetype)databaseWithURL:` which will returns a database instance from cache or create one.

mikehaney24:
The cache isn't really a cache--it's meant to prevent there from being
multiple GDTCORDatabase instances per-file. If you attempt to open a db
that's already been opened, I don't want it to return the existing
instance, I want it to return nil.

GoogleDataTransport/GDTCORTests/Unit/GDTCORDatabaseTest.m
```diff
@@ -0,0 +1,147 @@
...
+  [super tearDown];
+  dispatch_sync(_db.dbQueue, ^{
+                });
+  [_db close];
+  _db = nil;
+}
+
+/** */
+- (void)testInstantiation {
+  NSString *dbPath = [NSTemporaryDirectory() stringByAppendingPathComponent:@"test.sqlite3"];
```
dmandar: Do you want to randomize the db name for concurrently running tests?

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
mikehaney24: Yeah, though I'm not sure how this class could cause a main thread deadlock.

GoogleDataTransport/GDTCORTests/Unit/GDTCORDatabaseTest.m
mikehaney24: Most tests create in-memory stores, but these tests aren't going to be run concurrently anyway.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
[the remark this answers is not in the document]
mikehaney24: Fixed the cache management, but kept the initializer the same. I don't think the static initializer provides a lot of value in this case.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
mikehaney24: From [the docs](https://developer.apple.com/library/archive/documentation/General/Conceptual/ConcurrencyProgrammingGuide/OperationQueues/OperationQueues.html#//apple_ref/doc/uid/TP40008091-CH102-SW5)

> The system provides each application with four concurrent dispatch queues. These queues are global to the application and are differentiated only by their priority level. Because they are global, you do not create them explicitly. Instead, you ask for one of the queues using the dispatch_get_global_queue function.

That phrasing made me think that this was fine, but then I just found this in some [different docs](https://developer.apple.com/documentation/dispatch/1452797-dispatch_barrier_async)

> The queue you specify should be a concurrent queue that you create yourself using the dispatch_queue_create function. If the queue you pass to this function is a serial queue or one of the global concurrent queues, this function behaves like the dispatch_async function.

That's not actually what I want, so I reworked this.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
mikehaney24: Good catch.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORDatabase.h
mikehaney24: Moved to private header.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
maksymmalyhin: Ah, no I don't mean a deadlock, I mean a long operation on the main thread. Such an operation e.g. at app start may lead to a watchdog killing the app. Just wanted to confirm that it is intended for the class clients to run the methods of the main thread.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
```diff
@@ -0,0 +1,369 @@
...
+ * @param path The path to check.
+ * @return YES, if there's a db in memory operating on that path already.
+ */
+static BOOL InstanceExistsForPath(NSString *path) {
+  // Only 1 instance per file is allowed. Return nil if one is already open (including :memory:).
+  dispatch_once(&cacheToken, ^{
+    dbPathToInstance = [[NSMutableDictionary alloc] init];
+  });
+
+  // Check if this is a special type of path. If it begins and ends with :, it's ok to create.
```
maksymmalyhin: I wonder what is a special type of path?

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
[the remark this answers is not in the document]
maksymmalyhin: Yeah, if it is an intended behaviour not to return a references to an existing instance then I see no issues. Thanks for moving the cache management out to separate functions.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
mikehaney24: From https://www.sqlite.org/c3ref/open.html

_If the filename is ":memory:", then a private, temporary in-memory database is created for the connection. This in-memory database will vanish when the database connection is closed. **Future versions of SQLite might make use of additional special filenames that begin with the ":" character**. It is recommended that when a database filename actually does begin with a ":" character you should prefix the filename with a pathname such as "./" to avoid ambiguity._

We use it mostly for testing, but probably tvOS eventually.

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
mikehaney24: Yeah, if the main-thread blocks waiting for the result of a read, that's the adopter's problem, not this class's (same should be true of other db classes).

GoogleDataTransport/GDTCORLibrary/GDTCORDatabase.m
maksymmalyhin: Nice, thanks!

PR #4184, opened 2019-10-29
C++ migration: implement pretty-printing of Nanopb-generated messages

var-const:
Note: the output is indistinguishable from what is generated by the Objective-C library, with two exceptions:
* class names are different (due to using prefixes, e.g., `GCFS`);
* order in which map fields are outputted can be different (which I presume is to be expected).

I verified this by running integration tests and using simple diffing code to compare outputs: in the destructor of each `Message<T>` (if it's not moved-from), serialize the `Message` to bytes, create an Objective-C proto out of those bytes, convert both protos to their string form, compare the strings (ignoring minor differences in indentation, etc.). Out of 5.7K comparisons, there were just two cases of differences:
* one was in `SerializerTest.FailOnInvalidInputBytes`; since the input in question is invalid, this doesn't seem to require much attention;
* another one was due to differing floating point precision: C++ produces `1.79769e+308` while Objective-C produces `1.79769313486232e+308`. Added a TODO.

Example output:
```
2019-10-28 21:44:56.941749-0400 Firestore_Example_iOS[78883:2094242] 6.11.0 - [Firebase/Firestore][I-FST000001] WriteStream (7f903ca53d50) write request: <WriteRequest 0x70000991ef08>: {
    writes {
      update {
        name: "projects/varconst-11-a51fd/databases/(default)/documents/test-collection/u6QizHeYjqt8Xe61kBoi"
        fields {
          key: "key"
          value {
            string_value: "value"
          }
        }
      }
    }
    stream_token: "\031\020hB\002\201\364\265\265"
}
```


Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -188,16 +189,96 @@ def nanopb_parse_files(request, options):
...
 
-  return parsed_files
+    base_filename = fdesc.name.replace('.proto', '')
+    pretty_printing_info[base_filename] = FilePrettyPrintingInfo(parsed_file,
+                                                                 base_filename)
+
+  return parsed_files, pretty_printing_info
+
+
+class FilePrettyPrintingInfo:
```
var-const: I don't particularly like the names of these classes, it's more that I used them as sort of placeholders and then got used to them. Any renaming suggestions are very welcome.

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +329,309 @@ def nanopb_write(results):
...
+
+    for full_name in enum.members:
+      short_name = full_name.replace('%s_' % enum.name, '')
+      f.content += '''\
+    case %s:
+        return "%s";\n''' % (full_name, short_name)
+
+    f.content += '''
+    }
+    return "<unknown enum value>";
```
var-const: What do you think?

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +329,309 @@ def nanopb_write(results):
...
+    cc_name = parent_oneof.name + '.' + cc_name
+
+  if cc_name == 'delete':
+    cc_name = 'delete_'
+  if field.is_primitive:
+    display_name += ':'
+
+  function_name = ''
+  if field.is_enum:
+    function_name = 'PrintEnumField'
```
var-const: I originally used template metaprogramming for this, but tweaking generated code is probably simpler.

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +329,309 @@ def nanopb_write(results):
...
+  args = (
+    indent(indent_level), function_name, display_name, maybe_linebreak, cc_name,
+    always_print)
+
+  result = format_str % args
+  if len(result) <= line_width:
+    return result
+
+  # Best-effort attempt to fit within the expected line width.
+  maybe_linebreak = '\n' + indent(indent_level + 1)
```
var-const: Let me know if there's a nicer way to achieve this, without reinitializing `args`.

Firestore/Protos/nanopb/nanopb_pretty_printers.h
```diff
@@ -0,0 +1,108 @@
...
+  if (value == T{} && !always_print) {
+    return "";
+  }
+
+  return absl::StrCat(internal::Indent(indent), name, EnumToString(value), "\n");
+}
+
+template <typename T>
+std::string PrintHeader(bool is_root, absl::string_view message_name, const T* message_ptr) {
+  if (is_root) {
```
var-const: I suppose this logic could be moved to `Message::ToString`. Not sure if it would be any better, though.

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +329,308 @@ def nanopb_write(results):
   response = plugin_pb2.CodeGeneratorResponse()
 
   for result in results:
-    f = response.file.add()
-    f.name = result['headername']
-    f.content = result['headerdata']
+    base_filename = result['headername'].replace('.nanopb.h', '')
+    file_printers = pretty_printing_info[base_filename]
```
wilhuff: Isn't the thing returned here the `FilePrettyPrintingInfo` instance for the file? Shouldn't this be `file_printer`?

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +329,308 @@ def nanopb_write(results):
...
+  This is for changes to the code, as well as additions that cannot be made via
+  insertion points. Current fixups:
+  - rename fields named `delete` to `delete_`, because it's a keyword in C++.
+  """
+
+  delete_keyword = re.compile(r'\bdelete\b')
+  return delete_keyword.sub('delete_', file_contents)
+
+
+def create_insertion(files, file_name, insertion_point):
```
wilhuff: This file-list, file name pair, along with behavior related to it seems like it could be a class.

Here's an example of what I mean:

```python
class CurrentFile(object):

  def __init__(self, files, file_name) ...

  def create_insertion(self, insertion_point):
    f = self.files.add()
    f.name = self.file_name
    f.insertion_point = insertion_point
    return f
```

I think you likely could also move `add_contents` to be a member (though I think you should name it `set_contents`, because there can only be one main, right?

Also, all of the callers of this API immediately just call `f.content =`. Consider making this method `insert`, i.e.

```python
  def insert(self, insertion_point, content):
    f = self.files.add()
    f.name = self.file_name
    f.insertion_point = insertion_point
    f.content = content
```

Firestore/Protos/nanopb_pretty_printing/generators.py
```diff
@@ -0,0 +1,386 @@
...
+
+  def __init__(self, field_desc, message_desc):
+    """Args:
+      field_desc: nanopb.Field describing this field.
+      message_desc: nanopb.Message describing the message containing this field.
+    """
+    self.name = field_desc.name
+    self.tag = field_desc.tag
+
+    self.is_optional = field_desc.rules == 'OPTIONAL' and field_desc.allocation == 'STATIC'
```
var-const: For some reason, PyCharm doesn't want to break this line.

var-const:
@wilhuff Added unit tests for your reviewing pleasure.

Firestore/core/src/firebase/firestore/remote/serializer.h
```diff
@@ -61,11 +61,6 @@ class LocalSerializer;
 
 namespace remote {
 
-template <typename T>
```
var-const: I moved it to `nanopb/`, partially to be able to use it in tests without depending on `remote/`, but it also seems like a more appropriate place.

Firestore/core/src/firebase/firestore/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,11 @@ inline bool SafeReadBoolean(const bool& nanopb_boolean) {
   return absl::bit_cast<int8_t>(nanopb_boolean) != 0;
 }
 
+template <typename T>
+T* _Nonnull MakeArray(pb_size_t count) {
```
var-const: XCode demands nullability annotations in this header; any idea why this is happening? I can't see anything that would make it treat the file as Objective-C++.

Firestore/Protos/nanopb_pretty_printing/generators.py
```diff
@@ -0,0 +1,415 @@
+#!/usr/bin/env python
```
wilhuff: How about `lib/pretty_printing.py` as the filename for this?

I can imagine we'll have other things we can put in `lib`, but it's doubtful we'll put anything in `nanopb_pretty_printing`.

Note that all we're trying to do is disambiguate top-level scripts from the parts they use. I doubt we'd have any folder structure underneath `lib` unless the amount of code generation goes through the roof.

Firestore/Protos/nanopb_cpp_generator.py
```diff
@@ -246,16 +271,149 @@ def nanopb_write(results):
...
+    file_contents: The contents of the generated file as a single string. The
+      fixups will be applied without distinguishing between the code and the
+      comments.
+  """
+
+  delete_keyword = re.compile(r'\bdelete\b')
+  return delete_keyword.sub('delete_', file_contents)
+
+
+def nanopb_augment_header(request, pretty_printing_generator):
```
wilhuff: The `pretty_printing_generator` name suggests that it might work with any generator, when it really is the file one.

Consider renaming to `file_generator` to make the parameter name more clearly indicate the restriction.

Firestore/Protos/nanopb_pretty_printing/generators.py
```diff
@@ -0,0 +1,415 @@
...
+      result += _reindent(0, '''
+              std::string tail = PrintTail(indent);
+              return header + result + tail;
+          }\n\n''')
+
+    return result
+
+
+class FieldPrettyPrintingGenerator:
+  """Generates pretty-printing support for a oneof field.
```
wilhuff: for a *non-oneof* field, right?

Firestore/Protos/nanopb_pretty_printing/generators.py
```diff
@@ -0,0 +1,415 @@
...
+
+    self.name = field_desc.name
+    self.tag = field_desc.tag
+
+    self.is_optional = field_desc.rules == 'OPTIONAL' and field_desc.allocation == 'STATIC'
+    self.is_repeated = field_desc.rules == 'REPEATED'
+    self.is_primitive = field_desc.pbtype != 'MESSAGE'
+    self.is_enum = field_desc.pbtype in ['ENUM', 'UENUM']
+
+  def __str__(self):
```
wilhuff: I'm not opposed to this but it doesn't seem to fit exactly with the `generate_declaration`/`generate_definition` split above. It would be clearer which phase this is a part of if this were named similarly. This would also have the effect of making it clearer when we're recursing within the generator.

Firestore/Protos/nanopb_pretty_printing/generators.py
```diff
@@ -0,0 +1,415 @@
...
+    # runs separately from generating pretty printers; consequently, when pretty
+    # printers are being generated, all proto fields still have their unmodified
+    # names.
+    if cc_name == 'delete':
+      cc_name = 'delete_'
+
+    if self.is_repeated:
+      cc_name += '[i]'
+
+    if parent_oneof and not parent_oneof._is_anonymous:
```
wilhuff: You've accessing `_is_anonymous` from outside the class that defines it. This means that it shouldn't be a "private" member.

Firestore/core/test/firebase/firestore/nanopb/message_test.cc
```diff
@@ -1,7 +1,6 @@
 /*
  * Copyright 2019 Google
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
```
wilhuff: Restore this line?

Firestore/core/test/firebase/firestore/nanopb/pretty_printing_test.cc
```diff
@@ -0,0 +1,259 @@
...
+namespace {
+
+using ::testing::MatchesRegex;
+
+TEST(PrettyPrintingTest, PrintsInt) {
+  Message<firestore_client_WriteBatch> m;
+  m->batch_id = 123;
+
+  EXPECT_THAT(m.ToString(), MatchesRegex(
+                                R"(<WriteBatch 0x[0-9A-Fa-f]+>: {
```
wilhuff: Is the parenthesis immediately inside the quote required?

Firestore/core/test/firebase/firestore/nanopb/pretty_printing_test.cc
```diff
@@ -0,0 +1,259 @@
...
+  batch_id: 123
+})"));
+}
+
+TEST(PrettyPrintingTest, PrintsBool) {
+  Message<firestore_client_MaybeDocument> m;
+  m->has_committed_mutations = true;
+
+  EXPECT_THAT(m.ToString(), MatchesRegex(
```
wilhuff: You could slightly cut the verbosity down by writing a matcher.

Something like this:
```c++
MATCHER_P(HasToString, expected, "") {
  std::string arg_string = arg.ToString();
  return testing::Value(arg_string, MatchesRegex(expected));
}
```

This would allow you to write this as:

```c++
EXPECT_THAT(m, HasToString(R"(<MaybeDocument ..."));
```

It's not a big win, but would reduce some of the incantation burden.

Firestore/Protos/nanopb_cpp_generator.py
var-const: Thanks for the suggestions! While both `ToString` and `Pprint` seem reasonable, I still slightly prefer `PrettyPrinting` to either of them. However, dropping `Generator` seems like a win, and goes a long way towards solving the verbosity problem.

Firestore/Protos/nanopb_cpp_generator.py
var-const: Renamed to `file_pretty_printing`.

Firestore/core/src/firebase/firestore/nanopb/pretty_printing.h
```diff
@@ -0,0 +1,125 @@
...
+// If the field has its default value (e.g., zero), then:
+// - if `always_print` is false, an empty string is returned.
+// - if `always_print` is true, a string representing the default value is
+//   returned.
+template <typename T>
+std::string PrintPrimitiveField(absl::string_view name,
+                                T value,
+                                int indent_level,
+                                bool always_print) {
+  if (value == T{} && !always_print) {
```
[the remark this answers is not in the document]
var-const: Yes. My understanding is that in proto3, any primitive that has its default value is considered unset, and unset fields are not printed.

Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
```diff
@@ -74,6 +76,57 @@ PB_STATIC_ASSERT((pb_membersize(firestore_client_NoDocument, read_time) < 256 &&
...
+    case firestore_client_MaybeDocument_unknown_document_tag:
+        result += PrintMessageField("unknown_document ",
+            unknown_document, indent + 1, true);
+        break;
+    }
+    result += PrintPrimitiveField("has_committed_mutations: ",
+        has_committed_mutations, indent + 1, false);
+
+    bool is_root = indent == 0;
+    if (!result.empty() || is_root) {
```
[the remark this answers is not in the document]
var-const: This is admittedly not the clearest way (because we generate `has_` fields very sparingly), but it should achieve the right outcome (modulo any bugs).

By default, all submessages are unset. The only way we have of setting a submessage without modifying any of its members is by setting the `has_` flag to `true`. In the current implementation, an unset submessage and an empty-but-set submessage are printed differently (the former is not printed, and the latter is printed as empty braces). If we were to always print submessages, we would lose that distinction (the submessages would always be considered set). This would be inconsistent with the (presumably conformant) output that Objective-C library produces, and would also mean that the generated text representation would not roundtrip if it were given as input to a text proto parsing function.

> This means that in our logs a submessage might be omitted even though we're actually sending it.

Unless there's a bug, this shouldn't happen in the current implementation. For an "optional" field (the one that has a dedicated `has_` flag), the generated code will always print the value if it's set, even if the value is  the default one (via the `always_print` parameter).

> If the submessage is not optional, always print its contents, even if all the fields have default values.

This is the case in the current implementation, though the condition is in `PrintMessageField`, not `ToString`.

I've added a unit test to show an empty optional is printed, but the larger point is that for our integration test suite, the generated output is _exactly_ the same as what the Objective-C library would generate (except for 2-vs-4 space indentation of the root message and the "headers" (e.g., `<Write 0x123>: {` vs `<GCFSWrite 0x456>: {`). It is possible that there are cases not triggered by out integration tests, but it's very unlikely that the current implementation is simply broken.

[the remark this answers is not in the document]
wilhuff: The difference I'm trying to highlight is that fidelity to the Objective-C output is a non-goal. Fidelity to what we're actually sending is *the* feature we should be striving for here.

Objective-C protobuf treats all submessages as optional. nanopb treats them as optional only if we specially mark them up (which we've done only minimally). Because we're so conservative with making submessages optional (and minimally having `has_` accessors) we're potentially sending more submessages without meaningful content to the server. When the backend team files a bug against us I want the logs to reflect what we actually sent rather than what we might have sent if we were still using Objective-C protobuf.

Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
```diff
@@ -74,6 +76,57 @@ PB_STATIC_ASSERT((pb_membersize(firestore_client_NoDocument, read_time) < 256 &&
 #endif
 
 
+std::string firestore_client_NoDocument::ToString(int indent) const {
+    std::string header = PrintHeader(indent, "NoDocument", this);
```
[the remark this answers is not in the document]
wilhuff: If always set messages (i.e. those without `has_` accessors in the parent) are always printed (as I suggest elsewhere) then we never need to detect that there was no output in `result` and this distinction can be eliminated.

Firestore/Protos/nanopb/google/protobuf/timestamp.nanopb.cc
```diff
@@ -36,6 +38,22 @@ const pb_field_t google_protobuf_Timestamp_fields[3] = {
...
 
+std::string google_protobuf_Timestamp::ToString(int indent) const {
+    std::string header = PrintHeader(indent, "Timestamp", this);
+    std::string result;
+
+    result += PrintPrimitiveField("seconds: ", seconds, indent + 1, false);
+    result += PrintPrimitiveField("nanos: ", nanos, indent + 1, false);
+
+    bool is_root = indent == 0;
+    if (!result.empty() || is_root) {
```
wilhuff: To concretely follow up on the points being discussed above: in the context of timestamps this is never what we want. The backend frequently distinguishes Timestamps that are set but have zero values from Timestamps that are unset. Having this logic here defeats our ability to inspect via the logs what we actually sent.

Firestore/Protos/nanopb_pretty_printing/generators.py
wilhuff: Fair enough. Let's leave it as-is.

Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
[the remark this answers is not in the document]
wilhuff: Discussed offline: the last round of comments does not apply because I misunderstood the nature of nanopb's default proto3 support. By default nanopb elides submessages whose values are all default-values. Flagging these as proto2 and giving them explicit `has_` accessors allows us to consider a message set even though it might otherwise have been elided from the output.

This means that this code that emulates that behavior is correct.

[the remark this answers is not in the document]
var-const: Thanks for writing this up. I'll also copy the results of my experiments below, in case they come in handy later.

Imagine there's an interactive prompt that outputs the serialized form of the message after each operation, using `ByteString::ToString()` and also surrounding the output with quotes.

First, here's the current form of `StructuredQuery`, which treats `start_at` field (of type `Cursor`) as a proto3 field (that is, it doesn't have a `has_` flag for the field):
```
> Message<google_firestore_v1_StructuredQuery> m;
""
> m->start_at.before = true;
":\002\020\001"
```

Now, if I set options for that field to consider it a proto2 field, it would have a `has_start_at` flag generated for it, and the output would look like this:
```
> Message<google_firestore_v1_StructuredQuery> m;
""
> m->has_start_at = true;
":\000"
> m->start_at.before = true;
":\002\020\001"
> m->has_start_at = false;
""
```

PR #4185, opened 2019-10-29
Various minor fixes to make SDK compile on Linux

wu-hui:


Firestore/core/test/firebase/firestore/core/query_listener_test.cc
```diff
@@ -142,13 +142,15 @@ TEST_F(QueryListenerTest, RaisesCollectionEvents) {
...
+  std::vector<Status> accum;
+  auto* accum_ptr = &accum;
   Query query = testutil::Query("rooms/Eros");
 
   auto listener = QueryListener::Create(
-      query, ^(const StatusOr<ViewSnapshot>& maybe_snapshot) {
-        accum.push_back(maybe_snapshot.status());
-      });
+      query, EventListener<ViewSnapshot>::Create(
+                 [&accum_ptr](const StatusOr<ViewSnapshot>& maybe_snapshot) {
```
wilhuff: You could just pass `&accum` here. There's no reason to pass the pointer. (If you really needed to use a pointer for whatever reason, you could pass it by value to, rather than passing the pointer by reference as you've done here).

wu-hui: Fixed.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -0,0 +1,49 @@
...
+namespace firebase {
+namespace firestore {
+namespace local {
+
+using util::Path;
+
+Path LevelDbPersistence::AppDataDirectory() {
+  const char* home_dir;
+
+  if ((home_dir = getenv("HOME")) == NULL) {
```
wilhuff: If we're running on Linux (but not Android), we really should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for user specific data files. That is:

  * The check should be inside `#if __linux__ && !__ANDROID__`
  * getenv("XDG_DATA_HOME"); if set use it for `data_home`.
  * if unset find the home directory and then path append `.local/share` (this is what argues for `data_home` being a `Path`).

Only on non-Linux POSIX should we fall back on what you have here (though we have no supported platforms of this type there's nothing stopping someone from trying this on e.g. Solaris).

I suggest factoring out `HomeDirectory` that does the `$HOME` then fallback on `getpwuid` so you can reuse that in both Linux and non-Linux cases.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/firebase/firestore/local/leveldb_persistence.h"
+
+#include "Firestore/core/src/firebase/firestore/util/path.h"
+#include "Firestore/core/src/firebase/firestore/util/status.h"
+
+#if !defined(__APPLE__) && !defined(_WIN32)
```
var-const: @wilhuff Should this be determined at the build level rather than using preprocessor definitions? Currently, both implementations are included in the `cc_library` (unless they're somehow filtered out later).

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return Path::FromUtf8(pwd.pw_dir);
+  }
+
+  return Path::FromUtf8(getenv("HOME"));
+}
+
+StatusOr<Path> GetDataHomeDirectory() {
+  if (getenv("XDG_DATA_HOME") == nullptr) {
+    const Path& home_dir = GetHomeDirectory().ValueOrDie();
```
wilhuff: You can't call `ValueOrDie()` without first checking if the `StatusOr` is OK. This should be:

```c++
const char* data_home = getenv("XDG_DATA_HOME");
if (data_home) return Path::FromUtf8(data_home);

StatusOr<Path> maybe_home_dir = GetHomeDirectory();
if (!maybe_home_dir.ok()) return maybe_home_dir;

const Path& home_dir = maybe_home_dir.ValueOrDie();
return home_dir.AppendUtf8(".local/share");
```

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -0,0 +1,86 @@
...
+    return home_dir.AppendUtf8(".local/share");
+  }
+
+  return Path::FromUtf8(getenv("XDG_DATA_HOME"));
+}
+
+}  // namespace
+
+Path LevelDbPersistence::AppDataDirectory() {
```
wilhuff: This function should also return `StatusOr<Path>`.

In general you never want to call `ValueOrDie` unguarded. Even if you want to crash we don't want to crash that way, because it just aborts. We want to at least indicate why we're crashing.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -0,0 +1,86 @@
...
+  std::string dot_prefixed = absl::StrCat(".", kReservedPathComponent);
+
+#if defined(__linux__) && !defined(__ANDROID__)
+  const Path& dir = GetDataHomeDirectory().ValueOrDie();
+  return dir.AppendUtf8(dot_prefixed);
```
wilhuff: In the XDG case, this should append `kReservedPathComponent` and not the dot prefixed form. (The leading .local already makes this a dotfile).

Firestore/core/test/firebase/firestore/local/leveldb_persistence_test.cc
```diff
@@ -0,0 +1,39 @@
...
+namespace firestore {
+namespace local {
+
+using util::RecursivelyCreateDir;
+using util::RecursivelyDelete;
+
+TEST(LevelDbPersistenceTest, CanFindAppDataDirectory) {
+  const auto& path = LevelDbPersistence::AppDataDirectory();
+  EXPECT_TRUE(RecursivelyCreateDir(path).ok());
+  EXPECT_TRUE(RecursivelyDelete(path).ok());
```
wilhuff: You can't do this on a real host. Running this test would delete all the cached data for my firestore apps!

Firestore/core/src/firebase/firestore/local/leveldb_persistence_apple.mm
```diff
@@ -46,7 +47,8 @@
   return Path::FromNSString(NSHomeDirectory()).AppendUtf8(dot_prefixed);
 
 #else
-#error "Don't know where to store documents on this platform."
+  return Status{Error::Internal,
```
wilhuff: I don't think this change is right.

All known, supported configurations have been enumerated. Any other Apple platform should fail to compile because we haven't considered how to store data on it.

As it stands this change allows the build to succeed on an unknown platform, but then the default configuration (of persistence enabled) will always fail, which is not useful.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -20,8 +20,9 @@
 #include "Firestore/core/src/firebase/firestore/util/path.h"
 #include "Firestore/core/src/firebase/firestore/util/status.h"
 
-#if !defined(__APPLE__) && !defined(_WIN32)
+#if !__APPLE__ && !_WIN32
 
+#include <asm-generic/errno-base.h>
```
wilhuff: This is the wrong header. This should be the C++ header `<cerrno>`.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
```diff
@@ -20,8 +20,9 @@
...
 #include <pwd.h>
 #include <sys/types.h>
 #include <unistd.h>
```
wilhuff: C system headers should be separated from C++ standard library headers with a newline.

Firestore/core/src/firebase/firestore/local/leveldb_persistence.cc
```diff
@@ -123,7 +123,8 @@ LevelDbPersistence::LevelDbPersistence(std::unique_ptr<leveldb::DB> db,
 // MARK: - Storage location
 
 util::Path LevelDbPersistence::StorageDirectory(
-    const core::DatabaseInfo& database_info, const util::Path& documents_dir) {
+    const core::DatabaseInfo& database_info,
+    const util::StatusOr<util::Path>& documents_dir) {
```
wilhuff: The caller should handle the status not being OK and only call StorageDirectory with a valid Path. That way we can centralize how we fail in response to errors of any kind preventing the creation of the persistence instance.

wu-hui: Reverted.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_posix.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/local/leveldb_persistence_apple.mm
wu-hui: Yes, corrected.

Firestore/core/test/firebase/firestore/local/leveldb_persistence_test.cc
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/firebase/firestore/local/leveldb_persistence.h"
+#include <iostream>
```
wilhuff: Newline before iostream. None before gtest.

wu-hui: Done.

PR #4196, opened 2019-10-30
Send diagnostic data to the FLL backend instead of the CCT backend

mikehaney24:


mikehaney24:
Added redirect response handling along with setting a User-Agent string for traffic differentiation on the server.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
```diff
@@ -328,4 +353,23 @@ - (void)appWillTerminate:(GDTCORApplication *)application {
...
 }
 
+#pragma mark - NSURLSessionDelegate
+
+- (void)URLSession:(NSURLSession *)session
+                          task:(NSURLSessionTask *)task
+    willPerformHTTPRedirection:(NSHTTPURLResponse *)response
+                    newRequest:(NSURLRequest *)request
+             completionHandler:(void (^)(NSURLRequest *_Nullable))completionHandler {
+  if (!completionHandler) {
```
maksymmalyhin: nit: The parameter is not marked as an optional, not sure if we should validate it anyways, though it doesn't harm.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
```diff
@@ -328,4 +353,23 @@ - (void)appWillTerminate:(GDTCORApplication *)application {
...
+    return;
+  }
+  if (response.statusCode == 302 || response.statusCode == 301) {
+    NSURLRequest *newRequest = [self constructRequestWithURL:request.URL
+                                                        data:task.originalRequest.HTTPBody];
```
maksymmalyhin: while testing redirects yesterday I've seen different cases depending on the code. When the code was 307 there might be no `HTTPBody` in either `task.originalRequest`, `task.currentRequest` or `request`, but there was `request.HTTPBodyStream`. When the code was 302 I don't remember either `HTTPBody` or `HTTPBodyStream` for any of the requests instances. Could you double check this code is actually working?

maksymmalyhin: I see we keep track of `self.currentTask` meaning we know for sure which task is being processed. Can we store the body as well so we can safely reconstruct the request?

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
mikehaney24: Only newer sanitizer versions catch this and help you solve it. On older versions of Xcode, it'd just crash without much help. Since invoking a nil block is a hard crash, I'd rather keep the check in.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
mikehaney24: The way I instantiate the request (hopefully) keeps the HTTPBody property populated. I can't finish testing until the server begins returning 302s instead of 307s. I'd rather keep the body associated with the request to keep correct context if more than one task ever ends up being executed.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
maksymmalyhin: Why the block would be nil?

mikehaney24: Why wouldn't it be? Until recently, the annotations were more suggestions than hard rules. I'm not in control of what calls or passes data into this method, therefore, it's prudent to validate input.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
maksymmalyhin: Ah, I see, you didn't set the request HTTPBody filed itself. I didn't  notice it before. Can we create task with [this method](https://developer.apple.com/documentation/foundation/nsurlsession/1407613-datataskwithrequest?language=objc) then to remove ambiguity?

mikehaney24:
> LGTM, though I would prefer it to be fully tested before merging. Maybe a dummy HTTP server can help with it?

Doh! I forgot--I have a test HTTP server just for this purpose! I'll add a unit test, that should help increase confidence.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
```diff
@@ -309,6 +314,26 @@ - (nonnull NSData *)constructRequestProtoFromPackage:(GDTCORUploadPackage *)pack
...
+  NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:URL];
+  [request setValue:[self defaultAPIKey] forHTTPHeaderField:@"X-Goog-Api-Key"];
+  [request setValue:@"application/x-protobuf" forHTTPHeaderField:@"Content-Type"];
+  [request setValue:@"gzip" forHTTPHeaderField:@"Content-Encoding"];
+  [request setValue:@"gzip" forHTTPHeaderField:@"Accept-Encoding"];
+  NSString *userAgent = [NSString stringWithFormat:@"datatransport/%@ fllsupport/%@ apple/",
+                                                   kGDTCORVersion, kGDTCCTSupportSDKVersion];
+  [request setValue:userAgent forHTTPHeaderField:@"User-Agent"];
+  request.HTTPMethod = @"POST";
+  [request setHTTPBody:data];
```
maksymmalyhin: BTW, I suspect it may solve the issue even for 302 and the delegate method will be unnecessary. Do you think we can verify it?

mikehaney24: I just finished writing a unit test with a real HTTP server. With the delegate method implemented, it passes all redirects. Without the delegate method implemented, it fails on 301 and 302 redirects and passes on 307.

image

PR #4199, opened 2019-10-30
check if instanceID instance creates race condition

chliangGoogle:


chliangGoogle:
Looks like the test is still failing :(

ryanwilson:
I cherry-picked this commit to #4200 so this can be closed, both are necessary (plus other work I think).

Issue #4198, opened 2019-10-30
Fix IID/Messaging AutoInitEnabled test


ryanwilson:
Disabled temporarily until the tests can be fixed - tests pass locally but fail on pod lib lint.

cc @chliangGoogle

ryanwilson:
This wasn't actually disabled (the PR never landed), closing.

PR #4200, opened 2019-10-30
Temporarily disable testFCMAutoInitEnabled

ryanwilson:
This should be disabled to make Travis pass.

Tracked in #4198


ryanwilson:
This is out of date and passing on master, closing.

PR #4201, opened 2019-10-30
Add kokoro CI skeleton to firebase-ios-sdk

morganchen12:
Adds the barebones kokoro folder structure and a script that doesn't quite work (breaks because a travis env var is missing), along with instructions on how to add new build jobs.

My intention is to ultimately delete the `kokoro_build.sh` script once teams write their own tests; the script currently exists just to verify kokoro webhooks are working.


kokoro/macos_external/kokoro_build.sh
```diff
@@ -0,0 +1,20 @@
...
+set -x
+
+export LC_CTYPE=en_US.UTF-8
+export TRAVIS_PULL_REQUEST=""
+
+# Code under repo is checked out to ${KOKORO_ARTIFACTS_DIR}/github.
+# The final directory name in this path is determined by the scm name specified
+# in the job configuration.
+cd ${KOKORO_ARTIFACTS_DIR}/github/firebase-ios-sdk
+bundle install
```
wilhuff: This ignores essentially everything we've learned while getting Travis working. I don't think we should rewrite all the build scripts.

Builds in travis have two parts: [`install_prereqs.sh`](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/install_prereqs.sh) and [`build.sh`](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh). There's considerable shared code between projects that I don't think we should be repeating per product.

kokoro/macos_external/kokoro_build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
wilhuff: Is this the build for the Firebase pod? We should name this file after which pods it builds. There are going to be lots of them.

morganchen12:
Disregard travis failures, acquire kokoro

kokoro/macos_external/kokoro_build.sh
wilhuff: Would it be possible to move this script and associated cfg into the location that would actually be expected for a Core build of some kind? As it stands I can't really visualize how your instructions would actually work for any given build.

For example:
  * The readme mentions a `pull_request` directory but this doesn't use it
  * There's no equivalent directory requirement for continuous builds; is that intended?
  * Is the shell script name supposed to match the .cfg name? (If so kokoro_build.sh matches nothing)
  * Or maybe we're supposed to pass product/platform/method via build_params?

I'm having trouble seeing how this fits together.

kokoro/README.md
```diff
@@ -0,0 +1,22 @@
...
+Firebase uses a Google-internal continuous integration tool called "Kokoro"
+(a.k.a "internal CI") for running majority of its open source tests. This
+directory contains the external part of kokoro test job configurations (the
+actual job definitions live in Google's internal monorepo) and the shell
+scripts that act as entry points to execute the actual tests.
+
+## Kokoro job structure
+
+Each job name should follow the format `product_target_method_xcodeversion.cfg`
+(i.e. `Core_macOS_xcodebuild_Xcode10_1.cfg`). The Xcode version may be omitted
```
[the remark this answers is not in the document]
morganchen12: This was something I came up with by squishing the travis build preview parameters together into a snake case identifier, so I'm ok with changing it.

Omitting the default Xcode version seems like a good way to reduce visual clutter. Kokoro's docs mention that each xcode should be explicitly selected to prevent builds breaking when kokoro updates their default version and we'd have to rename all default Xcode builds every time the default version of Xcode changed, but this is not a huge amount of work and could probably be automated.

morganchen12:
I've added all of the outstanding discussion topics to the iOS Weekly meeting discussion on Tuesday (like job naming conventions).

Going to merge this as-is and address follow-ups with other iOS contributors next week.

PR #4202, opened 2019-10-30
Address issues found by repeated tsan runs

mikehaney24:


GoogleDataTransport/GDTCORTests/Common/Categories/GDTCORRegistrar+Testing.m
```diff
@@ -21,10 +21,10 @@
 @implementation GDTCORRegistrar (Testing)
 
 - (void)reset {
+  [self.targetToPrioritizer removeAllObjects];
+  [self.targetToUploader removeAllObjects];
   dispatch_sync(self.registrarQueue, ^{
-    [self.targetToPrioritizer removeAllObjects];
-    [self.targetToUploader removeAllObjects];
-  });
+                });
```
morganchen12: Now that this block is empty, can the whole dispatch statement be removed?

maksymmalyhin: I think it is needed to make sure all background tasks are finished before the method exits.

PR #4208, opened 2019-10-31
Port FSTMutationQueueTests to C++

wilhuff:
(note the diffbase)


Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+using testutil::Map;
+using testutil::Query;
+
+MutationQueueTestBase::MutationQueueTestBase(
+    std::unique_ptr<Persistence> persistence)
+    : persistence_(std::move(persistence)),
+      mutation_queue_(persistence_->GetMutationQueueForUser(User("user"))) {
+}
+
+MutationQueueTestBase::~MutationQueueTestBase() = default;
```
var-const: Is there any advantage to not define it in the header?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * A test fixture for implementing tests of the MutationQueue interface.
+ *
+ * This is separate from MutationQueueTest below in order to make additional
+ * implementation-specific tests.
+ */
+class MutationQueueTestBase : public testing::Test {
+ public:
+  explicit MutationQueueTestBase(std::unique_ptr<Persistence> persistence);
+  ~MutationQueueTestBase();
```
var-const: Question: should this be `virtual`?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    AddMutationBatch();
+
+    ASSERT_EQ(mutation_queue_->GetLastStreamToken(), stream_token1);
+
+    mutation_queue_->AcknowledgeBatch(batch1, stream_token2);
+    ASSERT_EQ(mutation_queue_->GetLastStreamToken(), stream_token2);
+  });
+}
+
+#pragma mark - Helpers
```
var-const: Seems like a leftover.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
```diff
@@ -0,0 +1,84 @@
...
+
+ protected:
+  model::MutationBatch AddMutationBatch(const std::string& key = "foo/bar");
```
var-const: Use `string_view` as the parameter?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  for (int i = 0; i < number; i++) {
+    MutationBatch batch = AddMutationBatch();
+    batches.push_back(batch);
+  }
+
+  return batches;
+}
+
+/** Returns the number of mutation batches in the mutation queue. */
+size_t MutationQueueTestBase::BatchCount() {
```
var-const: Optional: `GetBatchCount`? I personally only name getters without "get" (but the fact that they're getters is clear from the snake_case), though feel free to disagree, it's a completely optional suggestion.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+/**
+ * Creates an array of batches containing @a number dummy MutationBatches. Each
+ * has a different batch_id.
+ */
+std::vector<MutationBatch> MutationQueueTestBase::CreateBatches(int number) {
+  std::vector<MutationBatch> batches;
+
+  for (int i = 0; i < number; i++) {
+    MutationBatch batch = AddMutationBatch();
+    batches.push_back(batch);
```
var-const: Optional: `batches.push_back(AddMutationBatch());`? The intermediate variable doesn't seem to add much value.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+ * Removes the first n entries from the the given batches and returns them.
+ *
+ * @param n The number of batches to remove.
+ * @param batches The array to mutate, removing entries from it.
+ * @return A new array containing all the entries that were removed from @a
+ * batches.
+ */
+std::vector<MutationBatch> MutationQueueTestBase::RemoveFirstBatches(
+    size_t n, std::vector<MutationBatch>* batches) {
+  std::vector<MutationBatch> removed(batches->begin(), batches->begin() + n);
```
var-const: Optional: assert that `size() >= n`?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+
+MutationQueueTestBase::MutationQueueTestBase(
+    std::unique_ptr<Persistence> persistence)
+    : persistence_(std::move(persistence)),
+      mutation_queue_(persistence_->GetMutationQueueForUser(User("user"))) {
+}
+
+MutationQueueTestBase::~MutationQueueTestBase() = default;
+
+/**
```
var-const: Here and below -- should these comments be in the header?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
```diff
@@ -0,0 +1,84 @@
...
+  std::vector<model::MutationBatch> CreateBatches(int number);
+  size_t BatchCount();
+  std::vector<model::MutationBatch> RemoveFirstBatches(
+      size_t n, std::vector<model::MutationBatch>* batches);
+
+  std::unique_ptr<Persistence> persistence_;
+  MutationQueue* mutation_queue_;
```
var-const: Optional: `= nullptr`?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+ * Creates a new MutationBatch with the given key, the next batch ID and a set
+ * of dummy mutations.
+ */
+MutationBatch MutationQueueTestBase::AddMutationBatch(const std::string& key) {
+  SetMutation mutation = testutil::SetMutation(key, Map("a", 1));
+
+  MutationBatch batch =
```
var-const: Optional: omit the intermediate variable and return the return value of `AddMutationBatch` directly?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+      mutation_queue_->AddMutationBatch(Timestamp::Now(), {}, {mutation});
+  return batch;
+}
+
+/**
+ * Creates an array of batches containing @a number dummy MutationBatches. Each
```
var-const: Nit: use backticks instead of `@a`?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+/** Returns the number of mutation batches in the mutation queue. */
+size_t MutationQueueTestBase::BatchCount() {
+  return mutation_queue_->AllMutationBatches().size();
+}
+
+/**
+ * Removes the first n entries from the the given batches and returns them.
+ *
+ * @param n The number of batches to remove.
+ * @param batches The array to mutate, removing entries from it.
```
var-const: Nit: s/array/container/ (or similar)?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+ * @return A new array containing all the entries that were removed from @a
```
var-const: Nit: s/array/vector.

Nit: s/@a/backticks.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  }
+  return removed;
+}
+
+MutationQueueTest::MutationQueueTest() : MutationQueueTestBase(GetParam()()) {
+}
+
+TEST_P(MutationQueueTest, CountBatches) {
+  persistence_->Run("test_count_batches", [&] {
+    ASSERT_EQ(0, BatchCount());
```
var-const: I think we generally tend to use `EXPECT` instead of `ASSERT` in our tests. I'm not against it, just curious about the rationale.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  SetMutation mutation = testutil::SetMutation(key, Map("a", 1));
+
+  MutationBatch batch =
+      mutation_queue_->AddMutationBatch(Timestamp::Now(), {}, {mutation});
+  return batch;
+}
+
+/**
+ * Creates an array of batches containing @a number dummy MutationBatches. Each
+ * has a different batch_id.
```
var-const: Looking at the tests, it appears that the invariant is actually stronger -- each subsequent batch has a higher batch ID. Is it worth calling out?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+
+    mutation_queue_->AcknowledgeBatch(batch1, {});
+    mutation_queue_->RemoveMutationBatch(batch1);
+
+    ASSERT_EQ(BatchCount(), 0);
+  });
+}
+
+TEST_P(MutationQueueTest, LookupMutationBatch) {
+  // Searching on an empty queue should not find a non-existent batch
```
var-const: Should this comment be moved inside the lambda?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  persistence_->Run("test_lookup_mutation_batch", [&] {
+    absl::optional<MutationBatch> not_found =
+        mutation_queue_->LookupMutationBatch(42);
+    ASSERT_EQ(not_found, absl::nullopt);
+
+    std::vector<MutationBatch> batches = CreateBatches(10);
+    std::vector<MutationBatch> removed = RemoveFirstBatches(3, &batches);
+
+    // After removing, a batch should not be found
+    for (size_t i = 0; i < removed.size(); i++) {
```
var-const: Use range-based for loop?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    // Remaining entries should still be found
+    for (const MutationBatch& batch : batches) {
+      absl::optional<MutationBatch> found =
+          mutation_queue_->LookupMutationBatch(batch.batch_id());
+      ASSERT_EQ(found->batch_id(), batch.batch_id());
+    }
+
+    // Even on a nonempty queue searching should not find a non-existent batch
```
var-const: Ultranit: I _think_ there should be a comma before `searching`.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+      const MutationBatch& current = batches[i];
+      const MutationBatch& next = batches[i + 1];
+      absl::optional<MutationBatch> found =
+          mutation_queue_->NextMutationBatchAfterBatchId(current.batch_id());
+      ASSERT_EQ(found->batch_id(), next.batch_id());
+    }
+
+    for (size_t i = 0; i < removed.size(); i++) {
+      const MutationBatch& current = removed[i];
+      const MutationBatch& next = batches[0];
```
var-const: Should the index here depend on `i` as well?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  });
+}
+
+TEST_P(MutationQueueTest, NextMutationBatchAfterBatchID) {
+  persistence_->Run("test_next_mutation_batch_after_batch_id", [&] {
+    std::vector<MutationBatch> batches = CreateBatches(10);
+    std::vector<MutationBatch> removed = RemoveFirstBatches(3, &batches);
+
+    for (size_t i = 0; i < batches.size() - 1; i++) {
+      const MutationBatch& current = batches[i];
```
var-const: Very optional: it _seems_ like it would be very slightly less verbose with iterators:
```cpp
  for (auto current = batches.begin(); current != batches.end() - 1; ++current) {
      auto next = current + 1;
      absl::optional<MutationBatch> found =
          mutation_queue_->NextMutationBatchAfterBatchId(current->batch_id());
      ASSERT_EQ(found->batch_id(), next->batch_id());
    }
```

Applies to some of the following loops as well.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+
+    for (size_t i = 0; i < removed.size(); i++) {
+      const MutationBatch& current = removed[i];
+      const MutationBatch& next = batches[0];
+      absl::optional<MutationBatch> found =
+          mutation_queue_->NextMutationBatchAfterBatchId(current.batch_id());
+      ASSERT_EQ(found->batch_id(), next.batch_id());
+    }
+
+    const MutationBatch& first = batches[0];
```
var-const: Nit: s/batches[0]/front().

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    absl::optional<MutationBatch> found =
+        mutation_queue_->NextMutationBatchAfterBatchId(first.batch_id() - 42);
+    ASSERT_EQ(found->batch_id(), first.batch_id());
+
+    const MutationBatch& last = batches[batches.size() - 1];
```
var-const: Nit: s/[batches.size() - 1]/back(). `batches[batches.size() - 1]` is what languages with rich standard libraries do.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    std::vector<MutationBatch> expected{batches[1], batches[2]};
+    std::vector<MutationBatch> matches =
+        mutation_queue_->AllMutationBatchesAffectingDocumentKey(
+            testutil::Key("foo/bar"));
+
+    ASSERT_EQ(matches, expected);
+  });
+}
+
+TEST_P(MutationQueueTest, AllMutationBatchesAffectingDocumentKeys) {
```
var-const: Can this test's name be made less similar to the preceding test? It took me three readings of each name before I could spot the difference.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    std::vector<MutationBatch> expected{batches[1], batches[2], batches[4]};
+    std::vector<MutationBatch> matches =
+        mutation_queue_->AllMutationBatchesAffectingDocumentKeys(keys);
+
+    ASSERT_EQ(matches, expected);
+  });
+}
+
+TEST_P(MutationQueueTest,
+       AllMutationBatchesAffectingDocumentKeys_handlesOverlap) {
```
var-const: Ultranit: should `handles` be capitalized?

var-const: Note: having some clarification separated by an underscore is, AFAIK, Java-ism, though I don't mind.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+  persistence_->Run(
+      "test_all_mutation_batches_affecting_document_keys_handlesOverlap", [&] {
```
var-const: Ultranit: s/handlesOverlap/handles_overlap/?

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+
+    ASSERT_EQ(matches, expected);
+  });
+}
+
+TEST_P(MutationQueueTest, RemoveMutationBatches) {
+  persistence_->Run("test_remove_mutation_batches", [&] {
+    std::vector<MutationBatch> batches = CreateBatches(10);
+
+    mutation_queue_->RemoveMutationBatch(batches[0]);
```
var-const: Perhaps use `front`? Applies everywhere in this function where only the first element is removed "in one go".

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
```diff
@@ -0,0 +1,432 @@
...
+    batches.erase(batches.begin());
+
+    ASSERT_EQ(BatchCount(), 9);
+
+    std::vector<MutationBatch> found;
+
+    found = mutation_queue_->AllMutationBatches();
```
var-const: Nit: why not copy-initialize `found` above instead of default-constructing and reassigning?

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+ public:
+  LevelDbMutationQueueTest()
+      : MutationQueueTestBase(PersistenceFactory()),
+        db_(static_cast<LevelDbPersistence*>(persistence_.get())->ptr()) {
+  }
+
+ protected:
+  void SetDummyValueForKey(const std::string& key);
+
+  DB* db_;
```
var-const: Optional: `= nullptr`?

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Creates a key that's structurally the same as LevelDbMutationKey except it
+ * allows for nonstandard table names.
+ */
+std::string MutationLikeKey(absl::string_view table,
+                            absl::string_view user_id,
+                            BatchId batch_id) {
+  std::string key;
+  OrderedCode::WriteSignedNumIncreasing(&key, 5);  // TableName
```
var-const: Nit: can you add some newlines in this function?

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+  OrderedCode::WriteString(&key, table);
+  OrderedCode::WriteSignedNumIncreasing(&key, 13);  // UserId
+  OrderedCode::WriteString(&key, user_id);
+  OrderedCode::WriteSignedNumIncreasing(&key, 10);  // BatchId
+  OrderedCode::WriteSignedNumIncreasing(&key, batch_id);
+  OrderedCode::WriteSignedNumIncreasing(&key, 0);  // Terminator
+  return key;
+}
+
+TEST_F(LevelDbMutationQueueTest, LoadNextBatchID_zeroWhenTotallyEmpty) {
```
var-const: Question: should `zero` be capitalized? Applies throughout this function.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  SetDummyValueForKey(LevelDbMutationKey::Key("foo", 6));
+  SetDummyValueForKey(LevelDbMutationKey::Key("foo", 2));
+  SetDummyValueForKey(LevelDbMutationKey::Key("foo", 1));
+
+  ASSERT_EQ(LoadNextBatchIdFromDb(db_), 7);
+}
+
+TEST_F(LevelDbMutationQueueTest, LoadNextBatchID_onlyFindsMutations) {
+  // Write higher-valued batch_i_ds in nearby "tables"
```
var-const: Nit: extra underscore in `batch_i_ds`.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+  std::vector<std::string> tables{"mutatio", "mutationsa", "bears", "zombies"};
```
var-const: Mutations, zombies... Feels like Resident Evil. No idea what the bears are doing here.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
```diff
@@ -0,0 +1,188 @@
...
+
+#include "Firestore/Protos/nanopb/firestore/local/mutation.nanopb.h"
+#include "Firestore/Protos/nanopb/google/protobuf/empty.nanopb.h"
+#include "Firestore/core/src/firebase/firestore/auth/user.h"
+#include "Firestore/core/src/firebase/firestore/local/leveldb_key.h"
+#include "Firestore/core/src/firebase/firestore/local/leveldb_mutation_queue.h"
+#include "Firestore/core/src/firebase/firestore/local/leveldb_persistence.h"
+#include "Firestore/core/src/firebase/firestore/local/reference_set.h"
+#include "Firestore/core/src/firebase/firestore/nanopb/byte_string.h"
+#include "Firestore/core/src/firebase/firestore/nanopb/fields_array.h"
```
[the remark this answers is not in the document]
wilhuff: Removed.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
wilhuff: I removed all of these underscore separators (and then made the capitalization match).

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/leveldb_mutation_queue_test.cc
wilhuff: Stephen Colbert (on his satirical show, _The Colbert Report_) famously called bears "godless killing machines" and "the greatest threat to America".

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: While `std::unique_ptr` is broadly compatible with forward declared classes, it's destructor is not. Moving the destructor of this class here after including the header that declares `Persistence` allows the header to mostly forward declare its dependencies.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Moved.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Rephrased.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done. Objective-C is sufficiently more verbose that it was helpful there.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done. Makes sense.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done and done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: `EXPECT` makes sense when the assertions are independent. In this test, if the prior state is wrong the next assertions are also going to be broken. `ASSERT` short circuits once the rest of the test will become invalid.

I've changed a few to `EXPECT` below where there are no assertion dependencies.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: I opted not to do this. Dissimilarity to the other ports outweighs the slight gain.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: It should not. The assertion here is that searching for a removed, lower index, will find the first non-removed entry. I've added a comment.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Done: "AffectingMultipleDocumentKeys".

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Removed the underscore and capitalized.

wilhuff: Removed.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Fixed all these strings. My conversion script botched these. They really should match the name of the test case.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Since this test concerns itself with things in the first few positions by index I've opted not to treat the first specially.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Initially I did it this way so that each of the groupings would look more similar but yeah, you're right.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.cc
wilhuff: Removed.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
wilhuff: Done.

Firestore/core/test/firebase/firestore/local/mutation_queue_test.h
wilhuff: Done.

PR #4212, opened 2019-10-31
Don't use the address of a TimeSlot as a key for removal

wilhuff:
After a `TimeSlot` executes, it deletes itself and removes itself from `ExecutorLibdispatch`'s `schedule_`. The `DelayedOperation` still referred to this now freed address. If another `TimeSlot` is scheduled, operator `new` happens to allocate a `TimeSlot` that reuses the address, and the
`DelayedOperation` is subsequently canceled, this will cancel the new task before it has run.

This change fixes this by using a dedicated counter for `TimeSlot`s that effectively isn't reused. Note that while it is possible for the counter to wrap around, this would take tens of years of continuous execution to happen and require cancelation of a `TimeSlot` created tens of years ago to create the possibility of collision which is a sufficiently impractical problem as to be ignorable.

This fixes a flake in `FSTTransactionTests`, where we were actually seeing address reuse collisions in 1 out of ~50 test runs.

Also included are a few trivial fixes suggested by clang-tidy.

### DIFF
--- a/Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
+++ b/Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
@@ -274,8 +274,8 @@ DelayedOperation ExecutorLibdispatch::Schedule(const Milliseconds delay,
   return DelayedOperation{[this, time_slot_id] {
     // `time_slot` might have been destroyed by the time cancellation function
     // runs, in which case it's guaranteed to have been removed from the
-    // schedule_. If the `time_slot_id` refers to a slot that has been removed,
-    // the call to `RemoveFromSchedule` will be a no-op.
+    // `schedule_`. If the `time_slot_id` refers to a slot that has been
+    // removed, the call to `RemoveFromSchedule` will be a no-op.
     RemoveFromSchedule(time_slot_id);
   }};
 }
--- a/Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
+++ b/Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
@@ -135,7 +135,7 @@ class TimeSlot {
     done_ = true;
   }
 
-  static void InvokedByLibdispatch(void* const raw_self);
+  static void InvokedByLibdispatch(void* raw_self);
 
  private:
   void Execute();
@@ -175,7 +175,7 @@ Executor::TaggedOperation TimeSlot::Unschedule() {
   return std::move(tagged_);
 }
 
-void TimeSlot::InvokedByLibdispatch(void* const raw_self) {
+void TimeSlot::InvokedByLibdispatch(void* raw_self) {
   auto const self = static_cast<TimeSlot*>(raw_self);
   self->Execute();
   delete self;
@@ -289,7 +289,7 @@ absl::optional<Executor::TaggedOperation>
 ExecutorLibdispatch::PopFromSchedule() {
   absl::optional<Executor::TaggedOperation> result;
 
-  RunSynchronized(this, [this, &result] {
+  RunSynchronized(this, [this, &result]() -> void {
     if (schedule_.empty()) {
       return;
     }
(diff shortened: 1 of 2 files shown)


Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
```diff
@@ -246,28 +260,33 @@ void MarkDone() {
...
 
-  auto const time_slot = new TimeSlot{this, delay, std::move(operation)};
   dispatch_after_f(delay_ns, dispatch_queue(), time_slot,
                    TimeSlot::InvokedByLibdispatch);
-  RunSynchronized(this, [this, time_slot] { schedule_.push_back(time_slot); });
-  return DelayedOperation{[this, time_slot] {
+
+  return DelayedOperation{[this, time_slot_id] {
     // `time_slot` might be destroyed by the time cancellation function runs.
     // Therefore, don't access any methods on `time_slot`, only use it as
```
var-const: The comment is outdated now. Maybe something like:
```
// `time_slot` might have been destroyed by the time cancellation function runs,
// in which case it's guaranteed to be already removed from the schedule. If the
// `time_slot_id` refers to a slot that has been removed, the call to
// `RemoveFromSchedule` will be a no-op.
```

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
```diff
@@ -289,17 +309,18 @@ void MarkDone() {
 ExecutorLibdispatch::PopFromSchedule() {
   absl::optional<Executor::TaggedOperation> result;
 
-  RunSynchronized(this, [this, &result] {
+  RunSynchronized(this, [this, &result]() -> void {
```
var-const: Hmm, I'm surprised an explicit return type is needed here, or is it just linter complaining?

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
```diff
@@ -246,28 +260,33 @@ void MarkDone() {
   // guaranteed to outlive the executor, and it's possible for work to be
   // invoked by libdispatch after the executor is destroyed. Executor only
   // stores an observer pointer to the operation.
+  TimeSlot* time_slot = nullptr;
+  TimeSlotId time_slot_id = 0;
+  RunSynchronized(this, [this, delay, &operation, &time_slot, &time_slot_id] {
+    time_slot_id = time_slot_counter_++;
```
var-const: Note: overall, this approach is similar to what `ExecutorStd` does. In that case, the increment is done in a separate function, partially to make it stand out and partially to provide a natural place for the comment on wrapping: https://github.com/firebase/firebase-ios-sdk/blob/8c0ab30279d5aa347571b8a04375b9979c96633b/Firestore/core/src/firebase/firestore/util/executor_std.cc#L110-L116

Perhaps consider doing the same? It would make the two executor implementations slightly more similar, and it would formalize the comment from the PR description.

wilhuff: Rephrased to mostly match.

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
wilhuff: Rephrased.

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
wilhuff: `clang-tidy` was complaining.

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
```diff
@@ -272,9 +272,10 @@ void MarkDone() {
                    TimeSlot::InvokedByLibdispatch);
 
   return DelayedOperation{[this, time_slot_id] {
-    // `time_slot` might be destroyed by the time cancellation function runs.
-    // Therefore, don't access any methods on `time_slot`, only use it as
-    // a handle to remove from `schedule_`.
+    // `time_slot` might have been destroyed by the time cancellation function
+    // runs, in which case it's guaranteed to have been removed from the
+    // schedule_. If the `time_slot_id` refers to a slot that has been removed,
```
var-const: Ultranit: please add backticks around `schedule_`, for consistency with the rest of the comment.

wilhuff: Done.

PR #4223, opened 2019-11-01
Merging the 6.12.0 release into master

maksymmalyhin:


scripts/check_whitespace.sh
```diff
@@ -29,14 +29,16 @@ git grep "${options[@]}" -- \
     ':(exclude)Firebase/CoreDiagnostics/FIRCDLibrary/Protogen/nanopb' \
     ':(exclude)Firebase/CoreDiagnostics/ProtoSupport' \
     ':(exclude)CoreOnly/NOTICES' \
+    ':(exclude)Firebase/Firebase/NOTICES' \
```
paulb777: Still catching up, but we really should avoid a third NOTICES file.

paulb777:
When there are non-trivial cherry-picks, I've found that it's cleaner to make a branch from master and manually merge (and resolve any conflicts) from the release branch before sending the merge-back PR.  

Otherwise github will add a bunch of changes already in master to the PR.

scripts/check_whitespace.sh
maksymmalyhin: removed

maksymmalyhin:
The only Travis failure is now is Messaging which is still failing in master

PR #4224, opened 2019-11-01
FIS: Use IID auth token to authorize IID migration.

maksymmalyhin:
- for more details see go/authenticating-iid-migration-to-fis


FirebaseInstallations/Source/Library/IIDMigration/FIRInstallationsIIDTokenStore.m
```diff
@@ -0,0 +1,154 @@
...
+
+  return resultPromise;
+}
+
+- (FBLPromise<NSData *> *)IIDDefaultTokenData {
+  FBLPromise<NSData *> *resultPromise = [FBLPromise pendingPromise];
+
+  NSMutableDictionary *keychainQuery = [self IIDDefaultTokenDataKeychainQuery];
+  NSError *error;
+  NSData *data = [FIRInstallationsKeychainUtils getItemWithQuery:keychainQuery error:&error];
```
chliangGoogle: So IID put all the keychain operations in its own serial queue and would there be race conditions if FIS try to access the same keychain from another queue?

FirebaseInstallations/Source/Library/FIRInstallationsItem.h
```diff
@@ -44,8 +44,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, nullable) FIRInstallationsStoredAuthToken *authToken;
 @property(nonatomic, assign) FIRInstallationsStatus registrationStatus;
 
-/// Instance ID checkin data imported from IID checkin store as a part of IID migration.
-@property(nonatomic, nullable) FIRInstallationsStoredIIDCheckin *IIDCheckin;
+/// Instance ID default auth token imported from IID store as a part of IID migration.
+@property(nonatomic, nullable) NSString *IIDAuthToken;
```
chliangGoogle: It's more of a IIDToken, but I'm okay to leave it if it's a hassle to rename.

FirebaseInstallations/Source/Library/IIDMigration/FIRInstallationsIIDTokenStore.m
maksymmalyhin: Yeah, there might be a race, but in most of the cases there shouldn't be one. In the common case the sequence will be following:
1. IID starts default token update if needed. For that is asks FIS for FID and waits
2. FIS attempts  the migration (if needed) and potentially reads from IID keychain
3. FIS finishes all keychain operations and passes FID to IID
4. IID continues their operations. FIS does not touch the IID keychain after the migration attempt.

So the race condition may happen only when IID modifies keychain without requesting FID which it does not do in the IID with FIS version, though it may be accidentally introduced in the future.

FirebaseInstallations/Source/Library/FIRInstallationsItem.h
maksymmalyhin: I renamed it to `IIDDefaultToken`

FirebaseInstallations/Source/Library/IIDMigration/FIRInstallationsIIDTokenStore.m
maksymmalyhin: Please let me know if you still have concerns about it. I'll address them in a follow up PR if needed.

PR #4227, opened 2019-11-02
Prevent race conditions in DatastoreTest

wilhuff:
Previously it was possible for tasks on the async queue to outlive the 
instance of the DatastoreTest which could crash or cause TSAN violations.

This fixes a flake that appeared in the AuthOutlivesDatastore test case.

Also improve readability of the error classification tests


Firestore/core/test/firebase/firestore/remote/datastore_test.cc
```diff
@@ -123,6 +126,8 @@ class DatastoreTest : public testing::Test {
     if (!is_shut_down) {
       Shutdown();
     }
+    // Ensure that nothing remains on the AsyncQueue before destroying it.
+    worker_queue->EnqueueBlocking([] {});
```
var-const: Note: we use this pattern so much, I wonder if we should create a test-only helper function like `DrainWorkerQueue`.

PR #4231, opened 2019-11-04
check_whitespaces.sh: ignore notice files

maksymmalyhin:
It is needed to be able to commit notices files without modifications.

#no-changelog


maksymmalyhin:
@morganchen12 Thanks!

PR #4244, opened 2019-11-05
Update Core CHANGELOG for M59

morganchen12:


Firebase/Core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v6.3.3 -- M59
+- [changed] Carthage and zip file distributions are now built with Xcode 11.
```
maksymmalyhin: nit: Maybe we should be specific and specify Xcode 11.0 version just to clear up any confusion? @paulb777 WDYT?

paulb777: Definitely we should exactly specify the Xcode version.

paulb777: Add a sentence that Xcode 10.x is no longer supported with this release from the zip file and Carthage.

morganchen12: Done (internal CL updated as well).

morganchen12: Done, updated internal CL as well.

PR #4248, opened 2019-11-06
ZipBuilder README update

maksymmalyhin:


maksymmalyhin:
README change only, merging without waiting for CI.

PR #4249, opened 2019-11-06
Require CMake 3.5.1; upgrade Google Benchmark to 1.5.0

wilhuff:
CMake 3.5.1 is the version required by the latest versions of gRPC and Google Benchmark. See discussion google/benchmark#613.

Remove version check around find_package(XCTest).

Also, add a dependency on objc protos that was missing but somehow implicitly being added before.

#no-changelog

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 # Superbuild for Firebase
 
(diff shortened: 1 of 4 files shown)


wilhuff:
Ignoring Xcode 10.2 failures based on #4254.

PR #4250, opened 2019-11-06
Upgrade GoogleTest to 1.10.0

wilhuff:
TYPED_TEST_CASE and friends have been renamed TYPED_TEST_SUITE.

#no-changelog

PR #4189, opened 2019-10-29
C++ migration: remove the dependency on Objective-C `Protobuf` library

var-const:

PR #4252, opened 2019-11-06
Upgrade gRPC to 1.24.3; upgrade protobuf to 3.9.2

wilhuff:
This is a massive change according to the diff stats, but really it's a few small changes plus changes to generated code. Review one commit at a time.

Note:
  * The CocoaPod-based build was already upgraded to protobuf 3.9.2 in #3916.
  * This happened to work because the Objective-C code generation changes are trivial.
  * When building this PR locally in CMake, you'll need to clear your output directory first; upgrading gRPC destroys the things we've unpacked into its directory tree but they're not re-unpacked. I'll address this limitation separately (if it's possible).

Note the diffbase: this depends on the merge of #4249 and #4250.

#no-changelog


paulb777:
Why not go to the latest minor version - 3.10.0 - which is what currently gets installed by default with Firestore?

cmake/external/boringssl.cmake
```diff
@@ -22,10 +22,13 @@ endif()
 # the SOURCE_DIR when unpacking so this must come after grpc despite the fact
 # that grpc logically depends upon this.
 
-# grpc at v1.8.3 includes boringssl at be2ee342d3781ddb954f91f8a7e660c6f59e87e5
+# grpc v1.8.3 includes boringssl at be2ee342d3781ddb954f91f8a7e660c6f59e87e5
```
var-const: Hmm, if we depend on `v1.24.3` now, it's not clear why the comment references `v1.8.3`, of all gRPC versions. Is this part of the comment still relevant? Perhaps consider either amending this paragraph slightly to make it clear why it's talking about `v1.8.3`:
```
# grpc v1.8.3, on which Firestore originally depended, includes boringssl...
```
or removing this paragraph and slightly rewording the next one:
```
# grpc v1.24.3 includes boringssl at b29b21a81b32ec273f118f589f46d56ad3332420
# (2018-05-22). While it does include gtest as a dependency, the version is woefully
# out of date.
```

wilhuff:
> Why not go to the latest minor version - 3.10.0 - which is what currently gets installed by default with Firestore?

We no longer use Objective-C protocol buffers for much--just pretty printing, and only when debug logging is enabled--and even that's going away once #4184 and #4189 land.

We also don't use C++ protocol buffers in production: we use nanopb. We only use C++ protobuf for compatibility testing: to verify that our usage of nanopb will produce the expected results on the backend.

Therefore the version chosen here matches the version specified by gRPC at version 1.24.3 to absolutely minimize integration pain. If 3.9.2 is what gRPC is testing we're not going to expend any effort toward choosing anything different.

cmake/external/boringssl.cmake
wilhuff: Done. You're right that the history isn't useful here. All that matters is that gRPC is woefully out of date so we're not following it exactly.

PR #4258, opened 2019-11-06
FIS: IID incompatible version exception.

maksymmalyhin:
- throw and exception if FIS is used with an incompatible IID version.

b/144091500

### DIFF
--- a/FirebaseInstallations.podspec
+++ b/FirebaseInstallations.podspec
@@ -44,12 +44,16 @@ Pod::Spec.new do |s|
       unit_tests.requires_app_host = true
       unit_tests.dependency 'OCMock'
       unit_tests.dependency 'FirebaseInstanceID', '~> 4.2.0' # The version before FirebaseInstanceID updated to use FirebaseInstallations under the hood.
+      s.pod_target_xcconfig = {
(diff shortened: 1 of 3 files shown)


FirebaseInstallations.podspec
```diff
@@ -44,12 +44,16 @@ Pod::Spec.new do |s|
...
+    }
     end
 
     s.test_spec 'integration' do |int_tests|
       int_tests.source_files = base_dir + 'Tests/Integration/**/*.[mh]'
       int_tests.resources = base_dir + 'Tests/Resources/**/*'
-      s.pod_target_xcconfig = {
+      int_tests.pod_target_xcconfig = {
       'GCC_PREPROCESSOR_DEFINITIONS' =>
         'FIR_INSTALLATIONS_INTEGRATION_TESTS_REQUIRED=1'
```
paulb777: This is always defined.   You might want to consider an approach like https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleDataTransport.podspec#L45

maksymmalyhin: Good catch! I'll update it.

PR #4260, opened 2019-11-07
Do not activate if fetch returns NO_CHANGE

dmandar:
Fixes internal bug: 143622863. We activate only if fetched data has not resulted in a data change. We track this using the last time the ETag was changed (corresponds to a real change to the RC template). 
This change also mirrors current android behavior.

### DIFF
--- a/FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h
+++ b/FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h
@@ -76,6 +76,8 @@
 @property(nonatomic, readwrite, assign) NSTimeInterval lastSetDefaultsTimeInterval;
 /// The latest eTag value stored from the last successful response.
 @property(nonatomic, readwrite, assign) NSString *lastETag;
+/// The timestamp of the last eTag update.
+@property(nonatomic, readwrite, assign) NSTimeInterval lastETagUpdateTime;
 
 #pragma mark Throttling properties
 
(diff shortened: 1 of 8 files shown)


FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h
```diff
@@ -76,6 +76,8 @@
 @property(nonatomic, readwrite, assign) NSTimeInterval lastSetDefaultsTimeInterval;
 /// The latest eTag value stored from the last successful response.
 @property(nonatomic, readwrite, assign) NSString *lastETag;
+/// The timestamp of the last eTag update.
+@property(nonatomic, readwrite, assign) NSTimeInterval lastETagUpdateTime;
```
chliangGoogle: should we persist this value? cuz what happened before the current app launch.

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
```diff
@@ -560,6 +560,132 @@ - (void)testFetchConfigsFailed {
...
+    settings.lastETagUpdateTime = 100;
+    settings.lastApplyTimeInterval = 101;
+
+    dispatch_queue_t queue =
+        dispatch_queue_create([[NSString stringWithFormat:@"testNoStatusFetchQueue: %d", i]
+                                  cStringUsingEncoding:NSUTF8StringEncoding],
+                              DISPATCH_QUEUE_SERIAL);
+    _configFetch[i] = OCMPartialMock([[RCNConfigFetch alloc] initWithContent:configContent
+                                                                   DBManager:_DBManager
+                                                                    settings:settings
```
chliangGoogle: is this format right? Maybe run script/style.sh

FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h
dmandar: good pt. Done.

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
dmandar: This is with style.sh.

PR #4262, opened 2019-11-07
IID: indication of compatibility with FIS

maksymmalyhin:
b/144091500

NOTE: merge after #4258


maksymmalyhin:
NOTE: Auth xcodebuild tests are failing due to `FirebaseInstallations` unavailable on Cocoapods. This will be fixed closer to the release by pushing a `FirebaseInstallations.podspec` to the staging repo.

Firebase/InstanceID/FIRInstanceID+Private.m
```diff
@@ -57,4 +57,12 @@ - (NSString *)appInstanceID:(NSError **)outError {
   return instanceID;
 }
 
+#pragma mark - Firebase Installations Compatibility
+
+/// Presence of this method indicates that this version of IID uses FirebaseInstallations under the
+/// hood. It is checked by FirebaseInstallations SDK.
```
chliangGoogle: Should we only merge this when IID is released with FIS in public?

maksymmalyhin: Yeah, this PR is for `mm/fis-integration-master`, which will be merged only before IID+FIS release.

paulb777: How can FIS call IID?  I thought IID will depend on FIS?

maksymmalyhin: @paulb777  [Here](https://github.com/firebase/firebase-ios-sdk/pull/4258/files#diff-98cba1aed9149678f501b88c09a93b64R200) is usage of it. So basically FIS checks if the class is present in runtime and if yes if the method exists. This way of checking version compatibility looks like reliable enough as well as more flexible and easier to implement comparing to reading and validating of `FIRInstanceID_LIB_VERSION`. If you can see a better way, I am open for discussion.

paulb777: I see. Thanks! LGTM

PR #4266, opened 2019-11-08
Add a changelog entry on `Protobuf` removal.

var-const:
Removed in #4189.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+- [changed] If you're manually tracking dependencies, Firestore SDK no longer
+  depends on `Protobuf` framework (note, however, that other Firebase components 
+  might still require it).
 
 # v1.7.0
 - [feature] Added `whereField(_:in:)` and `whereField(_:arrayContainsAny:)` query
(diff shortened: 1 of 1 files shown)


var-const:
@wilhuff Seems like something worth calling out. What do you think?

var-const:
@wilhuff The [Travis failure](https://travis-ci.org/firebase/firebase-ios-sdk/jobs/609032369?utm_medium=notification&utm_source=github_status) doesn't make sense to me. Also, when I run the linter locally, it has no complaints. Any ideas?

paulb777:
Trailing space ...

Screen Shot 2019-11-07 at 7 07 15 PM

Firestore/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- [changed] If you're manually tracking dependencies, Firestore SDK no longer
```
wilhuff: This phrasing hides what actually changed. What changed should be first and how to potentially handle it should come after.

How about:

- [changed] Removed the Firestore's dependency on `Protobuf`. If you're manually tracking dependencies you may be able to remove it from your build (note, however, that other Firebase components may still require it).

var-const: Done.

var-const:
Paul, thanks for spotting this! Looks like my Vim isn't properly configured for Markdown files.

PR #4267, opened 2019-11-08
Simplify code by using `Message` where applicable

var-const:
Also:
* remove all direct references to Nanopb-generated `foo_bar_fields` arrays;
* minimize references to `Message<T>::fields()` (only in `MakeFoo` functions now);
* remove `FreeNanopbMessage` function (it would now only be called from `Message`'s destructor, which now instead calls `pb_release` directly).


Firestore/core/src/firebase/firestore/local/leveldb_migrations.cc
```diff
@@ -112,11 +113,8 @@ void ClearQueryCache(leveldb::DB* db) {
   LevelDbTransaction transaction(db, "Drop query cache");
 
   // Reset the target global entry too (to reset the target count).
-  firestore_client_TargetGlobal target_global{};
-
-  nanopb::StringWriter writer;
-  writer.Write(firestore_client_TargetGlobal_fields, &target_global);
-  transaction.Put(LevelDbTargetGlobalKey::Key(), writer.Release());
+  Message<firestore_client_TargetGlobal> target_global;
+  transaction.Put(LevelDbTargetGlobalKey::Key(), MakeStdString(target_global));
```
wilhuff: `Put` has an overload that takes `Message<T>` and does the `MakeStdString` for you already.

Firestore/core/test/firebase/firestore/remote/serializer_test.cc
```diff
@@ -243,19 +238,16 @@ class SerializerTest : public ::testing::Test {
...
-    google_firestore_v1_Document proto = serializer.EncodeDocument(key, value);
-    writer.Write(google_firestore_v1_Document_fields, &proto);
-    FreeNanopbMessage(google_firestore_v1_Document_fields, &proto);
-    return writer.Release();
+    return Encode(serializer.EncodeDocument(key, value));
+  }
+
+  template <typename T>
+  ByteString Encode(T&& nanopb_proto) {
+    return MakeByteString(MakeMessage(std::move(nanopb_proto)));
```
wilhuff: Given the argument is a forwarding reference, shouldn't this be `std::forward<T>`?

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -198,6 +195,23 @@ Message<T> Message<T>::TryParse(Reader* reader) {
...
+ * `std::make_pair`.
+ */
+// The resulting message takes ownership of the given proto, so only rvalue
+// references should be accepted. Unfortunately, because this function is
+// a template, the argument type is actually a forwarding reference, which would
+// bind to lvalues and rvalues alike. To work around this, use some
+// metaprogramming.
+template <typename T,
+          typename = absl::enable_if_t<!std::is_lvalue_reference<T>::value>>
+Message<T> MakeMessage(T&& proto) {
```
wilhuff: Given this method has exactly one caller and the complexity of the metaprogramming required to implement it, is this really worth it?

I guess I can see that this might be commonly used in combination with the serializer, though I think it leads to some potentially hard to reason about code:

```c++
auto message = MakeMessage(serializer_.EncodeDocument(...));
```

Should we add a move-assignment `operator=` so that we could write this?

```c++
Message<google_firestore_v1_Document> message = serializer_.EncodeDocument(...);
```

Firestore/core/src/firebase/firestore/nanopb/message.h
```diff
@@ -198,6 +195,23 @@ Message<T> Message<T>::TryParse(Reader* reader) {
...
+  Message<T> result;
+  *result = proto;
```
wilhuff: `std::move(proto)`  ?

paulb777:
Closing for inactivity. Please reopen to continue.

PR #4273, opened 2019-11-08
Remove workarounds for nanopb ODR violations

wilhuff:
Now that we depend upon gRPC 1.24+ (and it no longer includes nanopb) we
no longer need any build order workarounds.

Fixes b/111328563. #no-changelog


var-const:
Does this mean that we no longer _have_ to put the generated Nanopb code in a namespace?

wilhuff:
The generated messages still have to be in a namespace because there's a possibility that other nanopb-using components may also ship definitions for common types like `google_protobuf_Timestamp`. Unless we centralize the common definitions this would still lead to ODR violations on the messages, even if we're no longer ODR conflicting on the nanopb runtime itself.

PR #4275, opened 2019-11-08
Get tests passing on Linux

wilhuff:
`{` and `}` are metacharacters and need to be escaped to be used as literals.

Without this, the tests log warnings like this:

```
Regular expression "<Write 0x[0-9A-Fa-f]+>: {
  update_mask {
  }
}" is not a valid POSIX Extended regular expression.
```

and all the matches fail.

### DIFF
--- a/Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.cc
+++ b/Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.cc
@@ -52,7 +52,6 @@ using nanopb::Message;
 using nanopb::Reader;
 using remote::ByteBufferReader;
 using remote::Serializer;
-using util::MakeString;
 using util::Status;
 using util::StatusOr;
 using util::StringFormat;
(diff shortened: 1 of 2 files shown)


Firestore/core/src/firebase/firestore/remote/remote_objc_bridge.cc
```diff
@@ -52,7 +52,6 @@ using nanopb::Message;
 using nanopb::Reader;
 using remote::ByteBufferReader;
 using remote::Serializer;
-using util::MakeString;
```
var-const: Hmm, would it be possible to configure Clang-tidy to catch unused `using` declarations? (not for this PR)

wilhuff: I have yet to find a way to make `clang-tidy` available for CI, short of building and distributing our own binary. My attempts to build my own binary have been stymied by https://github.com/include-what-you-use/include-what-you-use/issues/100: clang-tidy can't find any headers available in the SDK leading to thousands of errors building and no useful feedback.

PR #4276, opened 2019-11-08
Run integration tests from CMake on mac

wu-hui:


Firestore/Source/CMakeLists.txt
```diff
@@ -90,4 +91,83 @@ if(APPLE)
     PUBLIC
     ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public
   )
-endif()
+
+  objc_framework(
+    FirebaseFirestore
+    HEADERS
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FirebaseFirestore.h
```
wu-hui: This is failing `check.sh`, I tried to use `Public/FirebaseFirestore.h` here but then the imports from integration tests fail

Firestore/Source/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ if(APPLE)
       API/FIRQuerySnapshot.mm
       API/FIRSnapshotMetadata+Internal.h
       API/FIRSnapshotMetadata.mm
+      API/FIRTimestamp.m
```
wu-hui: This was not included, was it intentional?

Firestore/Example/Tests/Util/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+find_package(XCTest REQUIRED)
+
+objc_framework(
+  firebase_firestore_tests_util
+  HEADERS
+    FSTEventAccumulator.h
+    # XCTestCase+Await.h
```
wu-hui: Note the hack here: I moved `XCTestCase+Await` into `FSTIntegrationTestCase` to avoid category..I could not figure out how to make that work.

Firestore/Example/App/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_executable(macOS_Example_App MACOSX_BUNDLE
+  macOS/main.m
+  macOS/AppDelegate.m
+  macOS/AppDelegate.h
+)
+
+add_objc_flags(macOS_Example_App
```
wu-hui: Note I have to call `add_objc_flags` with the same files as above, ideally we don't need to do this at all. Maybe we need a `objc_bundle` target?

Firestore/Example/Tests/Integration/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+find_package(XCTest REQUIRED)
+enable_testing()
+
+xctest_add_bundle(firebase_firestore_objc_integration_tests
```
wilhuff: The contents of this file should likely be wrapped in an `if(APPLE) ... endif()` to prevent this from breaking the Linux and Windows builds.

Firestore/Example/Tests/Util/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+find_package(XCTest REQUIRED)
+
+objc_framework(
```
wilhuff: Does this need to be an `objc_framework` or can it just be a `cc_library`?

Firestore/Source/CMakeLists.txt
wilhuff: An omission.

Firestore/Example/Tests/Util/CMakeLists.txt
wilhuff: The way to get this to work is to modify the linker command line to include the `-ObjC` flag. Use `target_link_options` like so:

```cmake
target_link_options(
  firebase_firestore_tests_util
  PUBLIC -ObjC
)
```

This might not work though, in which case this may need to be set as a private option on executables that link all this stuff together.

Firestore/Example/Tests/Util/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+  firebase_firestore_tests_util
```
wilhuff: Objective-C only targets get an `_objc` suffix elsewhere. I think this likely makes sense here too.

Firestore/Source/CMakeLists.txt
```diff
@@ -90,4 +91,83 @@ if(APPLE)
...
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRGeoPoint.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRListenerRegistration.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRQuery.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRQuerySnapshot.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRSnapshotMetadata.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRTimestamp.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRTransaction.h
+      ${FIREBASE_SOURCE_DIR}/Firestore/Source/Public/FIRWriteBatch.h
+    SOURCES
+      API/FIRCollectionReference+Internal.h
```
wilhuff: Is there any way to get this to depend on any of the existing libraries that we've already defined?

wu-hui:
Merged with PR #4279 and `origin/master`.

Firestore/Example/App/CMakeLists.txt
wilhuff: I'm not opposed, but we only have one of these at this point, so I'd suggest deferring this work until we have some idea how this works in more than one circumstance.

Note that this style of calling multiple functions to describe what's required to build a single target is common in CMake. Our concise helpers that make things look more like blaze/bazel are uncommon. 

In short, I think this is fine, but up to you.

Firestore/Example/Tests/Util/CMakeLists.txt
wilhuff: I've restored XCTestCase+Await into a separate file, but had to work around this by explicitly linking. If you agree this is OK, let's defer further work on this.

Firestore/Example/App/macOS/AppDelegate.m
```diff
@@ -18,16 +18,18 @@
 
 @interface AppDelegate ()
 
-@property(weak) IBOutlet NSWindow *window;
+@property(atomic) IBOutlet NSWindow *window;
```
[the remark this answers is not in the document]
wu-hui: I have to disable `-Wimplicit-atomic-properties` for this to compile though.

Firestore/Example/App/macOS/AppDelegate.m
```diff
@@ -18,16 +18,18 @@
...
 @end
 
 @implementation AppDelegate
 
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
+  [aNotification name];
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/Example/App/macOS/AppDelegate.m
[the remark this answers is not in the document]
wilhuff: That warning is flagging that fields have the atomic `atomic` attribute by default and is asking you to make it explicit. Properties have multiple attributes though, so this should /in addition to/ the `weak` attribute that's already there. That is, it should be:

```objc
@property(weak, atomic) IBOutlet NSWindow *window;
```

cmake/compiler_setup.cmake
```diff
@@ -83,7 +83,6 @@ if(APPLE)
     -Werror=objc-root-class
 
     -Wblock-capture-autoreleasing
-    -Wimplicit-atomic-properties
```
wilhuff: We can leave this in, I think.

wu-hui: Done.

cmake/cc_rules.cmake
```diff
@@ -329,6 +354,43 @@ function(objc_framework target)
...
+
+    xctest_add_test(
+      ${target}
+      ${target}
+    )
+
+    if(WITH_ASAN)
+      set_property(TEST ${target} APPEND PROPERTY
+        ENVIRONMENT
+        DYLD_INSERT_LIBRARIES=/usr/local/opt/llvm/lib/clang/8.0.0/lib/darwin/libclang_rt.asan_osx_dynamic.dylib)
```
wu-hui: I was able to fix the asan issue locally with this change. However, I could not figure out a way to set this ENV variable reliably both on my local machine and on travis. 

Ideally the compiler could tell us where this `libclang_rt.asan_osx` is, or we could infer from the compiler being used.

cmake/FindASANDylib.cmake
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Finds `libclang_rt.asan_osx_dynamic.dylib` on Apple platform, it is typically
+# under ${CMAKE_CXX_COMPILER}/../lib.
+get_filename_component(compiler_bin_dir ${CMAKE_CXX_COMPILER} DIRECTORY)
+get_filename_component(compiler_dir ${compiler_bin_dir} DIRECTORY)
+
+file(GLOB_RECURSE ASAN_DYLIB
```
wilhuff: We can refine this so that it logs what it's doing better, but let's do so in a follow-up PR.

PR #4281, opened 2019-11-11
FIRStorage.callbackQueue fixes

maksymmalyhin:
- fixes from cl/278915086 applied


Firebase/Storage/FIRStorage.m
```diff
@@ -236,6 +236,10 @@ - (FIRStorageReference *)referenceWithPath:(NSString *)string {
   return reference;
 }
 
+- (dispatch_queue_t)callbackQueue {
+  return _fetcherServiceForApp.callbackQueue;
```
schmidt-sebastian: Hm. Sorry for already approving this, but doesn't this break the ability for the developer to specify an override for the callback queue?

schmidt-sebastian: Apparently the solution to this mystery lies in line 244 :) Please ignore.

PR #4291, opened 2019-11-12
LastFetchTime is a read-only property. Mark it as such.

dmandar:


chliangGoogle:
My understanding this is a breaking change. We should have an API review on this.

dmandar:
Ack.

dmandar:
Will reopen at I/O time. Tracked in internal b/144366943

PR #4293, opened 2019-11-12
Update RC ReleaseNotes for M60.

dmandar:


FirebaseRemoteConfig/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.4.5
+- Do not re-activate if a NO_CHANGE has been signaled by the Remote Config backend. (#4260)
```
morganchen12: Prefix with `[changed]` and rephrase in past tense. Something like `Remote Config no longer re-activates the current config on fetch if it receives no changes from the backend. (#4260)`.

dmandar: missed seeing these..will follow up in another PR

PR #4296, opened 2019-11-12
Update CHANGELOG for Firestore v1.8.0

wilhuff:


Firestore/CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
 # Unreleased
-- [changed] Removed the Firestore's dependency on `Protobuf`. If you're manually
-  tracking dependencies, you may be able to remove it from your build (note,
-  however, that other Firebase components may still require it).
+
+# v1.7.1
+- [changed] Removed the Firestore's dependency on the `Protobuf` CocoaPod. If
```
morganchen12: nit: First `the` can be removed

paulb777:
It might be ambiguous, but my feeling is that a dependency change should be a minor version update - since it's definitely more than a bug fix and the resulting installation changes could expose extra issues in non-conventional build environments.

wilhuff:
> It might be ambiguous, but my feeling is that a dependency change should be a minor version update - since it's definitely more than a bug fix and the resulting installation changes could expose extra issues in non-conventional build environments.

Changed to 1.8.0.

PR #4264, opened 2019-11-07
Update nanopb to 0.3.9.5 (CocoaPods 1.30905.0)

paulb777:
Fix #5191 

See https://github.com/google/nanopb-podspec/pull/13 for a description of the new nanopb CocoaPods versioning scheme.

Updates test infrastructure to test preliminary versions of Analytics pod structure (but not functional) changes.

See #5455 for the corresponding no-op nanopb code gen changes.

Googlers - see also cl/308045137

#no-changelog

PR #4312, opened 2019-11-14
Upgrade gRPC-C++ to 1.28.0

wilhuff:
This unblocks nanopb changes proposed in #4264.

This upgrades everything in the gRPC constellation of dependencies to match gRPC at 1.28:

  * gRPC to 1.28.0 (note that gRPC 1.28.1, while released, is a fix for just the python binding and there's no gRPC-C++ CocoaPod at that version)
  * Abseil to 20200225 (as of gRPC 1.27 this is now a gRPC dependency too; there's a later patch but gRPC doesn't use it yet)
  * Protobuf C++ to 3.11.4 (only used for testing in Firestore)
  * C-Ares to 1.15.0
  * BoringSSL to match (no change between gRPC 1.27.0 and 1.28.0)

This is best reviewed as individual commits, largely ignoring 243bef4, which is a commit of generated code to keep our CocoaPod-based build simple.


wilhuff:
Note that I've rebased this change onto current master and updated the versions this tackles. So far this hasn't run into any of the issues I saw with gRPC 1.25.0.

wilhuff:
Argh. I've been re-running the integration tests repeatedly and can still see the occasional flake. This isn't safe to merge as-is :-(.

wilhuff:
PTAL, 5e74ae5e240f16bd225c21d1b758f926d54b050c represents a significant change.

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -120,6 +121,22 @@ HostConfigMap& Config() {
   return config_by_host_;
 }
 
+#if __APPLE__
+// Disable CFStream-based transport on Apple platforms due to b/133182964, where
+// CFStream will occasionally fail to raise a has-bytes-available events,
+// causing Firestore to appear to hang.
+//
+// Use a constructor of a globally scoped object to set the `grpc_cfstream`
```
var-const, line 129: This is ingenious, though I'd like to double-check if it's necessary. Does the gRPC team recommend doing it this way? Is there perhaps some well-defined point better suited for `setenv` calls? Note that this is mostly me playing the devil's advocate -- I'm pretty sure this is a good implementation.

wilhuff: gRPC team didn't have any more advice than "set `grpc_cfstream=0`". Here's their official documentation: https://github.com/grpc/grpc/blob/master/src/objective-c/README-CFSTREAM.md

In general, the best way to handle calls to `setenv` is to put them at the top of your `main` function. This avoids data races that are otherwise inherent in multiple threads setting and getting environment variables concurrently. As a library author we don't really control `main` so it's kind of a crapshoot as to whether or not we'll make the call before anyone else uses gRPC.

Indeed, my initial implementation was to put this in the `GrpcConnection` constructor:

```c++
#if __APPLE__
  static std::once_flag grpc_cfstream_disabled;

  std::call_once(grpc_cfstream_disabled, [] {
    setenv("grpc_cfstream", "0", 1);
  });
#endif  // __APPLE__
```

However, this didn't work. In gRPC-C++, the call to `grpc_init` happens implicitly and it seems to happen for essentially any API call. Other calls must have happened earlier and prevented the variable from being read.

I looked for other places insert this to make this earlier (for example, making a static `Datastore::InitGrpc` and calling it from the `api::Firestore` constructor (or similar) but this causes the problem that every different way we construct Firestore or FirestoreClient instances has to be adjusted or it will be subject to flakes if we forget. I didn't want to take the chance that we might forget to make this call so this approach seemed better.

var-const: Thanks for elaborating on the rationale. I think it's a great solution.

PR #4324, opened 2019-11-15
Use uncompressed proto data if gzip fails or is larger

mikehaney24:


GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTFLLUploader.m
```diff
@@ -321,10 +323,14 @@ - (nonnull NSData *)constructRequestProtoFromPackage:(GDTCORUploadPackage *)pack
  * @return A new NSURLRequest ready to be sent to FLL.
  */
 - (NSURLRequest *)constructRequestWithURL:(NSURL *)URL data:(NSData *)data {
+  const UInt8 *bytes = (const UInt8 *)data.bytes;
+  BOOL isGzipped = (data.length >= 2 && bytes[0] == 0x1f && bytes[1] == 0x8b);
```
maksymmalyhin: Can we just pass this flag to the method based on the way we get the data or the presence of the header in the original request? My guess it will be a more reliable approach.

mikehaney24: Initially I did that, but it won't work because in redirect cases whether the data was gzipped or not is lost information. Additionally, if this mechanism doesn't work, it's not gzip data by definition.

maksymmalyhin: Is `Content-Encoding` removed from the original request?

mikehaney24: Creating the original request also comes through this method, so the Content-Encoding will be set correctly in every context.

maksymmalyhin: Discussed offline. Agreed to add comments for the magical bytes.

Issue #4308, opened 2019-11-13
FirebaseDynamicLink crashes due to object deallocation on app launch.


alan-andrade:
# Environment
  * Xcode version: 11.2
  * Firebase SDK version: 6.12.0
  * Firebase Component:  DynamicDeeplinks
  * Component version: 4.0.5

# Problem
We are in the process of creating a shared framework that contains Firebase. When the framework is imported on the new app, the application crashes right away while trying to retrieve pending deeplinks.

# Steps to reproduce:

1. Create a shared framework with firebase in it.
2. Import it in a new project and `configure` firestore on th app delegate.
3. The main application will crash.

There aren't any custom domains on the Info.plist.

# Relevant Code:
Snippet from `FIRDLJavaScriptExecutor.m`
```objc
#pragma mark - WKNavigationDelegate

- (void)webView:(WKWebView *)webView
    didFinishNavigation:(null_unspecified WKNavigation *)navigation {
  __weak __typeof__(self) weakSelf = self;

  // Make sure that the javascript was loaded successfully before calling the method.
  [webView evaluateJavaScript:FIRDLTypeofFingerprintJSMethodNameString()
            completionHandler:^(id _Nullable typeofResult, NSError *_Nullable typeError) {
              if (typeError) {
                [weakSelf handleExecutionError:typeError];
                return;
              }
              if ([typeofResult isEqual:@"function"]) {
                [webView
                    evaluateJavaScript:GINFingerprintJSMethodString()
                     completionHandler:^(id _Nullable result, NSError *_Nullable functionError) {
                       if ([result isKindOfClass:[NSString class]]) {
                         [weakSelf handleExecutionResult:result];
                         // ^^^^^^^^ CRASH due to weakSelf being deallocated.
                       } else {
                         [weakSelf handleExecutionError:nil];
                       }
                     }];
              } else {
                [weakSelf handleExecutionError:nil];
              }
            }];
}
```

# Stacktrace
```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x54892ed86e80)
    frame #0: 0x00007fff50bad00b libobjc.A.dylib`objc_release + 11
[...]

PR #4325, opened 2019-11-15
Keeps a strong reference for handling dynamic deeplinks

alan-andrade:
### Discussion
- https://github.com/firebase/firebase-ios-sdk/issues/4308

Sending the proposed patch by admins. Wondering still if it's a good idea though.

### DIFF
--- a/Firebase/DynamicLinks/FIRDLJavaScriptExecutor.m
+++ b/Firebase/DynamicLinks/FIRDLJavaScriptExecutor.m
@@ -113,10 +113,11 @@ NSString *GINFingerprintJSMethodString() {
                 [webView
                     evaluateJavaScript:GINFingerprintJSMethodString()
                      completionHandler:^(id _Nullable result, NSError *_Nullable functionError) {
+                       __typeof__(self) strongSelf = weakSelf;
                        if ([result isKindOfClass:[NSString class]]) {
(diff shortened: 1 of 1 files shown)


Firebase/DynamicLinks/FIRDLJavaScriptExecutor.m
```diff
@@ -113,10 +113,11 @@ - (void)webView:(WKWebView *)webView
                 [webView
                     evaluateJavaScript:GINFingerprintJSMethodString()
                      completionHandler:^(id _Nullable result, NSError *_Nullable functionError) {
+                       __typeof__(self) strongSelf = weakSelf;
```
morganchen12: Taking a closer look at the code, I'm not actually convinced `self` needs to be weak in either of these blocks. Removing the weak self declaration and using `self` instead looks like a simpler solution. @dmandar what do you think?

alan-andrade:
@googlebot I signed it!

Firebase/DynamicLinks/FIRDLJavaScriptExecutor.m
dmandar: Unless there is a reference cycle, there isn;'t an explicit need. It is fine to add this since we already had the weakself.

paulb777:
Thanks for the contribution @alan-andrade !

PR #4328, opened 2019-11-15
Add +[FIRServerValues increment:]

inlined:
* Definition is currently in an "unreleased" extension to FIRServerValues
* Functions that resolve server values now need access to any existing snapshot; please help ensure that this is not done wastefully.
* Tests are added to FData.m (integration tests); please let me know if a more appropriate place exists.
* Some tests are renamed to make it clear that ServerValues != timestamps

### DIFF
--- a/Firebase/Database/Api/FIRServerValue_unreleased.h
+++ b/Firebase/Database/Api/FIRServerValue_unreleased.h
@@ -24,12 +24,13 @@
 /**
  * Adds the given delta to the current value at a location.
  *
- * The delta must be an long or a double value. If the current value is not an integer or
- * double, or if the data does not yet exist, the transformation will set the data
- * to the delta value. If either of the delta value or the existing data
- * are doubles, both values will be interpreted as doubles. Double arithmetic
- * and representation of double values follow IEEE 754 semantics. If there is
- * positive/negative integer overflow, the sum is calculated as a  a double.
+ * The delta must be an long or a double value. If the current value is not an
+ * integer or double, or if the data does not yet exist, the transformation will
+ * set the data to the delta value. If either of the delta value or the existing
+ * data are doubles, both values will be interpreted as doubles. Double
+ * arithmetic and representation of double values follow IEEE 754 semantics. If
+ * there is positive/negative integer overflow, the sum is calculated as a  a
+ * double.
  *
  * @param delta the amount to modify the current value atomically.
  * @return a placeholder value for modifying data atomically server-side.
--- a/Firebase/Database/Api/FIRServerValue_unreleased.h
+++ b/Firebase/Database/Api/FIRServerValue_unreleased.h
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 Google
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FIRServerValue_unreleased_h
+#define FIRServerValue_unreleased_h
+
+#import "FIRServerValue.h"
+
+@interface FIRServerValue (Unreleased)
+
+/**
+ * Adds the given delta to the ref's current value.
+ *
+ * This must be an long or a double value. If the ref is not an integer or
+ * double, or if the ref does not yet exist, the transformation will set the ref
+ * to the given value. If either of the given value or the current field value
+ * are doubles, both values will be interpreted as doubles. Double arithmetic
+ * and representation of double values follow IEEE 754 semantics. If there is
+ * positive/negative integer overflow, the field is resolved to a double.
+ * @param delta the amount to modify the ref's current value atomically.
+ * @return a placeholder value for modifying a ref atomically server-side.
+ */
++ (NSDictionary *)increment:(NSNumber *)delta NS_SWIFT_NAME(increment(_:));
+
+@end
+
+#endif /* FIRServerValue_unreleased_h */
--- a/Firebase/Database/Api/FIRServerValue_unreleased.h
+++ b/Firebase/Database/Api/FIRServerValue_unreleased.h
@@ -22,16 +22,17 @@
 @interface FIRServerValue (Unreleased)
 
 /**
- * Adds the given delta to the ref's current value.
+ * Adds the given delta to the current value at a location.
  *
- * This must be an long or a double value. If the ref is not an integer or
- * double, or if the ref does not yet exist, the transformation will set the ref
- * to the given value. If either of the given value or the current field value
+ * The delta must be an long or a double value. If the current value is not an integer or
+ * double, or if the data does not yet exist, the transformation will set the data
+ * to the delta value. If either of the delta value or the existing data
  * are doubles, both values will be interpreted as doubles. Double arithmetic
  * and representation of double values follow IEEE 754 semantics. If there is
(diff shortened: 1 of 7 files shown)


paulb777:
Looks like the wrong version of clang-format was run. See https://github.com/firebase/firebase-ios-sdk#code-formatting.  

Please revert the RemoteConfig and Installations changes

Firebase/Database/Core/FRepo.m
```diff
@@ -770,18 +781,23 @@ - (void)runOnDisconnectEvents {
     FFLog(@"I-RDB038019", @"Running onDisconnectEvents");
     NSDictionary *serverValues =
         [FServerValues generateServerValues:self.serverClock];
-    FSparseSnapshotTree *resolvedTree =
-        [FServerValues resolveDeferredValueTree:self.onDisconnect
-                               withServerValues:serverValues];
+    // TODO(inlined): Fix
```
schmidt-sebastian: +1 :)

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,102 @@
...
+
++ (id)resolveComplexServerOp:(NSDictionary *)op
+                withExisting:(id<FNode>)existing
+                serverValues:(NSDictionary *)serverValues {
+    // Only increment is supported as of now
+    if (op[kIncrement] == nil) {
+        return nil;
+    }
+
+    NSNumber *delta = op[kIncrement];
```
schmidt-sebastian: Do you want to include the comment from the JS PR?

``` // Incrementing a non-number sets the value to the incremented amount```

File: Firebase/Database/Core/FServerValues.m [hunk 3]
```diff
@@ -20,37 +20,102 @@
...
+    if (![existing isLeafNode]) {
+        return delta;
```
schmidt-sebastian: If I am not mistaken (and I may very well be) "op[kIncrement]" could return something that is not an NSNumber. We might need to verify that it is indeed of a numeric type.

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,102 @@
...
+    }
+    FLeafNode *existingLeaf = existing;
+    if (![existingLeaf.value isKindOfClass:NSNumber.class]) {
+        return delta;
+    }
+
+    BOOL incrLong = strcmp(delta.objCType, @encode(int)) == 0;
+    incrLong = incrLong || strcmp(delta.objCType, @encode(long)) == 0;
+
```
schmidt-sebastian: We use this in Firestore: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/API/FSTUserDataConverter.mm#L402

I didn't verify that your check is equivalent but wanted to check if you did? Thanks!

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,102 @@
...
+    if (incrLong && baseLong) {
+        long longSum = delta.longValue + existingNum.longValue;
+        // check for under/overflow
+        if (delta.longValue < 0 && existingNum.longValue < 0 && longSum > 0) {
+            return @(doubleSum);
+        }
+        if (delta.longValue > 0 && existingNum.longValue > 0 && longSum < 0) {
+            return @(doubleSum);
+        }
+        return @(doubleSum);
```
schmidt-sebastian: If I am not mistaken, this should be `@(longSum)`. Can you write a test that verified the output type?

Firebase/Database/Core/FServerValues.m
inlined: FYI: I tried using the more readable `@encode` but it breaks switch statements because @encode isn't a constant expression (too bad this isn't Obj-C++...). I used to know this but I guess I'm rusty 🤷‍♂️

Firebase/Database/Api/FIRServerValue_unreleased.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef FIRServerValue_unreleased_h
+#define FIRServerValue_unreleased_h
+
+#import "FIRServerValue.h"
+
+@interface FIRServerValue (Unreleased)
+
+/**
+ * Adds the given delta to the ref's current value.
```
schmidt-sebastian: Nit: Please replace usage of "ref's" with something else. Example would be: 

```
Adds the given delta to the current value at the current location.
```

We should probably not use "current" twice, but it is too early for me to look at a thesaurus.

Firebase/Database/Api/FIRServerValue_unreleased.h
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * This must be an long or a double value. If the ref is not an integer or
```
schmidt-sebastian: Can you clarify what "This" is?

also s/the ref/the current value/

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,118 @@
...
+NSNumber *overflowAwareAddLong(long x, long y) {
+    long r = x + y;
+
+    // See "Hacker's Delight" 2-12: Overflow if both arguments have the opposite
+    // sign of the result
+    if (((x ^ r) & (y ^ r)) >= 0) {
+        return @(r);
+    }
+
+  return [NSNumber numberWithDouble:(double)x + y];
```
schmidt-sebastian: The formatting seems off here.

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,118 @@
...
+    case 'L': // unsigned long; fallthrough
+    case 'Q': // unsigned long long; fallthrough
+        // Only use ulong(long) if there isn't an overflow.
+        if (num.unsignedLongLongValue > LONG_MAX) {
+            break;
+        }
+    default:
+        return YES;
+    }
+    return NO;
```
schmidt-sebastian: I think this could be more readable. What about something like this:


``` 
switch (num.objCType[0]) {
    case 'f': // float; fallthrough
    case 'd': // double
        return NO;
    case 'L': // unsigned long; fallthrough
    case 'Q': // unsigned long long; fallthrough
        // Only use ulong(long) if there isn't an overflow.
        if (num.unsignedLongLongValue > LONG_MAX) {
             return NO;;
        }
    }
    return YES;
}
```

I am mostly interested in getting rid of the "break" statements as the code flow is a little confusing.

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,118 @@
 #import "FLeafNode.h"
 #import "FSnapshotUtilities.h"
 
+const NSString *kTimestamp = @"timestamp";
+const NSString *kIncrement = @"increment";
+
+BOOL longCanHold(NSNumber *num) {
```
schmidt-sebastian: Not a huge fan of the name, but it's probably ok :)  I'd suggest something like `canBeRepresentedAsLong`.

Optional.

Firebase/Database/Core/FServerValues.m
```diff
@@ -20,37 +20,118 @@
...
+    if (![existingLeaf.value isKindOfClass:NSNumber.class]) {
+        return delta;
+    }
+
+    NSNumber *existingNum = existingLeaf.value;
+    BOOL incrLong = longCanHold(delta);
+    BOOL baseLong = longCanHold(existingNum);
+
+    if (incrLong && baseLong) {
+      return overflowAwareAddLong(delta.longValue, existingNum.longValue);
```
schmidt-sebastian: I'm also not a huge fan of this method name and would suggest to inline.

Optional.

Firebase/Database/Core/FServerValues.m
schmidt-sebastian: It's probably fine to just add an assert.

inlined: Actually this is an expected case. For example, a null value is not a leaf node and we want to set the value to delta in this case.

Firebase/Database/Core/FServerValues.m
inlined: inlining also simplifies, so yay

PR #4331, opened 2019-11-15
Reduce flakes

paulb777:
Make travis_retry more likely to succeed for flakes like https://travis-ci.org/firebase/firebase-ios-sdk/jobs/612615462

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -261,6 +261,9 @@ case "$product-$method-$platform" in
     ;;
 
   Firestore-xcodebuild-*)
+    # Make travis_retry more likely to succeed.
+    rm -rf ~/Library/Developer/Xcode/DerivedData
+
     "${firestore_emulator}" start
     trap '"${firestore_emulator}" stop' ERR EXIT
 


scripts/build.sh
```diff
@@ -261,6 +261,9 @@ case "$product-$method-$platform" in
     ;;
 
   Firestore-xcodebuild-*)
+    # Make travis_retry more likely to succeed.
+    rm -rf ~/Library/Developer/Xcode/DerivedData
```
wilhuff: This is a really steep price to pay. Are you certain this actually helps?

Note that we used to have this globally but it actually prevented Firestore from succeeding when retrying because rebuilding everything takes ~20 minutes on Travis.

FWIW, we've been tackling flakes in the Firestore build and found a bug in the emulator that was causing quite a few of them. Since we haven't exhausted this track I'd rather not do this nuclear thing--unless it's known to actually reduce flakiness somehow.

paulb777:
Lately, the Firestore xcode builds have been the most common culprits of post-submit builds failing - especially the tvOS build.  

Since the linked failure shows tries two and three failing immediately with the same bad simulator state, it seems worth doing this, since there would be time for at least one more full run - and since removing DerivedData helped with flakes on the Firebase xcodebuild test.

However, if you're actively working on flakes, I'm fine to hold off.

Issue #4284, opened 2019-11-11
Adding FirebaseFirestore pod dependency adds minutes to build time


jmagman:
* Xcode version: 11.2
  * Firebase Component: FirebaseFirestore
  * Component version: 1.7.0

Dependency on FirebaseFirestore adds minutes to build time.

#### Steps to reproduce:

Create new Xcode project (I called it TestFirebase)
New Podfile:
```ruby
target 'TestFirebase' do
  pod 'FirebaseFirestore', '1.7.0'
end
```
```shell
$ pod install
$ /usr/bin/xcodebuild -workspace TestFirebase.xcworkspace -scheme TestFirebase
```

[Build TestFirebase_2019-11-11T14-17-13.txt](https://github.com/firebase/firebase-ios-sdk/files/3833502/Build.TestFirebase_2019-11-11T14-17-13.txt)

I ran with a build timing summary on my 2018 15" MacBook Pro:
```
Build Timing Summary

CompileC (1114 tasks) | 1642.840 seconds
...
```
(I'm guessing that 1642.840 seconds is divided over my 6 cores). 

#### Impact

Investigating this on behalf of Flutter's cloud_firestore plugin iOS users (though I can reproduce the slow build times without Flutter or cloud_firestore).  See context at https://github.com/flutter/flutter/issues/37582 (55 👍 ) and https://github.com/FirebaseExtended/flutterfire/issues/349 (46 👍 ).

jmagman:
\cc @collinjackson

wilhuff:
This is expected:

  * Firestore itself is 60k LOC
  * BoringSSL is 194k LOC
  * gRPC (core and C++) is 210k LOC
  * LevelDB is 16k LOC
  * Abseil in 40k LOC

Half a million lines of mostly C++ isn't going to compile quickly on a single host. ~400k of this comes from gRPC and its dependencies, so even if you reduced code actually in Firestore to zero you'd still have a huge build. 

At least when building for iOS directly (using Xcode) you generally pay this cost once, not on every build.

paulb777:
Some options to investigate:
- [cocoapods-packager](https://guides.cocoapods.org/plugins/using-pods-for-closed-source-libs.html)
- [cocoapods-binary](https://guides.cocoapods.org/plugins/pre-compiling-dependencies.html)
- [cocoapods-rome](https://github.com/CocoaPods/Rome)

[...]

PR #4341, opened 2019-11-18
Zip Builder V2 Plan

paulb777:
Can potentially address #4284.

Implementation occurring in https://github.com/firebase/firebase-ios-sdk/tree/pb-any-zip

#no-changelog

### DIFF
--- a/ZipBuilder/ZipBuilderV2Plan.md
+++ b/ZipBuilder/ZipBuilderV2Plan.md
@@ -1,0 +1,58 @@
+V2 will add an option to generate a zip distribution of binary frameworks from an arbitrary list
+of source and binary CocoaPods.
+
+## Introduction
+
+The current [Zip Builder](https://github.com/firebase/firebase-ios-sdk/tree/master/ZipBuilder)
+is Firebase specific. This is a proposal and initial plan
+to evolve the Zip Builder into a multi-purpose Apple binary framework creation tool.
+
+It would be useful to have a generic 
+zip builder for other Google SDKs and other open source projects. In addition,
+providing a generic ZipBuilder would enable Firebase users to generate binary
+frameworks for configurations outside the standard Zip and Carthage distributions
+released in the standard Firebase release process.
+
+With the extra flexibility, the Zip Builder will be useful for both SDK distributors
+to package binary distributions and app developers who want a customized binary
+distribution that maps exactly to their app's requirements and provides a clean
+build time speed up.
+
+
+## Background
+
+A more flexible zip builder would enable the following scenarios:
+
+  * Creating a zip distribution from an arbitrary set of pods
+  * Support different building different sets of platform slices
+  * Support building with different Xcode versions
+  * Build only the subspecs needed for use case
+  * [Existing solutions](https://github.com/firebase/firebase-ios-sdk/issues/4284#issuecomment-552677044)
+  are intermittently maintained and written in Ruby. A Swift-implementation is
+  more accessible and maintainable by the Apple community.
+  * The Swift implementation will be easier to add Swift Package Manager support for which
+  there will likely be a need since Swift Package Manager is even more source-centric than
+  CocoaPods.
+
+## Plan
+
+  1. Add `--zipPods {JSON file}` option. The JSON file contains a list of CocoaPods
+  along with an optional version specifier. If the version is not specified, a CocoaPods
+  install will determine the version - typically the latest, unless another pod requires
+  something lower.
+
+  1. Add `--minimum-ios-version {version}` option. Specify the minimum iOS version to build for. Default
+  is 8.0.
+
+  1. Add `--archs {archs list}` option. Default is "arm64, arm64e, armv7, i386, x86_64"
+
+Unlike the Firebase zip build which builds a two-level zip file with a configurable set of
+installation, when `--zipPods` is specified a single zip file of frameworks will be created.
+
+## Future Directions
+
+  * Binary support for Catalyst
+  * Other Apple platforms besides iOS
+  * Swift Package Manager
+  * Dynamic Framework support
+  * More option customization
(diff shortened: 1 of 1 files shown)


ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
+V2 will add an option to generate a zip distribution of binary frameworks from an arbitrary list
```
ryanwilson: There are a few different capitalizations throughout (and throughout existing docs, which should be changed eventually). In an effort to keep it consistent, what do you think about:

- `zip` referring to a file that has the `.zip` extension, `Zip` if at the start of a sentence
- the `Zip Builder` (capitalized with a space) for the *product* that will be building `zip` files
- `ZipBuilder` for referring to the code and tool itself

I'm open to other suggestions as well but we should try to keep it consistent throughout if possible.

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+of source and binary CocoaPods.
+
+## Introduction
+
+The current [Zip Builder](https://github.com/firebase/firebase-ios-sdk/tree/master/ZipBuilder)
+is Firebase specific. This is a proposal and initial plan
+to evolve the Zip Builder into a multi-purpose Apple binary framework creation tool.
+
+It would be useful to have a generic
```
ryanwilson: Spacing seems wrong here, more can fit on this line

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+## Background
+
+A more flexible zip builder would enable the following scenarios:
+
+  * Creating a zip distribution from an arbitrary set of pods
+  * Support different building different sets of platform slices
+  * Support building with different Xcode versions
+  * Build only the subspecs needed for use case
+  * [Existing solutions](https://github.com/firebase/firebase-ios-sdk/issues/4284#issuecomment-552677044)
+  are intermittently maintained and written in Ruby. A Swift-implementation is
```
ryanwilson: Nit: no `-`?

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+  CocoaPods.
+
+## Plan
+
+  1. Add `--zipPods {JSON file}` option. The JSON file contains a list of CocoaPods
+  along with an optional version specifier. If the version is not specified, a CocoaPods
+  install will determine the version - typically the latest, unless another pod requires
+  something lower.
+
+  1. Add `--minimum-ios-version {version}` option. Specify the minimum iOS version to build for. Default
```
ryanwilson: I know it's awful for `iOS` but we should stick with the same camel-case convention for args, this should likely be `--minimumIOSVersion`, or we should switch over to use `-` everywhere in arguments (I'm fine with that as well).

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+  is 8.0.
+
+  1. Add `--archs {archs list}` option. Default is "arm64, arm64e, armv7, i386, x86_64"
```
ryanwilson: In the future when we add Catalyst support, how will we handle `x86_64`? Would it do both simulator and Catalyst, or would it rely on an extra flag being passed? 

I think default to simulator and needing an extra flag for Catalyst seems reasonable but thought I'd check.

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+
+Unlike the Firebase zip build which builds a two-level zip file with a configurable set of
+installation, when `--zipPods` is specified a single zip file of frameworks will be created.
```
ryanwilson: Will we have `Resources` embedded in the `.framework` files like Carthage or pull them out into a common directory like the Firebase zip file?

ZipBuilder/ZipBuilderV2Plan.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Future Directions
+
+  * Binary support for Catalyst
+  * Other Apple platforms besides iOS
+  * Swift Package Manager
```
ryanwilson: Is the idea here to make a binary for Swift package manifests? Or something else?

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Done.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Done.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Done.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Done.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Hmm. We'll need to think more about how to evolve this towards adding Catalyst and eventually only building xcframeworks instead of frameworks when we can drop Xcode 10.

I'll clarify this in Future Directions.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Since the non-Firebase zip is completely new, it seems reasonable to simplify and stop moving the `Resources` around in the distribution.

ZipBuilder/ZipBuilderV2Plan.md
paulb777: Yes. Swift Package Manager will likely only increase the desire for SDK distributors and app developers to have pre-compiled frameworks.

ZipBuilder/ZipBuilderV2Plan.md
ryanwilson: SGTM - just wanted to clarify that.

Issue #4343, opened 2019-11-18
Fix FirebaseCore test failures with OCMock 3.5.0


paulb777:
From https://travis-ci.org/firebase/firebase-ios-sdk/jobs/613701568:

https://console.firebase.googl/Users/travis/build/firebase/firebase-ios-sdk/Example/Core/Tests/FIRComponentContainerTest.m:81: error:
-[FIRComponentContainerTest testComponentsPopulatedOnInit] : failed: caught "NSInternalInconsistencyException", "Mock object was not used in OCMStub/OCMExpect/OCMReject. Did you accidentally use a real object?"

morganchen12:
Caused by https://github.com/erikdoe/ocmock/commit/40d3e22f29c20d1bb54e8b4f79aab55225cc4b72

PR #4344, opened 2019-11-18
Lock OCMock to 3.4 while investigating test failures with 3.5.0

paulb777:
OCMock caused unit tests to fail in seven libraries.

This PR locks the dependency to 3.4.0 while we investigate.

#no-changelog


paulb777:
This PR should be reverted in pieces as solutions to #4343 #4345 #4347, #4348, #4349, #4350 and #4351 are implemented.

Issue #3951, opened 2019-09-29
Non-ASCII character in app name causes crash in +[NSBundle bundleWithIdentifier:]


burkaslarry:
I use  firebase firestore to develop  . In this morning , no crash

as night, it gives exception at [NSBundle bundleWithIdentifier:@“org.cocoapods.grpcpp”] as screenshot stated . I have no clue on my pod settings or code... Would you please tell me what to fix? 

Below is my screen shot : 

[![enter image description here][1]][1]

Below is my code  : 

        Auth.auth().signInAnonymously() { (result , error) in
            guard let authResult = result else {
                return
            }
            let user = authResult.user
            let isAnonymous = user.isAnonymous  // true
            if(isAnonymous){
                let uid = user.uid
                User.shared.userId = uid
                print("user id : \(User.shared.userId)")
               self.writeProgressInitial()

            }
            // Do any additional setup after loading the view.
        }

 

       func writeProgressInitial() {
            let db = Firestore.firestore()
            let userRef = db.collection("user").document(User.shared.userId)
            userRef.getDocument { (querySnapshot, err) in
            if let err = err {
                print("Error getting documents: \(err)")
            } else {
                if (querySnapshot?.get("progress") == nil) {
                       var  dataDict : Dictionary<String , Any> = [:]
                       dataDict["progress"]  = 0
                       userRef.setData(dataDict)
                    }else{
                        let dict = querySnapshot?.data()
                        let sssresult = dict!.filter{ $0.key == "progress" }.first?.value as? Int ?? 0
                        User.shared.progress = sssresult
                    }
                }
            }
        }

  [1]: https://i.stack.imgur.com/BMPA7.png

Personally, I have tested that even using Dispatch async to update the progress but it fails .

Below is my Installed dependencies  : 

    Installing BoringSSL-GRPC (0.0.3)
    Installing Cosmos (19.0.3)
    Installing Firebase (6.9.0)
    Installing FirebaseAnalytics (6.1.2)
    Installing FirebaseAuth (6.2.3)
    Installing FirebaseAuthInterop (1.0.0)
    Installing FirebaseCore (6.3.0)
    Installing FirebaseCoreDiagnostics (1.1.0)
    Installing FirebaseCoreDiagnosticsInterop (1.0.0)
    Installing FirebaseFirestore (1.5.1)
    Installing FirebaseInstanceID (4.2.5)
    Installing GTMSessionFetcher (1.2.2)
[...]

PR #4352, opened 2019-11-18
Embed root certificates into the binary on Apple platforms as well

var-const:
Currently, on iOS Firestore is loading `roots.pem` file from the gRPC bundle. The mechanism has proved to be brittle, so instead we can use the same approach as the CMake build does -- parse the file at build time and embed its contents into the binary.

One downside of this approach is that now we would have to periodically update the generated files for embedding manually. Judging from history, `roots.pem` is updated quite often in gRPC repo, and can be expected to be modified on each release.

The command to regenerate the certificates is something like:
```sh
mkdir build # From the root of the repo
cd build
cmake .. && make -j 8 generate_grpc_root_certificates
```

TESTED=with Fireeats.


Firestore/core/src/firebase/firestore/remote/grpc_root_certificates_generated.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
```
paulb777: 2019

var-const: Done, thanks for noticing.

var-const:
@wilhuff Gil, is it possible to exclude the generated files from the linter?

var-const:
Fixes #3951

scripts/binary_to_array.py
```diff
@@ -56,6 +56,13 @@
                         help="Output source file, defining the array data.")
 arg_parser.add_argument("--output_header",
                         help="Output header file, declaring the array data.")
+arg_parser.add_argument("--logical_directory",
```
var-const: This is my attempt to make sure the script stays generic and is complaint with our linter. Do you think it's overkill?

var-const: Done, thanks. I really like the idea of using `git rev-parse` here.

I have originally wanted to also exclude the file from linter and `style.sh` because checking it caused a noticeable slowdown. However, now that I think about it, the generated files should only be checked if they are modified, so this shouldn't affect the regular development; therefore, I did not exclude them.

PR #4353, opened 2019-11-18
Add fis heartbeat support

VinayGuthal:
add fis support for the heartbeat feature


FirebaseInstallations/Source/Tests/Unit/FIRInstallationsAPIServiceTests.m
```diff
@@ -50,6 +53,10 @@ - (void)setUp {
   self.service = [[FIRInstallationsAPIService alloc] initWithURLSession:self.mockURLSession
                                                                  APIKey:self.APIKey
                                                               projectID:self.projectID];
+  NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
+  GULHeartbeatDateStorage *dataStorage =
+      [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
+  [[NSFileManager defaultManager] removeItemAtURL:[dataStorage fileURL] error:nil];
```
maksymmalyhin: The tests may be performed in parallel, so using a real heartbeat storage can interfere with other such tests and lead to flakes. We should mock FIRHeartbeatInfo class methods instead

VinayGuthal: Yes makes sense. Updated

FirebaseInstallations/Source/Tests/Unit/FIRInstallationsAPIServiceTests.m
```diff
@@ -50,6 +53,9 @@ - (void)setUp {
   self.service = [[FIRInstallationsAPIService alloc] initWithURLSession:self.mockURLSession
                                                                  APIKey:self.APIKey
                                                               projectID:self.projectID];
+  id heartbeatMock = OCMClassMock([FIRHeartbeatInfo class]);
```
maksymmalyhin: The mock should be stored to a property and set to nil in tearDown

PR #4369, opened 2019-11-20
Fix fuzz build

var-const:
There's an unused variable that makes fuzz build fail due to `-WError`.


wilhuff:
Could you also fix the main build so that we see this failure directly?

PR #4373, opened 2019-11-21
Fix Carthage CD overriding the Zip build.

ryanwilson:
This separates the frameworks built for Carthage from the frameworks
built for the zip file.


ryanwilson:
Rebased onto master.

paulb777:
Merging this now to prevent conflicts and enable testing along with V2 development.

PR #4377, opened 2019-11-21
Fix completely out-of-source CMake builds

wilhuff:
Visual Studio 2017 runs creates its CMake output directory in a location
outside the git checkout, which causes the check for a current git
repository to fail.

### DIFF
--- a/scripts/binary_to_array.py
+++ b/scripts/binary_to_array.py
@@ -227,12 +227,14 @@ def _get_repo_root(filename):
 
   # CMake builds can be run out of source tree, so use the directory containing
   # a source file as the starting point for the calculation.
-  if not os.path.isdir(filename):
-    filename = os.path.dirname(filename)
+  if os.path.isdir(filename):
+    dir_in_repo = filename
+  else:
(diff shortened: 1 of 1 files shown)


scripts/binary_to_array.py
```diff
@@ -218,11 +218,30 @@ def source(namespaces, array_name, array_size_name, fileid, filename,
...
+def _get_repo_root(filename):
+  """Returns the root of the repository containing the given file.
+
+  Args:
+    filename: A source file or directory in the repository.
+  """
+
+  # CMake builds can be run out of source tree, so use the directory containing
+  # a source file as the starting point for the calculation.
+  if not os.path.isdir(filename):
```
var-const: Nit: I think it would be slightly more readable if this used a new intermediate variable (I would expect `filename` to be a "leaf" file, not a directory). Consider something like:
```python
dir_in_repo = filename if os.path.isdir(filename) else os.path.dirname(filename)
```

wilhuff: The one-liner doesn't fit in the allowed line length, so I split it across multiple lines

PR #4379, opened 2019-11-21
lastFetchTime's accessor does not reference the backing ivar. Silence…

dmandar:
… the corresponding CLANG warning. Internal cl/278914880


chliangGoogle:
Thanks for doing this Mandar!

PR #3925, opened 2019-09-25
Handle UISceneDelegate changes in Auth

renkelvin:

PR #4380, opened 2019-11-22
Handle UISceneDelegate changes in Auth

renkelvin:
Seems there is issue with the building system for https://github.com/firebase/firebase-ios-sdk/pull/3925
Started a new PR.

### DIFF
--- a/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
+++ b/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
@@ -31,6 +31,11 @@ typedef BOOL (*GULRealOpenURLSourceApplicationAnnotationIMP)(
 typedef BOOL (*GULRealOpenURLOptionsIMP)(
     id, SEL, GULApplication *, NSURL *, NSDictionary<NSString *, id> *);
 
+#if UISCENE_SUPPORTED
+API_AVAILABLE(ios(13.0), tvos(13.0))
+typedef void (*GULOpenURLContextsIMP)(id, SEL, UIScene *, NSSet<UIOpenURLContext *> *);
+#endif  // UISCENE_SUPPORTED
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wstrict-prototypes"
 typedef void (*GULRealHandleEventsForBackgroundURLSessionIMP)(
@@ -289,6 +294,20 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
     id<GULApplicationDelegate> originalDelegate =
         [GULAppDelegateSwizzler sharedApplication].delegate;
     [GULAppDelegateSwizzler proxyAppDelegate:originalDelegate];
+
+#if UISCENE_SUPPORTED
+    if (@available(iOS 13.0, tvOS 13.0, *)) {
+      if (![GULAppDelegateSwizzler isAppDelegateProxyEnabled]) {
+        return;
+      } else {
+        [[NSNotificationCenter defaultCenter]
+            addObserver:self
+               selector:@selector(handleSceneWillConnectToNotification:)
+                   name:UISceneWillConnectNotification
+                 object:nil];
+      }
+    }
+#endif  // UISCENE_SUPPORTED
   });
 }
 
@@ -682,6 +701,15 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
   }];
 }
 
+#if UISCENE_SUPPORTED
++ (void)handleSceneWillConnectToNotification:(NSNotification *)notification {
+  if (@available(iOS 13.0, tvOS 13.0, *)) {
+    UIScene *scene = (UIScene *)notification.object;
+    [GULAppDelegateSwizzler proxySceneDelegate:scene];
+  }
+}
+#endif  // UISCENE_SUPPORTED
+
 // The methods below are donor methods which are added to the dynamic subclass of the App Delegate.
 // They are called within the scope of the real App Delegate so |self| does not refer to the
 // GULAppDelegateSwizzler instance but the real App Delegate instance.
@@ -760,6 +788,40 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
 }
 
 #endif  // TARGET_OS_IOS
+
+#pragma mark - [Donor Methods] UISceneDelegate URL handler
+
+#if UISCENE_SUPPORTED
+- (void)scene:(UIScene *)scene
+    openURLContexts:(NSSet<UIOpenURLContext *> *)URLContexts API_AVAILABLE(ios(13.0), tvos(13.0)) {
+  if (@available(iOS 13.0, tvOS 13.0, *)) {
+    SEL methodSelector = @selector(scene:openURLContexts:);
+    // Call the real implementation if the real App Delegate has any.
+    NSValue *openURLContextsIMPPointer =
+        [GULAppDelegateSwizzler originalImplementationForSelector:methodSelector object:self];
+    GULOpenURLContextsIMP openURLContextsIMP = [openURLContextsIMPPointer pointerValue];
+
+    // This is needed for the library to be warning free on iOS versions < 9.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability"
+    [GULAppDelegateSwizzler
+        notifyInterceptorsWithMethodSelector:methodSelector
+                                    callback:^(id<GULApplicationDelegate> interceptor) {
+                                      if ([interceptor
+                                              conformsToProtocol:@protocol(UISceneDelegate)]) {
+                                        id<UISceneDelegate> sceneInterceptor =
+                                            (id<UISceneDelegate>)interceptor;
+                                        [sceneInterceptor scene:scene openURLContexts:URLContexts];
+                                      }
+                                    }];
+#pragma clang diagnostic pop
+
+    if (openURLContextsIMP) {
+      openURLContextsIMP(self, methodSelector, scene, URLContexts);
+    }
+  }
+}
+#endif  // UISCENE_SUPPORTED
 
 #pragma mark - [Donor Methods] Network overridden handler methods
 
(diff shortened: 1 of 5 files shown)


GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -682,6 +701,15 @@ + (void)notifyInterceptorsWithMethodSelector:(SEL)methodSelector
   }];
 }
 
+#if UISCENE_SUPPORTED
++ (void)handleSceneWillConnectToNotification:(NSNotification *)notification {
+  if (@available(iOS 13.0, tvOS 13.0, *)) {
+    UIScene *scene = (UIScene *)notification.object;
```
maksymmalyhin: I think it will be safer to check the class of the `notification.object`. Even if presence and the class of the object is guaranteed in this iOS version, it may protect us from API changes in the future.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -761,6 +789,40 @@ - (BOOL)application:(GULApplication *)application
...
+- (void)scene:(UIScene *)scene
+    openURLContexts:(NSSet<UIOpenURLContext *> *)URLContexts API_AVAILABLE(ios(13.0), tvos(13.0)) {
+  if (@available(iOS 13.0, tvOS 13.0, *)) {
+    SEL methodSelector = @selector(scene:openURLContexts:);
+    // Call the real implementation if the real App Delegate has any.
+    NSValue *openURLContextsIMPPointer =
+        [GULAppDelegateSwizzler originalImplementationForSelector:methodSelector object:self];
+    GULOpenURLContextsIMP openURLContextsIMP = [openURLContextsIMPPointer pointerValue];
+
+    // This is needed for the library to be warning free on iOS versions < 9.
```
maksymmalyhin: The comment does not look up to date. I guess, `pragma clang diagnostic ignored "-Wunguarded-availability"` may not be needed here at all because of `#if UISCENE_SUPPORTED`

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -1002,6 +1064,89 @@ + (void)proxyAppDelegate:(id<GULApplicationDelegate>)appDelegate {
   }
 }
 
+#if UISCENE_SUPPORTED
++ (void)proxySceneDelegate:(UIScene *)scene {
+  Class realClass = [scene.delegate class];
+
+  // Skip proxying if the class has a prefix of kGULAppDelegatePrefix, which means it has been
+  // proxied before.
+  if ([NSStringFromClass(realClass) hasPrefix:kGULAppDelegatePrefix]) {
```
maksymmalyhin: nit: `NSStringFromClass(realClass)` is used multiple times. Can we assign the func result to a variable?

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -1002,6 +1064,89 @@ + (void)proxyAppDelegate:(id<GULApplicationDelegate>)appDelegate {
...
+                [NSString stringWithFormat:@"I-SWZ%06ld",
+                                           (long)kGULSwizzlerMessageCodeAppDelegateSwizzling007],
+                @"Cannot create subclass of App Delegate, because the created subclass is not the "
+                @"same size. %@",
+                NSStringFromClass(realClass));
+    NSAssert(NO, @"Classes must be the same size to swizzle isa");
+    return;
+  }
+
+  // Make the newly created class to be the subclass of the real App Delegate class.
```
maksymmalyhin: nit: "App Delegate" -> "Scene Delegate"

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -1002,6 +1064,89 @@ + (void)proxyAppDelegate:(id<GULApplicationDelegate>)appDelegate {
...
+
+  NSString *classNameWithPrefix =
+      [kGULAppDelegatePrefix stringByAppendingString:NSStringFromClass(realClass)];
+  NSString *newClassName =
+      [NSString stringWithFormat:@"%@-%@", classNameWithPrefix, [NSUUID UUID].UUIDString];
+
+  if (NSClassFromString(newClassName)) {
+    GULLogError(kGULLoggerSwizzler, NO,
+                [NSString stringWithFormat:@"I-SWZ%06ld",
+                                           (long)kGULSwizzlerMessageCodeAppDelegateSwizzling005],
```
maksymmalyhin: nit: Not sure if we should reuse the same message code. I think we should create separate message codes for this method. Also the message should be updated to replace "App Delegate" by "Scene Delegate"

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -1002,6 +1064,89 @@ + (void)proxyAppDelegate:(id<GULApplicationDelegate>)appDelegate {
...
+  SEL openURLContextsSEL = @selector(scene:openURLContexts:);
+  [self proxyDestinationSelector:openURLContextsSEL
+      implementationsFromSourceSelector:openURLContextsSEL
+                              fromClass:[GULAppDelegateSwizzler class]
+                                toClass:sceneDelegateSubClass
+                              realClass:realClass
+       storeDestinationImplementationTo:realImplementationsBySelector];
+
+  // Store original implementations to a fake property of the original delegate.
+  objc_setAssociatedObject(scene.delegate, &kGULRealIMPBySelectorKey,
```
maksymmalyhin: nit: It think readability of the class may be slightly improved if we create a "set" counterpart for `originalImplementationForSelector: object:` method and use it everywhere instead of referring to `kGULRealIMPBySelectorKey` directly. It may be a subject of a followup PR.

GoogleUtilities/Example/Tests/Swizzler/GULAppDelegateSwizzlerTest.m
```diff
@@ -1301,4 +1312,60 @@ - (void)testAppDelegateIsProxiedIncludingAPNSMethodsWhenEnabled {
...
+    Class realSceneDelegateClassBefore = [realSceneDelegate class];
+
+    // Proxy the scene delegate for the 2nd time.
+    [GULAppDelegateSwizzler proxySceneDelegate:mockSharedScene];
+
+    // Make sure that the class isn't changed.
+    XCTAssertEqualObjects([realSceneDelegate class], realSceneDelegateClassBefore);
+  }
+}
+
```
maksymmalyhin: Could you also add tests to make sure that interseptors are notified when the swizzled methods are called (similar to [this one](https://github.com/firebase/firebase-ios-sdk/blob/b05d8d186ee41cc5ec0f83967b865db10e7f1f41/GoogleUtilities/Example/Tests/Swizzler/GULAppDelegateSwizzlerTest.m#L1042), please.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
renkelvin: Yeah, you're right.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
renkelvin: Done.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
renkelvin: Good idea, done.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
renkelvin: Yeah, dedicated message codes should be better. Can I add it directly? or is there any review process for GULLogger?

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
renkelvin: I feel like using `kGULRealIMPBySelectorKey` may remind people that it's a runtime associated object and people will be more careful. While you suggestion totally makes sense. We can discuss more in a followup pr.

GoogleUtilities/Example/Tests/Swizzler/GULAppDelegateSwizzlerTest.m
renkelvin: Done.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
maksymmalyhin: Feel free to add them [here](https://github.com/firebase/firebase-ios-sdk/blob/080397209ec36f20d08f8fc931115bcd8cf82886/GoogleUtilities/Common/GULLoggerCodes.h#L19). A meaningful name is the recent recommendation (like `kGULSwizzlerMessageCodeAppDelegateSwizzlingInvalidAppDelegate` in the enum).

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -682,6 +701,17 @@ + (void)notifyInterceptorsWithMethodSelector:(SEL)methodSelector
   }];
 }
 
+#if UISCENE_SUPPORTED
++ (void)handleSceneWillConnectToNotification:(NSNotification *)notification {
+  if (@available(iOS 13.0, tvOS 13.0, *)) {
+    if ([notification.object isKindOfClass:[UIScene class]]) {
+      UIScene *scene = (UIScene *)notification.object;
+      [GULAppDelegateSwizzler proxySceneDelegate:scene];
```
maksymmalyhin: nit: I usually prefer adding suffix like `IfNeeded` to the method names if the action may or may not happen, e.g. `proxySceneDelegateIfNeeded:`. It makes the code a bit easier to read without looking at the method implementation. Though it's up to you, feel free to keep it as is.

renkelvin: Yeah, suffix added.

PR #4383, opened 2019-11-22
Fix the Linux build

wilhuff:
Reshuffling our dependencies in #4279 broke the build; apparently the
Linux linker is less forgiving than the one on macOS.


wilhuff:
(note the diffbase)

PR #4386, opened 2019-11-22
Extract AppDataDir from LevelDbPersistence

wilhuff:
This helps consolidate platform-specific behavior in leaf libraries in the build, making it easier to integrate into the blaze build in google3.


Firestore/core/src/firebase/firestore/util/filesystem.h
```diff
@@ -61,6 +61,25 @@ Status RecursivelyCreateDir(const Path& path);
  */
 Status RecursivelyDelete(const Path& path);
 
+/**
+ * Returns a system-defined best directory in which to create application data.
+ * Values vary wildly across platforms. They include:
```
var-const: > Values vary wildly

Tongue twister! Consider also `Values vary wildly; values vary very wildly`.

var-const:
New commit LGTM as well.

PR #4389, opened 2019-11-22
Update zip builder to use latest CocoaPods version

paulb777:
Adding `use_modular_headers!` seems to generate the same module map files as 1.5.3 except for diffs like:

```
diff -r 153-Pods/Target Support Files/nanopb/nanopb.modulemap Pods/Target Support Files/nanopb/nanopb.modulemap
1c1
< framework module nanopb {
---
> module nanopb {
```

I'm not sure the `framework` is needed?

If you agree, next step is to merge to master and do a test build and kokoro integration test run.

This should also dramatically speed up the build by using the CDN.

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -125,7 +125,7 @@ public enum CocoaPodUtils {
     }
 
     // Run pod install on the directory that contains the Podfile and blank Xcode project.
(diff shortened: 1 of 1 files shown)


paulb777:
This is wrong. System frameworks only show up in `OTHER_LDFLAGS` of `Pods/Target Support Files/Pods-FrameworkMaker/Pods-FrameworkMaker.release.xcconfig`.

A better approach would be to call `pod install` for each framework individually, step through `OTHER_LDFLAGS` excluding anything that is in a dependency.

PR #4390, opened 2019-11-25
Fix several small linter issues

var-const:
These fixes allow Firestore to build on Linux with IWYU checks:
* include some headers that are used, mostly `absl/memory/memory.h`;
* add some virtual destructors;
* fix one cyclic dependency;
* fix one potential case of ODR violation.

### DIFF
--- a/Firestore/core/src/firebase/firestore/util/log_stdio.cc
+++ b/Firestore/core/src/firebase/firestore/util/log_stdio.cc
@@ -25,7 +25,11 @@ namespace firebase {
 namespace firestore {
 namespace util {
 
+namespace {
+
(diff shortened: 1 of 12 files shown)


Firestore/core/src/firebase/firestore/util/log_stdio.cc
```diff
@@ -25,8 +25,12 @@ namespace firebase {
 namespace firestore {
 namespace util {
 
+namespace {
+
 LogLevel g_log_level = kLogLevelWarning;
```
var-const: According to Asan, this was somehow violating ODR, though I'm not sure what symbol it was conflicting with.

PR #4396, opened 2019-11-25
Fix TSAN test failures under CMake

wilhuff:


CMakeLists.txt
```diff
@@ -54,14 +54,15 @@ enable_language(CXX)
...
 list(INSERT CMAKE_MODULE_PATH 0 ${PROJECT_SOURCE_DIR}/cmake)
-include(cc_rules)
 include(compiler_id)
 include(external_rules)
 include(podspec_rules)
 include(archive_options)
 include(sanitizer_options)
 include(fuzzing_options)
 
+include(cc_rules)
```
var-const: Perhaps add a comment to note that `cc_rules` has to be included after `compiler_id`?

cmake/FindASANDylib.cmake
```diff
@@ -14,8 +14,21 @@
 
 # Finds `libclang_rt.asan_osx_dynamic.dylib` on Apple platform, it is typically
 # under ${CMAKE_CXX_COMPILER}/../lib.
```
var-const: Can you expand the comment to explain why finding the `dylib` file is necessary?

CMakeLists.txt
wilhuff: Done. And for consistency moved all the rules down here, clearly separating options processing from rule definition.

cmake/FindASANDylib.cmake
wilhuff: Done.

PR #4398, opened 2019-11-26
Fix compilation with older versions of Visual Studio

var-const:
Looks like Visual Studio 2015 has trouble with `constexpr` arrays:
```cpp
constexpr char Foo[] = "abc";
```
but not with `constexpr` pointers to `char`:
```cpp
constexpr const char* Foo = "abc";
```

### DIFF
--- a/Firestore/core/src/firebase/firestore/api/settings.h
+++ b/Firestore/core/src/firebase/firestore/api/settings.h
@@ -32,8 +32,8 @@ namespace api {
  */
 class Settings {
  public:
-  // Note: a constexpr array of char (`char[]`) doesn't work with older versions
-  // of Visual Studio.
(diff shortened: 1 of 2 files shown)


Firestore/core/src/firebase/firestore/api/settings.h
```diff
@@ -32,6 +32,8 @@ namespace api {
  */
 class Settings {
  public:
+  // Note: a constexpr array of char (`char[]`) doesn't work with older versions
```
wilhuff: You might want to make this more specific:

```
// ... doesn't work with Visual Studio 2015.
```

This makes it easier to know what can be unfixed when dropping support for older platforms

var-const: Good point, done.

PR #4404, opened 2019-11-26
Build Zip with modern CocoaPods

paulb777:
Instead of using the no-longer existent pod specific OTHER_LD_FLAGS in the all Pods build, build each pod separately to find the dependent libraries and frameworks.

Other changes:
- Manage minimumIOSVersion via an option instead of a pod specifier
- Change pod utilities functions to use `PodInfo` instead of the Firebase specific `CocoaPod`
- Use the CocoaPods CDN for a significant speed increase
- Refactor `pod install` setup code to CocoaPodUtils.podInstallPrepare

Testing:
Verified the generated zip matched the 6.13.0 zip with the following exception that libraries and frameworks are no longer generated in the modulemap if they're already in a  transitive dependency pod.  Also `TensorFlowLiteC.framework/Modules/module.modulemap` now includes `link "c++"` which looks like fixes a bug in the old zip generation.


ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -224,6 +204,13 @@ struct ZipBuilder {
     let carthageCoreDiagnostics: URL = createCarthageCoreDiagnostics(fromPods: installedPods,
                                                                      inProjectDir: projectDir)
 
+    do {
+      let coreDiagnosticModuleMap = frameworks["FirebaseCoreDiagnostics"]![0].appendingPathComponent("Modules")
```
ryanwilson: Instead of force unwrapping (which would likely give a not-very-helpful error) could we `guard let ` and handle the optional case? Also `first` is preferred over `[0]` since it'll fail gracefully.

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -180,8 +156,10 @@ struct ZipBuilder {
 
     // Break the `podsToInstall` into a variable since it's helpful when debugging builds to just
     // install a subset of pods, like the following line:
-//    let podsToInstall: [CocoaPod] = [.core, .analytics, .storage]
-    let podsToInstall = CocoaPod.allCases
+    //    let podsToInstall: [String] = ["", "Core", "Analytics", "Storage"]
+    let podsToInstall = CocoaPod.allCases.map { CocoaPodUtils.PodInfo(name: $0.rawValue,
+                                                                      version: "",
+                                                                      installedLocation: URL(fileURLWithPath: "dummy")) }
```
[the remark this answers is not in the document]
ryanwilson: Also, don't we want the actual pod name instead of the subspec name for `podsToInstall` now?

ZipBuilder/Sources/ZipBuilder/CocoaPod.swift
```diff
@@ -59,27 +60,17 @@ public enum CocoaPod: String, CaseIterable {
...
+      return "Firebase\(pod)"
     }
+    return pod
   }
 
   /// Describes the dependency on other frameworks for the README file.
-  public func readmeHeader() -> String {
-    var header = "## \(rawValue)"
-    if !(self == .analytics || self == .googleSignIn) {
+  public static func readmeHeader(pod: String) -> String {
```
ryanwilson: Nit: `podName` may be more appropriate here, was reading down below and assumed the type passed in wasn't a String but `PodInfo` (same with the function below)

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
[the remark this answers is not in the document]
paulb777: Made a VersionedPod struct to distinguish between input and output PodInfo

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -21,16 +21,28 @@ import Foundation
...
+    let name: String
 
     /// The version of the pod.
-    var version: String
+    let version: String
+  }
+
+  /// Information associated with an installed pod.
+  public struct PodInfo {
+    let versionedPod: VersionedPod
```
ryanwilson: This doesn't need to be changed at all but I think is a great time to mention a new feature from Swift 5.1 that would enable something cool.. you could essentially expose the properties of `VersionedPod` on `PodInfo` with autocomplete and everything with `@dynamicMemberLookup` (Swift 4.2) and `KeyPath`s (5.1).

```
@dynamicMemberLookup
public struct PodInfo {
  let versionedPod: VersionedPod
  
  subscript<T>(dynamicMember keyPath: KeyPath<VersionedPod, T>) -> T {
        return versionedPod[keyPath: keyPath]
    }
}
```

Would let you write `somePodInfo.name` and `somePodInfo.version` natively :) 

See the "Combining key paths with @dynamicMemberLookup" from this blog post for more info:
https://www.avanderlee.com/swift/dynamic-member-lookup/

I remember reading about it but couldn't think of many good use cases but this is a perfect one! Requires Swift 5.1 so we can't use it at the moment though :P

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -21,16 +21,28 @@ import Foundation
...
 
     /// The location of the pod on disk.
     var installedLocation: URL
+
+    func name() -> String {
```
ryanwilson: Nit: It'd be nice to have these as properties instead of functions with `lazy`:

```
lazy var name: String { return self.versionedPod.name }
lazy var version: String { return self.versionedPod.version }
```

Essentially identical but no need for the `()`

ZipBuilder/Sources/ZipBuilder/ModuleMapBuilder.swift
```diff
@@ -21,14 +21,16 @@ import Foundation
 struct ModuleMapBuilder {
   /// Information associated with a framework
   private class FrameworkInfo {
-    let isOpenSource: Bool
-    let podInfo: CocoaPodUtils.PodInfo
+    let isSourcePod: Bool
+    let installedLocation: URL
+    let versionedPod: CocoaPodUtils.VersionedPod
```
ryanwilson: Isn't this represented in a `PodInfo` type?

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -21,16 +21,28 @@ import Foundation
...
-  /// Information associated with an installed pod.
-  public struct PodInfo {
+  public struct VersionedPod {
     /// Public name of the pod.
-    var name: String
+    let name: String
 
     /// The version of the pod.
-    var version: String
+    let version: String
```
ryanwilson: Sorry should have caught this before - do we want to make version optional now since it's not always required when requesting a pod? I'm open to either but wanted to check with you.

paulb777: Done. nil is more Swifty than an empty string.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
paulb777: But still an improvement to add the property :)

ZipBuilder/Sources/ZipBuilder/ModuleMapBuilder.swift
ryanwilson: Got it... I was thinking that the `PodInfo` contains both `installedLocation` and the `VersionedPod` so I wasn't sure why we were using the types separately. Thanks!

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -160,8 +160,8 @@ struct ZipBuilder {
 
     // Break the `inputPods` into a variable since it's helpful when debugging builds to just
     // install a subset of pods, like the following line:
-    //    let inputPods: [String] = ["", "Core", "Analytics", "Storage"]
-    let inputPods = CocoaPod.allCases.map { $0.rawValue }
+    let inputPods: [String] = ["", "Core", "Analytics", "Storage"]
+    //let inputPods = CocoaPod.allCases.map { $0.rawValue }
```
ryanwilson: Whoops, probably want to take this back :)

paulb777: Done :)

PR #4405, opened 2019-11-27
Add archs option to ZipBuilder

paulb777:
Add an option to specify the architectures for the slices in the framework binaries.

#no-changelog

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/LaunchArgs.swift
+++ b/ZipBuilder/Sources/ZipBuilder/LaunchArgs.swift
@@ -87,7 +87,7 @@ struct LaunchArgs {
   }
 
   /// The list of architectures to build for.
-  let archs: [String]
+  let archs: [Architecture]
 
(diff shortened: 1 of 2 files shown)


ZipBuilder/Sources/ZipBuilder/LaunchArgs.swift
```diff
@@ -82,6 +86,9 @@ struct LaunchArgs {
     }
   }
 
+  /// The list of architectures to build for.
+  let archs: [String]
```
ryanwilson: It'd probably be nicer to store this directly as `[Architecture]`, do the unwrapping here (with a failable initializer: `guard let arch = Architecture(rawValue: someString) else { /* fail */ }` then we don't have to force unwrap at the usage point.

paulb777: Done.

PR #4409, opened 2019-11-27
Support concurrent Zip Builder runs

paulb777:
Use a unique temp folder per run so that multiple Zip Builder's can coexist on the same build/dev machine.

Add -keepBuildArtifacts option to keep the temp folder around. By default it will be deleted after a successful run.

Also integrate the cache into the temp folder since it is rebuilt every run anyway.

#no-changelog

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
@@ -102,7 +102,7 @@ public extension FileManager {
   }
 
   // Enable a single unique temporary workspace per execution.
-  static var uniq: String?
+  static var unique: String?
 
   /// Returns a deterministic path of a temporary directory for the given name. Note: This does
   /// *not* create the directory if it doesn't exist, merely generates the name for creation.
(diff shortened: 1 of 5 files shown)


ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
```diff
@@ -107,6 +101,9 @@ public extension FileManager {
     }
   }
 
+  // Enable a single unique temporary workspace per execution.
+  static var uniq: String?
```
ryanwilson: You can actually instantiate this right here and use `let` - `static let unique: String = UUID().uuidString`

Also, I think having the extra `ue` is worth it from a readability standpoint

ZipBuilder/Sources/ZipBuilder/main.swift
```diff
@@ -16,14 +16,6 @@
 
 import Foundation
 
-// Delete the cache directory, if it exists.
```
ryanwilson: Hmm, considering a Firebase zip build can be 1GB and Carthage another 1GB, should we at least add an option to wipe the cache beforehand? It'd be pretty easy to get out of control for disk space here.

ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
paulb777: "Static properties may only be declared on a type". Other comments will be addressed in next push.

ZipBuilder/Sources/ZipBuilder/main.swift
paulb777: Removing the working temp directory by default and keeping only with -keepBuildArtifacts true

ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
ryanwilson: Hmm, I had this in a Playground and it worked as expected:

```
extension FileManager {
  static let unique: String = UUID().uuidString
}

print(FileManager.unique)
print(FileManager.unique)
print(FileManager.unique)
```

paulb777: Done.

I misunderstood and thought you were asking to initialize inside the func. This is much cleaner. Thanks!

PR #4410, opened 2019-11-27
add heartbeat support for firestore

VinayGuthal:
Add a `GrpcMetadataProvider` class which provides both the user agent string and the heart beat information which is then used by grpc_connection to add their corresponding values to the header.


Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2018 Google
```
wilhuff: 2019. Here and throughout.

Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.mm
```diff
@@ -0,0 +1,38 @@
...
+#import <Foundation/Foundation.h>
+#include <string>
+
+NSString* const kFirebaseFirestoreHeartbeatTag = @"fire-fst";
+
+namespace firebase {
+namespace firestore {
+namespace remote {
+std::string GrpcMetadataProvider::getHeartbeatCode() {
+  return std::string(
```
wilhuff: Use util::MakeString() from Firestore/core/src/firebase/firestore/util/string_apple.h. In addition to being shorter, we've benchmarked it as being significantly less expensive than `[NSString UTF8String]`.

Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.mm
```diff
@@ -0,0 +1,38 @@
+/*
```
wilhuff: This should be `firebase_client_grpc_metadata_provider_apple.mm` to match `FirebaseClientGrpcMetadataProvider` on Android. The `_apple` in the filename allows us to have alternative implementations on non-apple platforms.

Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.h
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Class providing the userAgent and the hearbeat information.
+ */
+class GrpcMetadataProvider {
+ public:
+  /**
+   * Called by grpc_connection.cc. Provides the heartbeatcode corresponding to
+   * firestore.
+   */
+  static std::string getHeartbeatCode();
```
wilhuff: This should be a virtual function that adds metadata to a `grpc::ClientContext *` as required.

`virtual void UpdateMetadata(grpc::ClientContext* context) = 0`.

Once you add a virtual function you'll also need a virtual destructor, above:

`virtual ~GrpcMetadataProvider() = default;`

Firestore/core/src/firebase/firestore/remote/grpc_connection.cc
```diff
@@ -50,6 +51,8 @@ namespace {
 
 const char* const kXGoogAPIClientHeader = "x-goog-api-client";
 const char* const kGoogleCloudResourcePrefix = "google-cloud-resource-prefix";
+const char* const kFirebaseFirestoreHeartbeatKey = "X-firebase-client-log-type";
```
wilhuff: These keys should be hidden in the metadata provider specific to the firebase client, as we did on Android.

Firestore/core/src/firebase/firestore/remote/grpc_connection.cc
```diff
@@ -170,6 +171,10 @@ std::unique_ptr<grpc::ClientContext> GrpcConnection::CreateContext(
   context->AddMetadata(kGoogleCloudResourcePrefix,
                        StringFormat("projects/%s/databases/%s",
                                     db_id.project_id(), db_id.database_id()));
+
+  GrpcMetadataProvider * provider = GrpcMetadataProvider.create();
```
VinayGuthal: this complains telling its an abstract class

Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+class GrpcMetadataProvider {
+ public:
+  /** Creates a platform-specific GrpcMetadataProvider. */
+  static std::unique_ptr<GrpcMetadataProvider> Create();
+  /**
+   * Called by grpc_connection.cc. Updates the metadata with the
+   * heartbeat and the useragent header.
+   */
+  virtual void UpdateMetadata(
```
wilhuff: Making this an abstract method requires there to be a separate no-op implementation. If you make this just an empty method then you don't need a separate implementation:

```c++
virtual void UpdateMetadata(grpc::ClientContext* context) {
  (void)context;
}
```

Firestore/core/src/firebase/firestore/remote/grpc_metadata_provider.h
```diff
@@ -0,0 +1,47 @@
...
+#include "grpcpp/client_context.h"
+
+namespace firebase {
+namespace firestore {
+namespace remote {
+
+/**
+ * Class providing the userAgent and the hearbeat information.
+ */
+class GrpcMetadataProvider {
```
wilhuff: You'll need a .cc file that provides an implementation of `GrpcMetadataProvider::Create` for non-Apple platforms.
```c++
#if !__APPLE__

std::unique_ptr<GrpcMetadataProvider> GrpcMetadataProvider::Create() {
  return absl::make_unique<GrpcMetadataProvider>();
}

#endif  // !__APPLE__
```

ryanwilson:
Closing since there hasn't been progress in a very long time.

PR #4413, opened 2019-11-28
[1/4] Split Query and Target, and make LocalStore/RemoteStore work with Target.

wu-hui:
1. Added `Target` and `Query.ToTarget()`.
2. Make LocalStore and RemoteStore work with Target instead of Query, except `LocalStore.ExecuteQuery`.
3. Remove `resumeToken` from `QueryView`. This follows what Android does, and it's not used anywhere, this is not directly related to the change, but I would have to account for this in my change. Better delete it to say consistent with other SDKs.


Firestore/core/src/firebase/firestore/core/query.h
```diff
@@ -239,11 +245,12 @@ class Query {
...
   mutable OrderByList memoized_order_bys_;
 
-  int32_t limit_ = kNoLimit;
+  int32_t limit_ = Target::kNoLimit;
   std::shared_ptr<Bound> start_at_;
   std::shared_ptr<Bound> end_at_;
 
-  mutable std::string canonical_id_;
+  // The corresponding Target of this Query instance.
+  mutable std::shared_ptr<const Target> memoized_target;
```
wilhuff: Given that our only known divergence between Query and Target is going to be an additional property, it seems wasteful to duplicate the query properties (the path, filter list, etc) in both of these classes.

You could avoid having to keep this separate pointer by making Target a direct member and fully delegating all common properties to it.

Firestore/core/src/firebase/firestore/core/query.h
```diff
@@ -39,13 +40,12 @@ namespace core {
 using CollectionGroupId = std::shared_ptr<const std::string>;
 
 /**
- * Represents the internal structure of a Firestore Query. Query instances are
- * immutable.
+ * Encapsulates all the query attributes we support in the SDK. It can be run
```
wilhuff: One feature that distinguishes Queries vs Targets that isn't well captured in this comment is that `Query` represents the user-facing concept of the query where `Target` represents the query we might execute on the server. The comment says this indirectly now (in its discussion of the LocalStore vs RemoteStore) but making this explicit would make this clearer.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -85,7 +85,7 @@
 using firebase::firestore::remote::ExistenceFilter;
 using firebase::firestore::remote::DocumentWatchChange;
 using firebase::firestore::remote::ExistenceFilterWatchChange;
-using firebase::firestore::remote::InvalidQuery;
+using firebase::firestore::remote::InvalidTarget;
```
wilhuff: This appears unused now?

Firestore/core/src/firebase/firestore/local/query_data.h
```diff
@@ -92,14 +93,14 @@ class QueryData {
    */
   static QueryData Invalid();
 
-  /** The query being listened to. */
-  const core::Query& query() const {
-    return query_;
+  /** The target being listened to. */
+  const std::shared_ptr<const core::Target> target() const {
```
wilhuff: As noted elsewhere, this would be better off as a `const core::Target&`, hiding the `std::shared_ptr` detail (if we even need that).

Firestore/core/src/firebase/firestore/local/local_store.cc
```diff
@@ -429,17 +434,20 @@ QueryData LocalStore::AllocateQuery(Query query) {
...
   target_ids[target_id] = query_data;
   return query_data;
 }
+void LocalStore::ReleaseQuery(const core::Query& query) {
+  ReleaseTarget(*query.ToTarget());
+}
 
-void LocalStore::ReleaseQuery(const Query& query) {
+void LocalStore::ReleaseTarget(const Target& target) {
   persistence_->Run("Release query", [&] {
```
wilhuff: Also change these strings: "Release target" and "nonexistent target" (below).

Firestore/core/src/firebase/firestore/local/local_store.cc
```diff
@@ -411,13 +412,17 @@ BatchId LocalStore::GetHighestUnacknowledgedBatchId() {
     return mutation_queue_->GetHighestUnacknowledgedBatchId();
   });
 }
-
 QueryData LocalStore::AllocateQuery(Query query) {
+  return AllocateTarget(*query.ToTarget());
+}
+
+QueryData LocalStore::AllocateTarget(Target target) {
   QueryData query_data = persistence_->Run("Allocate query", [&] {
```
wilhuff: "Allocate target"

Firestore/core/src/firebase/firestore/local/local_serializer.cc
```diff
@@ -254,15 +255,15 @@ QueryData LocalSerializer::DecodeQueryData(
   SnapshotVersion version =
       rpc_serializer_.DecodeVersion(reader, proto.snapshot_version);
   ByteString resume_token(proto.resume_token);
-  Query query = InvalidQuery();
+  Target target = InvalidTarget();
```
wilhuff: All of our other types in here are just default constructible. This `InvalidTarget` thing doesn't seem that useful. Let's remove it and add a default constructor to `Target`.

Firestore/core/src/firebase/firestore/local/leveldb_query_cache.cc
```diff
@@ -129,18 +129,18 @@ void LevelDbQueryCache::RemoveTarget(const QueryData& query_data) {
...
   db_->current_transaction()->Delete(index_key);
 
   metadata_->target_count--;
   SaveMetadata();
 }
 
-absl::optional<QueryData> LevelDbQueryCache::GetTarget(const Query& query) {
+absl::optional<QueryData> LevelDbQueryCache::GetTarget(const Target& target) {
   // Scan the query-target index starting with a prefix starting with the given
   // query's canonical_id. Note that this is a scan rather than a get because
```
wilhuff: target's canonical_id

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -936,42 +938,42 @@ google_firestore_v1_Target_QueryTarget Serializer::EncodeQueryTarget(
...
+Target Serializer::DecodeQueryTarget(
     nanopb::Reader* reader,
     const google_firestore_v1_Target_QueryTarget& proto) const {
   // The QueryTarget oneof only has a single valid value.
   if (proto.which_query_type !=
       google_firestore_v1_Target_QueryTarget_structured_query_tag) {
     reader->Fail(
         StringFormat("Unknown query_type: %s", proto.which_query_type));
-    return InvalidQuery();
+    return InvalidTarget();
```
wilhuff: This can just be `return {}`, which we've used elsewhere for returning a default value after an error.

Here and below.

Firestore/core/src/firebase/firestore/core/query.cc
```diff
@@ -332,17 +298,23 @@ std::string Query::ToString() const {
   return absl::StrCat("Query(canonical_id=", CanonicalId(), ")");
 }
 
+const Target& Query::ToTarget() const& {
+  if (memoized_target == nullptr) {
+    // Not using `make_shared` because the constructor is private.
+    memoized_target = std::shared_ptr<Target>(
```
wilhuff: You could use make_shared with a move constructor that is public, like this:

```c++
    Target target(path(), collection_group(), filters(), order_bys(), limit(),
               start_at(), end_at());
    memoized_target = std::make_shared<Target>(std::move(target));
```

PR #4420, opened 2019-12-02
Change unit tests to avoid OCMock verifying a method call on a real object

### LINKED ISSUE #4349: Fix FIAM to work with OCMock 3.5.0
From https://travis-ci.org/firebase/firebase-ios-sdk/jobs/613701590

FIRIAMFetchFlowTests
  testAlwaysFetchForNewlyInstalledMode, failed: caught "NSInternalInconsistencyException", "Mock object was not used in OCMVerify. Did you accidentally use a real object?"
  /Users/travis/build/firebase/firebase-ios-sdk/InAppMessaging/Example/Tests/FIRIAMFetchFlowTests.m:279
  ```
  // we expect that the message cache is checked for app launch messages
  OCMVerify([self.flow checkForAppLaunchMessage]);

christibbs:
OCMock was verifying a method call on a real `FIRIAMDisplayExecutor` object.

This change refactors `FIRIAMFetchFlow` to get a display executor injected. The failing unit test class is updated to mock a display executor and pass that to `FIRIAMFetchFlow` on initialization.

#fixes #4349


Firebase/InAppMessaging/Runtime/FIRIAMRuntimeManager.m
```diff
@@ -347,6 +337,17 @@ - (void)internalStartRuntimeWithSDKSettings:(FIRIAMSDKSettings *)settings {
                                              activityLogger:self.activityLogger
                                        analyticsEventLogger:analyticsEventLogger];
 
+  self.fetchOnAppForegroundFlow =
```
ryanwilson: Are there any implications about this being moved to after the `displayExecutor` instantiation?

christibbs: None that I can see. I also tested the end to end flow and there are no changes.

PR #4428, opened 2019-12-03
Simplify Firebase pod naming

paulb777:
Get rid of the name adjustments for the Firebase pods and always calling them by the full name - eg "FirebaseAnalytics" instead of "Analytics". T he rename only increased complexity in both implementation and usage.

#no-changelog

### DIFF
--- a/ZipBuilder/Template/README.md
+++ b/ZipBuilder/Template/README.md
@@ -15,7 +15,7 @@ To integrate a Firebase SDK with your app:
 2. Make sure you have an Xcode project open in Xcode.
 3. In Xcode, hit `⌘-1` to open the Project Navigator pane. It will open on
    left side of the Xcode window if it wasn't already open.
-4. Drag each framework from the "Analytics" directory into the Project
+4. Drag each framework from the "FirebaseAnalytics" directory into the Project
    Navigator pane. In the dialog box that appears, make sure the target you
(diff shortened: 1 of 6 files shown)


ZipBuilder/Template/README.md
```diff
@@ -15,7 +15,7 @@ To integrate a Firebase SDK with your app:
 2. Make sure you have an Xcode project open in Xcode.
 3. In Xcode, hit `⌘-1` to open the Project Navigator pane. It will open on
    left side of the Xcode window if it wasn't already open.
-4. Drag each framework from the "Analytics" directory into the Project
+4. Drag each framework from the "FirebaseAnalytics" directory into the Project
```
ryanwilson: Do you think it's worthwhile removing the `Firebase` prefix in the names of folders to match our existing structure? It's a small worry but was thinking if some larger companies have written scripts about automating this, we'd be breaking them. I know we don't promise the names won't change but seems pretty trivial to be able to rename the folders and remove the "Firebase" prefix if there is one in the post assembled steps.

paulb777: I would be surprised if anyone has automated the script since using the zip can be tricky to get right even when using manually.  We do have to update our internal games script. 

My preference would be to go forward with reducing the name confusion and document in the release notes.

ryanwilson: SGTM

PR #4436, opened 2019-12-04
Allow multiple `configure` calls in an extension.

ryanwilson:
Some extensions have an entry point that gets called multiple times.
Instead of having to check if a default app has already been configured,
allow multiple `configure` calls if and only if it's in an extension and
the FIROptions contents are identical.

### DIFF
--- a/Firebase/Core/FIRApp.m
+++ b/Firebase/Core/FIRApp.m
@@ -161,18 +161,9 @@ static NSMutableDictionary *sLibraryVersions;
 
   if ([name isEqualToString:kFIRDefaultAppName]) {
     if (sDefaultApp) {
-      // The default app already exists - this may be okay if we're in an extension where the
-      // entry point is called multiple times.
-      if ([self canConfigureAppMultipleTimes:sDefaultApp usingOptions:options]) {
-        // Everything is identical but the extension's lifecycle triggered `configure` twice. Ignore
-        // duplicate calls and return since everthing should still be in a valid state.
-        FIRLogDebug(kFIRLoggerCore, @"I-COR000035",
-                    @"Ignoring second `configure` call in an extension.");
-        return;
-      } else {
-        [NSException raise:kFirebaseCoreErrorDomain
-                    format:@"Default app has already been configured."];
-      }
+      // The default app already exixts. Handle duplicate `configure` calls and return.
+      [self appWasConfiguredTwice:sDefaultApp usingOptions:options];
+      return;
     }
 
     FIRLogDebug(kFIRLoggerCore, @"I-COR000001", @"Configuring the default app.");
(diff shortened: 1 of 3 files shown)


Firebase/Core/FIRApp.m
```diff
@@ -159,8 +161,18 @@ + (void)configureWithName:(NSString *)name options:(FIROptions *)options {
 
   if ([name isEqualToString:kFIRDefaultAppName]) {
     if (sDefaultApp) {
-      [NSException raise:kFirebaseCoreErrorDomain
-                  format:@"Default app has already been configured."];
+      // The default app already exists - this may be okay if we're in an extension where the
+      // entry point is called multiple times.
+      if ([self canConfigureAppMultipleTimes:sDefaultApp usingOptions:options]) {
+        // Everything is identical but the extension's lifecycle triggered `configure` twice. Ignore
+        // duplicate calls and return since everthing should still be in a valid state.
```
paulb777: `everything` here and below

Firebase/Core/FIRApp.m
```diff
@@ -176,8 +188,16 @@ + (void)configureWithName:(NSString *)name options:(FIROptions *)options {
 
     @synchronized(self) {
       if (sAllApps && sAllApps[name]) {
-        [NSException raise:kFirebaseCoreErrorDomain
-                    format:@"App named %@ has already been configured.", name];
+        // The app already exists but check to see if multiple `configure` calls are okay.
```
paulb777: Consider a function for this duplicate code.

Firebase/Core/FIRApp.m
```diff
@@ -199,6 +219,17 @@ + (void)configureWithName:(NSString *)name options:(FIROptions *)options {
   }
 }
 
++ (BOOL)canConfigureAppMultipleTimes:(FIRApp *)app usingOptions:(FIROptions *)options {
+  // Only extensions should potentially be able to use
```
paulb777: Period or more to say?

Firebase/Core/FIRApp.m
ryanwilson: Nice catch, thanks.

Firebase/Core/FIRApp.m
ryanwilson: Broke it out into a separate function.

Firebase/Core/FIRApp.m
ryanwilson: Nice catch, thanks!

ryanwilson:
Will wait for actual extension testing to ensure this does the trick before closing.

Firebase/Core/FIRApp.m
```diff
@@ -159,8 +161,9 @@ + (void)configureWithName:(NSString *)name options:(FIROptions *)options {
 
   if ([name isEqualToString:kFIRDefaultAppName]) {
     if (sDefaultApp) {
```
maksymmalyhin: It looks like `sDefaultApp` is not protected from the race condition here, though it's protected with `@synchronized` in other places. Should we add `@synchronized(self)` here as well?

Firebase/Core/FIROptions.m
```diff
@@ -346,6 +346,56 @@ - (void)setAppGroupID:(NSString *)appGroupID {
...
+  // We don't care about the `editingLocked` or `usingOptionsFromDefaultPlist` properties since
+  // those relate to lifecycle and construction, we only care if the contents of the options
+  // themselves are equal.
+  return YES;
+}
+
+- (NSUInteger)hash {
+  // This is strongly recommended for any object that implements a custom `isEqual:` method to
+  // ensure that dictionary and set behavior matches other `isEqual:` checks.
+  return self.optionsDictionary.hash ^ self.deepLinkURLScheme.hash ^ self.appGroupID.hash;
```
maksymmalyhin: nit: Was `analyticsOptionsDictionary` omitted here intentionally?

ryanwilson: Yes, added a comment mentioning why that was the case. Thanks!

Firebase/Core/FIRApp.m
ryanwilson: There are a few other locations where it's not protected, so I think in this case it's probably best to leave it as is to avoid any extra latency and other issues. If we start seeing issues I'm fine to lock it down though! Does that sound like a good plan?

maksymmalyhin: Should be OK for now, but I feel we should bring consistency to the way we access the static variables in the class.

Firebase/Core/FIROptions.m
maksymmalyhin: Thank you!

PR #4438, opened 2019-12-04
Throw an error if the fetch response returns an error (irrespective o…

### LINKED ISSUE #4100: RemoteConfig.remoteConfig().fetch return .success while the app is offline
### Describe your environment

  * Xcode version: 11.0 (11A420a)
  * Firebase SDK version: 6.10.0
  * Component version: 6.10.0

### Describe the problem
`RemoteConfig.remoteConfig().fetch` returns a `.success` status (`FIRRemoteConfigFetchStatus`) even though I disabled the internet connection.

Note: I never configured defaults via `setDefaults`, and I ran the app for the first time to be sure no previous config was cached.

#### Steps to reproduce:
- Create a new project importing Firebase/RemoteConfig through Cocoapods
- Follow this [guide](https://firebase.google.com/docs/remote-config/use-config-ios) by skipping the step 2 consisting of setting in-app defaults. 
- Disable the internet connection
- Run the app, the remote config is fetched

I got a `.success` status and a `nil` error even though the app is offline and I didn't set any defaults. I expect to get an error instead.

dmandar:
…f status code).
Fixes #4100

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNFetch.m
+++ b/FirebaseRemoteConfig/Sources/RCNFetch.m
@@ -358,7 +358,7 @@ static RCNConfigFetcherTestBlock gGlobalTestBlock;
       NSInteger statusCode = [((NSHTTPURLResponse *)response) statusCode];
 
       if (error || (statusCode != kRCNFetchResponseHTTPStatusCodeOK)) {
-        // Update failure fetch time and database.
+        // Update metadata about fetch failure.
         [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO];
         if (error) {
           if (strongSelf->_settings.lastFetchStatus == FIRRemoteConfigFetchStatusSuccess) {
@@ -370,6 +370,9 @@ static RCNConfigFetcherTestBlock gGlobalTestBlock;
           }
         }
         if (statusCode != kRCNFetchResponseHTTPStatusCodeOK) {
+          // Update metadata about fetch failure.
+          [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO];
+
           FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000026",
                       @"RCN Fetch failure. Response http error code: %ld", (long)statusCode);
           // Response error code 429, 500, 503 will trigger exponential backoff mode.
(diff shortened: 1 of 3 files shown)


FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
```diff
@@ -558,6 +558,128 @@ - (void)testFetchConfigsFailed {
                                }];
 }
 
+- (void)testFetchConfigsFailedErrorNoNetwork {
```
maksymmalyhin: nit: I wonder if it is too big effort to break this test to several methods, e.g. refactor all mocks setup to a separate method that can be called from the for-loop (and may be potentially used by other tests). Also, separate methods for configuring individual mocks may improve readability and code reuse. 

Feel free to consider it as a separate improvements.

FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -370,6 +370,9 @@ - (void)fetchWithUserProperties:(NSDictionary *)userProperties
           }
         }
         if (statusCode != kRCNFetchResponseHTTPStatusCodeOK) {
+          // Update metadata about fetch failure.
+          [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO];
```
maksymmalyhin: Looks like by this line the method is already called at line 362. Do we need to call this method 2nd time?

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
dmandar: sg. Other tests have the mocks initialized in the setup section. It is only the testFetchConfigsFailed and now this test that need a different response code back. But yes, makes sense to refactor..will make that a separate change.

FirebaseRemoteConfig/Sources/RCNFetch.m
dmandar: right. Deleted.

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
chliangGoogle: maybe add a todo here.

dmandar: Added.

PR #4440, opened 2019-12-04
Update CMake build docs for Windows

wilhuff:


CMAKE.md
```diff
@@ -61,16 +61,21 @@ support for Visual Studio 2015.
...
-version.
+```cmd
+choco install git
+choco install cmake --installargs 'ADD_CMAKE_TO_PATH=System'
+choco install ninja
 
-```
-choco install cmake.portable
-choco install ninja  # optional
+# Build scripts use uses bash and python
```
var-const: `use uses` -- seems like there's an extra word?

wilhuff: Done.

PR #4446, opened 2019-12-05
Fix crash when displaying UIAlertController for a test on device message that errored

christibbs:
Previously we were calling

```
[[UIApplication sharedApplication].keyWindow.rootViewController presentViewController:alert
                                                                               animated:YES
                                                                             completion:nil];
```

on a background thread. This change puts the alert presentation logic on the UI thread and also uses a different approach since UIApplication keyWindow is deprecated in iOS 13.


Firebase/InAppMessaging/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -307,14 +309,33 @@ - (void)displayMessageLoadError:(NSError *)error {
...
                                                         }];
 
   [alert addAction:defaultAction];
 
-  [[UIApplication sharedApplication].keyWindow.rootViewController presentViewController:alert
-                                                                               animated:YES
-                                                                             completion:nil];
+  dispatch_async(dispatch_get_main_queue(), ^{
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    if (@available(iOS 13.0, *)) {
```
chliangGoogle: would this two conditions redundant?

Firebase/InAppMessaging/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -41,6 +41,8 @@ @interface FIRIAMDisplayExecutor () <FIRInAppMessagingDisplayDelegate>
 @property(nonatomic) BOOL impressionRecorded;
 @property(nonatomic, nonnull, readonly) id<FIRIAMAnalyticsEventLogger> analyticsEventLogger;
 @property(nonatomic, nonnull, readonly) FIRIAMActionURLFollower *actionURLFollower;
+// Used for displaying the test on device message error alert.
+@property(nonatomic, strong) UIWindow *alertWindow;
```
ryanwilson: This needs `#import <UIKit/UIKit.h>` at the top - I believe it's only working now due to implicit imports.

Firebase/InAppMessaging/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -307,14 +309,33 @@ - (void)displayMessageLoadError:(NSError *)error {
...
+          break;
+        }
+      }
+      self.alertWindow = [[UIWindow alloc] initWithWindowScene:foregroundedScene];
+    }
+#else  // defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    self.alertWindow = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+#endif
+    UIViewController *alertViewController = [[UIViewController alloc] init];
+    self.alertWindow.rootViewController = alertViewController;
```
ryanwilson: Is there a reason we're now adding a blank `UIViewController` to the root and presenting from it instead of presenting on the window itself?

christibbs: I tried setting the alert controller as the root but it was trying to stretch the alert to fit the whole screen.

Firebase/InAppMessaging/Flows/FIRIAMDisplayExecutor.m
christibbs: Yes, Xcode complains without the inner check, though.

Firebase/InAppMessaging/Flows/FIRIAMDisplayExecutor.m
christibbs: Done.

PR #4448, opened 2019-12-05
Add "swift" flag to `firebaseUserAgent`

maksymmalyhin:
- add "swift" flag to `firebaseUserAgent` when Swift runtime detected
- tests

### DIFF
--- a/Firebase/Core/FIRApp.m
+++ b/Firebase/Core/FIRApp.m
@@ -114,6 +114,7 @@ static NSMutableArray<Class<FIRLibrary>> *sRegisteredAsConfigurable;
 static NSMutableDictionary *sAllApps;
 static FIRApp *sDefaultApp;
 static NSMutableDictionary *sLibraryVersions;
+static dispatch_once_t sFirebaseUserAgentOnceToken;
 
 + (void)configure {
   FIROptions *options = [FIROptions defaultOptions];
@@ -243,6 +244,7 @@ static NSMutableDictionary *sLibraryVersions;
     sAllApps = nil;
     [sLibraryVersions removeAllObjects];
     sLibraryVersions = nil;
+    sFirebaseUserAgentOnceToken = 0;
   }
 }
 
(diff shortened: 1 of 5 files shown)


FirebaseCore.podspec
```diff
@@ -48,4 +48,12 @@ Firebase Core includes FIRApp and FIROptions which provide central configuration
     unit_tests.dependency 'OCMock'
     unit_tests.resources = 'Example/Core/App/GoogleService-Info.plist'
   end
+
+  s.test_spec 'swift-unit' do |swift_unit_tests|
+    swift_unit_tests.source_files = 'Example/Core/Tests/Swift/**/*.swift',
+                                    'Example/Core/Tests/Swift/**/*.h'
+    swift_unit_tests.pod_target_xcconfig = {
+      'SWIFT_OBJC_BRIDGING_HEADER' => '$(PODS_ROOT)/../../Example/Core/Tests/Swift/FirebaseCore-iOS-Unit-swift-unit-Bridging-Header.h'
```
paulb777: Should probably be `PODS_TARGET_SRCROOT`.

Firebase/Core/FIRApp.m
```diff
@@ -549,6 +557,10 @@ + (NSString *)firebaseUserAgent {
   }
 }
 
++ (BOOL)hasSwiftRuntime {
+  return objc_lookUpClass("Swift._SwiftObject") != nil;
```
ryanwilson: Great! Is there a resource you could add (even as a PR description) describing where you found this information and why this will work?

maksymmalyhin: Good point! I added comments to the code. The idea originated from the [internal discussion](https://groups.google.com/a/google.com/g/appledev-discuss/c/QTGnFJPJzOQ/m/N74DXXcpCwAJ).

FirebaseCore.podspec
maksymmalyhin: Great, that works! Thank you for the suggestion.

Firebase/Core/FIRApp.m
paulb777: Will this always be true when running on ios 12.4 and newer, even if there is no swift?

maksymmalyhin: It was not on simulator. I think even if it is, it will be fine. I assume, the main question we would like to answer is "how many applications/devices will have size bump if Swift is added". 

But anyways, let me check on a real device just in case.

maksymmalyhin: For me the class `Swift._SwiftObject` is found only when a swift file present in the project on my iPhone iOS 13.2.3

paulb777: Interesting. So we'll need to cross-reference the Swift flag with the device iOS version to know where Swift causes a size impact.

maksymmalyhin: Yeah, looks like it.

PR #4449, opened 2019-12-06
Speed up zip builder development iteration

paulb777:
Doing the full `pod install` can take a long time especially with a slow network.

This speeds up a Firebase build by three minutes on a fast connection and much more on a slow one.

#no-changelog


paulb777:
Hmm, that would be an issue.

We're not running with ` -updatePodRepo false` on respins are we?  We could have two separate options, but that seems like overkill for managing different pod installation states. 

My preference is one option that toggles between assuming everything needs to be regenned or nothing does.

ryanwilson:
Actually that may have only been before we were building on Kokoro, since Kokoro is a blank state each run. I'm good to change it in that case!

PR #4453, opened 2019-12-06
Add -localPodspecPath option

paulb777:
Enables specifying a path to local podspecs.

This enables testing the zip builder without pushing podspecs and can be the basis for CI for binary framework testing.

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -315,7 +315,7 @@ public enum CocoaPodUtils {
       podfile += "  pod '\(pod.name)'"
       // Check if we want to use a local version of the podspec.
       if let localURL = LaunchArgs.shared.localPodspecPath,
-         FileManager.default.fileExists(atPath: localURL.appendingPathComponent(pod.name + ".podspec").path) {
+        FileManager.default.fileExists(atPath: localURL.appendingPathComponent(pod.name + ".podspec").path) {
         podfile += ", :path => '" + localURL.path + "'"
       } else if pod.version != nil {
         podfile += ", '\(pod.version!)'"
@@ -331,8 +331,8 @@ public enum CocoaPodUtils {
     if !versionsSpecified, let localURL = LaunchArgs.shared.localPodspecPath {
       let podspecs = try! FileManager.default.contentsOfDirectory(atPath: localURL.path)
       for podspec in podspecs {
-        if podspec.starts(with:"Google") && podspec.hasSuffix(".podspec") {
-          let podName = podspec.replacingOccurrences(of:".podspec", with:"")
+        if podspec.starts(with: "Google"), podspec.hasSuffix(".podspec") {
+          let podName = podspec.replacingOccurrences(of: ".podspec", with: "")
           podfile += "  pod '\(podName)', :path => '\(localURL.path)/\(podspec)'\n"
         }
       }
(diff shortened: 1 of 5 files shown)


ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -166,10 +166,14 @@ public enum CocoaPodUtils {
     let podsDir = projectDir.appendingPathComponent("Pods")
     var installedPods: [String: PodInfo] = [:]
     for (podName, version) in pods {
-      let podDir = podsDir.appendingPathComponent(podName)
-      guard FileManager.default.directoryExists(at: podDir) else {
-        fatalError("Directory for \(podName) doesn't exist at \(podDir) - failed while getting " +
-          "information for installed Pods.")
+      var podDir: URL
+      podDir = podsDir.appendingPathComponent(podName)
```
ryanwilson: Nit: this could be on the same line, no need to break it out.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -166,10 +166,14 @@ public enum CocoaPodUtils {
...
+      if !FileManager.default.directoryExists(at: podDir) {
+        guard let repoDir = LaunchArgs.shared.localPodspecPath else {
```
ryanwilson: Nit: a comment here may be good to say something along the lines of, "The pod doesn't exist in a remote directory, check the local podspec path to see if a local podspec was provided."

I was confused when I first read the code assuming it'd always be required but after typing out a comment about it I realized the situation.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -299,12 +303,35 @@ public enum CocoaPodUtils {
...
+
     // Loop through the subspecs passed in and use the actual Pod name.
     for pod in pods {
-      let version = pod.version == nil ? "" : ", '\(pod.version!)'"
-      podfile += "  pod '\(pod.name)'" + version + "\n"
+      podfile += "  pod '\(pod.name)'"
+      // Check if we want to use a local version of the podspec.
+      if let localURL = LaunchArgs.shared.localPodspecPath,
+        FileManager.default.fileExists(atPath: localURL.appendingPathComponent(pod.name + ".podspec").path) {
+        podfile += ", :path => '" + localURL.path + "'"
```
ryanwilson: Nit: May be easier to read as `podfile += ", :path => '\(localURL.path)'"` to match the below formatting.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -299,12 +303,35 @@ public enum CocoaPodUtils {
...
+      } else if pod.version != nil {
```
ryanwilson: We can use `if let` instead of force unwrapping here:
```
else if let podVersion = pod.version {
  podfile += ", '\(podVersion)'"
}
```

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -299,12 +303,35 @@ public enum CocoaPodUtils {
...
+        podfile += ", '\(pod.version!)'"
+      }
+      podfile += "\n"
+      if pod.version != nil {
+        versionsSpecified = true
```
ryanwilson: Can we also assign this up in the block above? Presumably if there's a local podspec there wouldn't also be a version associated with it (and can't be, I think)

paulb777:
@ryanwilson Comments addressed, rebased to master, and reverified.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
ryanwilson: Not sure if this was missed or it didn't work, just wanted to mention it incase the comment was lost :) Non blocking for merge though.

paulb777:
Good catch! I missed that one. Fixed now. Thanks.

PR #4458, opened 2019-12-09
M61 Core Changelog

paulb777:
What if anything should be added for the heartbeat and Swift logging changes?

### DIFF
--- a/Firebase/Core/CHANGELOG.md
+++ b/Firebase/Core/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 - [changed] Speed up initialization by lazily registering for the user agent. (#1306)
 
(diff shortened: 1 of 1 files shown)


Firebase/Core/CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
-# Unreleased
+# v6.4.0 -- M61
```
maksymmalyhin: Do you think we should add something for #4448? E.g.
```
- [added] Added a Swift usage flag to `firebaseUserAgent`. The information will be used to support product decisions related to Swift, e.g. adding a Swift specific API, SDKs, etc. ( #4448)
```

PR #4462, opened 2019-12-09
Consolidate SecureCoding to avoid a new subspec

paulb777:
Consolidate SecureCoding to avoid a new subspec

Also move heartbeat files out of third_party folder.

#no-changelog


paulb777:
@wilhuff  Can you see what's causing the cmake failures?

paulb777:
@wilhuff nvm I think i see the issue.

PR #4482, opened 2019-12-10
Make -[FIRInstanceID appInstanceID:] safe to be called from any thread

maksymmalyhin:
- use cached value of FID to be returned by `-[FIRInstanceID appInstanceID:]` instead of blocking thread waiting for `FirebaseInstallations` response.

b/145993195

### DIFF
--- a/Firebase/InstanceID/FIRInstanceID.m
+++ b/Firebase/InstanceID/FIRInstanceID.m
@@ -122,6 +122,9 @@ typedef NS_ENUM(NSInteger, FIRInstanceIDAPNSTokenType) {
 @property(atomic, strong, nullable)
     FIRInstanceIDCombinedHandler<NSString *> *defaultTokenFetchHandler;
 
+/// A cached value of FID. Should be used only for `-[FIRInstanceID appInstanceID:]`.
+@property(atomic, copy, nullable) NSString *firebaseInstallationsID;
+
 @end
 
 // InstanceID doesn't provide any functionality to other components,
@@ -611,6 +614,8 @@ static FIRInstanceID *gInstanceID;
   self.fcmSenderID = GCMSenderID;
   self.firebaseAppID = options.googleAppID;
 
+  [self updateFirebaseInstallationID];
+
   // FCM generates a FCM token during app start for sending push notification to device.
   // This is not needed for app extension.
   if (![GULAppEnvironmentUtil isAppExtension]) {
@@ -699,6 +704,7 @@ static FIRInstanceID *gInstanceID;
              selector:@selector(notifyAPNSTokenIsSet:)
                  name:kFIRInstanceIDAPNSTokenNotification
                object:nil];
+  [self observeFirebaseInstallationIDChanges];
 }
 
 #pragma mark - Private Helpers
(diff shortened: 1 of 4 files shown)


Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -611,6 +614,8 @@ - (void)configureInstanceIDWithOptions:(FIROptions *)options {
   self.fcmSenderID = GCMSenderID;
   self.firebaseAppID = options.googleAppID;
 
+  [self updateFirebaseInstallationID];
```
chliangGoogle: This method currently triggers [FIRInstanceID didCompleteConfigure] which will get a IID and token in that process. I wonder if this branch is still the same?
In that case this method might not be needed we can just set the installationID from didCompleteConfigure method.

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -1092,4 +1098,35 @@ - (void)logAPNSConfigurationError:(NSString *)errorString {
...
 }
 
+#pragma mark - Sync InstanceID
+
+- (void)updateFirebaseInstallationID {
+  FIRInstanceID_WEAKIFY(self);
+  [self.installations
+      installationIDWithCompletion:^(NSString *_Nullable installationID, NSError *_Nullable error) {
+        FIRInstanceID_STRONGIFY(self);
+        self.firebaseInstallationsID = installationID;
```
chliangGoogle: This seems set async, so does it mean there's a chance that if analytics query it too fast, it might be nil?

maksymmalyhin: Yes, there is a chance that it will be `nil`. Particularly for analytics it will unlikely be an issue due to their use case - IID is requested before sending data which happens way after Firebase initialization (we discussed it with @htcgh some time ago). 

The sync method is deprecated in favour the public async method and will be deleted eventually. Taking this into account it looks better to me to provide safe and fast implementation which may return `nil` (which was a case even before) rather than keep slow and unsafe (may lead to a deadlock).

AFIK, the sync method is not be used by recent versions of other SDKs (except Analytics), so it is mostly to be on the safe side in the case when e.g. an old version of remote config is locked but IID is update. 

WDYT?

chliangGoogle: I wonder if we could just get rid of the sync method and ask analytics to call the async method instead. Otherwise we are adding extra code just to call async method from a sync one.

maksymmalyhin: Yeah, actually you're right. Removing the method will require a minor version bump for all dependent SDKs, but it is what we will do at the release time anyway. This is something we'll need to discuss with them though. BTW, Analytics will be doing it eventually. The implementation will be very similar on the Analytics side (cl/262474711).

Firebase/InstanceID/FIRInstanceID.m
maksymmalyhin: As far as I can see `[self didCompleteConfigure];` isn't called for extensions. I'm no sure if it's an issue though. I would prefer to keep it independent, because only the first FID request is relatively expensive and the consequent ones are chip because of the caching.

Firebase/InstanceID/FIRInstanceID.m
paulb777: Nope, we can't delete private APIs from IID until a major version update since locking to an old 6.x version of Analytics allows IID minor version updates of its IID dependency.

maksymmalyhin: @paulb777 Yeah, good point, there are some valid combinations of locks when the removed API may be called.

@chliangGoogle I created a bug #4486 to remove this code for the Firebase 7 release.

chliangGoogle: I'm okay with current change in that case. maybe we can add a todo here to say when major update, we can just get rid of this method?

chliangGoogle: This is a special case for Analytics or just in general we can't remove private API until major version update?

Firebase/InstanceID/FIRInstanceID.m
chliangGoogle: it is not called for extensions, but that's just to keep app extension less expensive so we don't have to worry about it for now.

Let's keep it independent then as that's an old function and is as always overly complicated.

Firebase/InstanceID/FIRInstanceID.m
paulb777: It's technically true for any private API used across pods in Firebase.

It's especially true for IID since that doesn't get included and locked with the Firebase pod.

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -122,6 +122,9 @@ @interface FIRInstanceID ()
 @property(atomic, strong, nullable)
     FIRInstanceIDCombinedHandler<NSString *> *defaultTokenFetchHandler;
 
+/// A cached value of FID. Should be used only for `-[FIRInstanceID appInstanceID:]`.
```
chliangGoogle: btw I don't see this firebaseInstallationsID is read anywhere, so what's the point of caching here?

maksymmalyhin: It's read in another file [here](https://github.com/firebase/firebase-ios-sdk/blob/0831e2d677be376495fe1fcb8fbb0e282afc505b/Firebase/InstanceID/FIRInstanceID%2BPrivate.m#L44)

PR #4491, opened 2019-12-12
[2/4]LimitToLast: Make SyncEngine work with Target

wu-hui:
A port of firebase/firebase-android-sdk#965.


Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -367,11 +373,18 @@ DocumentKeySet SyncEngine::GetRemoteKeys(TargetId target_id) const {
...
   } else {
-    auto found = query_views_by_target_.find(target_id);
-    if (found != query_views_by_target_.end()) {
-      return found->second->view().synced_documents();
+    DocumentKeySet keys;
+    if (queries_by_target_.count(target_id) == 0) {
+      return keys;
+    }
+
+    for (auto& query : queries_by_target_.at(target_id)) {
```
wilhuff: Since we don't need or want to mutate the query here (or the key below) these should be `const auto&`.

Firestore/core/src/firebase/firestore/local/local_store.cc
```diff
@@ -430,38 +427,20 @@ QueryData LocalStore::AllocateTarget(Target target) {
...
-
-    TargetId target_id = query_data->target_id();
-
-    auto found = target_ids.find(target_id);
-    if (found != target_ids.end()) {
-      const QueryData& cached_query_data = found->second;
-
-      if (cached_query_data.snapshot_version() >
-          query_data->snapshot_version()) {
-        // If we've been avoiding persisting the resume_token (see
```
wilhuff: Just to confirm: you're deleting this logic because it was deleted on the other platforms already and this just brings iOS into line, right?

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/local/local_store.cc
wu-hui: Correct.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wilhuff: Please make both `query` and `key` `const auto&`

wu-hui: Done.

PR #4498, opened 2019-12-13
Fix crash in Scene delegate swizzling when delegate is nil

maksymmalyhin:
- check on `UIScene.delegate` before swizzling

b/146216242


GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -1069,7 +1069,7 @@ + (void)proxySceneDelegateIfNeeded:(UIScene *)scene {
 
   // Skip proxying if the class has a prefix of kGULAppDelegatePrefix, which means it has been
```
renkelvin: Maybe change the comment as well.

maksymmalyhin: Yeah, it's a good idea. Sorry I forgot about it. I would prefer to do it separately to not disturb Travis.

renkelvin: No problem! Thanks for the fix!

PR #4505, opened 2019-12-16
[3/4] Implement LimitToLast

wu-hui:


Firestore/Example/Tests/Util/FSTIntegrationTestCase.h
```diff
@@ -102,6 +102,7 @@ extern "C" {
 - (void)updateDocumentRef:(FIRDocumentReference *)ref data:(NSDictionary<NSString *, id> *)data;
 
 - (void)deleteDocumentRef:(FIRDocumentReference *)ref;
+- (void)addDocumentRef:(FIRCollectionReference *)ref data:(NSDictionary<NSString *, id> *)data;
```
wilhuff: Newline between methods to match the rest of this file.

Firestore/Example/Tests/Integration/API/FIRQueryTests.mm
```diff
@@ -56,6 +56,89 @@ - (void)testLimitQueriesWithDescendingSortOrder {
                         (@[ @{@"k" : @"d", @"sort" : @2}, @{@"k" : @"c", @"sort" : @1} ]));
 }
 
+- (void)testLimitToLastMustAlsoHaveExplicitOrderBy {
+  FIRCollectionReference *collRef = [self collectionRefWithDocuments:@{}];
+  FIRQuery *query = [collRef queryLimitedToLast:2];
+  @try {
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/src/firebase/firestore/core/query.h
```diff
@@ -39,6 +39,8 @@ namespace core {
 
 using CollectionGroupId = std::shared_ptr<const std::string>;
 
+enum class LimitType { First, Last, None };
```
wilhuff: If you put the `None` value first in the list it will be the one that gets the zero value. Since this is a primitive value that's not as useful as it could be (because primitives don't normally get default initialized) but in cases where memory is zeroed out it's useful for that to create a `None`.

Firestore/core/src/firebase/firestore/remote/serializer.cc
```diff
@@ -1022,7 +1023,7 @@ Target Serializer::DecodeQueryTarget(
 
   return Query(std::move(path), std::move(collection_group),
                std::move(filter_by), std::move(order_by), limit,
-               std::move(start_at), std::move(end_at))
+               LimitType::None, std::move(start_at), std::move(end_at))
```
wilhuff: `LimitType::None` doesn't seem right here if `limit` ends up with a value other than `Target::kNoLimit`.

Should this be `LimitType::First`?

Firestore/core/src/firebase/firestore/core/query.h
wu-hui: > LGTM, with a minor suggestion and a last straggling maybe issue.

The issues raised are now fixed.

PR #4511, opened 2019-12-17
FirebaseInstallations: require Core version with heartbeat

maksymmalyhin:
- FIS uses heartbeat code from Firebase Core, so it should require proper version of it.

#no-changelog


maksymmalyhin:
Realized that the Firebase Core version with heartbeat support has not been released yet. Closing the PR so far, will reopen after the release with the proper version.

PR #4512, opened 2019-12-17
[4/4] Port spec tests for limitToLast

wu-hui:


Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
```diff
@@ -48,6 +50,11 @@ class Persistence;
 namespace core = firebase::firestore::core;
 namespace local = firebase::firestore::local;
 namespace model = firebase::firestore::model;
+namespace nanopb = firebase::firestore::nanopb;
+
+using ActiveTargetMap =
```
wilhuff: Could you add comments to this declaration indicating what this should contain? Previously this was just a map of target id to query data, but now I have no idea what's going on here.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
```diff
@@ -31,6 +32,7 @@
 #include "Firestore/core/src/firebase/firestore/model/mutation.h"
 #include "Firestore/core/src/firebase/firestore/model/snapshot_version.h"
 #include "Firestore/core/src/firebase/firestore/model/types.h"
+#include "Firestore/core/src/firebase/firestore/nanopb/nanopb_util.h"
```
wilhuff: Include nanopb/byte_string.h to get a declaration of `ByteString`.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
```diff
@@ -473,11 +475,11 @@ - (void)receiveWatchStreamError:(int)errorCode userInfo:(NSDictionary<NSString *
   return _datastore->ActiveTargets();
 }
 
-- (const std::unordered_map<TargetId, QueryData> &)expectedActiveTargets {
+- (const ActiveTargetMap &)expectedActiveTargets {
   return _expectedActiveTargets;
 }
 
-- (void)setExpectedActiveTargets:(const std::unordered_map<TargetId, QueryData> &)targets {
+- (void)setExpectedActiveTargets:(const ActiveTargetMap &)targets {
```
wilhuff: You could pass by value and `std::move(targets)` to potentially avoid an O(n) copy of the target map. (The caller would still have to `std::move` to save the copy).

Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -110,10 +110,23 @@ ViewSnapshot SyncEngine::InitializeViewAndComputeSnapshot(const Query& query,
   DocumentMap docs = local_store_->ExecuteQuery(query);
   DocumentKeySet remote_keys = local_store_->GetRemoteDocumentKeys(target_id);
 
+  // If there are already queries mapped to the target id, create a synthesized
+  // target change to apply the sync state from those queries to the new query.
+  auto current_sync_state = SyncState::None;
+  TargetChange synthesized_current_change;
```
wilhuff: This isn't equivalent to the android version. There this defaults to null, and only if there's a synthesized change does it get set to a non-null value.

Here you're creating a default TargetChange. This type should likely be `absl::optional<TargetChange>` (which is the type of the parameter to `ViewChange::ApplyChanges`.

Firestore/core/src/firebase/firestore/remote/remote_event.h
```diff
@@ -74,6 +74,9 @@ class TargetChange {
  public:
   TargetChange() = default;
 
+  explicit TargetChange(bool current) : current_{current} {
```
wilhuff: The Android version has this as a static factory named `createSynthesizedTargetChangeForCurrentChange`. This is considerably more explicit and self-documenting than just a constructor taking a boolean. Please port.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
wu-hui: Done.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.h
wu-hui: Added comment.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wu-hui: Done.

Firestore/core/src/firebase/firestore/remote/remote_event.h
wu-hui: Done.

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
wu-hui: Done.

PR #4514, opened 2019-12-17
Cherry-picks for Firestore 1.8.3

wilhuff:
This is everything we've found required to support C++ and Unity based on iOS M61.


wilhuff:
Note that if perform a regular merge of this PR (rather than squash/merge) then there will be no conflict with the corresponding changes on master).

wilhuff:
No need to change the config; I did it from the command-line.

PR #4518, opened 2019-12-18
GULAppDelegateSwizzler - comments updated

maksymmalyhin:


maksymmalyhin:
I am cancelling the rest of Travis jobs because there are no actual code changes.

maksymmalyhin:
I would say `delegate == nil` is the only common case way, but having two more methods (potentially more added?) in between made me think that validating the name itself is a good idea.

GoogleUtilities/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- Update comments in `GULAppDelegateSwizzler` (#4518)
```
wilhuff: Suggestion: these comment changes don't affect consumers of this package so don't bother your users with changelog entries. If users want to see everything we do they can browse git history. This should be a quick list of things that might affect them or might prove useful.

maksymmalyhin: Good point, I'll remove it then.

PR #4523, opened 2019-12-18
Work around spurious errors in GCC 4.8

wilhuff:
GCC shows a "chosen constructor is explicit in copy-initialization" that
shouldn't occur. Also, since we're now explicitly instantiating these
values, add type aliases in RemoteEvent.

This effectively a port of cl/286120322 with cleanup.

### DIFF
--- a/Firestore/core/src/firebase/firestore/core/sync_engine.cc
+++ b/Firestore/core/src/firebase/firestore/core/sync_engine.cc
@@ -281,13 +281,19 @@ void SyncEngine::HandleRejectedListen(TargetId target_id, Status error) {
     // store's invariants with another method.
     NoDocument doc(limbo_key, SnapshotVersion::None(),
                    /* has_committed_mutations= */ false);
-    DocumentKeySet limbo_documents = DocumentKeySet{limbo_key};
-    RemoteEvent event{SnapshotVersion::None(), /*target_changes=*/{},
-                      /*target_mismatches=*/{},
-                      /*document_updates=*/{{limbo_key, doc}},
+
+    // Explicitly instantiate these to work around a spurious "chosen
+    // constructor is explicit in copy-initialization" error in the RemoteEvent
+    // constructor on GCC 4.9.
+    DocumentKeySet limbo_documents{limbo_key};
+    RemoteEvent::TargetChangeMap target_changes;
+    RemoteEvent::TargetSet target_mismatches;
+    RemoteEvent::DocumentUpdateMap document_updates{{limbo_key, doc}};
+
+    RemoteEvent event{SnapshotVersion::None(), std::move(target_changes),
+                      std::move(target_mismatches), std::move(document_updates),
                       std::move(limbo_documents)};
     ApplyRemoteEvent(event);
-
   } else {
     auto found = query_views_by_target_.find(target_id);
     HARD_ASSERT(found != query_views_by_target_.end(), "Unknown target id: %s",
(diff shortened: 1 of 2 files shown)


Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -281,13 +281,19 @@ void SyncEngine::HandleRejectedListen(TargetId target_id, Status error) {
...
+    // Explicitly instantiate these to work around a spurious "chosen
+    // constructor is explicit in copy-initialization" error in the RemoteEvent
+    // constructor on GCC 4.9.
+    DocumentKeySet limbo_documents{limbo_key};
+    RemoteEvent::TargetChangeMap target_changes;
+    RemoteEvent::TargetSet target_mismatches;
+    RemoteEvent::DocumentUpdateMap document_updates{{limbo_key, doc}};
+
+    RemoteEvent event{SnapshotVersion::None(), std::move(target_changes),
+                      std::move(target_mismatches), std::move(document_updates),
```
var-const: Optional: you can instantiate these on the fly if you'd like:
```
RemoteEvent event{SnapshotVersion::None(), RemoteEvent::TargetChangeMap{},
  // ...and so on...
```
OTOH, perhaps this change would make the line too verbose, so feel free to keep as is.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -281,13 +281,19 @@ void SyncEngine::HandleRejectedListen(TargetId target_id, Status error) {
...
     NoDocument doc(limbo_key, SnapshotVersion::None(),
                    /* has_committed_mutations= */ false);
-    DocumentKeySet limbo_documents = DocumentKeySet{limbo_key};
-    RemoteEvent event{SnapshotVersion::None(), /*target_changes=*/{},
-                      /*target_mismatches=*/{},
-                      /*document_updates=*/{{limbo_key, doc}},
+
+    // Explicitly instantiate these to work around a spurious "chosen
+    // constructor is explicit in copy-initialization" error in the RemoteEvent
+    // constructor on GCC 4.9.
```
var-const: Nit: is the error in GCC 4.9? The log Jon posted seems to indicate GCC 4.8:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/unordered_map.h:138:7: note: explicit constructor declared here
      unordered_map(size_type __n = 10,
```

Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -281,13 +281,19 @@ void SyncEngine::HandleRejectedListen(TargetId target_id, Status error) {
     // store's invariants with another method.
     NoDocument doc(limbo_key, SnapshotVersion::None(),
                    /* has_committed_mutations= */ false);
-    DocumentKeySet limbo_documents = DocumentKeySet{limbo_key};
-    RemoteEvent event{SnapshotVersion::None(), /*target_changes=*/{},
-                      /*target_mismatches=*/{},
-                      /*document_updates=*/{{limbo_key, doc}},
+
+    // Explicitly instantiate these to work around a spurious "chosen
```
var-const: Note: what I think happens is that GCC implemented both the default constructor and the constructor taking a single integer argument (bucket count) as the same constructor with a default argument value. This is incorrect because one of those has to be `explicit` (the one taking an integer argument) and the other one has to be _not_ `explicit` (the default constructor). This implementation, however, has the effect of making both of them, default constructor included, `explicit`. An `explicit` default constructor prevents this form of initialization:
```
foo = {}; // Error if default constructor of `Foo` is explicit
foo = Foo{}; // Ok
```

It's just for the record -- feel free to incorporate any details you find useful into the comment, or just leave as is.

wilhuff: Good points, all. PTAL.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wilhuff: Updated.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wilhuff: Left as-is.

PR #4524, opened 2019-12-18
Add GULSceneDelegateSwizzler

renkelvin:


GoogleUtilities.podspec
```diff
@@ -77,6 +77,15 @@ other Google CocoaPods. They're not intended for direct public usage.
     adss.dependency 'GoogleUtilities/Environment'
   end
 
+  s.subspec 'SceneDelegateSwizzler' do |sdss|
```
maksymmalyhin: Cocoapods subspecs are actually broken and we are moving away from using them (will be done at I/O), so we should not introduce new subspecs. I think we should move the files to `AppDelegateSwizzler`.

cc @paulb777

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -0,0 +1,153 @@
...
+@implementation GULSceneDelegateSwizzlerTest
+
+- (void)setUp {
+  [super setUp];
+}
+
+- (void)tearDown {
+  //  [GULSceneDelegateSwizzler clearInterceptors];
+  [super tearDown];
+}
```
maksymmalyhin: Can we remove empty methods?

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -0,0 +1,153 @@
...
+    [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:interceptor2];
+
+    [realSceneDelegate scene:mockSharedScene openURLContexts:urlContexts];
+    OCMVerifyAll(interceptor);
+    OCMVerifyAll(interceptor2);
+  }
+}
+
+#endif
+
```
maksymmalyhin: I think we have to copy the tests for honoring flags as well (https://github.com/firebase/firebase-ios-sdk/blob/6fe312c451f0afb66b1782b38d38ccc6f857575b/GoogleUtilities/Example/Tests/Swizzler/GULAppDelegateSwizzlerTest.m#L1110), unless we can move this logic to a separate place and test it there.

GoogleUtilities.podspec
paulb777: +1 Please do not add another subspec

GoogleUtilities/Example/GoogleUtilities.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		40C06D7323AAABE40055BC8B /* GULSceneDelegateSwizzlerTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 40C06D7223AAABE40055BC8B /* GULSceneDelegateSwizzlerTest.m */; };
```
paulb777: We can delete/stop maintaining this file in favor of pure `pod gen` based development.

renkelvin: You mean the whole project.pbxproj file? Maybe do it in a separate pr.

GoogleUtilities.podspec
renkelvin: Done.

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
renkelvin: Alex suggested that we reuse the same flags. What do you think?

maksymmalyhin: Do you mean using GULAppDelegateSwizzler to get value in the scene swizzler? If yes - SGTM.

GoogleUtilities/Example/GoogleUtilities.xcodeproj/project.pbxproj
paulb777: Yep. That's fine. I wanted to let you know not to go to too much trouble to keep it working.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
```diff
@@ -0,0 +1,465 @@
...
+static char const *const kGULRealClassKey = "GUL_realClass";
+
+static GULLoggerService kGULLoggerSwizzler = @"[GoogleUtilities/SceneDelegateSwizzler]";
+
+// Since Firebase SDKs also use this for app delegate proxying, in order to not be a breaking change
+// we disable App Delegate proxying when either of these two flags are set to NO.
+
+/** Plist key that allows Firebase developers to disable Scene Delegate Proxying. */
+static NSString *const kGULFirebaseSceneDelegateProxyEnabledPlistKey =
+    @"FirebaseSceneDelegateProxyEnabled";
```
maksymmalyhin: Introducing a new Info.plist key actually should go through the API review. As for me, reusing the app delegate flag is a good option. In thins case we will spare developers from taking care of another flag also we will be able to keep the docs mostly unchanged.

@ryanwilson WDYT?

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -0,0 +1,126 @@
...
+    [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:interceptor];
+    [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:interceptor2];
+
+    [realSceneDelegate scene:mockSharedScene openURLContexts:urlContexts];
+    OCMVerifyAll(interceptor);
+    OCMVerifyAll(interceptor2);
+  }
+}
+
```
maksymmalyhin: I would add tests for the `NSNotificationCenter` related code as well.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
ryanwilson: +1, I think re-using the app delegate flag is the right option as well. Although it's now unfortunately named along with the AppDelegate, ultimately the purpose is to remove boilerplate for startup which SceneDelegate is now part of.

If there are valid use cases you can think of where a developer would want to disable AppDelegate swizzling but keep SceneDelegate (or vice-versa) we can re-evaluate.

renkelvin: Reused AppDelegate flags

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
renkelvin: Reused AppDelegate flags.

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -439,6 +445,11 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
           appCredentialManager:strongSelf->_appCredentialManager];
 
       [GULAppDelegateSwizzler registerAppDelegateInterceptor:strongSelf];
+      #if ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
+        if (@available(iOS 13, tvos 13, *)) {
+        [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:strongSelf];
+      }
+      #endif
```
maksymmalyhin: Could you please add annotations to `#endif` like `#endif // TARGET_OS_IOS`

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
```diff
@@ -0,0 +1,466 @@
...
+@end
+
+@implementation GULSceneZeroingWeakContainer
+@end
+
+@implementation GULSceneDelegateSwizzler
+
+#pragma mark - Public methods
+
++ (BOOL)isSceneDelegateProxyEnabled {
```
maksymmalyhin: I wonder if we can use `return [GULAppDelegateSwizzler isAppDelegateProxyEnabled]` as an implementation of the method to avoid code duplication?

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
maksymmalyhin: I think we still need tests for the logic protected by the flag even if we reuse the flag calculation itself.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
```diff
@@ -0,0 +1,466 @@
...
+  if ([isGoogleProxyEnabledPlistValue isKindOfClass:[NSNumber class]]) {
+    isGoogleUtilitiesSceneDelegateProxyEnabled = [isGoogleProxyEnabledPlistValue boolValue];
+  }
+
+  // Only deactivate the proxy if it is explicitly disabled by app developers using either one of
+  // the plist flags.
+  return isFirebaseSceneDelegateProxyEnabled && isGoogleUtilitiesSceneDelegateProxyEnabled;
+}
+
++ (GULSceneDelegateInterceptorID)registerSceneDelegateInterceptor:(id<UISceneDelegate>)interceptor {
```
maksymmalyhin: In the future we need to think how to avoid code duplication with `GULAppDelegateSwizzler`

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -384,6 +385,11 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
     UIApplication *application = [applicationClass sharedApplication];
 
     [GULAppDelegateSwizzler proxyOriginalDelegateIncludingAPNSMethods];
+    #if ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
```
baolocdo: Using Macro does not work in framework as the this will be built before developers can use anyway. Please use runtime check only

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -384,6 +385,11 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
...
+      if (@available(iOS 13, tvos 13, *)) {
+      [GULSceneDelegateSwizzler proxyOriginalSceneDelegate];
+    }
+    #endif
```
baolocdo: please use comment to describe which `#endif` belongs to which `#if`

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -439,6 +445,11 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
           appCredentialManager:strongSelf->_appCredentialManager];
 
       [GULAppDelegateSwizzler registerAppDelegateInterceptor:strongSelf];
+      #if ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
```
baolocdo: ditto, the macro check is not needed

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -0,0 +1,140 @@
...
+@implementation GULTestSceneDelegate
+@end
+
+@interface GULSceneDelegateSwizzlerTest : XCTestCase
+@end
+
+@implementation GULSceneDelegateSwizzlerTest
+
+- (void)testProxySceneDelegateWithNoSceneDelegate {
+  if (@available(iOS 13, tvOS 13, *)) {
```
baolocdo: can we only target some tests on some particular iOS > 12? This seems to be lots of ugly checks for each test. If not, please have an early return check 
```
if (!@available(iOS 13, tvOS 13, *) { return ;}
...
```

GoogleUtilities/SceneDelegateSwizzler/Private/GULSceneDelegateSwizzler.h
```diff
@@ -0,0 +1,81 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString *const GULSceneDelegateInterceptorID;
+
+/** This class contains methods that isa swizzle the scene delegate. */
+@interface GULSceneDelegateSwizzler : NSProxy
+
+#if UISCENE_SUPPORTED
+
+/** Registers an app delegate interceptor whose methods will be invoked as they're invoked on the
```
baolocdo: Should be "Registers a scene delegate". Here and below.

GoogleUtilities/SceneDelegateSwizzler/Private/GULSceneDelegateSwizzler.h
```diff
@@ -0,0 +1,81 @@
...
+    (id<UISceneDelegate>)interceptor API_AVAILABLE(ios(13.0), tvos(13.0));
+
+/** Unregisters an interceptor with the given ID if it exists.
+ *
+ *  @param interceptorID The object that was generated when the interceptor was registered.
+ */
++ (void)unregisterSceneDelegateInterceptorWithID:(GULSceneDelegateInterceptorID)interceptorID
+    API_AVAILABLE(ios(13.0), tvos(13.0));
+
+/** This method ensures that the original app delegate has been proxied. Call this before
```
baolocdo: Please update the comments

Firebase/Auth/Source/Auth/FIRAuth.m
paulb777: Both compile-time and run-time checks are needed. Most developers use Auth as a source-based CocoaPod.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
renkelvin: My general idea is to have a single class that takes care of Application delegate, Scene delegate and even more delegates in the future, since there are many shared logic.
While for now, to solve the scene delegate issue and at the same time minimize the impact on App delegate swizzler, we just created a standalone class for scene delegate.
It depends to duplicate the code or not. Please see my comments in each case above.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
renkelvin: I think we should keep AppDelegateSwizzler and SceneDelegateSwizzler separated. Reusing that method can be confusing.

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
renkelvin: It makes sense. I gonna duplicate those tests of AppDelegateSwizzler to SceneDelegateSwizzler.

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -384,6 +385,11 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
     UIApplication *application = [applicationClass sharedApplication];
 
     [GULAppDelegateSwizzler proxyOriginalDelegateIncludingAPNSMethods];
+    #if ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
+      if (@available(iOS 13, tvos 13, *)) {
+      [GULSceneDelegateSwizzler proxyOriginalSceneDelegate];
```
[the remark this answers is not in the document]
renkelvin: Make sense. Checks removed.

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
renkelvin: I just tried early return, however the complier raises warnings since the ios 13+ apis are not guarded with `@available`.

`@available does not guard availability here; use if (@available) instead`

Firebase/Auth/Source/Auth/FIRAuth.m
```diff
@@ -439,7 +441,12 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
           appCredentialManager:strongSelf->_appCredentialManager];
 
       [GULAppDelegateSwizzler registerAppDelegateInterceptor:strongSelf];
-      #endif
+      #if ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
+        if (@available(iOS 13, tvos 13, *)) {
```
maksymmalyhin: nit: I think this part can be changed in the same way as `[GULSceneDelegateSwizzler proxyOriginalSceneDelegate];`. Doesn't look like a blocker to me though.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
maksymmalyhin: As for me, as far as the implementation is identical, we should reuse it in that or another way. A more or less graceful solution could be introducing a separate class or function that encapsulates the implementation and is used by both App Delegate and Scene Delegate swizzlers. Then we will have cleaner code and less tests duplication. 

The downside of this approach is that we would have to change well tested and critical for Analytics App Delegate Swizzler which may introduce another issue. Taking this into account, using `[GULAppDelegateSwizzler isAppDelegateProxyEnabled]` as the implementation of `isSceneDelegateProxyEnabled` looks like a reasonable tradeoff, so we have less code duplication but don't touch  `GULAppDelegateSwizzler` just before the release.

I am kind of ok with the current code, though it feels like we will definitely forget to update one of the methods when the flag potentially changes  in the future.

WDTY?

Firebase/Auth/Source/Auth/FIRAuth.m
renkelvin: The interceptor is a class of id<UISceneDelegate>, let me see if we can do the same.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
renkelvin: Sounds good, I don't have a strong preference on this.

Firebase/Auth/Source/Auth/FIRAuth.m
renkelvin: Seems we have to keep the checks because of `UISceneDelegate`.

GoogleUtilities/Common/GULLoggerCodes.h
```diff
@@ -18,22 +18,38 @@
...
+  kGULSwizzlerMessageCodeAppDelegateSwizzling005 = 1005,                 // I-SWZ001005
+  kGULSwizzlerMessageCodeAppDelegateSwizzling006 = 1006,                 // I-SWZ001006
+  kGULSwizzlerMessageCodeAppDelegateSwizzling007 = 1007,                 // I-SWZ001007
+  kGULSwizzlerMessageCodeAppDelegateSwizzling008 = 1008,                 // I-SWZ001008
+  kGULSwizzlerMessageCodeAppDelegateSwizzling009 = 1009,                 // I-SWZ001009
+  kGULSwizzlerMessageCodeAppDelegateSwizzling010 = 1010,                 // I-SWZ001010
+  kGULSwizzlerMessageCodeAppDelegateSwizzling011 = 1011,                 // I-SWZ001011
+  kGULSwizzlerMessageCodeAppDelegateSwizzling012 = 1012,                 // I-SWZ001012
+  kGULSwizzlerMessageCodeAppDelegateSwizzling013 = 1013,                 // I-SWZ001013
+  kGULSwizzlerMessageCodeAppDelegateSwizzlingInvalidAppDelegate = 1014,  // I-SWZ001014
```
baolocdo: we should keep the format of the message code enum consistently. Otherwise, it will be difficult to search for if the users provide the log with the message code. Here and below

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -0,0 +1,140 @@
...
+    id interceptor2 = OCMProtocolMock(@protocol(TestSceneProtocol));
+    OCMExpect([interceptor2 scene:mockSharedScene openURLContexts:urlContexts]);
+
+    [GULSceneDelegateSwizzler proxySceneDelegateIfNeeded:mockSharedScene];
+
+    [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:interceptor];
+    [GULSceneDelegateSwizzler registerSceneDelegateInterceptor:interceptor2];
+
+    [realSceneDelegate scene:mockSharedScene openURLContexts:urlContexts];
+    OCMVerifyAll(interceptor);
```
baolocdo: OC Mock might lead to memory leaks across multiple tests. Our recommendation is to always stop mocking and explicitly set the mock object to nil at the end of each unit test. We have seen tests crashed once in a while and after doing that, the flakiness went away. This is also recommended by readability reviewers.

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -135,6 +147,149 @@ - (void)testNotificationCenterRegister {
   }
 }
 
+#pragma mark - Tests to test that Plist flag is honored
+
+/** Tests that app delegate proxy is enabled when there is no Info.plist dictionary. */
```
baolocdo: Please update all the comments from "App Delegate" to "Scene" here and below

GoogleUtilities/Example/Tests/Swizzler/GULSceneDelegateSwizzlerTest.m
```diff
@@ -135,6 +147,149 @@ - (void)testNotificationCenterRegister {
...
+ *  Info.plist key present.
+ */
+- (void)testAppProxyPlistFlag_NoSceneDelegateProxyKey {
+  // No app delegate disable key. If there is no key, the default should be enabled.
+  NSDictionary *mainDictionary = @{@"randomKey" : @"randomValue"};
+  id mainBundleMock = OCMPartialMock([NSBundle mainBundle]);
+  [[[mainBundleMock expect] andReturn:mainDictionary] infoDictionary];
+
+  XCTAssertTrue([GULSceneDelegateSwizzler isSceneDelegateProxyEnabled]);
+  [mainBundleMock stopMocking];
```
baolocdo: OC Mock might lead to memory leaks across multiple tests. Our recommendation is to always stop mocking and explicitly set the mock object to nil at the end of each unit test. We have seen tests crashed once in a while and after doing that, the flakiness went away. This is also recommended by readability reviewers.

renkelvin: Good catch. fixed.

GoogleUtilities/Common/GULLoggerCodes.h
renkelvin: Agreed. While this change is just because of formatting. `kGULSwizzlerMessageCodeAppDelegateSwizzlingInvalidAppDelegate = 1014` was there for a while.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
```diff
@@ -78,8 +78,30 @@ + (BOOL)isSceneDelegateProxyEnabled {
   return [GULAppDelegateSwizzler isAppDelegateProxyEnabled];
 }
 
++ (void)proxyOriginalSceneDelegate {
+  if ([GULAppEnvironmentUtil isAppExtension]) {
+    return;
+  }
+
 #if UISCENE_SUPPORTED
```
baolocdo: should this check be at line 82 instead? If there is no support for UIScene, the whole method should be a no-op. Otherwise, it will check for the app extension and it can take a few processing cycles.

GoogleUtilities/SceneDelegateSwizzler/GULSceneDelegateSwizzler.m
```diff
@@ -78,8 +78,30 @@ + (BOOL)isSceneDelegateProxyEnabled {
...
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    if (@available(iOS 13.0, tvOS 13.0, *)) {
+      if (![GULSceneDelegateSwizzler isSceneDelegateProxyEnabled]) {
+        return;
+      } else {
```
baolocdo: remove else since it already returns early

PR #4533, opened 2019-12-19
IID: version dependent on FirebaseInstallations

maksymmalyhin:
The contains:
- InstanceID version that uses FIS iOS SDK to generate store and retrieve IID
- related FCM tests changes

The changes in the branch have been already reviewed in the PRs into the branch, but a brief review is appreciated anyway.


Firebase/InstanceID/FIRInstanceIDKeychain.m
```diff
@@ -16,14 +16,10 @@
 
 #import "FIRInstanceIDKeychain.h"
```
chliangGoogle: This might be some thing I can help in another PR but I wonder if FIRInstanceIDKeychain is needed at all after FIS is in.

maksymmalyhin: As far as I can see it is still used by `FIRInstanceIDAuthKeyChain` which is used in the `FIRInstanceIDTokenManager`, `FIRInstanceIDCheckinStore` and couple more classes.

chliangGoogle: I see, so it's used to store token and checkin, not just IID. We can deal with that later then. Thanks!

Example/InstanceID/Tests/FIRInstanceIDTokenManagerTest.m
```diff
@@ -56,23 +59,23 @@ - (BOOL)checkTokenRefreshPolicyForIID:(NSString *)IID;
 - (void)updateToAPNSDeviceToken:(NSData *)deviceToken isSandbox:(BOOL)isSandbox;
 /**
  *  Create a fetch operation. This method can be stubbed to return a particular operation instance,
- *  which makes it easier to unit test different behaviors.
+ *  which makes it easier to unit test different behaviours.
```
paulb777: Hmm, an American to Canadian change.  @ryanwilson Do we have any standard guidance about English dialects in comments?

Firebase/InstanceID/FIRInstanceIDTokenManager.h
```diff
@@ -46,7 +45,7 @@ typedef NS_OPTIONS(NSUInteger, FIRInstanceIDInvalidTokenReason) {
  *
  *  @param authorizedEntity The authorized entity for the token, should not be nil.
  *  @param scope            The scope for the token, should not be nil.
- *  @param keyPair          The keyPair that represents the app identity.
+ *  @param instanceID   The unique string identifying the app instance.
```
paulb777: inconsistent indent

Firebase/InstanceID/FIRInstanceIDTokenManager.h
```diff
@@ -77,7 +76,7 @@ typedef NS_OPTIONS(NSUInteger, FIRInstanceIDInvalidTokenReason) {
...
```
paulb777: same here

Example/InstanceID/Tests/FIRInstanceIDTokenManagerTest.m
maksymmalyhin: I can revert it. I just followed Xcode suggestions

maksymmalyhin:
@paulb777 Thank you for the review! I pushed the fixes.

maksymmalyhin:
@paulb777 Initially I though not to keep the separate commits, though it is probably a good idea if we are fine with several extra commit pushed to the branch without PRs (the recent ones and conflict resolutions from the past). WDYT?

paulb777:
I'm ok either way. Even if you squash, you can still include the PR's from the commit message that GitHub generates automatically.

maksymmalyhin:
I'll squash and merge including the PR's in the commit message.

PR #4534, opened 2019-12-19
Change custom prioritization dictionary to bytes

mikehaney24:
Also fixes a couple of logs

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORStoredEvent.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORStoredEvent.m
@@ -31,7 +31,7 @@
     _target = @(event.target);
     _qosTier = event.qosTier;
     _clockSnapshot = event.clockSnapshot;
-    _customPrioritizationParams = event.customPrioritizationParams;
+    _customBytes = event.customBytes;
   }
   return self;
 }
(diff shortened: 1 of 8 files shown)


mikehaney24:
> What's the reason for changing to bytes?

A couple of things: 
1. It's easier to put bytes into a sqlite db than a dictionary of arbitrary objects.
2. Reduces the chance the NSCoding fails because of the objects inside of them not implementing NSCoding

GoogleDataTransport/GDTCORLibrary/GDTCORStoredEvent.m
```diff
@@ -78,8 +78,7 @@ - (nullable instancetype)initWithCoder:(nonnull NSCoder *)aDecoder {
     NSNumber *qosTier = [aDecoder decodeObjectOfClass:[NSNumber class] forKey:kQosTierKey];
     _qosTier = [qosTier intValue];
     _clockSnapshot = [aDecoder decodeObjectOfClass:[GDTCORClock class] forKey:kClockSnapshotKey];
-    _customPrioritizationParams = [aDecoder decodeObjectOfClass:[NSDictionary class]
-                                                         forKey:kCustomPrioritizationParamsKey];
+    _customBytes = [aDecoder decodeObjectOfClass:[NSDictionary class] forKey:kCustomBytesParamsKey];
```
maksymmalyhin: nit: Looks like we are still rely on `NSCoding` to encode/decode `customBytes`. Is it going to be replaced by JSON? If yes, maybe add JSON to the var name to be more explicit?

mikehaney24: Yeah, NSCoding will go away with the sqlite refactor. Using NSCoding for the bytes is fine because it's already bytes, whereas encoding a dictionary is worse because it will crash if the dictionary contains a value that doesn't implement the NSSecureCoding protocol.

PR #4538, opened 2019-12-21
Fix lint warnings spotted by google3's linter

wilhuff:
I haven't come up with a way to fix the warnings about include order of
pb_encode.h here.


wilhuff:
Agreed. Most of the remaining warnings are caused by differences in required include orderings between the open source linter and google3. Unfortunately, open source considers pb.h and pb_encode.h to be C system files but google3 considers them non-system files. Accommodating both views requires moving the headers from before the C++ system headers to after during copybara. I don't think this is worth it. Another possibility might be to change the way they're included in the open source built to match internal expectations, but I don't know exactly what that would look like.

PR #4545, opened 2019-12-23
style.sh: update clang-format formula in suggestions

maksymmalyhin:
- suggest the brew formula from README

#no-changelog


maksymmalyhin:
@paulb777  Yeah, makes sense. I wonder, do we have a place for such variables already?

PR #4573, opened 2020-01-02
Update Firebase Storage directory structure

paulb777:
See https://github.com/firebase/firebase-ios-sdk/blob/master/AddNewPod.md#directory-structure


morganchen12:
Please update the Danger PR autolabeler as well: https://github.com/firebase/firebase-ios-sdk/blob/master/Dangerfile#L90

paulb777:
Thanks for the push. There was an extra copy of those files. Now the diff should be right.

PR #4579, opened 2020-01-03
Revert "Disable broken Auth Api Tests"

renkelvin:
Reverts firebase/firebase-ios-sdk#4566


renkelvin:
The test failed because there is a user record error in the database. Now it's fixed.

PR #4412, opened 2019-11-27
Add sign in with Apple to the sample app

renkelvin:

PR #4583, opened 2020-01-03
Add apple sign in to sample code

renkelvin:


renkelvin:
Resubmit https://github.com/firebase/firebase-ios-sdk/pull/4412 There is weird build issue. I doubt it's because of the category name of "+Apple"

Example/Auth/Sample/MainViewController+OAuth.m
```diff
@@ -307,6 +318,86 @@ - (void)signInYahooHeadfulLite {
...
+    [controller performRequests];
+  }
+}
+
+- (void)authorizationController:(ASAuthorizationController *)controller didCompleteWithAuthorization:(ASAuthorization *)authorization API_AVAILABLE(ios(13.0)) {
+  ASAuthorizationAppleIDCredential* appleIDCredential = authorization.credential;
+  NSString *idToken = [NSString stringWithUTF8String:[appleIDCredential.identityToken bytes]];
+  FIROAuthCredential *credential = [FIROAuthProvider credentialWithProviderID:@"apple.com"
+                                                                      IDToken:idToken
+                                    // rawNonce:@"REPLACE_ME_WITH_YOUR_RAW_NONCE"
```
morganchen12: This looks like it will cause a build failure. I think you can avoid commenting out these placeholder nonce strings, they're pretty obviously placeholders.

renkelvin: Yeah, good catch!

morganchen12: To clarify, I meant you should leave the parameters in, but not commented out (unless that causes issues).

Will defer to your judgment here though, since I haven't actually built and ran the code.

renkelvin: It's able to build while can't actually sign in. I think leaving the parameters in makes sense.

PR #4584, opened 2020-01-03
Miscellaneous fixes

wilhuff:
See individual commits for descriptions.

The only ~serious issues are:

  * ba75105 Fixes a leaking stream token on every mutation acknowledgement
  * 6a34b23 Fixes leaking ByteStrings; clang wasn't using the by-value assignment operator correctly
  * f5f7980 Fixes leaking GrpcCompletions during shutdown.

With these changes in place, it's possible to get a relatively clean run of Instruments > Leaks during the integration tests, though occasionally the spec tests kick up a bunch of spurious leaks. Given that we're rewriting these in C++ I'm not inclined toward actually addressing this now.

### DIFF
--- a/Firestore/Source/API/FIRDocumentSnapshot.mm
+++ b/Firestore/Source/API/FIRDocumentSnapshot.mm
@@ -267,7 +267,7 @@ ServerTimestampBehavior InternalServerTimestampBehavior(FIRServerTimestampBehavi
 }
 
 - (id)convertedReference:(const FieldValue &)value {
-  auto ref = value.reference_value();
+  const auto &ref = value.reference_value();
   const DatabaseId &refDatabase = ref.database_id();
   const DatabaseId &database = _snapshot.firestore()->database_id();
   if (refDatabase != database) {
--- a/Firestore/Source/API/FIRDocumentSnapshot.mm
+++ b/Firestore/Source/API/FIRDocumentSnapshot.mm
@@ -53,11 +53,13 @@ using firebase::firestore::api::SnapshotMetadata;
 using firebase::firestore::model::DatabaseId;
 using firebase::firestore::model::Document;
 using firebase::firestore::model::DocumentKey;
+using firebase::firestore::model::FieldPath;
 using firebase::firestore::model::FieldValue;
 using firebase::firestore::model::FieldValueOptions;
 using firebase::firestore::model::ObjectValue;
 using firebase::firestore::model::ServerTimestampBehavior;
 using firebase::firestore::nanopb::MakeNSData;
+using firebase::firestore::util::MakeString;
 using firebase::firestore::util::ThrowInvalidArgument;
 
 NS_ASSUME_NONNULL_BEGIN
@@ -180,16 +182,16 @@ ServerTimestampBehavior InternalServerTimestampBehavior(FIRServerTimestampBehavi
 
 - (nullable id)valueForField:(id)field
      serverTimestampBehavior:(FIRServerTimestampBehavior)serverTimestampBehavior {
-  FIRFieldPath *fieldPath;
+  FieldPath fieldPath;
   if ([field isKindOfClass:[NSString class]]) {
-    fieldPath = [FIRFieldPath pathWithDotSeparatedString:field];
+    fieldPath = FieldPath::FromDotSeparatedString(MakeString(field));
   } else if ([field isKindOfClass:[FIRFieldPath class]]) {
-    fieldPath = field;
+    fieldPath = ((FIRFieldPath *)field).internalValue;
   } else {
     ThrowInvalidArgument("Subscript key must be an NSString or FIRFieldPath.");
   }
 
-  absl::optional<FieldValue> fieldValue = _snapshot.GetValue(fieldPath.internalValue);
+  absl::optional<FieldValue> fieldValue = _snapshot.GetValue(fieldPath);
   FieldValueOptions options = [self optionsForServerTimestampBehavior:serverTimestampBehavior];
   return !fieldValue ? nil : [self convertedValue:*fieldValue options:options];
 }
@@ -265,7 +267,7 @@ ServerTimestampBehavior InternalServerTimestampBehavior(FIRServerTimestampBehavi
 }
 
 - (id)convertedReference:(const FieldValue &)value {
-  const auto &ref = value.reference_value();
+  auto ref = value.reference_value();
   const DatabaseId &refDatabase = ref.database_id();
   const DatabaseId &database = _snapshot.firestore()->database_id();
   if (refDatabase != database) {
--- a/Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
+++ b/Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
@@ -71,7 +71,7 @@ ByteString& ByteString::operator=(const ByteString& other) {
   return *this;
 }
 
-ByteString& ByteString::operator=(ByteString&& other) {
+ByteString& ByteString::operator=(ByteString&& other) noexcept {
   if (bytes_ != other.bytes_) {
     std::free(bytes_);
     bytes_ = other.bytes_;
--- a/Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
+++ b/Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
@@ -55,11 +55,29 @@ ByteString::ByteString(const ByteString& other)
 }
 
 ByteString::ByteString(ByteString&& other) noexcept {
-  swap(*this, other);
+  bytes_ = other.bytes_;
+  other.bytes_ = nullptr;
 }
 
 ByteString::~ByteString() {
   std::free(bytes_);
+}
+
+ByteString& ByteString::operator=(const ByteString& other) {
+  if (bytes_ != other.bytes_) {
+    std::free(bytes_);
+    bytes_ = MakeBytesArray(other.data(), other.size());
+  }
+  return *this;
+}
+
+ByteString& ByteString::operator=(ByteString&& other) {
+  if (bytes_ != other.bytes_) {
+    std::free(bytes_);
+    bytes_ = other.bytes_;
+    other.bytes_ = nullptr;
+  }
+  return *this;
 }
 
 /* static */ ByteString ByteString::Take(pb_bytes_array_t* bytes) {
--- a/Firestore/core/src/firebase/firestore/nanopb/byte_string.h
+++ b/Firestore/core/src/firebase/firestore/nanopb/byte_string.h
@@ -80,10 +80,8 @@ class ByteString : public util::Comparable<ByteString> {
 
   ~ByteString();
 
-  ByteString& operator=(ByteString other) {
-    swap(*this, other);
-    return *this;
-  }
+  ByteString& operator=(const ByteString& other);
+  ByteString& operator=(ByteString&& other);
 
   friend void swap(ByteString& lhs, ByteString& rhs) noexcept;
 
--- a/Firestore/core/src/firebase/firestore/nanopb/byte_string.h
+++ b/Firestore/core/src/firebase/firestore/nanopb/byte_string.h
@@ -81,7 +81,7 @@ class ByteString : public util::Comparable<ByteString> {
   ~ByteString();
 
   ByteString& operator=(const ByteString& other);
-  ByteString& operator=(ByteString&& other);
+  ByteString& operator=(ByteString&& other) noexcept;
 
   friend void swap(ByteString& lhs, ByteString& rhs) noexcept;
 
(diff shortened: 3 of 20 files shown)


Firestore/core/src/firebase/firestore/remote/grpc_stream.cc
```diff
@@ -240,9 +240,16 @@ void GrpcStream::FastFinishCompletionsBlocking() {
     completion->Cancel();
   }
 
+  bool queue_shutting_down = worker_queue_->is_shutting_down();
+
   for (auto completion : completions_) {
     // This is blocking.
     completion->WaitUntilOffQueue();
+
+    // If the queue is shutting down it won't allow any completions to fire.
```
var-const: Ultranit: I think there should be a comma after `shutting down`.

Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
```diff
@@ -55,13 +55,31 @@ ByteString::ByteString(const ByteString& other)
...
 
+ByteString& ByteString::operator=(const ByteString& other) {
+  if (bytes_ != other.bytes_) {
+    std::free(bytes_);
+    bytes_ = MakeBytesArray(other.data(), other.size());
+  }
+  return *this;
+}
+
+ByteString& ByteString::operator=(ByteString&& other) {
```
var-const: Nit: make this `noexcept` as well.

Firestore/Source/API/FIRDocumentSnapshot.mm
```diff
@@ -265,7 +267,7 @@ - (id)convertedServerTimestamp:(const FieldValue &)value
 }
 
 - (id)convertedReference:(const FieldValue &)value {
-  const auto &ref = value.reference_value();
+  auto ref = value.reference_value();
```
var-const: Question: what's the reason for this change?

Firestore/core/test/firebase/firestore/nanopb/byte_string_test.cc
```diff
@@ -31,6 +31,22 @@ using testing::ContainerEq;
 
 namespace {
 
+struct free_deleter {
+  template <typename T>
+  void operator()(T* value) const {
+    using non_const_T = typename std::remove_const<T*>::type;
```
var-const: Very optional: Abseil has [`remove_const_t`](https://github.com/abseil/abseil-cpp/blob/master/absl/meta/type_traits.h#L559).

Firestore/core/test/firebase/firestore/nanopb/byte_string_test.cc
```diff
@@ -31,6 +31,22 @@ using testing::ContainerEq;
 
 namespace {
 
+struct free_deleter {
+  template <typename T>
+  void operator()(T* value) const {
```
var-const: A couple of notes on this function:
1. Since the code seems to remove `const` unconditionally, it could probably avoid any metaprogramming:
```cpp
void operator()(const T* value) const {
  std::free(const_cast<T*>(value));
}
```
2. All the pointers used in this file seem to be non-const, is removing `const` actually necessary?

Firestore/core/src/firebase/firestore/remote/grpc_stream.cc
```diff
@@ -240,9 +240,16 @@ void GrpcStream::FastFinishCompletionsBlocking() {
...
+    if (queue_shutting_down) {
```
var-const: Honestly, while this change is certainly the right thing to do, the fact that ownership is kinda vague here seems a little concerning. Do you think this is something we could improve long-term? Perhaps add some sort of a TODO?

Firestore/core/src/firebase/firestore/nanopb/byte_string.h
```diff
@@ -80,10 +80,8 @@ class ByteString : public util::Comparable<ByteString> {
 
   ~ByteString();
 
-  ByteString& operator=(ByteString other) {
```
var-const: Hmm, this is a little weird. IIUC, a function with `operator=(T)` signature should be considered a user-defined copy assignment operator[1], which should inhibit both the copy assignment operator and the move assignment operator from being implicitly defined. Even more confusingly, if the compiler were to implicitly define the copy assignment operator as `operator=(const T&)`, it would [create ambiguity](https://wandbox.org/permlink/9wDn0yThZnntwREQ). In other words, it seems like a compiler defect, and I'm not sure how it works (unless there's another accompanying defect that incorrectly resolves the ambiguity?).

[1] See [cppreference](https://en.cppreference.com/w/cpp/language/copy_assignment):
> A copy assignment operator of class `T` is a non-template non-static member function with the name `operator=` that takes exactly one parameter of type **`T`**, `T&`, `const T&`, `volatile T&`, or `const volatile T&`.

Firestore/core/src/firebase/firestore/remote/grpc_stream.cc
wilhuff: I added a TODO.

The refactoring that's required was more than I wanted to attempt here. The outline of a solution is that since we need to `new`/`delete` for the purposes of sharing with gRPC, `GrpcCompletion` objects should only exist for that purpose and should not otherwise be shared.

Concretely, this means that when the completion is called by gRPC, it should package up the values it wants to pass into the closure it submits onto the AsyncQueue rather than passing a pointer to itself. This would allow it to  unconditionally, immediately delete itself and clear this up.

I didn't make this change because (among other things) I don't entirely understand the relationship between the lifetime of the message and the callback onto our queue. Let's discuss and address in a separate PR.

Firestore/core/src/firebase/firestore/nanopb/byte_string.cc
wilhuff: Done.

Firestore/core/src/firebase/firestore/nanopb/byte_string.h
wilhuff: I completely agree that this *should* work, but I can't argue with the fact that the leak detector is definitely seeing a leak and that this change definitely addresses it. I suspect more investigation in a more controlled environment could pin this down but spending that time now didn't seem worthwhile given the ready availability of a cheap, unobtrusive fix.

Firestore/Source/API/FIRDocumentSnapshot.mm
wilhuff: This is a mistake. It has leaked forward from a change that made `reference_value` return a value.

Reverted.

Firestore/core/test/firebase/firestore/nanopb/byte_string_test.cc
wilhuff: All good thoughts. I started from a stack overflow description of the general solution without thinking too hard about it :-).

I've taken (2) to heart and just cut all the complexity out.

Firestore/core/test/firebase/firestore/nanopb/byte_string_test.cc
wilhuff: Obsolete.

Firestore/core/src/firebase/firestore/remote/grpc_stream.cc
var-const: SGTM.

wilhuff: The ASAN test failures indicate that while the old mode leaked on shutdown, the new mode uses-after-free. Neither are good.

I've reworked this to use `shared_ptr`. This allows either gRPC's hold or `GrpcStream`'s hold to exceed the other without negative consequence. See comments for exactly how this works.

Firestore/core/src/firebase/firestore/remote/grpc_completion.h
```diff
@@ -120,6 +138,37 @@ class GrpcCompletion {
...
+  //
+  // Under normal operation, this works as before: the completion's self-release
+  // just decrements the reference count because GrpcStream still holds a
+  // reference in its completion list. Then, removing the completion
+  // from the list destroys the completion on the worker queue.
+  //
+  // During shutdown, the GrpcStream can now cancel all the completions in the
+  // queue because its shared_ptrs guarantee liveness. It can then release its
+  // shared_ptrs and then gRPC completion (whenever it actually happens) will
+  // actually destroy the GrpcCompletion block.
```
var-const: I think this write-up does a great job of explaining the change; my only concern is that it might be a little too focused on the current (previous after this PR is merged) form of ownership that, in the fullness of time, will be otherwise forgotten. Perhaps consider if rewriting this comment focusing on just the problematic edge case could make it shorter without losing clarity.

Just as a starting point, here's a stab; feel free to modify/rewrite/reject as you see fit:
```
Ownership of a gRPC completion is shared between the Firestore gRPC wrapper object that
initiated the operation (e.g., a `GrpcStream`) and gRPC itself (for as long as the completion
is on the gRPC completion queue). While most of the time a completion gets removed from
the gRPC completion queue first, during shutdown it may be the case that the Firestore gRPC
wrapper is destroyed before the completion comes back from the gRPC completion queue.
To cover that case, a `GrpcCompletion` holds a shared pointer to itself for as long as it's on
the gRPC completion queue (modelling the fact that it's owned by gRPC during that time).
```

Firestore/core/src/firebase/firestore/remote/grpc_completion.cc
```diff
@@ -25,21 +25,32 @@ namespace remote {
...
+    Callback&& callback) {
+  // Construct in two steps to use the private constructor.
+  GrpcCompletion partial(type, worker_queue, std::move(callback));
+  auto completion = std::make_shared<GrpcCompletion>(std::move(partial));
+
+  // Prepare the `GrpcCompletion` for submission to gRPC.
+  //
+  // Note: this is a separate step from the constructor due to limitations in
+  // std::enable_shared_from_this. The internal weak_ptr that makes that work
+  // is is not initialized until after the shared_ptr for this object is
```
var-const: Nit: s/is is/is/.

Firestore/core/src/firebase/firestore/remote/grpc_completion.h
```diff
@@ -120,6 +138,37 @@ class GrpcCompletion {
...
+  // object must not be deleted.
+  //
+  // Previously we managed this by explicitly new/deleting GrpcCompletion but
+  // this creates timing problems during shutdown. During shutdown, the worker
+  // thread wants to cancel RPCs that are in flight, but this races with the
+  // completion event from gRPC.
+  //
+  // This used to work because we only allowed delete on the worker thread.
+  // Now that the AsyncQueue no longer admits tasks after shutdown has been
+  // started this now leaks the completion. If we called delete outside the
```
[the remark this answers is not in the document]
wilhuff: I initially considered this as a possibility but think it's problematic. EnqueueEvenAfterShutdown is a hack I'd really like to get rid of and going doing that path would likely cement it into place. I can be convinced otherwise, but this solution is viable, and changing this again/reviewing again will make us miss the release window, which I don't want to do.

Firestore/core/src/firebase/firestore/remote/grpc_completion.h
```diff
@@ -120,6 +138,37 @@ class GrpcCompletion {
...
+  // worker thread this could cause cancellation to use-after-free so that
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/core/src/firebase/firestore/remote/grpc_completion.h
```diff
@@ -120,6 +138,37 @@ class GrpcCompletion {
...
+  // doesn't work either.
+  //
+  // This field is an intentional retain cycle: while gRPC holds the
+  // GrpcCompletion this pointer is set and that keeps the raw pointer we give
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/core/src/firebase/firestore/remote/grpc_completion.h
wilhuff: Good idea. Rewrote this based on your suggestion.

Firestore/core/src/firebase/firestore/remote/grpc_completion.cc
```diff
@@ -25,21 +25,32 @@ namespace remote {
...
+    Type type,
+    const std::shared_ptr<util::AsyncQueue>& worker_queue,
+    Callback&& callback) {
+  // Construct in two steps to use the private constructor.
+  GrpcCompletion partial(type, worker_queue, std::move(callback));
+  auto completion = std::make_shared<GrpcCompletion>(std::move(partial));
+
+  // Prepare the `GrpcCompletion` for submission to gRPC.
+  //
+  // Note: this is a separate step from the constructor due to limitations in
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/core/src/firebase/firestore/remote/grpc_completion.cc
wilhuff: Obsolete.

PR #4587, opened 2020-01-03
Update swift-protobuf dependency

paulb777:
This also fixes the ZipBuilder build warnings.

Verified that a release build matches M61 as expected.

#no-changelog

### DIFF
--- a/ZipBuilder/Package.swift
+++ b/ZipBuilder/Package.swift
@@ -26,6 +26,8 @@ let package = Package(
     .executable(name: "UpdateFirebasePod", targets: ["UpdateFirebasePod"]),
   ],
   dependencies: [
+    // Keep the generated protos in sync with the version below.
+    // See https://github.com/firebase/firebase-ios-sdk/tree/master/ZipBuilder#updating-protobuf-generated-swift-files.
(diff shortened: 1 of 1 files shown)


ZipBuilder/Package.swift
```diff
@@ -26,7 +26,7 @@ let package = Package(
     .executable(name: "UpdateFirebasePod", targets: ["UpdateFirebasePod"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-protobuf.git", .exact("1.2.0")),
+    .package(url: "https://github.com/apple/swift-protobuf.git", .exact("1.7.0")),
```
ryanwilson: There's a recommendation that the protobufs used are generated with the same version of `swift-protobuf` tool. Do you mind checking to make sure there are no source changes?

(I realize now this should be in a comment above the dependency)

paulb777: Updated comment. 

It turns out that the last proto update in #4454 was already with 1.7.0:
`Warning: swift-protobuf 1.7.0 is already installed and up-to-date`

ryanwilson: Ahh got it, thanks!

PR #4588, opened 2020-01-03
Remove Firebase xcode project

paulb777:
Everything here should now be developed, built and tested with `pod gen` and `pod lib lint`

Removing now to make migration to the standard directory structure easier.

Note, that this removes the currently broken coverage collection mechanism.  Coverage will need to be addressed separately - ideally in conjunction with `pod lib lint`.

#no-changelog


scripts/if_changed.sh
```diff
@@ -45,6 +45,7 @@ elif [[ -z "$TRAVIS_COMMIT_RANGE" ]]; then
 
 else
   case "$PROJECT-$METHOD" in
+
```
ryanwilson: Should this be removed? Maybe was an accidental addition

paulb777: Done.

paulb777:
The final travis build (after comment change) was cancelled. The previous one succeeded.

PR #4593, opened 2020-01-04
Build zip in GHA cron

paulb777:
Build zip from GitHub Actions (takes 64 minutes - https://github.com/firebase/firebase-ios-sdk/runs/373751913)

Fix IID and Installations for local zip builids

It looks like the cron part of this can't be tested until this merges to master based on https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#scheduled-events-schedule

#no-changelog


paulb777:
17 minutes to build a single zip architecture on GHA versus 30 on travis!

Now building all architectures and enabling a pod repo update ....

PR #4598, opened 2020-01-06
Revert "FitebaseInstallations: fix create FID race condition"

maksymmalyhin:
Reverts firebase/firebase-ios-sdk#4576

The PR is to be reverted because:
- the test introduced in #4576 was incorrect and didn't reflect a real world scenario. With actual usage this type of race condition does not occur. 
- the fix introduces another issue(b/147230513) .

### DIFF
--- a/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
+++ b/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
@@ -177,7 +177,6 @@
   // 5.5. Verify registered installation was saved.
   OCMVerifyAll(self.mockInstallationsStore);
   OCMVerifyAll(self.mockIIDStore);
-  OCMVerifyAll(self.mockAPIService);
 }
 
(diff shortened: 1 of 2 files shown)


FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
```diff
@@ -334,17 +333,11 @@ - (void)testGetInstallationItem_WhenThereIsIIDAndNoFID_ThenFIDIsCreatedAndRegist
...
   OCMExpect([self.mockInstallationsStore installationForAppID:self.appID appName:self.appName])
       .andReturn(pendingStorePromise);
 
-  // 2. Expect Create FID API request to be sent once.
-  FBLPromise<FIRInstallationsItem *> *registrationErrorPromise = [FBLPromise pendingPromise];
-  [registrationErrorPromise reject:[FIRInstallationsErrorUtil APIErrorWithHTTPCode:400]];
-  OCMExpect([self.mockAPIService registerInstallation:storedInstallation1])
-      .andReturn(registrationErrorPromise);
-
   // 3. Request installation n times
```
ryanwilson: Nit: this number 3 should now be a 2.

Issue #4599, opened 2020-01-06
FR: Add FIRStorage Interop


leisurehound:
### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 11.3
  * Firebase SDK version: 6.14.0
  * Firebase Component: Storage
  * Component version: 3.4.3

### [REQUIRED] Step 3: Describe the problem

In attempting to build a SwiftLog implementation that posts client logs to Firebase Cloud Storage buckets as a Firebase Extended library, the SwiftLog impl will depend on Firebase Cloud Storage, however it should only depend on the storage object configured by Client App and not create its own object.  As suggested by Ryan Wilson, adding a FIRStorageInterop is the right solution to manage the dependency cleanly.

#### Steps to reproduce:
n/a

#### Relevant Code:
n/a see forthcoming PR

paulb777:
@leisurehound Is this still needed/planned?

leisurehound:
So, I've implemented around the need for this, tho this pushes more cognitive load on the client to use the solution.  I'm about to release the library on GitHub next week, so can we look at adoption and/or pushback on whether a more seamless solution to the client developer is warranted?

paulb777:
So you mean leave this open pending feedback on next week's release?

leisurehound:
yes, if that is possible.

paulb777:
SG - let me or @ryanwilson know if you'd like to come back to this. There have been some Interop structural changes since this PR was created.

leisurehound:
hmm...maybe it would be better then to just start a new PR if the need warrants it.

paulb777:
Perhaps. Maybe close #4600 and leave this feature request open?

leisurehound:
ah, yes, we can close that PR.  Will do so now.

paulb777:
Closing for staleness. Can reopen if the need re-arises.

PR #4600, opened 2020-01-06
Impl of FIRStorageInterop for SwiftLog impl using Firebase Cloud Storage as log storage backend.

leisurehound:
https://github.com/firebase/firebase-ios-sdk/issues/4599

For collaboration with wilsonryan@ per discussion about SwiftLog impl using Firebase Cloud Storage as a log storage backend.

### DIFF
--- a/FirebaseStorage/Sources/FIRStorage.m
+++ b/FirebaseStorage/Sources/FIRStorage.m
@@ -36,13 +36,51 @@ static NSMutableDictionary<
     NSMutableDictionary<NSString * /* bucket */, GTMSessionFetcherService *> *> *_fetcherServiceMap;
 static GTMSessionFetcherRetryBlock _retryWhenOffline;
 
-@interface FIRStorage () {
+//Should this be in FIRStorage_Private.h instead?
+@protocol FIRStorageInstanceProvider
+@end
+
+@interface FIRStorage () <FIRStorageInstanceProvider, FIRLibrary> {
   /// Stored Auth reference, if it exists. This needs to be stored for `copyWithZone:`.
   id<FIRAuthInterop> _Nullable _auth;
 }
 @end
 
 @implementation FIRStorage
+
++ (void)load {
+  // Register with Core as a library.  Adding FIRStorage as an Interop library enables
+  // proposed FirebaseExtended library SwiftLogForFireCloudStorage to depend on FIRStorage
+  // without making an explicit dependency of FIRStorage by SwiftLogForFireCloudStorage
+  [FIRApp registerInternalLibrary:(Class<FIRLibrary>)self
+                         withName:@"fire-storage"
+                      withVersion:[NSString stringWithUTF8String:FIRStorageVersionString]];
+}
+
+#pragma mark - Interoperability
+/// The components to register with Core.
++ (NSArray<FIRComponent *> *)componentsToRegister {
+  // Provide a component that will return an instance of `FIRStorage`.
+  FIRComponentCreationBlock storageCreationBlock =
+      ^id _Nullable(FIRComponentContainer *container, BOOL *isCacheable) {
+        // Cache so the same `FIRStorage` instance is returned each time.
+        *isCacheable = YES;
+        return [[FIRStorage alloc] initWithApp:container.app
+                                        bucket:@""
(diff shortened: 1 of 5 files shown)


FirebaseStorage/Sources/FIRStorage.m
```diff
@@ -36,14 +36,59 @@
...
+  // proposed FirebaseExtended library SwiftLogForFireCloudStorage to depend on FIRStorage
+  // without making an explicit dependency of FIRStorage by SwiftLogForFireCloudStorage
+  [FIRApp registerInternalLibrary:(Class<FIRLibrary>)self
+                         withName:@"fire-storage"
+                      withVersion:[NSString stringWithUTF8String:FIRStorageVersionString]];
+}
+
+#pragma mark - Interoperability
+/// The components to register with Core.
++ (NSArray<FIRComponent *> *)componentsToRegister {
```
ryanwilson: Storage already registers components here: https://github.com/firebase/firebase-ios-sdk/blob/d261cdeb9bec45317c2d5086336e3a3b5f33e651/FirebaseStorage/Sources/FIRStorageComponent.m#L62

FirebaseStorage/Sources/FIRStorage.m
```diff
@@ -36,14 +36,59 @@
...
 }
 @end
 
 @implementation FIRStorage
 
++ (void)load {
+  // Register with Core as a library.  Adding FIRStorage as an Interop library enables
+  // proposed FirebaseExtended library SwiftLogForFireCloudStorage to depend on FIRStorage
+  // without making an explicit dependency of FIRStorage by SwiftLogForFireCloudStorage
+  [FIRApp registerInternalLibrary:(Class<FIRLibrary>)self
```
ryanwilson: Storage already registers as a library here: https://github.com/firebase/firebase-ios-sdk/blob/d261cdeb9bec45317c2d5086336e3a3b5f33e651/FirebaseStorage/Sources/FIRStorageComponent.m#L55

So this can be removed.

FirebaseStorage/Sources/FIRStorage.m
```diff
@@ -36,14 +36,59 @@
...
+                                        bucket:@""
+                                          auth:nil];
+      };
+  FIRComponent *storageInterop =
+      [FIRComponent componentWithProtocol:@protocol(FIRStorageInterop)
+                            creationBlock:storageCreationBlock];
+  return @[storageInterop];
+}
+
+//Is this truly unique from storageForApp?
```
ryanwilson: Yeah this can go away, I think the instructions need to be updated with `For` instead of `With` since that's what the API actually is.

Thanks for catching this.

FirebaseStorage/Sources/FIRStorage_Private.h
```diff
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
+#import <FirebaseStorageInterop/FIRStorageInterop.h>
```
ryanwilson: I think in the case we're trying to solve here this needs to be in the public header, otherwise the type system won't be able to see if the Storage instance actually conforms to that protocol or not. I *could* be wrong here and please speak up if you think it can be private :)

FirebaseStorage/Sources/FIRStorage.m
```diff
@@ -36,14 +36,59 @@
     NSMutableDictionary<NSString * /* bucket */, GTMSessionFetcherService *> *> *_fetcherServiceMap;
 static GTMSessionFetcherRetryBlock _retryWhenOffline;
 
-@interface FIRStorage () {
+//Should this be in FIRStorage_Private.h instead?
+@protocol FIRStorageInstanceProvider
```
ryanwilson: This doesn't need to be here since there's already the `FIRStorageComponent` class taking care of this.

FirebaseStorage/Sources/FIRStorage.m
ryanwilson: https://github.com/firebase/firebase-ios-sdk/pull/4666

leisurehound:
Abandoning this PR as a less invasive impl was determined which incurred a small cost on the library end user, and the interop has changed much since this PR started.

PR #4608, opened 2020-01-07
Fix wrong Core version for M61

ryanwilson:


ryanwilson:
Merging before CI finishes since it's a no-op.

PR #4614, opened 2020-01-07
Have check.sh fallback to checking everything

wilhuff:
Sometimes PRs have been cut from so long ago that there's no merge base
with master in the shallow clone that Travis makes. In this case, fall
back on just checking everything.

This specifically is meant to fix #4602 which is currently failing because of this problem.

/cc @wu-hui

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -84,6 +84,7 @@ cd "${top_dir}"
 
 ALLOW_DIRTY=false
 COMMIT_METHOD="none"
+CHECK_DIFF=false
 START_REVISION="origin/master"
 TEST_ONLY=false
 VERBOSE=false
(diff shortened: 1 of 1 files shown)


scripts/check.sh
```diff
@@ -187,9 +191,24 @@ if [[ "${START_REVISION}" == *..* ]]; then
...
-  NEW_RANGE_START=$(git merge-base origin/master "${RANGE_END}")
-  START_REVISION="${START_REVISION/$RANGE_START/$NEW_RANGE_START}"
-  START_SHA="${START_REVISION}"
+  # Try to come up with a more accurate representation of the merge, so that
+  # checks will operate on just the differences the PR would merge into master.
+  # The start of the revision range that Travis supplies can sometimes be a
+  # seemingly random value.
+  NEW_RANGE_START=$(git merge-base origin/master "${RANGE_END}" || echo "")
+  if [[ -n "$NEW_RANGE_START" ]]; then
+    START_REVISION="${NEW_RANGE_START}..${RANGE_END}"
```
paulb777: Why is the then code different from the old version?

wilhuff: This more directly expresses the intent: previously this was substituting the new start into the old range string. This is now just constructing a new range from the new start and the old end.

PR #4616, opened 2020-01-07
Release LimitToLast

wu-hui:
Merge #4413 #4491 #4505 #4512 to master.

Only change not included in above PRs is `CHANGELOG.md`.


Firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+- [feature] Added `FIRQuery.limitedToLast(_:limit)`, which returns the last
```
wilhuff: The swift name for this is `limit(toLast:)` and the Objective-C is `queryLimitedToLast:`. We seem to be preferring the Swift names, so looking below it seems the most consistent way to spell this would be:

> Added a `limit(toLast:)` query operator, which returns the last matching documents up to the given limit.

?

wu-hui: Done.

PR #4618, opened 2020-01-07
Enable Index-Free queries

schmidt-sebastian:


Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Unreleased
 
+# v1.9.0
```
wilhuff: This should be in the "Unreleased" section.

schmidt-sebastian: Done. I was trying to suggest the version number. It is now under "Unreleased (v1.10.0)"

PR #4610, opened 2020-01-07
Open Sourcing FirebaseCrashlytics

samedson:
Co-authored-by: Jason Hu <jasonhu@google.com>
Co-authored-by: Bryan Klimt <klimt@google.com>

Fixes #4316

PR #4624, opened 2020-01-08
Crashlytics allow for both types of Promise imports

samedson:
In response to a tip from @maksymmalyhin https://github.com/firebase/firebase-ios-sdk/pull/4610#issuecomment-572124334, allow for both types of imports depending on what cocoapods configuration is being used.

### DIFF
--- a/Crashlytics/Crashlytics/Models/FIRCLSSettings.h
+++ b/Crashlytics/Crashlytics/Models/FIRCLSSettings.h
@@ -14,7 +14,11 @@
 
 #import <Foundation/Foundation.h>
 
-#import "FBLPromise.h"
+#if __has_include(<FBLPromises/FBLPromises.h>)
+#import <FBLPromises/FBLPromises.h>
+#else
+#import "FBLPromises.h"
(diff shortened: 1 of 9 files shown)


ryanwilson:
Can you please link to the context as to why this change was made in the description of the PR?

Crashlytics/Crashlytics/Models/FIRCLSSettings.h
```diff
@@ -14,7 +14,11 @@
 
 #import <Foundation/Foundation.h>
 
-#import "FBLPromise.h"
+#if __has_include(<FBLPromises/FBLPromises.h>)
+#import <FBLPromises/FBLPromises.h>
+#else
+#import "FBLPromises.h"
```
ryanwilson: This changed from `FBLPromise` to `FBLPromises` - is that intentional? Same as below.

samedson: Yeah we're being consistent and using the umbrella header in all cases

PR #4625, opened 2020-01-08
FirebaseMessaging changelog: FirebaseInstallations

maksymmalyhin:
- mention a new FIS dependency with the API Key restriction details
- FIS changelog header format consistent with [FirebaseCore](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/CHANGELOG.md)

### DIFF
--- a/Firebase/Messaging/CHANGELOG.md
+++ b/Firebase/Messaging/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 2020-01 -- v 4.2.0
 - [added] Added watchOS support for Firebase Messaging. This enables FCM push notification function on watch only app or independent watch app. (#4016)
+- [added] Added a new dependency [Firebase Installations SDK](../../FirebaseInstallations/CHANGELOG.md). The Firebase Installations SDK introduces the [Firebase Installations API](https://pantheon.corp.google.com/apis/library/firebaseinstallations.googleapis.com). Developers that use API-restrictions for their API-Key may experience blocked requests (https://stackoverflow.com/questions/58495985/). The solution is available [here](../../FirebaseInstallations/API_KEY_RESTRICTIONS.md).
 
 # 2019-12 -- v4.1.10
 - [fixed] Fix component startup time. (#4137)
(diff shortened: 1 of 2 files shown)


Firebase/Messaging/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 2020-01 -- v 4.2.0
 - [added] Added watchOS support for Firebase Messaging. This enables FCM push notification function on watch only app or independent watch app. (#4016)
+- [added] Added a new dependency [Firebase Installations SDK](../../FirebaseInstallations/CHANGELOG.md). The Firebase Installations SDK introduces the [Firebase Installations API](https://pantheon.corp.google.com/apis/library/firebaseinstallations.googleapis.com). Developers that use API-restrictions for their API-Key may experience blocked requests (https://stackoverflow.com/questions/58495985/). The solution is available [here](../../FirebaseInstallations/API_KEY_RESTRICTIONS.md).
```
paulb777: Perhaps "transitive dependency" instead of "dependency" since it isn't a direct dependency.

Firebase/Messaging/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 2020-01 -- v 4.2.0
 - [added] Added watchOS support for Firebase Messaging. This enables FCM push notification function on watch only app or independent watch app. (#4016)
+- [added] Added a new transitive dependency [Firebase Installations SDK](../../FirebaseInstallations/CHANGELOG.md). The Firebase Installations SDK introduces the [Firebase Installations API](https://pantheon.corp.google.com/apis/library/firebaseinstallations.googleapis.com). Developers that use API-restrictions for their API-Key may experience blocked requests (https://stackoverflow.com/questions/58495985/). The solution is available [here](../../FirebaseInstallations/API_KEY_RESTRICTIONS.md).
```
chliangGoogle: Are we doing the same for other SDKs as well?

maksymmalyhin: Yes, once we all agree on the text I'll copy it to all IID-dependent SDKs changelogs.

andirayo: How about "API-Keys" instead of "API-Key"?
Maybe developers have multiple apps with different API-Keys?

Instead of "The solution is available" I recommend to say "A solution is available".

FirebaseInstallations/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# 1.0.0
+# v1.0.0 -- M62
```
andirayo: just to double check: the uploaded SDK version will be "i:1.0.0", right?

maksymmalyhin: yes, the version string in the request is "i:1.0.0"

Firebase/Messaging/CHANGELOG.md
andirayo: How about "API-Keys" instead of "API-Key"?
Maybe developers have multiple apps with different API-Keys?

Instead of "The solution is available" I recommend to say "A solution is available".

maksymmalyhin: Applied the changes

PR #4630, opened 2020-01-08
Enable CI for Crashlytics

jasonhu-g:
Update scripts to support running CI for Crashlytics.


paulb777:
May want to move the `-use-libraries` and `-use-modular-headers` tests to a cron job to speed things up and failures should be rare.

See .travis.yml for examples of this.

PR #4636, opened 2020-01-08
Fix `vendor` directory being included in style.sh

ryanwilson:
#no-changelog Just a style script update.


ryanwilson:
Style passed on both Travis and GitHub Actions, cancelled the rest of the Travis build.

PR #4641, opened 2020-01-09
release branch testing

paulb777:
Now that we have the Firebase pod in the open, it should be possible to run CI on the release branch.

Making this PR to verify ....

### DIFF
--- a/Firebase.podspec
+++ b/Firebase.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'Firebase'
-  s.version          = '6.14.0'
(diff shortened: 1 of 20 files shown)


paulb777:
Looks like the Firestore maybe_local_pod mechanism doesn't support release branch testing. See https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Example/Podfile#L86 and https://travis-ci.org/firebase/firebase-ios-sdk/jobs/634539938

Also, FirebasePod testing fails because of needing a future version of FirebaseAnalytics.

PR #4645, opened 2020-01-09
Don't run unit test spec on watchOS.

ryanwilson:
This won't run because the required version of InstanceID doesn't work
on watchOS (and watchOS doesn't have XCTests). This is a blocker for
publishing the pod.

#no-changelog


maksymmalyhin:
@ryanwilson Thank you for fixing it!

PR #4652, opened 2020-01-10
Fix the version parsing to allow for betas.

ryanwilson:


ryanwilson:
Failures are an unrelated flake

PR #4658, opened 2020-01-10
Add verify support

thebrianchen:
Porting from [JS](https://github.com/firebase/firebase-js-sdk/pull/2436/files) and [Android](https://github.com/firebase/firebase-android-sdk/pull/1107/files)

### DIFF
--- a/Firestore/Example/Tests/Integration/FSTTransactionTests.mm
+++ b/Firestore/Example/Tests/Integration/FSTTransactionTests.mm
@@ -391,8 +391,7 @@ TransactionStage get = ^(FIRTransaction *transaction, FIRDocumentReference *doc)
 - (void)testIncrementTransactionally {
   // A barrier to make sure every transaction reaches the same spot.
   dispatch_semaphore_t writeBarrier = dispatch_semaphore_create(0);
-  auto counter_unique = absl::make_unique<std::atomic_int32_t>(0);
-  auto counter = counter_unique.get();
+  auto counter = std::make_shared<std::atomic_int>(0);
 
   FIRFirestore *firestore = [self firestore];
   FIRDocumentReference *doc = [[firestore collectionWithPath:@"counters"] documentWithAutoID];
@@ -437,8 +436,7 @@ TransactionStage get = ^(FIRTransaction *transaction, FIRDocumentReference *doc)
 - (void)testUpdateTransactionally {
   // A barrier to make sure every transaction reaches the same spot.
   dispatch_semaphore_t writeBarrier = dispatch_semaphore_create(0);
-  auto counter_unique = absl::make_unique<std::atomic_int32_t>(0);
-  auto counter = counter_unique.get();
+  auto counter = std::make_shared<std::atomic_int>(0);
 
   FIRFirestore *firestore = [self firestore];
   FIRDocumentReference *doc = [[firestore collectionWithPath:@"counters"] documentWithAutoID];
@@ -459,7 +457,7 @@ TransactionStage get = ^(FIRTransaction *transaction, FIRDocumentReference *doc)
           // Once all of the transactions have read, allow the first write. There should be 3
           // initial transaction runs.
           if (nowStarted == total) {
-            XCTAssertEqual(static_cast<int>(*counter), 3);
+            XCTAssertEqual(counter->load(), 3);
             dispatch_semaphore_signal(writeBarrier);
           }
 
@@ -478,7 +476,7 @@ TransactionStage get = ^(FIRTransaction *transaction, FIRDocumentReference *doc)
 
   [self awaitExpectations];
   // There should be a maximum of 3 retries: once for the 2nd update, and twice for the 3rd update.
-  XCTAssertLessThanOrEqual(static_cast<int>(*counter), 6);
+  XCTAssertLessThanOrEqual(counter->load(), 6);
   // Now all transaction should be completed, so check the result.
   FIRDocumentSnapshot *snapshot = [self readDocumentForRef:doc];
   XCTAssertEqualObjects(snapshot[@"count"], @(5.0 + total));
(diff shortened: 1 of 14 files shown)


Firestore/Example/Tests/Integration/FSTTransactionTests.mm
```diff
@@ -508,10 +485,12 @@ - (void)testUpdateTransactionally {
   XCTAssertEqualObjects(@"yes", snapshot[@"other"]);
 }
 
-- (void)testHandleReadingOneDocAndWritingAnother {
+- (void)testRetriesWhenDocumentThatWasReadWithoutBeingWrittenChanges {
   FIRFirestore *firestore = [self firestore];
   FIRDocumentReference *doc1 = [[firestore collectionWithPath:@"counters"] documentWithAutoID];
   FIRDocumentReference *doc2 = [[firestore collectionWithPath:@"counters"] documentWithAutoID];
+  auto counter_unique = absl::make_unique<std::atomic_int32_t>(0);
```
wilhuff: This doesn't seem like it needs to be dynamically allocated. Also, if you made this a `std::atomic_int` you wouldn't have to cast below.

```c++
std::atomic_int counter_value = 0;
std::atomic_int* counter = &counter_value;
```

Firestore/Example/Tests/Integration/FSTTransactionTests.mm
```diff
@@ -543,21 +523,13 @@ - (void)testHandleReadingOneDocAndWritingAnother {
...
-        // XCTAssertNil(error);
-        // XCTAssertEquals(0, tries);
-        // XCTAssertEqualObjects(@(1234), snapshot[@"count"]);
-        // snapshot = [transaction getDocument:doc2 error:&error];
-        // XCTAssertNil(error);
-        // XCTAssertEqualObjects(@(16), snapshot[@"count"]);
-        XCTAssertNotNil(error);
-        XCTAssertEqual(error.code, FIRFirestoreErrorCodeInvalidArgument);
+        XCTAssertNil(error);
+        XCTAssertEqual((int)(*counter), 2);
```
wilhuff: This cast should be unnecessary.

If you do find a need to cast, prefer C++ casts to C casts. In this case the C++ cast would be `static_cast<int>(*counter)`.

Firestore/core/src/firebase/firestore/model/verify_mutation.cc
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/firebase/firestore/model/verify_mutation.h"
+
+#include <cstdlib>
```
wilhuff: Unused?

Firestore/core/src/firebase/firestore/model/verify_mutation.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <utility>
+
+#include "Firestore/core/src/firebase/firestore/model/document.h"
```
wilhuff: It seems like most of these headers are unused (except hard_assert.h). Please remove.

Firestore/core/src/firebase/firestore/model/verify_mutation.cc
thebrianchen: Deleted unused imports.

Firestore/core/src/firebase/firestore/model/verify_mutation.cc
thebrianchen: Done.

Firestore/Example/Tests/Integration/FSTTransactionTests.mm
thebrianchen: Can you explain the rationale behind not having to dynamically allocate it vs. having to? The other tests all seem to create a unique pointer, and I blindly followed that pattern assuming that was the preferred way to use `atomic_int`.

Changing the declaration to your code snippet results in the following error in line declaring `counter_value`:
```
Copying variable of type 'std::atomic_int' (aka 'atomic<int>') invokes deleted constructor
```

wilhuff: There are a couple of issues here that were worked around in this file in a way that wasn't clear.

At the root of this is a conflict: `std::atomic`s (of any type) can't be copied (their copy constructors have been deleted). This is just a limitation of the standard library. On the other side is that Objective-C blocks copy the values of variables from the outer scope into the block when they're referenced inside the block.

This code is using unique_ptrs and casts to patch this together in a way that works but is ultimately not obvious. It's not obvious to you--someone who doesn't know much about this--and it's not obvious to me--someone who does.

The unique_ptr is unnecessary; it doesn't help anything. `std::atomic<T>` already has a deleted copy constructor so it can't be copied into the block. The blocks don't escape this test's frame so there's no benefit to making this a unique_ptr. Secondarily, even if the block does happen to escape, unique_ptr doesn't actually help: the unique_ptr will automatically delete the managed value as the outer frame closes, at which point you have a dangling pointer to freed memory.

The second issue is that `*counter` has a type of `std::atomic<T>`, but the expansion of the `XCTAssertEqual` causes a copy of that value. Casting to `int` in the assertion argument served a dual purpose: it fixed the `int32_t` vs `int` issue but also forces the conversion to `int` before the copy such that the copy will just be of an `int`, not a `std::atomic_int`. This wasn't obvious to me when I made the suggestion.

The way to refer to the counter value in a way that's more explicit is to call `load()` on it. That's the explicit read API that indicates you're forcing the read. That is, this is the preferable form:

```c++
XCTAssertEqual(counter->load(), 2);
```

... which is clearer and shorter than ...
```c++
XCTAssertEqual(static_cast<int>(*counter), 2);
```

Finally, rather than declaring a separate `counter_unique` or `counter_value` then using the raw address to work around any copies that Objective-C might happen to make, a better path is just to use a `std::shared_ptr` which is directly compatible with Objective-C block semantics. You can declare that like so, and it doesn't require two steps:

```c++
auto counter = std::make_shared<std::atomic_int>(0);
```

thebrianchen: Thanks for the explanation!

PR #4668, opened 2020-01-13
Use GHA paths

paulb777:
GitHub Actions has built in path support to trigger workflows if certain files have changed. See https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths

Starting to use this instead of the if_changed.sh script we use with travis.


.github/workflows/check.yml
```diff
@@ -0,0 +1,15 @@
...
+  check:
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Check
+      run: |
+        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/e3496d9/Formula/clang-format.rb
+        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/7963c3d/Formula/swiftformat.rb
+        pip install flake8
+        ./scripts/check.sh --test-only
```
paulb777: Still need to verify this fully works with GHA

.github/workflows/core.yml
```diff
@@ -0,0 +1,18 @@
+name: core
+
+on:
+  pull_request:
+    paths:
+    - 'FirebaseCore**'
```
ryanwilson: Do we also want to test when GoogleUtilities changes?

.github/workflows/dynamiclinks.yml
```diff
@@ -0,0 +1,20 @@
+name: dynamiclinks
+
+on:
+  pull_request:
+    paths:
+    - 'Firebase/DynamicLinks**'
```
ryanwilson: Should we also test on Core changes?

paulb777:
My preference is defer on triggering on all dependencies until we have more complete tests in place and are better able to assess the time/benefit trade-offs.

PR #4672, opened 2020-01-13
[Draft] M62 - IID rollback

maksymmalyhin:
This is a temporary placeholder in case M62 needs to be rolled back.


FirebaseInstanceID.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'FirebaseInstanceID'
-  s.version          = '4.2.8'
+  s.version          = '4.3.1'
```
paulb777: Changing APIs should be a minor version update

PR #4679, opened 2020-01-14
Put comments directly above enum definition

wilhuff:
This should fix a documentation generation bug that's causing
FIRDocumentChange to appear to be undocumented.

/cc @markarndt


wilhuff:
The test failure is in the Firestore quickstart and I've determined the root cause for the test flakiness. Addressing that separately.

PR #4683, opened 2020-01-15
Firebase Installations: validation of FIROptions parameters

maksymmalyhin:
Do validation of missing `FIROptions` fields at configuration as other Firebase SDKs do:
- throw in case of missing `FIROptions.APIKey`:
  - the parameter is required by Auth, Dynamic links, Firestore, etc.
  - this parameter was not required by the old InstanceID version, so there  is a slight chance there are projects that will have to missing `APIKey` after updating
- throw in case of missing `FIROptions.GCMSenderID`: 
  - this parameter is already validated by Firebase Core
- use `FIROptions.GCMSenderID` instead of  `FIROptions.projectID` (when unavailable)

See #4692 for additional details.

### DIFF
--- a/Example/InstanceID/Tests/FIRInstanceIDTest.m
+++ b/Example/InstanceID/Tests/FIRInstanceIDTest.m
@@ -156,6 +156,8 @@ static NSString *const kGoogleAppID = @"1:123:ios:123abc";
   // The shared instance relies on the default app being configured. Configure it.
   FIROptions *options = [[FIROptions alloc] initWithGoogleAppID:kGoogleAppID
                                                     GCMSenderID:kGCMSenderID];
+  options.APIKey = @"api-key";
+  options.projectID = @"project-id";
   [FIRApp configureWithName:kFIRDefaultAppName options:options];
   FIRInstanceID *instanceID = [FIRInstanceID instanceID];
   XCTAssertNotNil(instanceID);
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
+- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID` or `GCMSenderID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,5 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
-- [changed] [Firebase Installations](../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
-- [changed] [Firebase Installations](../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
+- [changed] [Firebase Installations](../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
-- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID` or `GCMSenderID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
+- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
-- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
+- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
-- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
+- [changed] [Firebase Installations](../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
 - [changed] The following SDKs introduce a new transitive dependency on the [Firebase Installations API](https://console.cloud.google.com/apis/library/firebaseinstallations.googleapis.com):
   - Analytics
   - Cloud Messaging
--- a/FirebaseInstallations/CHANGELOG.md
+++ b/FirebaseInstallations/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v1.1.0 -- M62.1
 
-- [changed] Throw an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID` or `GCMSenderID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/).  (#4683)
+- [changed] Throw an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/).  (#4683)
 
 # v1.0.0 -- M62
 
--- a/FirebaseInstallations/CHANGELOG.md
+++ b/FirebaseInstallations/CHANGELOG.md
@@ -1,6 +1,6 @@
-# v1.0.1 -- M62
+# v1.1.0 -- M62.1
 
-- [changed] Throw an exception when there are missing required `FIROptions` parameters. (#4683)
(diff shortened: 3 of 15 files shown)


FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -117,12 +118,38 @@ - (instancetype)initWithAppOptions:(FIROptions *)appOptions
...
+    [missingFields addObject:@"`FirebaseOptions.APIKey`"];
+  }
+  if (appOptions.projectID.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.projectID`"];
+  }
+  if (appOptions.googleAppID.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.googleAppID`"];
+  }
+  if (appOptions.GCMSenderID.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.googleAppID`"];
```
ryanwilson: This should be `GCMSenderID`

maksymmalyhin: Good catch!

maksymmalyhin:
@ryanwilson I may affect tests of the dependencies. I was sure, all dependent tests are run on Travis, but it looks like it's not the case (e.g. Remote Config tests)

ryanwilson:
> @ryanwilson I may affect tests of the dependencies. I was sure, all dependent tests are run on Travis, but it looks like it's not the case (e.g. Remote Config tests)

Can you please copybara and run a presubmit there as well before submitting? I also worry that Analytics tests will be affected.

maksymmalyhin:
@ryanwilson tap presubmit will not pass until b/147710489 is fixed. I created #4687 to be able to test open source  dependent SDKs on Travis.

maksymmalyhin:
See cl/289843920 for tap presubmit

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
```diff
@@ -59,15 +59,17 @@ NS_SWIFT_NAME(Installations)
...
+ * @throw Throws an exception if the default app is not configured yet or required  `FirebaseApp`
+ * options are missing.
  */
 + (FIRInstallations *)installations NS_SWIFT_NAME(installations());
 
 /**
  * Returns an instance of `Installations` for an application.
  * @param application A configured `FirebaseApp` instance.
  * @return Returns an instance of `Installations` corresponding to the passed application.
+ * @throw Throws an exception if required `FirebaseApp` options are missing.
```
andirayo: as above

Example/InstanceID/Tests/FIRInstanceIDTest.m
```diff
@@ -156,6 +156,8 @@ - (void)testSharedInstance {
   // The shared instance relies on the default app being configured. Configure it.
   FIROptions *options = [[FIROptions alloc] initWithGoogleAppID:kGoogleAppID
                                                     GCMSenderID:kGCMSenderID];
+  options.APIKey = @"api-key";
```
andirayo: out of curiosity: what does the @ symbol do?

FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
```diff
@@ -87,6 +87,46 @@ - (void)tearDown {
...
+      [[FIRInstallationsIDController alloc] initWithGoogleAppID:@"app-id"
+                                                        appName:@"app-name"
+                                                         APIKey:APIKey
+                                                      projectID:@""
+                                                    GCMSenderID:GCMSenderID
+                                                    accessGroup:nil];
+  XCTAssertNotNil(controller);
+
+  OCMVerifyAll(self.mockAPIService);
+}
```
andirayo: I didn't find a test that fails when no GCMSenderID and no projectID are set...
And a test that fails when API-Key is missing.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
```diff
@@ -75,8 +75,12 @@ - (instancetype)initWithGoogleAppID:(NSString *)appID
   FIRSecureStorage *secureStorage = [[FIRSecureStorage alloc] init];
   FIRInstallationsStore *installationsStore =
       [[FIRInstallationsStore alloc] initWithSecureStorage:secureStorage accessGroup:accessGroup];
+
+  // Use `GCMSenderID` as `projectID` is not available.
```
andirayo: Please change to:
"User `GCMSenderID` as project identifier, if `projectID` is not set"
(or "is not available")

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
```diff
@@ -59,15 +59,17 @@ NS_SWIFT_NAME(Installations)
 
 /**
  * Returns a default instance of `Installations`.
- * @return Returns an instance of `Installations` for `FirebaseApp.defaultApp(). Throws an exception
- * if the default app is not configured yet.
+ * @return Returns an instance of `Installations` for `FirebaseApp.defaultApp().
```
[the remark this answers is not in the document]
maksymmalyhin: Usage of `@return` is a bit inconsistent in Firebase iOS SDK, but I can find more without first `Returns`. I'll update it.

cc @ryanwilson

FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
maksymmalyhin: They are [here](https://github.com/firebase/firebase-ios-sdk/pull/4683/files/5a74f2e0b02315a51593f374f1e21f15fc77d687#diff-81c148814f62ebc1039b03a2c28027e2R240)

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
maksymmalyhin: Good catch!

Example/InstanceID/Tests/FIRInstanceIDTest.m
maksymmalyhin: `"a"` is a C string, `@"b"` is [NSString](https://developer.apple.com/documentation/foundation/nsstring?language=objc)

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
```diff
@@ -59,15 +59,17 @@ NS_SWIFT_NAME(Installations)
...
+ * @throw Throws an exception if the default app is not configured yet or required  `FirebaseApp`
```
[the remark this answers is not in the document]
maksymmalyhin: Good recommendation, but I'll probably keep it as is to be consistent with the rest of the AppleDocs in the repo. Objective C exceptions are not handled usually, so the exception type is not so important and is mostly informative for a crash report reader. Because of that we use just an SDK domain for most of exceptions.

@ryanwilson do you have an opinion on it?

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -117,12 +119,40 @@ - (instancetype)initWithAppOptions:(FIROptions *)appOptions
...
+    [missingFields addObject:@"`FirebaseApp.name`"];
+  }
+  if (appOptions.APIKey.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.APIKey`"];
+  }
+  if (appOptions.googleAppID.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.googleAppID`"];
+  }
+  if (appOptions.GCMSenderID.length < 1) {
+    [missingFields addObject:@"`FirebaseOptions.GCMSenderID`"];
```
ryanwilson: Just to confirm - we need both the `googleAppID` *and* `gcmSenderID` right?

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -117,12 +119,40 @@ - (instancetype)initWithAppOptions:(FIROptions *)appOptions
...
+  }
+
+  if (missingFields.count > 0) {
+    [NSException
+         raise:kFirebaseInstallationsErrorDomain
+        format:@"%@[%@] Could not configure Firebase Installations due to invalid FirebaseApp "
+               @"options. The following parameters are nil or empty: %@. If you use "
+               @"GoogleServices-Info.plist please download the most recent version from Firebase "
```
ryanwilson: nit: include the word "the" --> "from the Firebase console"

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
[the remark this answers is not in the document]
ryanwilson: Keeping it like this is fine I think - when rendered it's just a text field so having it start with "Throws" is fine.

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
```diff
@@ -59,15 +59,17 @@ NS_SWIFT_NAME(Installations)
 
 /**
  * Returns a default instance of `Installations`.
- * @return Returns an instance of `Installations` for `FirebaseApp.defaultApp(). Throws an exception
- * if the default app is not configured yet.
+ * @returns An instance of `Installations` for `FirebaseApp.defaultApp().
```
andirayo: I think the keyword is still @return (without trailing "s").

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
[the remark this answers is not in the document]
andirayo: Sounds good, thanks for the explanation.
I quickly googles Objective C documentation and apparently there is no @throw keyword, but rather a @warning keyword. How about using that instead?

FirebaseInstallations/Source/Library/FIRInstallations.m
andirayo: As far as I know, both AppID and SenderID are required parameters to initialize FIROptions.
FIS, however, is interested in AppID, ProjectID, and ApiKey (and not necessarily in SenderID).

andirayo: @Maksym:
Why are you checking for Sender-ID?
We need either Project-ID (preferred) or Sender-ID, but we only need one of the two.

Example/InstanceID/Tests/FIRInstanceIDTest.m
andirayo: cool, thank you!
NS is a bit of a weird acronym for me as a German... 
but I looked it up, NS stands for NextStep....

FirebaseInstallations/Source/Library/FIRInstallations.m
andirayo: I think, I mentioned this elsewhere, but we do NOT need GCMSenderID.
We only need GCMSenderID if projectID is not set.
Thus, I would recommend to change the check to look something like this:

if (appOptions.projectID.length < 1  &&  appOptions.GCMSenderID.length < 1) {
  [...]

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -117,12 +119,40 @@ - (instancetype)initWithAppOptions:(FIROptions *)appOptions
...
+               @"Console. If you configure Firebase in code, please make sure you specify all "
+               @"required parameters.",
```
andirayo: I recommend that we list the required parameters:
"APIKey and projectID"

FirebaseInstallations/Source/Library/FIRInstallations.m
andirayo: We do NOT necessarily need GCMSenderID.
We only need GCMSenderID if projectID is not set.
Thus, I would recommend to change the check to look something like this:

if (appOptions.projectID.length < 1 && appOptions.GCMSenderID.length < 1) {
  [missingFields addObject:@"`FirebaseOptions.projectID`"];

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -101,6 +102,7 @@ - (instancetype)initWithAppOptions:(FIROptions *)appOptions
                  prefetchAuthToken:(BOOL)prefetchAuthToken {
   self = [super init];
   if (self) {
+    [[self class] validateAppOptions:appOptions appName:appName];
```
andirayo: For Android, we realized that we break 648 TapPresubmit tests if we add this check to the initialization of the FIS instance.
If we add this check as first line/first command of methods #getId and #getToken, we only break 82 TapPresubmit tests.
I don't know how the situation is on iOS, but this is something to consider.

FirebaseInstallations/Source/Library/FIRInstallations.m
maksymmalyhin: Well, technically we can omit the check at all because in the current version of Firebase (6.x.x) `GCMSenderID` is required and enforced by Firebase Core, so`GCMSenderID` is always present and the `projectID` keeps being optional. But yeah, it may look a bit confusing, I'll update the check to be more explicit and add comments.

maksymmalyhin: BTW, `GCMSenderID` is also required to retrieve IID from the old storage, but I assume that we are fine to just skip IID migration if for some reasons (e.g. in the future versions) `GCMSenderID` is missing.

FirebaseInstallations/Source/Library/Public/FIRInstallations.h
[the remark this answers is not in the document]
maksymmalyhin: `@throw` renders well for me:

Screen Shot 2020-01-16 at 10 36 33 AM

FirebaseInstallations/Source/Library/FIRInstallations.m
maksymmalyhin: Updated and replied to the similar comment above.

FirebaseInstallations/Source/Library/FIRInstallations.m
maksymmalyhin: The list of the missing parameters will be added instead of `%@` at the string `"...The following parameters are nil or empty: %@..."` from `missingFields`

FirebaseInstallations/Source/Library/FIRInstallations.m
maksymmalyhin: OK for iOS (chatted offline)

FirebaseInstallations/Source/Library/FIRInstallations.m
maksymmalyhin: updated

FirebaseInstallations/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
-# v1.0.1 -- M62
+# v1.1.0 -- M62.1
 
-- [changed] Throw an exception when there are missing required `FIROptions` parameters. (#4683)
+- [changed] Throw an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID` or `GCMSenderID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/).  (#4683)
```
andirayo: nit: Throws  (with trailing "s")

I recommend to remove "GCMSenderID" and would not even mention it.
I think it has more potential to create confusion rather than help the developer understand what is happening.
If the developer changes something, we want him to add Project-ID.

Question: Is it called FirebaseOptions or FIROptions?

FirebaseCore/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
+- [changed] Firebase Installations(../../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID` or `GCMSenderID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downoaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
```
andirayo: downloaded

is missing an "L"

FirebaseInstallations/CHANGELOG.md
andirayo: downloaded

is missing an "L"

maksymmalyhin:
@ryanwilson updated the minor version and changelogs. PTAL.

FirebaseInstallations/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v1.1.0 -- M62.1
+
+- [changed] Throw an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/).  (#4683)
```
ryanwilson: The code as it is now doesn't throw when `projectID ` is unavailable - so I think that should be removed.  

Also: `s/are up to date/is up to date/`

FirebaseCore/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v6.6.0 -- M62
 - [changed] Reorganized directory structure.
+- [changed] [Firebase Installations](../FirebaseInstallations/CHANGELOG.md) throws an exception when there are missing required `FirebaseOptions` parameters (`APIKey`, `googleAppID`, and `projectID`). Please make sure your `GoogleServices-Info.plist` (or `FirebaseOptions` if you configure Firebase in code) are up to date. The file and settings can be downloaded from the [Firebase Console](https://console.firebase.google.com/). (#4683)
```
ryanwilson: I'm not sure this necessarily needs to be in the Core changelog, we can put it in the changelog online without having it mirrored here since it doesn't really belong in Core, we just want to make sure it's at the top level of the release notes online.

maksymmalyhin: SGTM if we can add it to devsite only. I'll remove it then

FirebaseInstallations/CHANGELOG.md
maksymmalyhin: done

maksymmalyhin:
I am going to merge the PR based on passed [build](https://travis-ci.org/firebase/firebase-ios-sdk/builds/637678633?utm_source=github_status&utm_medium=notification) and no code changes after it.

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -191,7 +225,7 @@ + (void)assertCompatibleIIDVersion {
   return;
 #else
   if (![self isIIDVersionCompatible]) {
-    [NSException raise:NSInternalInconsistencyException
+    [NSException raise:kFirebaseInstallationsErrorDomain
```
tiagoalmeida: I am kind of lost with this message and I have no idea what to do here. I have downloaded the info.plist again,  I have checked your other PR, I have checked the release notes, I have also look into the code and still no idea, so this is kind of my last resort 😅.

maksymmalyhin: @tiagoalmeida sounds like you are having issue with the Firebase update. I am sorry to hear if it is so. Would you like to create a ticket with a description of your issue so we can track all related details there and potentially help other people having similar problems?

PR #4687, opened 2020-01-15
if_changed.sh: add InstanceID and Installations to all dependent SDKs

maksymmalyhin:
#no-changelog

### DIFF
--- a/scripts/if_changed.sh
+++ b/scripts/if_changed.sh
@@ -78,7 +78,8 @@ else
       ;;
 
     Crashlytics-*)
-      check_changes '^(FirebaseCore|Firebase/InstanceID|Firebase/Installations|GoogleUtilities|Crashlytics|FirebaseCrashlytics.podspec)'
+      check_changes '^(FirebaseCore|GoogleUtilities|Crashlytics|FirebaseCrashlytics.podspec|Firebase/InstanceID|FirebaseInstanceID.podspec|'\
+'FirebaseInstallations|FirebaseInstallations.podspec)'
       ;;
 
     Database-*)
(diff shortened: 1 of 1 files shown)


scripts/if_changed.sh
```diff
@@ -78,7 +78,8 @@ else
       ;;
 
     Crashlytics-*)
-      check_changes '^(FirebaseCore|Firebase/InstanceID|Firebase/Installations|GoogleUtilities|Crashlytics|FirebaseCrashlytics.podspec)'
+      check_changes '^(FirebaseCore|GoogleUtilities|Crashlytics|FirebaseCrashlytics.podspec|Firebase/InstanceID|FirebaseInstanceID.podspec|'\
+'FirebaseInstallations|FirebaseInstallations.podspec)'
```
ryanwilson: I think `FirebaseInstallations.podspec` will be caught by the `FirebaseInstallations` regex itself, so we should be okay with just that. I think that's why `FirebaseCore` is alone without the `.podspec` too. Same as below.

maksymmalyhin: I think, ideally, we should add podspecs for dependencies as well in order to catch platforms, versions, etc. mismatches when they are changed in a dependency podspec. I feel like handling it in a more generic way separately in #4688 though.

ryanwilson: I agree with handling it a more general way, thanks for filing that. I was just noting that I think the regex itself catches `FirebaseInstallations.podspec` when you only include `FirebaseInstallations` and don't need to explicitly call out the podspec.

maksymmalyhin: Ah, you are right! I'll update it.

PR #4695, opened 2020-01-17
Source code changes for Segmentation SDK.

dmandar:
Source code changes only. Unit test and test app changes to follow. (Splitting original PR into multiple)


FirebaseSegmentation/Sources/FIRSegmentation.m
```diff
@@ -0,0 +1,77 @@
...
+      associateCustomInstallationIdentiferNamed:customInstallationID
+                                    firebaseApp:_firebaseAppName
+                                     completion:^(BOOL success, NSDictionary *result) {
+                                       if (!success) {
+                                         // TODO(dmandar) log; pass along internal error code.
+                                         NSError *error = [NSError
+                                             errorWithDomain:kFirebaseSegmentationErrorDomain
+                                                        code:FIRSegmentationErrorCodeInternal
+                                                    userInfo:result];
+                                         completionHandler(error);
```
GUserx: It would be more safety if you provide a condition that **completionHandler** is not nil otherwise it will cause a crash in addition its marked as nonnull in header file

FirebaseSegmentation/Sources/FIRSegmentation.m
```diff
@@ -0,0 +1,77 @@
...
+                                       } else {
+                                         completionHandler(nil);
```
GUserx: And here as well :)

FirebaseSegmentation/Sources/FIRSegmentation.m
```diff
@@ -0,0 +1,77 @@
...
+                                       }
+                                     }];
+}
+
+/// Designated initializer
+- (instancetype)initWithAppName:(NSString *)appName FIROptions:(FIROptions *)options {
+  self = [super init];
+  if (self) {
+    _firebaseAppName = appName;
```
GUserx: In my opinion `_firebaseAppName = [appName copy]` is more preferable so we can potentially secure firebaseAppName to not get a NSMutableString instance which can cause an undifned behaviour

FirebaseSegmentation/Sources/Private/FIRSegmentationComponent.h
```diff
@@ -0,0 +1,58 @@
...
+@class FIRSegmentation;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Provides and creates instances of Segmentation. Used in the
+/// interop registration process to keep track of Segmentation instances for each `FIRApp` instance.
+@protocol FIRSegmentationProvider
+
+/// Cached instances of Segmentation objects.
+@property(nonatomic, strong) NSMutableDictionary<NSString *, FIRSegmentation *> *instances;
```
GUserx: I would add an additional method like
`- (void)insertSegmentation:(FIRSegmentation *)segmentation withName:(NSString *)name`
so we can get rid of mutable dictionary and make it as simple **NSDictionary** which will make this property more consistency and more encapsulated.
In 
`- (void)insertSegmentation:(FIRSegmentation *)segmentation withName:(NSString *)name` would be something like:
```
synchronized(self) {
 if (name) {
self.mutableInstances[name] = segmentation;
}
}
```

dmandar:
Thanks for reviewing. This PR is abandoned but will incorporate those changes!

PR #4703, opened 2020-01-17
IID: fix flaky tests

maksymmalyhin:
- use mock `FIRInstallations` instead of a real object in `FIRInstanceIDTokenOperationsTest`

### DIFF
--- a/Example/InstanceID/Tests/FIRInstanceIDTokenOperationsTest.m
+++ b/Example/InstanceID/Tests/FIRInstanceIDTokenOperationsTest.m
@@ -17,7 +17,9 @@
 #import <XCTest/XCTest.h>
 
 #import <FirebaseInstanceID/FIRInstanceID.h>
+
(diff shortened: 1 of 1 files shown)


Example/InstanceID/Tests/FIRInstanceIDTokenOperationsTest.m
```diff
@@ -71,12 +79,24 @@ - (void)setUp {
                                                                     store:_mockStore];
   _instanceID = @"instanceID";
 
+  [self stubInstallations];
```
ryanwilson: Nit: can you please add a small comment here saying why it's necessary to stub installations?

maksymmalyhin: done

ryanwilson: Thanks!

PR #4710, opened 2020-01-21
Deprecating FCM Direct Channel API

chliangGoogle:


paulb777:
@chliangGoogle Is this ready for review?

chliangGoogle:
We are still waiting for launchcal but the code is ready for review.

Firebase/Messaging/Public/FIRMessaging.h
```diff
@@ -252,7 +252,8 @@ NS_SWIFT_NAME(MessagingDelegate)
 /// Handle data messages received via FCM direct channel (not via APNS).
 - (void)messaging:(FIRMessaging *)messaging
     didReceiveMessage:(FIRMessagingRemoteMessage *)remoteMessage
-NS_SWIFT_NAME(messaging(_:didReceive:));
+NS_SWIFT_NAME(messaging(_:didReceive:))
+__deprecated_msg("Downstream message in direct channel is deprecated, please use APNS channel.");
```
paulb777: Is there a URL that describes how to transition?

ryanwilson:
It looks like `FIRMessagingMessageStatus` was missed - please deprecate that as well. Also I believe this needs formatting, some of the lines with deprecation messages are quite long.

chliangGoogle:
FIRMessagingMessageStatus is used by FIRMessagingMessageInfo that is used by appDidReceiveMessage: that is not deprecated.
I've updated the API review doc for this change. And run ./script/format on the header file.

Firebase/Messaging/Public/FIRMessaging.h
chliangGoogle: Public doc probably launch at the same time, there's no direct alternative. Dev can use APNS for data message delivery and FireStore, firebase database, remote config for real time updates. But yeah we might be able to tuck in the public doc change for this.

Firebase/Messaging/CHANGELOG.md
```diff
@@ -1,12 +1,13 @@
-# unreleased
+# 2020-02 -- v4.3.0
+- [changed] Deprecated Messaging Direct Channel and upstream API. Please use APNS channel instead for data message delivery. Use FireStore, real time database or other services for real time updates. The deprecated API will be removed in Firebase 7. (#4710)
```
egilmorez: We never really "productized" this feature with capital letters.  Suggest just "Deprecated FCM direct messaging via `shouldEstablishDirectChannel`. Instead, use APNs for downstream message delivery. "

egilmorez: Also suggest separating direct channel and upstream.  Would that work?  Then you'd have also:

- [changed] Deprecated upstream messaging API.  For realtime updates, use Cloud Firestore, Realtime Database, or other services.  The deprecated . . . "

PR #4715, opened 2020-01-21
Use pre-release versions from release manifests.

ryanwilson:
This will allow installing pre-release versions since CocoaPods
complains otherwise. Note: this removes any version validation since
we're hardcoding the expected version as the version to install.

Also add `CustomDebugStringConvertible` support to `VersionedPod` for nicer printing when debugging.

#no-changelog

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
+++ b/ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
@@ -190,7 +190,7 @@ struct ZipBuilder {
     // versions with `alpha` or `beta` in it, we'll need to explicitly specify the version here so
     // CocoaPods installs it properly.
     let prereleases = expectedVersions().filter { _, version in
-      return version.contains("alpha") || version.contains("beta")
+      version.contains("alpha") || version.contains("beta")
     }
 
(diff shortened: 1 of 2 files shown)


ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -185,7 +185,20 @@ struct ZipBuilder {
     // install a subset of pods, like the following line:
     // let inputPods: [String] = ["Firebase", "FirebaseCore", "FirebaseAnalytics", "FirebaseStorage"]
     let inputPods = FirebasePods.allCases.map { $0.rawValue }
-    let podsToInstall = inputPods.map { CocoaPodUtils.VersionedPod(name: $0, version: nil) }
+
+    // Get the expected versions based on the release manifests, if there are any. If there are any
+    // versions with `alpha` or `beta` in it, we'll need to explicitly specify the version here so
+    // CocoaPods installs it properly.
+    let prereleases = expectedVersions().filter { _, version in
+      return version.contains("alpha") || version.contains("beta")
```
paulb777: also "rc"?

ryanwilson: Done, thanks!

PR #4718, opened 2020-01-21
Add compression to CCT

mikehaney24:
Also address a small bug in FLL whereby the compressed data was used no matter what (even if nil because of compression failure).

Currently, no Firebase SDKs are using the CCT backend (except older versions).


mikehaney24:
> How is it tested?

There are automated E2E integration tests sending actual requests to CCT and getting valid responses. I've also manually run the tests and test apps and examined the server's responses (they were fine). This should be a low-risk change because no Firebase SDK currently is sending data to the CCT backend.

Issue #4729, opened 2020-01-22
Restore Crashlytics quickstart testing


paulb777:
Restore Crashlytics quickstart testing in .travis.yml that was disabled in #4730.

paulb777:
Done in the GHA transition.

PR #4730, opened 2020-01-22
Disable Crashlytics quickstart testing

paulb777:
The Crashlytics quickstart test started consistently failing here after https://github.com/firebase/quickstart-ios/pull/871 even though the test doesn't fail in the quickstart repo.

More investigation is required.  In the meantime, this PR disables the test to restore green travis runs.

The QUICKSTART should be reenabled when #4729 is fixed.

Note that the before_install is not necessary as long as Crashlytics doesn't need anything there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -335,19 +335,17 @@ jobs:
 
     - stage: test
       env:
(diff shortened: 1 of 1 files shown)


paulb777:
Both Crashlytics travis builds are green. Merging ...

Issue #4444, opened 2019-12-05
Improve documentation for or remove usage of SHA1 in Instance ID


robimoltewhp:
Hello,

I am using Veracode Tool for code static analysis
https://www.veracode.com
and I am facing one security issue on FirebaseCore component
  
  * Xcode version: 10.3
  * Firebase SDK version: 6.13.0
  * Firebase Component: Core

The issue says:

File  - FIRInstanceIDCheckinService.m: 213

"Description: Standard random number generators do not provide a sufficient amount of entropy when used for security purposes. Attackers can brute force the output of pseudorandom number generators such as rand().
Remediation: If this random number is used where security is a concern, such as generating a session key or session identifier, use a trusted cryptographic random number generator instead. These can be found on the Windows platform in the CryptoAPI or in an open source library such as OpenSSL. In Java, use the SecureRandom object to ensure sufficient entropy."

chliangGoogle:
Thanks for reaching out. So the code you pointed is used only for logging purposes, and there is no risk of impersonation or access to associated data and so using a non-cryptographically secure random number for this isn't a risk.

We are working on to removing logging_ID or better documenting this.

PR #4732, opened 2020-01-23
Stop sending logging ID as it is no longer used in backend

chliangGoogle:
The logging ID also uses simple random number generator that confuse people #4444. Since it's not used, remove it from logging.


Firebase/InstanceID/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [changed] Stop collecting logging ID as it is not used anymore.(#4732)
```
paulb777: Issue number is preferable to PR number here. (#4444)

chliangGoogle: Done.

PR #4738, opened 2020-01-23
GHA cron updates

paulb777:
- Add a cron job for storage to insure it runs at least once a day
- More specific name for the zip builder cron job
- Setup the bundler in a separate script
- Standardize existing GHA yml files

#no-changelog

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -4,15 +4,21 @@ on:
   pull_request:
     paths:
     - 'FirebaseCore**'
+    - '.github/workflows/core.yml'
+  schedule:
+    # Run every day at 1am (PST) - cron uses UTC times
+    - cron:  '0 7 * * *'
 
 jobs:
   core:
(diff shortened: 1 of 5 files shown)


.github/workflows/core.yml
```diff
@@ -4,15 +4,21 @@ on:
   pull_request:
     paths:
     - 'FirebaseCore**'
+    - '.github/workflows/core.yml'
+  schedule:
+    # Run every day at 1am (PST) - cron uses UTC times
+    - cron:  '0 7 * * *'
```
maksymmalyhin: I wonder is there a way to move common parameters to kind of a common parent config?

paulb777: Maybe? I'll investigate in a future PR.

maksymmalyhin: Sure, it's my curiosity mostly and is nice to have but doesn't look required.

Issue #4461, opened 2019-12-09
157239639: dynamic_link_app_open and dynamic_link_first_open events are swapped


pmendesTekever:
I'm tracking events on DebugView and I noticed this:

**dynamic_link_app_open** is sent when the user opens the app for the first time through a DL.
**dynamic_link_first_open** is sent when the user opens the app through a DL (if it already had the app installed)

I debugged the issue and found that on **FIRDynamicLinks.m**, isFirstOpen is the following:

`BOOL isFirstOpen = ![_userDefaults boolForKey:kFIRDLReadDeepLinkAfterInstallKey];
`

and it should be:
`BOOL isFirstOpen = [_userDefaults boolForKey:kFIRDLReadDeepLinkAfterInstallKey]; `

Is there anything regarding this?
Thank you.

dmandar:
Hi, the code seems ok as is. kFIRDLReadDeepLinkAfterInstallKey is created upon install and is present thereafter. Hence, isFirstOpen will be true only when the key is absent (the first app open). Please reopen if you are still seeing issues with this behavior.

pmendesTekever:
Exactly, **isFirstOpen** should be **false** for users that have already used the app.
Every time my users open the app through a DL, firebase sends a **dynamic_link_first_open**.

![Screenshot 2020-01-31 at 14 29 25](https://user-images.githubusercontent.com/57406749/73547095-336e2280-4436-11ea-9189-b0dd25a199ef.png)

morganchen12:
The code snippet is fine, but it looks like something else is causing the dynamic links SDK to send a dynamic_link_first_open multiple times for the same app. `isFirstOpen` should be set to false if user defaults returns true, so either the bool is not being set or it's being set to false somewhere else after being set to true.

@dmandar, can you take another look?

morganchen12:
@rizafran, can you try to repro this issue? If it does repro, can you share the repro project here?

rizafran:
Noted on this, @morganchen12.

rizafran:
[...]

PR #4743, opened 2020-01-24
Fixing Dynamic Links open events.

pmendesTekever:
I'm tracking events on DebugView and I noticed this:

dynamic_link_app_open is sent when the user opens the app for the first time through a DL.
dynamic_link_first_open is sent when the user opens the app through a DL (if it already had the app installed)

This has beed discussed here:
[](https://github.com/firebase/firebase-ios-sdk/issues/4461)


paulb777:
@pmendesTekever Thanks for the PR. We'll need the CLA signed to take a look ...

paulb777:
Closing, since no CLA

Issue #4746, opened 2020-01-24
Crashlytics symbol collision issues


paulb777:
Crashlytics has several duplicate symbols in its library.

To reproduce build failures, add `-all_load` to `OTHER_LD_FLAGS` or add FirebaseCrashlytics to https://github.com/firebase/firebase-ios-sdk/blob/master/SymbolCollisionTest/Podfile.

See
```
    /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/SymbolCollisionTest-diifsqextkztahhhjcasjpsqkgwb/Build/Products/Debug-iphonesimulator/FirebaseInAppMessaging/libFirebaseInAppMessaging.a(FIRIAMTimeFetcher.o)
duplicate symbol '_CLSReportBinaryImageFile' in:
    /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/SymbolCollisionTest-diifsqextkztahhhjcasjpsqkgwb/Build/Products/Debug-iphonesimulator/FirebaseCrashlytics/libFirebaseCrashlytics.a(FIRCLSInternalReport.o)
    /Users/paulbeusterien/gh/firebase-ios-sdk/SymbolCollisionTest/Pods/Crashlytics/iOS/Crashlytics.framework/Crashlytics(CLSInternalReport.o)
duplicate symbol '_CLSReportCustomExceptionAFile' in:
    /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/SymbolCollisionTest-diifsqextkztahhhjcasjpsqkgwb/Build/Products/Debug-iphonesimulator/FirebaseCrashlytics/libFirebaseCrashlytics.a(FIRCLSInternalReport.o)
    /Users/paulbeusterien/gh/firebase-ios-sdk/SymbolCollisionTest/Pods/Crashlytics/iOS/Crashlytics.framework/Crashlytics(CLSInternalReport.o)
duplicate symbol '_CLSReportCustomExceptionBFile' in:
    /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/SymbolCollisionTest-diifsqextkztahhhjcasjpsqkgwb/Build/Products/Debug-iphonesimulator/FirebaseCrashlytics/libFirebaseCrashlytics.a(FIRCLSInternalReport.o)
[...]

PR #4753, opened 2020-01-27
Crashlytics resolve symbol collisions with old SDK

samedson:
- Resolves symbol collisions https://github.com/firebase/firebase-ios-sdk/issues/4746

Old and New SDK are not compatible in the same app, but symbol collisions are not a great way to discover that. Going to add a warning when both SDKs are included.

### DIFF
--- a/SymbolCollisionTest/Podfile
+++ b/SymbolCollisionTest/Podfile
@@ -12,7 +12,7 @@ target 'SymbolCollisionTest' do
     pod 'FirebaseAnalytics'
     pod 'FirebaseAuth', :path => '../'
     pod 'FirebaseCore', :path => '../'
-#    pod 'FirebaseCrashlytics', :path => '../'  #4746
+    pod 'FirebaseCrashlytics', :path => '../'  #4746
     pod 'FirebaseDatabase', :path => '../'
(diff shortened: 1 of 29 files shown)


paulb777:
Please uncomment https://github.com/firebase/firebase-ios-sdk/blob/master/SymbolCollisionTest/Podfile#L15

SymbolCollisionTest/Podfile
```diff
@@ -12,7 +12,7 @@ target 'SymbolCollisionTest' do
     pod 'FirebaseAnalytics'
     pod 'FirebaseAuth', :path => '../'
     pod 'FirebaseCore', :path => '../'
-#    pod 'FirebaseCrashlytics', :path => '../'  #4746
+    pod 'FirebaseCrashlytics', :path => '../'  #4746
```
paulb777: delete issue number comment. 

Note the SymbolCollisionTest only runs on the travis cron. To test locally in a PR, see https://github.com/firebase/firebase-ios-sdk/tree/master/SymbolCollisionTest#run-locally

samedson: Screen Shot 2020-01-28 at 10 44 54 AM

Looks good now

PR #4755, opened 2020-01-28
Crashlytics add warning about including new and old SDK

samedson:
Including multiple crash reporters tends to cause issues with registering exception handlers. Making this a warning so that we don't break builds. Putting in the public header so that zip install / Carthage users get the same warning.

Related to https://github.com/firebase/firebase-ios-sdk/pull/4753


samedson:
@ryanwilson gonna leave that statement off for now. We're still treating the `Crashlytics` SDK as non-deprecated until `FirebaseCrashlytics` is out of beta

PR #4760, opened 2020-01-28
Create ZipBuilder CI to compile tooling

ryanwilson:
This doesn't test, but does ensure that the zip tooling actually builds.

#no-changelog


ryanwilson:
From an earlier build, here's an uploaded artifact :)

zip_built

This *should* only build the zip during cron jobs, but will have to verify tomorrow to make sure it still ran as expected.

.github/workflows/zip.yml
```diff
@@ -9,8 +13,29 @@ jobs:
   build:
     runs-on: macOS-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
paulb777: What does v2 mean versus v1?  I haven't been able to find this in the docs?

.github/workflows/zip.yml
```diff
@@ -9,8 +13,29 @@ jobs:
...
+    - name: Build
+      run: |
+        cd ZipBuilder
+        swift build -v
+
+  package:
+    needs: build
+    runs-on: macOS-latest
+    # Only run the `package` job for the cron job, not PRs.
+    if: github.event_name == 'schedule'
```
paulb777: Nice find!

.github/workflows/zip.yml
```diff
@@ -9,8 +13,29 @@ jobs:
...
     - name: ZipBuildingTest
-      run: ./scripts/build_zip.sh
+      run: |
+         mkdir -p zip_output_dir
+         sh -x scripts/build_zip.sh zip_output_dir
+
+    - uses: actions/upload-artifact@v1
+      with:
+        name: Firebase-actions-dir
+        # Zip the entire output directory since the builder adds subdirectories we don't know the
```
paulb777: Isn't this the directory the zip file is copied into?

.github/workflows/zip.yml
ryanwilson: It's now the default for starting a new action - looks like Dec 13, 2019 it was released: https://github.com/actions/checkout/releases/tag/v2.0.0 

We should probably update the rest of the actions, but should do so in a separate PR.

.github/workflows/zip.yml
ryanwilson: Thanks! Hopefully it works... got this from the `if` syntax:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idif

and the "`github` Context" https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context

hoping that the name matches with the event trigger here:

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#scheduled-events-schedule

.github/workflows/zip.yml
ryanwilson: Not the case due to the Kokoro setup - the output dir is the root directory and artifacts get packaged in this structure:

```
# Zip is named `Firebase-latest-manual.zip` when no RC number is passed in.
<buildDir>/MAJOR_MINOR_PATCH/Firebase-latest-manual.zip
<buildDir>/carthage/MAJOR.MINOR.PATCH/latest-non-rc/<ALL_CARTHAGE_FILES>
```

This is used for the automated Kokoro setup, so unless we know the Firebase version we can't predict the full path of the output. I'm also fairly certain that regex doesn't work yet on the artifacts fetching, so we can't do something like `<buildDir>/*/Firebase-latest-manual.zip` currently.

PR #4770, opened 2020-01-29
Add unit tests for FIRCLSRecord*

jasonhu-g:
Add unit tests to validate the logic of reading the .clsrecord files


Crashlytics/UnitTests/FIRCLSRecordTests.m
```diff
@@ -0,0 +1,227 @@
...
+
+  // Verify first
+  XCTAssertTrue([adapter.keyValues[0].key isEqualToString:@"com.crashlytics.in-background"]);
+  XCTAssertTrue([adapter.keyValues[0].value isEqualToString:@"0"]);
+
+  // Verify last
+  XCTAssertTrue([adapter.keyValues[11].key isEqualToString:@"com.crashlytics.user-id"]);
+  XCTAssertTrue([adapter.keyValues[11].value
+      isEqualToString:@"test-user-28AE6E09-BC30-4CB3-9FA5-FE06828B8F3C"]);
+}
```
samedson: What do you think about storing a key-value line with an emoji in it, and then reading it out here to make sure they work?

Crashlytics/UnitTests/FIRCLSRecordTests.m
```diff
@@ -0,0 +1,227 @@
...
+@interface FIRCLSRecordTests : XCTestCase
+
+@end
+
+@implementation FIRCLSRecordTests
+
+/// It is important that crashes do not occur when reading persisted crash files before uploading
+/// Verify various invalid input cases
+- (void)testInvalidRecordCases {
+  id adapter __unused = [[FIRCLSRecordAdapter alloc] initWithPath:@"nonExistentPath"];
```
samedson: Could we add a test case where we initialize with a valid path, but files in the path are corrupted / invalid JSON?

Crashlytics/UnitTests/FIRCLSRecordTests.m
jasonhu-g: I remove this test case as I am now using FIRCLSFile's decode/encode.

This test case would work:
- (void)testHexDecoding {
  NSString *str = @"Hello world! 👻";
  NSString *encoded =  FIRCLSFileHexEncodeString([str UTF8String]);
  NSString *decoded =  FIRCLSFileHexDecodeString([encoded UTF8String]);
  XCTAssertTrue([str isEqualToString:decoded]);
}

Crashlytics/UnitTests/FIRCLSRecordTests.m
jasonhu-g: Added.

PR #4771, opened 2020-01-29
FCM: async remove database in FIRMessagingRmqManager

maksymmalyhin:
- tests for the deadlock b/148389485
- fix for the deadlock b/148389485
- test fixes to support async `FIRMessagingRmqManager` database removing
- `FIRTestsAssertionHandler` to override default assertion handler for a specific class

### DIFF
--- a/Example/Messaging/Tests/FIRMessagingRmqManagerTest.m
+++ b/Example/Messaging/Tests/FIRMessagingRmqManagerTest.m
@@ -329,10 +329,16 @@ static const NSTimeInterval kAsyncTestTimout = 0.5;
   XCTestExpectation *assertionFailureExpectation = [self expectationWithDescription:@"assertionFailureExpectation"];
   assertionFailureExpectation.assertForOverFulfill = NO;
 
+#ifndef FIR_MESSAGING_ASSERTIONS_BLOCKED
   [self.testAssertionHandler setMethodFailureHandlerForClass:[FIRMessagingRmqManager class]
                                                      handler:^(id object, NSString *fileName, NSInteger lineNumber) {
     [assertionFailureExpectation fulfill];
   }];
+#else
+  // If FIR_MESSAGING_ASSERTIONS_BLOCKED is defined, then no assertion handlers will be called,
+  // so don't wait for it.
+  [assertionFailureExpectation fulfill];
+#endif // FIR_MESSAGING_ASSERTIONS_BLOCKED
 
   // Create `FIRMessagingRmqManager` instance with a broken database.
   FIRMessagingRmqManager *manager = [[FIRMessagingRmqManager alloc] initWithDatabaseName:databaseName];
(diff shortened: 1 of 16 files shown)


Example/Messaging/Tests/FIRMessagingRmqManagerTest.m
```diff
@@ -302,6 +320,31 @@ - (void)testDeleteSyncMessage {
   XCTAssertNil([self.rmqManager querySyncMessageWithRmqID:rmqID]);
 }
 
+- (void)testInitWhenDatabaseIsBrokenThenDatabaseIsDeleted {
+  NSString *databaseName = @"invalid-database-file";
+  NSString *databasePath = [self createBrokenDatabaseWithName:databaseName];
+  XCTAssert([[NSFileManager defaultManager] fileExistsAtPath:databasePath]);
+
+  // Expect for at least one assertion.
```
ryanwilson: "Expect at least one"? or "Expect one assertion"?

ryanwilson: If there can only be one assertion in the `initWithDatabaseName`, it's likely worth having it only allow assertion. Otherwise we could be catching other assertions happening.

Example/Messaging/Tests/FIRTestsAssertionHandler.h
```diff
@@ -0,0 +1,37 @@
...
+* limitations under the License.
+*/
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void(^FIRTestsAssertionHandlerBlock)(id object, NSString *fileName, NSInteger lineNumber);
+
+@interface FIRTestsAssertionHandler : NSAssertionHandler
```
ryanwilson: I didn't know about `NSAssertionHandler`, neat!

Example/Messaging/Tests/FIRMessagingRmqManagerTest.m
maksymmalyhin: Makes sense. I'll take a closer look how many assertions are supposed to fail and update the test

maksymmalyhin: Currently there are two DB query initiated on the object init. It looks like if the DB file is damaged for some reason, the class will not work as expected until FCM stack is initialized (basically until the host app restart I guess). I am not sure it is the best strategy. I assume we can improve it, but I hesitate doing it without a discussion with @chliangGoogle.

Taking into account that this behaviour was implemented a while ago and the deadlock was introduced just recently, I would proceed with the small fix for the deadlock (and keep the tests tolerant to other potential assertions) and consider a better recovery strategy in the future. WDYT?

chliangGoogle: I think we added dispatch_async in removeDatabase recently for unit tests which is what i'm worried why deadlock happens.

maksymmalyhin:
@chliangGoogle For the PR I first added tests to reproduce the deadlock and then fixed it replacing `dispatch_sync` by `dispatch_async`. `dispatch_async` can be safely called on the same queue since it is a non-blocking function.

PR #4773, opened 2020-01-29
fix heartbeat

VinayGuthal:
use 1 less 0 so that it sends every day as per to once in 100 days

### DIFF
--- a/FirebaseCore/Sources/FIRHeartbeatInfo.m
+++ b/FirebaseCore/Sources/FIRHeartbeatInfo.m
@@ -16,7 +16,7 @@
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
 #import <GoogleUtilities/GULLogger.h>
 
-const static long secondsInDay = 864000;
+const static long secondsInDay = 86400;
 @implementation FIRHeartbeatInfo : NSObject
 
 /** Updates the storage with the heartbeat information corresponding to this tag.
--- a/FirebaseCore/Tests/Unit/FIRHeartbeatInfoTest.m
+++ b/FirebaseCore/Tests/Unit/FIRHeartbeatInfoTest.m
@@ -32,12 +32,12 @@
   NSDateComponents *componentsToAdd = [[NSDateComponents alloc] init];
   componentsToAdd.day = -1;
 
-  NSDate *pastTime = [[NSCalendar currentCalendar] dateByAddingComponents:componentsToAdd
+  NSDate *dayAgo = [[NSCalendar currentCalendar] dateByAddingComponents:componentsToAdd
                                                                    toDate:[NSDate date]
                                                                   options:0];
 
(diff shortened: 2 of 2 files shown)


FirebaseCore/Sources/FIRHeartbeatInfo.m
```diff
@@ -16,7 +16,7 @@
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
 #import <GoogleUtilities/GULLogger.h>
 
-const static long secondsInDay = 864000;
+const static long secondsInDay = 86400;
```
maksymmalyhin: We should add tests for this to make sure we won't miss a similar issue again.

FirebaseCore/Tests/Unit/FIRHeartbeatInfoTest.m
```diff
@@ -29,7 +29,7 @@ @implementation FIRHeartbeatInfoTest
 - (void)setUp {
   NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
   self.dataStorage = [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
-  NSDate *pastTime = [NSDate dateWithTimeIntervalSinceNow:-996400];
+  NSDate *pastTime = [NSDate dateWithTimeIntervalSinceNow:-99640];
```
maksymmalyhin: I would suggest using `NSCalendar` and `NSComponents` to calculate the date in a more readable way, e.g. 
```
NSDateComponents *componentsToAdd = [[NSDateComponents alloc] init];
componentsToAdd.day = -1;

NSDate *dateYesterday = [[NSCalendar currentCalendar] dateByAddingComponents:componentsToAdd toDate:[NSDate date] options:0];
```

It may look a bit overkill, but I think it is worth it taking into account the context of the issue.

FirebaseCore/Tests/Unit/FIRHeartbeatInfoTest.m
```diff
@@ -29,7 +29,13 @@ @implementation FIRHeartbeatInfoTest
 - (void)setUp {
   NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
   self.dataStorage = [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
-  NSDate *pastTime = [NSDate dateWithTimeIntervalSinceNow:-996400];
+  NSDateComponents *componentsToAdd = [[NSDateComponents alloc] init];
+  componentsToAdd.day = -1;
+
+  NSDate *pastTime = [[NSCalendar currentCalendar] dateByAddingComponents:componentsToAdd
```
maksymmalyhin: nit: maybe rename `pastTime` to smt like `dayAgo` or `yesterday`.  It looks a bit nicer to me, but I'm fine if you leave it as is.

PR #4767, opened 2020-01-29
FDL source, unit tests and app now live in FirebaseDynamicLinks folder.

dmandar:

PR #4774, opened 2020-01-30
Update dynamiclinks.yml

dmandar:


paulb777:
Looks right as part of or to merge to #4767

PR #4794, opened 2020-02-01
Add a storage completion handler to the event transport pipeline

mikehaney24:
In support of Crashlytics


jasonhu-g:
LGTM, thanks!

PR #4795, opened 2020-02-01
Fixes for issues discovered in the Unity desktop build

wilhuff:
Ensure that Firestore properly shuts down when the Unity AppDomain is unloaded. Improper shutdown manifested as a crash the next time Firestore was used because the LevelDB database files were still locked by the prior instance.

AppDomain unload differs from typical iOS termination because we want Dispose to be synchronous so that we signal shutdown is complete by the time Dispose returns. Without this, there's a race between shutdown in the current domain and startup in the next domain. This poses a problem though because disposers are run on the main thread, and callbacks from Firestore's async operations also typically run on the main thread. This means that by default there's no great way to implement Dispose without deadlock.

This change makes it such that the `api::Firestore` destructor now synchronously terminates the instance. There are a few moving parts to making this work:

  * The AsyncQueue now extends `std::enable_shared_from_this` captures a `shared_ptr` to itself whenever it enqueues an asynchronous operation. This ensures that if an operation is enqueued and the user doesn't wait for it to complete, the queue won't deallocate itself before the operation completes.

  * Makes ExecutorStd handle the case where its destructor might be running on a thread that it owns. This can now happen if an AsyncQueue Operation is the last thing holding a reference to itself.

  * Adds a new, synchronous FirestoreClient::Terminate method that does not have a callback. This is used to avoid the deadlock inherent with cross-queue notifications when the caller may be trying to block synchronously in the queue to which the notification would go.

  * Changes scheduled operations to hold their their references to FirestoreClient via `weak_ptr`, to avoid accidentally preserving the life of the FirestoreClient past Terminate.

  * Makes the api::Firestore destructor call a new synchronous Terminate method.


Firestore/core/src/firebase/firestore/core/firestore_client.h
```diff
@@ -172,6 +172,13 @@ class FirestoreClient : public std::enable_shared_from_this<FirestoreClient> {
   }
   bool is_terminated() const;
 
+#if __cplusplus < 201402L
```
var-const: Why not just have
```cpp
std::weak_ptr<FirestoreClient> weak_this = shared_from_this();
```
at the point of the call? While `weak_from_this` is the better solution in C++14, I'm not sure the benefits outweigh having this conditional logic.

Firestore/core/src/firebase/firestore/api/firestore.cc
```diff
@@ -62,6 +62,15 @@ Firestore::Firestore(model::DatabaseId database_id,
       extension_{extension} {
 }
 
+Firestore::~Firestore() {
+  // If the client hasn't been configured yet we don't need to create it just
```
var-const: Move this comment one line down?

Firestore/core/src/firebase/firestore/util/async_queue.cc
```diff
@@ -72,9 +72,7 @@ void AsyncQueue::Enqueue(const Operation& operation) {
 void AsyncQueue::EnqueueAndInitiateShutdown(const Operation& operation) {
   std::lock_guard<std::mutex> lock{shut_down_mutex_};
   VerifySequentialOrder();
-  if (is_shutting_down_) {
-    return;
-  }
```
var-const: Can you provide more context for this change?

Firestore/core/src/firebase/firestore/core/firestore_client.h
```diff
@@ -88,7 +88,8 @@ class FirestoreClient : public std::enable_shared_from_this<FirestoreClient> {
    * Terminates this client, cancels all writes / listeners, and releases all
    * resources.
    */
-  void Terminate(util::StatusCallback callback);
+  void TerminateAsync(util::StatusCallback callback);
+  void Terminate();
```
var-const: Can you add a comment explaining how `Terminate` is different?

Firestore/core/src/firebase/firestore/core/firestore_client.cc
```diff
@@ -248,30 +248,57 @@ void FirestoreClient::EnableNetwork(StatusCallback callback) {
...
+  });
+
+  LOG_DEBUG("Terminate: waiting");
+  terminated_condition.wait(outer_lock, [&] { return terminated; });
+
+  LOG_DEBUG("Terminate: complete");
+}
+
+void FirestoreClient::TerminateInternal() {
+  if (remote_store_) {
```
var-const: Optional: use early return?

Firestore/core/src/firebase/firestore/core/firestore_client.cc
```diff
@@ -248,30 +248,57 @@ void FirestoreClient::EnableNetwork(StatusCallback callback) {
...
 }
 
+void FirestoreClient::Terminate() {
+  std::mutex mutex;
+  bool terminated = false;
+  std::condition_variable terminated_condition;
+
+  std::unique_lock<std::mutex> outer_lock{mutex};
+
+  LOG_DEBUG("Terminate: starting");
```
var-const: Question: is this temporary logging, or do you intend to keep it?

Firestore/core/src/firebase/firestore/core/firestore_client.cc
```diff
@@ -248,30 +248,57 @@ void FirestoreClient::EnableNetwork(StatusCallback callback) {
...
+    TerminateInternal();
+
+    std::lock_guard<std::mutex> lock{mutex};
+    LOG_DEBUG("Terminate: notifying");
+    terminated = true;
+    terminated_condition.notify_one();
+  });
+
+  LOG_DEBUG("Terminate: waiting");
+  terminated_condition.wait(outer_lock, [&] { return terminated; });
```
var-const: Question: what's the advantage of using a conditional variable rather than a `promise`? (like here: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/util/executor_std.cc#L154-L159)

Firestore/core/src/firebase/firestore/core/firestore_client.h
wilhuff: I was making it such that the caller could look like C++14 and we could just drop this block when C++14 became the floor. I'm happy to do it the other way (review usages of `shared_from_this` to see if any could be made `weak_from_this`).

Firestore/core/src/firebase/firestore/api/firestore.cc
wilhuff: Done.

Firestore/core/src/firebase/firestore/core/firestore_client.cc
wilhuff: This was a mistake; removed.

Firestore/core/src/firebase/firestore/core/firestore_client.cc
wilhuff: There's no advantage. Thanks for the reminder! Fixed.

Firestore/core/src/firebase/firestore/core/firestore_client.cc
wilhuff: Done.

Firestore/core/src/firebase/firestore/core/firestore_client.h
wilhuff: It's a synchronous variant; I've added a comment.

Firestore/core/src/firebase/firestore/util/async_queue.cc
wilhuff: See this commit comment: https://github.com/firebase/firebase-ios-sdk/pull/4795/commits/a8c4136d9e50314730e3abd6fb3e971a6c47c8db

Essentially, it was previously the case that this short circuited if already shut down, but this caused the caller to have to `EnqueueEvenAfterShutdown` to ensure that the callback was still called. Removing this short circuit simplifies `FirestoreClient::TerminateAsync`.

Firestore/core/src/firebase/firestore/core/firestore_client.h
var-const: Thanks. Consider renaming the new function to `TerminateBlocking`; I think it would really highlight the difference.

Firestore/core/src/firebase/firestore/core/firestore_client.h
var-const: Not for this PR: perhaps we could have our own `enable_shared_from_this` for the time being. It would inherit from the standard library one and conditionally add `weak_from_this`. It's more work, but the benefit would be that the conditional logic is in just one place. No strong feelings about this, though.

Firestore/core/src/firebase/firestore/core/firestore_client.cc
var-const: Question: sounds like this would still trigger the issue on macOS where if the user queue is the main queue, the destructor would hang forever?

wilhuff: No--this implementation avoids that altogether.

There are now two Terminate methods: `TerminateAsync` calls back a callback on the main queue. `Terminate` does not call back on the queue at all.

Firestore/core/src/firebase/firestore/core/firestore_client.h
wilhuff: I considered that, but it seems worse than both iterations that are in this PR. Having a custom base class means that our classes no longer extend `std::enable_shared_from_this` which limits the self-documentary benefit.

Firestore/core/src/firebase/firestore/core/firestore_client.h
wilhuff: I'm partial to the convention of FooAsync/Foo pairs. This is already common in C# and C++ APIs to some extent.

PR #4752, opened 2020-01-27
Add a GitHub Actions workflow for Firestore

wilhuff:
#no-changelog

PR #4764, opened 2020-01-28
Prepare build scripts for building Firestore on GitHub Actions

wilhuff:
This includes a bunch of fixes that come primarily from the fact that our Travis build is configured by the environment while so far we've been trending toward configuring GHA explicitly.

Specifically:
  * This makes it possible to invoke `install_prereqs.sh` with command-line arguments;
  * Ensures that bundler is up-to-date by invoking `setup_bundler.sh` from `install_prereqs.sh`;
  * Adds Linux build support to `install_prereqs.sh` and `build.sh`; and
  * Converts Firestore CMake builds to use `ninja` instead of `make` for a ~20% improvement in build time.

These changes should make it straightforward to port any Travis build that requires `install_prereqs.sh` to GHA:

```yaml
  xcodebuild:
    runs-on: macos-latest

    strategy:
      matrix:
        target: [iOS, tvOS, macOS]

    steps:
    - uses: actions/checkout@v2

    - name: Setup build
      run:  scripts/install_prereqs.sh Firestore ${{ matrix.target }} xcodebuild

    - name: Build and test
      run:  scripts/third_party/travis/retry.sh scripts/build.sh Firestore ${{ matrix.target }} xcodebuild
```

PR #4796, opened 2020-02-01
Use matrix for Core GHA

paulb777:
Parallelize the `pod lib lint` platform runs with `matrix`

Inspired by #4752 by @wilhuff

#no-changelog

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -20,5 +20,5 @@ jobs:
     - uses: actions/checkout@v2
     - name: Setup Bundler
       run: ./scripts/setup_bundler.sh
-    - name: FirebaseCore iOS
+    - name: Build and test
       run: ./scripts/pod_lib_lint.rb FirebaseCore.podspec --platforms=${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


paulb777:
cc: @ryanwilson @maksymmalyhin @mikehaney24

ryanwilson:
Neat, great find!!

.github/workflows/core.yml
```diff
@@ -10,15 +10,15 @@ on:
...
-  core:
+  pod_lib_lint:
     runs-on: macOS-latest
+
+    strategy:
+      matrix:
+        target: [ios, tvos, macos]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
mikehaney24: why checkout@v2?

paulb777: See #4764 and https://github.com/actions/checkout

wilhuff: That PR doesn't really explain checkout@v2. There's some details in #4752 though, see specifically https://github.com/firebase/firebase-ios-sdk/pull/4752/commits/5ce0b2332cb6c4e39203689cda3e574af6aa39d4.

In short, the major difference between the versions is that v1 does a full clone including all branches, tags, and commits. checkout@v2 does a shallow clone of just one commit from the branch in question. This makes checkouts much faster.

PR #4812, opened 2020-02-04
Fixed iOS testapp dark mode

doudounan:
Fixed iOS testapp dark mode


doudounan:
Fixed dark mode display on normal GDT iOS testapp.

PR #4814, opened 2020-02-04
Update Spec test protos

schmidt-sebastian:
This pulls in firebase/firebase-js-sdk#2594


schmidt-sebastian:
> I presume it's okay that _all_ versions are increased? I guess the generator code could check for zero explicitly.

Yes. It doesn't make much of a difference, as long as there is no zero version.

PR #4816, opened 2020-02-04
Fix ignore paths for Check

ryanwilson:
Based on [the docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths), `paths` should be used for excluding only when it's also specifying which paths to include. Otherwise, `paths-ignore` should be used.


ryanwilson:
Manually cancelled Travis.

Issue #3686, opened 2019-08-26
Various Firebase types does not conform to NSSecureCoding


clausjoergensen:
### [READ] Step 1: Are you in the right place?
Yes

### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 10.3
  * Firebase SDK version: 5.16
  * Firebase Component: Core, Messaging
  * Component version: 5.16

### [REQUIRED] Step 3: Describe the problem

Our security team scanned our code and dependencies for vulnerabilities and found that the following classes did not conform to `NSSecureCoding` making them vulnerable to object substitution attacks:

`FIRMessagingTopicBatch`, `FIRInstanceIDAPNSInfo`, `FIRInstanceIDTokenInfo` and `FIRMessagingPendingTopicsList`

#### Steps to reproduce:

Open up the header files, note that they conform to `NSCoding`, and not `NSSecureCoding`.

#### Relevant Code:

- https://github.com/firebase/firebase-ios-sdk/blob/master/Firebase/Messaging/FIRMessagingPendingTopicsList.h
- https://github.com/firebase/firebase-ios-sdk/blob/master/Firebase/InstanceID/FIRInstanceIDAPNSInfo.h
- https://github.com/firebase/firebase-ios-sdk/blob/master/Firebase/InstanceID/FIRInstanceIDTokenInfo.h

chliangGoogle:
We removed NSCoding from Messaging but not InstanceID yet. InstanceID is deprecated and will be removed from Messaging soon. We will keep this open for now.

paulb777:
This was addressed for Messaging as noted above. InstanceID is no longer used by Firebase.

PR #4826, opened 2020-02-05
Use secure encoding for messaging's pending topics

chliangGoogle:
This fixes part of #3686

### DIFF
--- a/Firebase/Messaging/FIRMessagingPendingTopicsList.m
+++ b/Firebase/Messaging/FIRMessagingPendingTopicsList.m
@@ -48,7 +48,11 @@ NSString *const kPendingTopicsTimestampEncodingKey = @"ts";
   return self;
 }
 
-#pragma mark NSCoding
+#pragma mark NSSecureCoding
+
++ (BOOL)supportsSecureCoding {
+  return YES;
+}
 
 - (void)encodeWithCoder:(NSCoder *)aCoder {
   [aCoder encodeInteger:self.action forKey:kPendingTopicBatchActionKey];
@@ -56,7 +60,6 @@ NSString *const kPendingTopicsTimestampEncodingKey = @"ts";
 }
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder {
-
   // Ensure that our integer -> enum casting is safe
   NSInteger actionRawValue = [aDecoder decodeIntegerForKey:kPendingTopicBatchActionKey];
   FIRMessagingTopicAction action = FIRMessagingTopicActionSubscribe;
@@ -65,10 +68,7 @@ NSString *const kPendingTopicsTimestampEncodingKey = @"ts";
   }
 
   if (self = [self initWithAction:action]) {
-    NSSet *topics = [aDecoder decodeObjectForKey:kPendingTopicBatchTopicsKey];
-    if ([topics isKindOfClass:[NSSet class]]) {
-      _topics = [topics mutableCopy];
-    }
+    _topics = [aDecoder decodeObjectOfClass:[NSSet class] forKey:kPendingTopicBatchTopicsKey];
     _topicHandlers = [NSMutableDictionary dictionary];
   }
   return self;
@@ -109,7 +109,11 @@ NSString *const kPendingTopicsTimestampEncodingKey = @"ts";
   }
 }
 
-#pragma mark NSCoding
+#pragma mark NSSecureCoding
+
++ (BOOL)supportsSecureCoding {
+  return YES;
+}
 
 - (void)encodeWithCoder:(NSCoder *)aCoder {
   [aCoder encodeObject:[NSDate date] forKey:kPendingTopicsTimestampEncodingKey];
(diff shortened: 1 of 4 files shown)


Firebase/Messaging/FIRMessagingPubSub.m
```diff
@@ -183,10 +183,14 @@ - (BOOL)pendingTopicsListCanRequestTopicUpdates:(FIRMessagingPendingTopicsList *
...
 - (void)archivePendingTopicsList:(FIRMessagingPendingTopicsList *)topicsList {
   GULUserDefaults *defaults = [GULUserDefaults standardUserDefaults];
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-  NSData *pendingData = [NSKeyedArchiver archivedDataWithRootObject:topicsList];
-#pragma clang diagnostic pop
+
+  NSData *pendingData;
+  if (@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)) {
+    NSError *error;
```
ryanwilson: Small nit, not to block submission: We pass the error in here but we don't ever check it.  Same with below on line 205.

paulb777:
Merging for M64 as is.

@chliangGoogle Please follow up on the nit separately.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
```diff
@@ -65,10 +68,7 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
   }
 
   if (self = [self initWithAction:action]) {
-    NSSet *topics = [aDecoder decodeObjectForKey:kPendingTopicBatchTopicsKey];
-    if ([topics isKindOfClass:[NSSet class]]) {
-      _topics = [topics mutableCopy];
-    }
+    _topics = [aDecoder decodeObjectOfClass:[NSSet class] forKey:kPendingTopicBatchTopicsKey];
```
maksymmalyhin: I wonder if it's working correctly without specifying the all classes available for decoding with [unarchivedObjectOfClasses:fromData:error:](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/2962885-unarchivedobjectofclasses?language=objc)

Firebase/Messaging/FIRMessagingPubSub.m
```diff
@@ -195,16 +199,17 @@ - (void)restorePendingTopicsList {
...
+  if (pendingData) {
+    if (@available(macOS 10.13, iOS 11.0, tvOS 11.0, watchOS 4.0, *)) {
+      NSError *error;
+      subscriptions = [NSKeyedUnarchiver unarchivedObjectOfClass:[NSData class] fromData:pendingData error:&error];
+    } else {
+      @try {
+        subscriptions = [NSKeyedUnarchiver unarchiveObjectWithData:pendingData];
+      } @catch (NSException *exception) {
+        // Nothing we can do, just continue as if we don't have pending subscriptions
+      }
```
maksymmalyhin: I would suggest using [GULSecureCoding](https://github.com/firebase/firebase-ios-sdk/blob/b7fec9e2b82ef91fe4a2422d2f409cfd58c9b4fe/GoogleUtilities/Environment/Public/GULSecureCoding.h#L24) instead of a custom implementation to reduce code duplication.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
```diff
@@ -109,20 +109,22 @@ + (void)pruneTopicBatches:(NSMutableArray <FIRMessagingTopicBatch *> *)topicBatc
...
 
 - (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
-
   if (self = [self init]) {
-    _archiveDate = [aDecoder decodeObjectForKey:kPendingTopicsTimestampEncodingKey];
-    NSArray *archivedBatches = [aDecoder decodeObjectForKey:kPendingBatchesEncodingKey];
-    if (archivedBatches) {
-      _topicBatches = [archivedBatches mutableCopy];
+    _archiveDate = [aDecoder decodeObjectOfClass:[NSDate class] forKey:kPendingTopicsTimestampEncodingKey];
+    _topicBatches = [aDecoder decodeObjectOfClass:[NSMutableArray<FIRMessagingTopicBatch *> class] forKey:kPendingBatchesEncodingKey];
```
maksymmalyhin: The same here, we should pass `FIRMessagingTopicBatch` class like [here](https://github.com/firebase/firebase-ios-sdk/blob/f138a75a33cde26a58e586274e2dc7f8618fd924/GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m#L314)

Firebase/Messaging/FIRMessagingPendingTopicsList.m
chliangGoogle: what do you mean all the classes? isn't _topics are defined as NSSet only?

Firebase/Messaging/FIRMessagingPubSub.m
chliangGoogle: nice, didn't know there's a shared method already.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
chliangGoogle: you mean instead of NSMutableArray<FIRMessagingTopicBatch *>, use [NSSet setWithObjects:[NSMutableArray class], [FIRMessagingTopicBatch class], nil];

chliangGoogle:
Thanks for the review! I will make another PR once some of the questions have been resolved.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
maksymmalyhin: AFAIK the classes of the objects stored in `NSSet` also must be specified otherwise decoding will fail.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
maksymmalyhin: yep.

PR #4830, opened 2020-02-05
Actually don't run zip on markdown changes.

ryanwilson:
The nested folder regex did not work as expected.

#no-changelog


ryanwilson:
> Isn't the right fix `paths-ignore:` with the `**` version?

When using `paths`, you can't use `paths-ignore`.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#excluding-paths

PR #4836, opened 2020-02-05
Fix bits.h when compiled by MSVC after including <ciso646>

wilhuff:


Firestore/core/src/firebase/firestore/util/bits.h
```diff
@@ -24,6 +24,24 @@
 
 #include <cstdint>
 
+#if _MSC_VER
+// The Microsoft implementation of iso646.h defines alternative operator names
```
var-const: Nit: perhaps s/iso646.h/ciso646/?

wilhuff: I debated that. From a consumer's point of view it's either header, but really there's only one implementation and the offending macros are definitely in iso646.h.

var-const: Fair enough.

PR #4837, opened 2020-02-06
Fix bits.h when compiled by MSVC after including <ciso646> (#4836)

wilhuff:
This is a cherry-pick of #4836 for release-6.17.0.

This fixes builds on 32-bit Windows in google3 after cl/291012718, which added `#include <ciso646>` to `absl/base/options.h`.

This has no effect on any Apple platform so it doesn't warrant a respin of any iOS release candidates.

### DIFF
--- a/Firestore/core/src/firebase/firestore/util/bits.h
+++ b/Firestore/core/src/firebase/firestore/util/bits.h
@@ -24,6 +24,24 @@
 
 #include <cstdint>
 
+#if _MSC_VER
+// The Microsoft implementation of iso646.h defines alternative operator names
+// as macros and this interferes with the inline assembly, below, which uses the
+// `and` and `xor` instructions. Defining these as macros is conforming behavior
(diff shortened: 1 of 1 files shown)


paulb777:
Why the cherry-pick if it's not needed for release candidates?  Why isn't GHA green?

wilhuff:
We copybara release branches into google3 and cherry-picking this ensures that we won't forget this change on our next copybara.

wilhuff:
The xcodebuild targets are failing because of this:

```
[!] CocoaPods could not find compatible versions for pod "Firebase/CoreOnly":
  In Podfile:
    Firebase/CoreOnly (= 6.17.0)
```

... which doesn't make much sense because the Firebase pod on this branch clearly has that version. I'm investigating.

paulb777:
Approving, since the Podfile tests never have worked on the release branch prior to release.

PR #4840, opened 2020-02-06
Map the read values from user key value, error and log .clsrecords into the Report nanopb model

jasonhu-g:
This is a follow up to: https://github.com/firebase/firebase-ios-sdk/pull/4793

1. Map the read values from user key value, error and log .clsrecords into the Report nanopb models
2. Add TODOs to track potential logic that we have to move from protobuilder into the SDK

The next related PR will be to add unit tests for protobuilder logic that has been moved into the SDK.

### DIFF
--- a/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
+++ b/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
@@ -288,7 +288,6 @@
   NSMutableDictionary<NSString *, NSString *> *kvs = [self.userKeyValues mutableCopy];
   kvs[@"nserror-domain"] = error.domain;
   kvs[@"nserror-code"] = [NSString stringWithFormat:@"%li", (long)error.code];
-  ;
 
   return kvs;
 }
@@ -407,7 +406,7 @@
   }
 
   // Add crash event
-  if (numberOfEvents > 0) {
+  if ([self hasCrashed]) {
     events[numberOfEvents - 1] = [self protoEventForCrash];
   }
 
--- a/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
+++ b/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
@@ -42,8 +42,8 @@
     [self loadUserLogFiles];
     [self loadErrorFiles];
 
-    // TODO: Add support for mach_exception.clsrecord
-    // TODO: When implemented, add support for custom exceptions: custom_exception_a.clsrecord
+    // TODO: Add support for mach_exception.clsrecord (check Protobuf.scala:524)
+    // TODO: When implemented, add support for custom exceptions: custom_exception_a/b.clsrecord
 
     _report = [self protoReport];
   }
@@ -61,6 +61,7 @@
 /// Reads from binary_images.clsrecord
 - (void)loadBinaryImagesFile {
   NSString *path = [self.folderPath stringByAppendingPathComponent:FIRCLSReportBinaryImageFile];
+  // TODO: Should sort? If so, sort inside FIRCLSRecordBinaryImage. Protobuf:253
   self.binaryImages = [FIRCLSRecordBinaryImage
       binaryImagesFromDictionaries:[FIRCLSReportAdapter dictionariesFromEachLineOfFile:path]];
 }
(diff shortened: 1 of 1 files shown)


Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -351,31 +388,55 @@ - (google_crashlytics_Session_Device)protoSessionDevice {
...
+      malloc(sizeof(google_crashlytics_Session_Event) * numberOfEvents);
+
+  for (NSUInteger i = 0; i < self.errors.count; i++) {
+    events[i] = [self protoEventForError:self.errors[i]];
+  }
+
+  // Add crash event
+  if (numberOfEvents > 0) {
+    events[numberOfEvents - 1] = [self protoEventForCrash];
+  }
```
samedson: If we have 4 non-fatals, but no crash, I think this will still try to put a crash into the report?

jasonhu-g: Nice catch, this was an artifact of some manual testing that I was doing. I'll change this back to what I had:
`if ([self hasCrashed)`

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -388,13 +449,66 @@ - (google_crashlytics_Session_Event_Application)protoEventApplication {
...
+
+/// Generate an array of CustomAttributes from the user defined key values.
+/// For recorded errors, the error's nserror-domain and nserror-code are also added.
+/// @param error Error from recordError API
+- (google_crashlytics_CustomAttribute *)protoCustomAttributesWithError:(FIRCLSRecordError *)error {
+  NSDictionary<NSString *, NSString *> *kvs = [self KeyValuesWithError:error];
+  NSUInteger size = kvs.count;
+
+  google_crashlytics_CustomAttribute *attributes =
+      malloc(sizeof(google_crashlytics_CustomAttribute) * size);
```
samedson: If size is zero, might want to return? Just a suggestion but leave it up to you. Not sure what happens with nanopb in the case where it gets a zero-length, non NULL pointer (if that happens to be the semantics of malloc(0) on iOS).

https://stackoverflow.com/questions/2022335/whats-the-point-in-malloc0

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -388,13 +449,66 @@ - (google_crashlytics_Session_Event_Application)protoEventApplication {
   app.background = [self wasInBackground];
   app.ui_orientation = [self uiOrientation];
 
-  // TODO: Where do these come from?
-  //    app.custom_attributes
-  //    app.custom_attributes_count
+  app.custom_attributes = [self protoCustomAttributesWithError:nil];
+  app.custom_attributes_count = (pb_size_t)self.userKeyValues.count;
```
samedson: Remember crash_info_entry for Swift crashes? Should we add a comment for that?

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -388,13 +448,66 @@ - (google_crashlytics_Session_Event_Application)protoEventApplication {
...
+  thread.frames_count = (pb_size_t)error.stacktrace.count;
+  threads[0] = thread;
+  app.execution.threads = threads;
+  app.execution.threads_count = 1;
+
+  app.background = [self wasInBackground];
+  app.ui_orientation = [self uiOrientation];
+
+  app.custom_attributes = [self protoCustomAttributesWithError:error];
+  app.custom_attributes_count = (pb_size_t)self.userKeyValues.count + 2;
```
samedson: Here, line 452, and line 495 need to be reconciled - I'm worried we are gonna get our counts wrong with what's allocated in the arrays.
 - If we change the number inside `protoCustomAttributesWithError`, the +2 on this line may get out of sync
 - As a reader, I'm unsure the relationship between `self.userKeyValues.count` and `kvs.count`on 495.

A couple suggestions is to add an argument `* count` to `protoCustomAttributesWithError` and write to the value inside the method, or pass the count into `protoCustomAttributesWithError` as an argument, or multiple returns?

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
jasonhu-g: I just verified that when size = 0, nanopb treats the fields as not set and that I am able to still parse the nanopb bytes.

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
jasonhu-g: I now pass the key values into protoCustomAttributes

PR #4844, opened 2020-02-06
Make sure we turn of fetchInProgress flag on any fetch completion for…

dmandar:


FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -249,7 +252,8 @@ - (void)refreshInstanceIDTokenAndFetchCheckInInfoWithCompletionHandler:
               [NSString stringWithFormat:@"Error getting iid : %@.", error];
           FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000055", @"%@",
                       [NSString stringWithFormat:@"%@", errorDescription]);
-          return [self
+          strongSelfQueue->_settings.isFetchInProgress = NO;
```
maksymmalyhin: Are we sure `isFetchInProgress` is accessed in thread safe way?   Sorry, it's hard to see from the PR itself...

maksymmalyhin: Ah, ok, I can see that it is an `atomic` property now, so should be good.

maksymmalyhin:
@dmandar It looks like the impacted methods have several failable async operations chained to each other which is difficult to express and read with callback blocks. The code looks like a good candidate to be refactored using Promises to improve readability and simplify the code. Now since we have `FBLPromises` lib as a transitive dependency for Remote Config anyway, would you like to consider such refactoring? I am happy to help with it if have any questions.

dmandar:
> @dmandar It looks like the impacted methods have several failable async operations chained to each other which is difficult to express and read with callback blocks. The code looks like a good candidate to be refactored using Promises to improve readability and simplify the code. Now since we have `FBLPromises` lib as a transitive dependency for Remote Config anyway, would you like to consider such refactoring? I am happy to help with it if have any questions.

hey..yes, have been meaning to refactor for some time now and promises sound great. Noted!

PR #4848, opened 2020-02-06
use secure coding and secure archiving for messaging

chliangGoogle:


Firebase/Messaging/FIRMessagingPendingTopicsList.m
```diff
@@ -65,10 +68,9 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
   }
 
   if (self = [self initWithAction:action]) {
-    NSSet *topics = [aDecoder decodeObjectForKey:kPendingTopicBatchTopicsKey];
-    if ([topics isKindOfClass:[NSSet class]]) {
-      _topics = [topics mutableCopy];
-    }
+    _topics = [aDecoder decodeObjectOfClasses:
+                     [NSSet setWithObjects:NSSet.class, NSMutableString.class, nil]
```
maksymmalyhin: Should it be `NSMutableSet` and `NSString`?

Example/Messaging/Tests/FIRMessagingPubSubTest.m
```diff
@@ -87,4 +104,30 @@ - (void)testRemoveTopicPrefix {
...
+  id mockRmqManager = OCMClassMock([FIRMessagingRmqManager class]);
+  FIRMessagingClient *client = [[FIRMessagingClient alloc]
+                                initWithDelegate:mockClientDelegate
+                                reachability:mockReachability
+                                rmq2Manager:mockRmqManager];
+  FIRMessagingPubSub *pubSub = [[FIRMessagingPubSub alloc] initWithClient:client];
+  [pubSub archivePendingTopicsList:topicList];
+ 
+  [pubSub restorePendingTopicsList];
+  XCTAssertEqual(pubSub.pendingTopicUpdates.numberOfBatches, 3);
```
maksymmalyhin: Will it be difficult to validate content here to make sure archiving/unarchiving works as expected? Alternatively we may add tests for the models themself e.g. like [this one](https://github.com/firebase/firebase-ios-sdk/blob/9e5c9055fb2318071eb32fd07888af89e516df3e/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsStoredAuthTokenTests.m#L29).

maksymmalyhin:
@chliangGoogle It looks like FCM code is excluded from formatting by `./scripts/style.sh`. I wonder if you know why? Would you like to consider enabling formatting FCM code with the script? It will generate some significant amount of formatting changes on the first run but will simplify submitting future PRs.

chliangGoogle:
It's been disabled before me, we were talking about reenabling the formatting but at that time I was working on a long PR. Consider there will be huge amount of changes, I will re-enable it in a separate PR. Thanks for reminding me.

Firebase/Messaging/FIRMessagingPendingTopicsList.m
chliangGoogle: Good catch.

Example/Messaging/Tests/FIRMessagingPubSubTest.m
chliangGoogle: Done. Yeah that would be very helpful.

PR #4849, opened 2020-02-07
Set mutable data ref lengths, otherwise iOS will 0 out the buffer

mikehaney24:
iOS seems to do this when the buffer becomes > 16320 bytes.

### DIFF
--- a/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
+++ b/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
@@ -125,6 +125,7 @@ NSString *const kFIRCoreDiagnosticsHeartbeatDateFileName = @"FIREBASE_DIAGNOSTIC
   // Encode a 2nd time to actually get the bytes from it.
   size_t bufferSize = sizestream.bytes_written;
   CFMutableDataRef dataRef = CFDataCreateMutable(CFAllocatorGetDefault(), bufferSize);
+  CFDataSetLength(dataRef, bufferSize);
   pb_ostream_t ostream = pb_ostream_from_buffer((void *)CFDataGetBytePtr(dataRef), bufferSize);
   if (!pb_encode(&ostream, logs_proto_mobilesdk_ios_ICoreConfiguration_fields, &_config)) {
     GDTCORLogError(GDTCORMCETransportBytesError, @"Error in nanopb encoding for bytes: %s",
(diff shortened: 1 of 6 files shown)


Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
```diff
@@ -125,6 +125,7 @@ - (NSData *)transportBytes {
   // Encode a 2nd time to actually get the bytes from it.
   size_t bufferSize = sizestream.bytes_written;
   CFMutableDataRef dataRef = CFDataCreateMutable(CFAllocatorGetDefault(), bufferSize);
+  CFDataSetLength(dataRef, bufferSize);
```
wilhuff: Remove the second call, below?

mikehaney24: Yeah, not done yet. I just wrote this in github to remind myself to do something about it.

mikehaney24: Done, and unit test added. Confirmed that without these changes the unit test fails. Good catch!

GoogleDataTransportCCTSupport/GDTCCTTests/Unit/GDTCCTNanopbHelpersTest.m
```diff
@@ -130,4 +130,44 @@ - (void)testBytesAreDecodable {
...
+/** Tests that creating a message above the apparent threshold of 16320 bytes works. */
+- (void)testEncodingProtoAboveDefaultOSThreshold {
+  NSBundle *testBundle = [NSBundle bundleForClass:[self class]];
+  NSArray *testData = @[
+    @"message-32347456.dat", @"message-35458880.dat", @"message-39882816.dat",
+    @"message-40043840.dat", @"message-40657984.dat"
+  ];
+  NSMutableSet *storedEvents = [[NSMutableSet alloc] init];
+  // 250 messages results in a total size of 16337 which is > 16320, the apparent OS limit.
+  for (int i = 0; i < 250; i++) {  // Changing to 249 would've caused test to pass previously.
```
wilhuff: Consider pulling this comment up a line to make it a continuation of the comment on the prior line.

mikehaney24: Done.

paulb777:
Were you able to find an explanation for why 16320 is a threshold?

mikehaney24:
> Were you able to find an explanation for why 16320 is a threshold?

I really don't know...especially since I give the buffer size on creation. I couldn't find any reference to this value on the internet.

PR #4857, opened 2020-02-07
Decode documents in a background thread

wilhuff:
This is a port of https://github.com/firebase/firebase-android-sdk/pull/559.

This results in a mean speedup across all benchmarks of ~30%.

### DIFF
--- a/Firestore/Example/Benchmarks/remote_document_cache_benchmark.mm
+++ b/Firestore/Example/Benchmarks/remote_document_cache_benchmark.mm
@@ -157,8 +157,7 @@ void BM_QueryMatching(benchmark::State& state) {
   WriteDocs(collection, total_docs - matching_docs, /* match= */ false);
 
   for (auto _ : state) {
-    auto docs = GetDocumentsFromCache(
-        [collection queryWhereField:@"match" isEqualTo:@YES]);
+    auto docs = GetDocumentsFromCache([collection queryWhereField:@"match" isEqualTo:@YES]);
     (void)docs;
   }
 
--- a/Firestore/Example/Benchmarks/remote_document_cache_benchmark.mm
+++ b/Firestore/Example/Benchmarks/remote_document_cache_benchmark.mm
@@ -61,10 +61,23 @@ NSMutableDictionary<NSString*, id>* MakeDocumentData() {
   return doc;
 }
 
-FIRQuerySnapshot* GetDocuments(FIRQuery* query) {
+FIRQuerySnapshot* GetDocumentsFromCache(FIRQuery* query) {
   __block FIRQuerySnapshot* result;
   dispatch_semaphore_t done = dispatch_semaphore_create(0);
   [query getDocumentsWithSource:FIRFirestoreSourceCache
+                     completion:^(FIRQuerySnapshot* snap, NSError* error) {
+                       HARD_ASSERT(error == nil, "Failed: %s", MakeString([error description]));
+                       result = snap;
+                       dispatch_semaphore_signal(done);
+                     }];
+  dispatch_semaphore_wait(done, DISPATCH_TIME_FOREVER);
+  return result;
+}
+
+FIRQuerySnapshot* GetDocumentsFromServer(FIRQuery* query) {
+  __block FIRQuerySnapshot* result;
+  dispatch_semaphore_t done = dispatch_semaphore_create(0);
+  [query getDocumentsWithSource:FIRFirestoreSourceServer
                      completion:^(FIRQuerySnapshot* snap, NSError* error) {
                        HARD_ASSERT(error == nil, "Failed: %s", MakeString([error description]));
                        result = snap;
@@ -100,6 +113,40 @@ void Shutdown(FIRFirestore* db) {
   dispatch_semaphore_wait(done, DISPATCH_TIME_FOREVER);
 }
 
+void BM_QueryIndexFree(benchmark::State& state) {
+  int64_t matching_docs = state.range(0);
+  int64_t total_docs = state.range(1);
+
+  FIRFirestore* db = OpenFirestore();
+  auto collection = [db collectionWithPath:MakeNSString("docs-" + CreateAutoId())];
+  WriteDocs(collection, matching_docs, /* match= */ true);
+  WriteDocs(collection, total_docs - matching_docs, /* match= */ false);
+
+  FIRQuery* query = [collection queryWhereField:@"match" isEqualTo:@YES];
+
+  // Query the server to force the target tables to be updated.
+  GetDocumentsFromServer(query);
+
+  for (auto _ : state) {
+    auto docs = GetDocumentsFromCache(query);
+    (void)docs;
+  }
+
+  Shutdown(db);
+}
+BENCHMARK(BM_QueryIndexFree)
+    ->Unit(benchmark::kMicrosecond)
+    ->Args({0, 1})
+    ->Args({1, 1})
+    ->Args({1, 10})
+    ->Args({10, 10})
+    ->Args({1, 100})
+    ->Args({100, 100})
+    ->Args({1, 1000})
+    ->Args({10, 1000})
+    ->Args({100, 1000})
+    ->Args({1000, 1000});
+
 void BM_QueryMatching(benchmark::State& state) {
   int64_t matching_docs = state.range(0);
   int64_t total_docs = state.range(1);
@@ -110,7 +157,8 @@ void BM_QueryMatching(benchmark::State& state) {
   WriteDocs(collection, total_docs - matching_docs, /* match= */ false);
 
   for (auto _ : state) {
-    auto docs = GetDocuments([collection queryWhereField:@"match" isEqualTo:@YES]);
+    auto docs = GetDocumentsFromCache(
+        [collection queryWhereField:@"match" isEqualTo:@YES]);
     (void)docs;
   }
 
@@ -137,7 +185,7 @@ void BM_QueryAll(benchmark::State& state) {
   WriteDocs(collection, total_docs, /* match= */ true);
 
   for (auto _ : state) {
-    auto docs = GetDocuments(collection);
+    auto docs = GetDocumentsFromCache(collection);
     (void)docs;
   }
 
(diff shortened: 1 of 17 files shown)


Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -102,13 +146,22 @@ OptionalMaybeDocumentMap LevelDbRemoteDocumentCache::GetAll(
     it->Seek(LevelDbRemoteDocumentKey::Key(key));
     if (!it->Valid() || !current_key.Decode(it->key()) ||
         current_key.document_key() != key) {
-      results = results.insert(key, absl::nullopt);
+      results.Insert(std::make_pair(key, absl::nullopt));
```
schmidt-sebastian: More of a question regarding the consumers of this method: Could we change the callers so that we don't have to insert `null` values (they could instead look at the existence of the keys). This would simplify this method a bit: We would not need to insert `std::pair`s anymore and `AsyncResults` wouldn't need to use templates anymore (or if it did, there would only be one implementation of it).

wilhuff: It's possible, but messy. It may not be a net win.

The issue is that we use OptionalMaybeDocumentMaps to convey two pieces of information:

  * That a key was requested/affected
  * That a value exists for a key (and may be a NoDocument)

So for example, the local documents view can take this result and pipe it directly into GetLocalViewOfDocuments, which uses the keys of the map to go fetch affected mutations. Without this, we'd have to pass the keys separately to ensure that we find mutations that apply to documents that weren't in the remote document cache.

Similarly, rejecting a mutation batch uses the return value of GetDocuments to indicate the affected documents in a view. If the result doesn't contain the nullopts for things that were deleted, LocalStore::RejectMutationBatch would have to add them back in.

Adjusting this really becomes an exercise in getting https://github.com/firebase/firebase-js-sdk/pull/1886 right, which seems like more work than is warranted right now.

Firestore/Example/Benchmarks/remote_document_cache_benchmark.mm
```diff
@@ -100,6 +113,40 @@ void Shutdown(FIRFirestore* db) {
   dispatch_semaphore_wait(done, DISPATCH_TIME_FOREVER);
 }
 
+void BM_QueryIndexFree(benchmark::State& state) {
+  int64_t matching_docs = state.range(0);
+  int64_t total_docs = state.range(1);
+
+  FIRFirestore* db = OpenFirestore();
+  auto collection = [db collectionWithPath:MakeNSString("docs-" + CreateAutoId())];
+  WriteDocs(collection, matching_docs, /* match= */ true);
```
var-const: Nit: there should be no spaces in the comment indicating the parameter name (here and below).

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -102,13 +146,22 @@ OptionalMaybeDocumentMap LevelDbRemoteDocumentCache::GetAll(
...
     } else {
-      results = results.insert(key, DecodeMaybeDocument(it->value(), key));
+      const std::string& contents = it->value();
+      tasks.Execute([this, &results, key, contents] {
```
var-const: Optional: it looks like you can capture `key` by reference? (it seems like `key` is valid as long as `keys`, which is an argument to this function, is valid)

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -185,15 +239,26 @@ DocumentMap LevelDbRemoteDocumentCache::GetMatching(
...
-      MaybeDocument maybe_doc = DecodeMaybeDocument(it->value(), document_key);
-      if (!query_path.IsPrefixOf(maybe_doc.key().path())) {
+      if (!query_path.IsPrefixOf(document_key.path())) {
         break;
-      } else if (maybe_doc.is_document()) {
-        results = results.insert(maybe_doc.key(), Document(maybe_doc));
       }
+
+      const std::string& contents = it->value();
+      tasks.Execute([this, &results, document_key, contents] {
```
var-const: Same suggestion re. capturing `document_key` by reference.

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -48,12 +51,52 @@ using model::SnapshotVersion;
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    values_.push_back(std::move(value));
+  }
+
+  void Insert(const T& value) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    values_.push_back(value);
+  }
+
+  const std::vector<T>& Result() {
```
var-const: I think that returning a reference to (mutex-) protected data is generally something to avoid. Even with the current very limited interface it could result in potentially misleading code:
```cpp
if (foo.Result().empty()) {
  // Is `foo` actually empty at this point? Not necessarily
  foo.Insert("abc");
}
```
It could become dangerous if the interface were ever to grow (e.g., the classic `if (!foo.empty()) foo.pop();` problem).

One way to achieve that would be to move conversion to the expected end result (`DocumentMap`, etc.) into this class. Given that this class, at this point at least, is very much tailored to the needs of this file and isn't even exposed outside of this translation unit, I don't think it's inappropriate.

Alternatively: how much is the current code buying by doing some insertions on the current thread (the empty ones) and some on the background thread? I'm pretty sure that if all insertions were to be on the background thread, the mutex (and this class by extension) would be unnecessary -- `WaitForAll` would create the necessary "happens-before" relationship (of course, TSan can check these assumptions). I'm not sure if that would be the case with the current code structure, though (with some insertions happening on the current thread).

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -48,12 +51,52 @@ using model::SnapshotVersion;
...
+using util::BackgroundQueue;
+using util::Executor;
+
+/**
+ * An accumulator for results produced asynchronously. This accumulates
+ * values in a vector to avoid contention caused by accumulating into more
+ * complex structures like immutable::SortedMap.
+ */
+template <typename T>
+class AsyncResults {
```
var-const: Also, please wrap this class in an unnamed namespace (wrap the type aliases above too).

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
wilhuff: The issue is that it's background _threads_. The background executor is created with as many threads as are returned from `std::thread::hardware_concurrency`. Recent iPhones have 6 cores. My MacBook Pro has 12 (after hyperthreading). The mutex is required for contention among those threads, not just with the main worker.

As to moving the result-building into this class, that doesn't work well. In one case, we're building an `OptionalMaybeDocumentMap` based on `std::pair<DocumentKey, absl::optional<MaybeDocument>>`. In the other case we're building a `DocumentMap` based on `Document`. There's no unified way to process these. Meanwhile, a for-each style interface is cumbersome with the way our immutable maps accumulate.

The intent here is that all `Insert`s happen before the call to `Result` and that `Result` is only called once, after `BackgroundQueue::AwaitAll` has returned. There's nothing here constraining that to be the case but it's hard to ensure the cooperation of these classes because they're not aware of each other.

I could add a state machine here that prevents calling `Insert` after `Result` has been called but that feels like a hack. Short of that I'm having a hard time coming up with anything that substantively addresses the concern that doesn't add a lot of boilerplate in the caller.

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
wilhuff: This turns out not to work. In the first loop, the reference is is bound to an element that does indeed live in the collection, and has a lifetime that is at least as long as the background thread.

In this case, the reference is bound to the one that `LevelDbRemoteDocumentKey` parsed from the leveldb key for the current row. This reference instance is destroyed on the next iteration.

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
var-const: Ah, good point about multiple threads.

> The intent here is that <...> `Result` is only called once, after `BackgroundQueue::AwaitAll` has returned.

How about returning by value, then?

```cpp
std::vector<T> ConsumeResult() {
  std::lock_guard lock{m_};
  auto result = std::move(result_);
  result_.clear(); // Might be overkill
  return result;
}
```

It solves any potential concurrency problems and as an added benefit reflects the intended usage in the API itself. I don't expect the extra move to matter performance-wise (not to mention that accessing a variable using indirection also in theory could have some extra cost). What do you think?

var-const: Oh, and I think it would enable actually moving the docs into the resulting map.

wilhuff: I made it consume by value.

`immutable::SortedMap` cannot move into key or value. The complexity of making that work was stratospheric and I didn't bother...

Both DocumentKey and Document are reference-count bumps so I'm happy to leave this as-is.

Firestore/core/src/firebase/firestore/local/leveldb_remote_document_cache.cc
```diff
@@ -73,9 +73,10 @@ class AsyncResults {
     values_.push_back(value);
   }
 
-  const std::vector<T>& Result() {
+  std::vector<T> Result() {
     std::lock_guard<std::mutex> lock(mutex_);
-    return values_;
+    std::vector<T> result = std::move(values_);
```
var-const: A couple of final nitpicks:
1. I had an intermediate variable in my code example only so that the member variable could be `clear`ed before exiting the function. If you don't think it's worthwhile, then it could be just `return std::move(values_)`.
2. Can you please add a comment to this function explaining either that it clears the vector or that the object is not supposed to be used again after `Result` is called on it, depending on whichever aligns more with your intention?

PR #4858, opened 2020-02-07
re-enable style script for messaging

chliangGoogle:


paulb777:
The swiftformat version might still be wrong. If you don't have swift files, you could also revert the Firestore change.

PR #4859, opened 2020-02-07
Prep for forward declarations

wilhuff:
Fix a bunch of issues that make it hard to pervasively use forward declarations in our code.

Also, remove some dead code.

### DIFF
--- a/Firestore/core/src/firebase/firestore/core/sync_engine.h
+++ b/Firestore/core/src/firebase/firestore/core/sync_engine.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Firestore/core/src/firebase/firestore/core/sync_engine.h
+++ b/Firestore/core/src/firebase/firestore/core/sync_engine.h
@@ -41,21 +41,7 @@ namespace firebase {
 namespace firestore {
 namespace core {
 
-/**
- * Interface implemented by `EventManager` to handle notifications from
- * `SyncEngine`.
- */
-class SyncEngineCallback {
- public:
-  virtual ~SyncEngineCallback() = default;
-
-  /** Handles a change in online state. */
-  virtual void HandleOnlineStateChange(model::OnlineState online_state) = 0;
-  /** Handles new view snapshots. */
-  virtual void OnViewSnapshots(std::vector<core::ViewSnapshot>&& snapshots) = 0;
-  /** Handles the failure of a query. */
-  virtual void OnError(const core::Query& query, const util::Status& error) = 0;
-};
+class SyncEngineCallback;
 
 /**
  * Interface implemented by `SyncEngine` to receive requests from
--- a/Firestore/core/src/firebase/firestore/core/view_snapshot.h
+++ b/Firestore/core/src/firebase/firestore/core/view_snapshot.h
@@ -97,9 +97,6 @@ class DocumentViewChangeSet {
  */
 class ViewSnapshot {
  public:
-  using Listener = std::unique_ptr<EventListener<ViewSnapshot>>;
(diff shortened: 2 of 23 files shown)


Firestore/core/test/firebase/firestore/immutable/sorted_set_test.cc
```diff
@@ -94,21 +94,6 @@ TEST(SortedSetTest, IteratorsAreDefaultConstructible) {
       "is default constructible");
 }
 
-TEST(SortedSetTest, CanBeConstructedFromSortedMap) {
-  using Map = SortedMap<int, int>;
```
var-const: What's the reason to remove this?

Firestore/core/src/firebase/firestore/core/view_snapshot.h
```diff
@@ -179,6 +176,9 @@ class ViewSnapshot {
   bool excludes_metadata_changes_ = false;
 };
 
+using ViewSnapshotListener = std::unique_ptr<EventListener<ViewSnapshot>>;
```
var-const: Question: how does moving these aliases out of the class help?

Firestore/core/src/firebase/firestore/core/sync_engine.h
```diff
@@ -41,21 +41,7 @@ namespace firebase {
 namespace firestore {
 namespace core {
 
-/**
- * Interface implemented by `EventManager` to handle notifications from
- * `SyncEngine`.
- */
-class SyncEngineCallback {
```
var-const: Did we have it defined twice (here and in `sync_engine_callback.h`)?

Firestore/core/test/firebase/firestore/immutable/sorted_set_test.cc
wilhuff: This feature of being able to construct a `SortedSet` from an arbitrary `SortedMap` required `SortedSet` to have this declaration:

```c++
template <typename K,
          typename C = util::Comparator<K>,
          typename V = util::Empty,
          typename M = SortedMap<K, V, C>>
class SortedSet;
```

And that meant that forward declaring `DocumentKeySet` required all of this:

```c++
namespace util {
struct Empty;

template <typename T>
class Comparator;
}  // namespace util

namespace immutable {
template <typename K, typename V, typename C>
class SortedMap;

template <typename K, typename C, typename V, typename M>
class SortedSet;
}  // namespace immutable

namespace model {
class DocumentKey;

using DocumentKeySet =
    SortedSet<DocumentKey,
              util::Comparator<DocumentKey>,
              util::Empty,
              SortedMap<DocumentKey, util::Empty, util::Comparator<DocumentKey>>>;
}  // namespace model
```

With this simpler signature, forward declarations of `SortedSet` types become much simpler because we no longer have to mention `Empty` or `SortedMap`:

```c++
namespace util {
template <typename T>
class Comparator;
}  // namespace util

namespace immutable {
template <typename K, typename C>
class SortedSet;
}  // namespace immutable

namespace model {
class DocumentKey;

using DocumentKeySet = immutable::SortedSet<DocumentKey, util::Comparator<DocumentKey>>;
}  // namespace model
```

(Note that forward declarations can't have default template arguments because C++ doesn't let you redeclare template defaults. Were to do declare default parameters here we couldn't have defaults on the actual declaration.)

Firestore/core/src/firebase/firestore/core/view_snapshot.h
wilhuff: Nested classes cannot be forward declared. That is, this isn't legal:

```c++
class ViewSnapshot;
using ViewSnapshot::Listener = std::unique_ptr<EventListener<ViewSnapshot>>;
```

However, this is legal:

```c++
class ViewSnapshot;
using ViewSnapshotListener = std::unique_ptr<EventListener<ViewSnapshot>>;
```

Firestore/core/src/firebase/firestore/core/sync_engine.h
wilhuff: Yes.

PR #4863, opened 2020-02-08
Fix Xcode 11.4 analyze error

paulb777:
Fix new analyzer warning from Xcode 11.4

    - WARN  | [iOS] xcodebuild:  /Users/paulbeusterien/gh/firebase-ios-sdk/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTCompressionHelper.m:48:8: warning: Although the value stored to 'retCode' is used in the enclosing expression, the value is never actually read from 'retCode'

### DIFF
--- a/GoogleDataTransportCCTSupport/CHANGELOG.md
+++ b/GoogleDataTransportCCTSupport/CHANGELOG.md
@@ -1,8 +1,6 @@
-# Unreleased
-- Fix an Xcode 11.4 analyze error. (#4863)
-
 # v1.4.1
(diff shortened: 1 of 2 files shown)


GoogleDataTransportCCTSupport/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- Fix an Xcode 11.4 analyze error. (#4863)
```
mikehaney24: Add to v1.4.1 changes. 1.4.1 is currently unreleased

PR #4880, opened 2020-02-11
Delete GDTCORStoredEvent, modify apis and tests

doudounan:
1 Add GDTCORDataFuture as a property of GDTCOREvent
2 Change all GDTCORStoredEvent with GDTCOREvent
3 Refactor GDTCORUploader prioritizeEvent API to use a GDTCOREvent instead of GDTCORStoredEvent
4 Modify tests in GDT and GDTCCT modules


GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -74,8 +75,16 @@ - (void)setDataObject:(id<GDTCOREventDataObject>)dataObject {
...
 
-- (GDTCORStoredEvent *)storedEventWithDataFuture:(GDTCORDataFuture *)dataFuture {
-  return [[GDTCORStoredEvent alloc] initWithEvent:self dataFuture:dataFuture];
+- (void)setFileURL:(NSURL *)fileURL {
+  if (fileURL != _fileURL) {
+    _fileURL = fileURL;
+  }
+}
+
+- (void)clearDataObjectTransportBytes {
```
mikehaney24: I'd replace this with a method: `-writeToURL:` that calls [dataObject transportBytes] and writes the bytes to disk. Doing this allows you to get rid of the dataObjectTransportBytes property altogether. In GDTCORStorage, you'd create a file URL and then just call this method on the event.

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -114,15 +114,17 @@ - (void)storeEvent:(GDTCOREvent *)event
                                         eventHash:event.hash
                                             error:&error];
     GDTCORLogDebug("Event saved to disk: %@", eventFile);
-    GDTCORDataFuture *dataFuture = [[GDTCORDataFuture alloc] initWithFileURL:eventFile];
-    GDTCORStoredEvent *storedEvent = [event storedEventWithDataFuture:dataFuture];
+    [event setFileURL:eventFile];
```
mikehaney24: Change all setFileURL: calls to the more modern dot syntax: `event.fileURL = eventFile;`

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -114,15 +114,17 @@ - (void)storeEvent:(GDTCOREvent *)event
...
     completion(eventFile != nil, error);
 
+    // set dataObject to nil in current event
+    [event clearDataObjectTransportBytes];
```
mikehaney24: event.dataObjectTransportBytes = nil;

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -114,15 +114,17 @@ - (void)storeEvent:(GDTCOREvent *)event
                                         eventHash:event.hash
                                             error:&error];
     GDTCORLogDebug("Event saved to disk: %@", eventFile);
-    GDTCORDataFuture *dataFuture = [[GDTCORDataFuture alloc] initWithFileURL:eventFile];
-    GDTCORStoredEvent *storedEvent = [event storedEventWithDataFuture:dataFuture];
+    [event setFileURL:eventFile];
     completion(eventFile != nil, error);
 
+    // set dataObject to nil in current event
```
mikehaney24: This comment can be deleted, it's pretty clear what's happening

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -82,12 +85,14 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
 - (nullable instancetype)initWithMappingID:(NSString *)mappingID
                                     target:(NSInteger)target NS_DESIGNATED_INITIALIZER;
 
-/** Returns the GDTCORStoredEvent equivalent of self.
+/** Set _fileURL of the event
```
mikehaney24: You can remove this method declaration

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -82,12 +85,14 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
...
  *
- * @param dataFuture The data future representing the transport bytes of the original event.
- * @return An equivalent GDTCORStoredEvent.
+ *@param fileURL The URL of dataObjectTransportBytes stored in the disk
  */
-- (GDTCORStoredEvent *)storedEventWithDataFuture:(GDTCORDataFuture *)dataFuture;
+- (void)setFileURL:(NSURL *)fileURL;
+
+/** Clear the dataObject in event after stored to disk.*/
+- (void)clearDataObjectTransportBytes;
```
mikehaney24: You can also remove this method.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORPrioritizer.h
```diff
@@ -16,10 +16,11 @@
 
 #import <Foundation/Foundation.h>
 
+#import <GoogleDataTransport/GDTCOREvent.h>
 #import <GoogleDataTransport/GDTCORLifecycle.h>
 #import <GoogleDataTransport/GDTCORUploadPackage.h>
 
-@class GDTCORStoredEvent;
+@class GDTCOREvent;
```
mikehaney24: You imported GDTCOREvent--you don't need to forward declare it because you've imported it. I'd do one or the other. Something to think about: do you think adopters of this protocol should have to import the GDTCOREvent header, or should we just do it for them in this header? My opinion: headers that declare protocols should import the headers of classes required by the API so the adopting class only needs to import a single header--but maybe that's not such a good idea.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORUploadPackage.h
```diff
@@ -16,10 +16,11 @@
 
 #import <Foundation/Foundation.h>
 
+#import <GoogleDataTransport/GDTCOREvent.h>
```
mikehaney24: Probably get rid of the header import here

GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.h
```diff
@@ -43,25 +43,25 @@
...
+/** Generates a GDTCOReEvent, complete with a file specified in the eventFileURL property.
  *
  * @param qosTier The QoS tier the event should have.
  * @return A newly allocated fake stored event.
  */
-- (GDTCORStoredEvent *)generateStoredEvent:(GDTCOREventQoS)qosTier;
+- (GDTCOREvent *)generateEvent:(GDTCOREventQoS)qosTier;
 
-/** Generates a GDTCORStoredEvent, complete with a file specified in the eventFileURL property.
+/** Generates a GDTCOReEvent, complete with a file specified in the eventFileURL property.
```
mikehaney24: GDTCOREvent

GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.h
```diff
@@ -43,25 +43,25 @@
  */
 - (instancetype)initWithTarget:(GDTCORTarget)target NS_DESIGNATED_INITIALIZER;
 
-/** Generates a GDTCORStoredEvent, complete with a file specified in the eventFileURL property.
+/** Generates a GDTCOReEvent, complete with a file specified in the eventFileURL property.
```
mikehaney24: GDTCOREvent

GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.h
```diff
@@ -43,25 +43,25 @@
...
  * @param fileURL The file URL containing bytes of some event.
  * @return A newly allocated fake stored event.
  */
-- (GDTCORStoredEvent *)generateStoredEvent:(GDTCOREventQoS)qosTier fileURL:(NSURL *)fileURL;
+- (GDTCOREvent *)generateEvent:(GDTCOREventQoS)qosTier fileURL:(NSURL *)fileURL;
 
 /** Generates five consistent stored events.
  *
- * @return An array of five newly allocated but consistent GDTCORStoredEvents.
+ * @return An array of five newly allocated but consistent GDTCOReEvents.
```
mikehaney24: GDTCOREvents

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -63,6 +62,10 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
 /** The clock snapshot at the time of the event. */
 @property(nonatomic) GDTCORClock *clockSnapshot;
 
+/** The fileURL of clearDataObjectTransportBytes in disk. set when store the transport bytes to
+ * disk,*/
+@property(nonatomic) NSURL *fileURL;
```
mikehaney24: I'd suggest making this readonly so it can only be set by calling `-writeToURL:` (a method you should write, see comment above)

GoogleDataTransport/GDTCORTests/Unit/GDTCOREventTest.m
```diff
@@ -74,4 +74,52 @@ - (void)testArchiving {
   XCTAssertEqual(decodedEvent.clockSnapshot.timezoneOffsetSeconds, timezoneOffsetSeconds);
 }
 
+/** Tests encoding and decoding. */
+- (void)testNSSecureCoding {
```
mikehaney24: This doesn't actually test encoding and decoding. See the test above -testArchiving for an example. This just tests setting variables.

GoogleDataTransport/GDTCORTests/Unit/GDTCORStorageTest.m
```diff
@@ -275,35 +272,31 @@ - (void)testStoreMultipleEvents {
   });
 }
 
-/** Tests enforcing that a prioritizer does not retain an event in memory. */
+/** Tests enforcing that a prioritizer does not retain the DataObjectTransportBytes of an event in
+ * memory.*/
```
mikehaney24: nit: closing `*/` should be on the following line

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCORTestPrioritizer.m
```diff
@@ -15,7 +15,7 @@
  */
 
 #import "GDTCORTests/Unit/Helpers/GDTCORTestPrioritizer.h"
-
+#import <GoogleDataTransport/GDTCOREvent.h>
```
mikehaney24: Imports of different kinds should be grouped in blocks separated by newlines. I'm also not entirely sure that this import is needed.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -23,7 +23,7 @@
 #endif  // TARGET_OS_IOS || TARGET_OS_TV
 
 #import <GoogleDataTransport/GDTCORConsoleLogger.h>
-
+#import <GoogleDataTransport/GDTCOREvent.h>
```
mikehaney24: Separate the import groups with newlines

GoogleDataTransport/GDTCORTests/Unit/GDTCORStorageTest.m
```diff
@@ -16,13 +16,11 @@
...
 #import "GDTCORTests/Unit/GDTCORTestCase.h"
 
-#import <GoogleDataTransport/GDTCOREvent.h>
-#import <GoogleDataTransport/GDTCORStoredEvent.h>
-
 #import "GDTCORLibrary/Private/GDTCOREvent_Private.h"
 #import "GDTCORLibrary/Private/GDTCORRegistrar_Private.h"
 #import "GDTCORLibrary/Private/GDTCORStorage.h"
 #import "GDTCORLibrary/Private/GDTCORStorage_Private.h"
+#import "GDTCORLibrary/Public/GDTCOREvent.h"
```
mikehaney24: Separate the blocks of imports by path. Public headers should be in their own block of imports

mikehaney24:
I'd suggest merging master, because it'll remove the travis check.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -75,15 +75,17 @@ - (void)setDataObject:(id<GDTCOREventDataObject>)dataObject {
...
-  if (fileURL != _fileURL) {
-    _fileURL = fileURL;
-  }
-}
-
-- (void)clearDataObjectTransportBytes {
-  if (_dataObjectTransportBytes != nil) {
-    _dataObjectTransportBytes = nil;
+- (void)writeToURL:(NSURL *)fileURL error:(NSError **)error {
+  if (_dataObjectTransportBytes) {
```
mikehaney24: I'd get rid of the dataObjectTransportBytes property and just call [dataObject transportBytes] here

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -75,15 +75,17 @@ - (void)setDataObject:(id<GDTCOREventDataObject>)dataObject {
...
+    BOOL writingSuccess = [_dataObjectTransportBytes writeToURL:fileURL
+                                                        options:NSDataWritingAtomic
+                                                          error:error];
+    if (!writingSuccess) {
+      GDTCORLogError(GDTCORMCEFileWriteError, @"An event file could not be written: %@", fileURL);
+    } else {
+      _fileURL = fileURL;
```
mikehaney24: If writing fails, we should probably assign nil to _fileURL since that file doesn't exist

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -85,14 +85,13 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
...
  *
- *@param fileURL The URL of dataObjectTransportBytes stored in the disk
+ *  @param fileURL The fileURL that dataObjectTranspotBytes would be writen to.
+ *  @param error The error may be catched during writting to disk.
  */
-- (void)setFileURL:(NSURL *)fileURL;
-
-/** Clear the dataObject in event after stored to disk.*/
-- (void)clearDataObjectTransportBytes;
+- (void)writeToURL:(NSURL *)fileURL error:(NSError **)error;
```
mikehaney24: Make this return YES if writing was successful, NO otherwise.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORTransport.h
```diff
@@ -16,7 +16,7 @@
 
 #import <Foundation/Foundation.h>
 
-#import <GoogleDataTransport/GDTCOREventTransformer.h>
+#import "GDTCOREventTransformer.h"
```
mikehaney24: Public header imports need to be module-relative, so revert this change.

GoogleDataTransport/GDTCORTests/Unit/GDTCOREventTest.m
```diff
@@ -75,19 +75,17 @@ - (void)testArchiving {
 }
 
 /** Tests encoding and decoding. */
```
mikehaney24: Fix the comment

GoogleDataTransport/GDTCORTests/Unit/GDTCOREventTest.m
```diff
@@ -100,7 +98,8 @@ - (void)testIsEqualAndHash {
   [event1.clockSnapshot setValue:@(961141365197) forKeyPath:@"uptime"];
   event1.qosTier = GDTCOREventQosDefault;
   event1.customPrioritizationParams = @{@"customParam1" : @"aValue1"};
-  [event1 setFileURL:[NSURL fileURLWithPath:@"/tmp/fake.txt"]];
+  NSError *error1 = nil;
+  [event1 writeToURL:[NSURL fileURLWithPath:@"/tmp/fake.txt"] error:&error1];
```
mikehaney24: Use `XCTAssertNil(error)` to make sure the test fails if error is non-nil.

GoogleDataTransport/GDTCORTests/Unit/GDTCOREventTest.m
```diff
@@ -110,7 +109,8 @@ - (void)testIsEqualAndHash {
   [event2.clockSnapshot setValue:@(961141365197) forKeyPath:@"uptime"];
   event2.qosTier = GDTCOREventQosDefault;
   event2.customPrioritizationParams = @{@"customParam1" : @"aValue1"};
-  [event2 setFileURL:[NSURL fileURLWithPath:@"/tmp/fake.txt"]];
+  NSError *error2 = nil;
+  [event2 writeToURL:[NSURL fileURLWithPath:@"/tmp/fake.txt"] error:&error2];
```
mikehaney24: Same as above

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -70,12 +70,27 @@ - (void)setDataObject:(id<GDTCOREventDataObject>)dataObject {
   // dispatch_(barrier_ if concurrent)async here, and implement the getter with a dispatch_sync.
   if (dataObject != _dataObject) {
     _dataObject = dataObject;
-    _dataObjectTransportBytes = [dataObject transportBytes];
   }
 }
 
-- (GDTCORStoredEvent *)storedEventWithDataFuture:(GDTCORDataFuture *)dataFuture {
-  return [[GDTCORStoredEvent alloc] initWithEvent:self dataFuture:dataFuture];
+- (bool)writeToURL:(NSURL *)fileURL error:(NSError **)error {
```
mikehaney24: s/bool/BOOL/

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -70,12 +70,27 @@ - (void)setDataObject:(id<GDTCOREventDataObject>)dataObject {
...
+      _fileURL = nil;
+      return NO;
+    } else {
+      _fileURL = fileURL;
+      _dataObject = nil;
+      return YES;
+    }
+  } else {
+    return NO;
+  }
```
mikehaney24: You might make this easier to read by writing as such:

```
  if (dataTransportBytes == nil) {
    _fileURL = nil;
    _dataObject = nil;
    return NO;
  }
  BOOL writingSuccess = [dataTransportBytes writeToURL:fileURL
                                               options:NSDataWritingAtomic
                                                 error:error];
  if (!writingSuccess) {
    GDTCORLogError(GDTCORMCEFileWriteError, @"An event file could not be written: %@", fileURL);
    _fileURL = nil;
    return NO;
  }
  _fileURL = fileURL;
  _dataObject = nil;
  return YES;
```

Generally, if there's an error condition and early exit you want to try and make that logic apparent. In Swift they have 'guard' statements to help deal with this, but in ObjC we can only write code like this.

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -82,12 +85,14 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
...
+ * and set dataObject to nil.
  *
- * @param dataFuture The data future representing the transport bytes of the original event.
- * @return An equivalent GDTCORStoredEvent.
+ * @param fileURL The fileURL that dataObject would be writen to.
+ * @param error The error may be catched during writting to disk.
+ * @return YES if writing dataObject to disk is succeed, NO if not.
  */
-- (GDTCORStoredEvent *)storedEventWithDataFuture:(GDTCORDataFuture *)dataFuture;
+- (bool)writeToURL:(NSURL *)fileURL error:(NSError **)error;
```
mikehaney24: BOOL instead of bool

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -82,12 +85,14 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
...
 
-/** Returns the GDTCORStoredEvent equivalent of self.
+/** Write the dataObject to given URL, assign given URL to self->fileURL,
+ * and set dataObject to nil.
  *
- * @param dataFuture The data future representing the transport bytes of the original event.
- * @return An equivalent GDTCORStoredEvent.
+ * @param fileURL The fileURL that dataObject would be writen to.
+ * @param error The error may be catched during writting to disk.
+ * @return YES if writing dataObject to disk is succeed, NO if not.
```
mikehaney24: @return YES if writing dataObject to disk was successful, NO otherwise.

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -82,12 +85,14 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
 - (nullable instancetype)initWithMappingID:(NSString *)mappingID
                                     target:(NSInteger)target NS_DESIGNATED_INITIALIZER;
 
-/** Returns the GDTCORStoredEvent equivalent of self.
+/** Write the dataObject to given URL, assign given URL to self->fileURL,
+ * and set dataObject to nil.
```
mikehaney24: Writes [dataObject transportBytes] to the given URL, populates fileURL with the filename, then nils the dataObject property.

GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -63,6 +62,10 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
 /** The clock snapshot at the time of the event. */
 @property(nonatomic) GDTCORClock *clockSnapshot;
 
+/** The fileURL of dataObject in disk. set when store the transport bytes to
+ * disk,*/
```
mikehaney24: disk.
*/

GoogleDataTransport/GDTCORTests/Unit/GDTCOREventTest.m
```diff
@@ -67,11 +67,60 @@ - (void)testArchiving {
...
   XCTAssertEqual(decodedEvent.target, 42);
-  XCTAssertEqualObjects(decodedEvent.dataObjectTransportBytes,
-                        [@"someData" dataUsingEncoding:NSUTF8StringEncoding]);
+  event.dataObject = [[GDTCORDataObjectTesterSimple alloc] initWithString:@"someData"];
   XCTAssertEqual(decodedEvent.qosTier, GDTCOREventQoSTelemetry);
   XCTAssertEqual(decodedEvent.clockSnapshot.timeMillis, timeMillis);
   XCTAssertEqual(decodedEvent.clockSnapshot.timezoneOffsetSeconds, timezoneOffsetSeconds);
 }
 
+/** Tests testSettingVariables. */
```
mikehaney24: This comment should be a description of what the test is testing, not the test's name.

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCORDataObjectTesterClasses.h
```diff
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
 /** A string that will be turned into bytes. */
 @property(nonatomic) NSString *aString;
 
+/** Initializes an instance using the given String.
```
mikehaney24: s/String/string/

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCORDataObjectTesterClasses.h
```diff
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @param aString The give String.
```
mikehaney24: A string whose UTF8 encoded bytes will be used for -transportBytes.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-#import <GoogleDataTransport/GDTCOREvent.h>
+#import "GDTCORLibrary/Public/GDTCOREvent.h"
```
chliangGoogle: for public header you might have to use this format? 
#import <GoogleDataTransport/GDTCOREvent.h>

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -56,8 +56,10 @@ - (NSUInteger)hash {
   // This loses some precision, but it's probably fine.
   NSUInteger mappingIDHash = [_mappingID hash];
   NSUInteger timeHash = [_clockSnapshot hash];
-  NSUInteger dataObjectTransportBytesHash = [_dataObjectTransportBytes hash];
-  return mappingIDHash ^ _target ^ dataObjectTransportBytesHash ^ _qosTier ^ timeHash;
+  NSInteger dataObjectHash = [_dataObject hash];
```
chliangGoogle: if it's a property, use dot access:  _dataObject.hash

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -329,11 +319,11 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
 
 - (void)encodeWithCoder:(NSCoder *)aCoder {
   GDTCORStorage *sharedInstance = [self.class sharedInstance];
-  NSMutableOrderedSet<GDTCORStoredEvent *> *storedEvents = sharedInstance->_storedEvents;
+  NSMutableOrderedSet<GDTCOREvent *> *storedEvents = sharedInstance->_storedEvents;
```
chliangGoogle: do we need to worry about backward compatibility now that the encoding and decoding types have changed?
say previous version we've encoded GDTCORStoreEvent type in the files and now read it as GDTCOREvent.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: You only need to use that format in other public headers, here's it's fine.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: Doudou is just being consistent with what I do in other places. Since it's intra-class, ivar access is faster and the consequence is having KVO not work (which is fine, IMO).

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
mikehaney24: That is a very good question. It might not be backwards compatible, so we should need to change the encoding key for all fields that we've altered the class of.

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCORDataObjectTesterClasses.h
```diff
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
 /** A string that will be turned into bytes. */
 @property(nonatomic) NSString *aString;
 
+/** Initializes an instance using the given string.
+ *
+ * @param aString A string whose UTF8 encoded bytes will be used for -transportBytes..
```
mikehaney24: Extra period

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCOREventGenerator.m
```diff
@@ -15,10 +15,11 @@
  */
 
 #import "GDTCORTests/Unit/Helpers/GDTCOREventGenerator.h"
+#import "GDTCORTests/Unit/Helpers/GDTCORDataObjectTesterClasses.h"
```
mikehaney24: Move this down just below or just above the `#import "GDTCORLibrary/Private/GDTCOREvent_Private.h"`. This first line (or sometimes 2 lines) of import should be reserved for the corresponding header and/or private header.

http://go/objc-style#order-of-includes

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTPrioritizer.m
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#import <GDTCCTLibrary/Private/GDTCCTPrioritizer.h>
+#import "GDTCCTLibrary/Private/GDTCCTPrioritizer.h"
```
mikehaney24: Add a newline underneath this line.

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
chliangGoogle: I don't see this is being resolved. I guess it's a function question that would it be okay if such data is lost?
do we need test an old => new version transition to make sure the transition doesn't cause any error because of the inconsistency? Maybe it just returns null if it can't find the right format?

mikehaney24: I think it might be ok if data is lost here. When we transition to sqlite, this will occur again and then our migration story will be much more robust. We should definitely test the transition to make sure we understand the effects. My suggestion would be to base64 encode the bytes output of NSCoding singletons from a previous version of the SDK and attempt to read them in the newer version.

mikehaney24: I've written a regression test for Doudou and we ran through it together. The code currently, **with** her coding key name changes does not crash, it crashes **without** the coding key name changes, and it drops all events. Doudou is going to look into doing a class name migration to get the old data into the new format, but I've given her the greenlight to move forward with the data being lost if it can't be done.

doudounan: In order to decode the fieURL property in dataFuture of storedEvent objects, I recover the GDTCORDataFuture class and set its header in private folder for only decoding usage. Please review the newest commit.

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -224,12 +215,12 @@ - (NSURL *)saveEventBytesToDisk:(NSData *)transportBytes
  *
  * @param event The event to track.
  */
-- (void)addEventToTrackingCollections:(GDTCORStoredEvent *)event {
+- (void)addEventToTrackingCollections:(GDTCOREvent *)event {
   [_storedEvents addObject:event];
-  NSMutableSet<GDTCORStoredEvent *> *events = self.targetToEventSet[event.target];
+  NSMutableSet<GDTCOREvent *> *events = self.targetToEventSet[@(event.target)];
```
mikehaney24: save @(event.target) into a NSNumber variable so you only have to convert it to an NSNumber once

GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -309,15 +300,19 @@ + (BOOL)supportsSecureCoding {
 }
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder {
+  /**Sets a global translation mapping to decode GDTCORStoredEvent objects encoded as instances of
```
mikehaney24: These comments can begin with // instead of /**. Also, maintain a space between the comment delimiter and the beginning of your comment.

GoogleDataTransport/GDTCORTests/Unit/GDTCORStorageTest.m
```diff
@@ -275,35 +279,32 @@ - (void)testStoreMultipleEvents {
...
-                                                       XCTAssertNil(error);
-                                                       [writtenExpectation fulfill];
-                                                     }]);
-    [self waitForExpectations:@[ writtenExpectation ] timeout:10.0];
-    GDTCORDataFuture *dataFuture =
-        [[GDTCORDataFuture alloc] initWithFileURL:[NSURL fileURLWithPath:@"/test"]];
-    storedEvent = [event storedEventWithDataFuture:dataFuture];
-  }
+  __weak NSData *weakDataObjectTransportBytes;
+  GDTCOREvent *event = [[GDTCOREvent alloc] initWithMappingID:@"404" target:target];
```
mikehaney24: Similar to how the test was written before, wrap in an autoreleasepool because otherwise there will be a race condition in this test that will manifest as flakiness. (Sometimes the event will be deallocated, sometimes it won't--it'll be a race between dealloc and the dealloc check.)

PR #4892, opened 2020-02-12
Fixes to get crash report to be successfully processed in the new report endpoint

jasonhu-g:
Fixes:
1. Switch from using proto3 -> proto2. With proto3, fields with default values were being omitted which was an issue since the fields are required in the backend (proto2). 
2. Crash fatal event type from "crashed" -> "crash"
3. Set the thread and frame importance else the blame frame is incorrect and the backend will detect the event to be an internal issue.
4. Set the required organization id field in the proto.


Crashlytics/Protogen/nanopb/crashlytics.nanopb.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google
+ * Copyright 2019 Google
```
samedson: 2020

Crashlytics/Protogen/nanopb/crashlytics.nanopb.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google
+ * Copyright 2019 Google
```
samedson: 2020

PR #4894, opened 2020-02-12
Crashlytics GDT fill in Exception part of Proto

samedson:
- Sets the fields in the `Exception` part of the app execution.
 - Also fixes the analysis issue with `signalProto`

Does not fill in symbolicated files or importance for symbolicated.

### DIFF
--- a/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
+++ b/Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
@@ -672,9 +672,8 @@ const NSUInteger IMPORTANCE_FOR_EXCEPTION = 0;
 
 - (google_crashlytics_Session_Event_Application_Execution_Thread_Frame *)protoFramesWithFrames:
     (NSArray<FIRCLSRecordFrame *> *)frames {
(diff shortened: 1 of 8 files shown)


Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -26,6 +26,8 @@
 #import <nanopb/pb_decode.h>
 #import <nanopb/pb_encode.h>
 
+const NSUInteger IMPORTANCE_FOR_EXCEPTION = 0;
```
jasonhu-g: The naming for this may be confusing since the importance for exceptions be non-zero. Take a look at Protobuf.scala:384.

Maybe this should be DEFAULT_IMPORTANCE_FOR_EXCEPTION?

jasonhu-g: In FIRCLSRecordThread.m:21, we define IMPORTANCE_IN_CRASHED_THREAD. Does it make more sense to have this const be defined FIRCLSRecordFrame?

samedson: Yeah that probably makes more sense to organize it all there

Issue #4734, opened 2020-01-23
Remote config returning empty values


adrianortuzar:
Hi, we are realising that ours customers that transferred their data from a previous iPhone device to a new iPhone (more info here https://support.apple.com/en-us/HT210216) our app is not working correctly.

We reproduce the issue and all the values from Remote Config are empty. 

To solve the issue we have to ask ours customers to remove the app and install it again. What is not nice. Could you please take a look the issue and let us know how to solve it? Thanks

We are using those versions from Firebase:

```
- Crashlytics (3.12.0):
    - Fabric (~> 1.9.0)
  - Fabric (1.9.0)
  - Firebase (6.12.0):
    - Firebase/Core (= 6.12.0)
  - Firebase/Core (6.12.0):
    - Firebase/CoreOnly
    - FirebaseAnalytics (= 6.1.5)
  - Firebase/CoreOnly (6.12.0):
    - FirebaseCore (= 6.3.3)
  - Firebase/Messaging (6.12.0):
    - Firebase/CoreOnly
    - FirebaseMessaging (~> 4.1.8)
  - Firebase/Performance (6.12.0):
    - Firebase/CoreOnly
    - FirebasePerformance (~> 3.1.6)
  - Firebase/RemoteConfig (6.12.0):
    - Firebase/CoreOnly
    - FirebaseRemoteConfig (~> 4.4.4)
  - FirebaseABTesting (3.1.2):
    - FirebaseAnalyticsInterop (~> 1.3)
    - FirebaseCore (~> 6.1)
    - Protobuf (>= 3.9.2, ~> 3.9)
  - FirebaseAnalytics (6.1.5):
    - FirebaseCore (~> 6.3)
    - FirebaseInstanceID (~> 4.2)
    - GoogleAppMeasurement (= 6.1.5)
    - GoogleUtilities/AppDelegateSwizzler (~> 6.0)
    - GoogleUtilities/MethodSwizzler (~> 6.0)
    - GoogleUtilities/Network (~> 6.0)
    - "GoogleUtilities/NSData+zlib (~> 6.0)"
    - nanopb (= 0.3.9011)
  - FirebaseAnalyticsInterop (1.4.0)
  - FirebaseCore (6.3.3):
    - FirebaseCoreDiagnostics (~> 1.0)
    - FirebaseCoreDiagnosticsInterop (~> 1.0)
    - GoogleUtilities/Environment (~> 6.2)
[...]

Issue #4677, opened 2020-01-14
Restore system backup gives problem


SergeShkurko:
### Environment

  * Xcode version: 11.3.1
  * Firebase SDK version: 6.14.0
  * Firebase Component: FirebaseRemoteConfig
  * Component version: 4.4.6

### Description

I having an app with integrated Firebase remote config. I take a backup of all application and restore it again in the same iPhone/other iPhone. After the restoring backup, Firebase remote config is stopped working until the reinstall app again from AppStore, So some depended functions are not working in the live app.

The app is working normally after reinstalling from AppStore.

Getting error from Firebase Remote config is The operation couldn't be completed. (com.google.remoteconfig.ErrorDomain error 8003.)

Same issue [on stackoverflow](https://stackoverflow.com/questions/56495163/firebase-remote-config-is-stopped-working-after-restore-the-backup-in-iphone)

#### Steps to reproduce:

- Run app (remote config fetched success)
- Backup ios system
- Restore backup ios system
- Run app (remote config gives 0 keys)

rizafran:
Hi @SergeShkurko, apologies for the  late response.  I have filed this issue internally tracked at (b/148386739).

Goseong:
We are also experiencing this problem, any updates here? We have some key app feature relying on a valid remote config.

rizafran:
@Goseong, I'm sorry to hear if this issue is negatively affecting your users. I have made a follow up internally, and we'll inform you as soon as we have an update.

rizafran:
Hi all, this issue has been fixed in the latest release of SDK as shown [here](https://firebase.google.com/support/release-notes/ios#remote-config). Feel free to comment here if you're still experiencing an issue on restoring your backup, and we'll reopen this thread to continue investigating it.

Goseong:
Great, thank you for the work!

PR #4896, opened 2020-02-12
Reset RC per-instance userdefaults if config database was not found.

dmandar:
Addresses b/148386739, #4677 , #4734. Backgroubd: We explicitly disable backup of the sqlite database. Hence upon restore, we do not have any data. This causes a fetch to be made, but we send the last eTag (which is saved in userdefaults and cannot be opted out of backup) along with the fetch request which causes the RC backend to send a NO_CHANGE back. Fix is to identify if the database is gone, and if so, reset the userdefaults.

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
@@ -31,6 +31,7 @@
 #define RCNTableNameInternalMetadata "internal_metadata"
 #define RCNTableNameExperiment "experiment"
 
+static BOOL gIsNewDatabase;
 /// SQLite file name in versions 0, 1 and 2.
 static NSString *const RCNDatabaseName = @"RemoteConfig.sqlite3";
 /// The application support sub-directory that the Remote Config database resides in.
@@ -77,6 +78,7 @@ static BOOL RemoteConfigCreateFilePathIfNotExist(NSString *filePath) {
   }
   NSFileManager *fileManager = [NSFileManager defaultManager];
   if (![fileManager fileExistsAtPath:filePath]) {
+    gIsNewDatabase = YES;
     NSError *error;
     [fileManager createDirectoryAtPath:[filePath stringByDeletingLastPathComponent]
            withIntermediateDirectories:YES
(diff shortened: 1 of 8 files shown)


chliangGoogle:
LGTM on presubmit green

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
```diff
@@ -31,6 +31,7 @@
 #define RCNTableNameInternalMetadata "internal_metadata"
 #define RCNTableNameExperiment "experiment"
 
+static BOOL gIsNewDatabase;
```
paulb777: Does this need to be made thread-safe?

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
```diff
@@ -102,6 +102,14 @@ - (instancetype)initWithDatabaseManager:(RCNConfigDBManager *)manager
     _userDefaultsManager = [[RCNUserDefaultsManager alloc] initWithAppName:appName
                                                                   bundleID:_bundleIdentifier
                                                                  namespace:_FIRNamespace];
+
+    // Check if the config database is new. If so, clear the configs saved in userDefaults.
+    if ([_DBManager isNewDatabase]) {
```
paulb777: How can this test ever fail? I don't see gIsNewDatabase ever set to NO.

dmandar: The test case setup method above creates a new test database. We just want to verify that this flag is set to true whenever an existing database is not found.

chliangGoogle:
@ryanwilson  nvm, I was adding you and wondering if NSUserDefault can opt out for backup, looks like it's not available.

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
dmandar: it is. It runs on the database serial queue.

dmandar:
> Just to confirm, the issue happens when users uninstall the app and reinstall again but we thought config is fetched because user default is not reset?

It happens on a restore of a previous backup of the device. The config database is gone, but the last eTag is still remembered (since it is in the user defaults). Hence, a new fetch call succeeds but only returns NO_CHANGE since the eTag was sent up.

Issue #3687, opened 2019-08-26
AB tests not working when sending activation event right after activating remote config


domagojb:
### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 10.3
  * Firebase SDK version: 6.6.0
  * Firebase Component: Remote Config/AB Testing
  * Component version: 4.2.1/3.0.0

### [REQUIRED] Step 3: Describe the problem

We have an ongoing issue since last year where activating an experiment (i.e. sending an activation event) exactly after calling `[[FIRRemoteConfig remoteConfig] activateFetched]` didn't work. That is because activateFetched presumably did some property setting related to AB tests in the background. We figured that out by logging firebase messages to the console.

We found that also to be the case because doing this 

```
[[FIRRemoteConfig remoteConfig] activateFetched];

// This is a workaround for A/B tests not working when the activation event is sent too soon after `activateFetched`.
// Basically we delay further UI initialization and Analytics event dispatch enough that Firebase AB Testing framework has
// time to set-up a conditional user property internally used for marking ab test behaviour with activation event.
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{

    // send activation event here
});
```

made it work. But obviously this is not the way we wan't to use the API.

We were thrilled to see that `fetchAndActivateWithCompletionHandler` which should have async completion was added. But it seems the issue is still not fixed. AB tests still don't get activated.  

I was looking at the newly open sourced code of Remote Config and AB testing and I think I can pin point the problem:

```
**fetchAndActivateWithCompletionHandler** calls **activateFetched** which in turn calls **activateWithCompletionHandler**

**activateWithCompletionHandler**  calls **[strongSelf updateExperiments]** which does not have a completion handler but if you follow the  updateExperiments method down the rabbit hole you'll find a call like this in **FIRExperimentController**

- (void)updateExperimentsWithServiceOrigin:(NSString *)origin
                                    events:(FIRLifecycleEvents *)events
                                    policy:(ABTExperimentPayload_ExperimentOverflowPolicy)policy
                             lastStartTime:(NSTimeInterval)lastStartTime
                                  payloads:(NSArray<NSData *> *)payloads {
  FIRExperimentController *__weak weakSelf = self;
  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
[...]

PR #4897, opened 2020-02-12
Add a completion handler to FIRExperimentController's updateExperimentsWithServiceOrigin: method

christibbs:
Completion handler takes a nullable error parameter.

Internal bug ID 141921647.

Partial fix for #3687.


FirebaseABTesting/Sources/ABTConstants.h
```diff
@@ -33,3 +33,11 @@ static NSString *const kABTEventDictionaryNameKey = @"name";
 static NSString *const kABTEventDictionaryOriginKey = @"origin";
 static NSString *const kABTEventDictionaryParametersKey = @"parameters";
 static NSString *const kABTEventDictionaryTimestampKey = @"timestamp";
+
+#pragma mark - Errors
+
+static NSString *const kABTErrorDomain = @"com.google.abt";
+
```
morganchen12: nit: If this is a public header, you should add a doc comment to both the declaration and the value. If not ignore this comment.

FirebaseABTesting/Sources/ABTConstants.h
```diff
@@ -33,3 +33,11 @@ static NSString *const kABTEventDictionaryNameKey = @"name";
 static NSString *const kABTEventDictionaryOriginKey = @"origin";
 static NSString *const kABTEventDictionaryParametersKey = @"parameters";
 static NSString *const kABTEventDictionaryTimestampKey = @"timestamp";
+
+#pragma mark - Errors
+
+static NSString *const kABTErrorDomain = @"com.google.abt";
```
morganchen12: maybe `com.firebase.abtesting`?

FirebaseABTesting/Sources/Public/FIRExperimentController.h
```diff
@@ -45,11 +46,15 @@ NS_SWIFT_NAME(ExperimentController)
...
+/// @param completionHandler Code to be executed after experiments are updated in the background
+/// thread.
 - (void)updateExperimentsWithServiceOrigin:(NSString *)origin
                                     events:(FIRLifecycleEvents *)events
                                     policy:(ABTExperimentPayload_ExperimentOverflowPolicy)policy
                              lastStartTime:(NSTimeInterval)lastStartTime
-                                  payloads:(NSArray<NSData *> *)payloads;
+                                  payloads:(NSArray<NSData *> *)payloads
+                         completionHandler:
+                             (nullable void (^)(NSError *_Nullable error))completionHandler;
```
morganchen12: Definitely note that this is a breaking API change in the release notes.

FirebaseABTesting/Tests/Unit/FIRExperimentControllerTest.m
```diff
@@ -186,9 +191,13 @@ - (void)testUpdateExperiments {
...
                                                        ABTExperimentPayload_ExperimentOverflowPolicy_DiscardOldest  // NOLINT
                                             lastStartTime:now
-                                                 payloads:payloads];
+                                                 payloads:payloads
+                                        completionHandler:^(NSError *_Nullable error) {
+                                          completionHandlerCalled = YES;
+                                        }];
 
   XCTAssertEqual([_mockCUPController experimentsWithOrigin:gABTTestOrigin].count, 2);
+  XCTAssertTrue(completionHandlerCalled);
```
morganchen12: This seems flaky if the method above is async.

FirebaseABTesting/Tests/Unit/FIRExperimentControllerTest.m
```diff
@@ -326,13 +336,25 @@ - (void)testInvalidExperiments {
...
+                                        completionHandler:^(NSError *_Nullable error) {
+                                          if (error &&
+                                              error.code ==
+                                                  kABTInternalErrorFailedToFetchConditionalUserProperties) {
+                                            completionHandlerWithErrorCalled = YES;
+                                          }
+                                        }];
+
+  // Verify completion handler is still called.
+  XCTAssertTrue(completionHandlerWithErrorCalled);
```
morganchen12: Same as above comment

FirebaseABTesting/Sources/ABTConstants.h
christibbs: This isn't a public header.

FirebaseABTesting/Tests/Unit/FIRExperimentControllerTest.m
christibbs: I renamed this method, the method itself is not actually async... it's just called from a background thread. Name was confusing.

FirebaseABTesting/Tests/Unit/FIRExperimentControllerTest.m
christibbs: Same as above.

FirebaseABTesting/Sources/Public/FIRExperimentController.h
paulb777: Does the breaking change here mean that this PR should be tagged for Firebase 7 and merged then?

dmandar: A corresponding RC change will need to be made. Can we support the existing method so that the change is not tied with the same RC release?

morganchen12: Consider adding a new method with this new signature, and in the old method delegating to the new method with a completion handler of `nil`.

If you want to you can remove and deprecate the old method later but imo it's fine to leave as a convenience.

christibbs: Sounds good.

FirebaseABTesting/Sources/Public/FIRExperimentController.h
```diff
@@ -45,6 +46,30 @@ NS_SWIFT_NAME(ExperimentController)
...
+///                       Firebase Analytics FIREventOrigins.h.
+/// @param events         A list of event names to be used for logging experiment lifecycle events,
+///                       if they are not defined in the payload.
+/// @param policy         The policy to handle new experiments when slots are full.
+/// @param lastStartTime  The last known experiment start timestamp for this affected service.
+///                       (Timestamps are specified by the number of seconds from 00:00:00 UTC on 1
+///                       January 1970.).
+/// @param payloads       List of experiment metadata.
+/// thread.
 - (void)updateExperimentsWithServiceOrigin:(NSString *)origin
```
paulb777: Should this method be deprecated?

FirebaseABTesting/Sources/FIRExperimentController.m
```diff
@@ -195,6 +195,19 @@ - (void)updateExperimentsWithServiceOrigin:(NSString *)origin
...
+                                    events:(FIRLifecycleEvents *)events
+                                    policy:(ABTExperimentPayload_ExperimentOverflowPolicy)policy
+                             lastStartTime:(NSTimeInterval)lastStartTime
+                                  payloads:(NSArray<NSData *> *)payloads {
+  [self updateExperimentsWithServiceOrigin:origin
+                                    events:events
+                                    policy:policy
+                             lastStartTime:lastStartTime
+                                  payloads:payloads
+                         completionHandler:nil];
```
morganchen12: nit: Consider FIRLogging the error, so users who tab autocomplete into using this method or use other SDKs that call this method don't have the error omitted from logs.

christibbs: Users will get the error logged regardless of whether there's a completion handler (see line 230 of `FIRExperimentController.m` which is prior to the completionHandler nil check).

PR #4899, opened 2020-02-13
Remove deprecated FIAMDisplay from repo

### LINKED ISSUE #4728: Revert FIAMDisplay Zip Builder workaround
Revert the FIAMDisplay Zip Builder workaround #4727 once [FirebaseInAppMessagingDisplay](https://github.com/firebase/firebase-ios-sdk/blob/48aa5ed5ee30d01a184e757df806169195df00df/FirebaseInAppMessagingDisplay.podspec) is removed.

paulb777:
Fix #4728

The FirebaseInAppMessagingDisplay pod has been merged into FirebaseInAppMessaging.

#no-changelog


Firebase.podspec
```diff
@@ -105,12 +105,6 @@ Simplify your iOS development, grow your user base, and monetize more effectivel
     ss.ios.deployment_target = '9.0'
   end
 
-  s.subspec 'InAppMessagingDisplay' do |ss|
```
ryanwilson: I'm not sure we want to remove it from the `Firebase.podspec` yet, right? It has only been a single release so far, so if someone missed last release and updates to this one the error won't be nearly as user friendly. Or maybe more accurately, they'll be locked to the previous Firebase version?

paulb777: They're already locked to the currently published deprecated release.  We have not plans to do future releases, so no reason to keep it in the repository.

ryanwilson: But if someone still has `Firebase/InAppMessaging` in their podfile and we publish a new version of the `Firebase` pod without this subspec, what will happen? Will CocoaPods stay on the newest version that includes that subspec, will it try to get the newest `Firebase` without this subspec and fail, something else?

paulb777: Oh sorry. Now I see your point:
```

[!] CocoaPods could not find compatible versions for pod "Firebase/InAppMessagingDisplay":
  In Podfile:
    Firebase/InAppMessagingDisplay (from `../../../`)

None of your spec sources contain a spec satisfying the dependency: `Firebase/InAppMessagingDisplay (from `../../../`)`.

You have either:
```

paulb777:
Opened #4900 to come back to the Firebase pod subspec removal at Firebase 7.

PR #4901, opened 2020-02-13
Add union_with to sorted set.

wu-hui:
This is porting https://github.com/firebase/firebase-js-sdk/issues/2620.

### DIFF
--- a/Firestore/core/src/firebase/firestore/immutable/sorted_set.h
+++ b/Firestore/core/src/firebase/firestore/immutable/sorted_set.h
@@ -75,13 +75,13 @@ class SortedSet : public SortedContainer {
   }
 
   ABSL_MUST_USE_RESULT SortedSet union_with(const SortedSet& other) const {
-    auto result_ptr = const_cast<SortedSet*>(this);
-    auto other_ptr = const_cast<SortedSet*>(&other);
+    const SortedSet* result_ptr = this;
+    const SortedSet* other_ptr = &other;
 
     // Make sure `result_ptr` always points to the larger one of the two sets.
     if (result_ptr->size() < other_ptr->size()) {
       result_ptr = other_ptr;
-      other_ptr = const_cast<SortedSet*>(this);
+      other_ptr = this;
     }
 
     auto result = *result_ptr;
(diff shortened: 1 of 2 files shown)


Firestore/core/src/firebase/firestore/immutable/sorted_set.h
```diff
@@ -74,6 +74,22 @@ class SortedSet : public SortedContainer {
     return SortedSet{map_.insert(key, {})};
   }
 
+  ABSL_MUST_USE_RESULT SortedSet union_with(const SortedSet<K>& other) const {
```
wilhuff: I think you can drop the `<K>` here. You get it for free by virtue of being in the declaration of `SortedSet<K>`.

Firestore/core/src/firebase/firestore/immutable/sorted_set.h
```diff
@@ -74,6 +74,22 @@ class SortedSet : public SortedContainer {
...
+    const auto result_ptr = this;
```
wilhuff: Optional: since the type isn't onerous here (`const SortedSet*`), consider spelling it out. While explicit types are somewhat annoying they help reduce the cognitive load on the reader.

wu-hui: Here actually requires casting, which made `auto` seems OK now.

The PR does not build yet when you reviewed. But I also did not assign it yet, so it's not my fault :-)

Firestore/core/src/firebase/firestore/immutable/sorted_set.h
wu-hui: Done.

Firestore/core/src/firebase/firestore/immutable/sorted_set.h
```diff
@@ -74,6 +74,23 @@ class SortedSet : public SortedContainer {
     return SortedSet{map_.insert(key, {})};
   }
 
+  ABSL_MUST_USE_RESULT SortedSet union_with(const SortedSet& other) const {
+    auto result_ptr = const_cast<SortedSet*>(this);
```
wilhuff: `const_cast` is something to generally avoid. It's sometimes required when interoperating with broken C APIs but is something we should otherwise not use. Why not:

```c++
const SortedSet* result_ptr = this;
const SortedSet* other_ptr = &other;

if (...) {
  // swap
}

SortedSet result = *result_ptr;
...
```

PR #4903, opened 2020-02-13
Crashlytics GDT handle on-device symbolication

samedson:
macOS symbolicates on-device


samedson:
Going with https://github.com/firebase/firebase-ios-sdk/pull/4931 instead

PR #4907, opened 2020-02-14
Fix nullability of configValueForKey:source:

pyrtsa:
The implementation of `-[FIRRemoteConfig configValueForKey:namespace:source:]` returns `nil` if no such key exists.

Accordingly, this fix in nullability annotation avoids potential crashes in Swift apps and allows for distinguishing between empty (`""`) and missing remote config values (`nil`).

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
+++ b/FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
@@ -271,16 +271,18 @@ NS_SWIFT_NAME(RemoteConfig)
 /// Gets the config value of a given namespace and a given source.
 /// @param key              Config key.
 /// @param source           Config value source.
-- (nonnull FIRRemoteConfigValue *)configValueForKey:(nullable NSString *)key
-                                             source:(FIRRemoteConfigSource)source;
+/// @return                 Config value if the key exists in given source. Otherwise `nil`.
+- (nullable FIRRemoteConfigValue *)configValueForKey:(nullable NSString *)key
+                                              source:(FIRRemoteConfigSource)source;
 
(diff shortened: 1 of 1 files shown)


morganchen12:
@dmandar this seems like a real API issue.

dmandar:
This would be a breaking change for developers. What is the cause of the crash?

morganchen12:
In Swift, the type is bridged as `RemoteConfigValue` instead of `Optional<FIRRemoteConfigValue>`, which will cause a crash on access if the value is nil.

We should make this change for Firebase 7.

morganchen12:
Actually, currently it looks like the only method that can return nil is `configValueForKey:namespace:source:`, which is deprecated. We can address this if it comes up again, but for now I think it's ok to take no action.

pyrtsa:
The behaviour is the same for `configvalueforkey:source:` which calls the deprecated method as its implementation.

morganchen12:
You're right, and that method is not deprecated. @dmandar see [this declaration](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h#L274)

pyrtsa:
But speaking of source-compatibility (or breaking changes), a non-breaking change would be to keep the methods `nonnull`-returning but fix the implementation to always return an instance, constructed in the fallback case with `[NSData data]` as data.

In that case, I'd recommend (feature request!) adding a `BOOL`-returning method for testing whether a given key exists in a given source:

```objc
-(BOOL)hasKey:(nonnull NSString*)key inSource:(FIRRemoteConfigSource)source;
```

PR #4908, opened 2020-02-14
FirebaseInstallations import added to Firebase.h

maksymmalyhin:
#no-changelog


ryanwilson:
This doesn't have to wait for the Firebase 7 milestone - can be merged whenever, but it may be a reasonable time to add it for convenience.

ryanwilson:
If this is going to be added to the umbrella header we should also include it to the Firebase umbrella pod and the Swift Package Manifest as a product.

paulb777 (approved):
LGTM

The Firebase.podspec and Package.swift can be done here or in a separate PR.

maksymmalyhin:
Looks like SwiftPM product has been added already (see [here](https://cs.opensource.google/firebase-sdk/firebase-ios-sdk/+/master:Package.swift)).

paulb777 (approved):

Issue #4906, opened 2020-02-14
App Crashes On Calling FirebaseApp.app()?.delete


Aishwaryalaxmi:
On calling the FirebaseApp.app()?.delete **kFirebaseInstallationsErrorDomain** will be thrown and application will be crashed. Please find the attached logs.
==================================
Xcode version - 11.0 
FirebaseAnalytics Version - 6.2.2
Firebase/CoreOnly - 6.16.0
FirebaseCore - 6.6.1
Firebase/Database - 6.16.0

====================================

<img width="924" alt="Screenshot 2020-02-14 at 6 11 35 PM" src="https://user-images.githubusercontent.com/22444085/74533043-a6a78680-4f56-11ea-88d4-feef0dfc70ae.png">

maksymmalyhin:
@Aishwaryalaxmi I am sorry to see you have the issue. 

From the first glance it looks like the reason of the issue is in the way Firebase Performance handles FirebaseApp deletion. We will investigate further to provide a solution for the issue.

A quick question. You didn't mention Firebase Performance in the list of used products. Did you just forget mentioning it or Firebase Performance was not supposed to be included in the project?

Aishwaryalaxmi:
@maksymmalyhin  We are using Firebase Performance (6.16.0 version).

visumickey:
@maksymmalyhin Looking at the stack trace, I don't think Firebase Performance is the cause of the crash. Firebase Performance is requesting the installations ID before dispatching performance events and that request seem to have caused the crash. Reassigning this to Installations ID team to have a look at the crash.

andirayo:
+Maksym Malyhin <mmaksym@google.com>
+FIS Engineering Team <fis-eng-team@google.com>

paulb777:
See discussion in #4909

maksymmalyhin:
[...]

PR #4909, opened 2020-02-14
Nullable +[FIRInstallations installations]

maksymmalyhin:
A quick fix for issues like #4906. 
- `+[FIRInstallations installations]`, `+[FIRInstallations installationsWithApp:]`:  don't throw an exception if a corresponding app is not configured or is `nil`.

Before IID started using Firebase Installations the method `+[FIRInstanceID instanceID]` didn't throw an exception and returned `nil` in the case when `FIRApp` is not configured yet. Now `+[FIRInstanceID instanceID]` calls `+[FIRInstallations installations]` that throws an exception in this case. 

The exception revealed a race condition in `-[FIRApp deleteApp:]` method. Some SDKs like Analytics Performance, IID, etc. may have a background operation started before the app deletion and attempting to call `+[FIRInstallations installations]` after the deletion. The race condition will need to be addressed, but it may require significant changes so the current PR should provide a quick fix for the issue without a significant downside.  

cl/295207830 for global TAP presubmit

### DIFF
--- a/FirebaseInstallations/Source/Library/FIRInstallations.m
+++ b/FirebaseInstallations/Source/Library/FIRInstallations.m
@@ -153,21 +153,21 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Public
 
-+ (FIRInstallations *)installations {
-  FIRApp *defaultApp = [FIRApp defaultApp];
++ (nullable FIRInstallations *)installations {
+  FIRApp *defaultApp = [FIRApp defaultApp]; // Missing configure will be logged here.
   if (!defaultApp) {
-    [NSException raise:kFirebaseInstallationsErrorDomain
-                format:@"The default FirebaseApp instance must be configured before the default"
-                       @"FirebaseApp instance can be initialized. One way to ensure that is to "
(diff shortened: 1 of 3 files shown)


FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -153,21 +153,20 @@ + (void)validateAppOptions:(FIROptions *)appOptions appName:(NSString *)appName
...
++ (nullable FIRInstallations *)installations {
+  FIRApp *defaultApp = [FIRApp defaultApp];  // Missing configure will be logged here.
   if (!defaultApp) {
-    [NSException raise:kFirebaseInstallationsErrorDomain
-                format:@"The default FirebaseApp instance must be configured before the default"
-                       @"FirebaseApp instance can be initialized. One way to ensure that is to "
-                       @"call `[FIRApp configure];` (`FirebaseApp.configure()` in Swift) in the App"
-                       @" Delegate's `application:didFinishLaunchingWithOptions:` "
-                       @"(`application(_:didFinishLaunchingWithOptions:)` in Swift)."];
+    return nil;
```
wilhuff: The old exception text was quite helpful. Consider logging a warning with that text in it.

PR #4910, opened 2020-02-14
Add the ability to send to the new and legacy crash reporting endpoint

jasonhu-g:
The report manager has the ability to send to two different endpoints.

The old legacy flow for a report goes through the following states/folders:
1. active - .clsrecords optimized for crash time persistence
2. processing - .clsrecords with attempted symbolication
3. prepared-legacy - .multipartmime of compressed .clsrecords

The legacy flow for a report goes through the following states/folders:
1. active - .clsrecords optimized for crash time persistence
2. processing - .clsrecords with attempted symbolication
3. prepared - .clsrecords moved from processing with no changes

The code was designed so the report processing workflows are not dramatically different from one another. The design will help avoid having a lot of conditional code blocks throughout the codebase.

### DIFF
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
@@ -12,6 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//
+// The report manager has the ability to send to two different endpoints.
+//
+// The old legacy flow for a report goes through the following states/folders:
+// 1. active - .clsrecords optimized for crash time persistance
+// 2. processing - .clsrecords with attempted symbolication
+// 3. prepared-legacy - .multipartmime of compressed .clsrecords
+//
+// The new legacy flow for a report goes through the following states/folders:
+// 1. active - .clsrecords optimized for crash time persistance
+// 2. processing - .clsrecords with attempted symbolication
+// 3. prepared - .clsrecords moved from processing with no changes
+//
+// The code was designed so the report processing workflows are not dramatically different from one another.
+// The design will help avoid having a lot of conditional code blocks throughout the codebase.
+//
+
 #include <stdatomic.h>
 
 #if __has_include(<FBLPromises/FBLPromises.h>)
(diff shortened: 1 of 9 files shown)


Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -12,6 +12,23 @@
...
 
+//
+// The report manager has the ability to send to two different endpoints.
+//
+// The old legacy flow for a report goes through the following states/folders:
+// 1. active - .clsrecords optimized for crash time persistance
+// 2. processing - .clsrecords with attempted symbolication
+// 3. prepared-legacy - .multipartmime of compressed .clsrecords
+//
+// The new legacy flow for a report goes through the following states/folders:
```
morganchen12: drive-by comment: remove this extra `legacy`

samedson: giphy

Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
```diff
@@ -243,9 +198,67 @@ - (BOOL)uploadPackagedReportAtPath:(NSString *)path
...
   FIRCLSDeveloperLog("Crashlytics:Crash:Reports", @"Submitting report%@",
                      urgent ? @" as urgent" : @"");
-  return [self submitPackageMultipartMimeAtPath:path
-                            dataCollectionToken:dataCollectionToken
-                                  synchronously:urgent];
+
+  // Check with the legacy path as the new path will always be contained in the legacy path
+  BOOL isNewPreparedPath = ![path containsString:self.fileManager.legacyPreparedPath];
+
+  if (isNewPreparedPath && self.dataSource.settings.shouldUseNewReportEndpoint) {
```
samedson: Just a thought - should we just upload any reports that have been prepared, and just use the feature switch to determine how to prepare them?

jasonhu-g: The app will have to fail to trigger an upload and the experiment flag has to be toggled to have reports in 2 prepared folder. Since this is an edge case, I am pivoting towards reliability. If there's issue with preparing or sending with one of the endpoints, then attempting to send to both endpoints may cause an unexpected issue (even crash). I prefer to have the SDK send only to the endpoint it is configured to.

PR #4912, opened 2020-02-15
Force synchronous AsyncQueue shutdown

wilhuff:
Absolutely prevent tasks on the `AsyncQueue` from running after it has shut down.

Also rework the `AsyncQueue::Shutdown` interface to more closely match the actual behavior. Now instead of shutting down, the `AsyncQueue` enters a restricted mode--this more clearly conveys that work can continue, but now it's only restricted tasks.

This also adds an `AsyncQueue::Stop` interface after which no new work will be accepted.

With these changes it's possible to guarantee that tests will actually be done executing by the time they're torn down so this includes changes to the integration tests such that they'll reuse the database across tests, essentially validating that each test really is fully releasing its state.

Also fix the use of `HARD_ASSERT` in destructors by adding a `HARD_ASSERT_NOTHROW` variant. When exceptions are enabled, an assertion failure results in an immediate call to `std::terminate` because destructors are implicitly `noexcept`.

Finally, this also fixes a long-standing bug encountered on Travis where we'd see failures indicating that the test operation never finished bootstrapping. The issue was that we ran GoogleTest tests very early, essentially before XCTest had a chance to start. If there was a crash in the GoogleTests the test runner wouldn't count the test as properly having started. `FSTGoogleTestTests` now defers running tests until later.

### DIFF
--- a/Firestore/Example/Tests/Util/FSTIntegrationTestCase.mm
+++ b/Firestore/Example/Tests/Util/FSTIntegrationTestCase.mm
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google
+ * Copyright 2017 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,8 +63,8 @@ using firebase::firestore::auth::User;
 using firebase::firestore::core::DatabaseInfo;
 using firebase::firestore::local::LevelDbOpener;
 using firebase::firestore::model::DatabaseId;
-using firebase::firestore::testutil::AppForUnitTesting;
 using firebase::firestore::testutil::AsyncQueueForTesting;
+using firebase::firestore::testutil::OptionsForUnitTesting;
 using firebase::firestore::remote::GrpcConnection;
 using firebase::firestore::util::AsyncQueue;
 using firebase::firestore::util::CreateAutoId;
@@ -108,6 +108,8 @@ class FakeCredentialsProvider : public EmptyCredentialsProvider {
 };
 
 @implementation FSTIntegrationTestCase {
+  // A map of project ID to FIRApp.
+  NSMutableDictionary<NSString *, FIRApp *> *_apps;
   NSMutableArray<FIRFirestore *> *_firestores;
   std::shared_ptr<FakeCredentialsProvider> _fakeCredentialsProvider;
 }
@@ -122,7 +124,8 @@ class FakeCredentialsProvider : public EmptyCredentialsProvider {
   [self clearPersistenceOnce];
   [self primeBackend];
 
-  _firestores = [NSMutableArray array];
+  _apps = [[NSMutableDictionary alloc] init];
+  _firestores = [[NSMutableArray alloc] init];
   self.db = [self firestore];
   self.eventAccumulator = [FSTEventAccumulator accumulatorForTest:self];
 }
@@ -132,6 +135,9 @@ class FakeCredentialsProvider : public EmptyCredentialsProvider {
     for (FIRFirestore *firestore in _firestores) {
       [self terminateFirestore:firestore];
     }
+    [_apps enumerateKeysAndObjectsUsingBlock:^(NSString *projectID, FIRApp *app, BOOL *stop) {
+      [self deleteApp:app];
+    }];
   } @finally {
     _firestores = nil;
     [super tearDown];
@@ -261,8 +267,23 @@ class FakeCredentialsProvider : public EmptyCredentialsProvider {
   return defaultSettings;
 }
 
+- (FIRApp *)appWithProjectID:(NSString *)projectID {
+  FIRApp *app = _apps[projectID];
+  if (!app) {
+    // Use the same appName across integration tests to ensure that terminating
+    // Firestore is is actually completing before the next test starts.
+    NSString *appName = [NSString stringWithFormat:@"firestore-integration-tests-%@", projectID];
+    FIROptions *options = OptionsForUnitTesting(util::MakeString(projectID));
+
+    [FIRApp configureWithName:appName options:options];
+    app = [FIRApp appNamed:appName];
+    _apps[projectID] = app;
+  }
+  return app;
+}
+
 - (FIRFirestore *)firestoreWithProjectID:(NSString *)projectID {
-  FIRApp *app = AppForUnitTesting(util::MakeString(projectID));
+  FIRApp *app = [self appWithProjectID:projectID];
   return [self firestoreWithApp:app];
 }
 
--- a/Firestore/Example/Tests/Util/FSTIntegrationTestCase.mm
+++ b/Firestore/Example/Tests/Util/FSTIntegrationTestCase.mm
@@ -270,8 +270,8 @@ class FakeCredentialsProvider : public EmptyCredentialsProvider {
 - (FIRApp *)appWithProjectID:(NSString *)projectID {
   FIRApp *app = _apps[projectID];
   if (!app) {
-    // Use the same appName across integration tests to ensure that terminating
-    // Firestore is is actually completing before the next test starts.
+    // Use the same appName across integration tests to ensure that Firestore termination is
+    // actually complete before the next test starts.
     NSString *appName = [NSString stringWithFormat:@"firestore-integration-tests-%@", projectID];
     FIROptions *options = OptionsForUnitTesting(util::MakeString(projectID));
 
--- a/Firestore/core/src/firebase/firestore/api/firestore.cc
+++ b/Firestore/core/src/firebase/firestore/api/firestore.cc
@@ -63,13 +63,7 @@ Firestore::Firestore(model::DatabaseId database_id,
 }
 
 Firestore::~Firestore() {
-  std::lock_guard<std::mutex> lock{mutex_};
-
-  // If the client hasn't been configured yet we don't need to create it just
-  // to tear it down.
-  if (!client_) return;
-
-  client_->Dispose();
+  Dispose();
 }
 
 const std::shared_ptr<FirestoreClient>& Firestore::client() {
@@ -144,6 +138,17 @@ void Firestore::Terminate(util::StatusCallback callback) {
   // throws an exception.
   EnsureClientConfigured();
   client_->TerminateAsync(std::move(callback));
+}
+
+void Firestore::Dispose() {
+  std::lock_guard<std::mutex> lock{mutex_};
+
+  // If the client hasn't been configured yet we don't need to create it just
+  // to dispose of it.
+  if (!client_) return;
+
+  client_->Dispose();
+  client_.reset();
 }
 
 void Firestore::WaitForPendingWrites(util::StatusCallback callback) {
--- a/Firestore/core/src/firebase/firestore/api/firestore.cc
+++ b/Firestore/core/src/firebase/firestore/api/firestore.cc
@@ -69,7 +69,7 @@ Firestore::~Firestore() {
   // to tear it down.
   if (!client_) return;
 
-  client_->Stop();
+  client_->Dispose();
 }
 
 const std::shared_ptr<FirestoreClient>& Firestore::client() {
--- a/Firestore/core/src/firebase/firestore/api/firestore.cc
+++ b/Firestore/core/src/firebase/firestore/api/firestore.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -69,7 +69,7 @@ Firestore::~Firestore() {
   // to tear it down.
   if (!client_) return;
 
-  client_->Terminate();
+  client_->Stop();
 }
 
 const std::shared_ptr<FirestoreClient>& Firestore::client() {
@@ -152,7 +152,7 @@ void Firestore::WaitForPendingWrites(util::StatusCallback callback) {
 }
 
 void Firestore::ClearPersistence(util::StatusCallback callback) {
-  worker_queue()->EnqueueEvenAfterShutdown([this, callback] {
+  worker_queue()->EnqueueEvenWhileRestricted([this, callback] {
     auto Yield = [=](Status status) {
       if (callback) {
         this->user_executor_->Execute([=] { callback(status); });
--- a/Firestore/core/src/firebase/firestore/core/event_listener.h
+++ b/Firestore/core/src/firebase/firestore/core/event_listener.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,8 +17,8 @@
 #ifndef FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_CORE_EVENT_LISTENER_H_
 #define FIRESTORE_CORE_SRC_FIREBASE_FIRESTORE_CORE_EVENT_LISTENER_H_
 
-#include <atomic>
 #include <memory>
+#include <mutex>  // NOLINT(build/c++11)
 #include <utility>
 
 #include "Firestore/core/src/firebase/firestore/util/executor.h"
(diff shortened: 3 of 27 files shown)


Firestore/core/src/firebase/firestore/util/async_queue.h
```diff
@@ -109,22 +131,28 @@ class AsyncQueue : public std::enable_shared_from_this<AsyncQueue> {
   // calling `Enqueue` is a no-op.
   void Enqueue(const Operation& operation);
 
-  // Like `Enqueue`, but also starts the shutdown process. Once the shutdown
-  // process has started, calling any Enqueue* methods becomes a no-op
+  // Puts the `AsyncQueue` into restricted mode, where calling any Enqueue*
```
wu-hui: `into the restricted mode`?

Firestore/core/src/firebase/firestore/util/async_queue.h
```diff
@@ -109,22 +131,28 @@ class AsyncQueue : public std::enable_shared_from_this<AsyncQueue> {
...
+  // Puts the `AsyncQueue` into the stopped mode, where calling any Enqueue*
+  // methods becomes a no-op without exception.
   //
-  // The exception is `EnqueueEvenAfterShutdown`, operations requsted via
-  // this will still be scheduled.
-  void EnqueueAndInitiateShutdown(const Operation& operation);
+  // This also synchronously waits for the last pending operation to complete.
+  void Stop();
 
   // Like `Enqueue`, but it will proceed scheduling the requested operation
```
wu-hui: The comment needs to be updated to reflect the change.

Firestore/core/src/firebase/firestore/core/firestore_client.cc
```diff
@@ -317,7 +314,7 @@ bool FirestoreClient::is_terminated() const {
   // Technically, the worker queue is still running, but only accepting tasks
```
wu-hui: Comment needs to be updated.

Firestore/core/src/firebase/firestore/core/firestore_client.h
```diff
@@ -86,15 +86,16 @@ class FirestoreClient : public std::enable_shared_from_this<FirestoreClient> {
...
+   * resources. The client enters a restricted mode where only a few methods
+   * like ClearPersistence are allowed.
    */
   void TerminateAsync(util::StatusCallback callback);
 
   /**
-   * Synchronously terminates this client, cancels all writes / listeners, and
-   * releases all resources.
+   * Synchronously terminates this client, cancels all writes / listeners,
+   * releases all resources, and prepares for destruction.
```
wu-hui: I think we can also call out `Stop` returns when all operations are run, like what you did in the `AsyncQueue.Stop` comment.

Firestore/core/test/firebase/firestore/util/async_queue_test.cc
```diff
@@ -38,6 +38,12 @@ const TimerId kTimerId3 = TimerId::WriteStreamConnectionBackoff;
 
 }  // namespace
 
+AsyncQueueTest::~AsyncQueueTest() {
+  if (queue) {
+    queue->EnqueueBlocking([] {});
```
var-const: Optional: each time I see this pattern being used (and I think we use it quite a lot), I wonder if we should add a method like `AsyncQueue::Drain` or something.

Firestore/core/test/firebase/firestore/util/async_queue_test.cc
```diff
@@ -197,15 +203,20 @@ TEST_P(AsyncQueueTest, CanManuallyDrainSpecificDelayedOperationsForTesting) {
...
+  queue->EnqueueEvenWhileRestricted([&] { steps += '2'; });
   queue->Enqueue([&] { steps += '3'; });
-  queue->EnqueueEvenAfterShutdown([&] { steps += '4'; });
-  queue->EnqueueEvenAfterShutdown(ran.AsCallback());
+  queue->EnqueueEvenWhileRestricted([&] { steps += '4'; });
+  queue->EnqueueEvenWhileRestricted(ran.AsCallback());
+
+  queue->Stop();
+  queue->Enqueue([&] { steps += '5'; });
+  queue->EnqueueEvenWhileRestricted([&] { steps += '6'; });
```
var-const: No action required: technically, I think this test could pass even if the queue actually executed both of the operations enqueued after shutdown (because there is no synchronization to wait for them). On the other hand, I don't think it's possible to add that synchronization, so I guess there's not much we can do.

Firestore/core/test/firebase/firestore/remote/exponential_backoff_test.cc
```diff
@@ -42,6 +42,12 @@ class ExponentialBackoffTest : public testing::Test,
         backoff{queue, timer_id, 1.5, chr::seconds{5}, chr::seconds{30}} {
   }
 
+  void TearDown() override {
+    queue->Stop();
+
+    testing::Test::TearDown();
```
var-const: Question: is this necessary?

Firestore/core/test/firebase/firestore/FSTGoogleTestTests.mm
```diff
@@ -186,6 +188,8 @@ @interface GoogleTests : XCTestCase
  * this way here allows XCTAssert and friends to work.
  */
 void ReportTestResult(XCTestCase* self, SEL _cmd) {
```
var-const: Question: is this function called after the test was supposed to finish running? Or is using it to kick off the actual tests idiomatic?

Firestore/Example/Tests/Util/FSTIntegrationTestCase.mm
```diff
@@ -261,8 +267,23 @@ + (FIRFirestoreSettings *)settings {
   return defaultSettings;
 }
 
+- (FIRApp *)appWithProjectID:(NSString *)projectID {
+  FIRApp *app = _apps[projectID];
+  if (!app) {
+    // Use the same appName across integration tests to ensure that terminating
+    // Firestore is is actually completing before the next test starts.
```
var-const: Nit: `is` is repeated twice.

Firestore/core/src/firebase/firestore/core/firestore_client.h
```diff
@@ -86,15 +86,16 @@ class FirestoreClient : public std::enable_shared_from_this<FirestoreClient> {
 
   /**
    * Terminates this client, cancels all writes / listeners, and releases all
-   * resources.
+   * resources. The client enters a restricted mode where only a few methods
+   * like ClearPersistence are allowed.
    */
   void TerminateAsync(util::StatusCallback callback);
```
var-const: FWIW, to me, `Terminate` seems like a stronger term than `Stop`, so I find it a little surprising that `Stop` actually does more than `Terminate`. What do you think?

Firestore/core/src/firebase/firestore/util/exception.h
```diff
@@ -69,6 +87,17 @@ using ThrowHandler = void (*)(ExceptionType type,
                               int line,
                               const std::string& message);
 
+/**
+ * The default throw handler implementation, suitable for C++. If exceptions are
+ * enabled, throws a C++ exception, except for AssertionFailureNoThrow, which
+ * unconditionally aborts.
```
var-const: Ultranit: s/aborts/terminates/? Not that it's a very important distinction in our case, but the current wording suggests we call `std::abort`.

Firestore/core/src/firebase/firestore/util/async_queue.h
```diff
@@ -109,22 +131,28 @@ class AsyncQueue : public std::enable_shared_from_this<AsyncQueue> {
   // calling `Enqueue` is a no-op.
   void Enqueue(const Operation& operation);
 
-  // Like `Enqueue`, but also starts the shutdown process. Once the shutdown
-  // process has started, calling any Enqueue* methods becomes a no-op
+  // Puts the `AsyncQueue` into restricted mode, where calling any Enqueue*
+  // methods becomes a no-op.
+  //
+  // The exception is `EnqueueEvenWhileRestricted`, where operations are still
```
var-const: Optional: perhaps rephrase to something like `which would still enqueue operations even...`? (basically, active vs. passive voice).

Firestore/core/src/firebase/firestore/util/async_queue.h
```diff
@@ -89,8 +89,30 @@ class AsyncQueue : public std::enable_shared_from_this<AsyncQueue> {
   using Operation = Executor::Operation;
   using Milliseconds = Executor::Milliseconds;
 
+  enum class Mode {
+    /**
+     * The default mode of an AsyncQueue after creation. All tasks are allowed.
```
var-const: Ultranit: can you add backticks around `AsyncQueue` and `Mode::Restricted`? (here and below)

Firestore/core/src/firebase/firestore/util/async_queue.cc
```diff
@@ -127,9 +146,16 @@ AsyncQueue::Operation AsyncQueue::Wrap(const Operation& operation) {
...
 
-  // Note: can't move `operation` into lambda until C++14.
-  auto shared_this = shared_from_this();
-  return [shared_this, operation] { shared_this->ExecuteBlocking(operation); };
+  // Capture by weak_ptr in order to avoid implicitly keeping the AsyncQueue
+  // alive even though the owning FirestoreClient may have been destroyed
+  // already.
+  std::weak_ptr<AsyncQueue> weak_this = shared_from_this();
+  return [weak_this, operation] {
+    std::shared_ptr<AsyncQueue> shared_this = weak_this.lock();
```
var-const: Optional: use `auto`? (the type seems obvious enough, given that the `weak_ptr` declaration is just two lines above) No strong feelings, though.

Firestore/core/src/firebase/firestore/core/firestore_client.h
```diff
@@ -93,11 +95,15 @@ class FirestoreClient : public std::enable_shared_from_this<FirestoreClient> {
...
    * Synchronously terminates this client, cancels all writes / listeners,
-   * releases all resources, and prepares for destruction.
+   * releases all resources, synchronously waits for the last pending operation
+   * to complete, and all but destroys this instance.
    *
-   * This also synchronously waits for the last pending operation to complete.
+   * Note: Dispose exists separately from the destructor because shared_ptrs to
+   * FirestoreClient are captured frequently in ways that are hard to audit.
+   * This in turn makes it difficult to rely on the last shared owner to
+   * destroy the instance in a timely manner. All asynchronous operations
```
var-const: Comment trails off (_"All asynchronous operations..."_).

Firestore/core/src/firebase/firestore/util/async_queue.h
```diff
@@ -107,7 +107,7 @@ class AsyncQueue : public std::enable_shared_from_this<AsyncQueue> {
      * Finally, once the Firestore instance is in the process of being destroyed
```
var-const: Nit: I think there should be a comma here.

Firestore/core/src/firebase/firestore/util/executor_libdispatch.mm
```diff
@@ -60,16 +60,19 @@ void DispatchAsync(const dispatch_queue_t queue, std::function<void()>&& work) {
 }
 
 void DispatchSync(const dispatch_queue_t queue, std::function<void()> work) {
-  HARD_ASSERT(
-      GetCurrentQueueLabel() != GetQueueLabel(queue),
-      "Calling DispatchSync on the current queue will lead to a deadlock.");
+  if (GetCurrentQueueLabel() == GetQueueLabel(queue)) {
```
var-const: Note: this function is very similar to `RunSynchronized` below; it looks like `RunSynchronized` can now be removed (or perhaps this change can be avoided by using `RunSynchronized` -- though having a non-deadlocking `DispatchSync` is probably preferable).

Firestore/core/src/firebase/firestore/api/firestore.cc
```diff
@@ -146,6 +140,17 @@ void Firestore::Terminate(util::StatusCallback callback) {
...
 
+void Firestore::Dispose() {
+  std::lock_guard<std::mutex> lock{mutex_};
+
+  // If the client hasn't been configured yet we don't need to create it just
+  // to dispose of it.
+  if (!client_) return;
+
+  client_->Dispose();
+  client_.reset();
```
var-const: Question: what do you think about setting `extension_` to null here? It would have likely made the crash I've been seeing occur earlier and with a slightly more relevant stack trace. On the other hand, perhaps this is too specific to a particular problem.

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -132,6 +153,9 @@ void AsyncEventListener<T>::OnEvent(util::StatusOr<T> maybe_value) {
   std::shared_ptr<AsyncEventListener<T>> shared_this = this->shared_from_this();
 
   executor_->Execute([shared_this, maybe_value]() {
+    // Hold the lock while calling the delegate in order to prevent concurrent
```
var-const: Perhaps add a note like (`to prevent concurrent destruction of the Firestore instance (which will block on trying to mute the listener)`)? Without context, it's not immediately clear what the lock has to do with destroying a Firestore instance.

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -87,9 +84,32 @@ class AsyncEventListener
   void Mute();
 
  private:
-  std::atomic<bool> muted_;
   std::shared_ptr<util::Executor> executor_;
   DelegateListener delegate_;
+
+  // A mutex that protects both muting the AsyncEventListener and also calling
```
var-const: Nit: backtick `AsyncEventListener`.

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -87,9 +84,32 @@ class AsyncEventListener
...
+  // out to the delegate.
+  //
+  // Mute calls must be synchronized because users expect that when they call
+  // `ListenerRegistration::Remove` that they don't get notifications pretty
+  // much immediately upon return of that method. That is, we can't afford to
+  // wait for the `Remove` to be submitted through the `AsyncQueue`.
+  //
+  // The call to delegate_->OnEvent must also be protected in order to ensure
+  // that The `Firestore` instance isn't destroyed while we're calling out to
```
var-const: Nit: s/The/the

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -87,9 +84,32 @@ class AsyncEventListener
...
+
+  // A mutex that protects both muting the AsyncEventListener and also calling
+  // out to the delegate.
+  //
+  // Mute calls must be synchronized because users expect that when they call
+  // `ListenerRegistration::Remove` that they don't get notifications pretty
+  // much immediately upon return of that method. That is, we can't afford to
+  // wait for the `Remove` to be submitted through the `AsyncQueue`.
+  //
+  // The call to delegate_->OnEvent must also be protected in order to ensure
```
var-const: Nit: backticks.

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -87,9 +84,32 @@ class AsyncEventListener
...
+  // that The `Firestore` instance isn't destroyed while we're calling out to
+  // user code. `Firestore::Dispose` (eventually) calls `Mute` on each listener
+  // and forcing `Mute` and `OnEvent` to be mutually exclusive avoids a race.
+  //
+  // This must be a recursive mutex because the `DelegateListener` may be user
+  // code, and that we must allow that user code to invoke
```
var-const: Nit: the first `that` seems extraneous (`and we must allow that user code...` flows just as well).

Firestore/core/src/firebase/firestore/core/event_listener.h
```diff
@@ -87,9 +84,32 @@ class AsyncEventListener
...
+  // `ListenerRegistration::Remove` (which calls `Mute` on this class). If this
+  // were a non-recursive mutex such a call would deadlock.
```
var-const: Nit: I think there should be a comma after `mutex`.

wilhuff:
Abandoning this in favor of #5547 and #5620.

PR #4924, opened 2020-02-18
M65 changelog for FIAM SDK

rohandandavati:

### Discussion

### Testing

### API Changes

    can discuss it together.


FirebaseInAppMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2020-02-18 -- v0.17.0
+- [added] Added support for data bundles for in-app messages - additional key-value pairs that can be sent along with an in-app message (#4922).
```
paulb777: The dash is a bit confusing since it implies in-app messages are additional ...

How about - Added support for data bundles for in-app messages. Data bundles are additional key-value pairs that can be sent along with an in-app message (#4922).

rohandandavati: Good point. Thanks for the suggestion, just changed it.

PR #4932, opened 2020-02-19
Revert "Add a completion handler to FIRExperimentController's updateE…

ryanwilson:
…xperimentsWithServiceOrigin: method (#4897)"

This reverts commit 7587c3c2e4af3f1cd7e10a33eb075b81a18941ae.

This change needs to go through an API review first.

cc @christibbs


ryanwilson:
Unnecessary at the moment - will not be included in M65 so this doesn't need to be reverted.

PR #4935, opened 2020-02-19
Forward declare class to fix linting.

ryanwilson:
This will need to be excluded in the merge, but is required to get GDTCCTSupport linting.

```
- ERROR | xcodebuild:  GoogleDataTransportCCTSupport/GoogleDataTransportCCTSupport/GDTCCTLibrary/Private/GDTCCTPrioritizer.h:37:35: error: unknown type name 'GDTCORStoredEvent'

```


ryanwilson:
Would also be worth finding out why this didn't get caught in our tests.

PR #4938, opened 2020-02-19
Specify platforms for Swift unit tests

ryanwilson:
This is required since `watchOS` doesn't have `XCTest` available.

#no-changelog


ryanwilson:
Killed Travis build since all Core tests passed.

PR #4939, opened 2020-02-19
FIRApp: add apple platform to firebaseUserAgent

maksymmalyhin:
Collect stats on Firebase usage by Apple platform (b/146507104).

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -613,25 +613,17 @@ static dispatch_once_t sFirebaseUserAgentOnceToken;
 + (NSString *)applePlatform {
   NSString *applePlatform = @"unknown";
 
-#if TARGET_OS_IOS
-  applePlatform = @"ios";
-#endif  // TARGET_OS_IOS
-
-#if TARGET_OS_TV
-  applePlatform = @"tvos";
-#endif  // TARGET_OS_TV
-
-#if TARGET_OS_OSX
-  applePlatform = @"macos";
-#endif  // TARGET_OS_OSX
-
-#if TARGET_OS_WATCH
-  applePlatform = @"watchos";
-#endif  // TARGET_OS_WATCH
-
 #if TARGET_OS_MACCATALYST
   applePlatform = @"maccatalyst";
-#endif  // TARGET_OS_MACCATALYST
+#elif TARGET_OS_IOS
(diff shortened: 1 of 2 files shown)


FirebaseCore/Sources/FIRApp.m
```diff
@@ -606,6 +610,32 @@ + (BOOL)hasSwiftRuntime {
...
+#if TARGET_OS_OSX
+  applePlatform = @"macos";
+#endif  // TARGET_OS_OSX
+
+#if TARGET_OS_WATCH
+  applePlatform = @"watchos";
+#endif  // TARGET_OS_WATCH
+
+#if TARGET_OS_MACCATALYST
+  applePlatform = @"maccatalyst";
```
paulb777: Please check that this define is not combined with any of the others.  We might want it first with `#elif`s

maksymmalyhin: According to the current API docs it should not. I wonder if we can test it on Travis/GHA somehow? 
Screen Shot 2020-02-19 at 3 30 48 PM

maksymmalyhin: Ok, for Catalyst app run on macOS both `TARGET_OS_IOS` and `TARGET_OS_MACCATALYST` are set. I wonder what platform we would like to log in this case? I assume `maccatalyst`. WDYT?

ryanwilson: I would say `maccatalyst`.

maksymmalyhin: Updated with `#elif` accordingly.

FirebaseCore/Sources/FIRApp.m
```diff
@@ -606,6 +610,24 @@ + (BOOL)hasSwiftRuntime {
   return hasSwiftRuntime;
 }
 
++ (NSString *)applePlatform {
+  NSString *applePlatform = @"unknown";
+
+#if TARGET_OS_MACCATALYST
```
paulb777: Add a comment summarizing the discussion so we remember this is order-sensitive

maksymmalyhin: Added the comment. Is it what you meant?

paulb777: Yep. Thanks!

PR #4940, opened 2020-02-20
Zip Builder support for dependencies through subspecs

paulb777:
With these changes, it's now possible to build FirebaseUI with the ZipBuilder

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -57,13 +57,17 @@ enum CocoaPodUtils {
     /// Binary frameworks in this pod.
     let binaryFrameworks: [URL]
 
+    /// Subspecs installed for this pod.
+    let subspecs: Set<String>
+
     /// The contents of the module map for all frameworks associated with the pod.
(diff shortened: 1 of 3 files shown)


ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -205,11 +208,39 @@ enum CocoaPodUtils {
...
+        if let previousVersion = versions[podName], version != previousVersion {
+          fatalError("Different installed versions for \(podName)." +
+            "\(version) versus \(previousVersion)")
+        } else {
+          let basePodName = subspecArray[0]
+          versions[basePodName] = version
+          if subspecs[basePodName] == nil {
+            subspecs[basePodName] = Set()
+          }
+          subspecs[basePodName]?.insert(subspecArray[1])
```
ryanwilson: Here and below - you may be able to clean it up with the `subscript(_:default:)` call:

`subspecs[basePodName, default: Set()].insert(subspecArray[1])`

This will use the `default` value if it's currently `nil`.

paulb777: Much cleaner. Thanks! Done.

PR #4944, opened 2020-02-20
Generate network_connection_info and add tests

doudounan:
1 Generate gdt_cct_NetworkConnectionInfo in GDTCCTNanopbHelpers.
2 Modify GDTCCTPrioritizer to trigger network_connection_info generation.
3 Add a test for network_connection_info generation in GDTCCTPrioritizerTest.
4 Modify iOS testapp in GDTCCTSupport to generate events with network_connection_info.
4 Fix bug in GDTCOREvent encoding.
5 Add platform check methods in GDTCORPlatform.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -52,31 +52,35 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
 
 GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
 #if TARGET_OS_IOS
-  static NSDictionary<NSString *, NSNumber *> *ctRadioAccessTechnologyToNetworkSubTypeMessage;
+  static NSDictionary<NSString *, NSNumber *> *CTRadioAccessTechnologyToNetworkSubTypeMessage;
   static CTTelephonyNetworkInfo *networkInfo;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-    ctRadioAccessTechnologyToNetworkSubTypeMessage = @{
-      CTRadioAccessTechnologyGPRS : @(GDTCORNetworkMobileSubtype_GPRS),
-      CTRadioAccessTechnologyEdge : @(GDTCORNetworkMobileSubtype_Edge),
-      CTRadioAccessTechnologyWCDMA : @(GDTCORNetworkMobileSubtype_WCDMA),
-      CTRadioAccessTechnologyHSDPA : @(GDTCORNetworkMobileSubtype_HSDPA),
-      CTRadioAccessTechnologyHSUPA : @(GDTCORNetworkMobileSubtype_HSUPA),
-      CTRadioAccessTechnologyCDMA1x : @(GDTCORNetworkMobileSubtype_CDMA1x),
-      CTRadioAccessTechnologyCDMAEVDORev0 : @(GDTCORNetworkMobileSubtype_CDMAEVDORev0),
-      CTRadioAccessTechnologyCDMAEVDORevA : @(GDTCORNetworkMobileSubtype_CDMAEVDORevA),
-      CTRadioAccessTechnologyCDMAEVDORevB : @(GDTCORNetworkMobileSubtype_CDMAEVDORevB),
-      CTRadioAccessTechnologyeHRPD : @(GDTCORNetworkMobileSubtype_HRPD),
-      CTRadioAccessTechnologyLTE : @(GDTCORNetworkMobileSubtype_LTE),
+    CTRadioAccessTechnologyToNetworkSubTypeMessage = @{
+      CTRadioAccessTechnologyGPRS : @(GDTCORNetworkMobileSubtypeGPRS),
+      CTRadioAccessTechnologyEdge : @(GDTCORNetworkMobileSubtypeEdge),
+      CTRadioAccessTechnologyWCDMA : @(GDTCORNetworkMobileSubtypeWCDMA),
+      CTRadioAccessTechnologyHSDPA : @(GDTCORNetworkMobileSubtypeHSDPA),
+      CTRadioAccessTechnologyHSUPA : @(GDTCORNetworkMobileSubtypeHSUPA),
+      CTRadioAccessTechnologyCDMA1x : @(GDTCORNetworkMobileSubtypeCDMA1x),
+      CTRadioAccessTechnologyCDMAEVDORev0 : @(GDTCORNetworkMobileSubtypeCDMAEVDORev0),
+      CTRadioAccessTechnologyCDMAEVDORevA : @(GDTCORNetworkMobileSubtypeCDMAEVDORevA),
+      CTRadioAccessTechnologyCDMAEVDORevB : @(GDTCORNetworkMobileSubtypeCDMAEVDORevB),
+      CTRadioAccessTechnologyeHRPD : @(GDTCORNetworkMobileSubtypeHRPD),
+      CTRadioAccessTechnologyLTE : @(GDTCORNetworkMobileSubtypeLTE),
     };
     networkInfo = [[CTTelephonyNetworkInfo alloc] init];
   });
-  NSNumber *networkMobileSubtype =
-      ctRadioAccessTechnologyToNetworkSubTypeMessage[networkInfo.currentRadioAccessTechnology];
-
-  return networkMobileSubtype.intValue;
+  NSString *networkCurrentRadioAccessTechonlogy = networkInfo.currentRadioAccessTechnology;
+  if (networkCurrentRadioAccessTechonlogy) {
+    NSNumber *networkMobileSubtype =
+        CTRadioAccessTechnologyToNetworkSubTypeMessage[networkCurrentRadioAccessTechonlogy];
+    return networkMobileSubtype.intValue;
+  } else {
+    return GDTCORNetworkMobileSubtypeUNKNOWN;
+  }
 #else
-  return GDTCORNetworkMobileSubtype_UNKNOWN;
+  return GDTCORNetworkMobileSubtypeUNKNOWN;
 #endif
 }
 
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -49,6 +49,38 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
 }
 #endif  // !TARGET_OS_WATCH
 
+BOOL GDTCORPlatformIsIOS() {
+#if TARGET_OS_IOS
+  return YES;
+#else
+  return NO;
+#endif
+}
+
+BOOL GDTCORPlatformIsMacOS() {
+#if TARGET_OS_OSX
+  return YES;
+#else
+  return NO;
+#endif
+}
+
+BOOL GDTCORPlatformIsTvOS() {
+#if TARGET_OS_TV
+  return YES;
+#else
+  return NO;
+#endif
+}
+
+BOOL GDTCORPlatformIsWatchOS() {
+#if TARGET_OS_WATCH
+  return YES;
+#else
+  return NO;
+#endif
+}
+
 @interface GDTCORApplication ()
 /**
  Private flag to match the existing `readonly` public flag. This will be accurate for all platforms,
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -71,10 +71,10 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
     };
     networkInfo = [[CTTelephonyNetworkInfo alloc] init];
   });
-  NSString *networkCurrentRadioAccessTechonlogy = networkInfo.currentRadioAccessTechnology;
-  if (networkCurrentRadioAccessTechonlogy) {
+  NSString *networkCurrentRadioAccessTechnology = networkInfo.currentRadioAccessTechnology;
+  if (networkCurrentRadioAccessTechnology) {
     NSNumber *networkMobileSubtype =
-        CTRadioAccessTechnologyToNetworkSubTypeMessage[networkCurrentRadioAccessTechonlogy];
+        CTRadioAccessTechnologyToNetworkSubTypeMessage[networkCurrentRadioAccessTechnology];
     return networkMobileSubtype.intValue;
   } else {
     return GDTCORNetworkMobileSubtypeUNKNOWN;
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -18,6 +18,7 @@
 
 #import <GoogleDataTransport/GDTCORAssert.h>
 #import <GoogleDataTransport/GDTCORConsoleLogger.h>
+#import <GoogleDataTransport/GDTCORReachability.h>
 
 #import "GDTCORLibrary/Private/GDTCORRegistrar_Private.h"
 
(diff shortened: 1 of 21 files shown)


GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -136,8 +138,13 @@ gdt_cct_LogEvent GDTCCTConstructLogEvent(GDTCOREvent *event) {
   logEvent.has_event_uptime_ms = 1;
   logEvent.timezone_offset_seconds = event.clockSnapshot.timezoneOffsetSeconds;
   logEvent.has_timezone_offset_seconds = 1;
-  // TODO: Read network_connection_info from the custom params dict.
-
+  if (event.customPrioritizationParams[@"network_connection_info"]) {
+    NSData *networkConnectionInfoData =
+        event.customPrioritizationParams[@"network_connection_info"];
+    [networkConnectionInfoData getBytes:&logEvent.network_connection_info
+                                 length:sizeof(logEvent.network_connection_info)];
```
mikehaney24: I think length should be networkConnectionInfoData.length. The reason why this works is because the network_connection_info struct has no dynamic member variables and has a static length of 13 bytes.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -188,6 +195,76 @@ gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo() {
...
+      CTRadioAccessTechnologyeHRPD : @(gdt_cct_NetworkConnectionInfo_MobileSubtype_EHRPD),
+      CTRadioAccessTechnologyLTE : @(gdt_cct_NetworkConnectionInfo_MobileSubtype_LTE),
+    };
+  });
+  NSString *networkInfoString = GDTCCTNetworkInfo().currentRadioAccessTechnology;
+  NSNumber *networkMobileSubtype = ctRadioAccessTechnologyToNetworkSubType[networkInfoString];
+  return networkMobileSubtype.intValue;
+}
+
+CTTelephonyNetworkInfo *GDTCCTNetworkInfo() {
```
mikehaney24: for macOS, tvOS, and watchOS you'll need to write a wrapper class (probably in GDTCORPlatform) around CTTelephonyNetworkInfo to return from this function.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -188,6 +195,76 @@ gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo() {
...
+        if (networkConnectionInfo.mobile_subtype !=
+            gdt_cct_NetworkConnectionInfo_MobileSubtype_UNKNOWN_MOBILE_SUBTYPE) {
+          networkConnectionInfo.has_mobile_subtype = 1;
+        }
+      }
+    }
+  } else if (GDTCORPlatformIsMacOS() || GDTCORPlatformIsTvOS() || GDTCORPlatformIsWatchOS()) {
+    SCNetworkReachabilityFlags reachabilityFlags = [GDTCORReachability currentFlags];
+    // Currently set all non-iOS platform network type to WIFI if network is reachable.
+    if (reachabilityFlags & kSCNetworkReachabilityFlagsReachable) {
```
mikehaney24: if ((reachabilityFlags & kSCNetworkReachabilityFlagsReachable) == kSCNetworkReachabilityFlagsReachable)

Here's a good explanation why from http://sam.dods.co/blog/2014/02/01/checking-for-a-value-in-a-bit-mask/

Essentially, it's possible for reachabilityFlags & kSCNetworkReachabilityFlagsReachable to evaluate to YES because they have an overlapping bit, but not be equal to kSCNetworkReachabilityFlagsReachable (which requires all bits to overlap).

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -188,6 +195,76 @@ gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo() {
...
+    }
+  }
+  return networkConnectionInfo;
+}
+
+gdt_cct_NetworkConnectionInfo_NetworkType GDTCCTNetworkConnectonInfoNetworkType() {
+  gdt_cct_NetworkConnectionInfo_NetworkType networkType =
+      gdt_cct_NetworkConnectionInfo_NetworkType_NONE;
+  SCNetworkReachabilityFlags reachabilityFlags = [GDTCORReachability currentFlags];
+  if (reachabilityFlags & kSCNetworkReachabilityFlagsIsWWAN) {
```
mikehaney24: Same as above

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -188,6 +195,76 @@ gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo() {
...
+  }
+  return networkType;
+}
+
+gdt_cct_NetworkConnectionInfo_MobileSubtype GDTCCTNetworkConnectionInfoNetworkMobileSubtype() {
+  static NSDictionary<NSString *, NSNumber *> *ctRadioAccessTechnologyToNetworkSubType;
+  static dispatch_once_t onceToken = 0;
+  dispatch_once(&onceToken, ^{
+    ctRadioAccessTechnologyToNetworkSubType = @{
+      CTRadioAccessTechnologyGPRS : @(gdt_cct_NetworkConnectionInfo_MobileSubtype_GPRS),
```
mikehaney24: You'll need to wrap all the CTRadioAccess enums in a custom enum defined differently based on platform. It probably makes sense to write that part in GDTCORPlatform.

GoogleDataTransportCCTSupport/GDTCCTLibrary/Private/GDTCCTNanopbHelpers.h
```diff
@@ -96,6 +98,35 @@ gdt_cct_ClientInfo GDTCCTConstructClientInfo(void);
 FOUNDATION_EXPORT
 gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo(void);
 
+/** Constructs a gdt_cct_NetworkConnectionInfo representing the client nework connection
+ * information.
+ *
+ * @return The new gdt_cct_NetworkConnectionInfo object.
+ */
+FOUNDATION_EXPORT
+gdt_cct_NetworkConnectionInfo GDTCCTConstructNetWorkConnectionInfo(void);
```
mikehaney24: s/NetWork/Network/

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -188,6 +195,76 @@ gdt_cct_IosClientInfo GDTCCTConstructiOSClientInfo() {
   return iOSClientInfo;
 }
 
+gdt_cct_NetworkConnectionInfo GDTCCTConstructNetWorkConnectionInfo() {
```
mikehaney24: Something to consider: if the network connection info struct ever has a dynamic length member added to it, some of this code will break. Consider returning NSData from this method to help address this problem by making the problem manifest here, as opposed to a consumer of the API.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTPrioritizer.m
```diff
@@ -55,6 +57,11 @@ - (instancetype)init {
 #pragma mark - GDTCORPrioritizer Protocol
 
 - (void)prioritizeEvent:(GDTCOREvent *)event {
+  if (event.customPrioritizationParams[@"needs_network_connection_info"]) {
+    gdt_cct_NetworkConnectionInfo networkInfo = GDTCCTConstructNetWorkConnectionInfo();
+    NSData *networkInfoData = [NSData dataWithBytes:&networkInfo length:sizeof(networkInfo)];
```
mikehaney24: This works only because gdt_cct_NetworkConnectionInfo has a static length. My recommendation to change the return value of GDTCCTConstructNetWorkConnectionInfo to NSData is to help protect usage of the API in particular. This could be a harder bug to track down in the future if the NSData conversion occurs here.

GoogleDataTransportCCTSupport/GDTCCTLibrary/Private/GDTCCTNanopbHelpers.h
```diff
@@ -96,6 +98,35 @@ gdt_cct_ClientInfo GDTCCTConstructClientInfo(void);
...
+ */
+FOUNDATION_EXPORT
+gdt_cct_NetworkConnectionInfo_MobileSubtype GDTCCTNetworkConnectionInfoNetworkMobileSubtype(void);
+
+/** Constructs a CTTelephonyNetworkInfo object representing the client mobile nework information.
+ *
+ * @return The CTTelephonyNetworkInfo object.
+ */
+FOUNDATION_EXPORT
+CTTelephonyNetworkInfo *_Nullable GDTCCTNetworkInfo(void);
```
mikehaney24: Change to some wrapper type

GoogleDataTransportCCTSupport/GDTCCTTestApp/viewcontroller.swift
```diff
@@ -39,6 +39,7 @@ public extension ViewController {
     ]
     testMessage.root.subMessage.repeatedSubMessage[0].samplingPercentage = 13.37
     event.dataObject = testMessage
+    event.customPrioritizationParams = ["needs_network_connection_info": true]
```
mikehaney24: Here, it is fine for "needs_network_connection_info" to just be a normal string since GDTCCT's API is entirely private.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -49,6 +50,36 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
...
+      CTRadioAccessTechnologyCDMAEVDORev0 : @(GDTCORNetworkMobileSubtype_CDMAEVDORev0),
+      CTRadioAccessTechnologyCDMAEVDORevA : @(GDTCORNetworkMobileSubtype_CDMAEVDORevA),
+      CTRadioAccessTechnologyCDMAEVDORevB : @(GDTCORNetworkMobileSubtype_CDMAEVDORevB),
+      CTRadioAccessTechnologyeHRPD : @(GDTCORNetworkMobileSubtype_HRPD),
+      CTRadioAccessTechnologyLTE : @(GDTCORNetworkMobileSubtype_LTE),
+    };
+    networkInfo = [[CTTelephonyNetworkInfo alloc] init];
+  });
+  NSNumber *networkMobileSubtype =
+      ctRadioAccessTechnologyToNetworkSubTypeMessage[networkInfo.currentRadioAccessTechnology];
```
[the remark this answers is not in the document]
doudounan: If the device isn’t registered on any network, this property is nil. I will check it before fetch from dict.
ref: https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/1616895-currentradioaccesstechnology?language=objc

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -49,6 +50,36 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
...
+      CTRadioAccessTechnologyWCDMA : @(GDTCORNetworkMobileSubtype_WCDMA),
+      CTRadioAccessTechnologyHSDPA : @(GDTCORNetworkMobileSubtype_HSDPA),
+      CTRadioAccessTechnologyHSUPA : @(GDTCORNetworkMobileSubtype_HSUPA),
+      CTRadioAccessTechnologyCDMA1x : @(GDTCORNetworkMobileSubtype_CDMA1x),
+      CTRadioAccessTechnologyCDMAEVDORev0 : @(GDTCORNetworkMobileSubtype_CDMAEVDORev0),
+      CTRadioAccessTechnologyCDMAEVDORevA : @(GDTCORNetworkMobileSubtype_CDMAEVDORevA),
+      CTRadioAccessTechnologyCDMAEVDORevB : @(GDTCORNetworkMobileSubtype_CDMAEVDORevB),
+      CTRadioAccessTechnologyeHRPD : @(GDTCORNetworkMobileSubtype_HRPD),
+      CTRadioAccessTechnologyLTE : @(GDTCORNetworkMobileSubtype_LTE),
+    };
```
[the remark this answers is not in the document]
doudounan: Seems not. 
ref: https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/radio_access_technology_constants?language=objc

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -52,31 +52,35 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
 
 GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
 #if TARGET_OS_IOS
-  static NSDictionary<NSString *, NSNumber *> *ctRadioAccessTechnologyToNetworkSubTypeMessage;
+  static NSDictionary<NSString *, NSNumber *> *CTRadioAccessTechnologyToNetworkSubTypeMessage;
   static CTTelephonyNetworkInfo *networkInfo;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
```
mikehaney24: I think this is doable because all the members of the dictionary compile-time constants. I think you can probably discard the dispatch_once and just do
```
   static NSDictionary<NSString *, NSNumber *> *CTRadioAccessTechnologyToNetworkSubTypeMessage = @{
 <all the stuff here>
    };	   
 };
```

If that doesn't work you can leave as-is

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -52,31 +52,35 @@ BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags) {
...
+      CTRadioAccessTechnologyeHRPD : @(GDTCORNetworkMobileSubtypeHRPD),
+      CTRadioAccessTechnologyLTE : @(GDTCORNetworkMobileSubtypeLTE),
     };
     networkInfo = [[CTTelephonyNetworkInfo alloc] init];
   });
-  NSNumber *networkMobileSubtype =
-      ctRadioAccessTechnologyToNetworkSubTypeMessage[networkInfo.currentRadioAccessTechnology];
-
-  return networkMobileSubtype.intValue;
+  NSString *networkCurrentRadioAccessTechonlogy = networkInfo.currentRadioAccessTechnology;
```
mikehaney24: s/Techonlogy/Technology/

GoogleDataTransportCCTSupport/GDTCCTLibrary/Private/GDTCCTNanopbHelpers.h
```diff
@@ -23,6 +23,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** A string sets in customPrioritizationParams as a key paired to @YES if current event needs to
```
mikehaney24: */ on a newline by itself. here and below
```
/** A string sets in customPrioritizationParams as a key paired to @YES if current event needs to
 * populate network connection info data, @NO otherwise.
 */
```

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: It gets a error "initializer element is not a compile-time constant", I will leave as-is.

mikehaney24: SGTM

mikehaney24:
LGTM, Approval on green checks.

PR #4960, opened 2020-02-24
Fix radioacesstechnology api deprecated for ios 12.0 or newer

doudounan:
Fix radioacesstechnology api deprecated for ios 12.0 or newer

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -72,7 +72,7 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
     networkInfo = [[CTTelephonyNetworkInfo alloc] init];
   });
   NSString *networkCurrentRadioAccessTechnology;
-  if(@available(iOS 13, *)) {
+  if(@available(iOS 12, *)) {
     NSDictionary<NSString *, NSString *> *networkCurrentRadioAccessTechnologyDict = networkInfo.serviceCurrentRadioAccessTechnology;
     if(networkCurrentRadioAccessTechnologyDict){
       networkCurrentRadioAccessTechnology = networkCurrentRadioAccessTechnologyDict.allValues[0];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -75,7 +75,9 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
   if (@available(iOS 12, *)) {
     NSDictionary<NSString *, NSString *> *networkCurrentRadioAccessTechnologyDict =
         networkInfo.serviceCurrentRadioAccessTechnology;
-    if (networkCurrentRadioAccessTechnologyDict) {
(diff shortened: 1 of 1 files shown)


paulb777:
We don't yet have automated tests for Catalyst, so please test Catalyst manually.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -71,7 +71,16 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
     };
     networkInfo = [[CTTelephonyNetworkInfo alloc] init];
   });
-  NSString *networkCurrentRadioAccessTechnology = networkInfo.currentRadioAccessTechnology;
+  NSString *networkCurrentRadioAccessTechnology;
+  if (@available(iOS 12, *)) {
+    NSDictionary<NSString *, NSString *> *networkCurrentRadioAccessTechnologyDict =
+        networkInfo.serviceCurrentRadioAccessTechnology;
+    if (networkCurrentRadioAccessTechnologyDict) {
+      networkCurrentRadioAccessTechnology = networkCurrentRadioAccessTechnologyDict.allValues[0];
```
morganchen12: Based on the [documentation for this method](https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/3024510-servicecurrentradioaccesstechnol?language=objc) we should be checking the keys and not the values. 

>The keys for the `serviceCurrentRadioAccessTechnology` dictionary are `NSString` objects, each of which represents a service. The entry associated with a key is nil if the service is not registered on any network.
>
>Although the actual value of a key isn’t important, you can also use it to get the carrier information associated with the service. To do so, pass the key to the `serviceSubscriberCellularProviders` dictionary.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -75,7 +75,9 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
   if (@available(iOS 12, *)) {
     NSDictionary<NSString *, NSString *> *networkCurrentRadioAccessTechnologyDict =
         networkInfo.serviceCurrentRadioAccessTechnology;
-    if (networkCurrentRadioAccessTechnologyDict) {
+    if (networkCurrentRadioAccessTechnologyDict.count) {
+      // In iOS 12, multiple radio technologies can be captured. We prefer not particular radio
+      // tech to another, so we'll just return the first value in the dictionary.
       networkCurrentRadioAccessTechnology = networkCurrentRadioAccessTechnologyDict.allValues[0];
```
mikehaney24: This needs to be allKeys, not allValues

doudounan: The keys can be associated with other apis, like serviceSubscriberCellularProviders dictionary, but for the networkCurrentRadioAccessTechnology string we want is stored in the values.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Thank you, Morgan. I have changed the if condition to check dictionary.count. And for the networkCurrentRadioAccessTechnology string we want is stored in the values.

PR #4961, opened 2020-02-24
Remove Firebase Database specific log levels in favour FIRLoggerLevel

### LINKED ISSUE #4916: Path .info/connected returns NO with good connection

    file a Github issue.
    with the `firebase` tag.
    google group.

  * Xcode version: 11.3
  * Firebase SDK version: 6.16.0
  * Firebase Component: Database
  * Component version: 6.1.4

#### Steps to reproduce:

We use the `.info/connected` to make sure Firebase is online but it quite often returns NO even when the phone internet connection is working fine and without any issues. Not sure if its a bug, a mistake in our implementation or if thats normal to happen.

#### Relevant Code:

```
Database.database().reference(withPath: ".info/connected").observe(.value)
        {[weak self] (snapshot) in
            
            if let unwrappedIsConnected = snapshot.value as? Bool {
                
                if(unwrappedIsConnected){
                    
                    SwiftMessages.hide(id: "messageCantReachServer")
                    
                    if let unwrappedTask = self?.task {
                        
                        unwrappedTask.cancel()
                    }
                } else {
            
                    self?.task = DispatchWorkItem { self?.showMessageCantReachServer() }
                    
                    if let unwrappedTask = self?.task {
                        
                        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 3, execute: unwrappedTask)
                    }
                }
            }
        }
```

### LINKED ISSUE #4958: Firebase Database debug logs don't respect `-FIRDebugEnabled` flag.
Firebase Database debug logs are not enabled by `-FIRDebugEnabled`. Reproducible in Firebase Database 6.1.4.

Uncovered in #4916.

maksymmalyhin:
Fixes #4958.

Currently Firebase Database SDK uses additional log levels on top of `FIRLoggerLevel` which is a bit confusing and inconsistent with other Firebase SDK.

- `FLogLevel` and related code from Database SDK removed
- `+ [FIRDatabase setLoggingEnabled:]` method deprecated

### DIFF
--- a/Firebase/Database/Utilities/FUtilities.h
+++ b/Firebase/Database/Utilities/FUtilities.h
@@ -28,8 +28,6 @@
 + (NSError *)errorForStatus:(NSString *)status andReason:(NSString *)reason;
 + (NSNumber *)intForString:(NSString *)string;
 + (NSString *)ieee754StringForNumber:(NSNumber *)val;
-+ (void)setLoggingEnabled:(BOOL)enabled;
-+ (BOOL)getLoggingEnabled;
 
 + (NSString *)minName;
 + (NSString *)maxName;
@@ -41,14 +39,6 @@
 
 @end
 
-typedef enum {
-    FLogLevelDebug = 1,
-    FLogLevelInfo = 2,
-    FLogLevelWarn = 3,
-    FLogLevelError = 4,
-    FLogLevelNone = 5
-} FLogLevel;
-
 // Log tags
 FOUNDATION_EXPORT NSString *const kFPersistenceLogTag;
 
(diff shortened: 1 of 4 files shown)


paulb777:
Also needs an API review @ryanwilson

Firebase/Database/Utilities/FUtilities.h
```diff
@@ -41,42 +39,26 @@
...
-        }                                                                      \
+        FIRLogDebug(kFIRLoggerDatabase, (code), (format), ##__VA_ARGS__);      \
     } while (0)
 
 #define FFInfo(code, format, ...)                                              \
     do {                                                                       \
-        if (FFIsLoggingEnabled(FLogLevelInfo)) {                               \
-            FIRLogError(kFIRLoggerDatabase, (code), (format), ##__VA_ARGS__);  \
-        }                                                                      \
+        FIRLogError(kFIRLoggerDatabase, (code), (format), ##__VA_ARGS__);      \
```
ryanwilson: Presumably this should be `FIRLogInfo` instead of `FIRLogError` - I realize it was just a copy paste from before but that could also be why Database is particularly chatty during logs.

ryanwilson: Or, maybe since there's no `FFError`, we should rename it as such.

maksymmalyhin: Good catch! I'll update it to use proper FIRLogger level if we decide to move ahead with the change.

schmidt-sebastian: This seems like a bug more than a conscious decision. I can't come up with a reason why we would log INFOs at a higher level than WARN.

Can you update to use the proper level here? Please also add a changelog.

maksymmalyhin: I guess,  the intention was to make Database logging level to suppress the global logging level (if global level if error but DB level is Info, then DB logs are printed anyway). Doesn't look really readable though. @schmidt-sebastian Do you still think we will need to change it even if we decide to postpone the public API change? See also discussion in go/firebase-database-ios-log-level-api.

schmidt-sebastian: I can't get myself to recommend that you keep it as is, so I would suggest changing it. Info logs shouldn't log at a higher level than warning logs. Let's just make sure to make it very obvious in the release notes.

ryanwilson: If my search was right there’s only a single log that uses this function so it doesn’t have a super high impact either way.

maksymmalyhin:
Decided to postpone until a universal solution to control log level per component is implemented (see go/firebase-database-ios-log-level-api for details).

PR #4963, opened 2020-02-24
Move sample independent watch app under Example folder

chliangGoogle:
instead of the messaging's  unit test app folder


chliangGoogle:
revert this as I realize I pull from the wrong branch.

PR #4955, opened 2020-02-21
Make Firebase Storage compile and work in independent watch app

chliangGoogle:
I added a simple use case of Storage that apple watch is able to load image successfully.

PR #4965, opened 2020-02-25
Wrap SCNetworkReachabilityFlags and fix Catalyst warning

doudounan:
1 Wrap SCNetworkReachabilityFlags in GDTCORPlatform.
2 Fix Catalyst warning in GDTCORPlatform.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/Public/GDTCORPlatform.h
+++ b/GoogleDataTransport/GDTCORLibrary/Public/GDTCORPlatform.h
@@ -75,13 +75,13 @@ typedef NS_ENUM(NSInteger, GDTCORNetworkMobileSubtype) {
 BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags);
 #endif
 
-/** Generates an enum message GDTCORNetworkType representing Network connection type.
+/** Generates an enum message GDTCORNetworkType representing network connection type.
  *
  * @return A GDTCORNetworkType representing network connection type.
  */
 GDTCORNetworkType GDTCORNetworkTypeMessage(void);
 
-/** Generates an enum message GDTCORNetworkMobileSubtype representing Network connection mobile
+/** Generates an enum message GDTCORNetworkMobileSubtype representing network connection mobile
  * subtype.
  *
  * @return A GDTCORNetworkMobileSubtype representing network connection mobile subtype.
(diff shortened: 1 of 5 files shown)


chliangGoogle:
Can you also add watchos along with other platforms in https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/datatransport.yml? This helps prevent errors again like presubmit.

doudounan:
> Can you also add watchos in https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/datatransport.yml? This helps prevent errors again like presubmit.

Yes, we definitely should.

mikehaney24:
> #4955

Standard pod lib lint doesn't do anything, so making this change doesn't do anything. watchOS also doesn't support unit tests. The reason why it works in storage is because they use a different build and test system. Doudou or myself would have to replicate that infra here in GDT.

pod gen can't even generate a watchOS project, so I'm not sure anything is building.

```
 -> GoogleDataTransport
[!] Skipping test spec `GoogleDataTransport/Tests-Unit` on platform `watchos` since it is not supported.
 -> GoogleDataTransport (4.0.0)
    - NOTE  | xcodebuild:  note: Using new build system
```

GoogleDataTransport/GDTCORLibrary/Public/GDTCORPlatform.h
```diff
@@ -68,6 +75,12 @@ typedef NS_ENUM(NSInteger, GDTCORNetworkMobileSubtype) {
 BOOL GDTCORReachabilityFlagsContainWWAN(SCNetworkReachabilityFlags flags);
 #endif
 
+/** Generates an enum message GDTCORNetworkType representing Network connection type.
```
mikehaney24: network

chliangGoogle:
I see. Isn't storage just using the new github action? I thought as long as we copied the yml file, we've used the same build and test system.

mikehaney24:
> I see. Isn't storage just using the new github action? I thought as long as we copied the yml file, we've used the same build and test system.

They're still using GHA for CI, but it calls into `scripts/build.sh`, whereas GDT's workflow calls `scripts/pod_lib_lint.sh`

For GDT, we'd need to probably add something back to build.sh for our watchOS build that builds and runs a manually created watchOS test project

chliangGoogle:
I was planning just duplicate the yml in messaging, looks like there are more work needs to be done here. Thanks for the explanation! I will look into more for messaging as well.

PR #4970, opened 2020-02-25
Add github action for Messaging 

chliangGoogle:
Also move watchOS app under Example folder


.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+    - cron:  '0 7 * * *'
+
+jobs:
+  messaging:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Install Secret GoogleService-Info.plist
```
paulb777: Shouldn't need the secret for Messaging

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+    - '.github/workflows/messaging.yml'
+  schedule:
+    # Run every day at 11pm (PST) - cron uses UTC times
+    - cron:  '0 7 * * *'
+
+jobs:
+  messaging:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v1
```
wilhuff: Prefer `actions/checkout@v2`, which is much faster since it does a shallow clone.

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Install Secret GoogleService-Info.plist
+      env:
+         plist_secret: ${{ secrets.StoragePlistSecret }}
+      run: ./scripts/decrypt_gha_secret.sh scripts/gha-encrypted/storage-db-plist.gpg \
+          FirebaseStorage/Tests/Integration/Resources/GoogleService-Info.plist "$plist_secret"
+    - name: BuildAndTest # can be replaced with pod lib lint with CocoaPods 1.10
```
paulb777: Storage shouldn't be built from messaging.yml

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: scripts/third_party/travis/retry.sh scripts/build.sh Storage all
```
wilhuff: Did you mean "Messaging" instead of "Storage" here? (and possibly above)

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: scripts/setup_bundler.sh
+    - name: Build and test
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --skip-tests --platforms=watchos
+
+
+  messaging-cron-only:
+    runs-on: macos-latest
+    if: github.event_name == 'schedule'
+    strategy:
+      matrix:
```
paulb777: matrix doesn't work here since the --use-modular-headers with watchos won't work without --skip-tests

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: scripts/setup_bundler.sh
+    - name: Install Secret GoogleService-Info.plist
+      env:
+         plist_secret: ${{ secrets.StoragePlistSecret }}
+      run: ./scripts/decrypt_gha_secret.sh scripts/gha-encrypted/storage-db-plist.gpg \
+          FirebaseStorage/Tests/Integration/Resources/GoogleService-Info.plist "$plist_secret"
+    - name: BuildAndTest # can be replaced with pod lib lint with CocoaPods 1.10
+      run: scripts/third_party/travis/retry.sh scripts/build.sh Storage all
+
+  pod_lib_lint:
```
wilhuff: This name is spelled inconsistently with `messaging-cron-only` (i.e. snake_case vs kebab-case). Rename these to `pod-lib-lint`?

Also, if you're up for it, fix core.yml as well which also has this inconsistency.

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Build and test
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --skip-tests --platforms=watchos
+
+
+  messaging-cron-only:
+    runs-on: macos-latest
+    if: github.event_name == 'schedule'
+    strategy:
+      matrix:
+        target: [ios, tvos, macos, watchos]
```
wilhuff: This is going to run each of these in combination with the sets of flags below. This will include `watchos` in combination with `--use-modular-headers` which won't include `--skip-tests`. This probably isn't what you want. You probably want to have a `messaging-watchos-cron-only` target as well.

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,71 @@
...
+        flags: [
+          '--use-modular-headers',
+          '--skip-tests --use-libraries'
+        ]
+    needs: pod_lib_lint
```
wilhuff: Is this dependency necessary? If so you'll need to depend on `pod-lib-lint-watchos` when the target is watchos (another argument for a separate cron-only target).

.github/workflows/messaging.yml
chliangGoogle: I think it's because core also already has a 'pod-lib-lint'. I think the difference is one with cron job running pod lib lint?

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    strategy:
+      matrix:
+        target: [ios, tvos, macos]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Build and test
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }}
```
mikehaney24: it might be possible to add watchos to the matrix here with something like this:
```
if: matrix.target == 'watchos'
      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --skip-tests --platforms=${{ matrix.target }}
if: matrix.target == 'watchos'
      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }}
```

.github/workflows/messaging.yml
chliangGoogle: Oh you meant the format. Got it.

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,76 @@
...
+      run: scripts/setup_bundler.sh
+    - name: Build and test
+      if: matrix.target == 'watchos'
+        run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --skip-tests --platforms=${{ matrix.target }}
+      if: matrix.target != 'watchos'
+        run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }}
+
+  messaging-cron-only:
+    runs-on: macos-latest
+    if: github.event_name == 'schedule'
```
paulb777: To test, comment this out and uncomment when it goes green.

.github/workflows/messaging.yml
```diff
@@ -0,0 +1,76 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: PodLibLint Messaging Cron
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }} ${{ matrix.flags }}
+
+  messaging-watchos-cron-only:
+    runs-on: macos-latest
+    if: github.event_name == 'schedule'
```
paulb777: Here too

.github/workflows/messaging.yml
chliangGoogle: I wonder if the second condition will not happen if the first one is not passed. 
looking at the if: github.event_name == 'schedule'

mikehaney24: Totally possible! I suspect it can be done, but my knowledge of this syntax is so low, I'm not a good resource. Feel free to change it back if it's too much trouble.

.github/workflows/messaging.yml
chliangGoogle: comment it out and it went green.

.github/workflows/messaging.yml
chliangGoogle: Same here.

PR #4978, opened 2020-02-26
Add FIRCLSReportUploader unit tests for supporting multiple endpoints

jasonhu-g:


Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
```diff
@@ -113,13 +113,9 @@ - (BOOL)prepareAndSubmitReport:(FIRCLSInternalReport *)report
                        self.dataSource.settings.shouldUseNewReportEndpoint);
 
         if (self.dataSource.settings.shouldUseNewReportEndpoint) {
-          // For the new endpoint, just move the .clsrecords from "processing" -> "prepared"
```
jasonhu-g: Note to self: Add back comment

Crashlytics/UnitTests/Mocks/FIRCLSMockFileManager.h
```diff
@@ -26,13 +26,25 @@
...
 
+@property(nonatomic, copy) NSString *removedItemAtPath_path;
+
 // Overriding the method for testing Settings
 - (BOOL)removeItemAtPath:(NSString *)path;
 
+// Overrides moveItemAtPath if set
+@property(nonatomic) NSNumber *moveItemAtPathResult;
+@property(nonatomic, copy) NSString *moveItemAtPath_path;
+@property(nonatomic, copy) NSString *moveItemAtPath_destDir;
```
samedson: How do you feel about checking the files directly instead of setting variables?

jasonhu-g: Discussed offline:
I have a concern with making changes to the file system in unit tests. I have those tests cause unreliability to tests. One of the things I don’t like with the mock file manager is that it actually calls to the actual file manager. Mocks should not be calling into the product equivalent of itself. We have a good opportunity to fix these unit tests when we remove the experiment and perform a larger refactor.

PR #4984, opened 2020-02-26
Fixes for compatibility with builds in firebase-cpp-sdk

wilhuff:
This makes it possible to embed the Firestore CMake build contained in this repo in the public C++ CMake build.

This is enough to get the firebase-cpp-sdk build working with tests disabled. More changes may be required for tests enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,7 +199,6 @@ else()
   )
 endif()
 
-add_alias(protobuf::libprotobuf libprotobuf)
 if(CXX_CLANG OR CXX_GNU)
   target_compile_options(
     libprotobuf
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,12 @@ endif()
 enable_language(C)
 enable_language(CXX)
 
+option(
+  FIREBASE_IOS_BUILD_TESTS
+  "Enable building of C++ and Objective-C tests for this project"
+  ON
+)
+
 
 list(INSERT CMAKE_MODULE_PATH 0 ${PROJECT_SOURCE_DIR}/cmake)
 include(compiler_id)
@@ -251,7 +257,9 @@ if(APPLE)
 endif()
 
 
-enable_testing()
+if(FIREBASE_IOS_BUILD_TESTS)
+  enable_testing()
+endif()
 include(compiler_setup)
 
 add_subdirectory(FirebaseCore)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,12 @@ endif()
 
 
 # gRPC
+
+# libcurl and c-ares conflict in their usage of this variable. Prevent
+# libcurl's setting of this variable from affecting the c-ares build that's
+# pulled in indirectly via gRPC.
+unset(RANDOM_FILE CACHE)
+
 find_package(OpenSSL QUIET)
 if(OPENSSL_FOUND)
   set(gRPC_SSL_PROVIDER package CACHE STRING "Use external OpenSSL")
--- a/cmake/compiler_setup.cmake
+++ b/cmake/compiler_setup.cmake
@@ -117,14 +117,6 @@ if(MSVC)
 endif()
 
 foreach(flag ${common_flags} ${c_flags})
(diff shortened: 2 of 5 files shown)


cmake/compiler_setup.cmake
```diff
@@ -116,14 +116,6 @@ if(MSVC)
   )
 endif()
 
-foreach(flag ${common_flags} ${c_flags})
```
var-const: What's the reason to delete these lines?

cmake/external_rules.cmake
```diff
@@ -36,9 +36,22 @@ function(download_external_sources)
...
 function(add_external_subdirectory NAME)
-  add_subdirectory(
-    ${FIREBASE_BINARY_DIR}/external/src/${NAME}
-    ${FIREBASE_BINARY_DIR}/external/src/${NAME}-build
-    EXCLUDE_FROM_ALL
-  )
+  string(TOUPPER ${NAME} UPPER_NAME)
+  if (NOT EXISTS ${${UPPER_NAME}_SOURCE_DIR})
+    set(${UPPER_NAME}_SOURCE_DIR "${FIREBASE_BINARY_DIR}/external/src/${NAME}")
+    set(${UPPER_NAME}_SOURCE_DIR "${${UPPER_NAME}_SOURCE_DIR}" PARENT_SCOPE)
```
var-const: Can these two calls to `set` be rolled into one? (i.e. `set(${UPPER_NAME}_SOURCE_DIR "${FIREBASE_BINARY_DIR}/external/src/${NAME}" PARENT_SCOPE)`)

cmake/external_rules.cmake
```diff
@@ -36,9 +36,22 @@ function(download_external_sources)
...
+  endif()
+
+  if (NOT EXISTS ${${UPPER_NAME}_BINARY_DIR})
+    set(${UPPER_NAME}_BINARY_DIR "${${UPPER_NAME}_SOURCE_DIR}-build")
+    set(${UPPER_NAME}_BINARY_DIR "${${UPPER_NAME}_BINARY_DIR}" PARENT_SCOPE)
+  endif()
+
+  if (EXISTS "${${UPPER_NAME}_SOURCE_DIR}/CMakeLists.txt")
```
var-const: `else` branch does nothing -- it is an expected case as well, or should it report an error?

CMakeLists.txt
```diff
@@ -187,7 +199,6 @@ else()
   )
 endif()
 
-add_alias(protobuf::libprotobuf libprotobuf)
```
var-const: What's the reason to remove this line?

scripts/binary_to_array.py
```diff
@@ -48,8 +48,6 @@
 import logging
 import os
```
var-const: What was the problem with the previous state of this script?

cmake/compiler_setup.cmake
wilhuff: These change global state and apply our preferences for flags to all compile actions. This breaks `firebase-cpp-sdk` and its dependencies which don't hold themselves to anywhere near as high a standard.

cmake/external_rules.cmake
wilhuff: No. Each `set` call can only operate on a single scope at a time.

CMakeLists.txt
wilhuff: The upstream protobuf CMakeLists.txt now does this. Some previous upgrade failed to notice this and it's just cruft now.

scripts/binary_to_array.py
wilhuff: Most of the context for these changes was in the individual commit messages (e.g. [this one](https://github.com/firebase/firebase-ios-sdk/pull/4984/commits/07d1e5e1a72abb405e11d992b6f73789317ab4da)).

This code previously expected to be run from within a git repository but this assumption is flawed:

  * iOS users can download an archive of the source tree, in which case there's no `.git` folder
  * `firebase-cpp-sdk` actually does this to download its copy of `firebase-ios-sdk`

cmake/external_rules.cmake
wilhuff: Its an expected case that matches behavior in `firebase-cpp-sdk`. I may revisit this later.

PR #4993, opened 2020-02-28
Move UpdateFirebasePod to Swift's ArgumentParser library.

ryanwilson:
This was introduced in https://swift.org/blog/argument-parser/, leaving a nice clean way to parse command line arguments.

Example output with no arguments:

```
$ swift run UpdateFirebasePod
Error: Missing expected argument '--git-root <git-root>'
Usage: firebase-pod-updater --git-root <git-root> --releasing-pods <releasing-pods>
```


ZipBuilder/Sources/UpdateFirebasePod/main.swift
```diff
@@ -16,99 +16,120 @@
...
+        var versionStartIndex = contents.index(range.upperBound, offsetBy: 2)
+        while !contents[versionStartIndex].isWholeNumber {
+          versionStartIndex = contents.index(versionStartIndex, offsetBy: 1)
+        }
+        var versionEndIndex = contents.index(versionStartIndex, offsetBy: 1)
+        while contents[versionEndIndex] != "'" {
+          versionEndIndex = contents.index(versionEndIndex, offsetBy: 1)
+        }
+        contents.removeSubrange(versionStartIndex ... versionEndIndex)
+        contents.insert(contentsOf: version + "'", at: versionStartIndex)
```
ryanwilson: This just moved from the root of the file to inside the new type, along with not using the `FirebasePod` type.

ZipBuilder/Sources/UpdateFirebasePod/main.swift
```diff
@@ -16,99 +16,120 @@
...
+    let newVersions = getExpectedVersions()
+    updateFirebasePod(newVersions: newVersions)
+    print("Updating Firebase pod for version \(String(describing: newVersions["Firebase"]!))")
+
+    // Get the time since the tool start.
+    let secondsSinceStart = -Int(buildStart.timeIntervalSinceNow)
+    print("""
+    Time profile:
+      It took \(secondsSinceStart) seconds (~\(secondsSinceStart / 60)m) to update the Firebase pod.
+    """)
```
ryanwilson: This moved from the bottom of the file to the `run()` block.

ryanwilson:
I'll move the ZipBuilder itself over once we think this is at a reasonable place!

ZipBuilder/Package.swift
```diff
@@ -26,18 +26,19 @@ let package = Package(
     .executable(name: "UpdateFirebasePod", targets: ["UpdateFirebasePod"]),
   ],
   dependencies: [
+    .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
```
paulb777: Do we really want to float the dependency versions?

ryanwilson: I was going off their comment that says they're committed to source-stability within minor versions at the bottom of this section:

https://forums.swift.org/t/parsing-file-urls-with-argumentparser/34173

I'm happy to make this exact though, I'll do so now.

ryanwilson:
> The command is `swift run UpdateFirebasePod` but the usage is `firebase-pod-updater`. Which is it?

In `Package.swift` we still call it `UpdateFirebasePod` but the it looks like it expects to be called `firebase-pod-updater` (auto-generated from the struct name `FirebasePodUpdater`). I can change the executable in the package manifest to be called `firebase-pod-updater` if that's better, it seems that's the style they're going for.

ryanwilson:
> > The command is `swift run UpdateFirebasePod` but the usage is `firebase-pod-updater`. Which is it?
> 
> In `Package.swift` we still call it `UpdateFirebasePod` but the it looks like it expects to be called `firebase-pod-updater` (auto-generated from the struct name `FirebasePodUpdater`). I can change the executable in the package manifest to be called `firebase-pod-updater` if that's better, it seems that's the style they're going for.

Yep, based on https://github.com/apple/swift-argument-parser/blob/339baea0a5d8c9b26c6e9be47df8e713f6e26234/Package.swift#L31-L42 they're expecting the target to have the same name as the tool.

ryanwilson:
@paulb777 before merging, did you want me to change the target name in `Package.swift` to match `firebase-pod-updater`?

paulb777:
@ryanwilson It seems either that or fix the Usage message?

PR #5005, opened 2020-03-03
Support AB testing of Firebase In-App Messages

christibbs:
Supports experimental Firebase in-app messages (cross-functional with Firebase AB Testing).

Adds functionality to the ABT SDK:
- A method that cleans up old experiments that are no longer running
- A method that directly activates an experiment

FIAM SDK changes:
- Parses experimental metadata from FIAM backend, attaches the experiment payload to `InAppMessageDisplayMessage`
- On message fetch, cleans up experiments that are no longer running
- On message impression, sets the experiment to be activated
- ~~Removes deprecated initializers for message subclasses (internal bug ID 134409779).~~ Removed this for now, looks like the UI test app uses them

### DIFF
--- a/FirebaseABTesting/Sources/FIRExperimentController.m
+++ b/FirebaseABTesting/Sources/FIRExperimentController.m
@@ -291,4 +291,34 @@ NSArray *ABTExperimentsToClearFromPayloads(
   }
   return timestamp;
 }
+
+- (void)validateRunningExperimentsForServiceOrigin:(NSString *)origin
+                                          payloads:(NSArray<ABTExperimentPayload *> *)payloads {
+  ABTConditionalUserPropertyController *controller =
+      [ABTConditionalUserPropertyController sharedInstanceWithAnalytics:_analytics];
+  
+  FIRLifecycleEvents *lifecycleEvents = [[FIRLifecycleEvents alloc] init];
+
+  // Get the list of experiments from Firebase Analytics.
+  NSArray<NSDictionary <NSString *, NSString *> *> *activeExperiments =
+      [controller experimentsWithOrigin:origin];
+  
+  NSMutableSet *runningExperimentIDs = [NSMutableSet setWithCapacity:payloads.count];
+  for (ABTExperimentPayload *payload in payloads) {
+    [runningExperimentIDs addObject:payload.experimentId];
+  }
+
+  for (NSDictionary <NSString *, NSString *> *activeExperimentDictionary in activeExperiments) {
+    NSString *experimentID = activeExperimentDictionary[@"experimentId"];
+    if (![runningExperimentIDs containsObject:experimentID]) {
+      NSString *variantID = activeExperimentDictionary[@"variantId"];
+      
+      [controller clearExperiment:experimentID
+                        variantID:variantID
+                       withOrigin:origin
+                          payload:nil events:lifecycleEvents];
+    }
+  }
+}
+
 @end
--- a/FirebaseABTesting/Sources/FIRExperimentController.m
+++ b/FirebaseABTesting/Sources/FIRExperimentController.m
@@ -293,7 +293,7 @@ NSArray *ABTExperimentsToClearFromPayloads(
 }
 
 - (void)validateRunningExperimentsForServiceOrigin:(NSString *)origin
-                                          payloads:(NSArray<ABTExperimentPayload *> *)payloads {
+                         runningExperimentPayloads:(NSArray<ABTExperimentPayload *> *)payloads {
   ABTConditionalUserPropertyController *controller =
       [ABTConditionalUserPropertyController sharedInstanceWithAnalytics:_analytics];
 
@@ -309,9 +309,9 @@ NSArray *ABTExperimentsToClearFromPayloads(
   }
 
   for (NSDictionary<NSString *, NSString *> *activeExperimentDictionary in activeExperiments) {
-    NSString *experimentID = activeExperimentDictionary[@"experimentId"];
+    NSString *experimentID = activeExperimentDictionary[@"name"];
     if (![runningExperimentIDs containsObject:experimentID]) {
-      NSString *variantID = activeExperimentDictionary[@"variantId"];
+      NSString *variantID = activeExperimentDictionary[@"value"];
 
       [controller clearExperiment:experimentID
                         variantID:variantID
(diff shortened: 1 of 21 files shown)


FirebaseInAppMessaging/Sources/Flows/FIRIAMMsgFetcherUsingRestful.m
```diff
@@ -184,6 +186,19 @@ - (void)fetchMessagesWithImpressionList:(NSArray<FIRIAMImpressionRecord *> *)imp
                     FIRLogDebug(kFIRLoggerInAppMessaging, @"I-IAM130012",
                                 @"API request for fetching messages and parsing the response was "
                                  "successful.");
+
+                    // Validate running experiments with ABT.
+                    NSMutableArray *runningExperiments = [[NSArray array] mutableCopy];
+                    for (FIRIAMMessageDefinition *messageDefinition in messages) {
```
chliangGoogle: Why not just [[NSMutableArray alloc] init]

FirebaseInAppMessaging.podspec
```diff
@@ -44,6 +45,7 @@ See more product details at https://firebase.google.com/products/in-app-messagin
   s.ios.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
   s.dependency 'FirebaseInstanceID', '~> 4.0'
   s.dependency 'GoogleDataTransportCCTSupport', '~> 1.0'
+	s.dependency 'FirebaseABTesting'
```
chliangGoogle: The format seems off, I wonder if we need to specify a version and also bump up ABTesting and this podspec's version in case we forgot in next release

FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -209,6 +211,13 @@ - (void)impressionDetectedForMessage:(FIRInAppMessagingDisplayMessage *)inAppMes
     return;
   }
 
+  // If this is an experimental FIAM, activate the experiment.
+  if (inAppMessage.campaignInfo.experimentPayload) {
+    [[FIRExperimentController sharedInstance]
+        activateExperiment:inAppMessage.campaignInfo.experimentPayload
```
chliangGoogle: Just curious why not activate experiments when we get the payload in FIRIAMFetchUsingRestful class.

christibbs: When we get the payload, the messages are cached and we wait for the right time to show the message. We only activate the experiment after the message has been shown (here). Analogous to when an RC experiment gets its activation event fired.

FirebaseInAppMessaging.podspec
paulb777: Yes should depend on next ABTesting minor version and the ABTesting podspec should be bumped.

FirebaseABTesting/Sources/FIRExperimentController.m
```diff
@@ -291,4 +291,51 @@ - (NSTimeInterval)latestExperimentStartTimestampBetweenTimestamp:(NSTimeInterval
...
+    if (![runningExperimentIDs containsObject:experimentID]) {
+      NSString *variantID = activeExperimentDictionary[@"value"];
+
+      [controller clearExperiment:experimentID
+                        variantID:variantID
+                       withOrigin:origin
+                          payload:nil
+                           events:lifecycleEvents];
+    }
+  }
```
morganchen12: For code like this it may be better to make a lightweight `NSCopying` Experiment class. Then this method body would be more readable:

```objc
NSArray<FIAMExperiment *> *experiments = /* ... */

// ...

for (FIAMExperiment *experiment in experiments) {
  if (![runningExperimentIDs containsObject:experiment.identifier]) {
    [controller clearExperiment:experiment withOrigin:origin payload:nil events:lifecycleEvents];
  }
}
```

FirebaseABTesting/Sources/Public/FIRExperimentController.h
```diff
@@ -86,6 +88,21 @@ NS_SWIFT_NAME(ExperimentController)
 /// @param payloads   List of experiment metadata.
 - (NSTimeInterval)latestExperimentStartTimestampBetweenTimestamp:(NSTimeInterval)timestamp
                                                      andPayloads:(NSArray<NSData *> *)payloads;
+
+/// Expires experiments that aren't in the list of running experiment payloads.
+/// @param origin         The originating service affected by the experiment, it is defined at
+///                       Firebase Analytics FIREventOrigins.h.
```
ryanwilson: Since `FIREventOrigins.h` isn't a public header, it's probably better to remove it from this and the below method.

FirebaseABTesting/Sources/FIRExperimentController.m
christibbs: Even though FIAM is the only SDK that uses this method currently, I'd like to keep the methods in the ABT SDK generic. I'll probably refactor it down the road because right now it only consists of methods that RC uses and methods that FIAM uses.

FirebaseInAppMessaging/Sources/Public/FIRInAppMessagingRendering.h
```diff
@@ -16,6 +16,8 @@
 
 #import <UIKit/UIKit.h>
 
+@class ABTExperimentPayload;
+
```
ryanwilson: Nit: this can go away now.

FirebaseABTesting/Sources/Public/FIRExperimentController.h
```diff
@@ -86,6 +86,19 @@ NS_SWIFT_NAME(ExperimentController)
 /// @param payloads   List of experiment metadata.
 - (NSTimeInterval)latestExperimentStartTimestampBetweenTimestamp:(NSTimeInterval)timestamp
                                                      andPayloads:(NSArray<NSData *> *)payloads;
+
+/// Expires experiments that aren't in the list of running experiment payloads.
+/// @param origin         The originating service affected by the experiment.
+/// @param payloads     The list of valid, running experiments.
```
paulb777: nit: indenting

FirebaseInAppMessaging.podspec
```diff
@@ -44,6 +45,7 @@ See more product details at https://firebase.google.com/products/in-app-messagin
   s.ios.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
   s.dependency 'FirebaseInstanceID', '~> 4.0'
   s.dependency 'GoogleDataTransportCCTSupport', '~> 1.0'
+  s.dependency 'FirebaseABTesting', '~> 3.2.0'
```
paulb777: Firebase should allow minor version updates of Firebase deps. Change to '~> 3.2'

PR #5013, opened 2020-03-04
Core M66 changelog update

maksymmalyhin:


FirebaseCore/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
-# Unreleased
+# v6.6.4 -- M66
 - [fixed] Fix Zip Builder module map generation that could cause linker missing
```
paulb777: Didn't this first one go out in M65?

maksymmalyhin: Good catch! Updated.

PR #3546, opened 2019-08-09
Initial nanopb support infra for InAppMessaging

mikehaney24:
To assist with moving to GDT and GDTCCT.

Pre-reqs: protobuf

To re-compile the protos and pod gen the InAppMessaging project, run `Firebase/InAppMessaging/generate_project.sh`

PR #5019, opened 2020-03-04
M66 FIAM GoogleDataTransportCCTSupport dependency bump to '~> 2.0'

maksymmalyhin:


maksymmalyhin:
@christibbs @mikehaney24 I wonder why FIAM depends on GDT?

mikehaney24:
> @christibbs @mikehaney24 I wonder why FIAM depends on GDT?

FIAM was asked to switch from their ad-hoc Clearcut implementation to Firelog a long time ago (b/139060843). To help out, I did some of the work to help them depend on Firelog and put in nanopb infra (#3546). Their side of the work hasn't yet been completed, so the dependency isn't really doing anything, but it's also a non-issue since FirebaseCore has the dep anyway.

PR #5022, opened 2020-03-04
Refactor GDT's storage to be an interface

mikehaney24:
- Events are now removed by eventID
  - Events now generate an ID when they are alloc/inited
  - Copying events also copies their eventID
  - Event hashing takes into account the eventID
  - A static method in GDTCOREvent creates  event IDs as the concat of <session counter><event counter during session>
  - Session counters are incremented and stored in user defaults at app start
- Storage instances register themselves with GDTCORRegistrar like prioritizers and uploaders
- Storage fakes are refactored to be registered with GDTCORRegistrar
- Added some missing docs and fixed some typos
- GDTCORStorage will be renamed in the next PR,  I just wanted to keep this PR as small as possible.


GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -25,6 +25,37 @@
...
+    while (nextEventID >= tensPlace) {
+      tensPlace *= 10;
+    }
+    uint64_t eventID = sessionSequenceNumber * tensPlace + nextEventID;
+    nextEventID++;
+    // This isn't actually concatenating numbers because the first digits aren't the
+    // sessionSequenceNumber
+    result = @(eventID);
+  });
+  return result;
```
maksymmalyhin: I wonder how important is to increment event IDs? If the only requirement for the ID is to be unique, maybe [NSUUID](https://developer.apple.com/documentation/foundation/nsuuid) will provide a simpler and more reliable implementation.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -45,6 +44,12 @@
 
 @implementation GDTCORStorage
 
++ (void)load {
```
maksymmalyhin: AFAIK, any code in `+ (void)load` method contributes to the app startup time. I wonder, may we defer storage registration to later stages, ideally until it's actual first use.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -307,24 +323,33 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
...
+  if ([storedEvents isKindOfClass:[NSMutableOrderedSet class]]) {
+    [(NSMutableOrderedSet *)storedEvents
+        enumerateObjectsUsingBlock:^(GDTCOREvent *_Nonnull obj, NSUInteger idx,
+                                     BOOL *_Nonnull stop) {
+          events[obj.eventID] = obj;
+        }];
+  } else if ([storedEvents isKindOfClass:[NSMutableDictionary class]]) {
+    events = (NSMutableDictionary *)storedEvents;
+  }
+  sharedInstance->_storedEvents = events;
```
maksymmalyhin: Does it still have to be run on `storageQueue`?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: There may be a future requirement for event IDs to be generated sequentially for determining log loss, so this will do for now. Ultimately, I may move to a double like <session counter>.<event counter>.

maksymmalyhin: I think, in order to satisfy the sequential requirement we should take care of a potential (though maybe unlikely) integer overflow. Also, curious how determining log loss will work with multiple devices sending the same ID?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -25,6 +25,37 @@
 
 @implementation GDTCOREvent
 
++ (NSNumber *)nextEventID {
+  static NSInteger sessionSequenceNumber = 1;
```
paulb777: What's the point of initializing to one?  Isn't it always rewritten in the dispatch_once?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -25,6 +25,37 @@
...
+  static unsigned long long nextEventID = 0;
+  static dispatch_queue_t eventIDQueue;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    static NSString *kGDTCORSessionKey = @"GDTCORSession";
+    eventIDQueue = dispatch_queue_create("com.google.GDTCOREventIDQueue", DISPATCH_QUEUE_SERIAL);
+    sessionSequenceNumber = [[NSUserDefaults standardUserDefaults] integerForKey:kGDTCORSessionKey];
```
paulb777: Assuming 0 if doesn't exist?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
paulb777: +1

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -112,12 +116,17 @@ - (void)storeEvent:(GDTCOREvent *)event
                  @"The event should have been serialized to bytes");
     NSError *error = nil;
     NSURL *eventFile = [self saveEventBytesToDisk:event eventHash:event.hash error:&error];
-    GDTCORLogDebug("Event saved to disk: %@", eventFile);
-    completion(eventFile != nil, error);
+    if (!eventFile || error) {
+      GDTCORLogDebug("Event failed to save to disk: %@", error);
```
paulb777: It seems like these two log's should be different levels - info versus error

GoogleDataTransport/GDTCORLibrary/Private/GDTCOREvent_Private.h
```diff
@@ -31,6 +34,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)writeToURL:(NSURL *)fileURL error:(NSError **)error;
 
+/** Generates incrementing event IDs based on a persistent session counter stored in user defaults.
+ *
+ * @return An event ID that is the concatenation of a session counter and event counter. For
+ * example, if this is the 104th session and the 1234th event, the returned ID will be 1041234.
```
paulb777: How is that distinguished from the 41234th event of the 10th session?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: In this case, nothing will directly refer to GDTCORFlatFileStorage and so `+initialize` (the next-best option) will never be called.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: No, since appWillForeground was put onto the queue, this will cause a deadlock. appWillForeground is really the more appropriate place anyway.

GoogleDataTransport/GDTCORLibrary/Private/GDTCOREvent_Private.h
mikehaney24: Good question. I thought a lot about this over the weekend and I think the best solution will be to store the counter value into a file.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: The same devices generating duplicate eventIDs will be fine. Generally, we're hoping that app instances (an period of app usage in which the cache isn't deleted) generate sequential event  IDs. The event IDs joined with some additional information an SDK generates should be able to help determine log loss. I've moved saving the counter value to a file--it seems the best way in a flat file storage system to store a value robust against crashing and lifecycle events.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: Changed.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: Changed.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m

mikehaney24:
Any further comments? If not, I'll merge it later. I'm considering cherry-picking this into master.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
paulb777: I meant - shouldn't this be GDTCORLogError?

GoogleDataTransport/GDTCORLibrary/Private/GDTCOREvent_Private.h
paulb777: I'm not following how this is resolved.

mikehaney24: I need to update these docs, but now that I'm just incrementing a number in a file, there's no session number, it's just a number from 0 to N, so there's no conflict.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: It is, I think you might be looking at an older set of commits.

image

PR #5026, opened 2020-03-05
Fix assertions not printing correctly

mikehaney24:


mikehaney24:
> What was incorrect?

It would console print something like `Assertion failed: The problem is in the file %@`. Formatters didn't work correctly. Now it'll print `Assertion failed: The problem is in the file /Users/.../Somefile.m`

> Why is it in a public header?

The support libraries use the asserts as well.

PR #5027, opened 2020-03-05
Remove setting CMAKE_BUILD_TYPE in CMakeLists.txt

wilhuff:
This really needs to be set on the command-line if it's to be effective,
and setting it here breaks downstream projects that include
firebase-ios-sdk as a subdirectory.


wilhuff:
Travis failures are unrelated.

PR #5028, opened 2020-03-05
Modify watchOS task control for storage, transformer, and cctuploader

doudounan:
- Add beginActivityWithOptions and endActivity for watchOS task control.

- Modify relative methods calling in storage, transformer, cctuploader.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -194,6 +194,18 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
                            selector:@selector(macOSApplicationWillTerminate:)
                                name:NSApplicationWillTerminateNotification
                              object:nil];
+    
+#elif TARGET_OS_WATCH
+    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+    [notificationCenter addObserver:self
+                           selector:@selector(iOSApplicationDidEnterBackground:)
+                               name:NSExtensionHostDidEnterBackgroundNotification
+                             object:nil];
+    [notificationCenter addObserver:self
+                           selector:@selector(iOSApplicationWillEnterForeground:)
+                               name:NSExtensionHostWillEnterForegroundNotification
+                             object:nil];
+
 #endif  // TARGET_OS_IOS || TARGET_OS_TV
   }
   return self;
@@ -201,6 +213,7 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
 
 - (GDTCORBackgroundIdentifier)beginBackgroundTaskWithName:(NSString *)name
                                         expirationHandler:(void (^)(void))handler {
+#if !TARGET_OS_WATCH
   GDTCORBackgroundIdentifier bgID =
       [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
                                                            expirationHandler:handler];
@@ -210,14 +223,19 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
   }
 #endif  // !NDEBUG
   return bgID;
+#endif  // !TARGET_OS_WATCH
+  // TODO: WKExtension backgound tasks handling.
+  return GDTCORBackgroundIdentifierInvalid;
 }
 
 - (void)endBackgroundTask:(GDTCORBackgroundIdentifier)bgID {
+#if !TARGET_OS_WATCH
   if (bgID != GDTCORBackgroundIdentifierInvalid) {
     GDTCORLogDebug("Ending background task with ID:%ld was successful", (long)bgID);
     [[self sharedApplicationForBackgroundTask] endBackgroundTask:bgID];
     return;
   }
+#endif
 }
 
 #pragma mark - App environment helpers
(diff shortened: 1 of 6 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -249,35 +255,32 @@ - (BOOL)isAppExtension {
...
-  if (wkExtensionClass &&
-      [wkExtensionClass respondsToSelector:(NSSelectorFromString(@"sharedExtension"))]) {
-    sharedApplication = [wkExtensionClass sharedExtension];
+  // The processInfo class method returns the shared agent for the current process.
+  Class nsProcessInfoClass = NSClassFromString(@"NSProcessInfo");
+  if (nsProcessInfoClass &&
+      [nsProcessInfoClass respondsToSelector:(NSSelectorFromString(@"processInfo"))]) {
+    sharedInstance = [nsProcessInfoClass processInfo];
   }
 #endif
```
chliangGoogle: Just checking: are we going to take care of macOS here or it is supposed to be do nothing.

doudounan: This sharedInstance are taking responsibility for handling some task to be continually running when app goes into background, I think in macOS, this situation would not be considered. I am not sure about that. I would consult Michael as well.

chliangGoogle: I'm asking because it seems like you are handling macOS at line 265.

doudounan: Seems there are some waste checks for macOS here, I will carefully consider that. Thank you.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -211,9 +211,9 @@ - (instancetype)init {
   return self;
 }
 
+#if !TARGET_OS_WATCH
```
mikehaney24: I'm not sure that disabling this API is the right thing to do. Perhaps we can just implement the watchOS specific functionality in here?

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -223,21 +223,27 @@ - (GDTCORBackgroundIdentifier)beginBackgroundTaskWithName:(NSString *)name
...
     GDTCORLogDebug("Ending background task with ID:%ld was successful", (long)bgID);
     [[self sharedApplicationForBackgroundTask] endBackgroundTask:bgID];
     return;
   }
-#endif  // !TARGET_OS_WATCH
 }
 
+#elif TARGET_OS_WATCH
+- (id<NSObject>)beginActivityWithOptions:(NSActivityOptions)options reason:(NSString *)reason {
+  return [[self sharedNSProcessInfoForBackgroundTask] beginActivityWithOptions:options
```
mikehaney24: From the [watchOS lifecycle docs](https://developer.apple.com/documentation/watchkit/working_with_the_watchos_app_life_cycle?language=objc)
> Because the system can purge suspended apps without warning, use the extension delegate’s applicationDidEnterBackground method to save any data you need to recreate your app’s current state. If needed, you can request additional background execution time by calling the NSProcessInfo class’s performExpiringActivityWithReason:usingBlock: method.

That might mean that one possible solution is to just call the appWillTerminate callbacks when receiving a backgrounding notice on watchOS. Another might be implementing the handleBackgroundTasks method.

It also means that this might not be the right API to call--you might need to call `performExpiringActivityWithReason:usingBlock:` instead. If that's the case, it's not really possible to know how long we'll need to execute. It might be sufficient to just write a line in there that spins the current thread loop for like 10 seconds or something.

My hunch is that we'll want to call `performExpiringActivityWithReason:usingBlock:` and simultaneously call the `appWillTerminate` callbacks to try and force each queue to finish its work. @chliangGoogle what do you think?

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
mikehaney24: Just for clarity: macOS doesn't really have an app lifecycle analogous to anything we do on any other platforms. For our purposes: macOS doesn't have a background state, just a started and stopping state.

GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTUploader.m
```diff
@@ -174,6 +174,7 @@ - (NSString *)FLLAndCSHAPIKey {
 }
 
 - (void)uploadPackage:(GDTCORUploadPackage *)package {
+#if !TARGET_OS_WATCH
```
mikehaney24: The goal of GDTCORPlatform is to provide abstractions and APIs that prevent any other part of the codebase from having to have an #if TARGET_OS_XXXXX in it. Try to figure out an API, or how to use an existing API, to achieve this. We don't want TARGET_OS_XXXXXs all over the codebase. Beginning and ending activities should just happen when we call the appropriate backgroundTask APIs.

doudounan:
For adding wkExtensionDidEnterBackground, what I am trying to do is that let wk extension reacting to "didenterbackground" as "terminating". But I am not quite sure whether this would be working and still working on testing.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -211,9 +212,41 @@ - (instancetype)init {
...
+                                 if (expired && !seconds) {
+                                   [[NSRunLoop currentRunLoop]
+                                       runUntilDate:[NSDate dateWithTimeIntervalSinceNow:seconds]];
+                                 }
+                               }];
+#endif
+}
+
+#if !TARGET_OS_WATCH
 - (GDTCORBackgroundIdentifier)beginBackgroundTaskWithName:(NSString *)name
```
mikehaney24: Ok, so. I think I have an idea of how to keep this same API and have it function on watchOS. Essentially, we create a dispatch_semaphore and store it in a class-wide array mapping a generated ID (the next index) to the semaphore. We dispatch_wait in performExpiringActivity block, which should be called on a global queue so as not to block the current queue. If the block is called with expiring set to YES, we dispatch_signal the semaphore, effectively not waiting for the previous background task to finish. Otherwise, in endBackgroundTask, we find the semaphore we created, signal it, and that will effectively end the performExpiringActivity block. WDYT?

PR #5029, opened 2020-03-06
Use static frameworks to build zip

paulb777:
Take advantage of new CocoaPods 1.9.0 functionality to use static frameworks instead of static libs to build zip.  It now can build Swift pods, but still more to figure out for managing bridging headers.

The static library `pod install` is still necessary for calculating the module map dependencies.

Also some additional minor improvements:
- Use dashes instead of colons in temp file name so command line completion works
- Only adjust signing for building Catalyst library
- Stop assuming that arm64 is always built
- Support xcframeworks that don't have all possible slices built
- Handle umbrella header specification for non-Firebase pods
- Delete no longer necessary code for copying symbolic linked header directories - because Headers are not links in the framework build.

Future tasks:
- Figure out how to access binary Swift frameworks from the zip
- More robust handling of module maps specified by podspecs
- Include PrivateHeaders directories in zip archive

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -121,11 +121,14 @@ enum CocoaPodUtils {
   ///   - pods: List of VersionedPods to install
   ///   - directory: Destination directory for the pods.
   ///   - customSpecRepos: Additional spec repos to check for installation.
+  ///   - forceStaticLibs: Force a static library pod install. For the module map construction, we want pod names not module
+  ///                      names in the generated OTHER_LD_FLAGS options.
   /// - Returns: A dictionary of PodInfo's keyed by the pod name.
   @discardableResult
   static func installPods(_ pods: [VersionedPod],
                           inDir directory: URL,
-                          customSpecRepos: [URL]? = nil) -> [String: PodInfo] {
+                          customSpecRepos: [URL]?,
+                          forceStaticLibs: Bool = false) -> [String: PodInfo] {
     let fileManager = FileManager.default
     // Ensure the directory exists, otherwise we can't install all subspecs.
     guard fileManager.directoryExists(at: directory) else {
@@ -140,7 +143,10 @@ enum CocoaPodUtils {
 
     // Attempt to write the Podfile to disk.
     do {
-      try writePodfile(for: pods, toDirectory: directory, customSpecRepos: customSpecRepos)
+      try writePodfile(for: pods,
+                       toDirectory: directory,
+                       customSpecRepos: customSpecRepos,
+                       forceStaticLibs: forceStaticLibs)
     } catch let FileManager.FileError.directoryNotFound(path) {
       fatalError("Failed to write Podfile with pods \(pods) at path \(path)")
     } catch let FileManager.FileError.writeToFileFailed(path, error) {
@@ -150,7 +156,8 @@ enum CocoaPodUtils {
     }
 
     // Run pod install on the directory that contains the Podfile and blank Xcode project.
-    let result = Shell.executeCommandFromScript("pod _1.8.4_ install", workingDir: directory)
+    // At least 1.9.0 is required for `use_frameworks! :linkage => :static`
+    let result = Shell.executeCommandFromScript("pod _1.9.0_ install", workingDir: directory)
     switch result {
     case let .error(code, output):
       fatalError("""
@@ -354,7 +361,8 @@ enum CocoaPodUtils {
   /// Create the contents of a Podfile for an array of subspecs. This assumes the array of subspecs
   /// is not empty.
   private static func generatePodfile(for pods: [VersionedPod],
-                                      customSpecsRepos: [URL]? = nil) -> String {
+                                      customSpecsRepos: [URL]?,
+                                      forceStaticLibs: Bool) -> String {
     // Start assembling the Podfile.
     var podfile: String = ""
 
@@ -369,8 +377,12 @@ enum CocoaPodUtils {
       """ // Explicit newline above to ensure it's included in the String.
     }
 
-    if LaunchArgs.shared.dynamic {
+    if forceStaticLibs {
+      podfile += "  use_modular_headers!\n"
+    } else if LaunchArgs.shared.dynamic {
       podfile += "  use_frameworks!\n"
+    } else {
+      podfile += "  use_frameworks! :linkage => :static\n"
     }
     // Include the minimum iOS version.
     podfile += """
@@ -421,7 +433,8 @@ enum CocoaPodUtils {
   /// "Podfile".
   private static func writePodfile(for pods: [VersionedPod],
                                    toDirectory directory: URL,
-                                   customSpecRepos: [URL]?) throws {
+                                   customSpecRepos: [URL]?,
+                                   forceStaticLibs: Bool) throws {
     guard FileManager.default.directoryExists(at: directory) else {
       // Throw an error so the caller can provide a better error message.
       throw FileManager.FileError.directoryNotFound(path: directory.path)
@@ -429,7 +442,7 @@ enum CocoaPodUtils {
 
     // Generate the full path of the Podfile and attempt to write it to disk.
     let path = directory.appendingPathComponent("Podfile")
-    let podfile = generatePodfile(for: pods, customSpecsRepos: customSpecRepos)
+    let podfile = generatePodfile(for: pods, customSpecsRepos: customSpecRepos, forceStaticLibs: forceStaticLibs)
     do {
       try podfile.write(toFile: path.path, atomically: true, encoding: .utf8)
     } catch {
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -61,14 +61,13 @@ enum CocoaPodUtils {
     let subspecs: Set<String>
 
     /// The contents of the module map for all frameworks associated with the pod.
-    var moduleMapContents: String
+    var moduleMapContents: ModuleMapBuilder.ModuleMapContents?
 
     init(version: String, dependencies: [String], installedLocation: URL, subspecs: Set<String>) {
       self.version = version
       self.dependencies = dependencies
       self.installedLocation = installedLocation
       self.subspecs = subspecs
-      moduleMapContents = ""
 
       // Get all the frameworks contained in this directory.
       var binaryFrameworks: [URL] = []
--- a/ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
@@ -106,7 +106,8 @@ extension FileManager {
     if #available(OSX 10.12, *) {
       let formatter = ISO8601DateFormatter()
       formatter.formatOptions.insert(.withInternetDateTime)
-      return formatter.string(from: Date())
+      // Replace ":" with "-" so that bash file completion still works.
+      return formatter.string(from: Date()).replacingOccurrences(of: ":", with: "-")
     } else {
       return UUID().uuidString
     }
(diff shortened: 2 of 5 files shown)


ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
```diff
@@ -106,7 +106,8 @@ extension FileManager {
     if #available(OSX 10.12, *) {
       let formatter = ISO8601DateFormatter()
       formatter.formatOptions.insert(.withInternetDateTime)
-      return formatter.string(from: Date())
+      // Replace ":" with "-" so that bash file completion still works.
+      return formatter.string(from: Date()).replacingOccurrences(of: ":", with: "-")
```
maksymmalyhin: Would an explicit custom date format be more readable?

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -260,10 +262,20 @@ struct FrameworkBuilder {
...
+        let files = try FileManager.default.contentsOfDirectory(at: frameworkPath,
+                                                                includingPropertiesForKeys: nil).compactMap { $0.absoluteString }
+        let frameworkDir = files.filter { $0.contains(".framework") }
+        actualFramework = URL(fileURLWithPath: frameworkDir[0]).lastPathComponent
+      } catch {
+        fatalError("Error while enumerating files \(frameworkPath): \(error.localizedDescription)")
+      }
+      var libPath = frameworkPath.appendingPathComponent(actualFramework)
+      if !LaunchArgs.shared.dynamic {
+        libPath = libPath.appendingPathComponent(actualFramework.replacingOccurrences(of: ".framework", with: ""))
```
maksymmalyhin: Some brief explanation why this (and above) path transformation works will help us in the future. Maybe a add a dir structure in the comments or smth.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -406,32 +418,68 @@ struct FrameworkBuilder {
...
+    // TODO: copy PrivateHeaders directory as well if it exists. SDWebImage is an example pod.
+
     // Move all the Resources into .bundle directories in the destination Resources dir. The
     // Resources live are contained within the folder structure:
     // `projectDir/arch/Release-platform/FrameworkName`
-    let arch = Architecture.arm64
-    let contentsDir = projectDir.appendingPathComponents([arch.rawValue,
-                                                          "Release-\(arch.platform.rawValue)",
+
+    let contentsDir = projectDir.appendingPathComponents([anyArch.rawValue,
```
maksymmalyhin: Nit: I would suggest not to use `rawValue` directly but add a method with self-explanatory name to the enum to return it. It will improve readability and simplify modification in the future. 

It's probably outside of the scope of the PR. I just would like to bring this topic up to consider in the future.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -444,7 +492,9 @@ struct FrameworkBuilder {
     let xcframework = packageXCFramework(withName: framework,
                                          fromFolder: frameworkDir,
                                          thinArchives: thinArchives,
-                                         moduleMapContents: moduleMapContents)
+                                         moduleMapContents: moduleMapContents.replacingOccurrences(
+                                           of: "INSERT_UMBRELLA_HEADER_HERE", with: umbrellaHeader
```
maksymmalyhin: Should we declare a constant for "INSERT_UMBRELLA_HEADER_HERE"?

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -444,7 +492,9 @@ struct FrameworkBuilder {
     let xcframework = packageXCFramework(withName: framework,
                                          fromFolder: frameworkDir,
                                          thinArchives: thinArchives,
-                                         moduleMapContents: moduleMapContents)
+                                         moduleMapContents: moduleMapContents.replacingOccurrences(
```
maksymmalyhin: Nit: Will moving `moduleMapContents` calculation to a separate method improve readability?

ZipBuilder/Sources/ZipBuilder/FileManager+Utils.swift
paulb777: Current version looks like:

```
$ ls -lt | head
total 12536
drwxr-xr-x  12 paulbeusterien  primarygroup      384 Mar  5 15:21 2020-03-05T23-17-50Z
-rw-r--r--   1 paulbeusterien  primarygroup  6416981 Mar  5 15:21 Frameworks.zip
drwxr-xr-x  11 paulbeusterien  primarygroup      352 Mar  5 14:21 2020-03-05T22-19-43Z
drwxr-xr-x   5 paulbeusterien  primarygroup      160 Mar  5 14:00 2020-03-05T21-59-56Z
drwxr-xr-x   5 paulbeusterien  primarygroup      160 Mar  5 13:52 2020-03-05T21-50-17Z
drwxr-xr-x  12 paulbeusterien  primarygroup      384 Mar  5 13:42 2020-03-05T21-39-47Z
drwxr-xr-x   5 paulbeusterien  primarygroup      160 Mar  5 09:54 2020-03-05T17:54:33Z
drwxr-xr-x   5 paulbeusterien  primarygroup      160 Mar  5 09:46 2020-03-05T17-46-47Z
drwxr-xr-x   5 paulbeusterien  primarygroup      160 Mar  5 09:36 2020-03-05T17-36-15Z
```
It sortable, readable, and works in bash completion. I'm find with this or any other format that meets those three requirements.

maksymmalyhin: I mean instead of replacing symbols in the formatted string, if it would be better to use something like this:
```
let formatter = DateFormatter
formatter.dateFormat = "YYYY-MM-dd'T'HH-mm-ss"
...
```

paulb777: I see. Much cleaner. Thanks!

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -61,14 +61,13 @@ enum CocoaPodUtils {
     let subspecs: Set<String>
 
     /// The contents of the module map for all frameworks associated with the pod.
-    var moduleMapContents: String
+    var moduleMapContents: ModuleMapBuilder.ModuleMapContents?
```
maksymmalyhin: Why changed to Optional? (sorry if I'm missing it)

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -444,7 +493,8 @@ struct FrameworkBuilder {
     let xcframework = packageXCFramework(withName: framework,
                                          fromFolder: frameworkDir,
                                          thinArchives: thinArchives,
-                                         moduleMapContents: moduleMapContents)
+                                         moduleMapContents:
+                                         podInfo.moduleMapContents!.get(umbrellaHeader: umbrellaHeader))
```
maksymmalyhin: Are we 100% sure `moduleMapContents` exists at this point? If yes, may it be non-optional? It's better to avoid force unwrap unless it is necessary.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
paulb777: It was initialized to empty string before - that's no longer possible since the type is no longer String.  It's an improvement to have an uninitialized struct versus a wrong empty string.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -406,32 +419,68 @@ struct FrameworkBuilder {
...
+    } else {
+      var umbrellaHeaderURL: URL
+      do {
+        let files = try fileManager.contentsOfDirectory(at: headersDir,
+                                                        includingPropertiesForKeys: nil).compactMap { $0.absoluteString }
+        let umbrellas = files.filter { $0.contains("umbrella.h") }
+        if umbrellas.count != 1 {
+          fatalError("Did not find exactly one umbrella header in \(headersDir).")
+        }
+        umbrellaHeaderURL = URL(string: umbrellas[0])!
```
ryanwilson: Nit: this could be cleaned up a little bit by using `guard` and `guard let` (I typically try to use `guard` whenever some sort of validation is going on. Also using `first` since it's optional.

```
guard let firstUmbrella = umbrellas.first, 
    let foundHeader = URL(string: firstUmbrella) else { /* error */ }

umbrellaHeaderURL = foundHeader
```

You may even be able to combine them if `URL(string:)` takes an optional String but I don't think it does.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
paulb777: Yes. And if it's not there, we should crash since we'll otherwise generate a bad framework.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -406,32 +419,68 @@ struct FrameworkBuilder {
...
+        umbrellaHeader = umbrellaHeaderURL.lastPathComponent
       }
     }
-    // Copy the Headers over. Pass in the prefix to remove in order to generate the relative paths
-    // for some frameworks that have nested folders in their public headers.
+    // Copy the Headers over.
     let headersDestination = frameworkDir.appendingPathComponent("Headers")
     do {
-      try recursivelyCopyHeaders(from: headersDir, to: headersDestination)
+      try fileManager.copyItem(at: headersDir, to: headersDestination)
```
ryanwilson: Why did this change from recurisve?

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
maksymmalyhin: A crash from force unwrap is hardly readable. I would suggest adding a `guard` statement with a fatal error in `else` clause with an informative message.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
paulb777: We probably need a bigger refactor since the initialization of moduleMapContents is still ugly - but that's outside the scope of current PR

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
paulb777: Done. Thanks!

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
paulb777: `copyItem` is recursive.  There are no longer any symbolic links in the framework build, so the Swift provided method now meets the requirements.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
paulb777: Fixed.

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -562,6 +619,9 @@ struct FrameworkBuilder {
     for platform in Architecture.TargetPlatform.allCases {
       // Get all the slices that belong to the specific platform in order to lipo them together.
       let slices = thinArchives.filter { $0.key.platform == platform }
+      if slices.count == 0 {
```
maksymmalyhin: Very minor in this context. In general, it is recommended to use [isEmpty](https://developer.apple.com/documentation/swift/array/1688398-isempty) method instead of `count == 0` or similar for collections because `count` may be non-constant time for some collection implementations.

paulb777: Thanks - I'll fix for next PR

PR #5033, opened 2020-03-06
Crashlytics pin on newer version of GDT

samedson:
Need the change that allows for larger logs: https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleDataTransportCCTSupport/CHANGELOG.md#v141

### DIFF
--- a/FirebaseCrashlytics.podspec
+++ b/FirebaseCrashlytics.podspec
@@ -45,8 +45,8 @@ Pod::Spec.new do |s|
   s.dependency 'FirebaseInstallations', '~> 1.1'
   s.dependency 'FirebaseAnalyticsInterop', '~> 1.2'
   s.dependency 'PromisesObjC', '~> 1.2'
-  s.dependency 'GoogleDataTransport', '~> 4.0'
-  s.dependency 'GoogleDataTransportCCTSupport', '~> 1.3'
+  s.dependency 'GoogleDataTransport', '~> 4.0', '>= 4.0.1'
+  s.dependency 'GoogleDataTransportCCTSupport', '~> 1.4', '>= 1.4.1'
   s.dependency 'nanopb', '~> 0.3.901'
 
   s.libraries = 'c++', 'z'


FirebaseCrashlytics.podspec
```diff
@@ -45,8 +45,8 @@ Pod::Spec.new do |s|
   s.dependency 'FirebaseInstallations', '~> 1.1'
   s.dependency 'FirebaseAnalyticsInterop', '~> 1.2'
   s.dependency 'PromisesObjC', '~> 1.2'
-  s.dependency 'GoogleDataTransport', '~> 4.0'
-  s.dependency 'GoogleDataTransportCCTSupport', '~> 1.3'
+  s.dependency 'GoogleDataTransport', '~> 4.0', '>= 4.0.1'
```
mikehaney24: If you can hold off on submitting this, probably pin to the soon-to-be-released GDT 5.0 and GDTCCT 2.0 versions.

paulb777: Ideally internal testing should depend on GitHub latest.

paulb777: Since Firebase by default enables minor version updates of other pods in the repo - if a major version update is made by a low level pod, that pod should probably also manage the migration of its dependents so that the repo stays consistent at latest versions. Does that work for you @mikehaney24? cc: @ryanwilson @maksymmalyhin

mikehaney24: I don't think it's reasonable in the general case for a low-level library to manage the version dependencies of its adopters (what if they pinned to a lower version on purpose?) In this case, I can take care of it going forward, but only for Firebase OSS SDKs.

paulb777: Sorry, I should have been clear that I just meant dependencies in this repo.  Does that work?

mikehaney24: Yeah, maybe there's a way to make sure we always change it Firebase-wide by having a ruby config file with GDT versions that's loaded by the podspecs? For now, I can do a find-and-replace after major version changes.

samedson: For this change we'll hold off for V5 / V2

samedson:
Been updated already
https://github.com/firebase/firebase-ios-sdk/blame/master/FirebaseCrashlytics.podspec#L48-L49

PR #5040, opened 2020-03-07
Add Swift module support to ZipBuilder

paulb777:
Build Swift modules for zip distributions.  Swift libraries need to include a .swiftmodule directory in the Modules directory of the framework.  This PR sets it up:

Screen Shot 2020-03-07 at 8 29 45 AM

The resulting xcframeworks now can be dragged and dropped into the Firestore quickstart and it runs correctly for iOS and Catalyst.

#no-changelog

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
+++ b/ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
@@ -221,7 +221,7 @@ struct FrameworkBuilder {
                 "ARCHS=\(cleanArch)",
                 "VALID_ARCHS=\(cleanArch)",
                 "ONLY_ACTIVE_ARCH=NO",
-                // BUILD_LIBRARY_FOR_DISTRIBUTION=YES is Necessary for Swift libraries.
+                // BUILD_LIBRARY_FOR_DISTRIBUTION=YES is necessary for Swift libraries.
                 // See https://forums.developer.apple.com/thread/125646.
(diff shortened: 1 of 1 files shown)


paulb777:
Firestore quickstart running on Catalyst from the xcframework zip
Screen Shot 2020-03-07 at 9 21 28 AM

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -571,19 +574,76 @@ struct FrameworkBuilder {
...
-    let moduleDir = destination.appendingPathComponent("Modules")
-    do {
-      try FileManager.default.createDirectory(at: moduleDir, withIntermediateDirectories: true)
-    } catch {
-      fatalError("Could not create Modules directory for framework: \(framework). \(error)")
+
+    // For Swift modules, we use the modulemap from the xcodebuild. For Objective C modules,
+    // We use the constructed module map that includes required framework and library
+    // dependencies.
+    if !makeSwiftModuleMap(thinArchives: thinArchives, destination: destination) {
```
ryanwilson: Can you please break this out into separate lines with an appropriately named variable for the return statement? It's a little tough to read right now, reading it as `if don't make Swift module map` but the function itself has side-effects. Maybe something like:

```
let builtSwiftModules = makeSwiftModuleMap(thinArchives: thinArchives, destination: destination)
if !builtSwiftModules {
  // Contents of block...
}
```

paulb777:
I flushed out the comments and cleaned up the boolean logic.

PR #5041, opened 2020-03-07
Fix nightly zip GHA build

paulb777:
Fix nightly GHA build that started failing after repo switched to CocoaPods 1.9.0.

Remove the specification from the zip builder so it uses the CocoaPods installed by scripts/setup_bundler.sh.

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -115,7 +115,7 @@ enum CocoaPodUtils {
     }
 
     // Run pod install on the directory that contains the Podfile and blank Xcode project.
-    let result = Shell.executeCommandFromScript("pod _1.8.4_ install", workingDir: directory)
+    let result = Shell.executeCommandFromScript("pod install", workingDir: directory)
     switch result {
     case let .error(code, output):
       fatalError("""
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -116,7 +116,7 @@ enum CocoaPodUtils {
 
     // Run pod install on the directory that contains the Podfile and blank Xcode project.
     checkCocoaPodsVersion(directory: directory)
-    let result = Shell.executeCommandFromScript("pod install", workingDir: directory)
+    let result = Shell.executeCommandFromScript("bundle exec pod install", workingDir: directory)
     switch result {
     case let .error(code, output):
       fatalError("""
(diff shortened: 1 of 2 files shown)


ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -115,7 +115,8 @@ enum CocoaPodUtils {
     }
 
     // Run pod install on the directory that contains the Podfile and blank Xcode project.
-    let result = Shell.executeCommandFromScript("pod _1.8.4_ install", workingDir: directory)
+    checkCocoaPodsVersion(directory: directory)
+    let result = Shell.executeCommandFromScript("pod install", workingDir: directory)
```
maksymmalyhin: AFAIK just `pod install` is not necessarily uses Bundler version (it depends on other Ruby environment settings). `bundle exec pod install` should be more environment-independent version.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -381,4 +382,34 @@ enum CocoaPodUtils {
...
     }
   }
+
+  private static var checkedCocoaPodsVersion = false
+
+  /// At least 1.9.0 is required for `use_frameworks! :linkage => :static`
+  /// - Parameters:
+  ///   - directory: Destination directory for the pods.
+  private static func checkCocoaPodsVersion(directory: URL) {
+    if checkedCocoaPodsVersion {
```
maksymmalyhin: It doesn't look exactly correct, because the actual result may depend on `directory` in case if Bundler is used.

ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -381,4 +382,34 @@ enum CocoaPodUtils {
...
+      return
+    }
+    checkedCocoaPodsVersion = true
+    let podVersion = Shell.executeCommandFromScript("pod --version", workingDir: directory)
```
maksymmalyhin: `bundle exec pod --version`?

ryanwilson:
Do we want to allow updates to CocoaPods without us specifying it explicitly? I realize that having to update it twice is annoying but we could run into issues with module map generation or something else (regressions in CocoaPods) similar to what we saw in going from 1.5.3 to higher versions.

paulb777:
As it's currently set up, the ZipBuilder doesn't have access to the repo bundle install since it runs `pod install` in the temporary project directory.

The nightly build broke because GHA updated its default CocoaPods image to 1.9.0.

The build should now be more robust to future CocoaPods version updates.  If we'd still like to verify might be better to catch any issues with better testing versus forcing an old version in the zip builder - like adding automated diff testing for generated module maps.

I don't think there's a short solution to switch to the bundler or locking the CocoaPods version, so I'd prefer to merge this to fix the nightly build and open an issue if you think we need to do either of those.

maksymmalyhin:
As an option, we may copy `Gemfile` and `Gemfile.lock` from the main repo to the temporary project dir if we want Bundler working for it.

paulb777:
Agreed that copying Gemfile* is an option.  Are you ok to defer?

PR #5051, opened 2020-03-10
Refactor and clean up the CMake build ahead of CMake/iOS support

wilhuff:
This is a grab bag of minor fixes that I undertook while adding iOS support to the CMake build.

Individual commits describe the changes, but most notably, this:

  * Renames functions and variables to avoid conflicts in the global namespace. A corresponding change has been made to `firebase-cpp-sdk`. Unfortunately our definitions of `cc_test` had drifted and keeping them unified is harder than it's worth.
  * Renames the test host to be platform agnostic; a subsequent PR will add an iOS definition.
  * Simplifies `compiler_setup.cmake`. Prior to #4984, this script modified `CMAKE_CXX_FLAGS` so it was important to avoid imposing our preferences on projects we included. However, this also changed projects that included firebase-ios-sdk (notably, firebase-cpp-sdk) so that PR changed our rules to add `FIREBASE_CXX_FLAGS` to targets as appropriate. Once we did that though, there were no side-effects of `compiler_setup` so this change combines `compiler_id`, `archive_options`, and `compiler_setup` into the one file they really should be and moves it to the top.


Firestore/Example/App/CMakeLists.txt
```diff
@@ -16,35 +16,27 @@ if(NOT APPLE)
...
-  ${sources}
+  PRIVATE ${OBJC_FLAGS}
 )
 
 target_link_libraries(
   firestore_example_app_macos
-  PRIVATE "-framework AppKit"
+  PRIVATE
+  "-framework AppKit"
+  "-framework Foundation"
```
var-const: How did it work before, without specifying the dependency on `Foundation`?

Firestore/Example/App/CMakeLists.txt
```diff
@@ -17,26 +17,26 @@ if(NOT APPLE)
 endif()
 
 add_executable(
-  firestore_example_app_macos MACOSX_BUNDLE
+  firebase_firestore_example_app MACOSX_BUNDLE
```
var-const: Question: what is the rationale behind removing `macos` from the name?

cmake/compiler_setup.cmake
```diff
@@ -123,3 +129,18 @@ endforeach()
...
   list(APPEND FIREBASE_IOS_CXX_FLAGS ${flag})
 endforeach()
+
+if(APPLE)
+  # When building on Apple platforms, ranlib complains about "file has no
+  # symbols". Unfortunately, most of our dependencies implement their
+  # cross-platform build with preprocessor symbols so translation units that
+  # don't target the current platform end up empty (and trigger this warning).
+  #
+  # Note this can't be set in compiler_setup.cmake because that is included
```
var-const: This part of the comment is outdated.

Firestore/Example/App/CMakeLists.txt
wilhuff: It got the dependency implicitly because `AppKit` depends on `Foundation` too.

Firestore/Example/App/CMakeLists.txt
wilhuff: The idea is that our `objc_test` invocations will want to work on all supported platforms but they need to specify a host application in which to run. If I left the platform names in the app names, every `objc_test` would need a block like this above it:

```cmake
if(APPLE)
  if(IOS)
    set(host_app firestore_example_app_ios)
  else()
    set(host_app firestore_example_app_macos)
  endif()

  objc_test(
    firebase_firestore_objc_integration_test
    HOST ${host_app}
    ...
  )
endif()
```

Given that we can't actually build for macOS and iOS simultaneously changing this to a neutral name means this becomes:

```cmake
if(APPLE)
  objc_test(
    firebase_firestore_objc_integration_test
    HOST firebase_firestore_example_app
    ...
  )
endif()
```

cmake/compiler_setup.cmake
wilhuff: Fixed.

wilhuff:
All CMake-based builds are on GHA, so the Travis failures are spurious.

PR #5042, opened 2020-03-07
Fix compilation under Linux/GCC 9.2.1 in Release mode

wilhuff:
The `firebase_firestore_protos_libprotobuf` library contains protoc-generated code. Compile with standard flags rather than Firestore's default of `-Wall -Wextra -Werror`.

This avoids errors like these:

```
Firestore/Protos/cpp/google/firestore/v1/firestore.pb.cc:1669:11: error: ‘void* memset(void*, int, size_t)’ offset [57, 61] from the object at ‘google::firestore::v1::_ListDocumentsRequest_default_instance_’ is out of the bounds of referenced subobject ‘google::firestore::v1::ListDocumentsRequest::mask_’ with type ‘google::firestore::v1::DocumentMask*’ at offset 48 [-Werror=array-bounds]
 1669 |   ::memset(&mask_, 0, static_cast<size_t>(
      |   ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1670 |       reinterpret_cast<char*>(&show_missing_) -
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1671 |       reinterpret_cast<char*>(&mask_)) + sizeof(show_missing_));
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #5052, opened 2020-03-10
Add iOS build support to the CMake build

wilhuff:
Individual commits more-or-less say what they're doing.

At a high level, this change:

  * Allows a user to disable building tests and benchmarks
  * Fixes iOS-specific problems that arise from cross-compilation difficulty with CMake
  * Undoes the special case in #5042, adding a general `DISABLE_STRICT_WARNINGS` flag and removing the `-Wno-foo` blocks that were subtracting from the full warnings battery. This is ultimately a better choice for effectively third-party code.

Note:
  * this is diffbased on #5051;
  * I'll add a separate PR to add CI for this


wilhuff:
Diffbase removed.

CMakeLists.txt
```diff
@@ -98,19 +104,32 @@ download_external_sources()
...
-set(BENCHMARK_ENABLE_INSTALL OFF CACHE BOOL "Firestore disabled")
-set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "Firestore disabled")
-add_external_subdirectory(benchmark)
+if(FIREBASE_IOS_BUILD_BENCHMARKS)
+  set(BENCHMARK_ENABLE_TESTING OFF CACHE BOOL "Firestore disabled")
+  set(BENCHMARK_ENABLE_EXCEPTIONS OFF CACHE BOOL "Firestore disabled")
+  set(BENCHMARK_ENABLE_INSTALL OFF CACHE BOOL "Firestore disabled")
+  set(BENCHMARK_ENABLE_GTEST_TESTS OFF CACHE BOOL "Firestore disabled")
+
+  if(IOS)
```
a-maurice: FYI, IOS is only a standard CMake variable as of version 3.14.  For the C++ repo, we had to explicitly define IOS via the toolchain file to get around this (https://github.com/firebase/firebase-cpp-sdk/blob/master/cmake/toolchains/ios.cmake#L44), though will probably work on changing that to use a more explicit TARGET_PLATFORM_IOS, or something similar.  You might want to do something similar as well, to support older cmakes.

cmake/cc_rules.cmake
```diff
@@ -225,33 +233,59 @@ function(firebase_ios_maybe_remove_objc_sources output_var)
...
-# Adds FIREBASE_IOS_OBJC_FLAGS to the compile options of the given target if
-# any of the sources have filenames that indicate they are Objective-C.
-function(firebase_ios_add_objc_flags target)
-  set(_has_objc OFF)
+# If any of the sources have filenames that indicate they are Objective-C adds
+# Either FIREBASE_IOS_OBJC_FLAGS or FIREBASE_IOS_OBJC_FLAGS_STRICT depending on
+# whether or not DISABLE_STRICT_WARNINGS was passed.
+function(firebase_ios_add_compile_options target)
+  set(flag DISABLE_STRICT_WARNINGS)
+  cmake_parse_arguments(aco "${flag}" "" "${}" ${ARGN})
```
a-maurice: Is there a reason for the empty brackets?

CMakeLists.txt
wilhuff: The "standard" ios toolchain file also sets `IOS`: https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake#L466

Coding to modern CMake and adding fallback support for older CMake versions seems like a reasonable path forward. If/when `firebase-cpp-sdk` chooses to implement some other convention, we can revisit here too.

cmake/cc_rules.cmake
wilhuff: No. This is a mistake (though I'm surprised it works). Fixed.

wilhuff:
Travis failures are unrelated.

PR #5061, opened 2020-03-10
Add fuzz build to continuous integration checks

var-const:
#no-changelog


cmake/fuzzing_options.cmake
```diff
@@ -41,7 +41,9 @@ if(FUZZING AND NOT DEFINED ENV{LIB_FUZZING_ENGINE})
     # have libFuzzer by default.
     set(fuzzing_flags -fsanitize-coverage=trace-pc-guard)
   elseif(CXX_GNU)
-    set(fuzzing_flags -fsanitize-coverage=trace-pc)
+    # TODO(varconst): restore this flag. While this flag appears to be valid,
+    # setting it causes build to fail (at least when using GitHub Actions).
+    #set(fuzzing_flags -fsanitize-coverage=trace-pc)
```
var-const: This took a while to debug. Unfortunately, the [error messages](https://github.com/firebase/firebase-ios-sdk/pull/5061/checks?check_run_id=501673727) were absolutely unhelpful; I'm still not clear why exactly this flag breaks the build.

var-const:
Any suggestions on how to better configure this are very welcome.

var-const:
Example of a failed check: [link](https://github.com/firebase/firebase-ios-sdk/pull/5061/checks?check_run_id=501814927).
Example of a fix: [link](https://github.com/firebase/firebase-ios-sdk/pull/5061/checks?check_run_id=501822467).

cmake/fuzzing_options.cmake
wilhuff: The failure there is Google Benchmark's feature detection code failing to compile and it ultimately errors the build because it can't figure out how to run regexes. You can see that all the checks are failing, e.g.

```
-- Performing Test HAVE_CXX_FLAG_STD_CXX11
-- Performing Test HAVE_CXX_FLAG_STD_CXX11 - Failed
```

The `ubuntu-latest` GHA machines run GCC 7.4.0 while our gLinux machines have GCC 9+. The [GCC 7.4.0 docs](https://gcc.gnu.org/onlinedocs/gcc-7.4.0/gcc/Instrumentation-Options.html) suggest that `-fsanitize-coverage=trace-pc` is supported.

I created a docker machine with GCC 7.2.0 on it (`docker run -it --rm conanio/gcc7 /bin/bash`) and the failure message from `CMakeFiles/CMakeError.txt` looks like this:

```
: && /usr/bin/g++  -fsanitize=address -fno-omit-frame-pointer -fsanitize-coverage=trace-pc -DHAVE_CXX_FLAG_STD_CXX11  -std=c++11  -fsanitize=address CMakeFiles/cmTC_5a0b7.dir/src.cxx.o  -o cmTC_5a0b7   && :
CMakeFiles/cmTC_5a0b7.dir/src.cxx.o: In function `main':
src.cxx:(.text+0xa): undefined reference to `__sanitizer_cov_trace_pc'
src.cxx:(.text+0x14): undefined reference to `__sanitizer_cov_trace_pc'
CMakeFiles/cmTC_5a0b7.dir/src.cxx.o: In function `_GLOBAL__sub_I_00099_0_main':
src.cxx:(.text+0x26): undefined reference to `__sanitizer_cov_trace_pc'
collect2: error: ld returned 1 exit status
```

You'd be able to see this better from GHA if we changed `build.sh`, line 358 to invoke `cmake` like so:

```bash
      rc=0
      cmake -G Ninja "${cmake_options[@]}" .. || rc=$?
      if [[ $rc -ne 0 ]]; then
        cat CMakeFiles/CMakeError.log
        exit $rc
      fi
```

From there I think we could resolve this issue, though I'm not entirely sure why asan is failing to link.

paulb777:
Closing for inactivity. Please reopen to continue.

PR #5064, opened 2020-03-11
Bump Auth pod version

renkelvin:


Firebase/Auth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v6.5.0
+- [feature] Added support of Multi-factor Authentication. (#4823)
```
ryanwilson: @morganchen12 should this be `feature` or `added`?

renkelvin: Seems we used both before. My understanding is we use `feature` for minor releases? I'll let Morgan to decide.

renkelvin: @ryanwilson Talked to Morgan offline. Actually tehy're the same and they both show up as feature on devsite.

PR #5078, opened 2020-03-11
Remove cleanup of experiments that are no longer running

christibbs:
We've made a product decision to not clean these up for now.

Impact: if a user stops an experiment via the console, all events will continue to be striped with the CUP for that experiment.

Also fixed the version number for the ABT dependency, which somehow got reverted. FIAM should update on minor ABT releases.

### DIFF
--- a/FirebaseInAppMessaging.podspec
+++ b/FirebaseInAppMessaging.podspec
@@ -45,7 +45,7 @@ See more product details at https://firebase.google.com/products/in-app-messagin
   s.ios.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
   s.dependency 'FirebaseInstanceID', '~> 4.0'
   s.dependency 'GoogleDataTransportCCTSupport', '~> 2.0'
-  s.dependency 'FirebaseABTesting', '~> 3.2'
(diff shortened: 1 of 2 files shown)


FirebaseInAppMessaging.podspec
```diff
@@ -45,7 +45,7 @@ See more product details at https://firebase.google.com/products/in-app-messagin
   s.ios.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
   s.dependency 'FirebaseInstanceID', '~> 4.0'
   s.dependency 'GoogleDataTransportCCTSupport', '~> 2.0'
-  s.dependency 'FirebaseABTesting', '~> 3.2'
+  s.dependency 'FirebaseABTesting', '~> 3.2.0'
```
paulb777: Please revert. Firebase should only allow other pods in this repo to update minor versions.

morganchen12: Is this change necessary?

christibbs: Reverted.

JasonAHeron:
Do you need to change any tests?

christibbs:
> Do you need to change any tests?

Nope, apparently that was a hole in the unit test coverage.

PR #5080, opened 2020-03-11
Zip Carthage refactor

### LINKED ISSUE #4917: FirebaseCoreDiagnostic.framework is a not a valid framework bundle.

  * Xcode version: 10.2
  * Firebase SDK version: 6.9.0
  * Firebase Component: FirebaseCoreDiagnostic
  * Component version: N/A

#### Steps to reproduce:

Binary distribution with carthage is not working. The framework bundle is missing it's Info.plist

#### Relevant Code:

Screenshot 2020-02-17 at 23 01 38

paulb777:
Update and refactor Carthage build for the xcframework changes.

Carthage does not yet support xcframeworks, so this should generate the same build as before for Carthage.

In the refactor, Carthage builds and adaptations are pushed lower in the process to avoid complicated adjustments in the higher level functions that had to reach down to make changes into low-level generated build artifacts.

Both zip and Carthage share the same compile phases, but then separately package to xcframeworks and frameworks. Carthage rebuilds FirebaseCoreDiagnostics with the right flags and skips including the Catalyst slice.

Fix #4917

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
@@ -106,11 +106,12 @@ extension CarthageUtils {
 
       // Parse the JSON file, ensure that we're not trying to overwrite a release.
       var jsonManifest = parseJSONFile(fromDir: jsonDir, product: product)
-      //  guard jsonManifest[firebaseVersion] == nil else {
-      //     print("Carthage release for \(product) \(firebaseVersion) already exists - skipping.")
-      // temporary disable
-      //      continue
-      //   }
+      if !args.carthageSkipVersionCheck {
+        guard jsonManifest[firebaseVersion] == nil else {
+          print("Carthage release for \(product) \(firebaseVersion) already exists - skipping.")
+              continue
+        }
+      }
 
       // Make updates to all frameworks to make Carthage happy. We don't worry about xcframeworks
       // here.
--- a/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
@@ -22,7 +22,6 @@ import Foundation
 enum CarthageUtils {}
 
 extension CarthageUtils {
-
   /// Package all required files for a Carthage release.
   ///
   /// - Parameters:
@@ -88,10 +87,10 @@ extension CarthageUtils {
   ///   - outputDir: The directory where all artifacts should be created.
 
   private static func generateCarthageRelease(fromPackagedDir packagedDir: URL,
-                                      templateDir: URL,
-                                      jsonDir: URL,
-                                      artifacts: ZipBuilder.ReleaseArtifacts,
-                                      outputDir: URL) {
+                                              templateDir: URL,
+                                              jsonDir: URL,
+                                              artifacts: ZipBuilder.ReleaseArtifacts,
+                                              outputDir: URL) {
     let directories: [String]
     do {
       directories = try FileManager.default.contentsOfDirectory(atPath: packagedDir.path)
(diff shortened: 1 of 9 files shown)


ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -190,6 +198,27 @@ struct FrameworkBuilder {
     return .success(output: fullOutput)
   }
 
+  /// Build all thin slices for an open source pod.
+  /// - Parameter framework: The name of the framework to be built.
+  /// - Parameter logsDir: The path to the directory to place build logs.
+  /// - Returns: An dictionary of URLs to the built thin libraries keyed by architiecture
```
ryanwilson:
```suggestion
  /// - Returns: An dictionary of URLs to the built thin libraries keyed by architecture.
```

ZipBuilder/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -313,10 +343,10 @@ struct FrameworkBuilder {
   /// - Parameter framework: The name of the framework to be built.
   /// - Parameter logsOutputDir: The path to the directory to place build logs.
   /// - Parameter moduleMapContents: Module map contents for all frameworks in this pod.
-  /// - Returns: A path to the newly compiled framework (with any included Resources embedded).
+  /// - Returns: A path to the newly compiled framework and the Carthage version if needed)
```
ryanwilson: Nit: period at the end

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -158,12 +155,17 @@ struct ZipBuilder {
...
     for (framework, paths) in frameworks {
       print("Frameworks for pod: \(framework) were compiled at \(paths)")
     }
-    return (installedPods, frameworks)
+
+    // Create the CoreDiagnostics framework for Carthage, since it needs to be recompiled.
+//    let carthageCoreDiagnostics = createCarthageCoreDiagnostics(fromPods: installedPods,
+//                                                                frameworks: frameworks,
+//                                                                inProjectDir: projectDir)
+    return (installedPods, frameworks, carthageFrameworks)
```
ryanwilson: Nit: can this be removed? Or what's this for here?

ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -22,6 +22,60 @@ import Foundation
...
+  ///   - carthageJSONDir: Location of directory containing all JSON Carthage manifests.
+  ///   - artifacts: Release Artifacts from build.
+  ///   - rcNumber: The RC number.
+  /// - Returns: The path to the root of the Carthage installation.
+  static func packageCarthageRelease(templateDir: URL,
+                                     carthageJSONDir: URL,
+                                     artifacts: ZipBuilder.ReleaseArtifacts,
+                                     rcNumber: Int?) -> URL? {
+    if let zipLocation = artifacts.carthageDir {
+      var carthageRoot: URL?
```
maksymmalyhin: Nit: to avoid declaring `carthageRoot` as `var` you may wrap calculation of the value into a closure like here:
```
let carthageRoot: URL? = {
  do {
      return try doSomething()
  } catch {
      return nil
  }
}()
```

ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -51,23 +106,17 @@ extension CarthageUtils {
...
+          continue
+        }
       }
 
-      // Carthage will fail to install a framework if it doesn't have an Info.plist, even though
-      // they're not used for static frameworks. Generate one and write it to each framework.
-      let frameworks = allContents.filter { $0.hasSuffix(".framework") }
+      // Make updates to all frameworks to make Carthage happy. We don't worry about xcframeworks
+      // here.
+      let allFiles = FileManager.default.enumerator(atPath: fullPath.path)?.allObjects as! [String]
```
maksymmalyhin: Force cast `as!` will crash with an unclear error even if `.enumerator(atPath: fullPath.path)` is `nil`. I think, we should use `guard` here.

PR #5090, opened 2020-03-12
Remove smart quotes

morganchen12:


Firebase/Database/Constants/FConstants.m
```diff
@@ -41,7 +41,7 @@
 NSString *const kFWPASyncServerListenCancelled = @"c";
 NSString *const kFWPAsyncServerSecurityDebug = @"sd";
 NSString *const kFWPAsyncServerDataUpdateBodyPath =
-    @"p"; // {“a”: “d”, “b”: {“p”: “/”, “d”: “<data>”}}
+    @"p"; // {"a": "d", "b": {"p": "/", "d": "<data>"}}
```
paulb777: @renkelvin - Are there missing tests for this one?

morganchen12: This is in database. @schmidt-sebastian

paulb777: Oh, I missed that. If it's not a recent change, I'm less worried - but still worth a review by @schmidt-sebastian

renkelvin:
Looks good from Auth side

Firebase/Database/Constants/FConstants.m
schmidt-sebastian: LGTM. Looks like this is in the comment only.

PR #5096, opened 2020-03-13
Migrate from using IID SDK for Remote Config to the new FIS SDK

dmandar:
The RC SDK now uses the FIS SDK to pass up the installationID and installation token in the body of the request. The token is also included in the header.


maksymmalyhin:
Both merge conflicts and tests should are expected to be fixed by #5109.

FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -192,18 +190,24 @@ - (void)fetchConfigWithExpirationDuration:(NSTimeInterval)expirationDuration
...
+    [strongSelf refreshInstallationsTokenWithCompletionHandler:completionHandler];
   });
 }
 
 #pragma mark - Fetch helpers
 
-/// Refresh instance ID token before fetching config. Instance ID is now mandatory for fetch
+- (NSString *)FIRAppNameFromFullyQualifiedNamespace {
+  NSString *FIRAppName =
+      [_FIRNamespace substringFromIndex:[_FIRNamespace rangeOfString:@":"].location + 1];
```
maksymmalyhin: I am a bit worried that we compose the string at one place and parse it at another, because it increases probability of errors when the logic needs to be updated. Will it be better to just pass `appName` separately? 

If you still would like to parse the string, I would suggest a safer way to do it to avoid crashing if the `_FIRNamespace` contains an unexpected string for some reason, e.g.:
```
return [[_FIRNamespace componentsSeparatedByString:@":"] lastObject];
```

FirebaseRemoteConfig/Sources/RCNFetch.m
```diff
@@ -234,20 +239,21 @@ - (void)refreshInstanceIDTokenAndFetchCheckInInfoWithCompletionHandler:
...
       RCNConfigFetch *strongSelf = weakSelf;
 
       // Dispatch to the RC serial queue to update settings on the queue.
       dispatch_async(strongSelf->_lockQueue, ^{
         RCNConfigFetch *strongSelfQueue = weakSelf;
 
         // Update config settings with the IID and token.
-        strongSelfQueue->_settings.configInstanceIDToken = [token copy];
-        strongSelfQueue->_settings.configInstanceID = identity;
+        strongSelfQueue->_settings.configInstallationsToken = [tokenResult.authToken copy];
```
maksymmalyhin: Nit: `copy` is not required here because `configInstallationsToken` property is defined with `copy` attribute already.

dmandar: Done.

FirebaseRemoteConfig/Sources/RCNFetch.m
dmandar: Done.

PR #5100, opened 2020-03-13
In banner view, pin top elements to the bottom of the status bar

### LINKED ISSUE #4714: [InAppMessaging] Banner message display overlaps status bar on iPad Pro 11"

According to @christibbs yup

  * Xcode version: 11.3.1
  * Firebase SDK version: 6.15.0
  * Firebase Component: InAppMessaging
 * Component version: 0.16.0, InAppMessagingDisplay is 0.15.5

Banner message overlaps with status bar text, looks awful. See photo:

72654704-b4211f00-395e-11ea-8d85-caa7e1464adb

#### Steps to reproduce:

Display a banner style message on iPad Pro 11" (2018 model, the only one I think), running iOS 13.3

christibbs:
- Removes storyboard constraint (at build time) that pins top level UI elements to the top layout guide
- Adds programmatic constraint that pins them to the bottom of the status bar

#fixes #4714 

After fix:
new

Before fix:
old

### DIFF
--- a/FirebaseInAppMessaging/Resources/FIRInAppMessageDisplayStoryboard.storyboard
+++ b/FirebaseInAppMessaging/Resources/FIRInAppMessageDisplayStoryboard.storyboard
@@ -95,7 +95,7 @@
                             </constraint>
                             <constraint firstItem="GOK-vx-mU8" firstAttribute="leading" secondItem="rzo-9i-rXZ" secondAttribute="leading" id="J6L-hu-jV6"/>
                             <constraint firstItem="VfB-vw-7up" firstAttribute="leading" secondItem="vRb-yf-OWE" secondAttribute="leading" constant="5" id="gGt-ku-pPu"/>
-                            <constraint firstItem="rzo-9i-rXZ" firstAttribute="top" secondItem="lHf-Ux-dEc" secondAttribute="bottom" constant="3" id="hhi-1K-2YO"/>
+                            <constraint firstItem="rzo-9i-rXZ" firstAttribute="top" secondItem="lHf-Ux-dEc" secondAttribute="bottom" constant="3" placeholder="YES" id="hhi-1K-2YO"/>
                             <constraint firstItem="VfB-vw-7up" firstAttribute="top" secondItem="rzo-9i-rXZ" secondAttribute="top" id="nF9-EZ-bjS"/>
                             <constraint firstAttribute="trailingMargin" secondItem="rzo-9i-rXZ" secondAttribute="trailing" id="pOl-wC-mqq"/>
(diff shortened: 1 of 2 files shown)


ryanwilson:
Hmm how does this look on devices with a notch? (just to confirm).

Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?

christibbs:
> Hmm how does this look on devices with a notch? (just to confirm).
> 
> Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?

Looks good on notch devices (notch is included in status bar height calculations). Will look into using `safeAreaLayoutGuide` since that may be cleaner code.

christibbs:
> Hmm how does this look on devices with a notch? (just to confirm).
> 
> Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?



> > Hmm how does this look on devices with a notch? (just to confirm).
> > Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?
> 
> Looks good on notch devices (notch is included in status bar height calculations). Will look into using `safeAreaLayoutGuide` since that may be cleaner code.

@ryanwilson I tried using constraints to `safeAreaLayoutGuide`, but it looks like iPad Pro 11" (counterintuitively) doesn't take the status bar into account-- so the same bug appears.

ryanwilson:
> > Hmm how does this look on devices with a notch? (just to confirm).
> > Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?
> 
> > > Hmm how does this look on devices with a notch? (just to confirm).
> > > Could you use `[safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide?language=objc)` instead or was there a reason you avoided it?
> > 
> > 
> > Looks good on notch devices (notch is included in status bar height calculations). Will look into using `safeAreaLayoutGuide` since that may be cleaner code.
> 
> @ryanwilson I tried using constraints to `safeAreaLayoutGuide`, but it looks like iPad Pro 11" (counterintuitively) doesn't take the status bar into account-- so the same bug appears.

Ahh shucks, strange. Sg, thanks for looking into it.

PR #5101, opened 2020-03-13
Default to use new report upload endpoint when settings were failed to be fetched

jasonhu-g:
Successful tests:
1. Unit tests
2. Verified when there's no internet, the crash report is enqueued to GDT


Crashlytics/Crashlytics/Models/FIRCLSSettings.m
```diff
@@ -299,6 +299,15 @@ - (BOOL)shouldUseNewReportEndpoint {
   return value.intValue == 2;
 }
 
++ (BOOL)shouldUseNewReportEndpointWithSettings:(nullable FIRCLSSettings *)settings {
```
samedson: Can this just be a method?

jasonhu-g:
This PR being replaced by: https://github.com/firebase/firebase-ios-sdk/pull/5128/

PR #5103, opened 2020-03-13
Update the zip distribution from frameworks to xcframeworks

paulb777:
Update the distribution of source pods in the zip distribution from frameworks to xcframeworks.  The result is that the community-supported Catalyst support is now available via the zip distribution in addition the CocoaPods.

Everything in this PR has already been reviewed and the plan is to merge this from the command line to avoid squashing the individual PRs on the xcframework-master branch.

The release notes will be updated after this merges to master.

#no-changelog


steipete:
Thanks for finishing! This is looking great!

PR #5115, opened 2020-03-16
Add Catalyst testing for GHA

paulb777:
This is a start at adding Catalyst to GHA CI.

Building and unit tests are enabled for GDT and Storage.
Building is enabled for Core and Messaging.

One test had to be disabled in GDT for Catalyst because it didn't seem like an easy port and it relies upon the NSKeyedArchiver which is going away.

A future task is to figure out how to run tests for testspecs that have `requires_app_host` enabled.

[FirebaseCore](https://github.com/firebase/firebase-ios-sdk/runs/512270812?check_suite_focus=true) and [FirebaseMessaging](https://github.com/firebase/firebase-ios-sdk/runs/512270895?check_suite_focus=true) fail with:

```
Testing failed:
	Unable to load standard library for target 'x86_64h-apple-ios13.1-macabi'
	Testing cancelled because the build failed.

** TEST FAILED **
```

Specifying only the Objective C test target instead of the App target scheme is a workaround locally, but it still fails on GHA with:
 
```
2020-03-17 14:32:47.672 xcodebuild[1564:26786] [MT] IDETestOperationsObserverDebug: 0.589 elapsed -- Testing started completed.
2020-03-17 14:32:47.672 xcodebuild[1564:26786] [MT] IDETestOperationsObserverDebug: 0.000 sec, +0.000 sec -- start
2020-03-17 14:32:47.673 xcodebuild[1564:26786] [MT] IDETestOperationsObserverDebug: 0.589 sec, +0.589 sec -- end
2020-03-17 14:32:47.673 xcodebuild[1564:26786] Error Domain=IDELaunchErrorDomain Code=9 "Launch error" UserInfo={NSLocalizedDescription=Launch error, NSLocalizedRecoverySuggestion=There is a problem launching using posix_spawn (error code: 86).}
Testing failed:
	FirebaseCore-Unit-unit:
		AppHost-FirebaseCore-Unit-Tests.app encountered an error (Failed to install or launch the test runner. (Underlying error: Launch error))

** TEST FAILED **

##[error]Process completed with exit code 65.
```

#no-changelog

### DIFF
--- a/scripts/test_catalyst.sh
+++ b/scripts/test_catalyst.sh
@@ -23,9 +23,9 @@
 # and test.
 
 # TODO - Determine why test specs that include `requires_app_host` fail to
-# launch tests. Locally, they will pass if the unit test scheme is specified.
-# However, on GHA, they fail to launch both from the test scheme and the app
-# scheme.
+# launch tests. Locally, they will pass if the only Objective C unit test scheme
+# is specified. However, on GHA, they fail to launch both from the test scheme
+# and the app scheme.
 
 set -xeuo pipefail
 pod="$1"
--- a/scripts/test_catalyst.sh
+++ b/scripts/test_catalyst.sh
@@ -17,8 +17,15 @@
 
 # USAGE: test_catalyst.sh pod build_mode [scheme]
 #
-# Builds and run tests for Catalyst since it's not yet supported by `pod lib lint`
-# The second argument should be "build" or "test". "test" indicates both build and test.
+# Builds and run tests for Catalyst since it's not yet supported by
+# `pod lib lint`.
+# The second argument should be "build" or "test". "test" indicates both build
+# and test.
+
+# TODO - Determine why test specs that include `requires_app_host` fail to
+# launch tests. Locally, they will pass if the unit test scheme is specified.
+# However, on GHA, the they fail to launch both from the test scheme and the
+# app scheme.
 
 set -xeuo pipefail
 pod="$1"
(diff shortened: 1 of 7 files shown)


scripts/test_catalyst.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 Google
```
wilhuff: Should be "Copyright 2020 Google LLC". See go/copyright.

scripts/test_catalyst.sh
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# USAGE: test_catalyst.sh pod
+#
+# Builds and run tests for Catalyst since it's not yet supported by `pod lib lint`
+# The second argument should be "build" or "test". "Test" indicates both build and test.
+
+set -x
```
wilhuff: It would be helpful to perform some argument validation in the script. As it stands if you just run this it's going to fail trying to find `.podspec`.

Also you should either explicitly check errors or opt into bash's `errexit` mode. You also probably want to enable `pipefail` mode so that the pipe to `xcpretty` does not mask an `xcodebuild` error. The easiest way to do both is to add this at the top:

```bash
set -euo pipefail
```

Finally, referring to `$1` and `$2` throughout isn't very readable. As part of your validation please pull out some variables like `pod` and `build_mode` or something like that.

scripts/test_catalyst.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# USAGE: test_catalyst.sh pod build_mode
+#
+# Builds and run tests for Catalyst since it's not yet supported by `pod lib lint`
+# The second argument should be "build" or "test". "test" indicates both build and test.
+
+pod="$1"
+build_mode="$2"
+
+set -xeuo pipefail
```
wilhuff: Note: if you put this above the assignments this will catch `$1` and `$2` being unset for you.

paulb777:
Thanks, now I need to figure out why the test launches succeeds locally, but fail on GHA

scripts/test_catalyst.sh
```diff
@@ -0,0 +1,37 @@
...
+build_mode="$2"
+
+if [[ $# -gt 2 ]]; then
+  scheme="$3"
+else
+  scheme="$pod"
+fi
+
+bundle exec pod gen --local-sources=./ --sources=https://cdn.cocoapods.org/ "$pod".podspec --platforms=ios
+xcodebuild $build_mode -configuration Debug -workspace "gen/$pod/$pod.xcworkspace"  -scheme "$scheme"\
```
wilhuff: Whitespace is a little off in here. There are several runs of two spaces and there's no space between `"$scheme"` and the trailing `\`.

Also, not that matters hugely, but go/shell-style calls for 80 character line lengths.

Finally, consider grouping flags (`-scheme`, `-sdk`, `-destination`) separately from build settings.

scripts/test_catalyst.sh
```diff
@@ -0,0 +1,37 @@
...
+ ARCHS=x86_64h VALID_ARCHS=x86_64h ONLY_ACTIVE_ARCH=NO  SUPPORTS_MACCATALYST=YES  -sdk macosx \
```
wilhuff: Is `x86_64h` a typo? If not, consider adding a comment on its origin.

scripts/test_catalyst.sh
paulb777: Reorganized. Thanks for the suggestions.

paulb777:
Going to merge the Catalyst GHA support here.   Please feel free to add additional review comments and I'll address them in a subsequent PR.

PR #5116, opened 2020-03-16
The uploader should retry when there's an error

mikehaney24:
Uploaders weren't always retrying in all the situations they should've been, so this hopefully addresses that. Also does a slight readability refactor.

### DIFF
--- a/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTUploader.m
+++ b/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTUploader.m
@@ -202,51 +202,39 @@ NSNotificationName const GDTCCTUploadCompleteNotification = @"com.GDTCCTUploader
         GDTCORLogWarning(GDTCORMCWUploadFailed, @"There was an error uploading events: %@", error);
       }
       NSError *decodingError;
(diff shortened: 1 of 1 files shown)


GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTUploader.m
```diff
@@ -202,51 +202,39 @@ - (void)uploadPackage:(GDTCORUploadPackage *)package {
...
-
-            default:
-              break;
-          }
         }
         pb_release(gdt_cct_LogResponse_fields, &logResponse);
       }
+      if (!futureUploadTime) {
+        GDTCORLogDebug("%@", @"CCT: The backend response failed to parse, so the next request "
+                             @"won't occur until 15 minutes from now");
```
morganchen12: drive-by comment: looks like `decodingError` is not checked or logged after calling `GDTCCTDecodeLogResponse`, maybe log it here?

mikehaney24: Done.

mikehaney24:
> Is a test feasible?

Not super feasible for the retry change, but the other changes are well tested.

PR #5120, opened 2020-03-16
Make the prioritizer save state between app launches

mikehaney24:
The prioritizer currently does not maintain state between app restarts. It's possible for GDTCORStorage to have references to stored events that the prioritizer doesn't know about--this (sort of) fixes that. Ultimately, the prioritizer asking storage for events is probably the only reasonable solution long-term.

### DIFF
--- a/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTPrioritizer.m
+++ b/GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTPrioritizer.m
@@ -25,6 +25,21 @@
 
 const static int64_t kMillisPerDay = 8.64e+7;
 
+/** Creates and/or returns a singleton NSString that is the NSCoding file location.
+ *
+ * @return The NSCoding file path.
+ */
+static NSString *ArchivePath() {
+  static NSString *archivePath;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    NSString *cachePath =
+        NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)[0];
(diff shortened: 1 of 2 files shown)


GoogleDataTransportCCTSupport/GDTCCTLibrary/GDTCCTPrioritizer.m
```diff
@@ -238,9 +257,127 @@ typedef NS_ENUM(NSInteger, GDTCCTQoSTier) {
...
+  dispatch_sync(_queue, ^{
+    if (@available(macOS 10.13, iOS 11.0, tvOS 11.0, *)) {
+      NSError *error;
+      NSData *data = [NSKeyedArchiver archivedDataWithRootObject:self
+                                           requiringSecureCoding:YES
+                                                           error:&error];
+      [data writeToFile:ArchivePath() atomically:YES];
+    } else {
+#if !TARGET_OS_MACCATALYST && !TARGET_OS_WATCH
+      [NSKeyedArchiver archiveRootObject:self toFile:ArchivePath()];
```
maksymmalyhin: Secure coding is supported by previous OS versions. You may refer to [GoogleUtilities](https://github.com/firebase/firebase-ios-sdk/blob/dd70b46370dab27fb9e19853d586ee888c2c33b5/GoogleUtilities/Environment/GULSecureCoding.m#L30) as an example.

mikehaney24: I'm not sure I understand--this is still using secure coding, it's just avoiding calling a deprecated API on some platforms.

maksymmalyhin: Sorry, nm, mixed it with NSKeyedUnarchiver

PR #5121, opened 2020-03-16
Revert "Address a few potential security issues (#5059)"

mikehaney24:
Reverts #5059 

There's a subtle bug (currently unknown) that's causing events to not be decodable by the server.


mikehaney24:
> LGTM expect a wrong PR reference in the description :)

Fixed, I don't know why, but github loves messing that part up.

PR #5134, opened 2020-03-17
Test Quickstarts in GHA

paulb777:
Migrate quickstarts tests from travis to GHA for Storage, Messaging, and Firestore

Make a check_secrets.sh that test for secrets being available in either travis or GHA

Unlike the GHA documentation states - GITHUB_BASE_REF is defined on the main repo. I'll follow up in a subsequent PR to work through and update the logic for PRs from a forked repo.

#no-changelog


.github/workflows/firestore.yml
```diff
@@ -150,3 +150,24 @@ jobs:
         scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseFirestore.podspec \
             --platforms=ios \
             --allow-warnings
+
+  quickstart:
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
```
maksymmalyhin: Just a thought about keeping our scripts CI agnostic. I wonder will it be easier to port to another CI if we create a single script to run all steps so we less rely on a CI specific config?

scripts/gha-encrypted/FIREGSignInInfo.h.gpg
```diff
@@ -0,0 +1,3 @@
+�	��\��������]G4U��a�$�v7+@l$�|��v��zȎ��#�S��dq�����e���l�1��Ņ�vg��W��q�Y�P�B$Ԗs������U������3W�b�鬮�%7N����~N�s�	4�����]��*�O�,���8����a�F#,�jkz5+)*
```
maksymmalyhin: It would be nice to have a reference doc explaining how to update the files available somewhere in the repo. Maybe a go-link to it in the repo?

.github/workflows/firestore.yml
paulb777: The secrets management is very CI specific, so I wasn't able to come up with a scripting solution that wouldn't add complexity.

scripts/gha-encrypted/FIREGSignInInfo.h.gpg
paulb777: Added a README with a pointer to the GHA instructions.

maksymmalyhin: Thank you!

scripts/gha-encrypted/README
```diff
@@ -0,0 +1 @@
+See https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets
```
maksymmalyhin: Would you like also to add info where the original files are or how to find/generate them?

.github/workflows/firestore.yml
wilhuff: While I'm generally supportive of keeping these scripts CI agnostic, secret retrieval is inherently CI specific.

I see this as:
  * Setup (bundler then quickstart)
  * Install secrets  <- CI specific
  * Run tests

We could probably stand to just call `setup_bundler.sh` from `setup_quickstart.sh` so that's just one step but otherwise I think this is factored pretty well.

scripts/gha-encrypted/README
paulb777: Good point! Done.

.travis.yml
```diff
@@ -370,12 +351,11 @@ jobs:
 
     - stage: test
       env:
-        - PROJECT=Firestore PLATFORM=iOS METHOD=xcodebuild QUICKSTART=firestore
+        - PROJECT=Firestore PLATFORM=iOS METHOD=xcodebuild
```
wilhuff: You can strike this stanza altogether. Firestore does its xcodebuild testing on GHA already. All we need over here is Xcode 10.3, and even that hasn't turned up a single issue for Firestore so you could probably scratch that too.

scripts/gha-encrypted/README
maksymmalyhin: Thanks!

scripts/check_secrets.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+# Copyright 2020 Google
```
wilhuff: Copyright to Google LLC. See go/copyright

Please update your template.

scripts/check_secrets.sh
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Check if secrets are available for multiple CI's
+
+set -x
+echo $GITHUB_BASE_REF
```
wilhuff: Please either remove this debug logging or make it easier to understand by labeling the output, e.g.

```bash
echo "GITHUB_BASE_REF: ${GITHUB_BASE_REF}"
```

scripts/check_secrets.sh
```diff
@@ -0,0 +1,45 @@
...
+echo $GITHUB_HEAD_REF
+
+check_secrets()
+{
+  have_secrets=false
```
wilhuff: Rather than modifying a global variable like this, better to rename this function to `have_secrets` and return `0` if true and return `1` for false. Then all the callsite instead of

```bash
if [[ "$have_secrets" == true ]]; then
  # ...
fi
```

You can do

```bash
if have_secrets; then
  # ...
fi
```

This means you can also remove the call to `check_secrets` below.

scripts/setup_quickstart.sh
```diff
@@ -31,12 +34,16 @@ if [[ "$TRAVIS_PULL_REQUEST" == "false" ||
   # To test a branch, uncomment the following line
   # git checkout {BRANCH_NAME}
 
+  bundle update --bundler
+  bundle install
   bundle exec pod install
-  TRAVIS_PULL_REQUEST="$TRAVIS_PULL_REQUEST" TRAVIS_PULL_REQUEST_SLUG=$"TRAVIS_PULL_REQUEST_SLUG" \
-    ../scripts/install_prereqs/"$SAMPLE".sh
+  ../scripts/install_prereqs/"$SAMPLE".sh
```
wilhuff: This could be made less fragile by capturing the scripts and root directory at the top. Something like this:

```bash
scripts_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
root_dir="$(dirname "$scripts_dir")"
```

Now you can refer to scripts like so, regardless of the current directory:

```bash
"$scripts_dir/install_prerequs/$SAMPLE.sh"
```

And then you can copy from `$root_dir/Secrets"

scripts/test_quickstart.sh
```diff
@@ -16,11 +16,12 @@
 # Run a CI `script` phase to build the associated quickstart
 # sample and run its tests.
 
-if [[ "$TRAVIS_PULL_REQUEST" == "false" ||
-      "$TRAVIS_PULL_REQUEST_SLUG" == "$TRAVIS_REPO_SLUG" ]]; then
+set -xeuo pipefail
+
+# Set have_secrets to true or false.
+. scripts/check_secrets.sh
```
wilhuff: When sourcing scripts, consider spelling out `source` rather than `.`. It makes this more obvious to the reader that you're not just calling the script but including into the currently running script.

scripts/setup_quickstart.sh
paulb777: Done - except for install_prereqs which is from the quickstart repo's scripts

scripts/test_quickstart.sh
paulb777: Done

paulb777:
Thanks for the reviews @maksymmalyhin and @wilhuff.  I'm going to merge it here.  Feel free to add additional review comments and I'll address in a subsequent PR.

PR #5137, opened 2020-03-18
Refactor event fileURLs to be derived

mikehaney24:
Apparently, since iOS 8, it's possible for the app's bundle to change directories as often as every launch. This is a mechanism that's hopefully slightly more robust against that, as data is supposed to carry over, the UUID component of the app's absolute directory might just change.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
@@ -19,6 +19,7 @@
 #import <GoogleDataTransport/GDTCORAssert.h>
 #import <GoogleDataTransport/GDTCORClock.h>
 #import <GoogleDataTransport/GDTCORConsoleLogger.h>
+#import <GoogleDataTransport/GDTCORPlatform.h>
 
 #import "GDTCORLibrary/Private/GDTCORDataFuture.h"
 #import "GDTCORLibrary/Private/GDTCOREvent_Private.h"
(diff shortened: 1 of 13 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -152,7 +167,12 @@ - (id)initWithCoder:(NSCoder *)aDecoder {
   if (self) {
     _qosTier = [aDecoder decodeIntegerForKey:qosTierKey];
     _clockSnapshot = [aDecoder decodeObjectOfClass:[GDTCORClock class] forKey:clockSnapshotKey];
-    _fileURL = [aDecoder decodeObjectOfClass:[NSURL class] forKey:fileURLKey];
+    NSURL *fileURL = [aDecoder decodeObjectOfClass:[NSURL class] forKey:fileURLKey];
+    if (fileURL) {
```
jasonhu-g: Nice backward compatibility 👍

jasonhu-g:
I tested the change and it looks good:
1. Turn off device internet
2. Send an event
3. Close the app
4. Turn on device internet
5. Relaunch the app (and the persisted event is sent up)

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
ryanwilson: +1 - it may be nice to add a comment here (in a follow up) to say why this was done for future readers.

paulb777:
I'm going to merge to start on the release branch.

@mikehaney24 Please follow up on @ryanwilson's comment in another PR on master.

PR #5147, opened 2020-03-18
Libraries should save state if a completion block is requested

mikehaney24:
Currently, the onComplete block of sendXEvent:onComplete: is invoked when the event bytes are written to disk. However, if the app crashes before a lifecycle event occurs (trigger GDTCORStorage to serialize to disk), that event is effectively lost because the metadata has all been lost. This change triggers the saving of state of GDTCORStorage and prioritizers (if they implement it) if an onComplete block has been given.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
@@ -59,14 +59,17 @@
 }
 
 - (void)storeEvent:(GDTCOREvent *)event
-        onComplete:(void (^)(BOOL wasWritten, NSError *error))completion {
+        onComplete:(void (^_Nullable)(BOOL wasWritten, NSError *error))completion {
   GDTCORLogDebug("Saving event: %@", event);
   if (event == nil) {
     GDTCORLogDebug("%@", @"The event was nil, so it was not saved.");
     return;
   }
+  BOOL hadOriginalCompletion = completion != nil;
   if (!completion) {
     completion = ^(BOOL wasWritten, NSError *error) {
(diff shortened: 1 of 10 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORStorage.m
```diff
@@ -102,17 +105,26 @@ - (void)storeEvent:(GDTCOREvent *)event
...
+      GDTCORLogDebug(@"Prioritizer %@ has saved state due to an event's onComplete block.",
+                     prioritizer);
+    }
 
     // Check the QoS, if it's high priority, notify the target that it has a high priority event.
     if (event.qosTier == GDTCOREventQoSFast) {
       [self.uploadCoordinator forceUploadForTarget:target];
     }
 
     // Write state to disk if we're in the background.
```
jasonhu-g: Update comment

GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
```diff
@@ -48,8 +48,9 @@ - (instancetype)init {
 
 - (void)transformEvent:(GDTCOREvent *)event
       withTransformers:(NSArray<id<GDTCOREventTransformer>> *)transformers
-            onComplete:(nonnull void (^)(BOOL wasWritten, NSError *error))completion {
+            onComplete:(void (^_Nullable)(BOOL wasWritten, NSError *error))completion {
   GDTCORAssert(event, @"You can't write a nil event");
+  BOOL hadOriginalCompletion = completion != nil;
   if (!completion) {
```
jasonhu-g: if (! hadOriginalCompletion)

mikehaney24: This would work, but I prefer the semantics of !completion leading to reassigning completion.

PR #5160, opened 2020-03-19
Make `Error` a scoped enumeration and prefix its members with `k`

var-const:
The `k` prefix is required by the style guide.

Note that the changes are mostly mechanical.


Firestore/core/src/firebase/firestore/util/error_apple.h
```diff
@@ -42,7 +42,7 @@ namespace firestore {
 namespace util {
 
 // Translates a set of error_code and error_msg to an NSError.
-NSError* MakeNSError(int64_t error_code,
```
var-const: I'm not sure why it used `int64_t` to begin with.

wilhuff: NSError can take an arbitrary NSInteger value and `int64_t` models that. I guess in practice we don’t pass anything but Errors so maybe this is fine.

PR #5163, opened 2020-03-20
Create background task on watchOS platform with tracking semaphores

doudounan:
- Change sharedWKExtensionForBackgroundTask to sharedNSProcessInfo on watchOS for using api `performExpiringActivityWithReason usingBlock`

- Add class methods for mapping GDTCORBackgroundIdentifier to task semaphores

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -400,7 +400,7 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
 
 - (GDTCORBackgroundIdentifier)beginBackgroundTaskWithName:(NSString *)name
                                         expirationHandler:(void (^)(void))handler {
-  __block GDTCORBackgroundIdentifier bgID;
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
 #if !TARGET_OS_WATCH
   bgID = [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
                                                               expirationHandler:handler];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -330,21 +330,100 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
   return self;
 }
 
+/** Creates a NSMutableDictionary for mapping bgID to task semaphore.
+ *
+ * @return A NSMutableDictionary singleton instance for mapping bgID to task semaphore.
+ */
++ (NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *)getBgIDToSemaphoreMap {
+  static NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *sBgIDToSemephoreMap;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    sBgIDToSemephoreMap = [[NSMutableDictionary alloc] init];
+  });
+  return sBgIDToSemephoreMap;
+}
+
+/** Creates a queue on which all task semaphore will populate and remove from map.
+ *
+ * @return A dispatch queue singleton instance for the lifetime of an application.
+ */
++ (dispatch_queue_t)getSemaphoreQueue {
+  static dispatch_queue_t sSemaphoreQueue;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    sSemaphoreQueue = dispatch_queue_create("com.google.GDTCORApplication", DISPATCH_QUEUE_SERIAL);
+  });
+  return sSemaphoreQueue;
+}
+
+/** Populates a unique NSUInteger as GDTCORBackgroundIdentifier for given semaphore and maps it to
+ * the semaphore in NSMutableDictionary.
+ *
+ * @param semaphore The task semaphore needs to be mapping.
+ * @return A unique GDTCORBackgroundIdentifier mapped to the given semaphore.
+ */
++ (GDTCORBackgroundIdentifier)populateBgIDwithSemaphore:(dispatch_semaphore_t)semaphore {
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
+  if ([self getSemaphoreQueue]) {
+    dispatch_sync([self getSemaphoreQueue], ^{
+      NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *bgIDToSemaphoreMap =
+          [self getBgIDToSemaphoreMap];
+      bgID = arc4random();
+      while (bgID == GDTCORBackgroundIdentifierInvalid ||
+             bgIDToSemaphoreMap[[NSNumber numberWithLong:bgID]]) {
+        bgID = arc4random();
+      }
+      bgIDToSemaphoreMap[[NSNumber numberWithLong:bgID]] = semaphore;
+    });
+  }
+  return bgID;
+}
+
+/** Returns the semaphore mapped be given bgID and removed the key value pair in
+ * NSMutableDictionary.
+ *
+ * @param bgID The unique NSUInteger as GDTCORBackgroundIdentifier.
+ * @return The semaphore mapped by given bgID.
+ */
++ (dispatch_semaphore_t)semaphoreForBgID:(GDTCORBackgroundIdentifier)bgID {
+  __block dispatch_semaphore_t semaphore;
+  if ([self getSemaphoreQueue]) {
+    dispatch_sync([self getSemaphoreQueue], ^{
+      NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *bgIDToSemaphoreMap =
+          [self getBgIDToSemaphoreMap];
+      semaphore = bgIDToSemaphoreMap[[NSNumber numberWithLong:bgID]];
+      [bgIDToSemaphoreMap removeObjectForKey:[NSNumber numberWithLong:bgID]];
+    });
+  }
+  return semaphore;
+}
+
 - (GDTCORBackgroundIdentifier)beginBackgroundTaskWithName:(NSString *)name
                                         expirationHandler:(void (^)(void))handler {
+  __block GDTCORBackgroundIdentifier bgID;
 #if !TARGET_OS_WATCH
-  GDTCORBackgroundIdentifier bgID =
-      [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
-                                                           expirationHandler:handler];
+  bgID = [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
+                                                              expirationHandler:handler];
 #if !NDEBUG
   if (bgID != GDTCORBackgroundIdentifierInvalid) {
     GDTCORLogDebug("Creating background task with name:%@ bgID:%ld", name, (long)bgID);
   }
 #endif  // !NDEBUG
+#elif TARGET_OS_WATCH
+  [[self sharedNSProcessInfoForBackgroundTask]
+      performExpiringActivityWithReason:name
+                             usingBlock:^(BOOL expired) {
+                               dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+                               bgID = [GDTCORApplication populateBgIDwithSemaphore:semaphore];
+                               if (expired) {
+                                 handler();
+                                 dispatch_semaphore_signal(semaphore);
+                               }
+                               dispatch_semaphore_wait(
+                                   semaphore, dispatch_time(DISPATCH_TIME_NOW, 30 * NSEC_PER_SEC));
+                             }];
+#endif
   return bgID;
-#endif  // !TARGET_OS_WATCH
-  // TODO: WKExtension background tasks handling.
-  return GDTCORBackgroundIdentifierInvalid;
 }
 
 - (void)endBackgroundTask:(GDTCORBackgroundIdentifier)bgID {
@@ -353,6 +432,13 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
     GDTCORLogDebug("Ending background task with ID:%ld was successful", (long)bgID);
     [[self sharedApplicationForBackgroundTask] endBackgroundTask:bgID];
     return;
+  }
+#elif TARGET_OS_WATCH
+  if (bgID != GDTCORBackgroundIdentifierInvalid) {
+    dispatch_semaphore_t semaphore = [GDTCORApplication semaphoreForBgID:bgID];
+    if (semaphore) {
+      dispatch_semaphore_signal(semaphore);
+    }
   }
 #endif  // !TARGET_OS_WATCH
 }
@@ -364,14 +450,14 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
   return appExtension;
 }
 
-/** Returns a UIApplication or WKExtension instance if on the appropriate platform.
+/** Returns a UIApplication or NSProcessInfo instance if on the appropriate platform.
  *
- * @return The shared UIApplication or WKExtension if on the appropriate platform.
+ * @return The shared UIApplication or NSProcessInfo if on the appropriate platform.
  */
 #if TARGET_OS_IOS || TARGET_OS_TV
 - (nullable UIApplication *)sharedApplicationForBackgroundTask {
 #elif TARGET_OS_WATCH
-- (nullable WKExtension *)sharedApplicationForBackgroundTask {
+- (nullable NSProcessInfo *)sharedNSProcessInfoForBackgroundTask {
 #else
 - (nullable id)sharedApplicationForBackgroundTask {
 #endif
@@ -381,7 +467,7 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
     sharedInstance = [UIApplication sharedApplication];
   }
 #elif TARGET_OS_WATCH
-  sharedInstance = [WKExtension sharedExtension];
+  sharedInstance = [NSProcessInfo processInfo];
 #endif
   return sharedInstance;
 }
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -264,10 +264,10 @@ static NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *sBgIDToSemephoreMa
     dispatch_once(&onceToken, ^{
       sSemaphoreQueue =
           dispatch_queue_create("com.google.GDTCORApplication", DISPATCH_QUEUE_SERIAL);
-      NSLog(@"GDTCORApplication is initializing on watchOS, sSemaphoreQueue has been initialized");
+      NSLog(@"GDTCORApplication is initializing on watchOS, sSemaphoreQueue has been initialized.");
       sBgIDToSemephoreMap = [[NSMutableDictionary alloc] init];
       NSLog(@"GDTCORApplication is initializing on watchOS, sBgIDToSemephoreMap has been "
-            @"initialized");
+            @"initialized.");
     });
   }
 #endif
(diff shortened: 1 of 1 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +330,100 @@ - (instancetype)init {
...
+    sBgIDToSemephoreMap = [[NSMutableDictionary alloc] init];
+  });
+  return sBgIDToSemephoreMap;
+}
+
+/** Creates a queue on which all task semaphore will populate and remove from map.
+ *
+ * @return A dispatch queue singleton instance for the lifetime of an application.
+ */
++ (dispatch_queue_t)getSemaphoreQueue {
```
doudounan: Here I separate the queue init and dic init, because we need to manipulate the dic on the queue and the queue should init before dic.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +330,100 @@ - (instancetype)init {
   return self;
 }
 
+/** Creates a NSMutableDictionary for mapping bgID to task semaphore.
```
mikehaney24: Creates an NSMutableDictionary

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
mikehaney24: things that occur within a dispatch_once are thread-safe. If the queue and dictionary are initialized at the same time and before a get or set on the dictionary is called (where the queue is needed), then it should be fine. The difficulty is generally determining where "before a get or set on the dictionary is called" is. It could be in +load, +initialize, or using a shared dispatch_once_t with two dispatch_once calls each in the get and set methods.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +330,100 @@ - (instancetype)init {
...
+
+/** Populates a unique NSUInteger as GDTCORBackgroundIdentifier for given semaphore and maps it to
+ * the semaphore in NSMutableDictionary.
+ *
+ * @param semaphore The task semaphore needs to be mapping.
+ * @return A unique GDTCORBackgroundIdentifier mapped to the given semaphore.
+ */
++ (GDTCORBackgroundIdentifier)populateBgIDwithSemaphore:(dispatch_semaphore_t)semaphore {
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
+  if ([self getSemaphoreQueue]) {
```
mikehaney24: You should assign this to a variable for 2 reasons:
1. To reduce the number of times you send a message (call a method)
2. Without a strong reference to a variable, there's no guarantee that the object from the first method call is the same as the second, or that the 2nd time you call the method that it'll return the same thing.

Learning more about strong and weak variables could be useful. Here's one link to help learn more: https://stackoverflow.com/a/11013715

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +330,100 @@ - (instancetype)init {
...
   }
 #endif  // !NDEBUG
+#elif TARGET_OS_WATCH
+  [[self sharedNSProcessInfoForBackgroundTask]
+      performExpiringActivityWithReason:name
+                             usingBlock:^(BOOL expired) {
+                               dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+                               bgID = [GDTCORApplication populateBgIDwithSemaphore:semaphore];
+                               if (expired) {
+                                 handler();
```
mikehaney24: Check to see if handler is non NULL

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +330,100 @@ - (instancetype)init {
...
+                                 dispatch_semaphore_signal(semaphore);
+                               }
+                               dispatch_semaphore_wait(
+                                   semaphore, dispatch_time(DISPATCH_TIME_NOW, 30 * NSEC_PER_SEC));
```
mikehaney24: Any particular reasoning behind 30s?

doudounan: No idea. I thought it should be some explicit time expect FOREVER.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Moved initialization of queue and dictionary to +(void)initialize in GDTCORApplication.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed queue to a class wide static variable.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -237,6 +237,9 @@ @interface GDTCORApplication ()
 
 @implementation GDTCORApplication
 
+static dispatch_queue_t sSemaphoreQueue;
```
mikehaney24: Variables need documentation

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: I think every task we create should run very fast much less than 30s. Or should we change this waiting time less than 30s?

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -237,6 +237,9 @@ @interface GDTCORApplication ()
...
+static NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *sBgIDToSemephoreMap;
```
mikehaney24: Maybe rename sBackgroundIdentifierToSemaphoreMap

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -251,6 +254,25 @@ + (void)load {
...
++ (void)initialize {
+#if TARGET_OS_WATCH
+  if (self == [GDTCORApplication class]) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+      sSemaphoreQueue =
+          dispatch_queue_create("com.google.GDTCORApplication", DISPATCH_QUEUE_SERIAL);
+      NSLog(@"GDTCORApplication is initializing on watchOS, sSemaphoreQueue has been initialized");
+      sBgIDToSemephoreMap = [[NSMutableDictionary alloc] init];
+      NSLog(@"GDTCORApplication is initializing on watchOS, sBgIDToSemephoreMap has been "
```
mikehaney24: s/Seme/Sema/, or apply above suggested rename

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +352,77 @@ - (instancetype)init {
   return self;
 }
 
+/** Populates a unique NSUInteger as GDTCORBackgroundIdentifier for given semaphore and maps it to
+ * the semaphore in NSMutableDictionary.
+ *
+ * @param semaphore The task semaphore needs to be mapping.
+ * @return A unique GDTCORBackgroundIdentifier mapped to the given semaphore.
+ */
++ (GDTCORBackgroundIdentifier)populateBgIDwithSemaphore:(dispatch_semaphore_t)semaphore {
```
mikehaney24: maybe rename to createAndMapBackgroundIdentifierToSemaphore

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,21 +352,77 @@ - (instancetype)init {
...
+  return bgID;
+}
+
+/** Returns the semaphore mapped be given bgID and removed the key value pair in
+ * NSMutableDictionary.
+ *
+ * @param bgID The unique NSUInteger as GDTCORBackgroundIdentifier.
+ * @return The semaphore mapped by given bgID.
+ */
++ (dispatch_semaphore_t)semaphoreForBgID:(GDTCORBackgroundIdentifier)bgID {
```
mikehaney24: maybe rename to semaphoreForBackgroundIdentifier

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
mikehaney24: NSLogs are ok in test app code, use GDTConsoleLogger methods for library code

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Added

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Renamed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Renamed.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Renamed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -251,6 +259,24 @@ + (void)load {
...
+#if TARGET_OS_WATCH
+  if (self == [GDTCORApplication class]) {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+      sSemaphoreQueue =
+          dispatch_queue_create("com.google.GDTCORApplication", DISPATCH_QUEUE_SERIAL);
+      GDTCORLogDebug(
+          @"GDTCORApplication is initializing on watchOS, sSemaphoreQueue has been initialized.");
+      sBackgroundIdentifierToSemaphoreMap = [[NSMutableDictionary alloc] init];
+      GDTCORLogDebug(@"GDTCORApplication is initializing on watchOS, sBgIDToSemephoreMap has been "
```
mikehaney24: s/sBgIDToSemephoreMap/sBackgroundIdentifierToSemaphoreMap/

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -251,6 +259,24 @@ + (void)load {
   [self sharedApplication];
 }
 
+/** Initializes sSemaphoreQueue and sBackgroundIdentifierToSemaphoreMap. */
++ (void)initialize {
+#if TARGET_OS_WATCH
+  if (self == [GDTCORApplication class]) {
```
mikehaney24: Why the class equality check? If a subclass of GDTCORApplication were to be made, this wouldn't work anymore.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
+ * the semaphore in NSMutableDictionary.
+ *
+ * @param semaphore The task semaphore needs to be mapping.
+ * @return A unique GDTCORBackgroundIdentifier mapped to the given semaphore.
+ */
++ (GDTCORBackgroundIdentifier)createAndMapBackgroundIdentifierToSemaphore:
+    (dispatch_semaphore_t)semaphore {
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
+  if (sSemaphoreQueue && sBackgroundIdentifierToSemaphoreMap) {
+    dispatch_sync(sSemaphoreQueue, ^{
```
mikehaney24: It's technically possible for sSemaphoreQueue and sBackgroundIdentifierToSemaphoreMap to be deallocated between lines 368 and 369. The way you solve this is you read them into local variables (which will have a strong reference until the end of this method) and then check the values of the variables. That ensures that they can't be deallocated after you checked but before you used them.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -251,6 +259,24 @@ + (void)load {
   [self sharedApplication];
 }
 
+/** Initializes sSemaphoreQueue and sBackgroundIdentifierToSemaphoreMap. */
```
mikehaney24: a docstring (comments that look like `/** */`) aren't necessary here, +initialize is a method whose declaration exists elsewhere. You can just use an implementation comment (`//`)

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -237,6 +237,14 @@ @interface GDTCORApplication ()
 
 @implementation GDTCORApplication
 
+/** A dispatch queue on which all task semaphore will populate and remove from
+ * sBackgroundIdentifierToSemaphoreMap.
+ */
+static dispatch_queue_t sSemaphoreQueue;
+
+/** An NSMutableDictionary for mapping backgroundIdentifier to task semaphore. */
```
mikehaney24: `/** For mapping backgroundIdentifier to task semaphore. */`

You don't need to reiterate the type in the comment.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
   return self;
 }
 
+/** Populates a unique NSUInteger as GDTCORBackgroundIdentifier for given semaphore and maps it to
```
mikehaney24: Generates and maps a unique background identifier to the given semaphore.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
+ * the semaphore in NSMutableDictionary.
+ *
+ * @param semaphore The task semaphore needs to be mapping.
```
mikehaney24: The semaphore to map.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
+    (dispatch_semaphore_t)semaphore {
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
+  if (sSemaphoreQueue && sBackgroundIdentifierToSemaphoreMap) {
+    dispatch_sync(sSemaphoreQueue, ^{
+      bgID = arc4random();
+      while (bgID == GDTCORBackgroundIdentifierInvalid ||
+             sBackgroundIdentifierToSemaphoreMap[[NSNumber numberWithLong:bgID]]) {
+        bgID = arc4random();
+      }
+      sBackgroundIdentifierToSemaphoreMap[[NSNumber numberWithLong:bgID]] = semaphore;
```
mikehaney24: replace `[NSNumber numberWithLong:]` with just `@(bgID)`, here and several other places

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
+    });
+  }
+  return bgID;
+}
+
+/** Returns the semaphore mapped be given bgID and removes the key value pair in
```
mikehaney24: Returns the semaphore mapped to given bgID and removes the value from the map.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
+ * NSMutableDictionary.
+ *
+ * @param bgID The unique NSUInteger as GDTCORBackgroundIdentifier.
+ * @return The semaphore mapped by given bgID.
+ */
++ (dispatch_semaphore_t)semaphoreForBackgroundIdentifier:(GDTCORBackgroundIdentifier)bgID {
+  __block dispatch_semaphore_t semaphore;
+  if (sSemaphoreQueue && sBackgroundIdentifierToSemaphoreMap) {
+    dispatch_sync(sSemaphoreQueue, ^{
```
mikehaney24: same comment as above, with regard to variables

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
mikehaney24: Seems fine, I just wanted to know if there was reasoning. If we know about how long background tasks have to execute on watchOS before the expiration handler, we should set it to shorter than that.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +356,97 @@ - (instancetype)init {
...
   if (bgID != GDTCORBackgroundIdentifierInvalid) {
-    GDTCORLogDebug("Ending background task with ID:%ld was successful", (long)bgID);
+    GDTCORLogDebug("Ending background task with ID:%ld was successful.", (long)bgID);
     [[self sharedApplicationForBackgroundTask] endBackgroundTask:bgID];
     return;
   }
+#elif TARGET_OS_WATCH
+  if (bgID != GDTCORBackgroundIdentifierInvalid) {
+    dispatch_semaphore_t semaphore = [GDTCORApplication semaphoreForBackgroundIdentifier:bgID];
+    GDTCORLogDebug(@"Ending activity with bgID:%ld on watchOS.", (long)bgID);
```
mikehaney24: You print out if semaphore is nil.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: remove comment for initialize

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Changed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: changed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Removed class check.
- It depends on whether we want a subclass of GDTCORApplication to call the super initialize method. In our case, the queue and map will only initialize once for one application lifetime with dispatch once. So I think we can safely remove the class check.

doudounan:
Create another PR with same changes on new feature branch https://github.com/firebase/firebase-ios-sdk/pull/5185

PR #5172, opened 2020-03-23
Only delete events when package was successfully delivered

mikehaney24:
Fixes an issue where the upload coordinator removes events from disk when the upload was unsuccessful and should be retried in the future.

### DIFF
--- a/GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCOREventGenerator.m
+++ b/GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCOREventGenerator.m
@@ -26,7 +26,6 @@
 
 + (NSMutableSet<GDTCOREvent *> *)generate3Events {
   static NSUInteger counter = 0;
-  NSString *filePath = [NSString stringWithFormat:@"test-%ld.txt", (unsigned long)counter];
   int howManyToGenerate = 3;
   NSMutableSet<GDTCOREvent *> *set = [[NSMutableSet alloc] initWithCapacity:howManyToGenerate];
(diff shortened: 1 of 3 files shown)


mikehaney24:
I added a unit test and confirmed that the test fails without the changes I made.

GoogleDataTransport/GDTCORTests/Unit/Helpers/GDTCOREventGenerator.m
```diff
@@ -26,14 +26,14 @@ @implementation GDTCOREventGenerator
...
   static NSUInteger counter = 0;
-  NSString *filePath = [NSString stringWithFormat:@"test-%ld.txt", (unsigned long)counter];
   int howManyToGenerate = 3;
   NSMutableSet<GDTCOREvent *> *set = [[NSMutableSet alloc] initWithCapacity:howManyToGenerate];
   for (int i = 0; i < howManyToGenerate; i++) {
     GDTCOREvent *event = [[GDTCOREvent alloc] initWithMappingID:@"1337" target:50];
     event.clockSnapshot = [GDTCORClock snapshot];
     event.qosTier = GDTCOREventQosDefault;
     event.dataObject = [[GDTCORDataObjectTesterSimple alloc] initWithString:@"testing!"];
+    NSString *filePath = [NSString stringWithFormat:@"test-%ld.txt", (unsigned long)counter];
```
paulb777: Why not use `i` instead of `counter`?

mikehaney24: Because using i could result in creating duplicate event fileURLs across multiple calls to this method

PR #5179, opened 2020-03-24
Test GHA on forked PR

paulb777:

PR #5180, opened 2020-03-24
GHA for PRs from GitHub forks

paulb777:
This PR updates the storage GHA workflow to work on PRs from both the main repo and forked repos by disabling the tests that depend upon secrets when the associated environment variable is not set.

It would be nice to do this at the job level, but I wasn't able to find a way. 

The change is verified here for the main repo and in #5179 for a forked repo.

### DIFF
--- a/.github/workflows/dynamiclinks.yml
+++ b/.github/workflows/dynamiclinks.yml
@@ -13,7 +13,7 @@ jobs:
   pod_lib_lint:
     runs-on: macOS-latest
     steps:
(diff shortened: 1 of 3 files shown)


paulb777:
I started a discussion at https://github.community/t5/GitHub-Actions/How-can-I-test-if-secrets-are-available-in-an-action/m-p/51105/highlight/false#M8057 to check if there's a better way to do this. I'll wait on that thread before propagating this secrets PR solution to firestore and messaging.

PR #5183, opened 2020-03-24
Change FServerValues to just-in-time read from SyncTrees.

inlined:
Based on fix https://github.com/firebase/firebase-js-sdk/pull/2499 to bug https://github.com/firebase/firebase-js-sdk/issues/2487

Micro-benchmarking showed that N writes in succession led to N^2 performance when calcuating the resolved write tree for those writes (thankfully only at the subpath in this SDK). This change matches the optimization in JS that mitigated a 20% performance regression.

### DIFF
--- a/Firebase/Database/Core/FServerValues.h
+++ b/Firebase/Database/Core/FServerValues.h
@@ -18,19 +18,23 @@
 #import "FCompoundWrite.h"
 #import "FNode.h"
 #import "FSparseSnapshotTree.h"
+#import "FSyncTree.h"
 #import <Foundation/Foundation.h>
 
 @interface FServerValues : NSObject
 
 + (NSDictionary *)generateServerValues:(id<FClock>)clock;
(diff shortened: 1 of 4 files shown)


Firebase/Database/Core/FServerValues.h
```diff
@@ -18,19 +18,23 @@
...
-                           withExisting:(id<FNode>)existing
-                           serverValues:(NSDictionary *)serverValues;
+
++ (FCompoundWrite *)resolveDeferredValueCompoundWrite:(FCompoundWrite *)write
+                                         withSyncTree:(FSyncTree *)tree
+                                               atPath:(FPath *)path
+                                         serverValues:
+                                             (NSDictionary *)serverValues;
++ (id<FNode>)resolveDeferredValueSnapshot:(id<FNode>)node
+                             withSyncTree:(id<FNode>)existing
```
schmidt-sebastian: withExisting?

PR #5185, opened 2020-03-24
Handle tasks with semaphores on watchOS Platform

doudounan:
- Change sharedWKExtensionForBackgroundTask to sharedNSProcessInfo on watchOS for using api performExpiringActivityWithReason usingBlock
- Add static variable sSemaphoreQueue and sBackgroundIdentifierToSemaphoreMap
- Add class methods for mapping GDTCORBackgroundIdentifier to task semaphores

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
@@ -237,6 +237,14 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
 
 @implementation GDTCORApplication
 
+/** A dispatch queue on which all task semaphore will populate and remove from
+ * sBackgroundIdentifierToSemaphoreMap.
+ */
+static dispatch_queue_t sSemaphoreQueue;
+
+/** For mapping backgroundIdentifier to task semaphore. */
+static NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *sBackgroundIdentifierToSemaphoreMap;
+
 + (void)load {
   GDTCORLogDebug(
       "%@", @"GDT is initializing. Please note that if you quit the app via the "
@@ -249,6 +257,20 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
       @"GDTCORBackgroundIdentifierInvalid and UIBackgroundTaskInvalid should be the same.");
 #endif
   [self sharedApplication];
+}
+
++ (void)initialize {
+#if TARGET_OS_WATCH
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    sSemaphoreQueue = dispatch_queue_create("com.google.GDTCORApplication", DISPATCH_QUEUE_SERIAL);
+    GDTCORLogDebug(
+        @"GDTCORApplication is initializing on watchOS, sSemaphoreQueue has been initialized.");
+    sBackgroundIdentifierToSemaphoreMap = [[NSMutableDictionary alloc] init];
+    GDTCORLogDebug(@"GDTCORApplication is initializing on watchOS, "
+                   @"sBackgroundIdentifierToSemaphoreMap has been initialized.");
+  });
+#endif
 }
 
 + (nullable GDTCORApplication *)sharedApplication {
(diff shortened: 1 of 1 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -237,6 +237,14 @@ @interface GDTCORApplication ()
 
 @implementation GDTCORApplication
 
+/** A dispatch queue on which all task semaphore will populate and remove from
+ * sBackgroundIdentifierToSemaphoreMap.
+ */
+static dispatch_queue_t sSemaphoreQueue;
```
mikehaney24: This is my fault, but rename sSemaphoreQueue and sBackgroundIdentifierToSemaphoreMap to have a g prefix instead of an s.

go/objc-style#common-variable-names

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +352,100 @@ - (instancetype)init {
...
+ */
++ (GDTCORBackgroundIdentifier)createAndMapBackgroundIdentifierToSemaphore:
+    (dispatch_semaphore_t)semaphore {
+  __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
+  dispatch_queue_t queue = sSemaphoreQueue;
+  NSMutableDictionary<NSNumber *, dispatch_semaphore_t> *map = sBackgroundIdentifierToSemaphoreMap;
+  if (queue && map) {
+    dispatch_sync(queue, ^{
+      bgID = arc4random();
+      while (bgID == GDTCORBackgroundIdentifierInvalid || map[@(bgID)]) {
```
mikehaney24: You can save the result of @(bgID) into a variable so that you only call that method once.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +352,100 @@ - (instancetype)init {
...
 #if !TARGET_OS_WATCH
-  GDTCORBackgroundIdentifier bgID =
-      [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
-                                                           expirationHandler:handler];
+  bgID = [[self sharedApplicationForBackgroundTask] beginBackgroundTaskWithName:name
+                                                              expirationHandler:handler];
 #if !NDEBUG
   if (bgID != GDTCORBackgroundIdentifierInvalid) {
-    GDTCORLogDebug("Creating background task with name:%@ bgID:%ld", name, (long)bgID);
+    GDTCORLogDebug("Creating background task with name:%@ bgID:%ld.", name, (long)bgID);
```
mikehaney24: I don't recommend the period for this debug log because a space will be a better delimiter than a period since it's a value you might copy and paste.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -330,30 +352,100 @@ - (instancetype)init {
...
+                               if (expired) {
+                                 if (handler) {
+                                   handler();
+                                 }
+                                 dispatch_semaphore_signal(semaphore);
+                                 GDTCORLogDebug(
+                                     @"Activity with name:%@ bgID:%ld on watchOS is expiring.",
+                                     name, (long)bgID);
+                               }
+                               dispatch_semaphore_wait(
```
mikehaney24: This needs to be put into an else block, or the if (expired) block above needs to have a return call in there. Otherwise, dispatch_semaphore_wait will always be called and it'll freeze the app or force it to be terminated.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
doudounan: Fixed

PR #5187, opened 2020-03-25
Fix a couple of race conditions and module import of a private header

mikehaney24:
The header and API in question should be considered hidden. An alternative would be expose the API in the private header .

A race condition existed in the upload package itself--it should hand copies of itself and not self to prevent asynchronous mutation whilst a different queue is also executing an asynchronous mutation. One alternative would be to create mutable an immutable variants of the package. This could manifest as an EXC_BAD_ACCESS in GDTCCTPrioritizer.m

### DIFF
--- a/GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.m
+++ b/GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.m
@@ -18,7 +18,6 @@
 
 #import <GoogleDataTransport/GDTCORAssert.h>
 #import <GoogleDataTransport/GDTCOREventDataObject.h>
-#import <GoogleDataTransport/GDTCOREvent_Private.h>
 #import <GoogleDataTransport/GDTCORPlatform.h>
 #import <GoogleDataTransport/GDTCORTargets.h>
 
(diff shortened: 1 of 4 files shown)


GoogleDataTransportCCTSupport/GDTCCTTests/Unit/Helpers/GDTCCTEventGenerator.m
```diff
@@ -55,6 +54,15 @@ - (void)deleteGeneratedFilesFromDisk {
   }
 }
 
+- (void)writeEvent:(GDTCOREvent *)event toGDTPath:(NSString *)path error:(NSError **)error {
+  SEL sel = NSSelectorFromString(@"writeToGDTPath:error:");
+  IMP imp = [event methodForSelector:sel];
+  if (imp) {
```
maksymmalyhin: I wonder if using [respondsToSelector:](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418583-respondstoselector?language=objc) will work for this case? It would look a bit more clean and Objective C friendly. 

Another question, why the event may not provide an implementation for the method?

mikehaney24: It would work for the check, but I have to get the IMP anyway, so might as well check that (in test code).

maksymmalyhin: In this particular method you can go without IMP e.g.:
```
[(id)event writeToGDTPath:path error:error];
```

PR #5196, opened 2020-03-25
Change C++ default log level to `kLogLevelNotice`

var-const:
This is for consistency with [iOS](https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleUtilities/Logger/GULLogger.m#L66) and [C++ RTDB](https://github.com/firebase/firebase-cpp-sdk/blob/master/database/src/include/firebase/database.h#L175).


var-const:
Let me know if I'm missing something and there's a reason to keep `kLogLevelWarning`.

PR #5202, opened 2020-03-26
Log the total pass and skip count after running all spec tests

dconeybe:
This is a port of the following PR in the Android SDK: https://github.com/firebase/firebase-android-sdk/pull/1373.

### DIFF
--- a/Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
+++ b/Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
@@ -831,6 +831,8 @@ ByteString MakeResumeToken(NSString *specString) {
   }
 
   // Now iterate over them and run them.
+  __block NSUInteger testPassCount = 0;
+  __block NSUInteger testSkipCount = 0;
   __block bool ranAtLeastOneTest = NO;
   for (NSUInteger i = 0; i < specFiles.count; i++) {
     NSLog(@"Spec test file: %@", specFiles[i]);
@@ -853,7 +855,9 @@ ByteString MakeResumeToken(NSString *specString) {
         NSLog(@"  Spec test: %@", name);
         [self runSpecTestSteps:steps config:config];
         ranAtLeastOneTest = YES;
+        testPassCount++;
       } else {
+        testSkipCount++;
         NSLog(@"  [SKIPPED] Spec test: %@", name);
         NSString *comment = testDescription[@"comment"];
         if (comment) {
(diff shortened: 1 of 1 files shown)


Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -853,7 +855,9 @@ - (void)testSpecTests {
         NSLog(@"  Spec test: %@", name);
         [self runSpecTestSteps:steps config:config];
         ranAtLeastOneTest = YES;
+        testPassCount++;
```
schmidt-sebastian: Google style guide suggest:

Use the prefix form (++i) of the increment and decrement operators unless you need postfix semantics.

https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#Preincrement_and_Predecrement

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -831,6 +831,8 @@ - (void)testSpecTests {
   }
 
   // Now iterate over them and run them.
+  __block NSUInteger testPassCount = 0;
```
wilhuff: Avoid platform dependent types like `NSUInteger` unless you specifically need them to interact with Objective-C APIs. This can just be `int`.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -862,6 +866,8 @@ - (void)testSpecTests {
       }
     }];
   }
+  NSLog(@"%@ completed; pass=%@ skip=%@", NSStringFromClass([self class]), @(testPassCount),
```
wilhuff: Now that these values are plain ints, you can make the format `%d` and pass the value without boxing.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
dconeybe: Done.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
dconeybe: Done.  Changed to the prefix operator.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
dconeybe: Done.

PR #5218, opened 2020-03-28
Fix calloc calls, which had the argument order reversed

mikehaney24:
Also, maybe I should resign from software engineering

From `man calloc`, which was not the source I used (dumb).

```
     void *
     calloc(size_t count, size_t size);

```


mikehaney24:
> The implementation likely just multiplies the two values anyway.

I'd expect so: https://github.com/rofl0r/musl/blob/master/src/malloc/calloc.c

But we were seeing weird kind-of corrupt memory bugs parsing the data on the backend.

PR #5224, opened 2020-03-30
Run Danger from GHA

maksymmalyhin:
Move Danger job from Travis to GitHub Actions.

#no-changelog

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -8,10 +8,8 @@ jobs:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v2
-
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
-
     - name: Danger
       env:
         DANGER_GITHUB_API_TOKEN: ${{ secrets.DangerGitHubAPIToken }}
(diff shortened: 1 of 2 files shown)


paulb777:
What's different about the danger that is directly integrated to GitHub versus the danger run that is integrated into GHA (and was travis)?

maksymmalyhin:
@paulb777 TBH I was not aware of a GitHub danger version. Would you share a link please?

paulb777:
See the two danger runs in this PR - @morganchen12, do you know how danger is integrated?

Screen Shot 2020-03-30 at 9 47 07 AM

maksymmalyhin:
> See the two danger runs in this PR - @morganchen12, do you know how danger is integrated?
> 

Discussed with Morgan: the check is added by the Danger run from CI - it is not a separate run.

paulb777:
Do you know why the check and danger runs are now both doubled on the PR?

morganchen12:
Danger reports a separate check because it was initially designed to run from a travis `before_script`, which is unable to separate danger failures and main job script failures. It still does this in GHA, even though the GHA job is granular enough to report its own success/failure.

PR #5225, opened 2020-03-30
Add Swift Result type APIs to Storage

paulb777:
Add a new API that takes advantage of the [Swift Result enum type](https://developer.apple.com/documentation/swift/result). An API is added for the 10 StorageReference APIs that returned a callback with an optional type and optional error.  Instead a single Result enum is returned.

21 of the 25 integration tests used one of those APIs and are ported in this PR.

An initial FirebaseStorageSwift.podspec is created that is intended to work like FirebaseFirestoreSwift.podspec.

Documentation and CI will follow in subsequent PR(s).

To try it out:

- `git fetch`
- `git checkout pb-storage-result`
- `pod gen FirebaseStorageSwift.podspec --auto-open --local-sources=./ --platforms=ios`
- Test the FirebaseStorageSwiftUnit-integration scheme


FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,96 @@
...
+            completion: @escaping (Result<StorageListResult, Error>) -> Void) {
+    list(withMaxResults: maxResults,
+         completion: getResultCallback(completion: completion))
+  }
+
+  func listAll(completion: @escaping (Result<StorageListResult, Error>) -> Void) {
+    listAll(completion: getResultCallback(completion: completion))
+  }
+
+  func putData(uploadData: Data,
```
wilhuff: Labeling this as `uploadData` will force your callers to also label, when the underlying call does not. This should be

```swift
func putData(_ uploadData: Data, ...);
```

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,96 @@
...
+  func putFile(from: URL,
+               metadata: StorageMetadata?,
+               completion: @escaping (Result<StorageMetadata, Error>) -> Void)
+    -> StorageUploadTask {
+    return putFile(from: from,
+                   metadata: metadata,
+                   completion: getResultCallback(completion: completion))
+  }
+
+  func updateMetadata(metadata: StorageMetadata,
```
wilhuff: The label here is duplicative. This should be

```swift
func updateMetadata(_ metadata: StorageMetadata, ...);
```

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,96 @@
...
+                      completion: @escaping (Result<StorageMetadata, Error>) -> Void) {
+    return updateMetadata(metadata, completion: getResultCallback(completion: completion))
+  }
+
+  func writeToFile(toFile: URL, completion: @escaping (Result<URL, Error>)
```
wilhuff: `writeToFile(...)` is not the same as the underlying method. This should be:

```swift
func write(toFile: URL, ...);
```

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,598 @@
...
+      do {
+        let bundle = Bundle(for: StorageIntegration.self)
+        let filePath = try XCTUnwrap(bundle.path(forResource: "1mb", ofType: "dat"),
+                                     "Failed to get filePath")
+        let data = try XCTUnwrap(try Data(contentsOf: URL(fileURLWithPath: filePath)),
+                                 "Failed to load file")
+
+        for largeFile in largeFiles {
+          let ref = storage.reference().child(largeFile)
+          _ = ref.putData(uploadData: data, metadata: nil, completion: { result in
```
wilhuff: If it's expected that `metadata` will commonly be `nil`, we should add an overload that allows that label to be omitted.

wilhuff: You should write these invocations using trailing closure syntax to verify that works as expected--that's the way most users will want to use this API.

This feedback applies throughout. You should avoid explicitly labeling the completion because this can mask overload resolution bugs that users could encounter if we're not expliciting testing that trailing closure syntax works.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,598 @@
...
+            switch result {
+            case let .success(metadata):
+              XCTAssertEqual(metadata.name, "1mb")
+              setupExpectation.fulfill()
+            case let .failure(error):
+              XCTFail("Unexpected error \(error) from putData")
```
wilhuff: I believe you need to fulfill the expectation in either the success or failure case. Otherwise if this were to fail the test is going to additionally time out waiting for expectations.

Please manually test some of these failure cases (perhaps by breaking the storage SDK implementation) to verify that the test isn't timing out.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,598 @@
...
+            case let .success(metadata):
+              XCTAssertNotNil(metadata, "Metadata should not be nil")
+              setupExpectation.fulfill()
+            case let .failure(error):
+              XCTFail("Unexpected error \(error) from putData")
+            }
+          })
+        }
+      } catch {
+        XCTFail("Exception thrown setting up files in setUp")
```
wilhuff: Swift has errors, not exceptions.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,598 @@
...
+      }
+      waitForExpectations()
```
wilhuff: This is going to wait for expectations even after an error is thrown. This should probably be inside the `do`/`catch` block.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
maksymmalyhin: Good point. A more Swift friendly name probably will be :
```
func put(data: Data, ...)
```

similar to mentioned below `func write(toFile: URL, ...)`

paulb777: Which is more important? - Choose the more Swift-friendly `put` or keep `putData` to be consistent with the existing Swift API?

I'm leaning towards the former so that it is clear that the two `putData` API's are exactly the same except for their completion parameter.

wilhuff: Since you're trying to add an overload, I'd steer for more consistency with the existing API than anything else, though @ryanwilson and the API council's opinion probably matters more than mine.

ryanwilson: +1 for consistency, and I recall explicitly discussing something similar to `put(_ data: Data)` vs `putData(_ data: Data)` call with @wilhuff during the Firestore review process, and I think `putData(_ data: Data)` is cleaner and more explicit than a simple `put` call.

FWIW @maksymmalyhin I initially argued the same thing (`put` without `Data`) when we first started discussing it 😄

paulb777: Done

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
paulb777: Done

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
paulb777: Done

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
paulb777: Done

paulb777: Done

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
paulb777: reviewed and updated all fulfill's after verifying timeouts on fail cases.  I'll address the original tests in a separate PR.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
paulb777: Done.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,112 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import FirebaseStorage
+
+/// getResultCallback generates a closure that returns a Result type from a closure that returns an optional type and Error.
```
wilhuff: Please fix line length. go/swift-style calls for 100 cols.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,112 @@
...
+      }
+      completion(.failure(error))
+      return
+    }
+    completion(.success(metadata))
+  }
+}
+
+public extension StorageReference {
+  func downloadURL(completion: @escaping (Result<URL, Error>) -> Void) {
```
wilhuff: These methods should have comments on par with (or likely copies of) those found in the main class, e.g. https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/Public/FIRStorageReference.h#L194.

Feel free to address in a follow-up PR.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,620 @@
...
+          }
+        }
+        waitForExpectations()
+      } catch {
+        XCTFail("Error thrown setting up files in setUp")
+      }
+    }
+  }
+
+  override func tearDown() {
```
wilhuff: Is this actually required? I suspect Swift will destroy the test case instance naturally.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -0,0 +1,620 @@
...
+            case let .failure(error):
+              XCTFail("Unexpected error \(error) from putData")
+            }
+            setupExpectation.fulfill()
+          }
+        }
+        for emptyFile in emptyFiles {
+          let ref = storage.reference().child(emptyFile)
+          _ = ref.putData(data) { result in
+            switch result {
```
wilhuff: You could cut down on the verbosity of these tests by creating assertion helpers. Something like

```swift
func assertResultNotNil<T>(
    _ result: Result<T, Error>,
     file: StaticString = #file, line: UInt = #line) {
  switch result {
    case let .success(value):
      XCTAssertNotNil(value, file: file, line: line);
    case let .failure(error):
      XCTFail("Unexpected error \(error)")
  }
}
```

Now this whole mess can just be:

```swift
_ = ref.putData(data) { result in
  assertResultNotNil(result);
  setupExpectation.fulfill();
}
```

There are a few variations in here (equality, failure happened) but it seems like this would make these tests much more readable.

FirebaseStorageSwift.podspec
```diff
@@ -0,0 +1,44 @@
+Pod::Spec.new do |s|
+  s.name                    = 'FirebaseStorageSwift'
+  s.version                 = '0.1.0'
+  s.summary                 = 'Swift Extensions for Google Cloud Storage for iOS'
```
morganchen12: nit: maybe omit `for iOS`, since this supports a bunch of non-iOS platforms too

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
morganchen12: As of 2016, this was required to release resources since test cases were never deallocated. I'm not sure if this has changed since then. https://twitter.com/benjaminencz/status/758792216022126592

wilhuff: Makes sense, though to actually release the associated app, don't you need to call `-[FIRApp deleteApp]`? Without that `app = nil` seems misleading at best.

Firestore keeps a map of instances around associated with each app/database pair. I can't find equivalent code in [`FIRStorageComponent`](https://github.com/firebase/firebase-ios-sdk/blob/b6cd99e30a9ad87d3f933bfef62ec4d7c0d78cd5/FirebaseStorage/Sources/FIRStorageComponent.m). So I suspect clearing `storage` may actually have an effect.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
paulb777: Added assertResultSuccess to get the most common case.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
paulb777: Will leave this consistent with the other two integration test variations for now.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
wilhuff: Screen Shot 2020-03-30 at 4 21 56 PM

For the scheme you're using, click one of the elements under Source Editor, hit cmd-a to select all, then select the font you want to have for all elements. For myself, I find color helpful, but difference in font unhelpful so I set them all the same.

paulb777: Thanks! I made the more minimal change of only changing Documentation Markup to SF Mono Regular 
Screen Shot 2020-03-30 at 4 39 45 PM

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,113 @@
...
+
+import FirebaseStorage
+
+/// getResultCallback generates a closure that returns a Result type from a closure that returns an
+/// optional type and Error.
+private func getResultCallback<T>(
+  completion: @escaping (Result<T, Error>) -> Void
+) -> (_: T?, _: Error?) -> Void {
+  return { (metadata: T?, error: Error?) -> Void in
+    guard let metadata = metadata else {
```
schmidt-sebastian: Optional: This might  be easier to follow if you split this into two guard statement that are not nested.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,113 @@
...
+  func getData(maxSize: Int64, completion: @escaping (Result<Data, Error>) -> Void)
+    -> StorageDownloadTask {
+    return getData(maxSize: maxSize, completion: getResultCallback(completion: completion))
+  }
+
+  func getMetadata(completion: @escaping (Result<StorageMetadata, Error>) -> Void) {
+    getMetadata(completion: getResultCallback(completion: completion))
+  }
+
+  func list(withMaxResults maxResults: Int64,
```
schmidt-sebastian: `withMaxResults` should have been `maxResults` to begin with. If we do want to introduce some inconsistency, this might be a good candidate :)

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
paulb777: Good idea! This is an example that is better without using `guard` statements - with a plain old if then elseif, else, ....

Done.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
paulb777: I'll leave this to the API reviewers ...

schmidt-sebastian: Leaning towards fixing this here since it matches `getData(maxSize)` and will end up breaking fewer people when we do fix it in the upstream library.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -27,8 +27,9 @@ private func getResultCallback<T>(
     } else {
       completion(.failure(NSError(domain: "FirebaseStorageSwift",
                                   code: -1,
-                                  userInfo: ["Storage Result Generator":
-                                    "InternalError - Return type and Error code both nil"])))
+                                  userInfo: ["NSLocalizedDescription":
```
morganchen12: nit: Instead of using a string literal, use [`NSLocalizedDescriptionKey`](https://developer.apple.com/documentation/foundation/nslocalizeddescriptionkey).

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
```diff
@@ -0,0 +1,113 @@
+// Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
paulb777: Done. 

We can reorganize to a different file structure when we grow. Nothing in the file depends on the filename - everything public is an extension.

FirebaseStorageSwift/Sources/SwiftApiExtension.swift
paulb777: After discussion with @ryanwilson, we decided to make the change all at once, since it would be confusing to see the different labels in Xcode completion:

Screen Shot 2020-03-31 at 3 27 03 PM

PR #5241, opened 2020-04-01
Add API docs for FirebaseStorageSwift

paulb777:
Follows recommended formatting from https://nshipster.com/swift-documentation/ and kept consistent with FirebaseFirestoreSwift.

Content is adapted from https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/Public/FIRStorageReference.h

Looks like this in Xcode:

Screen Shot 2020-04-01 at 10 24 25 AM

### DIFF
--- a/FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
+++ b/FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
@@ -42,7 +42,7 @@ public extension StorageReference {
 
    - Parameters:
      - completion: A completion block returning a Result enum with either a URL or an NSError.
-  */
+   */
   func downloadURL(completion: @escaping (Result<URL, Error>) -> Void) {
     downloadURL(completion: getResultCallback(completion: completion))
   }
@@ -58,7 +58,7 @@ public extension StorageReference {
                    an NSError.
 
    - Returns: A StorageDownloadTask that can be used to monitor or manage the download.
-  */
+   */
   func getData(maxSize: Int64, completion: @escaping (Result<Data, Error>) -> Void)
     -> StorageDownloadTask {
     return getData(maxSize: maxSize, completion: getResultCallback(completion: completion))
@@ -70,7 +70,7 @@ public extension StorageReference {
    - Parameters:
      - completion: A completion block which returns a Result enum with either the
                    object metadata or an NSError.
-  */
+   */
   func getMetadata(completion: @escaping (Result<StorageMetadata, Error>) -> Void) {
     getMetadata(completion: getResultCallback(completion: completion))
   }
@@ -92,7 +92,7 @@ public extension StorageReference {
      - completion A completion handler that will be invoked with the next items and
                   prefixes under the current StorageReference. It returns a Result enum
                   with either the list or an NSError.
-  */
+   */
   func list(withMaxResults maxResults: Int64,
             pageToken: String,
             completion: @escaping (Result<StorageListResult, Error>) -> Void) {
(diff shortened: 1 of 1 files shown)


FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -35,19 +35,64 @@ private func getResultCallback<T>(
 }
 
 public extension StorageReference {
+  /**
```
wilhuff, line 38: I suspect you'll get the same Xcode auto-formatting benefit even if you used the more traditional leading-stars form:

```swift
/**
 * Text
 */
```

This is much more customary and aligns with the rest of the code we already have.

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -56,16 +101,60 @@ public extension StorageReference {
...
   }
 
+  /**
+    List up to `maxResults` items (files) and prefixes (folders) under this StorageReference.
+
+    "/" is treated as a path delimiter. Firebase Storage does not support unsupported object
+    paths that end with "/" or contain two consecutive "/"s. All invalid objects in GCS will be
+    filtered.
+
+   Only available for projects using Firebase Rules Version 2.
```
morganchen12, line 111: Maybe specify what happens if you call this (and the other list function) on a project using rules v1.

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
paulb777: I started with the leading stars format and that didn't work, but the three-slash doc style does. Converted to that.

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
paulb777: Out of scope for this PR. The same open question applies to the original list API docs.

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -35,64 +35,56 @@ private func getResultCallback<T>(
...
-                      greater than 0 and at most 1000.
-     - pageToken A page token from a previous call to list.
-     - completion A completion handler that will be invoked with the next items and
-                  prefixes under the current StorageReference. It returns a Result enum
-                  with either the list or an NSError.
-   */
+  ///  Resumes a previous `list` call, starting after a pagination token.
+  ///  Returns the next set of items (files) and prefixes (folders) under this StorageReference.
+  ///
+  /// "/" is treated as a path delimiter. Firebase Storage does not support unsupported object
```
morganchen12, line 75: nit: two spaces after `///`

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -101,60 +93,54 @@ public extension StorageReference {
...
-                      greater than 0 and at most 1000.
-     - completion A completion handler that will be invoked with the next items and
-                  prefixes under the current StorageReference. It returns a Result enum
-                  with either the list or an NSError.
-   */
+  ///   List up to `maxResults` items (files) and prefixes (folders) under this StorageReference.
+  ///
+  ///   "/" is treated as a path delimiter. Firebase Storage does not support unsupported object
+  ///   paths that end with "/" or contain two consecutive "/"s. All invalid objects in GCS will be
+  ///   filtered.
```
morganchen12, line 100: nit: two spaces after `///`

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -164,19 +150,17 @@ public extension StorageReference {
...
-              the upload.
-   */
+  ///   Asynchronously uploads a file to the currently specified FIRStorageReference.
+  ///
+  ///   - Parameters:
+  ///     - from A URL representing the system file path of the object to be uploaded.
+  ///     - metadata StorageMetadata containing additional information (MIME type, etc.)
+  ///                about the object being uploaded.
+  ///     - completion A completion block that returns a Result enum with either the
+  ///                  object metadata or an NSError.
```
morganchen12, line 160: nit: two spaces after `///`

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
paulb777: @morganchen12 Should it be one space or two? It looks like Firestore has one?

morganchen12: I said two here because it seemed to be the most common across this header file. We haven't standardized; using one to be consistent with FirestoreSwift makes sense.

wilhuff: As far as I'm aware, there should be one space between the three slashes and any text at the outer-most indentation level. This is what we did in `FirebaseFirestoreSwift`.

I can't say I really care too much, but at least let's be consistent. In this file there are one space, two space, and three leading space variations.

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -14,8 +14,8 @@
 
 import FirebaseStorage
 
-/// getResultCallback generates a closure that returns a Result type from a closure that returns an
-/// optional type and Error.
+///  getResultCallback generates a closure that returns a Result type from a closure that returns an
```
wilhuff, line 17: Start function comments with an implicit "this method ...", i.e.

> Generates a closure that ...

FirebaseStorageSwift/Sources/SwiftAPIExtension.swift
```diff
@@ -14,8 +14,14 @@
...
 import FirebaseStorage
 
-/// getResultCallback generates a closure that returns a Result type from a closure that returns an
-/// optional type and Error.
+/// Generates a closure that returns a Result type from a closure that returns an optional type and
+/// Error.
+///
+/// - Parameters:
+///   - completion: A completion block returning a Result enum with either a generic object or
+///                 an NSError.
```
wilhuff, line 22: Code references (like `NSError`) really should be in backticks, but I can see that this file is no worse off than the original storage headers.

In Swift shouldn't we substitute `Error` for `NSError`?

paulb777:
Thanks for the detailed feedback. It should be addressed now. Going to merge this one now. I'm happy to address any additional feedback in a subsequent PR.

PR #5242, opened 2020-04-01
WIP Run cron jobs only for public repo

maksymmalyhin:


paulb777:
@maksymmalyhin Close this one?

maksymmalyhin:
May return to this if we hit GHA quota. Let's keep configs more readable so far.

PR #5244, opened 2020-04-01
Use GutHub Actions token to run Danger

maksymmalyhin:


maksymmalyhin:
Configuring Danger for forks may be revisited after https://github.com/danger/danger/issues/1103 fix.

PR #5247, opened 2020-04-01
add missing return when delete token handler is nil

chliangGoogle:
This fixes crash if passing a nil handler


Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -367,6 +367,7 @@ - (void)deleteTokenWithAuthorizedEntity:(NSString *)authorizedEntity
   if (!handler) {
     FIRInstanceIDLoggerError(kFIRInstanceIDMessageCodeInstanceID001,
                              kFIRInstanceIDInvalidNilHandlerError);
+    return;
```
ryanwilson, line 370: Do we still want to delete the token if there's no completion handler, or leave it as is?

chliangGoogle: the API interface says it is nonnull here so it is supposed to return if passing anil.

PR #5259, opened 2020-04-02
Enable mac catalyst config for GDT and GDTCCT ios target

doudounan:
Enable mac catalyst config for GDT and GDTCCT ios target

#no-changelog


paulb777:
If you haven't already, see https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/test_catalyst.sh that is being used in some of the other workflows.

So far it is working for building for Catalyst and testing for test_specs that don't have `requires_app_host`.

mikehaney24:
This seems to break generating the pod: https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/670762436?utm_medium=notification&utm_source=github_status

I think we need to do this in generate_project. It's fine to abandon this PR for now in favor of working on extensions.

PR #5265, opened 2020-04-02
Upgrade clang-format

paulb777:
Version 9 2019-05-14 c13eda8

#no-changelog

### DIFF
--- a/Firestore/core/src/firebase/firestore/util/filesystem_posix.cc
+++ b/Firestore/core/src/firebase/firestore/util/filesystem_posix.cc
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-#include "Firestore/core/src/firebase/firestore/util/filesystem.h"
-
 #include <dirent.h>
 #include <pwd.h>
 #include <sys/stat.h>
@@ -27,6 +25,7 @@
 #include <deque>
 #include <string>
 
+#include "Firestore/core/src/firebase/firestore/util/filesystem.h"
 #include "Firestore/core/src/firebase/firestore/util/hard_assert.h"
 #include "Firestore/core/src/firebase/firestore/util/path.h"
 #include "Firestore/core/src/firebase/firestore/util/statusor.h"
--- a/Crashlytics/Crashlytics/Components/FIRCLSBinaryImage.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSBinaryImage.h
@@ -14,12 +14,12 @@
 
 #pragma once
 
+#include <stdbool.h>
+#include <stdint.h>
+
 #include "FIRCLSFeatures.h"
 #include "FIRCLSFile.h"
 #include "FIRCLSMachO.h"
-
-#include <stdbool.h>
-#include <stdint.h>
 
 __BEGIN_DECLS
 
--- a/Crashlytics/Crashlytics/Components/FIRCLSHost.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSHost.h
@@ -14,10 +14,10 @@
 
 #pragma once
 
-#include "FIRCLSFile.h"
-
 #include <mach/vm_types.h>
 #include <sys/cdefs.h>
+
+#include "FIRCLSFile.h"
 
 typedef struct {
   const char* documentDirectoryPath;
--- a/Crashlytics/Crashlytics/Components/FIRCLSProcess.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSProcess.h
@@ -14,10 +14,10 @@
 
 #pragma once
 
-#include "FIRCLSFile.h"
-
 #include <mach/mach.h>
 #include <stdbool.h>
+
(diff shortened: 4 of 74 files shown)


Firestore/core/src/firebase/firestore/util/filesystem_posix.cc
```diff
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-#include "Firestore/core/src/firebase/firestore/util/filesystem.h"
```
wilhuff, line 17: This change is wrong. The `filesystem.h` header is the the header for this `.cc` file and it should be first before system headers. There are similar concerns for `log.h`, `secure_random.h`, `status.h`, `filter.h`, and others.

It seems like clang-format is now being much more aggressive about re-grouping headers where previously it only sorted within groups and left grouping alone.

paulb777: Are the other changes useful?

Do you have an opinion about either trying to find the right option combination to fix or staying on the old version?

wilhuff: Looks like we need to specify an alternative `IncludeIsMainRegex` in `.clang-format` to preserve the existing grouping.

Maybe `"(_[[:alnum:]_]+)?$"`?

paulb777: That actually adds one more file change.  I'll do some more experimentation ...
```

$ git diff
diff --git a/.clang-format b/.clang-format
index 7fbcb06cb..1222cf96e 100644
--- a/.clang-format
+++ b/.clang-format
@@ -6,3 +6,4 @@ ObjCSpaceBeforeProtocolList: true
 SpacesInContainerLiterals: true
 PointerAlignment: Right
 AllowShortFunctionsOnASingleLine: None
+IncludeIsMainRegex: (_[[:alnum:]_]+)?$
diff --git a/Firestore/fuzzing/serializer_fuzzer.cc b/Firestore/fuzzing/serializer_fuzzer.cc
index 1dd456fe0..bc6fbf995 100644
--- a/Firestore/fuzzing/serializer_fuzzer.cc
+++ b/Firestore/fuzzing/serializer_fuzzer.cc
@@ -14,6 +14,8 @@
  */

+#include "Firestore/core/src/firebase/firestore/remote/serializer.h"
+
 #include <cstddef>
 #include <cstdint>

@@ -21,7 +23,6 @@
 #include "Firestore/core/src/firebase/firestore/model/database_id.h"
 #include "Firestore/core/src/firebase/firestore/nanopb/message.h"
 #include "Firestore/core/src/firebase/firestore/nanopb/reader.h"
-#include "Firestore/core/src/firebase/firestore/remote/serializer.h"

 using firebase::firestore::google_firestore_v1_Value;
 using firebase::firestore::model::DatabaseId;
```

wilhuff: That change is actually correct.

wilhuff: You probably need to revert the source files to their state on master to see any difference.

paulb777: But it still makes all the other changes.

This fix looks relevant - https://reviews.llvm.org/D66650 - but not yet available via brew. Should we wait?

paulb777: 3252  [2020-04-02 11:50:54 -0700] git checkout master -- Firestore
 3253  [2020-04-02 11:50:58 -0700] git checkout master -- Crashlytics/
 3254  [2020-04-02 11:51:05 -0700] git status
 3255  [2020-04-02 11:51:36 -0700] git restore --staged Firestore
 3256  [2020-04-02 11:51:40 -0700] git restore --staged Crashlytics/
 3257  [2020-04-02 11:51:43 -0700] git status
 3258  [2020-04-02 11:51:50 -0700] ./scripts/style.sh
 3259  [2020-04-02 11:53:40 -0700] git status
 3260  [2020-04-02 11:53:53 -0700] git diff

wilhuff: In general it did a good job pulling out main includes when we had forgotten to do it, reordering system headers in a few places, and cleaned up pointer dereferences from the results of Objective-C methods. Those are all good changes. It's the cases where it's folding in the main include that are the problem. 

Maybe setting `IncludeBlocks` to `preserve` would avoid this? We could revert the problem files, leave the other changes, and then still upgrade. If you want I can take a stab at adding that to this PR.

paulb777: Yes please.

wilhuff: I've pushed onto this PR. `IncludeBlocks: Preserve` did the trick.

PR #5266, opened 2020-04-02
Crashlytics collect on-device symbolication for GDT

samedson:
- On-device symbolication happens upon next startup, not during crash. These files are unique because they end in `.symbolicated` instead of `.clsrecord`. We need to collect and upload these files for GDT in addition to the `.clsrecord` files


Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -193,6 +193,9 @@ - (google_crashlytics_FilesPayload)protoFilesPayload {
     if ([filename.pathExtension.lowercaseString isEqualToString:@"clsrecord"]) {
       [clsRecords addObject:[self.folderPath stringByAppendingPathComponent:filename]];
     }
+    if ([filename.pathExtension.lowercaseString isEqualToString:@"symbolicated"]) {
```
jasonhu-g, line 196: Add an OR clause to the if statement above?

samedson: Done

PR #5273, opened 2020-04-02
FirebasePod, SymbolCollision and Interop tests to GHA

paulb777:
Migrate 3 sets of jobs from travis to GHA

- Firebase Pod tests
- Symbol Collision tests
- Interop pod lib lint tests

#no-changelog


.github/workflows/symbolcollision.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Prereqs
+      run: scripts/install_prereqs.sh FirebasePod iOS
+    - name: Build
+      run: scripts/build.sh FirebasePod iOS
```
ryanwilson, line 25: Does this need to be `SymbolCollision` instead of `FirebasePod`?

.github/workflows/firebasepod.yml
```diff
@@ -0,0 +1,24 @@
+name: firebasepod
+
```
ryanwilson, line 2: Nit: can you please add a comment here about what exactly it's testing?

.github/workflows/symbolcollision.yml
paulb777: Yes. Thanks!

.github/workflows/firebasepod.yml
paulb777: Done

PR #5279, opened 2020-04-03
Collect reports refetch

samedson:
- When collect_reports is false, Crashlytics currently doesn't start up at all. We want to continue to fetch settings in this situation, but never upload crash reports

### DIFF
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
@@ -326,12 +326,6 @@ static void (^reportSentCallback)(void);
     return [FBLPromise resolvedWith:@NO];
   }
 
-  if (![self.settings crashReportingEnabled]) {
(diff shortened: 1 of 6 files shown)


Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -621,11 +618,11 @@ - (void)submitReport:(FIRCLSInternalReport *)report
   [self didSubmitReport];
 }
 
-- (void)removeReport:(FIRCLSInternalReport *)report {
-  [_fileManager removeItemAtPath:report.path];
-}
+// This is the side-effect of calling deleteUnsentReports, or collect_reports setting
+// being galse
```
jasonhu-g: typo

samedson: Fixed 🤣

PR #5281, opened 2020-04-03
Update Firestore changelog for M68

morganchen12:


Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+# v1.11.3
+- [changed] Changed the default log level to `notice` instead of `warning` (#5196).
```
wilhuff, line 4: This only applies to non-Apple-platform users so it doesn't belong in this changelog.

Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+- [added] The Firestore iOS SDK can now be built via CMake (#5052).
```
wilhuff, line 5: This isn't supported. We have a CMake build to make it possible for us to combine this code with the C++ SDK, but that's an implementation detail of the C++ SDK, not a supported usage path for Apple platform users.

morganchen12: Removed.

Firestore/CHANGELOG.md
morganchen12: Removed.

Issue #5282, opened 2020-04-03
Undefined symbols for architecture armv7: "GULAppEnvironmentUtil"


SotoiGhost:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.3.1
  * Firebase SDK version: 6.21.0
  * Firebase Component: InAppMessaging
  * Component version: 0.19.1
  * Installation method: `CocoaPods` (`cocoapods-rome` plugin)

### [REQUIRED] Step 2: Describe the problem

When building the framework I get the following error:

```
Undefined symbols for architecture armv7:
  "_OBJC_CLASS_$_GULAppEnvironmentUtil", referenced from:
      objc-class-ref in FIRInAppMessaging+Bootstrap.o
ld: symbol(s) not found for architecture armv7
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

#### Steps to reproduce:

What happened? How can we make the problem occur?
Unzip the file and run `make` (`cocoapods-rome` plugin must be installed to build).

[InAppMessaging.zip](https://github.com/firebase/firebase-ios-sdk/files/4429945/InAppMessaging.zip)

SotoiGhost:
Workaround for now. Add these lines to your Podfile file:

```
post_install do |installer|
	installer.pods_project.targets.each do |pod|
		if pod.name == "FirebaseInAppMessaging"
			pod.build_configurations.each do |config|
				if config.name == 'Release'
					puts "Linking missing 'GoogleUtilities' framework to #{pod.name}"
					config.build_settings['OTHER_LDFLAGS'] ||= ['$(inherited)','-framework "GoogleUtilities"']
				end
			end
		end
	end
end
```

paulb777:
Thanks for sending the fix and workaround along with  the bug!

It's merged to master and is targeted for the M69-6.23.0 release.

PR #2312, opened 2019-01-25
Open source FIAM headless SDK

christibbs:
Merges intermediate branch `iam-staging` (containing files for headless SDK) into master.

PR #5284, opened 2020-04-03
Explicitly state InAppMessaging's dependency on GoogleUtilities

SotoiGhost:
In #2312 you added InAppMessaging to the repo and [here][1] you check if this is running on simulator using `GoogleUtilities`.

This adds `GoogleUtilities` explicitly to `FirebaseInAppMessaging.podspec` fixing #5282

[1]: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInAppMessaging/Sources/Runtime/FIRInAppMessaging%2BBootstrap.m#L105


paulb777:
Thanks for the fix!  LGTM. We'll merge once CI passes.

paulb777:
cc: @christibbs

PR #5301, opened 2020-04-06
Add a new firebase messaging test app using swiftui

chliangGoogle:
This will replace the old app and only has non-MCS functions. 
We start with basic functions that enable push notifications and a few buttons to test add/delete iid/fid/token flow.


Example/Messaging/Sample/Sample/AppDelegate.swift
```diff
@@ -0,0 +1,91 @@
...
+
+  // MARK: UISceneSession Lifecycle
+
+  func application(_ application: UIApplication, configurationForConnecting connectingSceneSession: UISceneSession, options: UIScene.ConnectionOptions) -> UISceneConfiguration {
+    // Called when a new scene session is being created.
+    // Use this method to select a configuration to create the new scene with.
+    return UISceneConfiguration(name: "Default Configuration", sessionRole: connectingSceneSession.role)
+  }
+
+  // MARK: - Core Data stack
```
morganchen12, line 45: nit: is the Core Data stack necessary?

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+import FirebaseInstallations
+
+struct ContentView: View {
+  @EnvironmentObject var identity: Identity
+  
+  var body: some View {
+    NavigationView {
+      List {
+        Button(action: getToken) {
+          HStack{
```
morganchen12, line 28: nit: Space between `HStack` and `{` (throughout file)

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+          .foregroundColor(.blue)
+        Text("Token: \(identity.token)")
+          .foregroundColor(.blue)
+        NavigationLink(destination: DetailView()) {
+          Text("Show Detail View")
+        }
+      }
+    }
+  }
+  func getToken() {
```
morganchen12, line 84: nit: newline before method declaration

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+      if result != nil && error == nil {
+        self.identity.token = result?.token ?? ""
+        self.identity.instanceID = result?.instanceID ?? ""
+      }
+    }
+  }
+  
+  func deleteToken() {
+    Messaging.messaging().deleteFCMToken(forSenderID: FirebaseApp.app()?.options.gcmSenderID ?? "") { (error) in
+      if (error != nil) {
```
morganchen12, line 95: nit: if statement doesn't require parens

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+    InstanceID.instanceID().instanceID { (result, error) in
+      if result != nil && error == nil {
+        self.identity.token = result?.token ?? ""
+        self.identity.instanceID = result?.instanceID ?? ""
+      }
+    }
+  }
+  
+  func deleteToken() {
+    Messaging.messaging().deleteFCMToken(forSenderID: FirebaseApp.app()?.options.gcmSenderID ?? "") { (error) in
```
morganchen12, line 94: Instead of using a nil coalesce here, this should be broken up into two statements.

```swift
guard let senderID = FirebaseApp.app()?.options.gcmSenderID else { return /* (or fatalError) */ }
Messaging.messaging().deleteFCMToken(forSenderID: senderID) { error in
  // ...
}
```

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+      }
+    }
+  }
+  func getToken() {
+    InstanceID.instanceID().instanceID { (result, error) in
+      if result != nil && error == nil {
+        self.identity.token = result?.token ?? ""
+        self.identity.instanceID = result?.instanceID ?? ""
+      }
```
morganchen12, line 89: Consider handling errors in this callback

Example/Messaging/Sample/Podfile
```diff
@@ -0,0 +1,18 @@
+use_frameworks!
+
+target 'Sample' do
+  platform :ios, '9.0'
```
paulb777, line 4: SwiftUI needs 13.0

Example/Messaging/Sample/Sample/SceneDelegate.swift
```diff
@@ -0,0 +1,68 @@
...
+  
+  func sceneDidEnterBackground(_ scene: UIScene) {
+    // Called as the scene transitions from the foreground to the background.
+    // Use this method to save data, release shared resources, and store enough scene-specific state information
+    // to restore the scene back to its current state.
+
+    // Save changes in the application's managed object context when the application transitions to the background.
+    (UIApplication.shared.delegate as? AppDelegate)?.saveContext()
+  }
+
```
paulb777, line 65: extra spaces

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,139 @@
...
+          .foregroundColor(.blue)
+        NavigationLink(destination: DetailView()) {
+          Text("Show Detail View")
+        }
+      }
+    }
+  }
+  func getToken() {
+    InstanceID.instanceID().instanceID { (result, error) in
+      if result != nil && error == nil {
```
maksymmalyhin, line 86: Please consider using `if let result  = result` or `guard let result = result else` to unwrap optional variables.

Example/Messaging/Sample/Sample/ContentView.swift
maksymmalyhin: I wonder why it is not fixed by `./scripts/style.sh`?

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: Done. Sorry forgot to run clang/format

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: Done.

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: Sorry I didn't run it. Updating now.

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,146 @@
...
+        NavigationLink(destination: DetailView()) {
+          Text("Show Detail View")
+        }
+      }
+    }
+  }
+
+  func getToken() {
+    InstanceID.instanceID().instanceID { result, error in
+      if error == nil {
```
maksymmalyhin, line 87: This looks like a typo. Did you mean something like
```
if let error = error {
  print(error)
  return
}
```
or 
```
guard error == nil else {
  print()
  return
}
```

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -81,40 +81,47 @@ struct ContentView: View {
...
-    InstanceID.instanceID().instanceID { (result, error) in
-      if result != nil && error == nil {
-        self.identity.token = result?.token ?? ""
-        self.identity.instanceID = result?.instanceID ?? ""
+    InstanceID.instanceID().instanceID { result, error in
+      if error == nil {
+        print("Failed getting iid and token: ", error ?? "")
+        return
       }
+      guard let result = result else {
```
maksymmalyhin, line 91: I think it can even be combined with lines 87-89 like:

```
guard 
  let let result = result,
  error == nil
else {
  print()
  return
}
```

Example/Messaging/Sample/Sample/AppDelegate.swift
morganchen12: That's just boilerplate generated by Xcode if you check the "Use Core Data" box. If there isn't any data being saved to Core Data, the save calls aren't doing anything and can also be removed.

chliangGoogle:
hmm I don't remember particularly enable "user core data" but they are being called in SceneDelegate to create the context and contentView object and I can't simply remove them, it will break the app.

Example/Messaging/Sample/Sample/AppDelegate.swift
chliangGoogle: Yes, but they are called in sceneDelegate and can't be simply removed. persistentContainer is used to create contentView in sceneDelegate.

chliangGoogle: I can remove saveContext tho.

morganchen12: SGTM, we can revisit this if it's found to be unnecessary in the future.

FirebaseMessaging.podspec
```diff
@@ -63,4 +63,12 @@ device, and it is completely free.
     }
     unit_tests.dependency 'OCMock'
   end
+
+  s.test_spec 'sample' do |build_test|
```
paulb777, line 67: Not necessary to do it this way when you already have a Podfile. You could also follow a pattern like https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/symbolcollision.yml

chliangGoogle: Done.

.github/workflows/messagingsample.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: macos-latest
```
paulb777, line 14: My preference would be adding another job in the existing messaging.yml file

chliangGoogle: You can only have one name and I wonder if I can just merge jobs and folders under one name. let me try that.

paulb777: Each job can be separately named

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,143 @@
...
+          HStack {
+            Image(systemName: "arrow.clockwise.circle.fill").font(.body)
+            Text("Get ID and Token")
+              .fontWeight(.semibold)
+          }
+          .padding(5)
+          .foregroundColor(.white)
+          .background(Color.purple)
+          .cornerRadius(40)
+        }
```
ryanwilson, line 37: These look really nice - just a few suggestions here.

There's a lot of duplication here and below, there are two ways I think this could be resolved:

1. Creating a custom button `struct PillButton` or something, that takes in the `title`, `systemImageName`, `backgroundColor`, and `action`. That way the actual layout is only in one place and you can modify it once for all buttons. I think a nice API would be like:

```
PillButton(title: "Get ID and Token",
           systemImageName:"arrow.clockwise.circle.fill",
           backgroundColor: .purple,
           action: getToken)
```
.. and then if it's a short action, it could be added inline:

```
PillButton(title: "Say Hello",
           systemImageName:"arrow.clockwise.circle.fill",
           backgroundColor: .purple) {
  print("Hello!")
}
```

Let me know if you need a hand creating it!

2. You could use a custom [`ButtonStyle`](https://developer.apple.com/documentation/swiftui/buttonstyle), which is really nice. Could let you do something like:

```
struct PillButton: ButtonStyle {
  // implementation...
}

// Usage:
Button("Get ID and Token", action: getToken)
  .buttonStyle(PillButton(icon: "arrow.clockwise.circle.fill", background: .purple))
```

Part of the problem with 2 is you lose the iOS provided pressing animation (going darker to indicate you've pressed it down). You can customize it since the ButtonStyle has a configuration that tells you when it's pressed, but it's more for really custom buttons that wouldn't act like a traditional button (scales when it's pressed, something else).

ryanwilson: Secondly, for the padding it's probably reasonable to keep the default value: `.padding()` that way it's a large enough touch target and it'll choose the right padding based on the platform (just a good practice, I know this isn't on other platforms).

Example/Messaging/Sample/Sample/Identity.swift
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+
+public final class Identity: ObservableObject {
+  @Published public var instanceID = ""
+  @Published public var installationID = ""
+  @Published public var token = ""
```
ryanwilson, line 20: Nit: can you please add some documentation here as to what each one is? Mostly targeting a user who is new to Messaging and has never heard of FIS/IID.

Example/Messaging/Sample/Sample/SceneDelegate.swift
```diff
@@ -0,0 +1,54 @@
...
+  var window: UIWindow?
+  let identity = Identity()
+
+  func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
+    // Use this method to optionally configure and attach the UIWindow `window` to the provided UIWindowScene `scene`.
+    // If using a storyboard, the `window` property will automatically be initialized and attached to the scene.
+    // This delegate does not imply the connecting scene or session are new (see `application:configurationForConnectingSceneSession` instead).
+
+    // Get the managed object context from the shared persistent container.
+    let context = (UIApplication.shared.delegate as! AppDelegate).persistentContainer.viewContext
```
ryanwilson, line 31: As mentioned in the previous comment, this could be deleted along with the `.environment` bit on line 35.

Example/Messaging/Sample/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+1. Go to the [Firebase Console](https://console.firebase.google.com/)
+2. Create a Firebase project if you don't have one already.
+3. Create a new iOS App if you don't have one already.
+4. Go to Project Overview -> General -> Your apps, select your iOS app and download the GoogleSerive-Info.plist file.
+
+
+#### Push notification provisioning profile
+
+If you need to test push notifications sent from FCM console or Sender API, you will need to run this test app on real device. In order to do so, you will need a provisioning profile enabled with push notifications. Following the [steps](https://firebase.google.com/docs/cloud-messaging/ios/certs) here to setup the properly APNs configuration to work with Firebase Messaging.
```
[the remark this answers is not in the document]
chliangGoogle: hmm I did on this file specifically, but no change seems to take effect.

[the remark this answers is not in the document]
maksymmalyhin: I don't think the script fixed .md files now. I ended up formatting .md in google3 (the file has to be a part of google3 doc target for `g4 fix` formats them).

Example/Messaging/Sample/Sample/Identity.swift
chliangGoogle: done. also realize only one ID is used, so remove installationID parameter.

Example/Messaging/Sample/Sample/SceneDelegate.swift
chliangGoogle: Done.

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: nice!

Example/Messaging/Sample/Sample/Identity.swift
chliangGoogle: done.

Example/Messaging/Sample/Sample/SceneDelegate.swift
chliangGoogle: done.

Example/Messaging/Sample/Sample/AppDelegate.swift
chliangGoogle: Done. I wonder what the use case this is. For messaging to update the data object identity, we need to implement MessagingDelegate in SceneDelegate. I wonder with this context, we can just do it in appDelegate.

Example/Messaging/Sample/README.md
paulb777, line 15: Please split up this line.

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,145 @@
+// Copyright 2020 Google
```
paulb777, line 1: New copyrights should be `Google LLC`. See go/copyright

Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -0,0 +1,145 @@
...
+        print("Failed getting iid and token: ", error ?? "")
+        return
+      }
+      self.identity.token = result.token
+      self.identity.instanceID = result.instanceID
+    }
+  }
+
+  func deleteToken() {
+    guard let senderID = FirebaseApp.app()?.options.gcmSenderID else { return }
```
paulb777, line 86: `guard let app = FirebaseApp.app()` might be cleaner.

Example/Messaging/Sample/Sample/Identity.swift
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 Google
```
paulb777, line 1: LLC here and all other new copyrights

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: Done.

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: Done.

Example/Messaging/Sample/Sample/Identity.swift
chliangGoogle: Done.

Example/Messaging/Sample/README.md
chliangGoogle: Done.

PR #5302, opened 2020-04-07
Append framework name to Carthage plist bundleID

### LINKED ISSUE #5299: Not able to run app in simulator due to framework identifier duplication in Info.plist
### Environment

  * Xcode version: 11.4
  * Firebase SDK version: 6.21.0
  * Firebase Component: Auth 
  * Component version: 6.21.0

#### Steps to reproduce:

After upgrading to the latest version of Firebase and after adding additional frameworks based on research as in the doc is not clear what to do the app compiles just fine, but when trying to run it I get this:

Screen Shot 2020-04-06 at 5 07 05 PM

It seems like the **GTMAppAuth.framework** and **abseil.framework** frameworks are both sharing the same app id in the Info.plist 

However I'm not sure if it's me doing something wrong or what.

Please see the 2 attachments below that shows the same id:

Screen Shot 2020-04-06 at 5 12 29 PM
Screen Shot 2020-04-06 at 5 12 05 PM

Thank you!

ryanwilson:
Addresses a symptom of #5299

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
@@ -249,7 +249,7 @@ extension CarthageUtils {
   }
 
   static func generatePlistContents(forName name: String) -> Data {
-    let plist: [String: String] = ["CFBundleIdentifier": "com.firebase.Firebase",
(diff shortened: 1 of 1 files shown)


ryanwilson:
Note: still need to test this :) but should do the trick.

paulb777:
So while it fixes one of the symptoms of #5299, I don't think it will actually fix it.

PR #5307, opened 2020-04-07
Merge master into gdtsqlite

mikehaney24:
Merges master and fixes issues with doing that. Will aid in cherry-picking the storage that converts the current storage API to a protocol.

There are a couple key changes, they're in the commits at the end of this PR:
- ca16d27 which addresses a race condition that was newly exposed (and tsan didn't catch)
- 6b58b06 which converts usage of a prior API customPrioritizationDict to customBytes and adds a category for easier implementation

**This PR will be merged as a normal git merge, not squashed!**


mikehaney24:
Since this was a master merge, all CLA requirements have already been met.

mikehaney24:
> Have you considered rebasing master instead of merging? If it works, it can be a lot more readable.

Frankly, I had forgotten that option. I'll keep that in mind for future feature branches.

PR #5310, opened 2020-04-08
Implement limbo resolution listen throttling

dconeybe:
This is a port of https://github.com/firebase/firebase-js-sdk/pull/2790 and https://github.com/firebase/firebase-android-sdk/pull/1374 to iOS.

The following description was copied from the first of those two PRs and pasted here for your convenience:

This change was motivated by the following bug report: https://github.com/firebase/firebase-js-sdk/issues/2683.  In summary, the issue was that when a large number of documents went into limbo (in this case 15,000 documents) then this would cause 15,000 document listens, which would result in "resource-exhausted" errors.  The client would then back off for a bit but would try the listens again, which would again result in "resource-exhausted" errors.  And the cycle would continue.  Worst of all, the customer was being billed for all of these reads that had no beneficial effects for the clients.

In order to avoid this situation, this PR modifies the limbo resolution logic to "throttle" the limbo resolutions.  It does this by allowing at most 100 concurrent limbo resolutions.  Limbo resolutions in excess of this limit are queued up and not started until another limbo resolution completes.  This fix should avoid the "resource-exhausted" errors and the infinite read loop.


Firestore/core/src/firebase/firestore/core/sync_engine.h
```diff
@@ -231,6 +240,16 @@ class SyncEngine : public remote::RemoteStoreCallback, public QueryEventSource {
 
   void TrackLimboChange(const LimboDocumentChange& limbo_change);
 
+  /**
```
wu-hui, line 243: Ultranit: I understand this is copied from other platforms. I'd still suggest to mention max_concurrent_limbo_resolutions_ in the comment.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
```diff
@@ -526,32 +529,42 @@ void SyncEngine::UpdateTrackedLimboDocuments(
...
-    Query query(key.path());
-    TargetData target_data(query.ToTarget(), limbo_target_id,
-                           kIrrelevantSequenceNumber,
-                           QueryPurpose::LimboResolution);
-    limbo_resolutions_by_target_.emplace(limbo_target_id, LimboResolution{key});
-    remote_store_->Listen(target_data);
-    limbo_targets_by_key_[key] = limbo_target_id;
+    active_limbo_resolutions_by_target_.emplace(limbo_target_id,
+                                                LimboResolution{key});
+    active_limbo_targets_by_key_[key] = limbo_target_id;
```
wu-hui, line 549: Is there special reason why `[]` is used here as opposed to `emplace` you used in the above line?

If not, we should be consistent here and use `emplace` in both?

dconeybe: So that particular line of code is not new but only changed due to the renaming of the variable `limbo_targets_by_key_` to `active_limbo_targets_by_key_`.  If I understand `emplace()` correctly, there is no benefit to using it over `operator[]` because `limbo_target_id` is just an int and therefore does not benefit from move semantics.  I was also trying to minimize the unrelated changes so if you'd still prefer that I change it to `emplace()` then I will do that in a follow-up PR.

wu-hui: I see. I am OK with a separate PR to make it more consistent.

Firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+- [changed] Firestore now limits the number of concurrent document lookups it
+  will perform when resolving inconsistencies in the local cache (#5310).
```
wilhuff, line 3: This really should link to the issue rather than the PR--it doesn't matter that the issue was in another repository. See below for a what a link looks like to an issue in firebase-js-sdk.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
wilhuff: For what it's worth, `emplace` on a `map` is constructing the `std::pair` representing the key-value pair, so even though the `limbo_target_id` is trivially copyable, the pair is not because the key is a `DocumentKey`.

This is something of an academic concern though: we've structured our code so that vocabulary types like `DocumentKey` are cheaply copyable. They're immutable values that hide a `std::shared_ptr`, making a copy just a reference count bump. In this case `emplace` vs `operator[]` without moving both make a copy so there's essentially no difference (`emplace` with a moved key does save the copy, but that doesn't work here because the key is used below).

In any case, making these consistent is still worthwhile, and feel free to just inline that into this PR. A separate PR just for that small change is more effort than this change is worth.

Firestore/CHANGELOG.md
dconeybe: Done.  Changed the PR number to the link to the issue https://github.com/firebase/firebase-js-sdk/issues/449.

Firestore/core/src/firebase/firestore/core/sync_engine.cc
dconeybe: Done.  It turns out that this was the *only* place where `[]` needed to be changed to `emplace`.

Firestore/core/src/firebase/firestore/core/sync_engine.h
```diff
@@ -90,7 +92,8 @@ class SyncEngine : public remote::RemoteStoreCallback, public QueryEventSource {
  public:
   SyncEngine(local::LocalStore* local_store,
              remote::RemoteStore* remote_store,
-             const auth::User& initial_user);
+             const auth::User& initial_user,
+             size_t maxConcurrentLimboResolutions);
```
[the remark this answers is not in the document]
dconeybe: Oops.  Fixed.  Good catch.

dconeybe:
Note that the ASAN check failure is a known issue with the build system: https://github.com/firebase/firebase-ios-sdk/pull/5316#issuecomment-611112974.  I will ignore that failure and go ahead and merge.

PR #5313, opened 2020-04-08
Update comments on `configure` call.

ryanwilson:
Currently it mentions that the calls are thread-safe. While technically true, `configure` should only be called from the main thread to ensure all SDKs are configured by the time they are used.


ryanwilson:
Removing the `public-api-change` label since it's a comment only change, and the contract is unaffected. The previous comment was inaccurate. No implementation has changed.

ryanwilson:
Cancelled travis.

PR #5315, opened 2020-04-08
Limit Swift line lengths to 100

paulb777:


CoreOnly/Tests/FirebasePodTest/FirebasePodTest/AppDelegate.swift
```diff
@@ -36,20 +39,31 @@ class StorageExists: Storage {}
...
     return true
   }
 
   // MARK: UISceneSession Lifecycle
 
-  func application(_ application: UIApplication, configurationForConnecting connectingSceneSession: UISceneSession, options: UIScene.ConnectionOptions) -> UISceneConfiguration {
+  func application(
+    _ application: UIApplication,
+    configurationForConnecting connectingSceneSession: UISceneSession,
+    options: UIScene.ConnectionOptions
```
ryanwilson, line 55: Not blocking, but can we see afterwards if we could have it wrap like this:

```
func application(_ application: UIApplication,
                 configurationForConnecting connectingSceneSession: UISceneSession,
                 option: UIScene.ConnectionOptions
) -> UISceneConfiguration
```

Seems a bit more readable having the first argument alongside the function name, I realize that wouldn't work with all functions though.

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -205,13 +206,18 @@ struct ZipBuilder {
     }
 
     let zipDir = try assembleDistributions(inProjectDir: projectDir, withPackageKind: "Firebase",
```
ryanwilson, line 208: Hmm should `withPackageKind:` also be wrapped?

CoreOnly/Tests/FirebasePodTest/FirebasePodTest/AppDelegate.swift
paulb777: Good catch!  Many changes look better after adding the option `--wrapparameters afterfirst`

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
paulb777: Seems like it, but I can't find the right `swiftformat`.  It's stable after manually updating.

PR #5316, opened 2020-04-08
Stop using built-in CMake Framework support

wilhuff:
CMake has built-in support for building Framework bundles (briefly: bundles of libraries, associated headers, and resources) but this doesn't work well for our needs.

1. In CMake 3.5.1 (our earliest supported CMake version) it doesn't copy headers properly at build time.

2. Even up to CMake 3.15.5 (the current version I'm testing with), CMake doesn't handle link-line multiplicity correctly. CMake passes the static library location within the framework and this seems to cause the linker to mishandle Objective-C metaclass information resulting in duplicate symbols. Building with `-framework` flags works correctly but there's no way to convince CMake to build this way without considerable additional complexity.

Previously we partially worked around the latter problem by building some frameworks as shared instead of static frameworks, but this is problematic because this mode of operation isn't actually intended and is unsupported as far as the rest of Firebase is concerned. This turned out to be fragile though and while performing an unrelated library refactoring this workaround broke.

Rather than piling on additional workarounds, this change converts the frameworks to simple library targets and emulates framework imports through include paths. This is reasonable because the frameworks we're depending upon just exist to test Firestore on iOS/macOS and aren't a part of the shipping product.

Additionally, this starts refactoring our cmake helper functions to make them more like built-in functions. This makes it possible to freely use different CMake features that were made difficult by trying to wrap everything up in a single CMake command. For example, previously all `target_compile_definitions` added via `firebase_ios_objc_framework`'s `DEFINES` argument were `PUBLIC` and now they can be `PRIVATE` where that's appropriate.

/cc @rafikhan

### DIFF
--- a/Firestore/Source/API/FIRCollectionReference.mm
+++ b/Firestore/Source/API/FIRCollectionReference.mm
@@ -54,6 +54,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 // Override the designated initializer from the super class.
 - (instancetype)initWithQuery:(api::Query &&)query {
+  (void)query;
+
   HARD_FAIL("Use FIRCollectionReference initWithPath: initializer.");
 }
 
--- a/cmake/cc_rules.cmake
+++ b/cmake/cc_rules.cmake
@@ -470,19 +470,21 @@ endfunction()
 #
 # All other arguments are ignored.
 function(firebase_ios_set_common_target_options target)
-  set(bool DISABLE_STRICT_WARNINGS)
-  cmake_parse_arguments(flag "${bool}" "" "" ${ARGN})
+  set(options DISABLE_STRICT_WARNINGS)
+  cmake_parse_arguments(flag "${options}" "" "" ${ARGN})
 
-  set(suffix _STRICT)
   if(flag_DISABLE_STRICT_WARNINGS)
-    set(suffix "")
+    set(cxx_flags ${FIREBASE_IOS_CXX_FLAGS})
+    set(objc_flags ${FIREBASE_IOS_OBJC_FLAGS})
+  else()
+    set(cxx_flags ${FIREBASE_IOS_CXX_FLAGS_STRICT})
+    set(objc_flags ${FIREBASE_IOS_OBJC_FLAGS_STRICT})
   endif()
+  message("objc_flags = ${objc_flags}")
 
-  target_compile_options(${target} PRIVATE ${FIREBASE_IOS_CXX_FLAGS${suffix}})
+  target_compile_options(${target} PRIVATE ${cxx_flags})
   if(APPLE)
-    target_compile_options(
-      ${target} PRIVATE ${FIREBASE_IOS_OBJC_FLAGS${suffix}}
-    )
+    target_compile_options(${target} PRIVATE ${objc_flags})
   endif()
 
   target_include_directories(
--- a/cmake/cc_rules.cmake
+++ b/cmake/cc_rules.cmake
@@ -15,6 +15,60 @@
 include(CMakeParseArguments)
 include(FindASANDylib)
 
+
+# firebase_ios_add_framework(
+#   target
+#   sources...
+# )
+#
+# Wraps a call to add_library (including arguments like EXCLUDE_FROM_ALL) that
+# additionally sets common options that apply to all frameworks in this repo.
+#
+# Note that this does not build an actual framework bundle--even recent versions
+# of CMake (3.15.5 as of this writing) produce problematic output when
+# frameworks are enabled. However `firebase_ios_add_framework` along with
+# `firebase_ios_framework_public_headers` produces a library target that
+# can be compiled against as if it *were* a framework. Notably, imports of the
+# form #import <Framework/Framework.h> and public headers that refer to each
+# other unqualified will work.
+function(firebase_ios_add_framework target)
+  firebase_ios_split_target_options(flag ${ARGN})
+
+  add_library(${target} ${flag_REST})
+  firebase_ios_set_common_target_options(${target} ${flag_OPTIONS})
+
+  target_link_libraries(${target} INTERFACE -ObjC)
+endfunction()
+
+# firebase_ios_framework_public_headers(
+#   target
+#   headers...
+# )
+#
+# Prepares the given headers for use as public headers of the given target.
+# This emulates CocoaPods behavior of flattening public headers by creating
+# a separate directory tree in the build output, consisting of symbolic links to
+# the actual header files. This makes it possible to use imports of these forms:
+#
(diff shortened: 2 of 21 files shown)


wilhuff:
The ASAN build failure reflects a preexisting condition. I have a follow-up PR that fixes this through a refactoring so I'd rather not spend the time to figure out how to make this work in the current state.

cmake/cc_rules.cmake
```diff
@@ -440,24 +408,88 @@ function(firebase_ios_objc_test target)
...
+
+# Sets common options for a given target. This includes:
+#
+#   * `DISABLE_STRICT_WARNINGS`: disables stricter warnings usually applied to
+#     targets in this repo. Use this to compile code that's local but doesn't
+#     hold itself to Firestore's warning strictness.
+#
+# All other arguments are ignored.
+function(firebase_ios_set_common_target_options target)
+  set(bool DISABLE_STRICT_WARNINGS)
```
dconeybe, line 473: Why are you setting a variable named `bool` here?  Couldn't the string `DISABLE_STRICT_WARNINGS` just be inlined into the call to `cmake_parse_arguments()`?  Or is this an attempt to allow for future expansion of `bool` as a list?  IMO the presence of this variable that is only used exactly once is confusing because it begs the question "why is this variable defined?".  If you think it adds value, then consider a more descriptive name than `bool`, such as `one_value_keywords`, to match the argument's name in the documentation for `cmake_parse_arguments()`.

cmake/cc_rules.cmake
```diff
@@ -440,24 +408,88 @@ function(firebase_ios_objc_test target)
...
+  cmake_parse_arguments(flag "${bool}" "" "" ${ARGN})
+
+  set(suffix _STRICT)
```
dconeybe, line 476: To improve readability, consider calling `set(suffix _STRICT)` in an `else` block instead of defining it before the `if` block, then potentially overwriting it.

cmake/cc_rules.cmake
```diff
@@ -440,24 +408,88 @@ function(firebase_ios_objc_test target)
...
+  if(flag_DISABLE_STRICT_WARNINGS)
+    set(suffix "")
+  endif()
+
+  target_compile_options(${target} PRIVATE ${FIREBASE_IOS_CXX_FLAGS${suffix}})
```
dconeybe, line 481: To reduce the code indirection and improve searchability, consider storing the  CXX_FLAGS and OBJC_FLAGS into respective variables in the `if` block above instead of the "suffix" method.

e.g.

```
if(flag_DISABLE_STRICT_WARNINGS)
  set(CXX_FLAGS ${FIREBASE_IOS_CXX_FLAGS})
  set(OBJC_FLAGS ${FIREBASE_IOS_OBJC_FLAGS})
else()
  set(CXX_FLAGS ${FIREBASE_IOS_CXX_FLAGS_STRICT})
  set(OBJC_FLAGS ${FIREBASE_IOS_OBJC_FLAGS_STRICT})
endif()
```

Then below, you can use these variables like this:

```
target_compile_options(
      ${target} PRIVATE ${OBJC_FLAGS}
    )
```

cmake/cc_rules.cmake
```diff
@@ -440,24 +408,88 @@ function(firebase_ios_objc_test target)
...
+# to `add_library`, `add_executable`, or similar.
+#
+# The resulting lists are set in the parent scope as `${prefix}_OPTIONS` for use
+# when calling `firebase_ios_set_common_target_options`, and `${prefix}_REST`
+# that contains any arguments that were unrecognized.
+#
+# See `firebase_ios_set_common_target_options` for details on which arguments
+# are split out.
+function(firebase_ios_split_target_options prefix)
+  set(option_names DISABLE_STRICT_WARNINGS)
```
dconeybe, line 447: Is there a way to define `option_names` as a global variable?  IMO Its presence here incorrectly suggests that it is a temporary variable that is manipulated in the function; however, it is closer to a global, immutable constant that logically makes sense to be defined exactly once.

Firestore/Source/API/FIRCollectionReference.mm
```diff
@@ -54,6 +54,8 @@ - (instancetype)initWithPath:(ResourcePath)path
 
 // Override the designated initializer from the super class.
 - (instancetype)initWithQuery:(api::Query &&)query {
+  (void)query;
```
dconeybe, line 57: Is this line here to avoid an "unused variable" warning? If so, please leave an inline comment to this effect since it is otherwise confusing (to me, at least). Ignore this comment if this method of avoiding "unused variable" warnings is conventional.  This comment applies to a few other files in this PR as well.

wilhuff: This is a fairly common convention in C and C++ (code search for `^\s*\(void\)\w+;` turns up ~36k hits just in google3). This also fairly common advice given on [stack overflow](https://stackoverflow.com/questions/40371426/how-to-suppress-the-compiler-warning-on-unused-variable-in-c). I'm inclined to leave this as-is.

Note: Abseil has an `ABSEIL_ATTRIBUTE_UNUSED` marker that you an use for this kind of thing, but it only has a definition for Clang and GCC and doesn't work on other compilers.

Where possible I've tried to avoid this by just leaving unused parameters unnamed, but Objective-C method declarations require an identifier so you have mark these unused like we're doing here.

cmake/cc_rules.cmake
```diff
@@ -470,19 +470,21 @@ endfunction()
...
-  set(suffix _STRICT)
   if(flag_DISABLE_STRICT_WARNINGS)
-    set(suffix "")
+    set(cxx_flags ${FIREBASE_IOS_CXX_FLAGS})
+    set(objc_flags ${FIREBASE_IOS_OBJC_FLAGS})
+  else()
+    set(cxx_flags ${FIREBASE_IOS_CXX_FLAGS_STRICT})
+    set(objc_flags ${FIREBASE_IOS_OBJC_FLAGS_STRICT})
   endif()
+  message("objc_flags = ${objc_flags}")
```
dconeybe, line 483: Should this message() line be removed?  It looks like it exists for debugging purposes.

cmake/cc_rules.cmake
dconeybe: That sounds good to me.  I agree that keeping it tamper-proof and defined int he function is of more value than defining it as a global constant.

cmake/cc_rules.cmake
dconeybe: SGTM.   The variable rename is a huge improvement.  Thank you for the detailed explanation.

cmake/cc_rules.cmake
wilhuff: Doh. Removed.

PR #5318, opened 2020-04-08
Migrate Auth to canonical directory structure

paulb777:
Mostly just a move and associated renames except for some restructuring of the Sample Xcode project.


paulb777:
Removing public API tag since the danger triggered it from a move - no actual API changes

paulb777:
The host apps in `Example/Auth/App` should all be deleted. Do you see any somewhere else?

PR #5321, opened 2020-04-09
Shorten CMake test targets/begin firebase_core consolidation

wilhuff:
This is the first in a series of PRs that aim to shorten target names and object paths with the goal of making it possible to build the project on Windows when the project root is nested within the firebase-cpp-sdk build, itself located within a source directory at a reasonable depth. By default, Windows limits path names to 255 characters, and our names result in paths that exceed this limit in that circumstance.

The top-level goal is to migrate the large variety of long target names like `firebase_firestore_remote_connectivity_monitor_noop` to things that are shorter, by:

  * Dropping `firebase_` prefixes; "firestore" is unique enough
  * Combining targets, so that they don't need long names to distinguish among them
  * Shortening paths
  * Choosing sources into a single library rather than building separate libraries and choosing the appropriate library afterward.

This PR shortens test target names and establishes the existence of three internal libraries:

  * `firestore_util`, a grouping of all libraries named like `firebase_firestore_util*`
  * `firestore_nanopb`, which contains `firebase_firestore_nanopb_runtime`
  * `firebase_core`, which contains everything else (including `firebase_firestore_remote_connectivity_monitor_noop`)

This PR does not implement those libraries; instead it creates CMake `INTERFACE` libraries that re-export the existing libraries and puts them to use. This simulates the existence of those libraries and adjusts all dependencies. Subsequent PRs will actually do the reimplementation.

Note the diffbase on #5316.


wilhuff:
Ugh. Need to fix the linter. The error is spurious here.

PR #5322, opened 2020-04-09
Merge storage refactor back to master

mikehaney24:


mikehaney24:
> Seems like this PR is a mix or previously reviewed changes and new changes. If it is so, would it be possible to have a separate PR for new changes to review so the feature branch can be merged without review after that?
> 
> Screen Shot 2020-04-09 at 4 02 02 PM

This change is generally all previously reviewed changes (the ones you screenshotted were reviewed in #5307). There _are_ a few  unreviewed changes, but they're usually inconsequential like changes to travis.yml.

PR #5323, opened 2020-04-09
Shorten CMake Objective-C targets

wilhuff:
This is a continuation of the work in #5321, renaming Objective-C and other nonessential targets from the point of view of the Windows build to match the pattern set out there.

The intent is that targets with a `firestore_objc_` prefix are built from Firestore/Sources and Firestore/Example. Those directories will eventually move to into Firestore/ObjC (see go/firestore-spm for details). This PR anticipates that move.


Firestore/Example/Tests/API/CMakeLists.txt
```diff
@@ -12,27 +12,18 @@
...
-      FIRTimestampTest.m
-      FSTAPIHelpers.h
-      FSTAPIHelpers.mm
-      FSTUserDataConverterTests.mm
-    DEPENDS
-      FirebaseFirestore
-      firebase_firestore_objc_test_util
-  )
-endif(APPLE)
+if(NOT FIREBASE_IOS_BUILD_TESTS OR NOT APPLE)
```
dconeybe, line 15: Testing the value of `FIREBASE_IOS_BUILD_TESTS` and `APPLE` is different from the original file, which only tested `APPLE`.  Can you confirm that this is an intended change?

Firestore/Example/Tests/SpecTests/CMakeLists.txt
```diff
@@ -12,26 +12,29 @@
...
-      FirebaseFirestore
-      firebase_firestore_objc_test_util
-      firestore_local_testing
-      firestore_remote_testing
-    # Force this to run in the project-wide binary directory so that relative
-    # paths used during compilation are usable.
-    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
-  )
-endif(APPLE)
+if(NOT FIREBASE_IOS_BUILD_TESTS OR NOT APPLE)
```
dconeybe, line 15: Same here... please confirm that checking `FIREBASE_IOS_BUILD_TESTS` is indeed intended.

Firestore/Example/Tests/Integration/CMakeLists.txt
```diff
@@ -16,26 +16,17 @@ if(NOT APPLE)
   return()
```
dconeybe, line 16: Should `FIREBASE_IOS_BUILD_TESTS` be checked in this file as well?

PR #5329, opened 2020-04-09
Move FIRSecureStorage to GoogleUtilities

maksymmalyhin:
- `FIRSecureStorage` moved to GoogleUtilities and renamed to `GULKeychainStorage`
- FIS refactored to use `GULKeychainStorage`

### DIFF
--- a/FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
+++ b/FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
@@ -22,6 +22,8 @@
 #import "FBLPromises.h"
 #endif
 
+#import <GoogleUtilities/GULKeychainStorage.h>
+
 #import <FirebaseCore/FIRAppInternal.h>
 
 #import "FIRInstallationsAPIService.h"
@@ -32,7 +34,6 @@
 #import "FIRInstallationsLogger.h"
 #import "FIRInstallationsSingleOperationPromiseCache.h"
 #import "FIRInstallationsStore.h"
-#import "FIRSecureStorage.h"
 
 #import "FIRInstallationsHTTPError.h"
 #import "FIRInstallationsStoredAuthToken.h"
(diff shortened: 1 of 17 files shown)


FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m
```diff
@@ -22,6 +22,8 @@
 #import "FBLPromises.h"
 #endif
 
+#import <GoogleUtilities/GULKeychainStorage.h>
```
paulb777, line 25: Nit: alphabetically below FirebaseCore

GoogleUtilities.podspec
```diff
@@ -34,6 +34,8 @@ other Google CocoaPods. They're not intended for direct public usage.
     es.source_files = 'GoogleUtilities/Environment/**/*.[mh]'
     es.public_header_files = 'GoogleUtilities/Environment/**/*.h'
     es.private_header_files = 'GoogleUtilities/Environment/**/*.h'
+
+    es.dependency 'PromisesObjC', '~> 1.2'
```
paulb777, line 38: Separate from this review - but are we comfortable allowing minor version updates of Promises. It might be safer to lock to patch version updates ...

GoogleUtilities/Environment/SecureStorage/GULKeychainStorage.m
```diff
@@ -23,32 +23,33 @@
...
 
-@implementation FIRSecureStorage
+@implementation GULKeychainStorage
 
 - (instancetype)init {
   NSCache *cache = [[NSCache alloc] init];
   // Cache up to 5 installations.
   cache.countLimit = 5;
-  return [self initWithService:@"com.firebase.FIRInstallations.installations" cache:cache];
+  return [self initWithService:@"com.gul.KeychainStorage" cache:cache];
```
chliangGoogle, line 43: Seems like the location has changed, how do we ensure the old keychain data is not wiped when update to this version? 
And do we care about backward compatibility?

maksymmalyhin: Good catch! Let me make sure the service name is alway set by client, so Installations will keep using the same service as before.

PR #5339, opened 2020-04-10
fix an issue that delete token does not trigger token refresh notification or delegate

### LINKED ISSUE #5338: Delete token doesn't trigger token refresh notification and delegate
When a token is deleted, it doesn't trigger  FIRMessagingRegistrationTokenRefreshedNotification notification or calling the FIRMessagingDelegate `messaging:didReceiveRegistrationToken:` method.

chliangGoogle:
Fix #5338: When a token is deleted by calling `deleteToken` method, messaging doesn't trigger FIRMessagingRegistrationTokenRefreshedNotification notification or calling the FIRMessagingDelegate `messaging:didReceiveRegistrationToken:` method.

Hence the token parameter in the delegate API `messaging:didReceiveRegistrationToken:`should be nullable, a change we might need to get in the next breaking change API.

Update:
I also cleanup the code and realize IID send out two notifications  (kFIRInstanceIDDefaultGCMTokenNotification, kFIRInstanceIDTokenRefreshNotification) when fcm token is updated. Sometimes it calls one, while other time it calls the other. These two notifications are essentially doing the exactly the same thing. So I replaced one that is not in the public API of IID.

Note: I was trying out combine with FIRMessagingRegistrationTokenRefreshedNotification and want to use token refresh notification/delegate to update the environment variable identity to reflect on UI  on demand rather than manual update, then I discover this issue.

### DIFF
--- a/Example/Messaging/Sample/Sample/Identity.swift
+++ b/Example/Messaging/Sample/Sample/Identity.swift
@@ -13,10 +13,22 @@
 // limitations under the License.
 
 import SwiftUI
+import FirebaseMessaging
+import FirebaseInstallations
 
 public final class Identity: ObservableObject {
   // Identity that is unique per app.
   @Published public var instanceID: String? = nil
   // The token that Firebase Messaging use to send notifications.
   @Published public var token: String? = nil
+
+  init() {
+    Installations.installations().installationID(completion: { fid, error in
+      if let error = error as NSError? {
+        print("Failed to get FID: ", error)
+        return
+      }
+      self.instanceID = fid ?? "None"
+    })
+  }
 }
--- a/Example/Messaging/Sample/Sample/SceneDelegate.swift
+++ b/Example/Messaging/Sample/Sample/SceneDelegate.swift
@@ -39,13 +39,5 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate, MessagingDelegate {
 
   func messaging(_ messaging: Messaging, didReceiveRegistrationToken fcmToken: String) {
     identity.token = fcmToken
-
-    Installations.installations().installationID(completion: { fid, error in
-      if let error = error as NSError? {
-        print("Failed to get FID: ", error)
-        return
-      }
-      self.identity.instanceID = fid ?? "None"
-    })
   }
 }
--- a/Example/Messaging/Sample/Sample/SceneDelegate.swift
+++ b/Example/Messaging/Sample/Sample/SceneDelegate.swift
@@ -39,8 +39,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate, MessagingDelegate {
 
   func messaging(_ messaging: Messaging, didReceiveRegistrationToken fcmToken: String) {
     identity.token = fcmToken
-    InstanceID.instanceID().instanceID { result, error in
-      self.identity.instanceID = result?.instanceID
-    }
+
+    Installations.installations().installationID(completion: { fid, error in
+      if let error = error as NSError? {
+        print("Failed to get FID: ", error)
+        return
+      }
+      self.identity.instanceID = fid ?? "None"
+    })
   }
 }
--- a/Example/Messaging/Sample/Sample/SceneDelegate.swift
+++ b/Example/Messaging/Sample/Sample/SceneDelegate.swift
@@ -26,7 +26,6 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate, MessagingDelegate {
              options connectionOptions: UIScene.ConnectionOptions) {
     let contentView = ContentView()
     // Use a UIHostingController as window root view controller.
-    Messaging.messaging().delegate = self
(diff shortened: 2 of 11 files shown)


Firebase/Messaging/FIRMessaging.m
```diff
@@ -955,18 +956,14 @@ - (void)didReceiveDefaultInstanceIDToken:(NSNotification *)notification {
...
-    NSNotificationCenter *center = [NSNotificationCenter defaultCenter];
-    [center postNotificationName:FIRMessagingRegistrationTokenRefreshedNotification object:nil];
+  NSString *oldToken = self.defaultFcmToken;
+  self.defaultFcmToken = [token copy];
+  if ((self.defaultFcmToken && ![self.defaultFcmToken isEqualToString:oldToken]) ||
+      (self.defaultFcmToken && !oldToken) || (!self.defaultFcmToken && oldToken)) {
+    [self notifyDelegateOfFCMTokenAvailability];
   }
+  NSNotificationCenter *center = [NSNotificationCenter defaultCenter];
+  [center postNotificationName:FIRMessagingRegistrationTokenRefreshedNotification object:nil];
```
paulb777, line 966: Is it intentional to always post here? There are no cases left where there shouldn't be a notification?

chliangGoogle:
yeah, I left the public API header change out of this PR.

Firebase/Messaging/FIRMessaging.m
chliangGoogle: Good catch! it should be called any time the delgate method is called, moving into the notify delegate method. 
I didn't test the case for notification only for delegate method. I will update with a integration test.

chliangGoogle: Looks like we need to encrypt a googleService-Info.pilst file to ensure the network call succeed in integration test. I might have to add it in another PR as I haven't figure out how to do that.

chliangGoogle:
I will kick out a separate PR for the integration test for this as we have configuration needs to be done.

Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -245,19 +245,21 @@ - (NSString *)cachedTokenIfAvailable {
...
 
 - (void)setDefaultFCMToken:(NSString *)defaultFCMToken {
-  if (_defaultFCMToken && defaultFCMToken && [defaultFCMToken isEqualToString:_defaultFCMToken]) {
-    return;
+  // Sending this notification out will ensure that FIRMessaging and FIRInstanceID has the updated
+  // default FCM token.
+  // Only notify of token refresh if we have a new valid token that's different than before
+  if ((defaultFCMToken.length && _defaultFCMToken.length &&
+       ![defaultFCMToken isEqualToString:_defaultFCMToken]) ||
+      (defaultFCMToken.length && !_defaultFCMToken.length) ||
```
paulb777, line 253: For these two lines, how can there be a length and not a length at the same time?

chliangGoogle: one is parameter that passes in(defaultFCMToken) that is not an empty string and the other is an  instance variable (_defaultFCMToken) that might be empty or nil, in this case, we still think token is changed. Does that make sense? 
Those two parameters are different adn we only assign  the updated value to instance variable after this check.

paulb777: Oh I see. I missed the underscore.

Would `defaultFCMToken.length !=_defaultFCMToken.length` be sufficient to replace lines 253 and 254?

chliangGoogle: Good catch! Updated.

Example/Messaging/Sample/Sample/SceneDelegate.swift
```diff
@@ -35,12 +34,27 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate, MessagingDelegate {
...
+      .receive(on: RunLoop.main)
+      .sink(receiveValue: {
+        Installations.installations().installationID(completion: { fid, error in
+          if let error = error as NSError? {
+            print("Failed to get FID: ", error)
+            return
+          }
+          self.identity.instanceID = fid ?? "None"
+          })
+        })
```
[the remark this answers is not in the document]
chliangGoogle: Done.

Example/Messaging/Sample/Sample/Identity.swift
```diff
@@ -13,10 +13,22 @@
...
   // The token that Firebase Messaging use to send notifications.
   @Published public var token: String? = nil
+
+  init() {
+    Installations.installations().installationID(completion: { fid, error in
+      if let error = error as NSError? {
+        print("Failed to get FID: ", error)
+        return
+      }
+      self.instanceID = fid ?? "None"
```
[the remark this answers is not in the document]
chliangGoogle: Done. Good catch!

PR #5341, opened 2020-04-10
Three ZipBuilder Fixes

paulb777:
Last night's zip build cron job failed because of a bug exposed by #5329 updating the `GoogleUtilities` version locally.  We were getting the podspec name from the podname without truncating subspec's from the podname.

Also fixed a warning from `pod install` if a subspec install was requested without a version along with a request to install the base spec with a version.

I also added a check to only allow `-buildDependencies false` option when `-zipPods` is used since it does not make sense (nor work) for Firebase product builds.

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -327,12 +327,12 @@ enum CocoaPodUtils {
                                                  in installedPods: [String: PodInfo])
     -> [VersionedPod] {
     return transitivePodDependencies(for: podName, in: installedPods).map {
-      var podVersion : String?
+      var podVersion: String?
       if let version = installedPods[$0]?.version {
         podVersion = version
       } else {
         // See if there's a version on the base pod.
-        let basePod = String($0.split(separator:"/")[0])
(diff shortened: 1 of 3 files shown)


ZipBuilder/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -327,7 +327,15 @@ enum CocoaPodUtils {
                                                  in installedPods: [String: PodInfo])
     -> [VersionedPod] {
     return transitivePodDependencies(for: podName, in: installedPods).map {
-      CocoaPodUtils.VersionedPod(name: $0, version: installedPods[$0]?.version)
+      var podVersion: String?
+      if let version = installedPods[$0]?.version {
+        podVersion = version
+      } else {
+        // See if there's a version on the base pod.
+        let basePod = String($0.split(separator: "/")[0])
```
morganchen12, line 335: Is it possible for this to out-of-index for dependencies that aren't subspecs, e.g. nanopb?

paulb777: My understanding of split is there will always be a 0th element for a non-nil string.

PR #5345, opened 2020-04-11
Workaround Carthage Protobuf parallel install issue

### LINKED ISSUE #5276: Multiple/Duplicate Protobuf Dependencies of Carthage Binary Firebase Components

  * Xcode version: 11.4
  * Firebase SDK version: 6.21.0
  * Firebase Component: Remote Config, Messaging (and all other components depending on Protobuf)
  * Component version: n/a

The usage description of the [Carthage documentation](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md) states that we should add a subset of the provided components to the Cartfile of a project that uses Firebase as their dependency.

The problem is: Multiple of those components internally depend on Protobuf and all of them bundle their own Protobuf.framework within the zip file used for binary distribution.

Example:
```
RemoteConfig
├── FirebaseABTesting.framework
├── FirebaseRemoteConfig.framework
└── Protobuf.framework

Messaging
├── FirebaseMessaging.framework
└── Protobuf.framework
```

When Carthage downloads these zip files it will extract its contents into the `Carthage/Build/<platform>` folder.

If a project depends on multiple Firebase components that depend on Protobuf, Carthage will always fail with an error like the following.

> Failed to write to /Users/ffittschen/Firebase_Test/Carthage/Build/iOS/Protobuf.framework: Error Domain=NSCocoaErrorDomain Code=513 "“Protobuf.framework” couldn’t be removed because you don’t have permission to access it." UserInfo={NSUserStringVariant=(
    Remove
), NSFilePath=/Users/ffittschen/Firebase_Test/Carthage/Build/iOS/Protobuf.framework, NSUnderlyingError=0x7fb827104db0 {Error Domain=NSPOSIXErrorDomain Code=66 "Directory not empty"}}

#### Steps to reproduce:

1. Create a Cartfile with this content:

```ruby
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseMessagingBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json"
```

2. Run `carthage update --platform iOS`

#### Proposed solution:

I would like to propose to either distribute Protobuf seperately instead of being bundled within multiple Firebase components, or bundle it with the Firebase Analytics component, which should always be included according to the [Carthage documentation](https://github.com/firebase/firebase-ios-sdk/blob/master/Carthage.md).

The Cartfile with a separate Protobuf component could look as follows. If Protobuf would be bundled within the Analytics component, the Cartfile would not look different than existing Cartfiles

<details>

<summary>Cartfile with separate Protobuf component</summary>

```ruby
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseAnalyticsBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/Protobuf.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseMessagingBinary.json"
binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseRemoteConfigBinary.json"
```

</details>

paulb777:
Fix #5276. Carthage can fail to install if two binary installs include the same framework. This PR factors Protobuf.framework out of the several bundles it was included in into a single Carthage bundle. 

The instructions are updated to describe when the new `FirebaseProtobufBinary.json` must now be included in the Cartfile.

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
+++ b/ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
@@ -75,7 +75,7 @@ extension CarthageUtils {
   /// Generates all required files for a Carthage release.
   ///
   /// - Parameters:
-  ///   - packagedDir: The packaged directory assembled for Carthage and Zip distribution.
+  ///   - packagedDir: The packaged directory assembled for the Carthage distribution.
   ///   - templateDir: The template project directory, contains the dummy Firebase library.
   ///   - jsonDir: Location of directory containing all JSON Carthage manifests.
   ///   - firebaseVersion: The version of the Firebase pod.
@@ -87,6 +87,7 @@ extension CarthageUtils {
                                               jsonDir: URL,
                                               artifacts: ZipBuilder.ReleaseArtifacts,
                                               outputDir: URL) {
+    factorProtobuf(inPackagedDir: packagedDir)
     let directories: [String]
     do {
       directories = try FileManager.default.contentsOfDirectory(atPath: packagedDir.path)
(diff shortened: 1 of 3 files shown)


ZipBuilder/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -190,6 +191,44 @@ extension CarthageUtils {
     }
   }
 
+  /// Factor Protobuf into a separate Carthage distribution to avoid Carthage install issues
+  /// trying to install the same framework from multiple bundles(#5276).
```
ryanwilson, line 195: Nit: space after `bundles`

paulb777:
Breakage is definitely a concern but the current breakage described in #5276 seems worse.  Mitigating the risk with documentation. And at least Protobuf is reasonably stable so is unlikely to break if people continue to use the old one and don't update to the new one.

PR #5353, opened 2020-04-13
Fix < 0  comparison for an unsigned long long

mikehaney24:


mikehaney24:
> line 43 should also be changed.
> 
> Otherwise if countChars is nil, nextEventID will be max ulll

Done.

Issue #5360, opened 2020-04-14
Memory Leaks in GoogleDataTransport


dimohamdy:
I check memory-graph in Xcode and leaks in Xcode Instruments  and I found leaks in [GDTCORUploadPackage.m](https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m)

<img width="1680" alt="Screen Shot 2020-04-14 at 12 18 28 PM" src="https://user-images.githubusercontent.com/1476754/79209744-7abc5a80-7e4c-11ea-81f3-ce3d587e01d0.png">
<img width="1680" alt="Screen Shot 2020-04-14 at 12 17 54 PM" src="https://user-images.githubusercontent.com/1476754/79209756-7ee87800-7e4c-11ea-83f1-8b89c22a2626.png">
<img width="1005" alt="Screen Shot 2020-04-13 at 11 19 35 PM" src="https://user-images.githubusercontent.com/1476754/79209811-8f005780-7e4c-11ea-8745-b30a556ad9f3.png">

PR #5362, opened 2020-04-14
Fix Memory leak of NSTimer when before use [self copy]

dimohamdy:
I make NSTimer to weak and force set it to nil after invalidate it 
I fix the leaks in this issue [https://github.com/firebase/firebase-ios-sdk/issues/5360](https://github.com/firebase/firebase-ios-sdk/issues/5360)

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
@@ -31,7 +31,7 @@
   BOOL _isHandled;
 
   /** A timer that will regularly check to see whether this package has expired or not. */
-  NSTimer *_expirationTimer;
+  __weak NSTimer *_expirationTimer;
 }
 
 - (instancetype)initWithTarget:(GDTCORTarget)target {
@@ -79,6 +79,7 @@
   if (!_isHandled && _handler &&
       [_handler respondsToSelector:@selector(packageDelivered:successful:)]) {
     [_expirationTimer invalidate];
+    _expirationTimer = nil;
     _isHandled = YES;
     [_handler packageDelivered:[self copy] successful:YES];
   }
@@ -89,6 +90,7 @@
   if (!_isHandled && _handler &&
       [_handler respondsToSelector:@selector(packageDelivered:successful:)]) {
     [_expirationTimer invalidate];
+    _expirationTimer = nil;
     _isHandled = YES;
     [_handler packageDelivered:[self copy] successful:NO];
   }


dimohamdy:
@googlebot I signed it!

dimohamdy:
@googlebot I signed it!

GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
```diff
@@ -31,7 +31,7 @@ @implementation GDTCORUploadPackage {
   BOOL _isHandled;
 
   /** A timer that will regularly check to see whether this package has expired or not. */
-  NSTimer *_expirationTimer;
+  __weak NSTimer *_expirationTimer;
```
wilhuff, line 34: This can't be the fix for this issue. This reference is owned by `GDTCORUploadPackage` and is the only reference to the timer. Making this weak will make it subject to immediate collection by ARC and will effectively disable the timer. This doesn't break `GDTCORUploadPackage` because nil dispatch will continue to succeed.

dimohamdy: The timer has a reference with `GDTCORUploadPackage` from the initialization of the `NSTimer` I only make the timer with nil after setting it as invalidate  and I am sure the `GDTCORUploadPackage` will be copied with `[self copy]`

morganchen12: Wouldn't `expirationTimer` be immediately `nil` after assigning it to the weak variable on line 44, regardless of copies?

dimohamdy: Screen Shot 2020-04-14 at 9 34 52 PM

 The initialization of NSTimer will make retain cycle with this style

wilhuff: I grant that there's a retain cycle, but the expirationTimer needs to be held strongly by `GDTCORUploadPackage` or it will be released essentially immediately. There needs to be a strong reference to the timer.

The problem is that `NSTimer` needs to keep a weak reference to the `GDTCORUploadPackage` instance.

This would be better written without using selectors.

```objc
__weak typeof(self) weakSelf = self;
_expirationTimer = [NSTimer scheduledTimerWithTimeInterval:5.0
                                                   repeats:YES
                                                   block:^(NSTimer *timer) {
  __strong typeof(self) strongSelf = weakSelf;
  if (strongSelf) {
    [strongSelf checkIfPackageIsExpired:timer];
  }
});
```

mikehaney24: Wil's suggestion is the correct way of doing this.

dimohamdy: it's not the solution by the way [scheduledtimerwithtimeinterval](https://developer.apple.com/documentation/foundation/nstimer/2091889-scheduledtimerwithtimeinterval?language=objc
) it's support **iOS 10+** 
and  `GoogleDataTransport` support from **iOS 8** depend [GoogleDataTransport.podspec](https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleDataTransport.podspec)

dimohamdy: I try this solution 
```
    __weak id weakSelf = self;
    _expirationTimer = [NSTimer scheduledTimerWithTimeInterval:5.0
                                                        target:weakSelf
                                                      selector:@selector(checkIfPackageIsExpired:)
                                                      userInfo:nil
                                                       repeats:YES];
```
but unfortunately doesn't work also

Screen Shot 2020-04-15 at 11 29 39 AM

dimohamdy: How I can check  the expire packet scenario ?

morganchen12: Passing a weak reference into the timer's `target:` argument won't cause the target to not be strongly retained, since ARC can't conditionally remove `retain`s from the body of the timer method (but it can avoid inserting `retain`s when constructing a block literal with a weak reference).

The memory leak is caused by the timer not being invalidated in some cases. Gil's fix resolves this by weakening the timer's reference to `self`, so when `self` is deallocated the timer will be invalidated in the `dealloc` method. The best way to achieve the same thing in iOS versions below 10 is to use a helper object that holds a weak reference to the GDTCORUploadPackage instance, and then passing that object to the timer instead of `self`.

I'm not sure how to check the packet expiration method, but maybe unit tests would be sufficient for testing this leak.

mikehaney24:
I am writing both the fix and a unit test as we speak.

GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
dimohamdy: @mikehaney24 Which solution will you implement the one that uses block?

mikehaney24:
We have to keep compatibility back to iOS 8 for now, so I'm using a holding
object. This issue will be tagged as fixed by the upcoming PR.

morganchen12:
Hey @dimohamdy, ultimately we decided to merge #5382 instead of this PR because we're rushing this change into the next release, as this bug affects a significant number of apps. Though this PR isn't being merged, it contributed to our understanding of the memory leak and resulted in the fix that @mikehaney24 contributed. Thanks for all of your help, and sorry for closing this PR after all the time you spent on it!

mikehaney24:
Yeah, I want to second this--this is the first public contribution for GoogleDataTransport, and I'm incredibly excited about that! The code freeze for the next release was yesterday EoD, and I really wanted to get this fixed for you by next release. Please test that other PR and let me know if it fixed the problem for you or not! Thanks again!

PR #5369, opened 2020-04-14
GoogleUtilities: fix public header imports

maksymmalyhin:
- use modular style import for public headers


maksymmalyhin:
Submitting the PR now because all relevant tests have passed.

PR #5373, opened 2020-04-14
Update CHANGELOG for Firestore v1.12.1

wilhuff:
Also fix the version entry for the previous release. The last Firestore
release was 1.12.0, not 1.11.3.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -4,7 +4,9 @@
 - [changed] Internal improvements for future C++ and Unity support.
 
 # v1.12.0
-- [changed] Internal improvements.
+- [changed] Internal improvements for future C++ and Unity support. The minor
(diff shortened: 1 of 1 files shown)


Firestore/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Unreleased
 
-# v1.11.3
-- [changed] Internal changes.
+# v1.12.1
+- [changed] Internal improvements for future C++ and Unity support.
+
+# v1.12.0
+- [changed] Internal improvements.
```
morganchen12, line 7: I remember this change being difficult to document, but if you're able to succinctly specify why this version was a minor version increase instead of a patch version increase, it would be good to add here.

wilhuff: Took a crack at this, but ultimately it comes down to: this was breaking for (alpha) C++ but not for CocoaPods/Carthage/etc users.

PR #5375, opened 2020-04-15
add settings to test app

chliangGoogle:
Add a setting view (re-use the detail view) to toggle autoInitEnable flag and to choose between delegate or notification for token refresh. This covers all the token refresh tests in original test app.


Example/Messaging/Sample/Sample/ContentView.swift
```diff
@@ -42,8 +43,8 @@ struct ContentView: View {
               .layoutPriority(1)
           }
 
-          NavigationLink(destination: DetailView()) {
-            Text("Show Detail View")
+          NavigationLink(destination: SettingsView(settings: self.settings)) {
```
ryanwilson, line 46: Since `settings` is passed in via the environment, you should be able to ignore it in this struct entirely and just put `@EnvironmentObject var settings: UserSettings` in the `SettingsView` itself.

Example/Messaging/Sample/Sample/SceneDelegate.swift
```diff
@@ -22,28 +22,37 @@ import FirebaseInstallations
...
-    // Subscribe to token refresh
-    _ = NotificationCenter.default
-      .publisher(for: Notification.Name.MessagingRegistrationTokenRefreshed)
-      .map { $0.object as? String }
-      .receive(on: RunLoop.main)
-      .assign(to: \Identity.token, on: identity)
-      .store(in: &cancellables)
+    if !settings.shouldUseDelegateThanNotification {
+      // Subscribe to token refresh
+      _ = NotificationCenter.default
```
ryanwilson, line 49: I think since it's stored in cancellables set you can remove the `_ = `. Sorry for not catching before.

Example/Messaging/Sample/Sample/UserSettings.swift
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import FirebaseMessaging
+
+public final class UserSettings: ObservableObject {
+  @Published public var isAutoInitEnabled: Bool = Messaging.messaging().isAutoInitEnabled {
```
ryanwilson, line 19: This may not work super well depending on what time it's initialized - potentially injecting Messaging to get the initial value? 

It also won't be in sync with `Messaging.messaging().isAutoInitEnabled` if it gets modified separately (I realize this likely won't happen, but it's something that could happen). I think it's fine for now given this is a sample but potentially refactor afterwards!

Example/Messaging/Sample/Sample/ContentView.swift
chliangGoogle: But UserSettings is also used in SessionDelegate so I created the instance and passing in as a second environmentObject in contentView. (first one is identity)

ryanwilson: Right - adding an `.environmentObject()` on a view adds it for all children of that view.  The `settings` variable here isn't actually used in the `ContentView` struct (we don't access the variable outside of passing it through to `SettingsView`) but since `SettingsView` has `ContentView()` as an ancestor which has `UserSettings` in it's environment, it'll automatically get access to that environment variable as well. The resultant state (concatenated, missing other parts) can be like:

In the `SceneDelegate`:
```
UIHostingController(rootView: contentView
  // Pass in the settings as an environmentObject to `contentView`
  .environmentObject(settings)
)
```

In the `ContentView` struct:
```
// No need to declare the `@EnvironmentObject var settings: UserSettings` in this struct
// since it's not used.

// No need to pass in the `UserSettings` here since it already has a copy in the environment.
NavigationLink(destination: SettingsView()) {
  Text("Settings")
}
```

In the `SettingsView` struct:
```
// Pull the settings from the environment here, no need to inject in the constructor
@EnvironmentObject var settings: UserSettings

// Remove the @ObservedObject since we have access through the environment.
```

chliangGoogle: Nice it works! So @environmentObject is kinda global in this file?

Example/Messaging/Sample/Sample/SceneDelegate.swift
chliangGoogle: Done. Nice!

Example/Messaging/Sample/Sample/UserSettings.swift
chliangGoogle: Sounds good. I will keep in mind and observe initialization would cause any issues.

Example/Messaging/Sample/Sample/ContentView.swift
ryanwilson: Global for all subviews - yes!

PR #5379, opened 2020-04-15
Crashlytics 4.0.0 Changelog

samedson:


Crashlytics/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.0.0
+
+ - [changed] Firebase Crashlytics SDKs are available for general access.
```
RelaxedPear, line 3: "The Firebase Crashlytics SDK is now available for general access."

This should be singular since this release note is specific to iOS, right?

samedson: Yeah that makes sense 👍

PR #5382, opened 2020-04-15
Don't leak upload packages or deadlock on lifecycle events

### LINKED ISSUE #5360: Memory Leaks in GoogleDataTransport

mikehaney24:
GDTCORUploadPackages would call targetToStorage in -initWithTarget. When this method was running during app terminate, this would cause a deadlock because calling targetToStorage is re-entrant while GDTCORRegistrar is sending lifecycle events to all the instances it's tracking.

NSTimer's invalidate method is also called whenever a package is completed, retried, or expired, without regard to the presence of a package handler.

A unit test is added to ensure that GDTCORUploadPackage's are not leaked by a retain

TSAN caught no issues

Fixes #5360

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
@@ -122,63 +122,66 @@
 #pragma mark - GDTCORLifecycleProtocol
 
 - (void)appWillBackground:(nonnull GDTCORApplication *)app {
-  dispatch_async(_registrarQueue, ^{
-    for (id<GDTCORUploader> uploader in [self->_targetToUploader allValues]) {
-      if ([uploader respondsToSelector:@selector(appWillBackground:)]) {
-        [uploader appWillBackground:app];
-      }
+  NSArray<id<GDTCORUploader>> *uploaders = [self.targetToUploader allValues];
+  for (id<GDTCORUploader> uploader in uploaders) {
+    if ([uploader respondsToSelector:@selector(appWillBackground:)]) {
+      [uploader appWillBackground:app];
     }
-    for (id<GDTCORPrioritizer> prioritizer in [self->_targetToPrioritizer allValues]) {
-      if ([prioritizer respondsToSelector:@selector(appWillBackground:)]) {
-        [prioritizer appWillBackground:app];
-      }
+  }
+  NSArray<id<GDTCORPrioritizer>> *prioritizers = [self.targetToPrioritizer allValues];
+  for (id<GDTCORPrioritizer> prioritizer in prioritizers) {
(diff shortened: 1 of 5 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
```diff
@@ -76,30 +113,30 @@ - (void)completeDelivery {
...
     _isHandled = YES;
     [_handler packageDelivered:[self copy] successful:NO];
   }
   GDTCORLogDebug(@"Upload package will retry in the future: %@", self);
 }
 
-- (void)checkIfPackageIsExpired:(NSTimer *)timer {
+- (void)checkIfPackageIsExpired {
   if ([[GDTCORClock snapshot] isAfter:_deliverByTime]) {
+    [_expirationTimer invalidate];
```
maksymmalyhin, line 137: Would be nice to have a follow up PR with tests for the changes

GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
```diff
@@ -122,63 +122,66 @@ - (void)registerPrioritizer:(id<GDTCORPrioritizer>)prioritizer target:(GDTCORTar
 #pragma mark - GDTCORLifecycleProtocol
 
 - (void)appWillBackground:(nonnull GDTCORApplication *)app {
-  dispatch_async(_registrarQueue, ^{
-    for (id<GDTCORUploader> uploader in [self->_targetToUploader allValues]) {
-      if ([uploader respondsToSelector:@selector(appWillBackground:)]) {
-        [uploader appWillBackground:app];
-      }
+  NSArray<id<GDTCORUploader>> *uploaders = [self.targetToUploader allValues];
```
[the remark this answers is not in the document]
mikehaney24: Yeah sure! The getters (self.targetToXXXX) all run on the queue, which is why the deadlock would occur. `-allValues` creates a new immutable dictionary of all the keys, so it's safe to operate on that.

GoogleDataTransport/GDTCORLibrary/GDTCORUploadPackage.m
mikehaney24: A unit test for a memory leak is included in this PR. This change is associated with that.

PR #5387, opened 2020-04-15
Synchronize useUserAccessGroup

### LINKED ISSUE #4175: [Bug?] Unsafe to call useUserAccessGroup immediately after FirebaseApp.configure()

  * Xcode version: 11.1 (11A1027)
  * Firebase SDK version: 6.11
  * Firebase Component: Auth
  * Component version: 6.3.1

#### Steps to reproduce:

1. Follow the steps in [Enabling cross-app authentication with shared iOS Keychain](https://firebase.google.com/docs/auth/ios/single-sign-on)
2. Do the following:

```swift
// ... inside application(_:didFinishLaunchingWithOptions:):
FirebaseApp.configure()

do {
    try Auth.auth().useUserAccessGroup("TEAMID.com.example.group1")
    print("useUserAccessGroup OK")
} catch let error as NSError {
    print("useUserAccessGroup failed with \(error)")
}
```

**Expected output:**
useUserAccessGroup OK

**Actual output:**
useUserAccessGroup failed with Error Domain=Foundation._GenericObjCError Code=0 "(null)"

### My analysis
It seems like there is some async setup going on which prevents this from working. If I wrap the `useUserAccessGroup` call with 

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
	do {
	    try Auth.auth().useUserAccessGroup("TEAMID.com.example.group1")
	    print("useUserAccessGroup OK")
	} catch let error as NSError {
	    print("useUserAccessGroup failed with \()")
	}
}
```

or move it out of `application(_:didFinishLaunchingWithOptions:):` and call it later it works as expected.

I think this is a bug, but if it's not, I would suggest improving the error message to something more helpful.

paulb777:
Fix #4175

### DIFF
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -431,11 +431,6 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
           UIApplicationDidEnterBackgroundNotification.
    */
   id<NSObject> _applicationDidEnterBackgroundObserver;
-
-  /** @var initializingKeychain
-      @brief Locked until the keychain is initialized.
-   */
-  dispatch_semaphore_t _initializingKeychain;
 }
 
 + (void)load {
@@ -486,7 +481,6 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
     _listenerHandles = [NSMutableArray array];
     _requestConfiguration = [[FIRAuthRequestConfiguration alloc] initWithAPIKey:APIKey];
     _firebaseAppName = [appName copy];
-    _initializingKeychain = dispatch_semaphore_create(0);
 #if TARGET_OS_IOS
     _settings = [[FIRAuthSettings alloc] init];
     static Class applicationClass = nil;
@@ -535,7 +529,7 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
                         @"Error loading saved user when starting up: %@", error);
           }
         } else {
-          [strongSelf internalUseUserAccessGroup:storedUserAccessGroup error:&error];
+          [strongSelf useUserAccessGroup:storedUserAccessGroup error:&error];
           if (error) {
             FIRLogError(kFIRLoggerAuth, @"I-AUT000001",
                         @"Error loading saved user when starting up: %@", error);
@@ -564,7 +558,6 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
       }
 #endif  // ((TARGET_OS_IOS || TARGET_OS_TV) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 130000))
 #endif  // TARGET_OS_IOS
-      dispatch_semaphore_signal(self->_initializingKeychain);
     });
   }
   return self;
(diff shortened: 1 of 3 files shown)


FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2171,6 +2171,12 @@ - (nullable NSString *)getUserID {
 
 - (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
                      error:(NSError *_Nullable *_Nullable)outError {
+  // self.storedUserManager is initialized asynchronously.
+  __block FIRAuthStoredUserManager *storedUserManager;
+  dispatch_sync(FIRAuthGlobalWorkQueue(), ^{
```
wilhuff, line 2176: This creates the potential for deadlock because `useUserAccessGroup:error:` is called from code already on the `FIRAuthGlobalWorkQueue()`

See https://github.com/firebase/firebase-ios-sdk/blob/f41bcaf177b7bca5f8987cdb382484e50c2072aa/FirebaseAuth/Sources/Auth/FIRAuth.m#L532

paulb777: Thanks @wilhuff Good catch! PTAL at the update and let me know if you have a suggestion for a better way.

paulb777:
I'll see if I can add a unit test as well before merging.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2196,6 +2203,15 @@ - (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
   return YES;
 }
 
+- (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
+                     error:(NSError *_Nullable *_Nullable)outError {
+  // Make sure that the keychain has been initialized.
+  dispatch_semaphore_wait(_initializingKeychain, DISPATCH_TIME_FOREVER);
```
wilhuff, line 2209: This approach of splitting internal and external usage would work just as well with your `dispatch_sync` approach and would avoid needing to keep an extra semaphore (or keep waiting/re-signaling it).

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2169,8 +2169,8 @@ - (nullable NSString *)getUserID {
 
 #pragma mark - Keychain sharing
 
-- (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
-                     error:(NSError *_Nullable *_Nullable)outError {
+- (BOOL)internalUseUserAccessGroup:(NSString *_Nullable)accessGroup
```
wilhuff, line 2172: Line 532 needs to change to use the internal version and you're done.

https://github.com/firebase/firebase-ios-sdk/blob/dba4325fd26b4a40a94339b425e6297e4a4f1f96/FirebaseAuth/Sources/Auth/FIRAuth.m#L532

The test needs to set up a stored user access group to effectively verify that the deadlock during initialization isn't happening.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2196,6 +2196,14 @@ - (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
   return YES;
 }
 
+- (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
+                     error:(NSError *_Nullable *_Nullable)outError {
+  // self.storedUserManager is initialized asynchronously. Make sure it is done.
+  dispatch_sync(FIRAuthGlobalWorkQueue(), ^{
```
renkelvin, line 2202: I'm thinking maybe it makes sense to load storedUserManager synchronously, but it requires more investigation on the impact. This should be good for now.

PR #5389, opened 2020-04-16
Fix unused variable warnings

wilhuff:
A forthcoming change will change the CMake build so that `libprotobuf` no longer exports a `-Wno-unused-parameter` compiler command-line flag to any target that depends on it. This fixes all the tests
and other code that came to have this error disabled.

Note there are a few cases where this doesn't just silence the warning:

  * FIRFirestoreTests makes an assertion about the value
  * Fixes a bug in `writeDocumentRef:data:` that ignored the `ref` argument
  * Removes a never used `path` argument in `readerAndWriterOnDocumentRef:`
  * Removes a never used `fields` argument in `SerializationTest`'s `Decode`


Firestore/Example/Tests/Integration/FSTDatastoreTests.mm
```diff
@@ -144,10 +144,13 @@ - (void)applySuccessfulWriteWithResult:(const MutationBatchResult &)batchResult
 }
 
 - (void)rejectFailedWriteWithBatchID:(BatchId)batchID error:(NSError *)error {
+  (void)batchID;
```
thebrianchen, line 147: It appears that this method is only being called in `HandleRejectedWrite`, and it's also "not implemented". 

If this method is never going to be implemented for tests, why do we bother to convert the `Status` into `error.ToNSError()` when passing it into `rejectFailedWriteWithBatchID`, instead of having the method take in `Status`?

Firestore/Example/Tests/Integration/FSTDatastoreTests.mm
```diff
@@ -144,10 +144,13 @@ - (void)applySuccessfulWriteWithResult:(const MutationBatchResult &)batchResult
...
+  (void)error;
```
thebrianchen, line 148: Also, is there a term that refers to this patter of calling `(void)unusedParam`? I'm curious to know if this the commonly accepted way to handle unused vars that get passed down, or if there are other ways to do so.

Coming from TS patterns, this seems more excessive than just disabling the unused var linter for this line.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -204,6 +204,7 @@ @implementation FSTSpecTests {
                         userInfo:nil];
 
 - (std::unique_ptr<Persistence>)persistenceWithGCEnabled:(BOOL)GCEnabled {
+  (void)GCEnabled;
```
thebrianchen, line 207: From what I see in Xcode, this method is being used in two places, and there seems to be code that facilitates storing the `GCEnabled` state for the tests. However, it does not appear that `GCEnabled` is being used other than when passing it into `persistenceWithGCEnabled`, where it goes unused.

Could we just remove `GCEnabled` from this method and tests entirely, rather than having to add the `void(GCEnabled)`?

PR #5407, opened 2020-04-17
Add LLDB type summaries for Firestore types

wilhuff:
There are some particular pain points debugging Firestore today because contents of variables are obscured due to information hiding and shared representation techniques we use. Also, we use elements of the Abseil library to supplement what's available in the C++11 standard library.

Taken together, these make our default experience in LLDB and the Xcode debugger quite painful: getting the path of a document currently takes 5 clicks. This change pulls the document path into the summary line, making it available with no clicks.

Only a small handful of types are included (see individual commits).

When printing a `FIRDocumentSnapshot`, here was before:

```
(lldb) p *snapshot
(FIRDocumentSnapshot) $1 = {
  NSObject = {
    isa = FIRDocumentSnapshot
  }
  _snapshot = {
    firestore_ = std::__1::shared_ptr<firebase::firestore::api::Firestore>::element_type @ 0x000060000233f318 strong=4 weak=2 {
      __ptr_ = 0x000060000233f318
    }
    internal_key_ = {
      path_ = std::__1::shared_ptr<const firebase::firestore::model::ResourcePath>::element_type @ 0x0000600001016f58 strong=3 weak=1 {
        __ptr_ = 0x0000600001016f58
      }
    }
    internal_document_ = {
      absl::lts_2020_02_25::optional_internal::optional_data<firebase::firestore::model::Document, false> = {
        absl::lts_2020_02_25::optional_internal::optional_data_base<firebase::firestore::model::Document> = {
          absl::lts_2020_02_25::optional_internal::optional_data_dtor_base<firebase::firestore::model::Document, false> = {
            engaged_ = true
             = {
              data_ = {
                firebase::firestore::model::MaybeDocument = {
                  rep_ = std::__1::shared_ptr<const firebase::firestore::model::MaybeDocument::Rep>::element_type @ 0x0000600003448da8 strong=5 weak=1 {
                    __ptr_ = 0x0000600003448da8
                  }
                }
              }
              dummy_ = {}
            }
          }
        }
      }
    }
    metadata_ = (pending_writes_ = false, from_cache_ = true)
  }
  _cachedMetadata = nil
}
```

... and after:

```
(lldb) p *snapshot
(FIRDocumentSnapshot) $0 = "test-collection/SXJOSv8YRvhc6EJih1xr" {
  NSObject = {
    isa = FIRDocumentSnapshot
  }
  _snapshot = "test-collection/SXJOSv8YRvhc6EJih1xr" {
    firestore_ = std::__1::shared_ptr<firebase::firestore::api::Firestore>::element_type @ 0x000060000233f318 strong=4 weak=2 {
      __ptr_ = 0x000060000233f318
    }
    internal_key_ = "test-collection/SXJOSv8YRvhc6EJih1xr"
    internal_document_ = engaged=true {
      engaged_ = true
      data_ = {
        firebase::firestore::model::MaybeDocument = {
          rep_ = std::__1::shared_ptr<const firebase::firestore::model::MaybeDocument::Rep>::element_type @ 0x0000600003448da8 strong=5 weak=1 {
            __ptr_ = 0x0000600003448da8
          }
        }
      }
    }
    metadata_ = (pending_writes_ = false, from_cache_ = true)
  }
  _cachedMetadata = nil
}
```

If we agree on approach, this can be extended to more types.

### DIFF
--- a/Firestore/README.md
+++ b/Firestore/README.md
@@ -10,6 +10,17 @@
   * Select the Firestore_Tests_iOS scheme
   * ⌘-u to build and run the unit tests
 
+### Improving the debuger experience
+
+You can install a set of type formatters to improve the presentation of
+Firestore internals in LLDB and Xcode. Add the following to your `~/.lldbinit` file:
+
+```
+command script import ~/path/to/firebase-ios-sdk/scripts/lldb/firestore.py
+```
+
+(substitute the location of your checkout of the firebase-ios-sdk.)
+
 ### Running Integration Tests
 
   * [Set up a `GoogleServices-Info.plist`](//github.com/firebase/firebase-ios-sdk#running-sample-apps)
--- a/Firestore/README.md
+++ b/Firestore/README.md
@@ -10,7 +10,7 @@
   * Select the Firestore_Tests_iOS scheme
   * ⌘-u to build and run the unit tests
 
-### Improving the debuger experience
+### Improving the debugger experience
 
 You can install a set of type formatters to improve the presentation of
 Firestore internals in LLDB and Xcode. Add the following to your `~/.lldbinit` file:
--- a/scripts/lldb/firestore.py
+++ b/scripts/lldb/firestore.py
@@ -39,6 +39,67 @@ we're making use of:
 """
 
 
+# Abseil
+
+class AbseilOptional_SynthProvider(object):
+  """A synthetic child provider that hides the internals of absl::optional.
+  """
+
+  def __init__(self, value, params):
+    self.value = value
+    self.engaged = None
+    self.data = None
+
+  def update(self):
+    # Unwrap all the internal optional_data and similar types
+    value = self.value
+    while True:
+      if value.GetNumChildren() <= 0:
+        break
+
+      child = value.GetChildAtIndex(0)
+      if not child.IsValid():
+        break
+
+      if 'optional_internal' not in child.GetType().GetName():
+        break
+
+      value = child
+
+    # value should now point to the innermost absl::optional container type.
+    self.engaged = value.GetChildMemberWithName('engaged_')
+
+    if self.has_children():
+      self.data = value.GetChildMemberWithName('data_')
+
+    else:
+      self.data = None
+
+  def has_children(self):
+    return self.engaged.GetValueAsUnsigned(0) != 0
+
+  def num_children(self):
+    return 2 if self.has_children() else 1
+
+  def get_child_index(self, name):
+    if name == 'engaged_':
+      return 0
+    if name == 'data_':
+      return 1
+    return -1
+
+  def get_child_at_index(self, index):
+    if index == 0:
+      return self.engaged
+    if index == 1:
+      return self.data
+
+
+def AbseilOptional_SummaryProvider(value, params):
+  # Operates on the synthetic children above, calling has_children.
+  return 'engaged={0}'.format(format_bool(value.MightHaveChildren()))
+
+
 # model
 
 def DocumentKey_SummaryProvider(value, params):
@@ -95,6 +156,11 @@ def format_string(string):
   return json.dumps(string)
 
 
+def format_bool(value):
+  """Formats a Python value as a C++ bool literal."""
+  return 'true' if value else 'false'
+
+
 def deref_shared(value):
   """Dereference a shared_ptr."""
   return value.GetChildMemberWithName('__ptr_').Dereference()
@@ -108,6 +174,15 @@ def __lldb_init_module(debugger, params):
     args = ' '.join(args)
     run('type summary add -w firestore -F {0} {1} {2}'.format(
       qname(provider), args, typename))
+
+  def add_synthetic(provider, typename, *args):
+    args = ' '.join(args)
+    run('type synthetic add -l {0} -w firestore {1} {2}'.format(
+      qname(provider), args, typename))
+
+  optional_matcher = '-x absl::[^:]*::optional<.*>'
+  add_summary(AbseilOptional_SummaryProvider, optional_matcher, '-e')
+  add_synthetic(AbseilOptional_SynthProvider, optional_matcher)
 
   api = 'firebase::firestore::api::'
   add_summary(DocumentReference_SummaryProvider, api + 'DocumentReference')
--- a/scripts/lldb/firestore.py
+++ b/scripts/lldb/firestore.py
@@ -62,6 +62,18 @@ def ResourcePath_SummaryProvider(value, params):
   return text
 
 
+# api
+
+def DocumentReference_SummaryProvider(value, params):
+  return value.GetChildMemberWithName('key_').GetSummary()
+
+
+# Objective-C
+
+def FIRDocumentReference_SummaryProvider(value, params):
+  return value.GetChildMemberWithName('_documentReference').GetSummary()
+
+
 def get_string(value):
   """Returns a Python string from the underlying LLDB SBValue."""
   # TODO(wilhuff): This is gross hack. Actually use the SBData API to get this.
@@ -89,9 +101,14 @@ def __lldb_init_module(debugger, params):
     run('type summary add -w firestore -F {0} {1} {2}'.format(
       qname(provider), args, typename))
 
+  api = 'firebase::firestore::api::'
+  add_summary(DocumentReference_SummaryProvider, api + 'DocumentReference')
+
   model = 'firebase::firestore::model::'
   add_summary(DocumentKey_SummaryProvider, model + 'DocumentKey')
   add_summary(ResourcePath_SummaryProvider, model + 'ResourcePath')
+
+  add_summary(FIRDocumentReference_SummaryProvider, 'FIRDocumentReference')
 
   run('type category enable firestore')
 
--- a/scripts/lldb/firestore.py
+++ b/scripts/lldb/firestore.py
@@ -1,0 +1,101 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+import json
+
+"""
+LLDB type summary providers for common Firestore types.
+
+This is primarily useful for debugging Firestore internals. It will add useful
+summaries and consolidate common types in a way that makes them easier to
+observe in the debugger.
+
+Use this by adding the following to your ~/.lldbinit file:
+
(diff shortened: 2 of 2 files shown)


scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+LLDB type summary providers for common Firestore types.
+
+This is primarily useful for debugging Firestore internals. It will add useful
+summaries and consolidate common types in a way that makes them easier to
+observe in the debugger.
+
+Use this by adding the following to your ~/.lldbinit file:
+
+    command script import ~/path/to/firebase-ios-sdk/scripts/lldb/firestore.py
```
var-const, line 27: I wonder if this would be better extracted to a `README` file to be placed in `scripts/lldb`? Or perhaps leave as is, but duplicate this information in a readme file?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+# model
+
+def DocumentKey_SummaryProvider(value, params):
+  """Summarizes DocumentKey as if path_->segments_ were inline and a single
+  string."""
+  return deref_shared(value.GetChildMemberWithName('path_')).GetSummary()
+
+
+def ResourcePath_SummaryProvider(value, params):
+  """Summarizes ResourcePath as if segments_ is a single string."""
```
var-const, line 155: "as if segments_ _were_ a single string"?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+    view of the data. The children that are synthesized here show up in the
+    graphical debugger.
+"""
+
+
+# model
+
+def DocumentKey_SummaryProvider(value, params):
+  """Summarizes DocumentKey as if path_->segments_ were inline and a single
+  string."""
```
var-const, line 150: Optional: consider adding an example of what the summary might look like.

scripts/lldb/firestore.py
var-const, line 155: Similarly to the comment above, consider adding an example.

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+  return deref_shared(value.GetChildMemberWithName('path_')).GetSummary()
+
+
+def ResourcePath_SummaryProvider(value, params):
+  """Summarizes ResourcePath as if segments_ is a single string."""
+
+  segments = value.GetChildMemberWithName('segments_')
+  count = segments.GetNumChildren()
```
var-const, line 158: Nit: consider inlining this variable (`for i in range(0, segments.GetNumChildren())`).

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+
+  segment_text = []
+  for i in range(0, count):
+    child = segments.GetChildAtIndex(i)
+    segment_text.append(get_string(child))
+
+  text = format_string('/'.join(segment_text))
```
var-const, line 165: Optional: perhaps return the expression and bypass the intermediate variable?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,101 @@
...
+
+def ResourcePath_SummaryProvider(value, params):
+  """Summarizes ResourcePath as if segments_ is a single string."""
+
+  segments = value.GetChildMemberWithName('segments_')
+  count = segments.GetNumChildren()
+
+  segment_text = []
+  for i in range(0, count):
+    child = segments.GetChildAtIndex(i)
```
var-const, line 162: Hmm, looking at `lldb.SBValue`, it seems possible to [iterate](https://lldb.llvm.org/python_reference/lldb-pysrc.html#SBValue.__iter__) over them, so this probably could be something like:
```
segment_text = [get_string(child) for child in segments]
```
(Of course, I might be missing some reason why this wouldn't work)

scripts/lldb/firestore.py
```diff
@@ -103,13 +111,16 @@ def add_summary(provider, typename, *args):
 
   api = 'firebase::firestore::api::'
   add_summary(DocumentReference_SummaryProvider, api + 'DocumentReference')
+  add_summary(DocumentSnapshot_SummaryProvider, api + 'DocumentSnapshot', '-e')
```
var-const, line 232: What does `-e` flag do?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+
+def AbseilOptional_SummaryProvider(value, params):
+  # Operates on the synthetic children above, calling has_children.
+  return 'engaged={0}'.format(format_bool(value.MightHaveChildren()))
+
+
+# model
+
+class DatabaseId_SynthProvider(ForwardingSynthProvider):
+  """Makes DatabaseId behave as if `*rep_` were inline."""
```
var-const, line 134: Nit: perhaps add a little bit of context to make what `rep_` is more immediately obvious, e.g. `Hides the fact that DatabaseId uses PIMPL by making it behave as if...`?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+    child = segments.GetChildAtIndex(i)
+    segment_text.append(get_string(child))
+
+  text = format_string('/'.join(segment_text))
+  return text
+
+
+# api
+
+def DocumentReference_SummaryProvider(value, params):
```
var-const, line 171: Nit: these few summaries lack a doc comment. If the comments are a pain to write, I think just an example is sufficient.

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+    return self.delegate().GetNumChildren()
+
+  def get_child_index(self, name):
+    return self.delegate().GetIndexOfChildWithName(name)
+
+  def get_child_at_index(self, index):
+    return self.delegate().GetChildAtIndex(index)
+
+  def update(self):
+    pass
```
var-const, line 67: Perhaps add a comment to make it clear why this method is a no-op?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+  * Synthetic Children Providers: these are classes that provide an alternative
+    view of the data. The children that are synthesized here show up in the
+    graphical debugger.
+"""
+
+
+class ForwardingSynthProvider(object):
+  """A synthetic child provider that forwards all methods to another provider.
+
+  Override the `delegate` method to customize the target to which this forwards.
```
var-const, line 45: Does using this class on its own (without overriding `delegate`) ever make sense? If no, perhaps `delegate` should throw rather than have a default behavior?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+  def __init__(self, value, params):
+    self.value = value
+    self.engaged = None
+    self.data = None
+
+  def update(self):
+    # Unwrap all the internal optional_data and similar types
+    value = self.value
+    while True:
+      if value.GetNumChildren() <= 0:
```
var-const, line 85: Question: can the number of children actually be less than zero?

scripts/lldb/firestore.py
```diff
@@ -0,0 +1,250 @@
...
+  return value.GetChildMemberWithName('_documentReference').GetSummary()
+
+
+def FIRDocumentSnapshot_SummaryProvider(value, params):
+  return value.GetChildMemberWithName('_snapshot').GetSummary()
+
+
+def get_string(value):
+  """Returns a Python string from the underlying LLDB SBValue."""
+  # TODO(wilhuff): This is gross hack. Actually use the SBData API to get this.
```
var-const, line 191: Question: what's the reason to use a hack? If you expect it to exist for some time, it might be worthwhile to expand the comment a little bit to explain why the hack is in place.

scripts/lldb/firestore.py
wilhuff: I copied the instructions for how to install this into our `Firestore/README.md`. Until there are more of these scripts (and I suspect there won't be) it seems better to keep this self-contained.

scripts/lldb/firestore.py
wilhuff: The summary is just a path string, so fully cooking up an example showing the equivalence between the C++ structure and the result seemed overkill. I've incorporated a trivial example into the text.

scripts/lldb/firestore.py
wilhuff: Done.

wilhuff: Handled as above.

scripts/lldb/firestore.py
wilhuff: Eliminated this by taking the suggestion below.

scripts/lldb/firestore.py
wilhuff: Good idea. Done.

scripts/lldb/firestore.py
wilhuff: Done.

scripts/lldb/firestore.py
wilhuff: Pulling out the string contents requires dealing with the memory access API which is more than I wanted to bite off here. Instead, this gets lldb to format that into a Python string containing the summary and then evaluates the string as if it were a string literal to remote quotes and escape sequences. 

I've reformulated this note.

scripts/lldb/firestore.py
wilhuff: Expands children by default when printing. A `DocumentReference` can straightforwardly be summarized with just the path. `DocumentSnapshot` has more going on, so this makes the children visible by default when printing.

scripts/lldb/firestore.py
wilhuff: I've almost convinced myself that a forwarding synth provider is a bad idea (it masks the actual path to the variables). I'd prefer not to invest much more in it until I'm more certain this will stick around.

Using this in its default configuration does no harm; it's just a no-op.

scripts/lldb/firestore.py
wilhuff: Done.

scripts/lldb/firestore.py
wilhuff: The contract isn't well specified and I've seen some implementations use negative numbers to indicate an error.

scripts/lldb/firestore.py
wilhuff: This is sort of veering into overkill. I don't really want to comment on the rationale for why something is laid out the way it is here. I did add some text to indicate that this hides the implementation details.

scripts/lldb/firestore.py
wilhuff: Given that these were one-liners indicating that the summary was derived from a single field I considered these fairly self-explanatory.

I've added comments but I suspect I'm not quite capturing what you're after.

Firestore/README.md
```diff
@@ -10,6 +10,17 @@
   * Select the Firestore_Tests_iOS scheme
   * ⌘-u to build and run the unit tests
 
+### Improving the debuger experience
```
var-const, line 13: Nit: s/debuger/debugger/

scripts/lldb/firestore.py
var-const: I like the new comments, thanks for adding them.

Firestore/README.md
wilhuff: 🤦

PR #5358, opened 2020-04-14
add a token refresh integration test in messaging

chliangGoogle:

PR #5408, opened 2020-04-17
Fix issues in Messaging GHA yml file

paulb777:
Fix build breakage. I'm not sure how/why this worked in #5358

#no-changelog


chliangGoogle:
I think messaging build wasn't triggered in #5358, which is why this error isn't shown.

PR #5411, opened 2020-04-20
M69 changelog for GUL and FIS

maksymmalyhin:


FirebaseInstallations/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v1.1.2 -- M69
```
paulb777, line 1: Should have been a minor version update since dependencies are changing. Probably not worth fixing this late though.

PR #5413, opened 2020-04-20
Cleanup logging and error handling for initial auth persistence implementation

schnecle:
* Remove NSLog statements from AuthPersistence
* Add error pointers to all calls into AuthPersistence
* Add logging to calling methods based on errors bubbled up but do not bubble errors up to the user
* Add TODO's for adding in the FIRLogger

### Testing

  * Running locally e2e - will post screen grab

### API Changes

  * N/A

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -61,7 +61,12 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
     [GULAppDelegateSwizzler registerAppDelegateInterceptor:interceptor];
   }
 
-  self.authState = [FIRAppDistributionAuthPersistence retrieveAuthState];
+  NSError *authRetrievalError;
+  self.authState = [FIRAppDistributionAuthPersistence retrieveAuthState error:&authRetrievalError];
+  //TODO (schnecle): replace NSLog statement with FIRLogger log statement
+  if(authRetrievalError){
+    NSLog(@"Error retrieving token from keychain: %@", [authRetrievalError localizedDescription]);
+  }
   self.isTesterSignedIn = self.authState ? YES : NO;
   return self;
 }
@@ -129,7 +134,12 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
 }
 
 - (void)signOutTester {
-  [FIRAppDistributionAuthPersistence clearAuthState];
+  NSError *error;
+  [FIRAppDistributionAuthPersistence clearAuthState error:&error];
+  // TODO (schnecle): Add in FIRLogger to report when we have failed to clear auth state
+  if(error){
+    NSLog(@"Error clearing token from keychain: %@", [error localizedDescription]);
+  }
   self.authState = nil;
   self.isTesterSignedIn = false;
 }
@@ -140,6 +150,7 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
                                                NSString *_Nonnull idToken,
                                                NSError *_Nullable error) {
       if (error) {
+        // TODO (schnecle): Add in FIRLogger log statement
         NSLog(@"Error fetching fresh tokens: %@", [error localizedDescription]);
         [self signOutTester];
         return;
@@ -187,7 +198,7 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
        appDistributionSignInCompletion:(void (^)(NSError *_Nullable error))completion {
   if (!configuration) {
     // TODO: Handle when we cannot get configuration
-      NSLog(@"ERROR - Cannot discover oauth config");
+    NSLog(@"ERROR - Cannot discover oauth config");
     @throw([NSException exceptionWithName:@"NotImplementedException"
                                    reason:@"This code path is not implemented yet"
                                  userInfo:nil]);
@@ -214,8 +225,16 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
                                                      callback:^(OIDAuthState *_Nullable authState,
                                                                 NSError *_Nullable error) {
           self.authState = authState;
+
+          // Capture errors in persistence but do not bubble them up
+          NSError *authPersistenceError;
           if(authState) {
-              [FIRAppDistributionAuthPersistence persistAuthState:authState];
+            [FIRAppDistributionAuthPersistence persistAuthState:authState error:&authPersistenceError];
+          }
+
+          // TODO (schnecle): Log errors in persistence using FIRLogger
+          if(authPersistenceError) {
+            NSLog(@"Error persisting token to keychain: %@", [error localizedDescription]);
           }
           self.isTesterSignedIn = self.authState ? YES : NO;
           completion(error);
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 #import "FIRAppDistribution+Private.h"
-#import "FIRAppDistributionRelease+Private.h"
-#import "FIRAppDistributionMachO+Private.h"
 #import "FIRAppDistributionAuthPersistence+Private.h"
+#import "FIRAppDistributionMachO+Private.h"
+#import "FIRAppDistributionRelease+Private.h"
 
 #import <FirebaseCore/FIRAppInternal.h>
 #import <FirebaseCore/FIRComponent.h>
@@ -62,9 +62,9 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
   }
 
   NSError *authRetrievalError;
-  self.authState = [FIRAppDistributionAuthPersistence retrieveAuthState error:&authRetrievalError];
-  //TODO (schnecle): replace NSLog statement with FIRLogger log statement
-  if(authRetrievalError){
+  self.authState = [FIRAppDistributionAuthPersistence retrieveAuthState:&authRetrievalError];
+  // TODO (schnecle): replace NSLog statement with FIRLogger log statement
+  if (authRetrievalError) {
     NSLog(@"Error retrieving token from keychain: %@", [authRetrievalError localizedDescription]);
   }
   self.isTesterSignedIn = self.authState ? YES : NO;
@@ -135,9 +135,9 @@ NSString *const kAppDistroLibraryName = @"fire-fad";
 
 - (void)signOutTester {
   NSError *error;
-  [FIRAppDistributionAuthPersistence clearAuthState error:&error];
+  [FIRAppDistributionAuthPersistence clearAuthState:&error];
   // TODO (schnecle): Add in FIRLogger to report when we have failed to clear auth state
-  if(error){
+  if (error) {
     NSLog(@"Error clearing token from keychain: %@", [error localizedDescription]);
   }
   self.authState = nil;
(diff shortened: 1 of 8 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
+    result = [(__bridge NSData *)passwordData copy];
+  } else if (error) {
+    NSString *desc = NSLocalizedString(
+        @"Failed to retrieve auth state from keychain. Tester will have to sign in again.",
+        @"Error message for failure to retrieve auth state from keychain");
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey : desc};
+    *error = [NSError errorWithDomain:kFIRAppDistributionKeychainErrorDomain
+                                 code:FIRAppDistributionErrorTokenRetrievalFailure
+                             userInfo:userInfo];
+  }
```
samedson, line 57: 2 things:
 - error is a pointer to a pointer. This should probably be checking that the pointer we're setting (*error) is not null. Or else this will always be true.
 - This seems odd to re-set the error when you pass an error to this method. Could it be we forgot to rename a variable that got deleted?

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
+
+  OIDAuthState *authState = nil;
+  if (result) {
+    authState = (OIDAuthState *)[NSKeyedUnarchiver unarchiveObjectWithData:result];
+  } else if (error) {
```
samedson, line 62: Same here with the above 2 points

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
++ (OIDAuthState *)retrieveAuthState:(NSError **_Nullable)error {
+  NSMutableDictionary *keychainQuery = [self getKeyChainQuery];
+  [keychainQuery setObject:(id)kCFBooleanTrue forKey:(id)kSecReturnData];
+  [keychainQuery setObject:(id)kSecMatchLimitOne forKey:(id)kSecMatchLimit];
+  CFDataRef passwordData = NULL;
+  NSData *result = nil;
+  OSStatus status = SecItemCopyMatching((CFDictionaryRef)keychainQuery, (CFTypeRef *)&passwordData);
+
+  if (status == noErr && 0 < [(__bridge NSData *)passwordData length]) {
+    result = [(__bridge NSData *)passwordData copy];
```
samedson, line 48: So it's weird, but make passwordData a `NSData *`, and cast it to `void *` when passing to `SecItemCopyMatching`. https://stackoverflow.com/questions/17822895/how-to-cast-an-indirect-pointer-in-objective-c

Apple's toll-free bridge is only one way here it seems

samedson: In general it's nicer to stick with the NS- APIs than the CF ones so you don't need to deal with memory management

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
samedson: Nevermind - you want to see if we passed in an error variable... Ignore this

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
+  if (status != errSecSuccess && status != errSecItemNotFound && error) {
+    NSString *desc = NSLocalizedString(
+        @"Failed to clear auth state from keychain. Tester will overwrite data on sign in.",
+        @"Error message for failure to retrieve auth state from keychain");
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey : desc};
+    *error = [NSError errorWithDomain:kFIRAppDistributionKeychainErrorDomain
+                                 code:FIRAppDistributionErrorTokenDeletionFailure
+                             userInfo:userInfo];
+  }
+  return errSecSuccess || status == errSecItemNotFound;
```
samedson, line 36: If statement and return values do not line up (isn't `errSecSuccess` a constant?).

There's 2 ways to address this:

1) For these kindof coding patterns, it helps to always return a constant `YES` or `NO`. That way you only need the check in one place. With this pattern you also need to check the `if (error)` elsewhere, and I would recommend making a function to do this everywhere in your codebase, instead of building it into the logic. This requires discipline to always remember to use that function to set `**errors` (eg: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInstallations/Source/Library/Errors/FIRInstallationsErrorUtil.m#L23)
2) The other way is to just make a variable, and use it in the if statement and return value. To me this requires more discipline than 1) 

For here' we'd `return NO` in the error

samedson: For 1), you would audit your code and never write `if (error)`, unless you were checking an `NSError *` (single * pointer)

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
+    authState = (OIDAuthState *)[NSKeyedUnarchiver unarchiveObjectWithData:result];
+  } else if (error) {
+    NSString *desc =
+        NSLocalizedString(@"Failed to unarchive auth state. Tester will have to sign in again.",
+                          @"Error message for failure to copy password data from keychain");
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey : desc};
+    *error = [NSError errorWithDomain:kFIRAppDistributionKeychainErrorDomain
+                                 code:FIRAppDistributionErrorTokenRetrievalFailure
+                             userInfo:userInfo];
+  }
```
samedson, line 70: To me it's odd that this will overwrite the error you built on line 54 every time result is null

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
samedson: This advice applies to all the methods in this file that take an `NSError **`

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
```diff
@@ -16,77 +16,99 @@
...
+  if (status == noErr && 0 < [(__bridge NSData *)passwordData length]) {
+    result = [(__bridge NSData *)passwordData copy];
+  } else if (error) {
+    NSString *desc = NSLocalizedString(
+        @"Failed to retrieve auth state from keychain. Tester will have to sign in again.",
+        @"Error message for failure to retrieve auth state from keychain");
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey : desc};
+    *error = [NSError errorWithDomain:kFIRAppDistributionKeychainErrorDomain
+                                 code:FIRAppDistributionErrorTokenRetrievalFailure
+                             userInfo:userInfo];
```
samedson, line 56: It's usually nice to return after you construct an error, if possible. That prevents you from later overwriting the error.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -129,65 +134,69 @@ - (void)signInTesterWithCompletion:(void (^)(NSError *_Nullable error))completio
...
+            }
+
+            NSHTTPURLResponse *HTTPResponse = (NSHTTPURLResponse *)response;
+
+            if (HTTPResponse.statusCode == 200) {
+              [self handleReleasesAPIResponseWithData:data completion:completion];
+            } else {
+              // TODO: Handle non-200 http response
+              NSLog(@"ERROR - Non 200 service response - %@", HTTPResponse);
+              @throw([NSException exceptionWithName:@"NotImplementedException"
```
samedson, line 184: I'm assuming this is just a reminder, but would recommend to never ever throw in Objective-C

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
schnecle: 👍

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: Pranav is actively working on this 👍

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
schnecle: Added an error handling method to this class to leverage in all of the places that we are populating this error and restructured the code a bit to make it so we return immediately.

FirebaseAppDistribution/Sources/FIRAppDistributionAuthPersistence.m
schnecle: This really cleaned up the code 👍

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -129,65 +134,69 @@ - (void)signInTesterWithCompletion:(void (^)(NSError *_Nullable error))completio
...
+    NSURLSessionDataTask *listReleasesDataTask = [URLSession
+        dataTaskWithRequest:request
+          completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
+            if (error) {
+              // TODO: Reformat error into error code
+              completion(nil, error);
+              return;
+            }
+
+            NSHTTPURLResponse *HTTPResponse = (NSHTTPURLResponse *)response;
```
samedson, line 177: These seem to be the same type you're casting to

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -129,65 +134,69 @@ - (void)signInTesterWithCompletion:(void (^)(NSError *_Nullable error))completio
...
-                  @throw([NSException exceptionWithName:@"NotImplementedException"
-                                                 reason:@"This code path is not implemented yet"
-                                               userInfo:nil]);
-                }
-              }];
-
-        [listReleasesDataTask resume];
-    }];
+    [listReleasesDataTask resume];
+  }];
```
samedson, line 191: JFYI: The formatter doesn't seem to play well with blocks (seems to horrendously indent them). Since this is double indented it might fly off the page after you run `style.sh`.

I sometimes resolve this by moving the whole block into a method.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -209,20 +218,33 @@ - (void)handleOauthDiscoveryCompletion:(OIDServiceConfiguration *_Nullable)confi
...
+                                             }
+
+                                             // TODO (schnecle): Log errors in persistence using
+                                             // FIRLogger
+                                             if (authPersistenceError) {
+                                               NSLog(@"Error persisting token to keychain: %@",
+                                                     [error localizedDescription]);
+                                             }
+                                             self.isTesterSignedIn = self.authState ? YES : NO;
+                                             completion(error);
```
samedson, line 244: If there's an `authPersistenceError` but not an `error`, do we want to surface that in this completion handler?

samedson: You can also apply the same pattern here where you check `if (error)` at the very beginning of this block, and if so, `self.isTesterSignedIn = NO; self.authState = nil; completion(error)`, then do the same thing right after the call to `persistAuthState`

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -240,28 +262,27 @@ - (UIWindow *)createUIWindowForLogin {
...
+      NSString *codeHash = [releaseDict objectForKey:@"codeHash"];
+      NSString *executablePath = [[NSBundle mainBundle] executablePath];
+      FIRAppDistributionMachO *machO =
+          [[FIRAppDistributionMachO alloc] initWithPath:executablePath];
+
+      if (![codeHash isEqualToString:machO.codeHash]) {
+        FIRAppDistributionRelease *release =
+            [[FIRAppDistributionRelease alloc] initWithDictionary:releaseDict];
+        dispatch_async(dispatch_get_main_queue(), ^{
+          completion(release, nil);
```
samedson, line 279: Is it ok that this sometimes is never called?

samedson: This method should be pretty easy to write tests for

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: For this case, we explicitly do not want to surface this to the user or to the developer because the authState may be valid for this session even though there was an issue with persistence. So we want to log it, but not actually raise it because the action we would want a developer to take in this case are pretty unclear. I think it is worth logging from an audit perspective so that when a developer writes in to us saying that their users have to sign in every time they open the app we can get some signal on where it's going wrong.

schnecle:
I've added these comments to b/154110693 and will follow up in a subsequent PR to add unittests and clean up some of the completion blocks.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: Will follow up in a subsequent PR to this feature branch

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: Will follow up in a subsequent PR to this branch.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: We believe these are different types, but we'll check it out and see.

Issue #5020, opened 2020-03-04
AppStore Distribution / Release app does not send crashes due to Exported Symbols File missing __mh_execute_header


kondratk:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.3.1
  * Firebase SDK version:  Firebase/Crashlytics (6.15.0):
  * Firebase Component: Crashlytics
  * Component version: 4.0.0-beta.4
  * Installation method: `CocoaPods` (select one)

### [REQUIRED] Step 2: Describe the problem

We've identified, that after signing the app with AppStore certificate, Firebase Crashlytics stops working. The problem does not occur while signing the app with AdHoc certificate. We are using Apple Distribution Certificate.

The errors we've found in the console:
<Forced crash, re-ran the app>
[Firebase/Crashlytics] Version 4.0.0-beta.4
..
[Firebase/Crashlytics][I-CLS000000] An application must have a valid bundle identifier in its Info.plist
[Firebase/Crashlytics][I-CLS000000] Crash reporting could not be initialized
[Firebase/InstanceID][I-IID003014] Error while reading embedded mobileprovision Error Domain=NSCocoaErrorDomain Code=260 "The file “embedded.mobileprovision” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/private/var/containers/Bundle/Application/71738F66-30A3-4D4E-A60E-85F7FB3614AC/X.app/embedded.mobileprovision, NSUnderlyingError=0x282bc5b00 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}

I've unzipped the ipa file and:
- there is Info.plist, it contains bundle identifier
- there is mobileprovision file, looks good.

The app would not even run without files above, so these errors seems like incorrect ones.

#### Steps to reproduce:

1) Build the app with AdHoc profile, force a crash - works fine. Crashes are sent to Crashlytics.
2) Build the app with AppStore profile, distribute via testflight.
3) Install it from TestFlight, run, force a crash
4) Re-open the app. Crash should be sent- it is not. Check out logs:
5) Errors from Step 2 are visible.

JayachandraA:
Crashlytics, I'm also experiencing same problem. This is sending me the crash report in debug mode, but when I use signing the build and push to TestFlight it is not sending me the report.

yasirmturk:
Same here 👎

SeeTheInvisible:
We also have the same problem

[...]

PR #5414, opened 2020-04-20
Crashlytics/issues/5020

samedson:
Reported in: https://github.com/firebase/firebase-ios-sdk/issues/5020

Adds a clear error message for customers to self-resolve Crashlytics not starting up (usually in Release Builds) because Exported Symbols File is set in Build Settings, and the file is missing `__mh_execute_header`.

### DIFF
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
@@ -527,7 +527,7 @@ static void (^reportSentCallback)(void);
   // failing computeExecutableInfo. This can happen if the user sets the
   // Exported Symbols File in Build Settings, and leaves off the one symbol
   // that Crashlytics needs, "__mh_execute_header" (wich is defined in mach-o/ldsyms.h as
-  // _MH_EXECUTE_SYM) From https://github.com/firebase/firebase-ios-sdk/issues/5020
+  // _MH_EXECUTE_SYM). From https://github.com/firebase/firebase-ios-sdk/issues/5020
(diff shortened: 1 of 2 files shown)


Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -522,7 +522,20 @@ - (void)crashReportingSetupCompleted:(FIRCLSProfileMark)mark {
...
+  // When the ApplicationIdentifierModel fails to initialize, it is usually due to
+  // failing computeExecutableInfo. This can happen if the user sets the
+  // Exported Symbols File in Build Settings, and leaves off the one symbol
+  // that Crashlytics needs, "__mh_execute_header" defined as _MH_EXECUTE_SYM
+  // From https://github.com/firebase/firebase-ios-sdk/issues/5020
+  if (!self.appIDModel) {
+    FIRCLSErrorLog(
+        @"Crashlytics could not find the main file symbol, and cannot start up. This can happen "
+        @"when Exported Symbols File is set in Build Settings. To resolve this, add "
+        @"\"__mh_execute_header\" as a newline to your Exported Symbols File.");
```
paulb777, line 535: The message seems inconsistent with the comment.

Reading the comment, I would think I need to add:
`__mh_execute_header=_MH_EXECUTE_SYM`

Reading the message:
`__mh_execute_header`

samedson: I'll update the comment - should be just `__mh_execute_header`

PR #5420, opened 2020-04-20
Manually specify the authorization header

### LINKED ISSUE #5072: Firestore Emulator Security Rule `request.auth` is always null even after authed

  * Xcode version: 11.3.1
  * Firebase SDK version: 6.19.0
  * Firebase Component: Auth, Firestore
  * Component version: Auth: 6.19.0, Firestore: 6.19.0

Firestore Emulator Security Rule `request.auth` is always null even after authed.

I tried it on Android project/emulator, and it works fine, so I suspect  it is iOS SDK's bug.

#### Steps to reproduce:

I created the reproduced project: https://github.com/mono0926/firestore-emulator-sample

#### 1. Prepare Firestore Emulator

Execute this:

```sh
git clone https://github.com/mono0926/firestore-emulator-sample.git
cd firestore-emulator-sample/firebase
firebase emulators:start
```

If you create your own Firebase project, these steps are needed:

Enable Anonymous Auth:
image

Enable Firestore:
image

#### 2. Run iOS application

Execute this:

```sh
cd firestore-emulator-sample/iOS/ 
pod install
open EmulatorSample.xcworkspace
```

And Run app on iOS simulator, and push the button:

image

The error log will be output:

```
error: Optional(Error Domain=FIRFirestoreErrorDomain Code=7 "
false for 'get' @ L5" UserInfo={NSLocalizedDescription=
false for 'get' @ L5})
```

##### Security Rule

This security rule allows `/users/{userId}` `get` only if `request.auth != null`

https://github.com/mono0926/firestore-emulator-sample/blob/633cd0d48c8b6eaf228d1fb3f978420df1fc4e6d/firebase/firestore.rules#L5

```json
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    match /users/{userId} {
      allow get: if request.auth != null;
    }
  }
}
```

##### iOS initialization

At `didFinishLaunchingWithOptions`, set to connect to Firestore emulator.

https://github.com/mono0926/firestore-emulator-sample/blob/633cd0d48c8b6eaf228d1fb3f978420df1fc4e6d/iOS/EmulatorSample/AppDelegate.swift#L20-L24

```swift
        let settings = Firestore.firestore().settings
        settings.host = "localhost:8080"
        settings.isPersistenceEnabled = false
        settings.isSSLEnabled = false
        Firestore.firestore().settings = settings
```

##### iOS logic

1. `Auth.auth().signInAnonymously`
1. ` Firestore.firestore().collection("users").document(uid).getDocument()`

https://github.com/mono0926/firestore-emulator-sample/blob/32f5c5bdf59eba8098c3bede21c6da430568b799/iOS/EmulatorSample/ViewController.swift#L19-L39

```swift
    @IBAction func buttonDidTap(_ sender: Any) {
        Auth.auth().signInAnonymously { (result, error) in
            assert(error == nil)
            
            guard let result = result else {
                assert(false)
                return;
            }
            
            let uid = result.user.uid
            print(uid)
            Firestore.firestore().collection("users").document(uid).getDocument { (snap, error) in
//                "snap: nil"
                print("snap: \(snap?.data())")
//                error: Optional(Error Domain=FIRFirestoreErrorDomain Code=7 "
//                false for 'get' @ L5" UserInfo={NSLocalizedDescription=
//                false for 'get' @ L5})
                print("error: \(error)")
            }
        }
    }
```

This should succeed without any errors because the user is authed and firestore security rule only checks `request.auth != null` and this condition is satisfied.

If [setting to connect to emulator codes](https://github.com/mono0926/firestore-emulator-sample/blob/633cd0d48c8b6eaf228d1fb3f978420df1fc4e6d/iOS/EmulatorSample/AppDelegate.swift#L20-L24) is commented out, the app will connect to real Firestore backend(same security rule deployed), and it works as expected.

wilhuff:
This matches the [Android](https://github.com/firebase/firebase-android-sdk/blob/205b52314bfc1b6bac44958e5411134d6003a85c/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreCallCredentials.java#L55) and [JS](https://github.com/firebase/firebase-js-sdk/blob/ed9a7be7d5f65aa37c0c478021b15bd18166a515/packages/firestore/src/platform_browser/webchannel_connection.ts#L96) SDKs and works around a limitation in gRPC, whereby it silently drops credentials if the connection isn't using TLS. In Firestore's case, developers routinely run against an emulator on the local host with TLS disabled (because requiring developers to provision certificates for local development is too onerous).

Fixes #5072.

### DIFF
--- a/Firestore/core/src/remote/grpc_connection.cc
+++ b/Firestore/core/src/remote/grpc_connection.cc
@@ -32,6 +32,7 @@
 #include "Firestore/core/src/util/statusor.h"
 #include "Firestore/core/src/util/string_format.h"
 #include "absl/memory/memory.h"
+#include "absl/strings/str_cat.h"
 #include "grpcpp/create_channel.h"
 
 namespace firebase {
@@ -49,6 +50,7 @@ using util::StringFormat;
 
 namespace {
 
+const char* const kAuthorizationHeader = "authorization";
 const char* const kXGoogAPIClientHeader = "x-goog-api-client";
 const char* const kGoogleCloudResourcePrefix = "google-cloud-resource-prefix";
 
@@ -151,7 +153,7 @@ std::unique_ptr<grpc::ClientContext> GrpcConnection::CreateContext(
 
   auto context = absl::make_unique<grpc::ClientContext>();
   if (token.data()) {
-    context->set_credentials(grpc::AccessTokenCredentials(MakeString(token)));
+    context->AddMetadata(kAuthorizationHeader, absl::StrCat("Bearer ", token));
   }
 
   // TODO(dimond): This should ideally also include the gRPC version, however,
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -2,6 +2,8 @@
 - [changed] Firestore now limits the number of concurrent document lookups it
   will perform when resolving inconsistencies in the local cache
   (https://github.com/firebase/firebase-js-sdk/issues/2683).
+- [fixed] Firestore will now send Auth credentials to the Firestore Emulator
+  (#5072).
 
 # v1.12.1
 - [changed] Internal improvements for future C++ and Unity support.


Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -151,7 +153,7 @@ std::unique_ptr<grpc::ClientContext> GrpcConnection::CreateContext(
 
   auto context = absl::make_unique<grpc::ClientContext>();
   if (token.data()) {
-    context->set_credentials(grpc::AccessTokenCredentials(MakeString(token)));
```
var-const, line 154: Is it safe to not call `set_credentials`; in other words, is it guaranteed that the only thing `set_credentials` does is to add the header? Otherwise, perhaps the header should be added _in addition_ to calling `set_credentials`, not instead of it?

wilhuff: It's safe to not call `set_credentials`. The only observable effect it has on the HTTP messages that are sent is to add the `authorization` header.

Issue #5152, opened 2020-03-19
-[FIRInstanceIDTokenInfo isFreshWithIID:]: unrecognized selector sent to instance


AndreasLS:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 10.1
  * Firebase SDK version: 6.20.0
  * Firebase Component: FirebaseInstanceID
  * Component version: 4.3.2 (was 4.2.6)
  * Installation method: CocoaPods

#### Steps to reproduce:

After updating from 6.11 to 6.20, it started to crash and reported the error:

-[FIRInstanceIDTokenInfo isFreshWithIID:]: unrecognized selector sent to instance

I did some search and find out that in version 6.12, the method isFreshWithIID: came as replacement to the previous isFresh.

I put some breakpoint at the crash point and printed the object to check if it responds to the new selector and old selector, then I found this:

(lldb) po [tokenInfo respondsToSelector:@selector(isFresh)]
YES

(lldb) po [tokenInfo respondsToSelector:@selector(isFreshWithIID)]
NO

It seems that the object archived in keychain is an old version of the object, that doesn't have the new method. Since its use the NSKeyedUnarchiver to unarchive the object, it remains the object that was saved before, crashing when trying to access the new method.

#### Relevant Code:

```
//FIRInstanceIDTokenStore.m

+ (nullable FIRInstanceIDTokenInfo *)tokenInfoFromKeychainItem:(NSData *)item {
  // Check if it is saved as an archived FIRInstanceIDTokenInfo, otherwise return nil.
  FIRInstanceIDTokenInfo *tokenInfo = nil;
  // NOTE: Passing in nil to unarchiveObjectWithData will result in an iOS error logged
  // in the console on iOS 10 and below. Avoid by checking item.data's existence.
  if (item) {
    // TODO(chliangGoogle: Use the new API and secureCoding protocol.
    @try {
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
      tokenInfo = [NSKeyedUnarchiver unarchiveObjectWithData:item];
#pragma clang diagnostic pop

    } @catch (NSException *exception) {
      FIRInstanceIDLoggerDebug(kFIRInstanceIDMessageCodeTokenStoreExceptionUnarchivingTokenInfo,
                               @"Unable to parse token info from Keychain item; item was in an "
                               @"invalid format");
      tokenInfo = nil;
    } @finally {
    }
  }
[...]

PR #5441, opened 2020-04-22
Edit Firebase from frameworks tech note

markarndt:
Cheers.


docs/firebase_in_libraries.md
```diff
@@ -1,31 +1,28 @@
...
-(especially when different versions of the static framework are linking to the
-app and the dynamic framework). For example, a `dispatch_once` may or may not do
-the right initialization since there are now two entities to initialize. Here
-are couple more examples of issues related to this undefined behavior:
+A dynamic framework is a bundle containing dynamic libraries and other resources. The dynamic
+libraries bundled in the framework can themselves be compiled from static libraries, like
+Firebase core and product libraries, meaning all of the symbols of the static libraries are
+part of the dynamic framework bundle. What if your core app already directly links a static
+Firebase library when you link to the same library indirectly from a dynamic framework? This
+leads to undefined behavior (especially when different versions of the static framework are
```
maksymmalyhin, line 18: From the first glance it doesn't look exactly clear that the undefined behaviour is a result two(or more) copies of symbols are present in its runtime when such an application is launched. Do you think we should/can reword it to highlight this fact?

docs/firebase_in_libraries.md
```diff
@@ -1,31 +1,28 @@
...
+linked to the app and the dynamic framework). For example, a `dispatch_once` may or may not
+perform the correct initialization since there are now two entities to initialize. Here are
+a couple more examples of issues related to this undefined behavior:
 [#4315](https://github.com/firebase/firebase-ios-sdk/issues/4315),
```
maksymmalyhin, line 22: As suggested at b/154619039 we may add the exception message from issues like [this](https://stackoverflow.com/questions/60753233/the-default-firebaseapp-instance-must-be-configured-before-the-defaultfirebaseap) to make it searchable:
"The default FirebaseApp instance must be configured before the defaultFirebaseApp instance can be initialized".

docs/firebase_in_libraries.md
```diff
@@ -1,31 +1,28 @@
...
-[#5152](https://github.com/firebase/firebase-ios-sdk/issues/4315), etc.
+[#5152](https://github.com/firebase/firebase-ios-sdk/issues/4315).
```
maksymmalyhin, line 23: The link should be https://github.com/firebase/firebase-ios-sdk/issues/5152 ?

markarndt: Git it.

maksymmalyhin: Spotting my own errors 😄

docs/firebase_in_libraries.md
markarndt: Thanks!

docs/firebase_in_libraries.md
markarndt: Woops missed that from the gDoc. Got it!

maksymmalyhin: Thanks!

PR #5445, opened 2020-04-22
create firebase messaging folder

chliangGoogle:
To follow the current folder style in Firebase.


paulb777:
Other option would be update scripts/check_no_module_imports.sh to exclude FirebaseMessaging/Tests

chliangGoogle:
Does it have to use module? I just update the import syntax.

paulb777:
Yep that's fine to change the syntax.  The CI check is really meant for the libraries though which aren't allowed to have module imports. In theory, `@import` is fine in tests.

chliangGoogle:
- Remove FirebaseMessaging/Tests/App

Didn't know this is no longer in use

- Can be in a later PR - move Example/Messaging/Sample to FirebaseMessaging/Tests/
Yeah I plan to move this after Ryan's session this friday

chliangGoogle:
yeah, that's what I will work on next. We also need to adjust internal test app path, etc.

PR #5471, opened 2020-04-27
Fix Docs Regarding Carthage Integration

Blackjacx:
`$(SRCROOT)/$(BUILT_PRODUCTS_DIR)` is not a valid directory combination as far as I know. In my project I use just $(BUILT_PRODUCTS_DIR) and it seems to work.


Blackjacx:
Somebody has to update this also on https://firebase.google.com/docs/crashlytics/get-started-new-sdk

paulb777:
@Blackjacx Thanks for the PR. Please sign the CLA and we'll take a look. Otherwise based on the description, we might be able to handle this as an issue instead.

paulb777:
Replaced by #5472

PR #5478, opened 2020-04-27
Initial dummy publisher for Combine experimentation.

ryanwilson:
This currently fails `pod lib lint` but seems to work in usage. Opened the PR so folks can continue to work on it today if desired, or I'll continue tomorrow.

### DIFF
--- a/Firestore/Swift/Source/Combine/DocumentReference+Combine.swift
+++ b/Firestore/Swift/Source/Combine/DocumentReference+Combine.swift
@@ -15,13 +15,13 @@
  */
 
 #if canImport(Combine)
-import Combine
+  import Combine
 #endif
 import Foundation
 import FirebaseFirestore
 
 extension DocumentReference {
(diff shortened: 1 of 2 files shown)


Firestore/Swift/Source/Combine/DocumentReference+Combine.swift
```diff
@@ -0,0 +1,28 @@
...
+
+#if canImport(Combine)
+  import Combine
+#endif
+import Foundation
+import FirebaseFirestore
+
+extension DocumentReference {
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+  func dummyPublisher() -> AnyPublisher<Int, Never> {
```
peterfriese, line 25: This needs to be public.

ryanwilson: Right, thanks! Had this on Storage but not here :)

PR #5483, opened 2020-04-28
Integration tests need SpecStaging

paulb777:
Fix build failures caused by needing pre-released nanopb and FirebaseAnalytics

#no-changelog


maksymmalyhin:
Slightly unrelated question. I wonder what prevents us from running the integration tests from `pod lib lint` ?

paulb777:
These are tests that depend on Sample Apps - so we'd need set up CocoaPods app_specs as well and likely fix some CocoaPods feature gaps and bugs.

PR #5484, opened 2020-04-28
Fixed linting for Combine APIs.

ryanwilson:
This will likely cause a merge conflict if you've added anything to this class. In that case, I'm fine just closing this once a new API is added with this change integrated as well (moving the `#endif` to the end of the file)


peterfriese:
Close this in favour of the work in https://github.com/firebase/firebase-ios-sdk/projects/3 , @ryanwilson ?

PR #5487, opened 2020-04-28
Introduce Query.getDocument() publisher

peterfriese:
Here are two proposals for a publisher for Query.getDocuments().

There are two versions, one yielding `QuerySnapshot`s, the other one yielding `DocumentSnapshot`s.

Here is a code snippet using the traditional APIs:

```swift
    db.collection("books").getDocuments { (querySnapshot, error) in
      guard let documents = querySnapshot?.documents else {
        print("No documents")
        return
      }

      self.books = documents.compactMap { queryDocumentSnapshot -> Book? in
        return try? queryDocumentSnapshot.data(as: Book.self)
      }
    }
```

Here is how this looks with the publisher yielding `QuerySnapshot`s:

```swift
    let noBooks = [Book]()
    db.collection("books").getDocuments()
      .map { querySnapshot in
        querySnapshot.documents.compactMap { (queryDocumentSnapshot) in
          return try? queryDocumentSnapshot.data(as: Book.self)
        }
      }
      .replaceError(with: noBooks)
      .assign(to: \.books, on: self)
      .store(in: &cancellables)
```

And here is how it looks for the publisher yielding `DocumentSnapshot`s:

```swift
    db.collection("books").getDocuments2()
      .map { documentSnapshots in
        documentSnapshots.compactMap { documentSnapshot -> Book? in
          return try? documentSnapshot.data(as: Book.self)
        }
      }
      .replaceError(with: noBooks)
      .assign(to: \.books, on: self)
      .store(in: &cancellables)
```

### DIFF
--- a/Firestore/Swift/Source/Combine/Query+Combine.swift
+++ b/Firestore/Swift/Source/Combine/Query+Combine.swift
@@ -27,8 +27,8 @@ extension Query {
    * Reads the documents matching this query.
    *
    * This method returns a publisher that yields an array
-   * of `DocumentSnapshot`s, allowing the user to easily iterate over
-   * the documents themselves:
+   * of `QuerySnapshot`s, requiring the user to extract the underlying
+   * `DocumentSnapshot`s before using them:
    *
    * ```
    * let noBooks = [Book]()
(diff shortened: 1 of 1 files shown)


wilhuff:
Please make sure to target a non-master branch

PR #5490, opened 2020-04-28
Update integration testing docs

wilhuff:
Prefer the emulator by default but give alternate instructions for prod.

### DIFF
--- a/Firestore/README.md
+++ b/Firestore/README.md
@@ -23,9 +23,9 @@ command script import ~/path/to/firebase-ios-sdk/scripts/lldb/firestore.py
 
 ### Running Integration Tests
 
-By default, run integration tests against the Firestore Emulator. This is much
-faster than running against production and does not require you to configure a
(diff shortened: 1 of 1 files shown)


Firestore/README.md
```diff
@@ -23,35 +23,30 @@ command script import ~/path/to/firebase-ios-sdk/scripts/lldb/firestore.py
 
 ### Running Integration Tests
 
+By default, run integration tests against the Firestore Emulator. This is much
```
dconeybe, line 26: Consider rewording this sentence to begin with "Prefer running the integration tests..."

I find that by saying "By default" this suggests that it is a default setting that you get unless you specify otherwise. But if I understand correctly, the purpose of this sentence is to encourage people to use the emulator in lieu of production.

wilhuff: Done!

(Note that it kind of is a default setting--out of the box the the source tree is configured for running against the emulator.)

PR #5493, opened 2020-04-28
Minimal test_spec for FirebaseFirestoreSwift

paulb777:
#no-changelog

### DIFF
--- a/FirebaseFirestoreSwift.podspec
+++ b/FirebaseFirestoreSwift.podspec
@@ -37,9 +37,27 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   ]
 
   s.dependency 'FirebaseFirestore', '~> 1.6', '>= 1.6.1'
+  s.dependency 'GoogleTest', '1.10.0'
 
-  s.test_spec 'unit' do |unit_tests|
-    # The other Swift tests require FSTDocRef, gtest, and other dependencies.
-    unit_tests.source_files = 'Firestore/Swift/Tests/API/*.swift'
+  s.test_spec 'unit' do |int_tests|
+    int_tests.source_files = 'Firestore/Swift/Tests/API/*.swift',
+                             'Firestore/Swift/Tests/Codable/*.swift',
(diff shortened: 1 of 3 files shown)


FirebaseFirestoreSwift.podspec
```diff
@@ -37,4 +37,9 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   ]
 
   s.dependency 'FirebaseFirestore', '~> 1.6', '>= 1.6.1'
+
+  s.test_spec 'unit' do |unit_tests|
+    # The other Swift tests require FSTDocRef, gtest, and other dependencies.
+    unit_tests.source_files = 'Firestore/Swift/Tests/API/*.swift'
```
wilhuff, line 43: I'm not enthusiastic about this. It suggests that this is a valid way to test this pod, but it's really not. It's skipping all the hard parts that do most of the heavy lifting.

The way to test this code is via the Firestore/Example/Podfile, which fully exercises all of it.

paulb777: Yep, I was hoping that it would be feasible to include more of the tests here.

The counter argument is that the test spec will lower the bar to contribute Swift-only extensions.  @ryanwilson Any insights about this after investigating Firestore and combine?

wilhuff: There are no Swift-only things you can write that don't in some way depend upon the core Firestore API. The only way to write those kinds of tests today is to use the test helpers out of the Firestore testutils.

paulb777:
I'm going to close it here with all the Swift tests back in the podspec, as well as a dependency to a copy of GoogleTest.podspec in the root.  The build fails with five undefined references to `"firebase::firestore::testutil::Field(absl::lts_2020_02_25::string_view)"`

Next steps if this gets picked back up:

- Resolve undefined symbol
- Merge the two GoogleTest.podspec's
- Update `pod gen` version to handle local dependencies from testspecs. (should be done by updating Firebase from using 1.6.0 to 2.0.0)

PR #5498, opened 2020-04-28
Fix Firestore build broken by M70 version changes

paulb777:
#no-changelog


paulb777:
cc: @mikehaney24 See the question above.

mikehaney24:
> LGTM
> 
> Can we merge GoogleDataTransport and GoogleDataTransportCCTSupport? When would you ever want one without the other? (I should add that I mean this for the future--not this PR.)

The idea is that one could implement their own backend and share the bulk of the transport code. There are plans for this in the works, so I'd prefer to keep them separate for now and reassess later.

PR #5500, opened 2020-04-29
Proposal for a snapshot publisher

peterfriese:
Proposal for a snapshot publisher.

Sample code:

Without Combine:
```swift
    listenerRegistration = db.collection("books").addSnapshotListener { (querySnapshot, error) in
      guard let documents = querySnapshot?.documents else {
        print("No documents")
        return
      }

      self.books = documents.compactMap { queryDocumentSnapshot -> Book? in
        return try? queryDocumentSnapshot.data(as: Book.self)
      }
    }
```

With snapshot publisher:
```swift
    db.collection("books")
      .snapshotPublisher()
      .map { querySnapshot in
        return querySnapshot.documents.compactMap { queryDocumentSnapshot -> Book? in
          return try? queryDocumentSnapshot.data(as: Book.self)
        }
      }
      .replaceError(with: noBooks)
      .assign(to: \.books, on: self)
      .store(in: &cancellables)
```

I also added an extension to make mapping "easier":
```swift
    db.collection("books")
      .snapshotPublisher()
      .mapData()
      .replaceError(with: noBooks)
      .assign(to: \.books, on: self)
      .store(in: &cancellables)
```

### DIFF
--- a/Firestore/Swift/Source/Combine/Query+Combine.swift
+++ b/Firestore/Swift/Source/Combine/Query+Combine.swift
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(Combine)
+  import Combine
+
+  import Foundation
+  import FirebaseFirestore
+
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+  extension Query {
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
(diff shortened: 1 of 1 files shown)


Firestore/Swift/Source/Combine/Query+Combine.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public func snapshotPublisher() -> QuerySnapshotPublisher {
+      return QuerySnapshotPublisher(self)
+    }
+  }
+
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+  extension Query.QuerySnapshotPublisher {
+    public func mapData<T: Decodable>() -> AnyPublisher<[T], Error> {
```
peterfriese, line 77: I tried the following to make this method DRY, but failed:

```swift
    public func mapData2<T: Decodable>() -> AnyPublisher<[T], Error> {
      map { querySnapshot in
        querySnapshot.documents.compactMap { queryDocumentSnapshot -> T? in
          return try? queryDocumentSnapshot.data(as: T.Type) // (1)
        }
      }
      .eraseToAnyPublisher()
    }
```

The compiler complained that `Value of type 'NSObject' has no member 'data'` (see comment (1) above) - no idea why...

morganchen12: Can you manually specify the type in the parameter of the closure?
```swift
{ (queryDocumentSnapshot: QueryDocumentSnapshot) -> T? in /* ... */ }
```

peterfriese:
As per our discussion on GVC, I removed the data mapping portion of the code in favour of a solution potentially using `TopLevelDecoder`.

PR #5515, opened 2020-05-01
Add an integration test for calling clearPersistence() on a new Firestore instance

dconeybe:


Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
```diff
@@ -1342,6 +1342,42 @@ - (void)testCanClearPersistenceAfterRestarting {
...
+  [self awaitExpectations];
+  [self deleteApp:app];
+
+  // We restart the app with the same name and options to check that the previous instance's
+  // persistent storage is actually cleared after the restart. Calling [self firestore] here would
+  // create a new instance of firestore, which defeats the purpose of this test.
+  [FIRApp configureWithName:appName options:options];
+  FIRApp *app2 = [FIRApp appNamed:appName];
+  FIRFirestore *firestore2 = [self firestoreWithApp:app2];
+  [firestore2
```
dconeybe, line 1366: Note that the only material difference between this test case and the one above it is that the location of the call to clearPersistenceWithCompletion() has moved to be invoked on the newly-created firestore2 object.

dconeybe:
Note that the same unit test is being add to the js repository: https://github.com/firebase/firebase-js-sdk/pull/3005

Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
```diff
@@ -1320,7 +1320,7 @@ - (void)testCanClearPersistenceAfterRestarting {
   [self terminateFirestore:firestore];
   [self.firestores removeObject:firestore];
   [firestore
-      clearPersistenceWithCompletion:[self completionForExpectationWithName:@"Enable network"]];
+      clearPersistenceWithCompletion:[self completionForExpectationWithName:@"ClearPersistence"]];
```
thebrianchen, line 1323: Thanks for catching that!

PR #5524, opened 2020-05-04
Fixes for GHA from a forked repo

### LINKED ISSUE #5523: Broken GHA from forked PRs
The have_secrets script is failing and causing tests that require secrets to run. See https://github.com/firebase/firebase-ios-sdk/pull/5520/checks?check_run_id=643479328.

It looks like the GITHUB Actions environment variables are no longer including the repo:

```
++ source scripts/check_secrets.sh
+++ set -x
+++ echo 'GITHUB_BASE_REF: master'
+++ echo 'GITHUB_HEAD_REF: master'
GITHUB_BASE_REF: master
GITHUB_HEAD_REF: master
++ check_secrets
```

paulb777:
Fix #5523 

#no-changelog

### DIFF
--- a/.github/workflows/firestore.yml
+++ b/.github/workflows/firestore.yml
@@ -173,4 +173,5 @@ jobs:
       run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/FIREGSignInInfo.h.gpg \
           quickstart-ios/TestUtils/FIREGSignInInfo.h "$signin_secret"
     - name: Test quickstart
-      run: scripts/third_party/travis/retry.sh scripts/test_quickstart.sh Firestore
+      run: ([ -z $plist_secret ] || \
+            scripts/third_party/travis/retry.sh scripts/test_quickstart.sh Firestore)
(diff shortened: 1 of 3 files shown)


.github/workflows/firestore.yml
```diff
@@ -173,4 +173,5 @@ jobs:
       run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/FIREGSignInInfo.h.gpg \
           quickstart-ios/TestUtils/FIREGSignInInfo.h "$signin_secret"
     - name: Test quickstart
-      run: scripts/third_party/travis/retry.sh scripts/test_quickstart.sh Firestore
+      run: ([ -z $plist_secret ] || \
```
wilhuff, line 176: Doesn't this have to be in the `env` for the task? As it stands this looks like this will never run `test_quickstart.sh`.

paulb777: See https://github.com/firebase/firebase-ios-sdk/runs/641984399?check_suite_focus=true for a working example for Storage.

paulb777: I see. Storage sets up `env` for the job instead of the specific steps. That change needs to be made also. Making that change and verifying here and in #5525. Thanks!

paulb777:
Verified the disabled fork tests still run on the non-fork PR in #5525.

PR #5525, opened 2020-05-04
Test 5524 on unforked repo

paulb777:


paulb777:
Successful verification that the fork-disabled tests of #5524 still run on the non-forked repo.

PR #5537, opened 2020-05-06
Add a workflow to run remote config on GHA

granluo:
Add workflow to run remote config quickstart.


granluo:
The config loader does not work properly. Will test the config loader.

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,44 @@
...
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/qs-auth.plist.gpg \
+          quickstart-ios/config/GoogleService-Info.plist "$plist_secret"
+    - name: Install Secret FIREGSignInInfo.h
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/FIREGSignInInfo.h.gpg \
+          quickstart-ios/TestUtils/FIREGSignInInfo.h "$signin_secret"
+    - name: Test quickstart
+      run: |
+        cd ./quickstart-ios/config/
+        xcodebuild -workspace *.xcworkspace -list
+        xcodebuild test -scheme ConfigExample -workspace ConfigExample.xcworkspace -destination "platform=${{env.PLATFORM}},name=${{env.DEVICE_NAME}},OS=${{env.OS_VERSION}}"
```
ryanwilson, line 43: We have a script to do this for us already without the config:

https://github.com/firebase/firebase-ios-sdk/blob/24e23931b43c7791b456d92eede67ebd88ae525d/.github/workflows/storage.yml#L58-L59

(replacing `Storage` with `Config` or whatever the sample is named).

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,44 @@
...
+
+
+  quickstart:
+    env:
+      plist_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+      signin_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: technote-space/load-config-action@v1
```
ryanwilson, line 26: We should try to avoid depending on external actions from a security and reliability standpoint if possible - please remove this dependency.

.github/workflows/config.yml
```diff
@@ -0,0 +1,4 @@
+XCODE_VERSION: 11.3
+OS_VERSION: 13.3
+DEVICE_NAME: iPhone 11
+PLATFORM: iOS Simulator
```
ryanwilson, line 4: This isn't needed - we don't specify in the other quickstart tests either.

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: scripts/setup_quickstart.sh config
+    - name: Install Secret GoogleService-Info.plist
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/qs-auth.plist.gpg \
+          quickstart-ios/config/GoogleService-Info.plist "$plist_secret"
+    - name: Install Secret FIREGSignInInfo.h
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/FIREGSignInInfo.h.gpg \
+          quickstart-ios/TestUtils/FIREGSignInInfo.h "$signin_secret"
+    - name: Test quickstart
+      run: |
+        scripts/test_quickstart.sh Config
```
ryanwilson, line 38: Style nit: This can fit on one line so the `|` isn't necessary. 

Also can you prefix this with `([ -z $plist_secret ] || scripts/third_party/travis/retry.sh` similar to the other quickstarts? The first is a check to see if the secret is `null` in which case we'll fail early since the right info isn't available, and second will retry any failures in case it's a flake.

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,38 @@
...
+    paths:
+    - 'FirebaseRemoteConfig**'
+    - '.github/workflows/remoteconfig.yml'
+    - 'Gemfile'
+  schedule:
+    # Run every day at 11pm (PST) - cron uses UTC times
+    - cron:  '0 7 * * *'
+
+jobs:
+
```
paulb777, line 17: nit: extra blank line

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Setup quickstart
+      run: scripts/setup_quickstart.sh config
+    - name: Install Secret GoogleService-Info.plist
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/qs-auth.plist.gpg \
+          quickstart-ios/config/GoogleService-Info.plist "$plist_secret"
+    - name: Install Secret FIREGSignInInfo.h
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/FIREGSignInInfo.h.gpg \
+          quickstart-ios/TestUtils/FIREGSignInInfo.h "$signin_secret"
+    - name: Test quickstart
+      run: ([ -z $plist_secret ] || scripts/test_quickstart.sh Config)
```
paulb777, line 37: Still needs the retry for flakes.

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  quickstart:
+    env:
+      plist_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+      signin_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set xcode version
+      run: sudo xcode-select -switch /Applications/Xcode_11.3.app
```
paulb777, line 26: Is setting the Xcode version necessary?

Ideally I'd like to test on the GHA default - typically latest - so we catch issues early.

In the future, we could add more explicit Xcode version coverage.

granluo: will remove the setting if the GHA default device is sufficient.

.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,35 @@
...
+    env:
+      plist_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+      signin_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup quickstart
+      run: scripts/setup_quickstart.sh config
+    - name: Install Secret GoogleService-Info.plist
+      run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/qs-auth.plist.gpg \
```
ryanwilson, line 28: This needs to be `qs-config.plist.gpg` not `qs-auth`

granluo: yep, updated.

PR #5544, opened 2020-05-07
Add quickstart tests

granluo:
Add quickstart testings.


.github/workflows/remoteconfig.yml
```diff
@@ -0,0 +1,35 @@
+name: remoteconfig
```
paulb777, line 1: Looks like master wasn't updated before this branch was created.

One way to fix would be `git fetch` and `git rebase origin/master`

.github/workflows/dynamiclinks.yml
```diff
@@ -36,3 +36,22 @@ jobs:
       run: scripts/setup_bundler.sh
     - name: PodLibLint Storage Cron
       run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseDynamicLinks.podspec --platforms=ios ${{ matrix.flags }}
+
+  quickstart:
```
paulb777, line 40: The test is failing. There may be missing steps in the setup. It should be sorted out or removed for a subsequent PR.

.github/workflows/remoteconfig.yml
paulb777, line 1: It looks like master wasn't updated before this PR.

One way to fix would be to `git fetch` and `git rebase origin/master` on add_quickstart_tests branch.

.github/workflows/dynamiclinks.yml
granluo: remove it for now will add it back later.

.github/workflows/dynamiclinks.yml
```diff
@@ -3,8 +3,8 @@ name: dynamiclinks
 on:
   pull_request:
```
paulb777, line 4: Undo changes in this file - `git checkout master -- .github/workflows/dynamiclinks.yml`

PR #5547, opened 2020-05-08
Make Executors have deterministic shutdown

wilhuff:
In particular:

  * All tasks are guaranteed to either run to completion or not run at all;
  * Executors can now be destroyed from the threads they own (if applicable); and
  * Many common concepts between the Executor implementations have been harmonized and shared.

Note:

  * This change does not yet implement `Executor::Dispose` or clean up any of the higher level componentry to take advantage of these changes.
  * The executors implement shutdown by canceling any tasks that haven't started yet. This diverges from the plan (which called for a separate `UserCallbackExecutor` to handle this), but in the end it turns out to be simpler to just cancel everything and make cancel handle tasks-in-progress in a reasonable way. LMK if you have grave objections to this, in which case I can rework this.

I've tried to pull this apart into comprehensible commits, but the cutover from executor-specific tracking to `Task` in ceb3aa9f70d1b9934eeb27ae1a6c1c5b9955b5a1 wasn't something I could split without a lot of additional scaffolding. This commit remains a bear; sorry.

This is the foundation for addressing firebase/quickstart-unity#638.


Firestore/core/test/unit/util/task_test.cc
```diff
@@ -0,0 +1,241 @@
...
+
+    Await(task_can_complete);
+    steps += "3";
+  });
+
+  // Start the task on a separate thread; this will block until the Task
+  // completes.
+  Async([&] { task->Execute(); });
+
+  // Cancel on yet another thread because this also will block
```
var-const, line 165: Ultranit: missing full stop.

Firestore/core/test/unit/util/task_test.cc
```diff
@@ -0,0 +1,241 @@
...
+  task->Execute();
+  ASSERT_EQ(state.op_executed, 1);
+  ASSERT_EQ(state.op_destroyed, 1);
+  ASSERT_EQ(state.task_destroyed, 0);
+
+  task->Release();
+  ASSERT_EQ(state.task_destroyed, 1);
+}
+
+TEST_F(TaskTest, OwnedReleaseThenRelease) {
```
var-const, line 202: The actual sequence is `OwnedReleaseThenExecute`.

Firestore/core/src/util/executor_std.cc
```diff
@@ -38,10 +56,17 @@ std::string ThreadIdToString(const std::thread::id thread_id) {
 
 }  // namespace
 
+class ExecutorStd::SharedState {
+ public:
+  // Operations scheduled for immediate execution are also put on the schedule
+  // (with due time set to `Immediate`).
+  class Schedule schedule_;
```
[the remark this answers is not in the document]
wilhuff: `class` is necessary here because of the conflict between the `Executor::Schedule` method and the `Schedule` class, both of which are in scope here. Absent this, C++ prefers the method when resolving the unannotated name.

Firestore/core/src/util/executor_std.cc
```diff
@@ -38,10 +56,17 @@ std::string ThreadIdToString(const std::thread::id thread_id) {
 
 }  // namespace
 
+class ExecutorStd::SharedState {
```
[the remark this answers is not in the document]
wilhuff: I've gone back and forth about this about three times during the time it took me to get this working and then minimize it.

Initially there was a mutex in here and some additional things. While sketching what `Dispose` looked like I also tucked some state in here as well. Each time I've reversed `std::shared_ptr<Schedule>` and re-added the `SharedState`.

If it's all the same to you, I'd like to defer that cleanup until after I'm certain how dispose works.

Firestore/core/src/util/executor_std.cc
```diff
@@ -79,58 +109,58 @@ ExecutorStd::~ExecutorStd() {
...
-  const auto id =
-      PushOnSchedule(std::move(tagged.operation), now + delay, tagged.tag);
+  const auto target_time = MakeTargetTime(delay);
+  const auto id = PushOnSchedule(target_time, tag, std::move(operation));
+  return DelayedOperation(this, id);
+}
 
-  return DelayedOperation{[this, id] { TryCancel(id); }};
+void ExecutorStd::Complete(Task*) {
+  // No-op in this implementation
```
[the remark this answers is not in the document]
wilhuff: There are a few alternatives:

  * Make `Task` take a completion callback instead of an `executor`
  * Make `ExecutorLibdispatch` wrap the operation in another function-like thing (though I avoided this here because it's tricky to propagate cancelation without making operation some type other than `std::function`.

I think the second option isn't terribly viable but I can see a way to making the former work if you prefer it.

Firestore/core/src/util/executor_std.cc
```diff
@@ -79,58 +109,58 @@ ExecutorStd::~ExecutorStd() {
...
 }
 
-void ExecutorStd::TryCancel(const Id operation_id) {
-  schedule_.RemoveIf(
-      [operation_id](const Entry& e) { return e.id == operation_id; });
+void ExecutorStd::Cancel(const Id operation_id) {
+  auto removed = state_->schedule_.RemoveIf(
+      [operation_id](const Task& t) { return t.id() == operation_id; });
+  if (removed) {
+    removed->Release();
```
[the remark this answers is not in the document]
wilhuff: Good idea. Done.

Firestore/core/src/util/schedule.h
```diff
@@ -0,0 +1,130 @@
...
+    std::lock_guard<std::mutex> lock{mutex_};
+
+    for (auto iter = scheduled_.begin(), end = scheduled_.end(); iter != end;
+         ++iter) {
+      Task* task = *iter;
+      if (pred(*task)) {
+        return ExtractLocked(iter);
+      }
+    }
+    return {};
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/core/test/unit/util/task_test.cc
wilhuff: Done.

Firestore/core/test/unit/util/task_test.cc
wilhuff: Fixed. Also added `OwnedReleaseThenRelease`.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
+  {
+    TASK_TRACE("Executor::Complete %s task %s", this, task);
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    auto found = async_tasks_.find(task);
+    if (found != async_tasks_.end()) {
+      should_release = true;
+      async_tasks_.erase(found);
+
+      if (!task->is_immediate()) {
```
var-const, line 179: Nit: it's okay to try to `erase` a non-existent key, do you think having an explicit conditional is more readable?

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
+}
+
+void ExecutorLibdispatch::Complete(Task* task) {
+  bool should_release = false;
+  {
+    TASK_TRACE("Executor::Complete %s task %s", this, task);
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    auto found = async_tasks_.find(task);
+    if (found != async_tasks_.end()) {
```
var-const, line 175: Can you add a comment on why a `Task` might not be found?

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
+  }
+}
+
+void ExecutorLibdispatch::Cancel(Id operation_id) {
+  Task* found_task = nullptr;
+  {
+    TASK_TRACE("Executor::Cancel %s task %s", this, task);
+    std::lock_guard<std::mutex> lock(mutex_);
+
     // `time_slot` might have been destroyed by the time cancellation function
```
var-const, line 196: Nit: this comment references several no-longer-existent entities (e.g. `time_slot`).

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
-bool ExecutorLibdispatch::IsScheduled(const Tag tag) const {
-  bool result = false;
-  RunSynchronized(this, [this, tag, &result] {
-    result = std::any_of(schedule_.begin(), schedule_.end(),
-                         [&tag](const ScheduleEntry& operation) {
-                           return *operation.second == tag;
-                         });
-  });
-  return result;
+void ExecutorLibdispatch::InvokeSync(void* raw_task) {
```
var-const, line 222: These two functions have the same implementation. Do you think it's worthwhile keeping them separate (e.g. to be more future-proof)?

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
+
+  if (found_task) {
+    found_task->Cancel();
+    found_task->Release();
+  }
 }
 
-// Test-only methods
+void ExecutorLibdispatch::InvokeAsync(void* raw_task) {
+  auto task = static_cast<Task*>(raw_task);
```
var-const, line 218: Nit: consider qualifying `auto` with an asterisk to make it (even) clearer that it's a pointer.

Firestore/core/src/util/async_queue.cc
```diff
@@ -162,10 +163,13 @@ void AsyncQueue::RunScheduledOperationsUntil(const TimerId last_timer_id) {
         "Attempted to run scheduled operations until missing timer id: %s",
         last_timer_id);
 
-    for (auto next = executor_->PopFromSchedule(); next.has_value();
+    for (auto next = executor_->PopFromSchedule(); next != nullptr;
```
var-const, line 166: Yet another nit re. `auto*`.

Firestore/core/src/util/task.cc
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Returns the initial reference count for a Task based on whether or not the
+ * task shares ownership with the executor that created it.
+ *
+ * @param executor The executor that owns the Task, or `nullptr` if the Task
+ *     owns itself.
+ * @return The initial reference count value.
+ */
+int InitialRefCount(Executor* executor) {
+  return executor ? 2 : 1;
```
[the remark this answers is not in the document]
var-const: From the options you outlined, number 2 seems preferable (`ExecutorLibdispatch pays for it with an extra Retain`) -- the starting state of a `Task` is reasonable and the special case is confined to the class that needs it (`ExecutorLibdispatch`). So I'm mildly in favor of this option, but only assuming it's straightforward to implement/doesn't create any unexpected issues.

Firestore/core/src/util/executor_libdispatch.h
```diff
@@ -72,28 +57,36 @@ class ExecutorLibdispatch : public Executor {
...
+  static void InvokeAsync(void* raw_task);
+  static void InvokeSync(void* raw_task);
+
+  Id NextIdLocked();
+
+  mutable std::mutex mutex_;
 
   dispatch_queue_t dispatch_queue_;
-  // Stores non-owned pointers to `TimeSlot`s.
-  // Invariant: if a `TimeSlot` is in `schedule_`, it's a valid pointer.
```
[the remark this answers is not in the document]
wilhuff: I've reformulated this.

Firestore/core/src/util/executor_libdispatch.h
```diff
@@ -72,28 +57,36 @@ class ExecutorLibdispatch : public Executor {
...
+
+  std::unordered_set<Task*> async_tasks_;
```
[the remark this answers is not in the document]
wilhuff: `async_tasks_` (which I've renamed to `tasks_` because synchronous tasks are present too) tracks all tasks.

`schedule_` is an index of tasks by their Id, and only contains scheduled tasks. Cancellation of scheduled tasks is common (e.g. the idle timer task is canceled on every write), and forcing this to be a linear search of everything in the async queue is more expensive than I want it to be.

A single bimap would be sufficient for both usages, though it would require non-scheduled tasks to also have an Id.

I suspect we could get rid of the schedule altogether if `DelayedOperation` retained a pointer to the backing task, but trying to implement that is something I want to defer.

I've added comments describing the separation.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -45,51 +49,6 @@
...
-  } else {
-    DispatchSync(executor->dispatch_queue(), std::forward<Work>(work));
-  }
-}
-
-}  // namespace
-
 // MARK: - TimeSlot
 
 // Represents a "busy" time slot on the schedule.
```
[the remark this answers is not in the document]
wilhuff: I meant to attach this to the header innards but forgot to do so. This has been reformulated and is now attached to `tasks_`.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
 }
 
 void ExecutorLibdispatch::Execute(Operation&& operation) {
-  DispatchAsync(dispatch_queue(), std::move(operation));
+  auto task = Task::Create(this, std::move(operation));
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +113,184 @@ void MarkDone() {
...
+  }
+
+  dispatch_async_f(dispatch_queue_, task, InvokeAsync);
 }
+
 void ExecutorLibdispatch::ExecuteBlocking(Operation&& operation) {
-  DispatchSync(dispatch_queue(), std::move(operation));
+  HARD_ASSERT(
+      GetCurrentQueueLabel() != GetQueueLabel(dispatch_queue_),
+      "Calling DispatchSync on the current queue will lead to a deadlock.");
```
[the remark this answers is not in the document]
wilhuff: Fixed.

Firestore/core/src/util/executor_libdispatch.mm
wilhuff: There's an overlap between the first scheduled task (with id zero) and all non-scheduled tasks, which have no id and default to zero. Removing this check makes it such that the completion of the first non-scheduled task can mark the first scheduled task complete. I've added a comment.

Firestore/core/src/util/executor_libdispatch.mm
wilhuff: I can't think of a good reason why it might not be found--concurrent access to the executor is forbidden and tasks should only complete themselves once.

Given that everything else in this vicinity handles duplicate invocation it seemed sensible to do so here as well. If you prefer, I can make the assertion that the task is found. I'm slightly wary of this though, since the consequences are dire and the behavior on not found seems reasonable.

Firestore/core/src/util/executor_libdispatch.mm
wilhuff: Fixed. Sorry.

Firestore/core/src/util/executor_libdispatch.mm
wilhuff: I kept them separate because stack traces look better--it's purely cosmetic. I've added a comment so that a future do-gooder doesn't undo this.

Firestore/core/src/util/async_queue.cc
wilhuff: Done.

Firestore/core/src/util/task.h
```diff
@@ -0,0 +1,227 @@
...
+#include <thread>  // NOLINT(build/c++11)
+
+#include "Firestore/core/src/util/executor.h"
+
+namespace firebase {
+namespace firestore {
+namespace util {
+
+// For testing
+class TrackingTask;
```
var-const, line 33: IIRC, a friend declaration serves as a forward declaration as well, so this can be omitted.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +94,231 @@ void MarkDone() {
...
+  }
+
+  return tag_scheduled;
 }
 
-absl::optional<Executor::TaggedOperation>
-ExecutorLibdispatch::PopFromSchedule() {
-  absl::optional<Executor::TaggedOperation> result;
+bool ExecutorLibdispatch::IsIdScheduled(Id id) const {
+  Task* found_task = nullptr;
```
var-const, line 272: Looking at these functions, it seems like `Schedule` is a clearly "separatable" abstraction (similar to `ExecutorStd`) that could wrap the mutex and the two containers and potentially return `TaskRef`s. It might be overkill, though.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -243,91 +94,231 @@ void MarkDone() {
...
+      // Only try to remove scheduled tasks here because non-scheduled tasks
+      // all have id 0, which overlaps with the first scheduled task.
+      if (!task->is_immediate()) {
+        schedule_.erase(task->id());
+      }
+    }
+  }
+
+  if (should_release) {
+    task->Release();
```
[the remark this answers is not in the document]
wilhuff: In the presence of calls between objects that each have their own locks, a general deadlock avoidance protocol is to avoid making calls to others while holding your own lock.

`Release` does not acquire the `Task`'s lock and does not call back to the executor so I guess it's safe to call while holding the executor's lock, but I'm wary of abandoning the general rule on the basis of this knowledge. Task could change in such a way that it called back into executor, or even worse could change such that it blocked on some other thread's callback into the executor and then it would introduce the possibility of deadlock (the former case can be solved by using recursive locks, the latter cannot).

I've added a comment here stating:

> Avoid calling potentially locking methods on the task while holding the executor's lock.

Firestore/core/src/util/executor_libdispatch.mm
wilhuff: I thought about this too. For now I've left it as is just to avoid further iterating on this PR.

Firestore/core/src/util/task.cc
```diff
@@ -0,0 +1,210 @@
...
+    if (this_thread != executing_thread_) {
+      AwaitLocked(lock);
+    }
+  }
+  return true;
+}
+
+void Task::AwaitLocked(std::unique_lock<std::mutex>& lock) {
+  TASK_TRACE("Task::Await %s", this);
+  is_complete_.wait(lock, [this] {
```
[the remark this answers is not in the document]
var-const: Ah, you're right, sorry about the false alarm.

Firestore/core/src/util/executor_libdispatch.h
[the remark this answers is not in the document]
var-const: That's a good point. Do you think it's worth appending this to the existing comment? Rough sketch: `(while under normal operation the number of tasks on the schedule is small and doesn't grow much, unusual usage of the SDK could result in thousands of tasks being added)`.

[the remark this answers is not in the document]
wilhuff: I think it's fairly reasonable (and our usual practice) to just pick the structures that give the right asymptotic complexity. It's only when intentionally choosing a structure that isn't intuitively right that it's worth commenting on why that's the right choice. In this case, if we only had the `tasks_` set I'd expect a comment on why the linear search is reasonable.

PR #5551, opened 2020-05-08
Add some missing steps to AddNewPod instructions

paulb777:
#no-changelog

### DIFF
--- a/AddNewPod.md
+++ b/AddNewPod.md
@@ -65,12 +65,5 @@ For GitHub tag management and public header change detection, add a GitHub api t
 * Update [Firebase.podspec](Firebase.podspec).
 * Register library via registerInternalLibrary API like this
   [Storage example](FirebaseStorage/Sources/FIRStorageComponent.m).
-* Register service in
-  [firebasecore.proto](Firebase/CoreDiagnostics/ProtoSupport/Protos/firebasecore.proto) along with
-  corresponding internal icore.proto.
-* Regenerate the iCore nanopb protos by running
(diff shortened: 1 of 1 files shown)


AddNewPod.md
```diff
@@ -58,3 +58,19 @@ the GitHub Action check phase. Run [scripts/style.sh](scripts/style.sh).
...
+* Make sure the public headers are imported via [Firebase.h](CoreOnly/Sources/Firebase.h).
+* Update [Firebase.podspec](Firebase.podspec).
+* Register library via registerInternalLibrary API like this
+  [Storage example](FirebaseStorage/Sources/FIRStorageComponent.m).
+* Register service in
+  [firebasecore.proto](Firebase/CoreDiagnostics/ProtoSupport/Protos/firebasecore.proto) along with
+  corresponding internal icore.proto.
+* Regenerate the iCore nanopb protos by running
+  [generate_project.sh](Firebase/CoreDiagnostics/generate_project.sh).
+* Add logging code to
```
ryanwilson, line 73: Is this still required with the user agent?

paulb777: + @maksymmalyhin

maksymmalyhin: Good question. This is used to identify a sender of an event and is not related directly to the user agent. As far as I'm aware currently the `ICORE` value is the only important as an indicator of a daily heartbeat. Also, no other SDKs  (except core) use CoreDiagnostics. So, I think, this step is safe to be deleted from the instruction.

PR #5554, opened 2020-05-08
Add defer, a general facility for ad-hoc RAII.

wilhuff:
Also, use it in a few places where it helps.

### DIFF
--- a/Firestore/core/src/util/defer.h
+++ b/Firestore/core/src/util/defer.h
@@ -27,13 +27,18 @@ template <typename Action>
 class Deferred;
 
 /**
- * Creates a deferred action that will be destroyed at the close of the lexical
- * scope containing the result of this function call. This is useful for
- * performing ad-hoc RAII-style actions, without having to create the wrapper
- * class. For example:
+ * Creates a `Deferred<Action>` that will execute the given `action` when the
+ * returned object is destroyed at the close of the lexical scope containing it.
+ * The result of the call to `Defer` must be assigned to a variable for the
+ * `Deferred` object to have any useful effect; otherwise the `Deferred` is
+ * destroyed at the end of the statement, which is equivalent to just directly
+ * running the `action`.
+ *
+ * `Defer` is useful for performing ad-hoc RAII-style actions, without having to
+ * create the wrapper class. For example:
  *
  *     FILE* file = fopen(filename, "rb");
- *     auto cleanup = defer([&] {
+ *     auto cleanup = Defer([&] {
  *       if (file) {
  *         fclose(file);
  *       }
@@ -44,7 +49,7 @@ class Deferred;
  *     the restriction that exists on destructors generally.
  */
 template <typename Action>
-Deferred<Action> defer(Action&& action) {
+Deferred<Action> Defer(Action&& action) {
   return Deferred<Action>(std::forward<Action>(action));
 }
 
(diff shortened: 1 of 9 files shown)


Firestore/core/src/util/defer.h
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Storage for a deferred action. The `action` is invoked during the destructor
+ * of the `Deferred`.
+ */
+template <typename Action>
+class Deferred {
+ public:
+  explicit Deferred(Action&& action) : action_(std::move(action)) {
+  }
+
```
var-const, line 60: Should this class forbid copying (and probably moving as well)?

Firestore/core/src/util/defer.h
```diff
@@ -0,0 +1,73 @@
...
+namespace firebase {
+namespace firestore {
+namespace util {
+
+template <typename Action>
+class Deferred;
+
+/**
+ * Creates a deferred action that will be destroyed at the close of the lexical
```
var-const, line 30: Looks like `destroyed` should be `executed`/`run`/`invoked`/etc.

Firestore/core/src/util/defer.h
```diff
@@ -0,0 +1,73 @@
...
+ * scope containing the result of this function call. This is useful for
```
var-const, line 31: Optional: perhaps highlight that any meaningful use of this function requires assigning its return value to a variable?

Firestore/core/src/util/defer.h
```diff
@@ -0,0 +1,73 @@
...
+ *         fclose(file);
+ *       }
+ *     });
+ *
+ * @param action a callable object; usually a lambda. Even if exceptions are
+ *     enabled, when `action` is invoked it must not throw. This is similar to
+ *     the restriction that exists on destructors generally.
+ */
+template <typename Action>
+Deferred<Action> defer(Action&& action) {
```
var-const, line 47: Question: what's the rationale to use a lowercase name for this function? (generally, the style guide requires functions to be `CamelCase`)

Firestore/core/src/util/defer.h
wilhuff: Fixed.

Firestore/core/src/util/defer.h
wilhuff: Done.

Firestore/core/src/util/defer.h
wilhuff: It's an oversight. Fixed.

Firestore/core/src/util/defer.h
wilhuff: In order to return this by value the type must at least be moveable. Only once we get to C++17 with guaranteed elision can we return a non-copyable, non-moveable object by value. I've made it move-only.

The other alternative is to make this class concrete, taking a `std::function<void()>` as an argument. Then the constructor can be invoked directly and the class could be made non-copyable and non-moveable (like `std::lock_guard`). Usage would look like this:

```c++
FILE* file = fopen(filename, "rb");
Defer cleanup([&] {
  if (file) {
    fclose(file);
  }
}
```

Unfortunately, the major downside of that approach is that `std::function` must perform a heap allocation whereas the inferred implementation defined type of `Action` here likely does not.

PR #5564, opened 2020-05-11
Only log the URL follow event in messageClicked:withAction: if there's a URL to follow

christibbs:
Fix for internal bug #155959022 to not count clicks if an in-app messaging campaign has no URL to follow.


FirebaseInAppMessaging/Tests/Unit/FIRIAMDisplayExecutorTests.m
```diff
@@ -599,8 +599,12 @@ - (void)testAnalyticsTrackingOnMessageClickCase {
 
   [self.clientMessageCache setMessageData:@[ self.m2 ]];
```
JasonAHeron, line 601: this seems like a test for the positive case where a buttion with an action *does* log. Do we also have a test for a button without an action *not* logging?

christibbs: Added this in 8d26898.

JasonAHeron:
LGTM

PR #5581, opened 2020-05-12
update changelog for messaging

chliangGoogle:


FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# unreleased
+# 2020-05 -- v4.4.1
 - [changed] Updated NSError with a failure reason to give more details on the error.
```
morganchen12, line 2: nit: Can you add a bug or PR number to this?

chliangGoogle: Good catch!

PR #5586, opened 2020-05-12
Make sure `timestamp.h` compiles even in presence of `min` and `max` macros

var-const:
It is a well-known issue that `<windows.h>` Windows header defines macros `min` and `max`, causing compilation errors for code that uses `std::min/max`, `std::numeric_limits::min/max`, etc While users may work around the issue by defining the `NOMINMAX` macro before including Windows headers, this PR adds some preprocessor directives to temporarily undefine these macros then restore their previous values. While the relevant pragmas are non-standard, they are implemented in all Firestore-supported compilers (Visual Studio, Clang, GCC).

### DIFF
--- a/Firestore/core/include/firebase/firestore/timestamp.h
+++ b/Firestore/core/include/firebase/firestore/timestamp.h
@@ -212,6 +212,15 @@ inline bool operator==(const Timestamp& lhs, const Timestamp& rhs) {
 }
 
 #if !defined(_STLPORT_VERSION)
+
+// Make sure the header compiles even when included after `<windows.h>` without
+// `NOMINMAX` defined. `push/pop_macro` pragmas are supported by Visual Studio
+// as well as Clang and GCC.
+#pragma push_macro("min")
+#pragma push_macro("max")
+#undef min
+#undef max
+
 template <typename Clock, typename Duration>
 std::chrono::time_point<Clock, Duration> Timestamp::ToTimePoint() const {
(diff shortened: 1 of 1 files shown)


Firestore/core/include/firebase/firestore/timestamp.h
```diff
@@ -212,6 +212,15 @@ inline bool operator==(const Timestamp& lhs, const Timestamp& rhs) {
 }
 
 #if !defined(_STLPORT_VERSION)
+
+// Make sure the header compiles even when included after `<windows.h>` without
+// `NOMINMAX` defined. `push/pop_macro` pragmas are supported by Visual Studio
+// as well as Clang and GCC.
+#pragma push_macro("min")
```
wilhuff, line 219: Just to confirm, this works correctly even if there's no current definition for the macros (i.e. that undefined gets pushed on the stack and then undefined is popped back out, below)?

var-const: Unfortunately, these are very lightly documented in both Visual Studio and GCC (and I couldn't find Clang docs at all). Experimentally, it seems to be the case: [link](https://wandbox.org/permlink/MyeekNZMaEuD4f5e).

wilhuff: I played with your example and verified it works for function-like macros too: [link](https://wandbox.org/permlink/tkWNatCa8pTu6E0S).

PR #5600, opened 2020-05-13
Fix race condition between setting developer mode and config fetch

### LINKED ISSUE #4740: Remote Config fetchAndActivate does not honour minimumFetchInterval in RemoteConfigSettings

  * Xcode version: 11.3
  * Firebase SDK version: 6.14.0
  * Component version: 6.14.0

#### Steps to reproduce:
On app launch, set `minimumFetchInterval` to 10 sec,
Call `fetchAndActivate` with a completion handler.
The first time it will always fetch from the server.
Then wait for a couple of min and re launch the app
You can see from the log
`[Firebase/RemoteConfig][I-RCN000067] Successfully set configSettings. Developer Mode: false, Minimum Fetch Interval:10.000000, Fetch timeout:60.000000`
`[Firebase/RemoteConfig][I-RCN000051] Returning cached data.`
`[Firebase/RemoteConfig][I-RCN000069] Most recently fetched config is already activated.`
It does not make a call to the backend to retrieve the latest Config.

The `minimumFetchInterval` is not taking effect.
#### Relevant Code:

```
let remoteConfig = RemoteConfig.remoteConfig()
settings.minimumFetchInterval = 10 
remoteConfig.setDefaults(fromPlist: "RemoteConfigDefault")
remoteConfig.configSettings = settings
remoteConfig.fetchAndActivate(completionHandler: { (status, error) in
      ....
})
```

paulb777:
Fix #4740 

Insure that minimumFetchInterval setting is applied before fetching.  Otherwise there will be indeterministic behavior when trying to verify console changes in an app.

I've verified in a sample and will work on adding a unit test.

I'd like to get the review going since I'd like to get this fix in M71.

### DIFF
--- a/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
+++ b/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
@@ -214,8 +214,11 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
 #pragma mark - fetch
 
 - (void)fetchWithCompletionHandler:(FIRRemoteConfigFetchCompletion)completionHandler {
-  [self fetchWithExpirationDuration:_settings.minimumFetchInterval
-                  completionHandler:completionHandler];
+  __block NSTimeInterval minimumFetchInterval;
+  dispatch_sync(_queue, ^{
+    minimumFetchInterval = self->_settings.minimumFetchInterval;
+  });
+  [self fetchWithExpirationDuration:minimumFetchInterval completionHandler:completionHandler];
 }
 
 - (void)fetchWithExpirationDuration:(NSTimeInterval)expirationDuration
--- a/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
+++ b/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
@@ -214,11 +214,10 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
 #pragma mark - fetch
 
 - (void)fetchWithCompletionHandler:(FIRRemoteConfigFetchCompletion)completionHandler {
-  __block NSTimeInterval minimumFetchInterval;
-  dispatch_sync(_queue, ^{
-    minimumFetchInterval = self->_settings.minimumFetchInterval;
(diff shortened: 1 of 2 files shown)


paulb777:
`- (void)fetchWithCompletionHandler:(FIRRemoteConfigFetchCompletion)completionHandler {` should be fixed.

More investigation still needed for 

```
- (void)fetchWithExpirationDuration:(NSTimeInterval)expirationDuration
                  completionHandler:(FIRRemoteConfigFetchCompletion)completionHandler {
```

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -214,8 +214,11 @@ - (void)ensureInitializedWithCompletionHandler:
 #pragma mark - fetch
 
 - (void)fetchWithCompletionHandler:(FIRRemoteConfigFetchCompletion)completionHandler {
-  [self fetchWithExpirationDuration:_settings.minimumFetchInterval
-                  completionHandler:completionHandler];
+  __block NSTimeInterval minimumFetchInterval;
+  dispatch_sync(_queue, ^{
```
maksymmalyhin, line 218: In general `dispatch_sync` should be avoided unless required because the more `dispatch_sync` is used the easier it is to introduce a deadlock even when modifying code not directly related. I wonder if it can be updated to either:
1. Use `atomic` properties or `_Atomic` vars for variables involved in the race condition
2. Use `dispatch_async` e.g. : 
```
dispatch_async(_queue, ^{
  [self fetchWithExpirationDuration:self->_settings.minimumFetchInterval completionHandler:completionHandler];
}
```

chliangGoogle: #2 is probably better as remote config's mutex lock is all guarded by this queue. The fetch is assigned to async again inside https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/RCNFetch.m#L139
So using another mechanism might be more expensive since we already have a queue just for that.

paulb777:
Testing was done manually with the new API tests in #5602. They'll be merged once automation is added for them.

All other automated tests passed on commit before changelog update.

PR #5604, opened 2020-05-13
Bump stream closing log to WARN.

wu-hui:
b/155897356

### DIFF
--- a/Firestore/core/src/remote/stream.cc
+++ b/Firestore/core/src/remote/stream.cc
@@ -295,7 +295,11 @@ void Stream::HandleErrorStatus(const Status& status) {
 void Stream::OnStreamFinish(const Status& status) {
   EnsureOnQueue();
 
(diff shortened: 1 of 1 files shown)


Firestore/core/src/remote/stream.cc
```diff
@@ -295,7 +295,11 @@ void Stream::HandleErrorStatus(const Status& status) {
 void Stream::OnStreamFinish(const Status& status) {
   EnsureOnQueue();
 
-  LOG_DEBUG("%s Stream error: '%s'", GetDebugDescription(), status.ToString());
+  if(!status.ok()) {
```
var-const, line 298: Nit: there should be a space between `if` and the opening brace, you probably need to run `style.sh`.

wu-hui: Done.

PR #5605, opened 2020-05-13
Generate the OSS manifest from the internal textprotos.

ryanwilson:
This replaces the internal Python tool that offers the same functionality.

Can be run by calling:

```
swift run oss-manifest-generator  --git-root $(pwd)/.. \
  --releasing-pods <PATH_TO_CURRENT_RELEASE_PROTO> \
  --existing-versions <PATH_TO_ALL_SDKS_PROTO>
```


ZipBuilder/Sources/ManifestReader/FirebaseSDKs.pb.swift
```diff
@@ -59,7 +59,7 @@ public struct ZipBuilder_SDK {
   }
 
   /// MPM name for the blueprint.
-  var mpmName: String {
+  public var mpmName: String {
```
paulb777, line 62: If I remember right, I got warnings for these `public`s. Is there a reason they're added back?

ZipBuilder/Sources/oss-manifest-generator/main.swift
```diff
@@ -0,0 +1,121 @@
...
+  @Option(help: "The file path to a textproto file containing all existing Pods, of type `ZipBuilder_FirebaseSDKs.",
+          transform: URL.init(fileURLWithPath:))
+  var existingVersions: URL
+
+  mutating func validate() throws {
+    guard FileManager.default.fileExists(atPath: gitRoot.path) else {
+      throw ValidationError("git-root does not exist: \(gitRoot.path)")
+    }
+
+    guard FileManager.default.fileExists(atPath: currentRelease.path) else {
```
paulb777, line 44: Optional idea - replace both of these with a g3 repo root - with default being head

ZipBuilder/Sources/ManifestReader/FirebaseSDKs.pb.swift
ryanwilson: I needed access to the `openSource` variable but for whatever previous reason it was private. Nothing was specified in the `.proto` itself. I didn't see any warnings, so I'm not sure what it could have been previously. Maybe this was before it was split up into a separate module?

I can manually remove the `public` designations for the variables we're not using, but I haven't been able to see any warnings yet in any of the targets.

ZipBuilder/Sources/oss-manifest-generator/main.swift
ryanwilson: I do like that idea although then we'll have to use the google3 relative path in the code and I think that's not preferred.

ZipBuilder/Sources/ManifestReader/FirebaseSDKs.pb.swift
paulb777: No need. I may be misremembering what it was ....

ryanwilson: I'll fix it if anything comes up!

PR #5615, opened 2020-05-15
Danger - fix missing change log warning

maksymmalyhin:


maksymmalyhin:
I debugged and tested different combinations of changes. With the current Dangerfile I got the warning when had a change in a public header and didn't get it when the change was reverted.

Issue #3586, opened 2019-08-14
"Most recently fetched config is already activated" after remote config activate() call


b-onc:
## [REQUIRED] Step 2: Describe your environment

  * Xcode version: 10.3
  * Firebase SDK version: 6.6.0
  * Firebase Component: Remote Config
  * Component version: 4.2.1

### [REQUIRED] Step 3: Describe the problem
We fetch and activate remote config with the following code:
```
let expirationDuration: TimeInterval
#if DEBUG
expirationDuration = 0
#else
expirationDuration = 3600
#endif
firebaseRemoteConfig.fetch(withExpirationDuration: expirationDuration) { [unowned self] (status, error) in
    if status == .success {
        log.info("Remote Config fetched")
        self.firebaseRemoteConfig.activate(completionHandler: { (error) in
            if let error = error {
                log.error(error)
            }
        })
    } else {
        log.error("Remote Config could not be fetched. \(String(describing: error))")
    }
}
```

We call this part of the code only one, in `AppDelegate` during `didFinishLaunchingWithOptions` 

In the wild, we see lots of Remote config errors reported, they're pretty straightforward:

> ActivationFailureReason: Most recently fetched config is already activated
> Code: 8003

Is it an error to try to activate fetched config when it's the same as active one? If yes, which approach should we take?
We can't use the `fetchAndActivate` since it doesn't have the `expirationDuration` parameter.

morganchen12:
If your goal is to make sure the values in RemoteConfig are up-to-date, this code will work and you can ignore the error.

@dmandar is this intentional? If we have an error that we recommend users ignore, can we downgrade the error to an info/debug log instead?

dmandar:
The error object always contains a value when activation cannot succeed. In this case, the already fetched config (present in the cache) is also already activated and no new activation can take place.  If you prefer, you can use fetchAndActivate by pre-configuring the fetchInterval in the configSettings at initialization time which returns more granular status regarding whether both remote fetch and activate succeeded.

dmandar:
Thanks for the feedback. We are working on improving the return status of the existing API. In the meantime,  if appropriate, please use the new fetchAndActivate API which should have more granular status.

paulb777:
There is already a warning logged at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L298.

[...]

PR #5617, opened 2020-05-15
Migrate FetchAndActivate from deprecated implementation

paulb777:
Migrate `fetchAndActivateWithCompletionHandler:` implementation from using the deprecated `activateFetched` API to the async `activateWithCompletionHandler:`

Working on integration tests, I was getting a deadlock in `activateFetched` so thought it would be good to move this forward.

### DIFF
--- a/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
+++ b/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
@@ -244,18 +244,19 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
         // Fetch completed. We are being called on the main queue.
         // If fetch is successful, try to activate the fetched config
         if (fetchStatus == FIRRemoteConfigFetchStatusSuccess && !fetchError) {
-          [strongSelf activateWithCompletionHandler:^(NSError * _Nullable activateError) {
+          [strongSelf activateWithCompletionHandler:^(NSError *_Nullable activateError) {
             if (completionHandler) {
-              FIRRemoteConfigFetchAndActivateStatus status = activateError ?
-                FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData :
-                FIRRemoteConfigFetchAndActivateStatusSuccessFetchedFromRemote;
+              FIRRemoteConfigFetchAndActivateStatus status =
+                  activateError ? FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData
+                                : FIRRemoteConfigFetchAndActivateStatusSuccessFetchedFromRemote;
               completionHandler(status, fetchError);
             }
           }];
         } else if (completionHandler) {
-          FIRRemoteConfigFetchAndActivateStatus status = fetchStatus ?
-            FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData :
-            FIRRemoteConfigFetchAndActivateStatusError;
+          FIRRemoteConfigFetchAndActivateStatus status =
+              fetchStatus == FIRRemoteConfigFetchStatusSuccess
+                  ? FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData
+                  : FIRRemoteConfigFetchAndActivateStatusError;
           completionHandler(status, fetchError);
         }
       };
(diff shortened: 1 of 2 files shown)


FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -236,27 +236,28 @@ - (void)fetchAndActivateWithCompletionHandler:
...
-          if (fetchStatus == FIRRemoteConfigFetchStatusSuccess) {
-            status = didActivate ? FIRRemoteConfigFetchAndActivateStatusSuccessFetchedFromRemote
-                                 : FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData;
-          } else {
-            status = FIRRemoteConfigFetchAndActivateStatusError;
-          }
-          // Pass along the fetch error e.g. throttled.
-          completionHandler(status, error);
+        if (fetchStatus == FIRRemoteConfigFetchStatusSuccess && !fetchError) {
+          [strongSelf activateWithCompletionHandler:^(NSError *_Nullable activateError) {
```
maksymmalyhin, line 247: 👍

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -236,27 +236,28 @@ - (void)fetchAndActivateWithCompletionHandler:
...
+            if (completionHandler) {
+              FIRRemoteConfigFetchAndActivateStatus status =
+                  activateError ? FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData
+                                : FIRRemoteConfigFetchAndActivateStatusSuccessFetchedFromRemote;
+              completionHandler(status, fetchError);
```
maksymmalyhin, line 252: Should we use `activateError` instead of `fetchError` here?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -236,27 +236,28 @@ - (void)fetchAndActivateWithCompletionHandler:
     (FIRRemoteConfigFetchAndActivateCompletion)completionHandler {
   __weak FIRRemoteConfig *weakSelf = self;
   FIRRemoteConfigFetchCompletion fetchCompletion =
-      ^(FIRRemoteConfigFetchStatus fetchStatus, NSError *error) {
+      ^(FIRRemoteConfigFetchStatus fetchStatus, NSError *fetchError) {
```
maksymmalyhin, line 239: This method is a great candidate to be refactored with Promises. We can consider it for future PRs.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Maybe, but `fetchError` matches the previous behavior and changing it would be a breaking change.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Agreed! This and several other RC methods.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
maksymmalyhin: If I read the code correctly, `fetchError` is `nil` here, so can we just pass `nil`?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -236,27 +236,28 @@ - (void)fetchAndActivateWithCompletionHandler:
...
+            if (completionHandler) {
+              FIRRemoteConfigFetchAndActivateStatus status =
+                  activateError ? FIRRemoteConfigFetchAndActivateStatusSuccessUsingPreFetchedData
+                                : FIRRemoteConfigFetchAndActivateStatusSuccessFetchedFromRemote;
+              completionHandler(status, fetchError);
+            }
+          }];
+        } else if (completionHandler) {
+          FIRRemoteConfigFetchAndActivateStatus status =
+              fetchStatus == FIRRemoteConfigFetchStatusSuccess
```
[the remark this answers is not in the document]
paulb777: Yes - if the config didn't change. See https://github.com/firebase/firebase-ios-sdk/issues/3586.  I'm planning to deprecate that API and add a new API that returns a bool like Android and JS instead of an error - after I get some integration test infra set up.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Yep - good catch. Thanks!

PR #5620, opened 2020-05-15
Implement safe, synchronous destruction in Firestore

wilhuff:
Now that the Executor makes strong guarantees that tasks submitted to it will either be completely run or not start at all once destruction starts (#5547), build a Dispose protocol on top of this to allow Firestore to shut itself down leaf first.

Now that Firestore calls the Dispose chain down to the Executor, it's no longer possible for tasks to run while the instance is partially destructed nor can user callbacks execute after the instance has been destroyed.

This is the other half of the iOS changes required to address firebase/quickstart-unity#638.

There are a few follow-on tasks that this PR does not address:

  * Ownership of FirestoreClient and AsyncQueue are still shared though they no longer need to be.
  * The tests are still overly defensive about keeping different test cases isolated from each other. Much of this can be removed now that we're guaranteed to release resources by the time the test case shuts down (at least in GoogleTest; XCTest remains to be seen).


Firestore/core/src/util/executor_std.h
```diff
@@ -62,24 +64,29 @@ class ExecutorStd : public Executor {
...
+
+  Id PushOnScheduleLocked(TimePoint when, Tag tag, Operation&& operation);
 
   void OnCompletion(Task* task) override;
   void Cancel(Id operation_id) override;
 
-  void PollingThread();
-  Id NextId();
+  static void PollingThread(std::shared_ptr<SharedState> state);
+  Id NextIdLocked();
```
var-const, line 75: I liked the pattern used in `Task` where `*Locked` functions would take a `unique_lock` as the first argument. Can the same be done here?

Firestore/core/src/util/executor_std.cc
```diff
@@ -145,19 +164,19 @@ void ExecutorStd::Cancel(const Id operation_id) {
...
+  const auto id = NextIdLocked();
   state_->schedule_.Push(
       Task::Create(nullptr, when, tag, id, std::move(operation)));
   return id;
 }
 
-void ExecutorStd::PollingThread() {
-  std::shared_ptr<SharedState> local_state = state_;
+void ExecutorStd::PollingThread(std::shared_ptr<SharedState> state) {
+  std::shared_ptr<SharedState> local_state = state;
```
var-const, line 179: How does the other copy of the shared pointer help? It seems they have exactly the same lifetime.

Firestore/core/src/util/executor_std.cc
```diff
@@ -131,8 +143,15 @@ void ExecutorStd::OnCompletion(Task*) {
 }
 
 void ExecutorStd::Cancel(const Id operation_id) {
-  auto removed = state_->schedule_.RemoveIf(
-      [operation_id](const Task& t) { return t.id() == operation_id; });
+  Task* removed = nullptr;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
```
var-const, line 148: `schedule_` already has an internal mutex. Is this additional mutex to protect `state_` pointer itself?

Firestore/core/src/util/executor_std.h
```diff
@@ -62,24 +64,29 @@ class ExecutorStd : public Executor {
...
 
   // State shared with workers. Note that if the Executor's destructor is called
   // from a worker thread, this state will outlive the nominally owning
-  // Executor.
+  // Executor. `mutex_` does not protect this state.
   std::shared_ptr<SharedState> state_;
-
-  std::vector<std::thread> worker_thread_pool_;
-
-  std::atomic<Id> current_id_{0};
```
var-const, line 82: What's the advantage of not using an atomic?

Firestore/core/test/unit/util/executor_test.cc
```diff
@@ -379,6 +379,55 @@ TEST_P(ExecutorTest, DestructorAvoidsDeadlockWhenDeletingSelf) {
   ASSERT_EQ(result, "12");
 }
 
+TEST_P(ExecutorTest, DipsoseBlocksTaskSubmission) {
+  // Verify there's no crash for an idempotent invocation.
```
var-const, line 383: Nit: move this comment one line down?

Firestore/core/test/unit/util/executor_test.cc
```diff
@@ -379,6 +379,55 @@ TEST_P(ExecutorTest, DestructorAvoidsDeadlockWhenDeletingSelf) {
   ASSERT_EQ(result, "12");
 }
 
+TEST_P(ExecutorTest, DipsoseBlocksTaskSubmission) {
```
var-const, line 382: Nit: `s/Dipsose/Dispose/`.

Firestore/core/test/unit/util/executor_test.cc
```diff
@@ -379,6 +379,55 @@ TEST_P(ExecutorTest, DestructorAvoidsDeadlockWhenDeletingSelf) {
...
+  executor->Dispose();
+
+  Expectation ran;
+  executor->Execute(ran.AsCallback());
+
+  auto status = ran.get_future().wait_for(Executor::Milliseconds(50));
+  ASSERT_EQ(status, std::future_status::timeout);
+}
+
+TEST_P(ExecutorTest, DipsoseBlocksConcurrentTaskSubmission) {
```
var-const, line 394: Same nit.

Firestore/core/test/unit/util/executor_test.cc
```diff
@@ -379,6 +379,55 @@ TEST_P(ExecutorTest, DestructorAvoidsDeadlockWhenDeletingSelf) {
...
+    executor->Dispose();
+    dispose_complete.Fulfill();
+  });
+
+  Await(dispose_running);
+  Expectation execute_running;
+  Expectation ran;
+  Async([&] {
+    execute_running.Fulfill();
+    executor->Execute(ran.AsCallback());
```
var-const, line 420: Consider adding a brief comment here, similar to the ones above, to explain the intent of this `Execute`.

Firestore/core/src/util/executor_libdispatch.mm
```diff
@@ -157,6 +174,7 @@
   {
     TASK_TRACE("Executor::OnCompletion %s task %s", this, task);
     std::lock_guard<std::mutex> lock(mutex_);
+    // No need to check `disposed_`: `tasks_` was cleared.
```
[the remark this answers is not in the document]
wilhuff: I'm starting to think this comment is altogether unnecessary. Only methods that add tasks need to check `disposed_`. What do you think of removing it?

In the meantime I've taken your suggestion.

Firestore/core/src/util/executor_std.cc
wilhuff: Correct.

We could probably remove the internal mutex on schedule (though it makes `PopBlocking` something that will require a `unique_lock` as an argument.

Firestore/core/src/util/executor_std.cc
wilhuff: It doesn't. I should have cleaned this up when changing this to pass the state rather than rely on `this`.

Firestore/core/src/util/executor_std.h
wilhuff: I only did that for the one method in `Task::AwaitLocked` and did so specifically because that method needed to wait on a condition variable. The `Locked` suffix alone is enough of an indication that the lock needs to be held so I'm wary of also requiring that the lock be passed even when the method isn't going to use it.

Firestore/core/src/util/executor_std.h
wilhuff: Once we need to acquire a mutex to check whether or not the executor has been disposed, it's harder to increment not holding the lock than to just continue holding it while incrementing. At that point the atomic is a waste.

Firestore/core/test/unit/util/executor_test.cc
wilhuff: 🤦

Firestore/core/test/unit/util/executor_test.cc
wilhuff: Done.

Firestore/core/test/unit/util/executor_test.cc
wilhuff: Done.

Firestore/core/test/unit/util/executor_test.cc
wilhuff: This was actually pretty critical--thanks for pushing on this.

Firestore/core/src/util/executor_std.cc
var-const: I actually prefer `Schedule` being self-contained (efficiency concerns may of course override that, but I don't think we're at that point yet), so I'm certainly not pushing for something like that.

Firestore/core/src/util/executor.h
```diff
@@ -72,6 +72,25 @@ class Executor {
 
   virtual ~Executor() = default;
 
+  // Explicitly destroy this Executor, canceling pending tasks, and waiting for
+  // any tasks that are currently running.
+  //
+  // Dispose exists as a separate step to facilitate a leaf-first destruction
+  // order. Normally the root-most object in a hierarchy runs its destructor and
+  // then the objects that make it up are destroyed. If tasks referring to the
+  // root are running while the Executor's destructor is running, there's no
```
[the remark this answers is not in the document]
var-const: Thanks for adding the clarification. One last nit: `this` is a little ambiguous in `this prevents any tasks from running`, consider replacing it with a more specific description (is it calling `Dispose` that prevents tasks from running? Some specific part of the implementation of `Dispose`? etc.).

Firestore/core/src/util/executor_libdispatch.mm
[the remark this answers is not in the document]
var-const: I don't have a strong preference either way. Feel free to remove if you think it's extraneous.

Firestore/core/src/util/executor.h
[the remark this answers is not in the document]
wilhuff: I've reordered the sentence to make this clearer. The pattern itself ensures the outcome:

> By calling `Dispose` in its destructor, the root-most object prevents any tasks from running that could observe a partially, destroyed object graph.

PR #5631, opened 2020-05-18
Fix auth persistence logging and invalid type casting

pranavrajgopal:


FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -76,7 +76,7 @@ - (instancetype)initWithApp:(FIRApp *)app appInfo:(NSDictionary *)appInfo {
   self.authState = [FIRAppDistributionAuthPersistence retrieveAuthState:&authRetrievalError];
   // TODO (schnecle): replace NSLog statement with FIRLogger log statement
   if (authRetrievalError) {
-    NSLog(@"Found no tester auth token in keychain on intitialization");
+    NSLog(@"Found no tester auth token in keychain on intitialization. error? - %@", [authRetrievalError localizedDescription]);
```
schnecle, line 79: I don't think we want to error here - we expect this to find no tester in the keychain when they first start up. This isn't necessarily an error.

pranavrajgopal: Yeah. But I wanted to print the error just in case there was an actual error retrieving the keychain.

PR #5632, opened 2020-05-18
Make RemoveSnapshotsInSyncListener run on the async queue

wilhuff:


Firestore/core/src/core/firestore_client.cc
```diff
@@ -499,7 +499,9 @@ void FirestoreClient::AddSnapshotsInSyncListener(
 
 void FirestoreClient::RemoveSnapshotsInSyncListener(
     const std::shared_ptr<EventListener<Empty>>& user_listener) {
-  event_manager_->RemoveSnapshotsInSyncListener(user_listener);
+  worker_queue_->Enqueue([this, user_listener] {
+    event_manager_->RemoveSnapshotsInSyncListener(user_listener);
```
thebrianchen, line 503: Quick question: Is the main reason that we don't `std::move` the listener pointer on remove because we no longer care about reference counts when it's about to be deleted?

wilhuff: Moving is only useful when the parameter is accepted by value (or by rvalue reference). This gives the caller's copy away.

In this case, `RemoveSnapshotsInSyncListener` takes a const reference to the `user_listener` so moving doesn't do anything, but it's not bumping the reference count anyway.

https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/core/event_manager.h#L68

To answer your follow-up question, `RemoveSnapshotsInSyncListener` takes the listener by const reference because it's not going to store the listener anyway. It's uses the reference passed to find associated data but forgets about it immediately. You only want to take something like that by value if your method will keep it. In that case std::move in the caller gives the caller's copy away.

PR #5636, opened 2020-05-18
Add end to end integration tests with RC+ABT

christibbs:
Adds an integration testing app that fetches an A/B tested parameter from Remote Config using development versions of the RC and ABT SDKs. There is also functionality to fire custom events to Analytics.

Links up to an existing Firebase project, but API keys in `GoogleService-Info.plist` are obfuscated.

#no-changelog


christibbs:
> Nice!
> 
> nit(optional): you can also add a workflow to ensure your test app build: https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/messaging.yml#L127

Good idea! Will do in a follow-up PR.

PR #5637, opened 2020-05-18
Provide an arbitrary data persistence mechanism on storage instances

mikehaney24:
This could allow most singletons to no longer need NSCoding, they just store the individual fields they need with an appropriately unique key.

#no-changelog

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -211,12 +211,14 @@
 #pragma mark - GDTCORStorageProtocol
 
 - (void)libraryDataForKey:(nonnull NSString *)key
-               onComplete:(nonnull void (^)(NSData *_Nullable))onComplete {
+               onComplete:
+                   (nonnull void (^)(NSData *_Nullable, NSError *_Nullable error))onComplete {
   dispatch_async(_storageQueue, ^{
     NSString *dataPath = [[[self class] libraryDataPath] stringByAppendingPathComponent:key];
-    NSData *data = [NSData dataWithContentsOfFile:dataPath];
+    NSError *error;
+    NSData *data = [NSData dataWithContentsOfFile:dataPath options:0 error:&error];
     if (onComplete) {
-      onComplete(data);
+      onComplete(data, error);
     }
   });
 }
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -51,6 +51,27 @@
   return archivePath;
 }
 
++ (NSString *)libraryDataPath {
+  static NSString *libraryDataPath;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    libraryDataPath =
+        [GDTCORRootDirectory() URLByAppendingPathComponent:NSStringFromClass([self class])
+                                               isDirectory:YES]
+            .path;
+    libraryDataPath = [libraryDataPath stringByAppendingPathComponent:@"gdt_library_data"];
+    if (![[NSFileManager defaultManager] fileExistsAtPath:libraryDataPath isDirectory:NULL]) {
+      NSError *error;
+      [[NSFileManager defaultManager] createDirectoryAtPath:libraryDataPath
+                                withIntermediateDirectories:YES
+                                                 attributes:0
+                                                      error:&error];
+      GDTCORAssert(error == nil, @"Creating the library data path failed: %@", error);
+    }
+  });
+  return libraryDataPath;
+}
+
 + (instancetype)sharedInstance {
   static GDTCORFlatFileStorage *sharedStorage;
   static dispatch_once_t onceToken;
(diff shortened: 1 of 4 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -187,6 +208,39 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
+  dispatch_async(_storageQueue, ^{
+    NSError *error;
+    NSString *dataPath = [[[self class] libraryDataPath] stringByAppendingPathComponent:key];
+    if (data && data.length > 0) {
+      [data writeToFile:dataPath options:NSDataWritingAtomic error:&error];
+      if (onComplete) {
+        onComplete(error);
+      }
+    } else {
+      if ([[NSFileManager defaultManager] fileExistsAtPath:dataPath]) {
```
paulb777, line 234: Would the API be easier to use if there was a `removeLibraryData` method instead of a zero lengthed `NSData`?

Either way, please add a test for this path.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -51,6 +51,27 @@ + (NSString *)archivePath {
   return archivePath;
 }
 
++ (NSString *)libraryDataPath {
```
maksymmalyhin, line 54: nit: Since the method has a side effect (may create a directory) it may be useful to add mention about it in the method name to improve readability, e.g. `libraryDataPathCreateIfNotExists`. I don't have a strong opinion here, so leave it up to you.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
```diff
@@ -36,6 +38,31 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (void)storeLibraryData:(NSData *)data
+                  forKey:(NSString *)key
+              onComplete:(void (^)(NSError *_Nullable error))onComplete;
+
+/** Retrieves the stored data for the given key.
+ *
+ * @param key The key corresponding to the desired data.
+ * @param onComplete The callback to invoke with the data once it's retrieved.
+ */
+- (void)libraryDataForKey:(NSString *)key onComplete:(void (^)(NSData *_Nullable data))onComplete;
```
maksymmalyhin, line 56: Are there any potential read errors we may want to expose in the completion?

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
```diff
@@ -36,6 +38,31 @@ NS_ASSUME_NONNULL_BEGIN
 /** Removes the events from storage. */
 - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs;
 
+/** Persists the given data with the given key.
```
maksymmalyhin, line 41: The new methods look unrelated to the previously existing ones. Do you think a separate protocol (e.g. `GDTCORKeyValueStorage`) and implementation may be a better abstraction from this perspective?

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
```diff
@@ -36,6 +38,31 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @param data The data to store.
+ * @param key The unique key to store it to.
+ * @param onComplete An block to be run when storage of the data is complete.
+ */
+- (void)storeLibraryData:(NSData *)data
```
maksymmalyhin, line 47: I wonder what is the intended usage of the storage. Is the client still expected to use `NSKeyedArchiver` to create data from e.g. an `NSArray` instance? If so, do you think we may use `id<NSSecureCoding>` instead of `NSData` type for values and encapsulate `NSKeyed(Un)Archiver` logic in the storage?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
mikehaney24: That's not a bad idea--but these are methods that I want to require of all future storage implementations.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
mikehaney24: It's meant to store library state. A caller _can_ implement NSSecureCoding to generate its bytes, but it's not a hard requirement. How the caller generates bytes is a decision left to that specific class. Because several of the callers will be singletons, it's created a lot of NSCoding weirdness that this can help to avoid. I prefer to keep it as NSData because then NSCoding can be used when wanted, or, it can just be a string or whatever.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
mikehaney24: Done.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: You're right, but the way it is is more consistent with other similar folder functions in GDT.

paulb777:
Please fix GHA before merging.

PR #5639, opened 2020-05-18
Remove unused source file

### LINKED ISSUE #4334: Exception and crash due to forwarding class AndroidConfigFetchProto which is neither defined, nor implemented
Yes, seems to me in the right place.

Environment

  * Xcode version: Version 11.2 (11B52)
  * Firebase SDK version: 6.13.0
  * Component version: 4.4.3, 4.4.5

#### Steps to reproduce:

1. Download Firebase quickstart sample project for iOS
2. Add GoogleService-Info.plist into quickstart sample project with valid keys, so that project can launch without issues
3. Install pods
4. Replace source in AppDelegate.m with code from AppDelegate.txt attached
5. Run ConfigExample target on iOS simulator
6. Observe app crashes due to exception. Please see Xcode logs in file attached 

Please note in this example app crashes for simulator, but the same will happen for device as well. The real problem is that AndroidConfigFetchProto class is neither defined, nor implemented  but referenced in Config.pbobjc.h and Config.pbobjc.m in FirebaseRemoteConfig pod. 

#### Relevant Code:

[xcode_log_crash.txt](https://github.com/firebase/firebase-ios-sdk/files/3856200/xcode_log_crash.txt)
[AppDelegate.txt](https://github.com/firebase/firebase-ios-sdk/files/3856205/AppDelegate.txt)

paulb777:
In a discussion with @chliangGoogle, we realized that RemoteConfig no longer uses its own protos. It only uses protos from ABTesting. This PR removes the unused source and has the side effect of fixing #4334.

Fix #4334


FirebaseRemoteConfig.podspec
```diff
@@ -28,7 +28,6 @@ app update.
 
   base_dir = "FirebaseRemoteConfig/Sources/"
   s.source_files = base_dir + '**/*.[mh]'
-  s.requires_arc = base_dir + '*.m'
```
paulb777, line 31: This flag indicates that the other files should use arc.  Removing the flag indicates all files follow the default of building with arc.

FirebaseRemoteConfig.podspec
```diff
@@ -44,7 +43,6 @@ app update.
   s.dependency 'FirebaseInstallations', '~> 1.1'
   s.dependency 'GoogleUtilities/Environment', '~> 6.2'
   s.dependency 'GoogleUtilities/NSData+zlib', '~> 6.2'
-  s.dependency 'Protobuf', '~> 3.9', '>= 3.9.2'
```
chliangGoogle, line 47: do we still need this for abt?

paulb777: ABT defines the dependency itself. We still need the Protobuf related define flag though for including the protobuf headers from ABT.

paulb777:
Going to merge this one now. If there's additional feedback, I'll address in a subsequent PR.

PR #5640, opened 2020-05-19
Added StorageIntegration swift tests

ncooke3:
Adding several swift integration tests to FirebaseStorage

• Attempt to get metadata from an file without having permission
• Attempt to delete non-existing file
• Upload blank .jpg file to storage container
• Get file using [`write(toFile: completion:)`](https://github.com/firebase/firebase-ios-sdk/blob/b6cd99e30a9ad87d3f933bfef62ec4d7c0d78cd5/FirebaseStorage/Sources/Public/FIRStorageReference.h#L219) with a non-null `completion`

### DIFF
--- a/FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
+++ b/FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
@@ -423,14 +423,19 @@ class StorageIntegration: XCTestCase {
       let fileURL = tmpDirURL.appendingPathComponent("cookie.txt")
       ref.write(toFile: fileURL) { url, error in
         XCTAssertNil(error, "Error should be nil")
+
+        guard let url = url else {
+          XCTFail("Failed to unwrap url")
+          return
+        }
+        XCTAssertEqual(fileURL, url)
+
         do {
-          let url = try XCTUnwrap(url, "Failed to unwrap url")
-          XCTAssertEqual(fileURL, url)
           let stringData = try String(contentsOf: fileURL, encoding: .utf8)
           XCTAssertEqual(stringData, cookieString)
           expectation.fulfill()
         } catch {
-          XCTFail("Throw in url completion block")
+          XCTFail("Could not get String contents of fetched data")
         }
       }
     })
--- a/FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
+++ b/FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
@@ -92,6 +92,18 @@ class StorageIntegration: XCTestCase {
     waitForExpectations()
   }
 
+  func testUnauthenticatedGetMetadataUnauthorized() {
+    let expectation = self.expectation(description: #function)
+    let ref = storage.reference().child("ios/private/secretfile.txt")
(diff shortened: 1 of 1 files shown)


FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
```diff
@@ -357,6 +409,33 @@ class StorageIntegration: XCTestCase {
     waitForExpectations()
   }
 
+  func testUnauthenticatedSimpleGetFileWithCompletion() throws {
+    let expectation = self.expectation(description: #function)
+    let ref = storage.reference(withPath: "ios/public/cookie")
+    let tmpDirURL = URL(fileURLWithPath: NSTemporaryDirectory())
+    let fileURL = tmpDirURL.appendingPathComponent("cookie.txt")
+    let data = try XCTUnwrap("Here's a 🍪, yay!".data(using: .utf8), "Data construction failed")
```
paulb777, line 417: Nice cookie! :)

FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
```diff
@@ -357,6 +409,33 @@ class StorageIntegration: XCTestCase {
...
+      XCTAssertNotNil(metadata, "Metadata should not be nil")
+      XCTAssertNil(error, "Error should be nil")
+
+      ref.write(toFile: fileURL) { url, error in
+        XCTAssertNil(error, "Error should be nil")
+        do {
+          let url = try XCTUnwrap(url, "Failed to unwrap url")
+          XCTAssertEqual(fileURL, url)
+          let stringData = try String(contentsOf: fileURL, encoding: .utf8)
+          XCTAssertEqual(stringData, "Here's a 🍪, yay!")
```
ryanwilson, line 429: Nit: it's probably good to have this in a single place - defined as a String above before we get the data from it. Then we can compare string data to the constant instead of having the value in two different places.

FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
```diff
@@ -357,6 +409,33 @@ class StorageIntegration: XCTestCase {
...
+          expectation.fulfill()
+        } catch {
+          XCTFail("Throw in url completion block")
```
ryanwilson, line 432: Instead of wrapping this in a `do/catch` block you should be able to just throw since the function already mentions it can throw (on line 412).

ncooke3: @ryanwilson  I'm a little confused so I think I'm missing something. How do I throw the error if it happens in the closure? I don't think the closure is defined to throw right? I was following along what was done in this method [here](https://github.com/firebase/firebase-ios-sdk/blob/b90ddf1339c8ec4b2004faf416f100beb59d0ccc/FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift#L375)

ryanwilson: My bad, I was the confused one :) you're right, thanks. Maybe updating the comment to something like "Could not get String contents for data fetched" or something?

PR #5643, opened 2020-05-19
Use secure rules for Firebase Storage Integration tests

schmidt-sebastian:
This changes the Storage Integration Test to use Firebase Auth with Email/Password sign in. I also removed some duplicate tests.

Addresses internal b/155892305

#no-changelog


FirebaseStorage/Tests/Integration/FIRStorageIntegrationTests.m
```diff
@@ -136,37 +151,8 @@ - (void)testUnauthenticatedGetMetadata {
   [self waitForExpectations];
 }
 
-- (void)testUnauthenticatedUpdateMetadata {
```
schmidt-sebastian, line 139: There is already an update metadata test,

FirebaseStorage/Tests/Integration/FIRStorageIntegrationTests.m
```diff
@@ -205,27 +191,8 @@ - (void)testDeleteWithNilCompletion {
   [self waitForExpectations];
 }
 
-- (void)testUnauthenticatedSimplePutData {
```
schmidt-sebastian, line 208: This test also exists twice.

FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
```diff
@@ -12,23 +12,31 @@
...
+  static var app: FirebaseApp!
+  static var auth: Auth!
   var storage: Storage!
   static var once = false
 
   override class func setUp() {
     FirebaseApp.configure()
+    app = FirebaseApp.app()
+    auth = Auth.auth(app:app)
+    auth.signIn(withEmail: "test@example.com", password: "testing") { result, error in
```
schmidt-sebastian, line 32: Do we need to block on this?

paulb777: Yes. Good catch!

FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
```diff
@@ -705,6 +716,16 @@ class StorageIntegration: XCTestCase {
     waitForExpectations()
   }
 
+  private func signInAndWait() {
+    let expectation = self.expectation(description: #function)
```
maksymmalyhin, line 720: I wonder if `self` can be omitted here?

FirebaseStorage/Tests/Integration/FIRStorageIntegrationTests.m
```diff
@@ -23,32 +24,37 @@
...
- * }
+ * You also need to enable email/password sign in and add a test user in your
+ * Firebase Authentication settings. Your account credentials need to match
+ * the credentials defined in `kTestUser` and `kTestPassword`.
  *
  * You can define these access rights in the Firebase Console of your project.
  */
+
+NSString *const kTestUser = @"test@example.com";
+NSString *const kTestPassword = @"testing";
```
maksymmalyhin, line 52: BTW are public credentials an issue from b/155892305 perspective? We can move credentials to secrets (e.g. like Auth do).

schmidt-sebastian: We are planning to do so as well.

ncooke3:
Some of the Storage Swift Integration tests are now failing since they test with unauthenticated access. What should we do about the unauthenticated tests?

paulb777:
@ncooke3 Yep - Working on it now trying to figure out why the tests are still failing on CI, even though they succeed locally.

ncooke3:
Hmm does it have anything to do with the fact that the rules have changed for the storage container on the firebase console? When I run them locally, any current tests that does any writing fails because it is unauthenticated and the rules on the console have since changed to only allow writes if authenticated. Maybe this is what is making the CI fail?

paulb777:
@ncooke3 Yes - the rules have changed, but the tests should now be signing in to authenticate

paulb777:
Maybe the issue is auth is an instance, but signing in is singleton ... trying to fix that now - although I still don't understand why signing in doesn't seem to matter when running locally.

FirebaseStorage/Tests/SwiftIntegration/StorageIntegration.swift
paulb777: nope. Variable used within its own initial value

PR #5645, opened 2020-05-19
Fix xcresult log output in Travis

wilhuff:
This change:

  * Removes code folding from Travis/GHA output. Code folds were interacting poorly with Travis and causing output to be lost. This was causing xcresult log output to be lost.
  * Removes raw log output on failure. Travis terminates jobs that exceed its maximum log output length, so showing raw log output could cause failing jobs to be killed for unrelated reasons, obscuring the actual failure.
  * Fixes `xcresult_logs.py` when run against non-Run configurations. The leading part of the name can be "Run" or "Test" or others.


wilhuff:
Ideally we get at least one flake in the travis output so we get to see this working :-).

paulb777:
The Firestore travis log is very tedious to scroll through, but probably not worth worrying about since we can likely drop travis for Firestore in the next few months.  The Storage failures can be ignored or solved by updating to master.

wilhuff:
Agreed that it's tedious, but at least now it's showing the test output and if you look hard enough you can see which test case actually flaked.

wilhuff:
Database quickstarts are reliably timing out but that's unrelated to this change.

PR #5646, opened 2020-05-19
Better completion callback for activate

### LINKED ISSUE #3586: "Most recently fetched config is already activated" after remote config activate() call

paulb777:
Adds an activate API with a changed parameter in the callback. It replaces a deprecated API that set an internal error if there was no change.

Fixes #3586 

The integration test infrastructure is implemented in #5633 

Googlers: See accompanying API proposal.

### DIFF
--- a/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
+++ b/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
@@ -278,7 +278,16 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
   return didActivate;
 }
 
+- (void)activateWithCompletion:(FIRRemoteConfigActivateChangeCompletion)completion {
+  [self activateWithEitherHandler:completion deprecatedHandler:nil];
+}
+
 - (void)activateWithCompletionHandler:(FIRRemoteConfigActivateCompletion)completionHandler {
+  [self activateWithEitherHandler:nil deprecatedHandler:completionHandler];
+}
+
+- (void)activateWithEitherHandler:(FIRRemoteConfigActivateChangeCompletion)completion
+                deprecatedHandler:(FIRRemoteConfigActivateCompletion)deprecatedHandler {
   __weak FIRRemoteConfig *weakSelf = self;
   void (^applyBlock)(void) = ^(void) {
     FIRRemoteConfig *strongSelf = weakSelf;
@@ -286,29 +295,36 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
       NSError *error = [NSError errorWithDomain:FIRRemoteConfigErrorDomain
                                            code:FIRRemoteConfigErrorInternalError
                                        userInfo:@{@"ActivationFailureReason" : @"Internal Error."}];
-      if (completionHandler) {
+      if (completion) {
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-          completionHandler(error);
+          completion(NO, error);
         });
+      } else if (deprecatedHandler) {
+        deprecatedHandler(error);
       }
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000068", @"Internal error activating config.");
       return;
     }
     // Check if the last fetched config has already been activated. Fetches with no data change are
     // ignored.
+    NSLog(@"%f %f", strongSelf->_settings.lastETagUpdateTime, strongSelf->_settings.lastApplyTimeInterval);
     if (strongSelf->_settings.lastETagUpdateTime == 0 ||
         strongSelf->_settings.lastETagUpdateTime <= strongSelf->_settings.lastApplyTimeInterval) {
       FIRLogWarning(kFIRLoggerRemoteConfig, @"I-RCN000069",
                     @"Most recently fetched config is already activated.");
-      NSError *error = [NSError
-          errorWithDomain:FIRRemoteConfigErrorDomain
-                     code:FIRRemoteConfigErrorInternalError
-                 userInfo:@{
-                   @"ActivationFailureReason" : @"Most recently fetched config is already activated"
-                 }];
-      if (completionHandler) {
+      if (completion) {
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-          completionHandler(error);
+          completion(NO, nil);
+        });
+      } else if (deprecatedHandler) {
+        NSError *error = [NSError
+            errorWithDomain:FIRRemoteConfigErrorDomain
+                       code:FIRRemoteConfigErrorInternalError
+                   userInfo:@{
+                     @"ActivationFailureReason" : @"Most recently fetched config already activated"
+                   }];
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+          deprecatedHandler(error);
         });
       }
       return;
@@ -319,9 +335,13 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
     [strongSelf updateExperiments];
     strongSelf->_settings.lastApplyTimeInterval = [[NSDate date] timeIntervalSince1970];
     FIRLogDebug(kFIRLoggerRemoteConfig, @"I-RCN000069", @"Config activated.");
-    if (completionHandler) {
+    if (completion) {
       dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        completionHandler(nil);
+        completion(YES, nil);
+      });
+    } else if (deprecatedHandler) {
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        deprecatedHandler(nil);
       });
     }
   };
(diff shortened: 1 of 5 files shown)


FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
```diff
@@ -232,18 +238,23 @@ NS_SWIFT_NAME(RemoteConfig)
...
+/// of the app to take effect (depending on how config data is used in the app).
+/// @param completion Activate operation callback.
+- (void)activateWithCompletion:(nullable FIRRemoteConfigActivateChangeCompletion)completion;
+
 /// Applies Fetched Config data to the Active Config, causing updates to the behavior and appearance
 /// of the app to take effect (depending on how config data is used in the app).
 /// @param completionHandler Activate operation callback.
-- (void)activateWithCompletionHandler:(nullable FIRRemoteConfigActivateCompletion)completionHandler;
+- (void)activateWithCompletionHandler:(nullable FIRRemoteConfigActivateCompletion)completionHandler
+    DEPRECATED_MSG_ATTRIBUTE("Use -[FIRRemoteConfig activateWithHandler:] instead.");
```
paulb777, line 250: Swift friendly message.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -278,37 +278,54 @@ - (BOOL)activateFetched {
...
   __weak FIRRemoteConfig *weakSelf = self;
   void (^applyBlock)(void) = ^(void) {
     FIRRemoteConfig *strongSelf = weakSelf;
     if (!strongSelf) {
       NSError *error = [NSError errorWithDomain:FIRRemoteConfigErrorDomain
                                            code:FIRRemoteConfigErrorInternalError
                                        userInfo:@{@"ActivationFailureReason" : @"Internal Error."}];
-      if (completionHandler) {
+      if (completion) {
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
chliangGoogle, line 299: callback should be called in main queue, otherwise might block UI. (altho seems not related to this PR, although seems like it's using a bad practice for all following places)

FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
```diff
@@ -87,6 +87,12 @@ typedef void (^FIRRemoteConfigFetchCompletion)(FIRRemoteConfigFetchStatus status
 typedef void (^FIRRemoteConfigActivateCompletion)(NSError *_Nullable error)
     NS_SWIFT_NAME(RemoteConfigActivateCompletion);
 
+/// Completion handler invoked by activate method upon completion.
+/// @param changed  Flag indicating whether or not the configuration changed.
+/// @param error  Error message on failure. Nil if activation was successful.
+typedef void (^FIRRemoteConfigActivateChangeCompletion)(BOOL changed, NSError *_Nullable error)
```
chliangGoogle, line 93: can you illustrate a bit on what the advantage is. I'm guessing using the boolean as first parameter will improve checking in swift but I don't know what exactly it is.

FirebaseRemoteConfig/Tests/HermeticAPI/APITests.swift
```diff
@@ -130,6 +125,41 @@ class APITests: XCTestCase {
     waitForExpectations()
   }
 
+  // Test New API.
+  func testUnchangedActivateWillFlag() {
+    let expectation = self.expectation(description: #function)
+    config.fetch { status, error in
```
chliangGoogle, line 131: Will this also help us adding ResultType?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Probably should be a separate PR, right?

FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
paulb777: It gives feedback to the API client about whether or not the config changed without an error being thrown. See #3586 for more on the issues caused by the other API.

FirebaseRemoteConfig/Tests/HermeticAPI/APITests.swift
paulb777: Yep, but adding a Swift extension for RemoteConfig is out of scope for now.

FirebaseRemoteConfig/Sources/Public/FIRRemoteConfig.h
chliangGoogle: Nice, so this will align with Android.

paulb777: And JS :)

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -110,6 +105,41 @@ class APITests: APITestBase {
...
+    let expectation = self.expectation(description: #function)
+    config.fetch { status, error in
+      if let error = error {
+        XCTFail("Fetch Error \(error)")
+      }
+      XCTAssertEqual(status, RemoteConfigFetchStatus.success)
+      self.config.activate { changed, error in
+        XCTAssertTrue(!APITests.useFakeConfig || changed)
+        if let error = error {
+          print("Activate Error \(error)")
```
karenyz, line 120: should this use `XCTFail` like it does above?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -278,37 +278,56 @@ - (BOOL)activateFetched {
...
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000068", @"Internal error activating config.");
       return;
     }
     // Check if the last fetched config has already been activated. Fetches with no data change are
     // ignored.
+    NSLog(@"%f %f", strongSelf->_settings.lastETagUpdateTime,
+          strongSelf->_settings.lastApplyTimeInterval);
     if (strongSelf->_settings.lastETagUpdateTime == 0 ||
         strongSelf->_settings.lastETagUpdateTime <= strongSelf->_settings.lastApplyTimeInterval) {
       FIRLogWarning(kFIRLoggerRemoteConfig, @"I-RCN000069",
```
karenyz, line 316: since we're no longer treating this as an error I wonder if we should downgrade the log severity

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
paulb777: This new API should assert on error being nil. Good catch!

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Done. Thanks!

FirebaseRemoteConfig/Tests/FakeConsole/FakeConsoleTests.swift
```diff
@@ -59,6 +60,43 @@ class FakeConsoleTests: APITestBase {
...
+  func testChangedActivateWillNotFlag() {
+    let expectation = self.expectation(description: #function)
+    config.fetch { status, error in
+      if let error = error {
+        XCTFail("Fetch Error \(error)")
+      }
+      XCTAssertEqual(status, RemoteConfigFetchStatus.success)
+      self.config.activate { changed, error in
+        XCTAssertNil(error)
+        XCTAssert(changed)
```
ryanwilson, line 74: Tiny nit: Can you change this to `XCTAssertTrue`? I did a double take on it, I think explicitly saying `True` makes it more clear.

FirebaseRemoteConfig/Tests/FakeConsole/FakeConsoleTests.swift
```diff
@@ -59,6 +60,43 @@ class FakeConsoleTests: APITestBase {
     waitForExpectations()
   }
 
+  // Test New API.
+  // Contrast with testUnchangedActivateWillFlag in APITests.swift.
+  func testChangedActivateWillNotFlag() {
```
ryanwilson, line 65: Should this also test the `false, nil` case where activate is called but no changes happened?

FirebaseRemoteConfig/Tests/FakeConsole/FakeConsoleTests.swift
paulb777: Agreed! Done

FirebaseRemoteConfig/Tests/FakeConsole/FakeConsoleTests.swift
paulb777: Yes, but that test doesn't need the FakeConsole and was added to APITests.swift

PR #5656, opened 2020-05-20
FIS: disable Auth token auto fetch

maksymmalyhin:
b/156746574


maksymmalyhin:
> This will be consistent with android right?

It should be. @ankitaj224 could you please confirm?

PR #5665, opened 2020-05-21
Fixing unwanted pending dynamic links checks on subsequent app restarts

eldhosembabu:
Fix for b/154589338

Most of the times, when the app is cold started, we are making api calls to:

https://firebasedynamiclinks-ipv4.googleapis.com/v1
https://firebasedynamiclinks-ipv6.googleapis.com/v1

Until we get a successful response. But sometimes, if the device doesn’t have support for v6 endpoint and if the v4 endpoint doesn’t retrieve a valid dynamic link, it will return error. This mostly happens when user directly goes to App Store and installs an app which has fdl sdk integrated.

For dynamic links, we want the app to retrieve the pending dynamic link on the very first start of the app after install so that 3p developers can deeplink to a particular page in the app. If the pending links retrieval failed during the first start, there is no point in retrying the retrieval during next restart since it could result in an un expected  deep link for the user when they opens up the app for subsequent restarts (if a valid dynamic link is returned).

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
@@ -554,9 +554,14 @@ static const NSInteger FIRErrorCodeDurableDeepLinkFailed = -119;
   self.retrievingPendingDynamicLink = NO;
   _retrievalProcess = nil;
 
-  if (!result.error && ![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
+  if (![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
+    //Once we complete the Pending dynamic link retrival, regardless of whether the retrival is
+    //success or failure, we don't want to do the retrival again on next app start.
+    //If we try to redo the retrival again because of some error, the user will experiance unwanted
+    //deeplinking when they restart the app next time.
     [_userDefaults setBool:YES forKey:kFIRDLOpenURLKey];
   }
+
   NSURL *linkToPassToApp = [result URLWithCustomURLScheme:_URLScheme];
   [self passRetrievedDynamicLinkToApplication:linkToPassToApp];
 }
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
@@ -555,10 +555,10 @@ static const NSInteger FIRErrorCodeDurableDeepLinkFailed = -119;
   _retrievalProcess = nil;
 
   if (![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
-    //Once we complete the Pending dynamic link retrival, regardless of whether the retrival is
-    //success or failure, we don't want to do the retrival again on next app start.
-    //If we try to redo the retrival again because of some error, the user will experiance unwanted
-    //deeplinking when they restart the app next time.
+    // Once we complete the Pending dynamic link retrival, regardless of whether the retrival is
+    // success or failure, we don't want to do the retrival again on next app start.
+    // If we try to redo the retrival again because of some error, the user will experiance unwanted
+    // deeplinking when they restart the app next time.
     [_userDefaults setBool:YES forKey:kFIRDLOpenURLKey];
   }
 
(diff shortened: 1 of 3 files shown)


FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -554,9 +554,14 @@ - (void)retrievalProcess:(id<FIRDLRetrievalProcessProtocol>)retrievalProcess
   self.retrievingPendingDynamicLink = NO;
   _retrievalProcess = nil;
 
-  if (!result.error && ![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
```
paulb777, line 557: Should all errors be disregarded here or just errors specific to ipv4/ipv6 not being found?

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -554,9 +554,14 @@ - (void)retrievalProcess:(id<FIRDLRetrievalProcessProtocol>)retrievalProcess
...
+  if (![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
+    //Once we complete the Pending dynamic link retrival, regardless of whether the retrival is
```
paulb777, line 558: s/retrival/retrieval/

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: It makes sense for all errors to be disregarded ; otherwise it will result in retry on next app start and if it returns a pending link during retry, it would be an unexpected deeplink for the user.
@dmandar let me know your thoughts if you think otherwise.

paulb777: Makes sense to me.

eldhosembabu:
> Thanks for the PR @eldhosembabu!
> 
> Please investigate adding a unit test.
> **Added a unit test** 
> Please fix the GHA issues by running clang-format. See https://github.com/firebase/firebase-ios-sdk#code-formatting
> **Done**
> While ok to put internal bug numbers in PRs, it should not be in the title. Instead a short description of the bug fix should be there.
> **Done, please let me know the description looks good.**
> Also the PR description should include a note about the GitHub issue being fixed. With "Fix #{bug number}", GitHub will automatically close the related bug when the PR merges.
> **The fix was based on an internal bug, didn't find a github issues to mention here.**

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: Done

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -554,9 +554,14 @@ - (void)retrievalProcess:(id<FIRDLRetrievalProcessProtocol>)retrievalProcess
   self.retrievingPendingDynamicLink = NO;
   _retrievalProcess = nil;
 
-  if (!result.error && ![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
+  if (![_userDefaults boolForKey:kFIRDLOpenURLKey]) {
+    // Once we complete the Pending dynamic link retrieval, regardless of whether the retrival is
+    // success or failure, we don't want to do the retrival again on next app start.
```
paulb777, line 559: Also "retrieval" here and on next line.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -554,9 +554,14 @@ - (void)retrievalProcess:(id<FIRDLRetrievalProcessProtocol>)retrievalProcess
...
+    // If we try to redo the retrival again because of some error, the user will experiance unwanted
```
paulb777, line 560: s/experiance/experience/

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: oops...didn't check that...done now

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: Done

eldhosembabu:
> Thanks for the followup!
> 
> I thought this was also a fix for #5032?
> **Its not actually, for fixing #5032 we need to optimize the calls i guess for that. Currently we are doing ipv4 and ipv6 calls parallel and consolidate the results which needs to be serialized. In this current fix, I'm just making sure we don't repeat checkforpending calls when there is an error in the previous one.**
> The description looks good. Please also add it to https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDynamicLinks/CHANGELOG.md. The version can be "Unreleased" for now.
> 
> Plus three more spelling nits
>**Done**

PR #5667, opened 2020-05-21
use UNNotificationRequest to schedule local notification for iOS 10 and above

chliangGoogle:
This fixes b/156310608 also as the UILocalNotification doesn't properly process "%", where it escape it in notification title and body.

Also UILocalNotification is deprecated, so use them for iOS 10 above.

### DIFF
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2020-06 -- v4.4.2
-- [changed] Use UNNotificationRequest to schedule local notification for local timezone notification for iOS 10 ang above. (#5667)
+- [changed] Use UNNotificationRequest to schedule local notification for local timezone notification for iOS 10 and above. This should also fix the issue that '%' was not properly shown in title and body. (#5667)
 
 # 2020-05 -- v4.4.1
 - [changed] Updated NSError with a failure reason to give more details on the error. (#5511)
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2020-06 -- v4.4.2
+# unreleased
 - [changed] Use UNNotificationRequest to schedule local notification for local timezone notification for iOS 10 and above. This should also fix the issue that '%' was not properly shown in title and body. (#5667)
 
 # 2020-05 -- v4.4.1
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2020-06 -- v4.4.2
+- [changed] Use UNNotificationRequest to schedule local notification for local timezone notification for iOS 10 ang above. (#5667)
+
 # 2020-05 -- v4.4.1
 - [changed] Updated NSError with a failure reason to give more details on the error. (#5511)
 
--- a/FirebaseMessaging/Sources/FIRMMessageCode.h
+++ b/FirebaseMessaging/Sources/FIRMMessageCode.h
@@ -96,7 +96,7 @@ typedef NS_ENUM(NSInteger, FIRMessagingMessageCode) {
   kFIRMessagingMessageCodeContextManagerService003 = 6003,                  // I-FCM006003
   kFIRMessagingMessageCodeContextManagerService004 = 6004,                  // I-FCM006004
   kFIRMessagingMessageCodeContextManagerService005 = 6005,                  // I-FCM006005
-  kFIRMessagingMessageCodeContextManagerServiceFailedLocalSchedule = 6006,  // I-FCM006005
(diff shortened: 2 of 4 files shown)


FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2020-06 -- v4.4.2
+- [changed] Use UNNotificationRequest to schedule local notification for local timezone notification for iOS 10 ang above. (#5667)
```
paulb777, line 2: s/ang/and/

FirebaseMessaging/Sources/FIRMMessageCode.h
```diff
@@ -90,12 +90,13 @@ typedef NS_ENUM(NSInteger, FIRMessagingMessageCode) {
...
-  kFIRMessagingMessageCodeContextManagerService003 = 6003,  // I-FCM006003
-  kFIRMessagingMessageCodeContextManagerService004 = 6004,  // I-FCM006004
-  kFIRMessagingMessageCodeContextManagerService005 = 6005,  // I-FCM006005
+  kFIRMessagingMessageCodeContextManagerService000 = 6000,                  // I-FCM006000
+  kFIRMessagingMessageCodeContextManagerService001 = 6001,                  // I-FCM006001
+  kFIRMessagingMessageCodeContextManagerService002 = 6002,                  // I-FCM006002
+  kFIRMessagingMessageCodeContextManagerService003 = 6003,                  // I-FCM006003
+  kFIRMessagingMessageCodeContextManagerService004 = 6004,                  // I-FCM006004
+  kFIRMessagingMessageCodeContextManagerService005 = 6005,                  // I-FCM006005
+  kFIRMessagingMessageCodeContextManagerServiceFailedLocalSchedule = 6006,  // I-FCM006005
```
paulb777, line 99: s/5/6/

FirebaseMessaging/CHANGELOG.md
paulb777, line 2: Should fixing the problem with the percent character be mentioned?

FirebaseMessaging/Sources/FIRMessagingContextManagerService.m
```diff
@@ -129,7 +136,70 @@ + (BOOL)handleContextManagerLocalTimeMessage:(NSDictionary *)message {
...
 
++ (void)scheduleiOS10LocalNotificationForMessage:(NSDictionary *)message atDate:(NSDate *)date {
+  NSCalendar *calendar = [NSCalendar currentCalendar];
+  if (@available(macOS 10.14, iOS 10.0, watchOS 3.0, tvOS 10.0, *)) {
+    NSCalendarUnit unit = NSCalendarUnitYear | NSCalendarUnitMonth | NSCalendarUnitDay |
+                          NSCalendarUnitHour | NSCalendarUnitMinute | NSCalendarUnitSecond;
+    NSDateComponents *dateComponents = [calendar components:(NSCalendarUnit)unit fromDate:date];
+    UNCalendarNotificationTrigger *trigger =
+        [UNCalendarNotificationTrigger triggerWithDateMatchingComponents:dateComponents
+                                                                 repeats:YES];
```
maksymmalyhin, line 147: I wonder why `repeats` is `YES`? It looks like it won't repeat any time soon with the specified calendar units (maybe in the next era maybe :) ).

FirebaseMessaging/Sources/FIRMessagingContextManagerService.m
```diff
@@ -129,7 +136,70 @@ + (BOOL)handleContextManagerLocalTimeMessage:(NSDictionary *)message {
...
+                     kFIRMessagingMessageCodeContextManagerServiceFailedLocalSchedule,
+                     @"Failed scheduling local timezone notification: %@.", error);
+               }
+             }];
+  }
+}
+
 + (void)scheduleLocalNotificationForMessage:(NSDictionary *)message atDate:(NSDate *)date {
+  if (@available(macOS 10.14, iOS 10.0, watchOS 3.0, tvOS 10.0, *)) {
+    [self scheduleiOS10LocalNotificationForMessage:message atDate:date];
```
maksymmalyhin, line 200: I wonder if we need to make any validation of the date (e.g. if it is in the past or in a distant future)?

FirebaseMessaging/Sources/FIRMessagingContextManagerService.m
chliangGoogle: Good catch. I think this should be no as backend every day will send a new scheduled notification. And client has no way to set when to end, so this should be a backend control thing. I will double test again today and tomorrow.

FirebaseMessaging/Sources/FIRMessagingContextManagerService.m
chliangGoogle: We do check it here: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/FIRMessagingContextManagerService.m#L99

And if it's very near future we schedule right away.

FirebaseMessaging/CHANGELOG.md

PR #5669, opened 2020-05-22
Added Core FirebaseApp swift tests

ncooke3:
- Adding `FirebaseAppTests.swift` that aims to test the [FirebaseApp](https://firebase.google.com/docs/reference/swift/firebasecore/api/reference/Classes/FirebaseApp) api

- Adding an **objc** class, `ExceptionCatcher`, that you can call in Swift to wrap code that may throw objc originating exceptions. You could then handle the exception as a normal swift error. 

- Changed the `FirebaseCore.podspec` file to provide the `ExceptionCatcher`class and sample `GoogleService-Info.plist` for testing


FirebaseCore/Tests/Unit/Swift/FirebaseAppTests.swift
```diff
@@ -0,0 +1,415 @@
...
+  func testConfigureDefaultAppTwice() {
+    FirebaseApp.configure()
+
+    let secondConfigurationAttempt = {
+      try ExceptionCatcher.catchException {
+        FirebaseApp.configure()
+      }
+    }
+
+    XCTAssertThrowsError(try secondConfigurationAttempt())
```
ncooke3, line 92: This test shows how you could use the `ExceptionCatcher` class. Configuring a FirebaseApp twice throws an exception, and I couldn't test the Swift API for this because the thrown objc exception can't be handled by in swift. This class allows you to provide its `catchException` method with a closure that contains code that might throw an objc exception and then handle it sorta? elegantly in swift.

FirebaseCore/Tests/Unit/Swift/SwiftTestingUtilities/ExceptionCatcher.h
```diff
@@ -12,11 +12,17 @@
 // See the License for the specific language governing permissions and
```
ncooke3, line 12: I'm not sure why it says I renamed `FIRAppTests.swift` to `ExceptionCatcher.h`. I used `git mv` to keep the `FIRAppTests.swift` history when I renamed it to `FirebaseAppTests.swift`. I kept trying to correct it but couldn't get it right. I read that git can diff based on the contents and I think that is part of the problem because every time I got the history corrected, it would revert back when I added the contents of each file back in :(

ncooke3:
Closing in favor of #5678

PR #5681, opened 2020-05-26
Migrate RTDB to GHA

paulb777:
Migrate RTDB CI to GHA

Disable two flaky integration tests that got flaky on travis with emulator switch and even more flaky on GHA.

Disable an always failing macOS unit test related to floating point double underflow

### DIFF
--- a/.github/workflows/database.yml
+++ b/.github/workflows/database.yml
@@ -14,6 +14,29 @@ on:
     - cron:  '0 7 * * *'
 
 jobs:
+  database:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        target: [iOS, tvOS, macOS]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Setup build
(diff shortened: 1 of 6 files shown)


.github/workflows/database.yml
```diff
@@ -14,6 +14,39 @@ on:
...
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        target: [iOS, tvOS, macOS]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: BuildAndTest # can be replaced with pod lib lint with CocoaPods 1.10
+      run: scripts/third_party/travis/retry.sh scripts/build.sh Database ${{ matrix.target }} unit
```
paulb777, line 27: Still doing the retry here - but yep - subsequent tries typically get worse - there seems to be something fundamental that needs to be fixed with these tests - but not clear to me after a bit of debugging ...

schmidt-sebastian: After retrying for a couple hours last week I realized I should have spent that time fixing the flakes. But then the tests passed :)

PR #5682, opened 2020-05-26
Add a global function to get storage instances

mikehaney24:
Needed to find storage instances internal _and_ external to this library.

#no-changelog

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
@@ -19,6 +19,10 @@
 
 #import "GDTCORLibrary/Public/GDTCORConsoleLogger.h"
 
+id<GDTCORStorageProtocol> _Nullable GDTCORStorageInstanceForTarget(GDTCORTarget target) {
+  return [GDTCORRegistrar sharedInstance].targetToStorage[@(target)];
+}
+
 @implementation GDTCORRegistrar {
   /** Backing ivar for targetToUploader property. */
   NSMutableDictionary<NSNumber *, id<GDTCORUploader>> *_targetToUploader;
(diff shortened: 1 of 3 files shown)


paulb777:
By "global", do you mean "public"?

Shouldn't a new public API have a Changelog?

GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
```diff
@@ -19,6 +19,10 @@
 
 #import "GDTCORLibrary/Public/GDTCORConsoleLogger.h"
 
+id<GDTCORStorageProtocol> _Nullable GDTCORStorageInstanceForTarget(GDTCORTarget target) {
```
maksymmalyhin, line 22: Just curiosity: what are advantages of using a global C function vs a static method on e.g. `GDTCORRegistrar` for this case?

mikehaney24:
> By "global", do you mean "public"?
> 
> Shouldn't a new public API have a Changelog?

Added a CHANGELOG entry, though I'm leaving the version the same since I'm planning on landing a public API change when this library next ships (whenever that might be). I'll probably change the CHANGELOG at that point to declare the larger change.

GoogleDataTransport/GDTCORLibrary/GDTCORRegistrar.m
mikehaney24: The GDTCORRegistrar class doesn't expose any other internals, and since exposing the storage instances is only a thing needed as part of using an id<GDTCORStorageProtocol> it seemed to make sense to include it in that header. Just seemed like a more coherent API.

maksymmalyhin: Makes sense, thank you for explaining!

paulb777:
Our policy is to always ship master, so it really should be a 6.2.0 for tomorrow's code freeze or hold off on merging until after that.

mikehaney24:
> Our policy is to always ship master, so it really should be a 6.2.0 for tomorrow's code freeze or hold off on merging until after that.

I'm going to hold off on merging until after the next code freeze.

PR #5685, opened 2020-05-27
Import updates in unported tests

paulb777:
Fix imports in unit tests that have not yet been ported from v1 to v2

This addresses deleted file reference presubmit errors after copybara


paulb777:
cc: @karenyz @maksymmalyhin

PR #5690, opened 2020-05-27
Update GDT CHANGELOG and version

mikehaney24:


GoogleDataTransport/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v6.2.0
+- Added an API for arbitrary data persistence on storage instances.
```
maksymmalyhin, line 2: Do you think we know the eventual PR number for the change?

mikehaney24: Unfortunately, no.

PR #5692, opened 2020-05-28
Migrate Installations to GHA

paulb777:
Also delete some additional unneeded travis-specific code.

### DIFF
--- a/.github/workflows/installations.yml
+++ b/.github/workflows/installations.yml
@@ -35,8 +35,10 @@ jobs:
       run: echo "::set-output name=val::$([[ -z $plist_secret ]] && echo "0" || echo "1")"
     - name: Build and test
       run: |
-       export FIR_IID_INTEGRATION_TESTS_REQUIRED=${{ steps.secrets.outputs.val }}
-       scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseInstallations.podspec --platforms=${{ matrix.target }}
+       export FIS_INTEGRATION_TESTS_REQUIRED=${{ steps.secrets.outputs.val }}
+       export FIS_ALLOWS_INCOMPATIBLE_IID_VERSION="1"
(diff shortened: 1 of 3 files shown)


scripts/setup_quickstart.sh
```diff
@@ -46,23 +49,5 @@ if check_secrets; then
 
   # Add GoogleService-Info.plist to Xcode project
   ruby ../scripts/info_script.rb "${SAMPLE}"
-
```
paulb777, line 49: Quickstarts are now all run on GHA

.travis.yml
```diff
@@ -114,21 +114,6 @@ jobs:
       script:
         - travis_retry ./scripts/if_changed.sh ./scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=ios
 
-    - stage: test
```
paulb777, line 117: Already migrated to gha

.github/workflows/installations.yml
```diff
@@ -0,0 +1,102 @@
...
+      id: secrets
+      run: echo "::set-output name=val::$([[ -z $plist_secret ]] && echo "0" || echo "1")"
+    - name: Build and test
+      run: |
+       export FIS_INTEGRATION_TESTS_REQUIRED=${{ steps.secrets.outputs.val }}
+       export FIS_ALLOWS_INCOMPATIBLE_IID_VERSION="1"
+       scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseInstallations.podspec \
+         --platforms=${{ matrix.target }} --ignore-local-podspecs=FirebaseInstanceID.podspec
+
+  # TODO - Catalyst is disabled because `pod gen` does not have a way to get some dependent pods
```
maksymmalyhin, line 43: I think we can introduce an additional flag to disable specific tests requiring a specific IID version. We can address this later.

paulb777: SG - or do we need to keep testing the old version of IID?

PR #5698, opened 2020-05-29
Add FirebaseFirestoreSwift readme

morganchen12:


Firestore/Swift/README.md
```diff
@@ -0,0 +1,38 @@
...
+file that depends on it, since it's not automatically bundled with the
+Firestore module.
+
+```swift
+import Firebase
+import FirebaseFirestoreSwift
+```
+
+The FirebaseFirestoreSwift module does not provide any additional utility to
+Objective-C projects, and is thus not recommended for non-Swift projects.
```
wilhuff, line 25: "thus" is about the how of something and "therefore" is more about why--expressing a causal relationship (you mentally replace it with "because of this"). I think therefore is preferred here.

Also consider keeping the "is not" together.

Firestore/Swift/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+The FirebaseFirestoreSwift module is currently only available for installation
+through CocoaPods. To add FirebaseFirestoreSwift to your project, add the
+following to your Podfile
+
+```ruby
+pod 'FirebaseFirestoreSwift'
+```
+
+You will also need to import the FirebaseFirestoreSwift module in any source
```
paulb777, line 15: "also" redundant?

morganchen12: Done.

Firestore/Swift/README.md
morganchen12: Done.

PR #5710, opened 2020-06-01
Deprecate Remote Config typedefs

paulb777:
Stop using typedefs for completion blocks in public APIs so that Xcode is more usable in Swift:

Screen Shot 2020-06-01 at 7 26 29 AM

There is no change/impact on Objective C usability.

The API council suggested this change.  We're piloting in RemoteConfig and will follow up with the other libraries.

### DIFF
--- a/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
+++ b/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
@@ -235,7 +235,7 @@ static NSMutableDictionary<NSString *, NSMutableDictionary<NSString *, FIRRemote
 - (void)fetchAndActivateWithCompletionHandler:
     (FIRRemoteConfigFetchAndActivateCompletion)completionHandler {
   __weak FIRRemoteConfig *weakSelf = self;
-  FIRRemoteConfigFetchCompletion fetchCompletion =
+  void (^fetchCompletion)(FIRRemoteConfigFetchStatus status, NSError *error) =
       ^(FIRRemoteConfigFetchStatus fetchStatus, NSError *fetchError) {
         FIRRemoteConfig *strongSelf = weakSelf;
         if (!strongSelf) {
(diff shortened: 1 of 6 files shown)


AlbGarciam:
@ryanwilson @paulb777 Have you considered to adopt Result instead of having two separate variables for the error and success?

ryanwilson:
> @ryanwilson @paulb777 Have you considered to adopt Result instead of having two separate variables for the error and success?

Thanks for the question! We've done it so far for some other Firebase SDKs but haven't done this one yet. For Objective-C compatibility reasons we can't add Swift directly to the SDK but we'll have to add an additional pod (as you can see in [FirebaseStorageSwift.podspec](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorageSwift.podspec)). 

If this is something you're interested in we can certainly take a look at bumping the priority or can work together on a PR if you'd be interested in contributing as well.

paulb777:
The usability improvement has been migrated to #5748. The typedef deprecation part of this PR will remain open here pending an API review.

paulb777:
@ryanwilson As I revisit this PR, I'm wondering if it's overkill to deprecate the typedefs, since it makes Objective C usability more complicated. 

Would it be sufficient to stop using the typedefs in the public APIs so we improve Swift usability, but don't downgrade Objective C usability?

ryanwilson:
> Would it be sufficient to stop using the typedefs in the public APIs so we improve Swift usability, but don't downgrade Objective C usability?

So you mean keep the typedefs there for convenience for developers but not use them in the method signature?

I have mixed feelings there... there wouldn't always be an obvious type to choose, if the names were something like `FIRVoidBoolErrorBlock` it may be better? Although I feel like everyone can visit http://goshdarnblocksyntax.com/ for their own typedef. I guess having it in ObjC isn't terrible and just marking it as unavailable for Swift, I'd be fine with that but I don't know how helpful it would be.

paulb777:
After discussion, we're going to leave the typedefs for legacy Objective C apps. Closing.

PR #5711, opened 2020-06-01
CocoaPods-specific tag for the Firebase pod

paulb777:
To avoid confusion with Swift Package Manager


ryanwilson:
Also for any future readers: Swift Package Manager will automatically pull in the last tag that matches sem-ver, so the *moment* we tag a `X.Y.Z` that becomes available as an update for developers, which doesn't go well with our release process right now where we tag and stage the pending release before publishing the other release artifacts at a later date.

PR #5715, opened 2020-06-01
Standardize headers for SPM and Auth

paulb777:
Simple name for public headers in the library
Module imports for public headers from dependencies like:

```
#if SWIFT_PACKAGE
@import FirebaseCore;
#else
#import <FirebaseCore/FIRApp.h>
#import <FirebaseCore/FIROptions.h>
#endif
```

Also restore Sample to master until we're ready to get it working with SPM

### DIFF
--- a/FirebaseAuth/Tests/Sample/ApiTests/FIRAuthApiTestsBase.h
+++ b/FirebaseAuth/Tests/Sample/ApiTests/FIRAuthApiTestsBase.h
@@ -16,9 +16,9 @@
 
 #import <XCTest/XCTest.h>
 
-#import "FIRApp.h"
+#import <FirebaseCore/FIRApp.h>
 #import "AuthCredentials.h"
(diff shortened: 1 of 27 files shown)


FirebaseAuth/Tests/Sample/ApiTests/FIRAuthApiTestsBase.h
```diff
@@ -16,9 +16,9 @@
 
 #import <XCTest/XCTest.h>
 
-#import "FIRApp.h"
+#import <FirebaseCore/FIRApp.h>
```
ryanwilson, line 19: Does this worth with SwiftPM? Or are these tests just not running yet?

paulb777: Yep, no tests are running with SwiftPM - so this is just a revert back to master for now.

Issue #5686, opened 2020-05-27
Thrown exception in [FLevelDBStorageEngine runMigration]


gapl:
### Environment

  * Xcode version: v11.4.1
  * Firebase SDK version: v6.24.0
  * Firebase Component: Database
  * Component version: v6.24.0
  * Installation method: `Carthage`

### Exception

We're seeing the following exception raised in our Crashlytics reporting.

We're aware that this issue has been raised a couple of times ([#4471](https://github.com/firebase/firebase-ios-sdk/issues/4471), [#1171](https://github.com/firebase/firebase-ios-sdk/issues/1171)), but they've been simply closed due to lack of additional information. We're happy to provide any additional information that might help. The crash is occurring since we added the Database dependency to the project. Users experiencing the crash are on non-jailbroken devices and are experiencing the crash in foreground.

These are not new users, but users who have updated from the previous version of the app that did not include Firebase Database, it did however include other components from Firebase.

```
Fatal Exception: NSInternalInconsistencyException
Unrecognized database version:

0  CoreFoundation                 0x1bfcb0794 (Missing)
1  libobjc.A.dylib                0x1bf9d2bcc (Missing)
2  CoreFoundation                 0x1bfba6b98 (Missing)
3  AppName                        0x10481cd5c -[FLevelDBStorageEngine runMigration]
4  AppName                        0x10481cc48 -[FLevelDBStorageEngine initWithPath:]
5  AppName                        0x1048362d4 -[FRepo deferredInit]
6  libdispatch.dylib              0x1bf9759a8 (Missing)
7  libdispatch.dylib              0x1bf976524 (Missing)
8  libdispatch.dylib              0x1bf9228a4 (Missing)
9  libdispatch.dylib              0x1bf923294 (Missing)
10 libdispatch.dylib              0x1bf92c78c (Missing)
11 libsystem_pthread.dylib        0x1bf9c7b74 (Missing)
12 libsystem_pthread.dylib        0x1bf9ca740 (Missing)
```

schmidt-sebastian:
@gapl Thanks for filing this. Based on your error message, the "version" file is empty. This is a special file that the SDK uses to denote the database schema version. This file sometimes shows up in iCloud Drive and may have been edited by the user.

To mitigate the crash, I will add an exception to ignore empty version files.

PR #5719, opened 2020-06-02
Don't crash RTDB SDK if version file is empty

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-ios-sdk/issues/5686

### DIFF
--- a/Firebase/Database/CHANGELOG.md
+++ b/Firebase/Database/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v6.2.2
+- [fixed] Addressed crash that prevented the SDK from opening due to a corrupted versioning file.
+  (#3120)
+
 # v6.2.1
 - [fixed] Fixed documentation typos. (#5406, #5418)
 
--- a/Firebase/Database/CHANGELOG.md
+++ b/Firebase/Database/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v6.2.2
-- [fixed] Addressed crash that prevented the SDK from opening due to a corrupted versioning file.
-  (#3120)
+- [fixed] Addressed crash that prevented the SDK from opening when the versioning file was
+  corrupted. (#3120)
 
 # v6.2.1
 - [fixed] Fixed documentation typos. (#5406, #5418)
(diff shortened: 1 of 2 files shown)


Firebase/Database/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v6.2.2
+- [fixed] Addressed crash that prevented the SDK from opening when the versioning file was
+  corrupted. (#3120)
```
paulb777, line 3: Should be #5686

Firebase/Database/Persistence/FLevelDBStorageEngine.m
```diff
@@ -132,6 +132,9 @@ - (void)runMigration {
         }
     } else if ([oldVersion isEqualToString:kFPersistenceVersion]) {
         // Everythings fine no need for migration
+    } else if ([oldVersion length] == 0) {
```
paulb777, line 135: Should we instead eliminate `runMigration` completely, get rid of the file management, and add it back in the unlikely case that there is ever a version 2?

schmidt-sebastian: I think the code is meant to crash for downgrades - e.g. when we add version 2, version 1 clients are supposed to crash.

Firebase/Database/CHANGELOG.md
schmidt-sebastian: Thanks for catching.

Firebase/Database/Persistence/FLevelDBStorageEngine.m
paulb777: Understood, but isn't it unlikely we'll ever add a version 2? In which case it would be nice to stop managing something in the filesystem.  We could always turn this code on again if and when version 2 happens.

I'm ok with whatever you decide.

schmidt-sebastian: I do think it is unlikely, but I prefer that we leave this check since we cannot turn this check back on in code that we already shipped. Without this check, it would be a much more expensive to add schema v2, as it that would require us to ship an SDK with version 1 that rejected version 2 schema before even considering a migration.

PR #5725, opened 2020-06-02
Update quota exceeded error message

renkelvin:


FirebaseAuth/Sources/Utilities/FIRAuthErrorUtils.m
```diff
@@ -383,8 +383,8 @@
 /** @var kFIRAuthErrorMessageQuotaExceeded
     @brief Message for @c FIRAuthErrorCodeQuotaExceeded error code.
  */
-static NSString *const kFIRAuthErrorMessageQuotaExceeded = @"The phone verification quota for this "
-                                                            "project has been exceeded.";
+static NSString *const kFIRAuthErrorMessageQuotaExceeded = @"The quota for this operation "
```
paulb777, line 386: Will the definition of "operation" be clear to those who get the message?

renkelvin: It should be clear by checking where the error is raised. While we can distinguish different operations purely based on the error.

PR #5734, opened 2020-06-03
Build GDT and nanopb with Swift Package Manager

paulb777:
Public header Imports in .m's use relative path
PubIic imports in .h's use simple file name for intra-module publics
Public imports from other modules use module import for SPM

Using forked repo for nanopb for now.


paulb777:
Ack, I'll sort out the copybara implications before merging back to master. Thanks

PR #5736, opened 2020-06-03
Upload zip artifact to gcs after build.

granluo:
Zip build workflow will upload framework zip to GCS after the zip file is successfully built. The zip file will later be downloaded in sdk testing jobs.


.github/workflows/zip.yml
```diff
@@ -41,3 +43,16 @@ jobs:
         # Zip the entire output directory since the builder adds subdirectories we don't know the
         # name of.
         path: zip_output_dir
+    - name: Install gcloud tool
+      if: ${{ always() }}
+      run: scripts/install_access_gcloud.sh
+    - name: Access gcloud account
+      if: ${{ always() }}
```
ryanwilson, line 50: I think we can omit the `if` and it should always run - I could be wrong though.

.github/workflows/zip.yml
```diff
@@ -41,3 +43,16 @@ jobs:
         # Zip the entire output directory since the builder adds subdirectories we don't know the
         # name of.
         path: zip_output_dir
+    - name: Install gcloud tool
+      if: ${{ always() }}
+      run: scripts/install_access_gcloud.sh
```
ryanwilson, line 48: The script name is incorrect - should be `install_gcloud.sh`

.github/workflows/zip.yml
granluo: I think it would only run the rest if previous steps successfully passed. 
GHA have[ a doc ](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions)about if successful/failed steps

scripts/install_gcloud.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Copyright 2020 Google
```
paulb777, line 3: Google LLC

scripts/upload_to_gcs.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Copyright 2020 Google
```
paulb777, line 3: LLC

granluo:
> Have you turned off schedule and tested success and failure cases?

Not yet, will make couple test commits see if both success and failure cases work .

granluo:
Have both cases, success and failure, tested. Should be ready to submit.

PR #5737, opened 2020-06-03
Send pod SDK version instead of hardcoded value in RC fetch request

karenyz:
Currently we are sending an "internal" SDK version that is never updated. Using `FIRRemoteConfigPodVersion` to reflect actual SDK version usage

Tracked in internal bug 157147675

This change only affects internal logging #no-changelog

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigConstants.h
+++ b/FirebaseRemoteConfig/Sources/RCNConfigConstants.h
@@ -19,13 +19,6 @@
 #define RCN_SEC_PER_MIN 60
 #define RCN_MSEC_PER_SEC 1000
 
-/// Remote Config SDK internal version that is different than
-/// FIRRemoteConfigPodVersion. This is for config server to track down iOS
-/// client app version. Each version can only go up to 99.
-static const int kRCNMajorVersion = 1;
-static const int kRCNMinorVersion = 2;
-static const int kRCNPatchVersion = 10;
-
 /// Key prefix applied to all the packages (bundle IDs) in internal metadata.
 static NSString *const RCNInternalMetadataAllPackagesPrefix = @"all_packages";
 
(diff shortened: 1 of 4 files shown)


FirebaseRemoteConfig/Sources/RCNConfigConstants.h
```diff
@@ -19,13 +19,6 @@
 #define RCN_SEC_PER_MIN 60
 #define RCN_MSEC_PER_SEC 1000
 
-/// Remote Config SDK internal version that is different than
-/// FIRRemoteConfigPodVersion. This is for config server to track down iOS
```
chliangGoogle, line 23: Based on the description, it seems RC backend does track a separate version that is different than the pod version. I would double check with RC backend team to make sure this is no longer needed. 
Also considering this field seems expect a number, e.g. for v1.2 it's 10200. I wonder if this field should be passed in at all now that we are passing in a string "1.2.0"

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
```diff
@@ -354,8 +354,8 @@ - (NSString *)nextRequestWithUserProperties:(NSDictionary *)userProperties {
                                                                 FIRRemoteConfigAppVersion()]];
   ret = [ret stringByAppendingString:[NSString stringWithFormat:@", app_build:'%@'",
                                                                 FIRRemoteConfigAppBuildVersion()]];
-  ret = [ret stringByAppendingString:[NSString stringWithFormat:@", sdk_version:'%d'",
-                                                                FIRRemoteConfigSDKVersion()]];
+  ret = [ret stringByAppendingString:[NSString stringWithFormat:@", sdk_version:'%@'",
+                                                                FIRRemoteConfigPodVersion()]];
```
paulb777, line 358: Assuming you've confirmed the backend can handle a version string that includes periods versus the pure integer that was sent before?

FirebaseRemoteConfig/Sources/RCNConfigConstants.h
karenyz: Checked in the backend and we don't use `sdk_version` anywhere except for logging/metrics, and we recently fixed our logging to handle this field as a string (updated bug with details)

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
karenyz: yep! updated the internal bug with details

FirebaseRemoteConfig/Sources/RCNConfigConstants.h
chliangGoogle: lol looks like this should be updated for a long while. Thanks for doing it!

PR #5738, opened 2020-06-03
Implement helpers for storing and finding events

mikehaney24:
#no-changelog because more changes are incoming.

Essentially, this change is to allow the storing of events to the filesystem in 3 ways:
- A direct write of a GDTCOREvent to the root data path, using the eventID as the filename. This will allow fast finding of events by target.
- A symbolic link to the direct write that allows fast finding of events via the mappingID.
- A symbolic link to the direct write that allow fast finding of events via the qosTier.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -79,7 +79,7 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
   return eventDataPath;
 }
 
-+ (NSString *)libraryDataPath {
++ (NSString *)libraryDataStoragePath {
   static NSString *libraryDataPath;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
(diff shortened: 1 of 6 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -72,6 +100,66 @@ + (NSString *)libraryDataPath {
...
+    return [NSString stringWithFormat:@"%@/%ld/%@", baseEventPath, (long)target, qosTier];
+  }
+
+  // If a target, mappingID, and qosTier were all given, return a single target/qosTier/mappingID
+  // directory.
+  return
+      [NSString stringWithFormat:@"%@/%ld/%@/%@", baseEventPath, (long)target, qosTier, mappingID];
+}
+
++ (nullable NSArray<NSString *> *)searchPathsWithEventSelector:
```
maksymmalyhin, line 144: Is `nullable` actually required here? Doesn't look like the method ever returns `nil`

mikehaney24: Removed nullability.

mikehaney24:
> Nice. I wonder if there is a design doc or some high-level overview of the storage structure. Now it looks more or less understandable after reading the PR description, but I'm afraid it may be more difficult in a month or so.

Added some clarifying comments after I realized I was actually returning the wrong paths.

PR #3803, opened 2019-09-09
Fix header issues exposed by :generate_multiple_pod_projects

paulb777:
Fix #3796

- Fix imports in Private headers. 
- Stop referencing an internal typedef from a private header.

PR #5740, opened 2020-06-03
SPM for FIS and IID

paulb777:
Neither IID nor FIS follow the recommended structure at https://github.com/firebase/firebase-ios-sdk/blob/master/AddNewPod.md so this is an incremental update sufficient to get a build going in Swift Package Manager.

Public header Imports in .m's use relative path
PubIic imports in .h's use simple file name for intra-module publics
Public imports from other modules use module import for SPM
All other imports should be repo-relative

#no-changelog

### DIFF
--- a/Firebase/InstanceID/Private/FIRInstanceID+Private.h
+++ b/Firebase/InstanceID/Private/FIRInstanceID+Private.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#import <FirebaseInstanceID/FIRInstanceID.h>
-#import <FirebaseInstanceID/FIRInstanceIDCheckinPreferences.h>
+#import "FIRInstanceID.h"
+#import "Firebase/InstanceID/Private/FIRInstanceIDCheckinPreferences.h"
 
 /**
  *  @related FIRInstanceIDCheckinService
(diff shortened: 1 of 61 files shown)


paulb777:
I'll follow up with copybara script updates before merging back to master.  There may be more file organizational changes while getting SwiftPM working end-to-end.

Firebase/InstanceID/Private/FIRInstanceID+Private.h
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#import <FirebaseInstanceID/FIRInstanceID.h>
-#import <FirebaseInstanceID/FIRInstanceIDCheckinPreferences.h>
+#import "FIRInstanceID.h"
```
maksymmalyhin, line 17: It looks like a revert of [the change](https://github.com/firebase/firebase-ios-sdk/pull/3803/files#diff-6c6af05d6b145c06596c62b94d30aa18) I wonder if anything has changed since #3803 so using different import styles for the same header is not an issue any more?

paulb777: That's a good catch - I'll see if I can add a test for `:generate_multiple_pod_projects` to CI

PR #5741, opened 2020-06-03
Run remoteconfig framework testing.

granluo:
Run remoteconfig framework testing by tracking a latest commit hash zip file on GCS.


.github/workflows/remoteconfig.yml
```diff
@@ -61,6 +61,59 @@ jobs:
       # Only build the unit tests on Catalyst. Test stopped working when GHA moved to Xcode 11.4.1.
       run: scripts/test_catalyst.sh FirebaseRemoteConfig build FirebaseRemoteConfig-Unit-unit
 
+  quickstart_framework:
```
paulb777, line 64: This shouldn't be in this file since it has nothing to do with the current state of the remote config sources.

It should be in `zip.yml` or another file.

Otherwise, LGTM

granluo: Could we set a schedule like how zip.yml did with zip building? Then this job will only be triggered nightly.
If we put this job in zip.yml, then a nightly test of framework cannot light up the badge of remoteconfig.

paulb777: That would work. We should change the schedule so it runs at least 2 hours after the zip build - or even better - checks if there is a new zip archive.

granluo: Yep
1. will update the schedule to be 3 hours after zip building.
2. The zip workflow will update files on GCS and make sure all SDK testings have the latest zip framework to test. SDK zip testing will fail if zip workflow fails.

paulb777: Change the PR title to 4 hours.

PR #5742, opened 2020-06-03
Expand the storage protocol and provide an event iterator

mikehaney24:
#no-changelog because more changes are incoming.

This change expands the storage protocol API so that prioritizers don't need to exist. An uploader can complete the following by itself:

- Find the storage instance for the target it cares about
- Query storage if there are events
- Get an event iterator and process the events to construct a request
- Store uploader state using the library data API

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -408,8 +408,7 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
     NSString *libraryDataPath = [GDTCORFlatFileStorage libraryDataStoragePath];
     NSDirectoryEnumerator *enumerator =
         [[NSFileManager defaultManager] enumeratorAtPath:eventDataPath];
-    NSString *nextFile;
-    while ((nextFile = [enumerator nextObject])) {
+    while ([enumerator nextObject]) {
       NSFileAttributeType fileType = enumerator.fileAttributes[NSFileType];
       if ([fileType isEqual:NSFileTypeDirectory] == NO &&
           [fileType isEqual:NSFileTypeSymbolicLink] == NO) {
@@ -418,8 +417,7 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
       }
     }
     enumerator = [[NSFileManager defaultManager] enumeratorAtPath:libraryDataPath];
-    nextFile = nil;
-    while ((nextFile = [enumerator nextObject])) {
+    while ([enumerator nextObject]) {
       NSFileAttributeType fileType = enumerator.fileAttributes[NSFileType];
       if ([fileType isEqual:NSFileTypeDirectory] == NO &&
           [fileType isEqual:NSFileTypeSymbolicLink] == NO) {
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -361,8 +361,7 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
   NSFileManager *fm = [NSFileManager defaultManager];
   NSDirectoryEnumerator *enumerator = [fm enumeratorAtPath:searchPath];
   while ([enumerator nextObject]) {
-    if ([enumerator.fileAttributes[NSFileType] isEqual:NSFileTypeDirectory] == NO &&
-        [enumerator.fileAttributes[NSFileType] isEqual:NSFileTypeSymbolicLink] == NO) {
+    if ([enumerator.fileAttributes[NSFileType] isEqual:NSFileTypeRegular]) {
       return YES;
     }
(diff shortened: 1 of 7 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -354,6 +356,80 @@ - (void)removeLibraryDataForKey:(nonnull NSString *)key
   });
 }
 
+- (BOOL)hasEventsForTarget:(GDTCORTarget)target {
+  NSString *searchPath = [GDTCORFlatFileStorage pathForTarget:target qosTier:nil mappingID:nil];
+  NSFileManager *fm = [NSFileManager defaultManager];
+  NSDirectoryEnumerator *enumerator = [fm enumeratorAtPath:searchPath];
+  while ([enumerator nextObject]) {
+    if ([enumerator.fileAttributes[NSFileType] isEqual:NSFileTypeDirectory] == NO &&
```
paulb777, line 364: Why not test for `NSFileTypeRegular` instead?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -354,6 +356,80 @@ - (void)removeLibraryDataForKey:(nonnull NSString *)key
...
+                                                                 queue:_storageQueue];
+      iterator.eventFiles = filePaths;
+    }
+  });
+  return iterator;
+}
+
+- (void)purgeEventsFromBefore:(GDTCORClock *)beforeSnapshot
+                   onComplete:(void (^)(NSError *_Nullable error))onComplete {
+  // TODO(mikehaney24): Figure out how we're going to deal with an NS
```
paulb777, line 400: NS?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: That's probably a lot smarter.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Must've gotten sidetracked!

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -354,6 +356,76 @@ - (void)removeLibraryDataForKey:(nonnull NSString *)key
...
+  NSDirectoryEnumerator *enumerator = [fm enumeratorAtPath:searchPath];
+  while ([enumerator nextObject]) {
+    if ([enumerator.fileAttributes[NSFileType] isEqual:NSFileTypeRegular]) {
+      return YES;
+    }
+  }
+  return NO;
+}
+
+- (nullable id<GDTCORStorageEventIterator>)iteratorWithSelector:
```
maksymmalyhin, line 371: It would be nice to know why we need the iterator. Would enumeration using a block (e.g. like [NSArray](https://developer.apple.com/documentation/foundation/nsarray/1415846-enumerateobjectsusingblock?language=objc)) work better? Is there a way to avoid  using `dispatch_sync`?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorageIterator.m
```diff
@@ -0,0 +1,66 @@
...
+    return nil;
+  }
+  dispatch_queue_t queue = _queue;
+  if (!queue) {
+    GDTCORLogDebug(@"%@", @"iterator queue was nil, so -nextEvent will be nil.");
+    return nil;
+  }
+  __block GDTCOREvent *nextEvent;
+  dispatch_sync(queue, ^{
+    NSData *data = [NSData dataWithContentsOfFile:self->_eventFiles[self->_currentIndex]];
```
maksymmalyhin, line 51: It looks like it is possible that the file may become unavailable (deleted) by the time it is requested. Is it intended?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Mostly because it could be costly (in terms of memory and CPU) to block a thread while a set of events is recreated from storage. Doing it one by one means that an event can be recreated and destroyed one at a time.

maksymmalyhin: I should've asked my 3 questions separately to get answers to all of them 😄.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorageIterator.m
mikehaney24: Yes, this is a good point. I'm not sure how I'm going to address it yet. I have several options:
- check for file exists in -nextEvent
- introduce API to 'lock' files to prevent them from showing up in iterators
- defer finding file paths until -nextEvent (as you suggested), though this may not work because -nextEvent doesn't run on (and block) the storage queue, so the same problem could still occur.

PR #5749, opened 2020-06-04
Build Crashlytics in Swift Package Manager

paulb777:
Primarily done by converting to repo-relative headers.

Also changed GDT and GDTCCT to be one SPM target since I couldn't find a way to keep them separate as a dependency in SPM.

Open question is SPM may not have a good way to set the linker option - `-sectcreate __TEXT __info_plist`

Got a successful upload from the Crashlytics quickstart.


GoogleDataTransport.podspec
```diff
@@ -34,7 +34,7 @@ Shared library for iOS SDK data transport needs.
   s.tvos.frameworks = 'SystemConfiguration'
 
   header_search_paths = {
-    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}/GoogleDataTransport/"'
+    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}/"'
```
mikehaney24, line 37: Any particular reason for this change?

paulb777: To handle the repo-relative imports that enable GDT and GDTCCT to be in the same SPM target.

paulb777:
I'll redo this in more incremental PRs

Issue #5720, opened 2020-06-02
Upload file failed: Error Domain=NSPOSIXErrorDomain Code=100 "Protocol error"


thomasdao:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Version 11.5 (11E608c)
  * Firebase SDK version: Version 6.25.0
  * Firebase Component: Auth, Core, Firestore, Storage, Functions
  * Component version: complete Pod.lock file is at https://drive.google.com/file/d/1GIz_tK5zr67hTvBz1dpSUJ7f6c1sj0A-/view?usp=sharing
  * Installation method: CocoaPods
  * My iPhone: iPhone 7, iOS 13.4.1

#### Steps to reproduce:
My app let user picks a file from UIDocumentPickerViewController and upload to Firebase storage. However:
- If I picked images, `.txt` or `.json` files, they can be uploaded successfully to Firebase Storage
- If I selected a custom file like `Home Improvement.numbers` file from `iCloud Drive/Numbers` folder, the upload failed with below error. This happened with other extensions like `.noto`.

Task <B5FBCC04-CBF6-49BB-9D6F-D8B0BA153D1B>.<1> finished with error [100] Error Domain=NSPOSIXErrorDomain Code=100 "Protocol error" UserInfo={_NSURLErrorFailingURLSessionTaskErrorKey=BackgroundUploadTask <B5FBCC04-CBF6-49BB-9D6F-D8B0BA153D1B>.<1>, _kCFStreamErrorDomainKey=1, NSErrorPeerAddressKey={length = 28, bytes = 0x1c1e01bb 00000000 24046800 40050801 ... 0000200a 00000000 }, _kCFStreamErrorCodeKey=100, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "BackgroundUploadTask <B5FBCC04-CBF6-49BB-9D6F-D8B0BA153D1B>.<1>",
    "LocalUploadTask <B5FBCC04-CBF6-49BB-9D6F-D8B0BA153D1B>.<1>"
)}

#### Relevant Code:
Pick file from UIDocumentPickerViewController:
```swift
    func pickFile() {
        let filePicker = UIDocumentPickerViewController(documentTypes: ["public.item"], in: .import)
        filePicker.allowsMultipleSelection = false
        filePicker.delegate = self
        present(filePicker, animated: true)
    }
```

Upload file to Firebase storage
```swift
[...]

PR #5750, opened 2020-06-04
Add error for attempt to upload directory

ncooke3:
In issue #5720, [`putFile`](https://github.com/firebase/firebase-ios-sdk/blob/d6f3f8dc3f5514c23e6c9f251fae471f97884b60/FirebaseStorage/Sources/FIRStorageReference.m#L215) was used to upload a `.numbers` document. Since `.numbers` is actually a directory and not a file, the method resulted in an error. The error is a bit misleading as it appears to come from the URLSession upload task.

In an attempt to provide a better error for such cases, I added a check that validates that the `fileURL` passed into `putFile` is indeed a file and not a directory.

Since this is just a draft PR, I figured I could? include two documents (`Home Improvement.numbers` and `star_wars.pdf`) that support some initial tests I wrote for the addition I made to [`FIRStorageReference.m`](https://github.com/firebase/firebase-ios-sdk/compare/nc-storage-directory-upload-attempt-error?expand=1#diff-0861433ffc33da3e380b4bed37b1673e). If we decide to move forward with this, I would definitely take those out before opening a real PR 

I appreciate any feedback!


FirebaseStorage/Sources/FIRStorageReference.m
```diff
@@ -215,6 +215,13 @@ - (FIRStorageUploadTask *)putFile:(NSURL *)fileURL
 - (FIRStorageUploadTask *)putFile:(NSURL *)fileURL
                          metadata:(nullable FIRStorageMetadata *)metadata
                        completion:(nullable FIRStorageVoidMetadataError)completion {
+  if ([fileURL frs_hasDirectoryPath]) {
```
paulb777, line 218: Should we check that it is a file versus not a directory?  Like https://github.com/firebase/firebase-ios-sdk/pull/5742/files?

FirebaseStorage/Sources/FIRStorageUtils.h
```diff
@@ -108,4 +108,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface NSURL (FIRStorageNSURLHelpers)
+
+/**
+ * Returns a boolean indicating that the NSURL instance represents a file directory
+ * @return True if the NSURL object represents a file directory, and false if otherwise
+ */
+- (BOOL)frs_hasDirectoryPath;
```
schmidt-sebastian, line 117: Nit: The Firebase prefix for all Firebase API classes is `fir`. Should we use the same prefix here?

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -223,6 +223,20 @@ class StorageIntegration: XCTestCase {
...
+    let fileName = "star_wars.pdf"
+    let bundle = Bundle(for: StorageIntegration.self)
+    let filePath = try XCTUnwrap(bundle.path(forResource: fileName, ofType: ""),
+                                 "Failed to get filePath")
+    let ref = storage.reference(withPath: "ios/public/" + fileName)
+    ref.putFile(from: URL(fileURLWithPath: filePath)) { result in
+      self.assertResultSuccess(result)
+      expectation.fulfill()
+    }
+    waitForExpectations()
```
schmidt-sebastian, line 237: I wonder if this is substantially different from testSimplePutFile(). Do you think we should keep it?

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -256,6 +270,27 @@ class StorageIntegration: XCTestCase {
     waitForExpectations()
   }
 
+  func testAttemptToUploadDirectory() throws {
```
schmidt-sebastian, line 273: The test name should convey that the attempt is going to fail.

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -256,6 +270,27 @@ class StorageIntegration: XCTestCase {
...
+    let expectation = self.expectation(description: #function)
+
+    let fileName = "Home Improvement.numbers"
+    let bundle = Bundle(for: StorageIntegration.self)
+    let fileURL = try XCTUnwrap(bundle.url(forResource: fileName, withExtension: ""),
+                                 "Failed to get filePath")
+    let ref = storage.reference(withPath: "ios/public/" + fileName)
+    ref.putFile(from: fileURL) { result in
+      switch result {
```
schmidt-sebastian, line 282: I am not a Swift export, but is there a more succinct way to verify that the task has failed? If might also be possible to validate the error message, which would make this test very much self-explanatory.

ncooke3: True, I'll work on it!

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
ncooke3: Probably not! I was just trying things out with some different file types. I'll remove this test and `star_wars.pdf`. I agree that the current `testSimplePutFile()` is enough here.

I'll also remove the contents of the `Home Improvement.numbers` so we can have a test for attempting to upload a directory.

FirebaseStorage/Sources/FIRStorageUtils.h
ncooke3: Good point, I can fix that. FWIW, I prefixed it with `frs_` because I noticed the other categories in `FIRStorageUtils.h` had [methods prefixed with `frs_`](https://github.com/firebase/firebase-ios-sdk/blob/8d1e2464255365a88619225740ffedcf01195428/FirebaseStorage/Sources/FIRStorageUtils.h#L107). Just curious, was there a reason some of the other category methods in `FIRStorageUtils.h` were prefixed with `frs_`?

FirebaseStorage/Sources/FIRStorageReference.m
ncooke3: I will play with this. Maybe instead of the check looking like this:
```
 if fileUrl is a directory -> return error
```

We could instead have something like?:

```
 if !(fileUrl is a file) -> return error
```

maksymmalyhin: Good point, I think the 2nd option is preferable because there may be other types of URLs that are not handled now, e.g. a symlink or not a file system URL.

schmidt-sebastian: SGTM

ncooke3:
Looks like I broke some of the objc unit tests by making the change to check that the passed in `fileURL` is a file rather than a directory. The good news is the ones that failed were `testReferenceWithNilFileURLFailsWithCompletion` and `testReferenceWithNonexistentFileURLFailsWithCompletion` which were caught by the new isFile check. Should I fix the objc tests so that the assertions match?

paulb777:
Updating the tests SGTM since its improving the tests with a more accurate error message, right?

ncooke3:
> Updating the tests SGTM since its improving the tests with a more accurate error message, right?

Yup, I believe so. Up until now, it looks like errors related to an invalid `fileURL` input occur sometime during the upload process. With this check, if we start the upload process, then we should be able to assume that the `fileURL` is valid and any other errors are related to the actual upload process.

paulb777:
On second thought, I'm not completely certain, if the error message

FirebaseStorage/Sources/FIRStorageUploadTask.m
```diff
@@ -193,11 +193,14 @@ - (BOOL)isContentToUploadValid:(NSError **)outError {
   }
 
   NSError *fileReachabilityError;
-  if (![_fileURL checkResourceIsReachableAndReturnError:&fileReachabilityError]) {
+  if (![_fileURL checkResourceIsReachableAndReturnError:&fileReachabilityError] ||
+      ![self fileURLisFile:_fileURL]) {
```
ncooke3, line 197: The [`checkResourceIsReachableAndReturnError`](https://developer.apple.com/documentation/foundation/nsurl/1410411-checkpromiseditemisreachableandr?language=objc) returns true or false depending on if the `_fileURL` can be reached. This is why attempts to upload directories weren't being caught in the first place, because they are reachable. Instead of doing a check in `putFile:`, I figured I should move the `fileURLisFile` helper method to  `FIRStorageUploadTask.m` and call it in this `isContentToUploadValid` method so we have a single check to verify that the file we are trying to upload is both a reachable resource and **it is a file**.

FirebaseStorage/Sources/FIRStorageUploadTask.m
```diff
@@ -193,11 +193,14 @@ - (BOOL)isContentToUploadValid:(NSError **)outError {
...
     if (outError != NULL) {
       NSMutableDictionary *userInfo = [NSMutableDictionary dictionaryWithCapacity:2];
-      userInfo[NSLocalizedDescriptionKey] =
-          [NSString stringWithFormat:@"File at URL: %@ is not reachable.", _fileURL.absoluteString];
+      userInfo[NSLocalizedDescriptionKey] = [NSString
+          stringWithFormat:@"File at URL: %@ is not reachable. "
+                           @"Ensure file URL is not a directory, symbolic link, or invalid url.",
+                           _fileURL.absoluteString];
```
ncooke3, line 203: I added some extra context to the error since we are now explicitly checking if the url points to a file.

schmidt-sebastian: Looks good!

PR #5754, opened 2020-06-05
SPM for Functions

paulb777:
And remove Core, IID, and FIS from selectable products.

### DIFF
--- a/Functions/FirebaseFunctions/FIRFunctions.m
+++ b/Functions/FirebaseFunctions/FIRFunctions.m
@@ -15,11 +15,32 @@
 #import "FIRFunctions.h"
 #import "FIRFunctions+Internal.h"
 
+#if SWIFT_PACKAGE
+@import FirebaseCore;
+@import GTMSessionFetcherCore;
+#import "FirebaseCore/Sources/Private/FIRAppInternal.h"
+#import "FirebaseCore/Sources/Private/FIRComponent.h"
+#import "FirebaseCore/Sources/Private/FIRComponentContainer.h"
+#import "FirebaseCore/Sources/Private/FIRDependency.h"
+#import "FirebaseCore/Sources/Private/FIRLibrary.h"
+#import "Interop/Auth/Public/FIRAuthInterop.h"
+#else
 #import <FirebaseAuthInterop/FIRAuthInterop.h>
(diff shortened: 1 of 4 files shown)


Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -15,11 +15,32 @@
 #import "FIRFunctions.h"
 #import "FIRFunctions+Internal.h"
 
+#if SWIFT_PACKAGE
+@import FirebaseCore;
+@import GTMSessionFetcherCore;
+#import "FirebaseCore/Sources/Private/FIRAppInternal.h"
```
wilhuff, line 21: Why not adopt repo-relative imports for all private headers?

And for that matter, if that works, why not just use repo-relative imports for all the headers?

As it stands this is an unmaintainable mess. People will develop one way and fail to update the imports on the other side. This doesn't seem long term viable without finding a way to just have one set of imports.

paulb777: We should migrate to all repo-relative internal headers, but I'd prefer to work on standardization on master and separate from the Swift Package Manager branch.

Public headers cannot be repo relative because they'll generate conflicts with module-imports.  The exception is public header imports from a non-header source in the same target.

wilhuff: But surely there has to be a way to hide this so that not every client of FirebaseCore has to have this mess inside it.

Consider an alternative:

FirebaseCore has a private header that contains:

```objc
#if SWIFT_PACKAGE
@import FirebaseCore;
#else
#import <FirebaseCore/FirebaseCore.h>
#endif
```

Now functions is none the wiser about this; we just adopt the convention that imports from one SDK to another always use our private umbrella header.

As it stands, this is creating a mass of technical debt. Honestly, I can't even read what's going on here and have a hard time understanding whether or not imports are being handled equivalently on both sides of the `#else`. Now that storage is done I think we're past the point of prototyping and we need to address this issue.

paulb777: OK, addressing in #5758 and will rebase or reimplement this PR after that.

paulb777:
Replaced by #5858

PR #5755, opened 2020-06-05
Lazy initialize auth state

pranavrajgopal:
1. Added a Getter for isTesterSignedIn and lazy initialized the auth state in the get method
2. Also added auth state lazy initialization logic in signInTester
3. Also added a log message for when we don't find the auth state in the keychain

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -56,6 +56,13 @@ NSString *const kCodeHashKey = @"codeHash";
 NSString *const kAuthErrorMessage = @"Unable to authenticate the tester";
 NSString *const kAuthCancelledErrorMessage = @"Tester cancelled sign-in";
 
+@synthesize isTesterSignedIn = _isTesterSignedIn;
+
+- (BOOL)isTesterSignedIn {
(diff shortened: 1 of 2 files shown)


FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -346,6 +347,27 @@ - (void)handleOauthDiscoveryCompletion:(OIDServiceConfiguration *_Nullable)confi
...
+  if (self.authState) {
+    NSLog(@"Auth state already initialized.");
+    return true;
+  }
+
+  NSError *authRetrievalError;
+  self.authState = [self.authPersistence retrieveAuthState:&authRetrievalError];
+  // TODO (schnecle): replace NSLog statement with FIRLogger log statement
+  if (authRetrievalError) {
+    NSLog(@"Error retrieving tester auth token");
```
schnecle, line 362: I still think that this log statement is misleading because we will error anytime we don't find the authState. We only want this to tell us there is an error if we know you have signed in. This is only good enough to tell us that we were unable to retrieve the authState but not whether or not we should have been able to.

pranavrajgopal: Fair point. I'll fix it to not be an error log.

Issue #3777, opened 2019-09-05
FirebaseCore Private header cleanup


paulb777:
The private headers FIRErrors.h and FIRErrorCode.h is no longer being used and should be deleted in the next major release.

Some of the other headers are internal and should be moved out of private and repackaged as internal.

ryanwilson:
Just to confirm, moving out of private and repackaged as internal will match SwiftPM style, right?

paulb777:
Yes, both CocoaPods and SwiftPM support internal headers.

If SwiftPM doesn't add support for private, we'll likely also have to convert private headers to public - but that's outside the scope of this issue.

paulb777:
FIRBundleUtil.h and FIRVersion.h are the only private headers not currently used outside of Core.

paulb777:
FIRBundleUtil.h and FIRVersion.h are addressed in #4590

The header removal should still happen with Firebase 7

paulb777:
The reorganization is addressed in #5759. The FIRError.h cleanup still requires cleanup in internal Remote Config sources.

Related to b/147219276

paulb777:
Fixed by #6439 and will release with Firebase 7.

PR #5759, opened 2020-06-07
Separate Core private versus internal headers

paulb777:
Fix the reorganization part of #3777.

Based on testing in #5758, move Core headers only used by Core from "Private" to "Sources".


FirebaseCore/Tests/Unit/FIRAnalyticsConfigurationTest.m
```diff
@@ -16,7 +16,7 @@
 
 #import "FirebaseCore/Tests/Unit/FIRTestCase.h"
 
-#import <FirebaseCore/FIRAnalyticsConfiguration.h>
+#import "FirebaseCore/Sources/FIRAnalyticsConfiguration.h"
```
ryanwilson, line 19: May be a silly question but just to confirm for here and other tests - is it alright that we're not using framework imports in test code?

paulb777: Yes. The tests are in the repo and using repo-relative imports should allow Swift Package Manager to build them without change. We should leave framework import testing for integration tests.

PR #5760, opened 2020-06-07
Migrate from interop pods to source_files header access

paulb777:
Eliminate Interop pods and instead access their headers via a podspec `source_files` attribute.

- Change all interop imports to be repo-relative
- Go back to module-based imports for non Swift Package manager for public header imports from private headers. In addition to not working for Objective C++, they also don't work in some forward compatibility cases for using older pods with an updated FirebaseCore.


paulb777:
Ready for review. The one test failure is the known master MLKit symbol collision bug

FirebaseABTesting.podspec
```diff
@@ -33,14 +33,14 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
...
   s.private_header_files = base_dir + 'Protos/developers/mobile/abt/proto/*.h'
+  s.preserve_paths = 'Interop/Analytics/Public/*.h', 'FirebaseCore/Sources/Private/*.h'
   s.pod_target_xcconfig = {
     'GCC_C_LANGUAGE_STANDARD' => 'c99',
     'GCC_PREPROCESSOR_DEFINITIONS' =>
       'GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS=1 ' +
       'FIRABTesting_VERSION=' + String(s.version),
     'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
   }
-  s.dependency 'FirebaseAnalyticsInterop', '~> 1.3'
```
maksymmalyhin, line 43: I wonder if removing explicit versioned interop dependencies can make possible installing incompatible versions in case when developers lock particular pod versions? E.g. if FirebaseABTesting is locked to a version that requires FirebaseAnalyticsInterop  1.3  but a new Analytics version compatible with FirebaseAnalyticsInterop 2.0 is installed we may not get an error.

paulb777: There is no longer the concept of an Interop version. However, any breaking changes should be caught by CI.

maksymmalyhin: I see. I can imagine a case when an API change will be difficult to be caught by CI, e.g. if a new error is thrown but the method signature hasn't changed. Probably it's not too big concern but something to remember when making changes to API.

paulb777: Agreed. It should be standard practice to review all clients when making a breaking API change.

FirebaseCore.podspec
```diff
@@ -28,13 +28,14 @@ Firebase Core includes FIRApp and FIROptions which provide central configuration
   s.source_files = 'FirebaseCore/Sources/**/*.[mh]'
   s.public_header_files = 'FirebaseCore/Sources/Public/*.h', 'FirebaseCore/Sources/Private/*.h'
   s.private_header_files = 'FirebaseCore/Sources/Private/*.h'
+  s.preserve_paths = 'Interop/CoreDiagnostics/Public/*.h'
```
maksymmalyhin, line 31: @paulb777  I has tried this approach for another SDK and noticed that `preserve_paths` are not added to the project by `pod gen`. Do you think there is a nice workaround for this? Will it be appropriate to add this file to sources as well?

FirebaseCore/Sources/Private/FIRAppInternal.h
```diff
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 
-#if __cplusplus
-#import <FirebaseCore/FIRApp.h>
-#else
+#if SWIFT_PACKAGE
 #import "FIRApp.h"
+#else
+// Above is not findable in some CocoaPods scenarios, including Objective C++
```
wilhuff, line 20: This comment is talking about CocoaPods, C++, and Firebase versions, but the `#if` is referring to `SWIFT_PACKAGE`. Since It seems like SPM is the exception and we'd be interested in just finding the single common way to import things, maybe it be worth discussing why SPM can't use the module-qualified form? 

In its current form, I don't really understand this comment and it doesn't really help me get context. In which CocoaPods builds does this not work? If it's at all complicated, it's probably worth spelling this out in detail in your headers/imports doc and just linking here.

Also this is phrased in a way that seems to indicate all three of these conditions are required when I think it's any of them, right? At the very least change "and using" to "or using", but also consider splitting this into bullets and giving more details on what's wrong in each case.

Finally, what's special about FirebaseCore 6.27.0? Did it change something? It may be worth describing what changed about it that makes this version special.

FirebaseCoreDiagnostics.podspec
```diff
@@ -43,9 +44,10 @@ non-Cocoapod integration. This library also respects the Firebase global data co
   }.merge(header_search_paths)
 
   s.source_files = 'Firebase/CoreDiagnostics/FIRCDLibrary/**/*.[cmh]'
+  s.preserve_paths = 'Interop/CoreDiagnostics/Public/*.h'
```
wilhuff: Pods that implement an interop protocol should probably have these headers as part of their sources.

FirebaseABTesting.podspec
paulb777: That situation only gets caught now if the person remembers to update the podspec's major version and can only happen at a Firebase major version change.  In the future, if an interop header does a break all dependencies will also need a major version bump along with the Firebase pod major version bump.

FirebaseCore.podspec
paulb777: Switched from `preserve_paths` to `source_files`.

.github/workflows/firestore.yml
```diff
@@ -20,6 +20,9 @@ on:
     # Firestore sources
     - 'Firestore/**'
 
+    # Interop headers
+    - 'Interop/Auth/Public/*.h'
```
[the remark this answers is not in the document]
paulb777: I'm not sure. I've been avoiding adding all the transitive deps to avoid CI backlogs in favor of forcing build all's via a Gemfile change. The last time I researched I saw that GHA was planning to add a feature to force all tests to run, but it was not there yet.

FirebaseABTesting.podspec
```diff
@@ -33,14 +33,14 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
   s.requires_arc = base_dir + '*.m'
   s.public_header_files = base_dir + 'Public/*.h', base_dir + 'Protos/developers/mobile/abt/proto/*.h'
   s.private_header_files = base_dir + 'Protos/developers/mobile/abt/proto/*.h'
+  s.preserve_paths = 'Interop/Analytics/Public/*.h', 'FirebaseCore/Sources/Private/*.h'
```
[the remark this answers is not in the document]
paulb777: Done

FirebaseCore/Sources/Private/FIRAppInternal.h
paulb777: Removed confusing comment in favor of a TODO when I get back to Swift PM.  For now, these three headers are unchanged for CocoaPods.

FirebaseCoreDiagnostics.podspec
paulb777: Agreed - done.

Interop/FirebaseComponentSystem.md
```diff
@@ -350,8 +347,10 @@ get back to this spot as it lays the groundwork necessary to understand this sec
 Adding dependencies is easy once components are registered with Core. Let's take the example from
 Functions above and add a dependency to `FIRAuthInterop` defined above.
 
-**Important**: You will also need to add a dependency on the `FirebaseAuthInterop` pod to your
-               product's podspec and any package manager supported.
+**Important**: You will also need to add `FirebaseAuthInterop` headers to your
+               product's podspec `preserved_paths` attribute for CocoaPods and something
```
maksymmalyhin, line 351: Should be updated as well as the PR title, I guess.

PR #5765, opened 2020-06-08
Header/Import Guidelines

paulb777:


HeadersImports.md
```diff
@@ -0,0 +1,68 @@
...
+In Swift Package Manager, it's a library target.
+
+## Header File Types and Locations
+
+* *Public Headers* - Headers that define the library's API. They should be located in
+  `FirebaseFoo/Sources/Public`. Any additions require a minor version update. Any changes or
+  deletions require a major version update.
+
+* *Public Umbrella Header* - A single header that includes the full library's public API located at
+  `FirebaseFoo/Sources/Public/FirebaseFoo.h`. This header should be included in
```
wilhuff, line 18: Minor nit: even though we're talking about headers and they are included by the compiler, "included in" still reads as if you're talking about copying it into `Firebase.h`, when you really mean a new reference should be added.

Consider "included from" or rephrasing to more directly/concretely state what you want people to do, e.g.

> Add a new `__has_include` stanza for your umbrella header to Firebase.h.

Also, this is written as if this is unconditional, but really this should only apply to new products, right? For example, when adding GoogleDataTransport, we shouldn't add it to `Firebase.h`, right?

HeadersImports.md
```diff
@@ -0,0 +1,68 @@
...
+   [Firebase.h](CoreOnly/Sources/Firebase.h).
+
+* *Library Internal Headers* - Headers that are only used by the enclosing library. These headers
```
wilhuff, line 21: It seems like this section should be last. It would make these ordered by public/private/internal

HeadersImports.md
```diff
@@ -0,0 +1,68 @@
...
+  should be located among the source files. [Xcode](https://stackoverflow.com/a/8016333) refers to
+  these as "Project Headers".
+
+* *Private Headers* - Headers that are available to other libraries in the repo, but are not part
+  of the public API. These should be located in `FirebaseFoo/Sources/Private`.
+  [Xcode](https://stackoverflow.com/a/8016333) and CocoaPods refer to these as "Private Headers".
+  Note that we are deprecating the usage of CocoaPods `private_headers` and should instead
```
wilhuff, line 28: This note should be elsewhere in the section about how to import headers from another library in the repo.

HeadersImports.md
```diff
@@ -0,0 +1,68 @@
...
+  publish them with `preserve_paths` and access them with a repo-relative import.
+
+## Imports
+
+* *Headers within the Library* - Use a repo-relative path. The one exception is
```
wilhuff, line 33: You spend a bunch of time above breaking down the taxonomy of headers but then don't actually use those terms here. It's confusing. You called this "Library Internal Headers" above; use that term here?

Also, pull out the bit about public headers into its own bullet to match the structure above and to make it easier to skim.

HeadersImports.md
```diff
@@ -0,0 +1,68 @@
...
+  that public header imports from other public headers should do an unqualified
+  import like `import "publicHeader.h"` to avoid colliding with the public module import.
+
+* *Headers within the Repo* - Import an umbrella header like
```
wilhuff, line 37: "Headers within the Repo" isn't mentioned above. I think you mean "private headers from another library in the repo".

Also, nowhere else do you describe what a private umbrella header is, nor how to write one. A section about how to share internal APIs with other Firebase libraries would be helpful.

This paragraph should also mention that you need to add references to the private headers you're using to your CocoaPods `preserve_paths`. This would be a good place for your note about deprecating `private_headers`.

HeadersImports.md
paulb777: I clarified and moved the `Firebase.h` info to `AddNewPod.md`.

HeadersImports.md
paulb777: Done.

HeadersImports.md
paulb777: Flushed out.

HeadersImports.md
paulb777: I was originally going to organize like that and then realized that thinking about creating and managing headers is much different and more complicated than being a consumer of them. I tried to make it more clear the document is organized like that.

HeadersImports.md
paulb777: Clarified

HeadersImports.md
```diff
@@ -8,35 +8,44 @@ guidelines are designed to support a wide range of build systems and usage scena
...
-  that public header imports from other public headers should do an unqualified
-  import like `import "publicHeader.h"` to avoid colliding with the public module import.
+## Imports - For Header File Consumers
 
-* *Headers within the Repo* - Import an umbrella header like
-  `FirebaseCore/Sources/Private/FirebaseCoreInternal.h`. Any package manager
-  complexity should be localized to the internal umbrella.
+* *Headers within the Library* - Use a repo-relative path for all of the header types above.
+  * *Exception* - Public header imports from other public headers should do an unqualified
+  import like `import "publicHeader.h"` to avoid public module collisions.
```
wilhuff, line 41: Since there's both `#import` and `@import`, you should make clear which you mean here.

PR #5768, opened 2020-06-08
Introduce new enum for Data Collection Controls.

ryanwilson:
This deprecates the old boolean flag and allows developers to reset back
to defaults if needed. This will also be used in other SDKs to allow for
migration from the old always-persistent flag.

Testing wise: this adds tests to make sure the existing underlying
storage is honoured and duplicates previous tests with the new flag. The
old tests can be removed once the deprecated API is deleted but it's
better to have overlapping and repeated tests than to miss something in
this case.


ryanwilson:
An update so folks know the status: there's a bit of confusion about what should be returned when reading the flag. Do we fall back to the Plist value and return "Disabled" if it's disabled there, even though the runtime flag hasn't been set? Working with the Android folks to figure out what their implementation does.

ryanwilson:
Closing for now until the discussion is resolved.

PR #5773, opened 2020-06-09
Refactor event saving and searching

mikehaney24:
No more symlinks to try and speed searching. Just get a directory's contents and check the filename, which will contain metadata about the event.

#no-changelog because it's parter of a larger change

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -359,7 +359,7 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
                                               mappingIDs:nil
                                                 qosTiers:nil];
   id<GDTCORStorageEventIterator> iter = [self iteratorWithSelector:eventSelector];
-  return [iter nextEvent];
+  return [iter nextEvent] != nil;
 }
 
 - (nullable id<GDTCORStorageEventIterator>)iteratorWithSelector:
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -102,21 +102,67 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
   return libraryDataPath;
 }
 
-+ (NSDictionary<NSString *, NSString *> *)pathsForEvent:(GDTCOREvent *)event {
-  NSString *dataPath =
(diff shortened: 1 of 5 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -102,74 +102,69 @@ + (NSString *)libraryDataStoragePath {
...
+  NSFileManager *fileManager = [NSFileManager defaultManager];
+  NSString *targetPath = [NSString
+      stringWithFormat:@"%@/%ld", [GDTCORFlatFileStorage baseEventStoragePath], (long)target];
+  NSError *error;
+  NSArray<NSString *> *dirPaths = [fileManager contentsOfDirectoryAtPath:targetPath error:&error];
+  if (error) {
+    GDTCORLogDebug(@"There was an error reading the contents of the target path: %@", error);
+    return paths;
+  }
+  BOOL checkingIDs = eventIDs != nil && eventIDs.count > 0;
```
maksymmalyhin, line 128: nit: Taking into account Objective C `nil` behaviour  `eventIDs != nil` looks redundant, because `nil` casted to an integer as `0`.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -102,74 +102,69 @@ + (NSString *)libraryDataStoragePath {
...
-    BOOL isDirectory;
-    if ([[NSFileManager defaultManager] fileExistsAtPath:searchPath isDirectory:&isDirectory]) {
-      if (isDirectory) {
-        [searchPaths addObject:searchPath];
-      }
+    NSString *filename = [path lastPathComponent];
+    NSArray<NSString *> *components = [filename componentsSeparatedByString:@"."];
+    if (components.count != 3) {
+      GDTCORLogDebug(@"There was an error reading the filename components: %@", components);
+      return paths;
```
maksymmalyhin, line 141: I wonder why we are skipping other files if one of the files has an incorrect name?

mikehaney24:
> How are you measuring performance?

Right now, I'm not. The whole library is async and I'm currently more concerned with robustness than performance.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: doh, yeah, good catch

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Good point--perhaps we can just delete those files and continue.

PR #5767, opened 2020-06-08
Add support for CCT EventCode field

JJLionHeart:
Issue: https://github.com/firebase/firebase-ios-sdk/issues/5766

PR #5780, opened 2020-06-09
Update GDTCCT CHANGELOG and version

mikehaney24:


GoogleDataTransportCCTSupport/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v3.2.0
+- Expose a new public API to collect network or event code data.
```
paulb777, line 2: What are the clients of the new public API?  I'm worried about adding new API surface before sorting out the circular dependencies between GDT and GDTCCT.

mikehaney24: An internal team needs this functionality and implemented this feature on their own in #5767. Also, the deps aren't circular. Adopters need GDTCCT and GDTCCT needs GDT. Adopters -> GDTCCT -> GDT

paulb777: But the public API surface is in GDT so the header dependency tree doesn't match the linker dependency tree.  I'll get back to Swift PM in the next week or so and flush this out in more detail and hopefully find what I'm missing.

PR #5782, opened 2020-06-09
update changelog for messaging and iid

chliangGoogle:


Firebase/InstanceID/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2020-06 -- 4.4.0
+# [changed] Standardize FirebaseCore import headers. (#5758)
```
paulb777, line 2: I'll leave to @morganchen12  about how best to centralize or de-centralize this for public consumption, but it's fine to stay in CHANGELOG even if it doesn't go out in the release notes.

chliangGoogle: it's the only change for iid so we need an entry here but i'm open for other wording.

morganchen12: Given the thorough explanation in the core changelog, I'm ok with making this as succinct as possible. This should be a list item, though, not a header. (Replace `#` with `-`)

chliangGoogle: Done.

PR #5788, opened 2020-06-10
Add Installations framework testing.

granluo:
This will add multiple framework testing.

### DIFF
--- a/.github/workflows/installations.yml
+++ b/.github/workflows/installations.yml
@@ -53,6 +53,7 @@ jobs:
   #   - name: Setup project and Build for Catalyst
   #     # Only build the unit tests on Catalyst. Test stopped working when GHA moved to Xcode 11.4.1.
   #     run: scripts/test_catalyst.sh FirebaseInstallations build FirebaseInstallations-Unit-unit
+
   quickstart_framework:
     env:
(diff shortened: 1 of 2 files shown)


granluo:
The framework test is passed in commit 3c87822

.github/workflows/installations.yml
```diff
@@ -10,8 +10,9 @@ on:
     - '.github/workflows/installations.yml'
     - 'Gemfile'
   schedule:
-    # Run every day at 11pm (PST) - cron uses UTC times
-    - cron:  '0 7 * * *'
+    # Run every day at 3am (PST) - cron uses UTC times
+    # This is set to 3 hours after zip workflow so zip testing can run after.
```
paulb777, line 14: This is set to 3 hours after zip workflow finishes so zip testing can run after.

scripts/setup_quickstart_framework.sh
```diff
@@ -12,8 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SAMPLE="$1"
+set -e
```
paulb777, line 15: Perhaps, one line - `set -ex`

PR #5790, opened 2020-06-10
Store events using the flat file storage system

mikehaney24:
Resolves many TODOs in the flat file storage system
Bug fixes found by unit tests are also included.

Doesn't remove any code that has now been made duplicate.

#no-changelog because this is a feature branch

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -111,62 +111,69 @@ NSString *const gGDTCORFlatFileStorageQoSTierPathKey = @"QoSTierPath";
                                  (long)target, eventID, qosTier, mappingID];
 }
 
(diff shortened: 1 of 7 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -227,7 +245,7 @@ - (void)storeEvent:(GDTCOREvent *)event
     // Write the transport bytes to disk, get a filename.
     GDTCORAssert([event.dataObject transportBytes],
                  @"The event should have been serialized to bytes");
-    NSError *error = nil;
+    error = nil;
```
paulb777, line 248: The nil initialization is only necessary for clarity - but if done, it should be at the declaration above - https://stackoverflow.com/a/10022295/556617

mikehaney24: This is to clear the error object from above if it's non-nil for some reason.

PR #5791, opened 2020-06-11
Refactor workflow to be concise

granluo:
This is to simplify the workflow to make the workflow more readable.


scripts/remove_data.sh
```diff
@@ -14,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-curl https://sdk.cloud.google.com > install.sh
-bash install.sh --disable-prompts
-echo "::add-path::${HOME}/google-cloud-sdk/bin/"
+set -x
+
+SDK="$1"
+rm -f quickstart-ios/"${SDK}"/GoogleSerivce-info.plist
```
paulb777, line 18: typo

PR #5706, opened 2020-05-30
SwiftPM Prototype for Storage and Auth

paulb777:
Add this branch via `Xcode -> File -> Swift Packages -> AddPackage Dependencies ...` to the Xcode project https://github.com/firebase/quickstart-ios/tree/master/storage and see the ObjC and Swift quickstarts run

PR #5795, opened 2020-06-11
Squash spm master2020

paulb777:
Test only PR to test a squashed rebase of the spm-master2020 branch before updating it.


paulb777:
Closing now that #5706 is now functional with the tested rebase.

PR #5796, opened 2020-06-11
Add ABTesting framework testing.

granluo:
Add ABTesting framework testing.


granluo:
The framerwork test is in commit b4f0391.

PR #5798, opened 2020-06-12
Add Crashlytics framework testing.

granluo:
Add Crashlytics Framework testing.


granluo:
Test is done in 89d1dbd

granluo:
> Looks like the schedule needs to be updated ...

Good catch, just updated.

PR #5799, opened 2020-06-12
Add Database framework testing.

granluo:
Add Database framework testing.


granluo:
Test is done in 33c0c26.

PR #5804, opened 2020-06-12
Remove Crashlytics OTHER_LD_FLAGS for Swift Package Manager

samedson:
This linker flag creates a section in the binary containing the Info.plist.

 - Swift Package Manager can't set `OTHER_LD_FLAGS` as of now
 - This may be unnecessary

Took a look at the old codebase - I believe it was a mistake to add this to the cocoapod in the first place. The use case for this rule is creating a section in the binary when no Info.plist is present, but there will always be an Info.plist in the bundle for iOS / macOS / tvOS apps.

This rule is used for upload-symbols so that it can read its own version. Since upload-symbols is binary only, there's no Info.plist, so this linker rule is necessary.

We must have just accidentally applied the linker rule to the SDK too.

### DIFF
--- a/Crashlytics/CHANGELOG.md
+++ b/Crashlytics/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- [changed] Removed an unnecessary linker rule for embedding the Info.plist (#5804)
+
 # v4.1.1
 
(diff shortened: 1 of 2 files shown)


samedson:
Took me a while to get to this - I'm on-call this week.

Just running through some manual tests.

samedson:
Ooh lemme just do that real quick.

PR #5810, opened 2020-06-12
GoogleUtilities repo-relative imports

paulb777:
Output of the scripts/change_headers.swift

The GoogleUtilities clients are updated in the subsequent #5824 

scripts/change_headers.swift will be updated to a fully commented import verification step later in this series of PRs.

PR #5824, opened 2020-06-15
Update GoogleUtilities' clients with repo-relative imports

paulb777:
- Update client imports to GoogleUtilities to use repo-relative imports - following up #5810 
  - The clients were updated with the next incarnation of `scripts/change_headers.swift`
- Initial Swift Package Manager CI for GoogleUtilities and FirebaseCore - (productizing prototype code in the spm-master2020 branch.
- Bump GoogleUtilities minor version and its clients podspec dependency
- Updated GoogleUtilities header structure to minimize deltas between CocoaPods and SPM
  - Where possible, treated the CocoaPods `Private` folder as public.
  - When not possible because of intra-pod imports in the Private headers, created a Public/dummy.h to treat the Privates as Private.
  - The GoogleUtilities header organization should be revisited and made more consistent when breaking changes are allowed with Firebase 7. There is not a consistent definition between public and private header usage currently.


GoogleUtilities/Environment/SecureStorage/GULKeychainStorage.m
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#import "GoogleUtilities/Environment/Public/GULKeychainStorage.h"
+#import "GoogleUtilities/Environment/Private/GULKeychainStorage.h"
```
maksymmalyhin, line 17: I wonder if there is a way to implement test apps like #5828 with this structure.

scripts/change_headers.swift
```diff
@@ -107,16 +113,22 @@ let headerMap = getHeaderMap(url)
 for root in changeImports {
   let rootURL = url.appendingPathComponent(root)
   let enumerator = FileManager.default.enumerator(atPath: rootURL.path)
-  while let file = enumerator?.nextObject() as? String {
+  whileLoop: while let file = enumerator?.nextObject() as? String {
```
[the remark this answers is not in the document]
paulb777: The label enables the continue from a doubly nested loop.  See line 128.

GoogleUtilities/Environment/SecureStorage/GULKeychainStorage.m
paulb777: I don't think that test apps should depend on the repo structure. My intent is to only change the library sources and unit test imports.

maksymmalyhin: The issue (existing one already) that there are no actual public header in GoogleUtilities/Environment (they are marked as private now). I was going to make them public. Do you think it may be an issue?

paulb777: Yes. We should avoid public headers since they require ifdef'ing between Swift Package Manager and CocoaPods.  See the Logger and NSData changes in this PR for details.

maksymmalyhin: It feels a bit weird to have a library without public API, but it's probably out of scope of the PR, we may discuss it later

paulb777:
cc: @eldhosembabu @chliangGoogle @karenyz @christibbs

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -22,10 +22,10 @@
 #import <FirebaseInstanceID/FirebaseInstanceID.h>
 #import <FirebaseMessaging/FIRMessaging.h>
 #import <FirebaseMessaging/FIRMessagingExtensionHelper.h>
-#import <GoogleUtilities/GULAppDelegateSwizzler.h>
```
mikehardy, line 25: Hey @paulb777 :-), turns out react-native-firebase was using the GULAppDelegateSwizzler:

https://github.com/invertase/react-native-firebase/blob/67c52a8aaeac32d2253e73769313e49700bc497c/packages/messaging/ios/RNFBMessaging/RNFBMessaging%2BAppDelegate.m#L20

And with 6.7.0 of GoogleUtilities / 6.28.0 of firebase-ios-sdk we get a build error https://github.com/invertase/react-native-firebase/issues/3938 as GULAppDelegateSwizzler.h tries to pull in GULApplication.h but it's not found:

> ios/Pods/Headers/Private/GoogleUtilities/GULAppDelegateSwizzler.h:19:9: 'GoogleUtilities/AppDelegateSwizzler/Private/GULApplication.h' file not found

I'm ace at the Java side of the react-native house but my Objective-C skills sadly leave something to be desired.

I attempted to alter the import in our repo to match the change here (` #import "GoogleUtilities/AppDelegateSwizzler/Private/GULAppDelegateSwizzler.h"`) without success

Is there a simple trick I'm missing that will resolve the headers and allow our auto-swizzling for react-native library consumers to work, or is this a more serious structural issue?

Sorry to bother, but thanks in advance for any help

alejobrainz: Turns out that the Facebook SDK for Unity was also using GULAppDelegateSwizzler. :/. We temporarily fixed the issue by adding this entry to the Podfile on our project:

`pod 'GoogleUtilities', '6.6.0'`

Hope this is fixed soon.

mikehardy: At least our repo isn't alone @alejobrainz :-), we have a user report confirming that pinning the GoogleUtilities dep at 6.6.0 works for us as well, at least for firebase-ios-sdk pods 6.27.1, still building against 6.28.0 but with GoogleUtilities pinned on 6.6.0 to see

ryanwilson: Sorry for the issue - I see it's been reported at #6047 so we'll make all progress updates there and continue the discussion for greater visibility.

PR #5827, opened 2020-06-15
Add Storage framework testing.

granluo:
Add Storage framework testing.


granluo:
Objc test is passed in f0132ec

PR #5829, opened 2020-06-15
Get rid of event iterators and add separate folders for batching

mikehaney24:
This resolves the issue of events being moved or deleted while another operation is in progress. When an event is batched, a batchID is created and all the events in that batch are moved to that folder.

#no-changelog because the feature branch itself will have the changelog


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
+                                  withIntermediateDirectories:YES
+                                                   attributes:nil
+                                                        error:nil];
+        NSString *destinationPath = [batchPath stringByAppendingPathComponent:fileName];
+        error = nil;
+        [[NSFileManager defaultManager] moveItemAtPath:eventPath
+                                                toPath:destinationPath
+                                                 error:&error];
+        if (error) {
+          GDTCORLogDebug(@"An event file wasn't not moveable into the batch directory: %@", error);
```
paulb777, line 402: was moveable

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
+    if (error) {
+      if (onComplete) {
+        onComplete();
+      }
+      return;
+    }
+    for (NSString *path in batches) {
+      NSArray<NSString *> *components = [[path lastPathComponent] componentsSeparatedByString:@"."];
+      if (components.count != 2) {
+        GDTCORLogDebug(@"The name of the batch folder couldn't be parsed: %@", path);
```
paulb777, line 432: break or return?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
   dispatch_async(_storageQueue, ^{
     NSString *dataPath = [[[self class] libraryDataStoragePath] stringByAppendingPathComponent:key];
     NSError *error;
     NSData *data = [NSData dataWithContentsOfFile:dataPath options:0 error:&error];
     if (onComplete) {
-      onComplete(data, error);
+      NSData *newValue = onComplete(data, error);
+      // The -isKindOfClass check is necessary because without an explicit 'return nil' in the block
+      // the implicit return value will be the block itself. The compiler doesn't detect this.
+      if (newValue && [newValue isKindOfClass:[NSData class]] && newValue.length) {
```
paulb777, line 468: What if it's not an NSData?

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -168,7 +168,7 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
     resultData = [NSKeyedArchiver archivedDataWithRootObject:obj
                                        requiringSecureCoding:YES
                                                        error:error];
-    if (*error) {
+    if (resultData == nil && *error) {
```
paulb777, line 171: Isn't resultData == nil **OR** *error an encoding failure?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: I'm not sure what you mean. If it's not an NSData then we do nothing. It shouldn't be an error since it's allowable from the compiler's perspective.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
+        continue;
+      } else {
+        NSString *fileName = [eventPath lastPathComponent];
+        NSString *batchPath = [GDTCORFlatFileStorage batchPathForTarget:eventSelector.selectedTarget
+                                                                batchID:batchID];
+        [[NSFileManager defaultManager] createDirectoryAtPath:batchPath
+                                  withIntermediateDirectories:YES
+                                                   attributes:nil
+                                                        error:nil];
+        NSString *destinationPath = [batchPath stringByAppendingPathComponent:fileName];
```
maksymmalyhin, line 396: Some comments describing where files are stored will be really helpful.

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
```diff
@@ -863,6 +830,167 @@ - (void)testStorageSizeWithCallback {
...
+  XCTAssertNotNil(batchID);
+  XCTAssertEqualObjects(batchID, expectedBatchID);
+
+  expectedBatchID = @1;
+  batchID = [[GDTCORFlatFileStorage sharedInstance] nextBatchID];
+  XCTAssertNotNil(batchID);
+  XCTAssertEqualObjects(batchID, expectedBatchID);
+
+  for (int i = 0; i < 1000; i++) {
+    NSNumber *batchID = [[GDTCORFlatFileStorage sharedInstance] nextBatchID];
```
maksymmalyhin, line 848: It may be useful to have a test for concurrent calling of `nextBatchID` to assert its thread safety.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -475,6 +602,27 @@ - (void)addEventToTrackingCollections:(GDTCOREvent *)event {
...
+                   [data getBytes:(void *)&lastBatchID length:sizeof(int32_t)];
+                 }
+                 if (data == nil) {
+                   lastBatchID = 0;
+                 }
+                 dispatch_semaphore_signal(sema);
+                 int32_t incrementedValue = lastBatchID + 1;
+                 return [NSData dataWithBytes:&incrementedValue length:sizeof(int32_t)];
+               }];
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 10 * NSEC_PER_SEC)) != 0) {
```
maksymmalyhin, line 620: IMO it would be much safer to have async method semantic and avoid blocking the caller even for 10 sec. All callers use completion to return the result so it is not a huge issue

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
paulb777: Isn't it controllable what gets returned from `onComplete`?   What else could be returned besides an `NSData`?

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
mikehaney24: they should coincide, but the real problem was an occasional deference of a NULL pointer. Really, I should be `resultData == nil || (checking error != NULL && *error != nil)`

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: The completion handler has a specific prototype (returning `nullable NSData *`), but the compiler isn't enforcing it.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
   });
 }
 
-#pragma mark - GDTCORStorageProtocol
+- (nullable NSNumber *)
+    batchWithEventSelector:(nonnull GDTCORStorageEventSelector *)eventSelector
+           batchExpiration:(nonnull GDTCORClock *)expiration
```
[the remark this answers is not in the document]
mikehaney24: It's going to be addressed in the next PR.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -323,17 +350,128 @@ - (void)removeEvents:(NSSet<NSNumber *> *)eventIDs {
...
+  __block NSSet<NSString *> *eventPaths;
+  [self pathsForTarget:eventSelector.selectedTarget
+              eventIDs:eventSelector.selectedEventIDs
+              qosTiers:eventSelector.selectedQosTiers
+            mappingIDs:eventSelector.selectedMappingIDs
+            onComplete:^(NSSet<NSString *> *_Nonnull paths) {
+              eventPaths = paths;
+              dispatch_semaphore_signal(sema);
+            }];
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 10 * NSEC_PER_SEC)) != 0) {
```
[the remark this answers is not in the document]
mikehaney24: The call to pathsForTarget::::: also happens on the queue, so it needs to be outside the dispatch_async block below. Without the semaphore, the current thread will block anyway. The semaphore provides a timeout mechanism if there are too many operations queued on on the queue. 10 seconds just seemed like a reasonable number. I'm open to alternatives.

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
mikehaney24: I think calling it 1000 times almost as fast as possible is sufficient (It regularly failed when it wasn't thread-safe during development).

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: I'll make this change, but it's going to lead to onComplete hell.

maksymmalyhin: yeah, onComplete hell is annoying. We use a Google promises lib in FIS, Crashlytics (and couple more as far as I remember) to deal with it. It's pretty solid and small, maybe it's worth considering it as a dependency to GDT as well?

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
maksymmalyhin: The current test does the calls one by one and will not fail if we remove semaphore from `nextBatchID` implementation. I was thinking of using e.g. `dispatch_async` on a concurrent queue e.g. like [here](https://github.com/firebase/firebase-ios-sdk/blob/5f370d6395c51a8616157d7b2cbe8a5737e8647e/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsSingleOperationPromiseCacheTests.m#L51)

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Sounds like a good first project for you ;-)

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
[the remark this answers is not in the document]
maksymmalyhin: Since `dispatch_async` calls can be safely nested even for a serial queue, I think, executing the remaining work can be done from the completion. We can try to deduce the callback hell if we split the implementation into separate methods. Though this approach won't provide the timeout feature if it is a requirement.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Done

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Fixed.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORStorageProtocol.h
```diff
@@ -72,13 +82,15 @@ NS_ASSUME_NONNULL_BEGIN
...
+/** Retrieves the stored data for the given key and optionally sets a new value.
  *
  * @param key The key corresponding to the desired data.
- * @param onComplete The callback to invoke with the data once it's retrieved.
+ * @param onComplete The callback to invoke with the data once it's retrieved. If NSData is returned
+ * by the completion block, the library data will be set with this new value.
  */
 - (void)libraryDataForKey:(NSString *)key
-               onComplete:(void (^)(NSData *_Nullable data, NSError *_Nullable error))onComplete;
+               onComplete:(nullable NSData * (^)(NSData *_Nullable data,
```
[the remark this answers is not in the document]
mikehaney24: Done.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
mikehaney24: Added a test.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Made async.

PR #5835, opened 2020-06-16
remove ios 7 logic as sdk minimum support is iOS 8 now

chliangGoogle:


Firebase/InstanceID/FIRInstanceIDURLQueryItem.h
```diff
@@ -18,7 +18,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// Stand-in for NSURLQueryItem, which is only available on iOS 8.0 and up.
 @interface FIRInstanceIDURLQueryItem : NSObject
```
paulb777, line 21: Could the clients use NSURLQueryItem instead and delete the whole class?

chliangGoogle: Good idea! Didn't realize that.

FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [changed] Remove logic that is executed for iOS 7 and below. (5835)
```
paulb777, line 2: Missing '#'

PR #5843, opened 2020-06-17
Initial check imports script

paulb777:
Verify functional and style of imports and includes.

### DIFF
--- a/scripts/check_imports.swift
+++ b/scripts/check_imports.swift
@@ -15,18 +15,10 @@
  * limitations under the License.
  */
 
-// Utility script for updating to repo-relative headers.
+// Utility script for verifying `import` and `include` syntax. This ensures a
+// consistent style as well as functionality across multiple package managers.
 
 import Foundation
-
-// Update with directories in which to find headers.
-let findHeaders = ["FirebaseInstallations"]
-
-// Update with directories in which to change imports.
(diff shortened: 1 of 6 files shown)


scripts/check_imports.swift
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/swift
```
wilhuff, line 1: Writing this kind of check in Swift means these checks can't run on Windows. While most of Firebase doesn't support Windows from this repo, Firestore does. Please reconsider this.

scripts/check_imports.swift
```diff
@@ -0,0 +1,177 @@
...
+/*
+ * Copyright 2020 LLC Google
```
wilhuff, line 3: Google LLC

scripts/check_imports.swift
paulb777: Swift looks to be coming soon to Windows - next week perhaps?  https://swift.org/blog/5-3-release-process/.

scripts/check_imports.swift
paulb777: Thanks

paulb777:
Maybe?  

I'm prioritizing more integration testing first in expanding the script's coverage across the repo first.

PR #5849, opened 2020-06-18
Don't run GHA on private repo until it is for a PR

maksymmalyhin:


.github/workflows/abtesting.yml
```diff
@@ -17,6 +17,9 @@ on:
 
 jobs:
   pod-lib-lint:
+    # Don't run on private repo unless it is a PR.
```
paulb777, line 20: Also need to delete the on: push: above - here and other .ymls that have it.

maksymmalyhin: Do you mean we would like to get rid of post-submit checks even in public repo?

paulb777: Yes, they're inconsistently used and not a lot of value.

maksymmalyhin: done

maksymmalyhin:
> Did you cancel all the GHA runs for the last commit? If so, did you find a way other than individually cancelling them in the console?

I did and it took some time actually. I found using [this view](https://github.com/firebase/firebase-ios-sdk/actions) the fastest way, but it is far from ideal.

PR #5851, opened 2020-06-18
Storage: Repo-relative headers and SPM

paulb777:
#no-changelog - Will be part of a global M74 changelog


scripts/change_headers.swift
```diff
@@ -20,13 +20,13 @@
 import Foundation
```
paulb777, line 20: This is a short-lived script not intended to stay in the repo long - but checked in to associate with the repo-relative changes in the PR.

PR #5852, opened 2020-06-18
Fix buildDependencies false and subspecs

paulb777:
pods specified in the json that included subspec were ignored when the buildDependencies option was false.

#no-changelog


paulb777:
This code runs after the `pod install` so it will only build the parts of FirebaseUI requested.  The problem was that the built pod is named `FirebaseUI` no matter what subspecs were specified.  I did a build with `FirebaseUI/Database` to confirm.

PR #5853, opened 2020-06-19
Manually created dynamic links should be subject to allowed/blocked check

palexsua:
- **Description**: Dynamic links can be validated against a block/allow list, which is defined server side in the Firebase Console. This verification should apply to both short and long dynamic links, and be supported by both iOS and Android SDKs.
- **What’s expected**: It’s expected that both iOS and Android will verify both kinds of dynamic link, and if it passes the verification rules defined server side, allow the link to be processed.
- **What’s observed**: On iOS, long dynamic links are not verified, and it’s possible for a link that doesn’t pass the "allowed link" rules to be treated as verified. Short dynamic links are verified, that’s not an issue. This also isn’t an issue on Android, which verifies both short and long dynamic links.
- **Proposed solution**: Ensure the error response from `reopenAttribution` (which contains information to whether the link was verified or not), should be propagated further up for long dynamic links, so the appropriate decision on whether it’s a verified link can be made. Some method definitions have changed as well to accommodate this, which have now made them asynchronous.

According to the source code all short links are checked when a call to `FIRDynamicLinkNetworking` class method `resolveShortLink:completion:` takes place. This method does a HTTP call to `reopenAttribution` endpoint and receives a response with an error, if the URL is not allowed. The problem is that the same is not true for long dynamic links (aka [manually created links](https://firebase.google.com/docs/dynamic-links/create-manually)) - no call to `resolveShortLink:completion:` method (or its long links counterpart) is done. Thus all URLs are accepted and considered allowed.
Another part of the problem is that when a call to `reopenAttribution` endpoint is done the response with the error description is ignored and and error equal to `nil` is passed to completion callback. Thus, all traces of the issue (the URL being not allowed) are lost.

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
@@ -41,6 +41,7 @@ static NSString *const kFDLAnalyticsDataSourceKey = @"utmSource";
 static NSString *const kFDLAnalyticsDataMediumKey = @"utmMedium";
 static NSString *const kFDLAnalyticsDataCampaignKey = @"utmCampaign";
 static NSString *const kHeaderIosBundleIdentifier = @"X-Ios-Bundle-Identifier";
+static NSString *const kGenericErrorDomain = @"com.firebase.dynamicLinks";
 
 typedef NSDictionary *_Nullable (^FIRDLNetworkingParserBlock)(
     NSString *requestURLString,
@@ -67,7 +68,7 @@ void FIRMakeHTTPRequest(NSURLRequest *request, FIRNetworkRequestCompletionHandle
       [session dataTaskWithRequest:request
                  completionHandler:^(NSData *_Nullable data, NSURLResponse *_Nullable response,
                                      NSError *_Nullable error) {
-                   completion(data, error);
+                   completion(data, response, error);
                  }];
   [dataTask resume];
 }
@@ -108,33 +109,66 @@ NSData *_Nullable FIRDataWithDictionary(NSDictionary *dictionary, NSError **_Nul
     @"sdk_version" : FDLSDKVersion
   };
 
-  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSError *error) {
-    NSURL *resolvedURL;
+  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSURLResponse *response,
+                                                             NSError *error) {
+    NSURL *resolvedURL = nil;
+    NSHTTPURLResponse *resp = nil;
 
-    if (!error && data) {
-      NSDictionary *result = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
-      if ([result isKindOfClass:[NSDictionary class]]) {
-        id invitationIDObject = [result objectForKey:@"invitationId"];
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+      NSError *err = [NSError
+          errorWithDomain:kGenericErrorDomain
+                     code:0
+                 userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+      handler(resolvedURL, error ? error : err);
+    } else {
+      resp = (NSHTTPURLResponse *)response;
+      if (resp.statusCode >= 200 && resp.statusCode < 300) {
+        if (!error && data) {
+          NSDictionary *result = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
+          if ([result isKindOfClass:[NSDictionary class]]) {
+            id invitationIDObject = [result objectForKey:@"invitationId"];
 
-        NSString *invitationIDString;
-        if ([invitationIDObject isKindOfClass:[NSDictionary class]]) {
-          NSDictionary *invitationIDDictionary = invitationIDObject;
-          invitationIDString = invitationIDDictionary[@"id"];
-        } else if ([invitationIDObject isKindOfClass:[NSString class]]) {
-          invitationIDString = invitationIDObject;
+            NSString *invitationIDString;
+            if ([invitationIDObject isKindOfClass:[NSDictionary class]]) {
+              NSDictionary *invitationIDDictionary = invitationIDObject;
+              invitationIDString = invitationIDDictionary[@"id"];
+            } else if ([invitationIDObject isKindOfClass:[NSString class]]) {
+              invitationIDString = invitationIDObject;
+            }
+
+            NSString *deepLinkString = result[kFDLResolvedLinkDeepLinkURLKey];
+            NSString *minAppVersion = result[kFDLResolvedLinkMinAppVersionKey];
+            NSString *utmSource = result[kFDLAnalyticsDataSourceKey];
+            NSString *utmMedium = result[kFDLAnalyticsDataMediumKey];
+            NSString *utmCampaign = result[kFDLAnalyticsDataCampaignKey];
+            resolvedURL = FIRDLDeepLinkURLWithInviteID(invitationIDString, deepLinkString,
+                                                       utmSource, utmMedium, utmCampaign, NO, nil,
+                                                       minAppVersion, self->_URLScheme, nil);
+          }
         }
-
-        NSString *deepLinkString = result[kFDLResolvedLinkDeepLinkURLKey];
-        NSString *minAppVersion = result[kFDLResolvedLinkMinAppVersionKey];
-        NSString *utmSource = result[kFDLAnalyticsDataSourceKey];
-        NSString *utmMedium = result[kFDLAnalyticsDataMediumKey];
-        NSString *utmCampaign = result[kFDLAnalyticsDataCampaignKey];
-        resolvedURL = FIRDLDeepLinkURLWithInviteID(invitationIDString, deepLinkString, utmSource,
-                                                   utmMedium, utmCampaign, NO, nil, minAppVersion,
-                                                   self->_URLScheme, nil);
+        handler(resolvedURL, error);
+      } else if (error) {
+        handler(resolvedURL, error);
+      } else {
+        NSDictionary *result = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
+        if ([result isKindOfClass:[NSDictionary class]] && [result objectForKey:@"error"]) {
+          id err = [result objectForKey:@"error"];
+          NSError *customError = [NSError errorWithDomain:kGenericErrorDomain
+                                                     code:resp.statusCode
+                                                 userInfo:err];
+          handler(resolvedURL, customError);
+        } else {
+          NSError *err = [NSError
+              errorWithDomain:kGenericErrorDomain
+                         code:0
+                     userInfo:@{
+                       @"message" : [NSString
+                           stringWithFormat:@"Failed to resolve link: %@", url.absoluteString]
+                     }];
+          handler(resolvedURL, err);
+        }
       }
     }
-    handler(resolvedURL, error);
   };
 
   NSString *requestURLString =
@@ -242,13 +276,14 @@ NSData *_Nullable FIRDataWithDictionary(NSDictionary *dictionary, NSError **_Nul
     }
   };
 
-  FIRNetworkRequestCompletionHandler convertInvitationCallback = ^(NSData *data, NSError *error) {
-    if (handler) {
-      dispatch_async(dispatch_get_main_queue(), ^{
-        handler(error);
-      });
-    }
-  };
+  FIRNetworkRequestCompletionHandler convertInvitationCallback =
+      ^(NSData *data, NSURLResponse *response, NSError *error) {
+        if (handler) {
+          dispatch_async(dispatch_get_main_queue(), ^{
+            handler(error);
+          });
+        }
+      };
 
   NSString *requestURL = [NSString stringWithFormat:@"%@/convertInvitation%@", kApiaryRestBaseUrl,
                                                     FIRDynamicLinkAPIKeyParameter(_APIKey)];
@@ -269,13 +304,37 @@ NSData *_Nullable FIRDataWithDictionary(NSDictionary *dictionary, NSError **_Nul
   NSString *requestURLString = [NSString
       stringWithFormat:@"%@/%@%@", baseURL, endpointPath, FIRDynamicLinkAPIKeyParameter(_APIKey)];
 
-  FIRNetworkRequestCompletionHandler completeInvitationByDeviceCallback =
-      ^(NSData *data, NSError *error) {
-        if (error || !data) {
-          dispatch_async(dispatch_get_main_queue(), ^{
-            handler(nil, nil, error);
-          });
-          return;
+  FIRNetworkRequestCompletionHandler completeInvitationByDeviceCallback = ^(
+      NSData *data, NSURLResponse *response, NSError *error) {
+    if (error || !data) {
+      dispatch_async(dispatch_get_main_queue(), ^{
+        handler(nil, nil, error);
+      });
+      return;
+    }
+    NSString *matchMessage = nil;
+    NSError *parsingError = nil;
+    NSDictionary *parsedDynamicLinkParameters =
+        parserBlock(requestURLString, data, &matchMessage, &parsingError);
+
+    // If request was made with pasteboard contents, verify if we got a unique match. If we got
+    // a "none" match, we were unable to get a unique match or deduce using fingerprinting.
+    // In this case, resend requests to IPV4 and IPV6 endpoints for fingerprinting. b/79704203
+    if (requestBody[@"uniqueMatchLinkToCheck"] && parsedDynamicLinkParameters &&
+        (!parsedDynamicLinkParameters[kFIRDLParameterMatchType] ||
+         [parsedDynamicLinkParameters[kFIRDLParameterMatchType] isEqualToString:@"none"])) {
+      NSMutableDictionary *requestBodyMutable = [requestBody mutableCopy];
+      [requestBodyMutable removeObjectForKey:@"uniqueMatchLinkToCheck"];
+      NSMutableArray *baseURLs =
+          [@[ kIosPostInstallAttributionRestBaseUrlIPV4, kIosPostInstallAttributionRestBaseUrlIPV6 ]
+              mutableCopy];
+      if (parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]) {
+        if ([parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]
+                isEqualToString:@"IP_V4"]) {
+          [baseURLs removeObject:kIosPostInstallAttributionRestBaseUrlIPV4];
+        } else if ([parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]
+                       isEqualToString:@"IP_V6"]) {
+          [baseURLs removeObject:kIosPostInstallAttributionRestBaseUrlIPV6];
         }
 
         NSString *matchMessage = nil;
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
@@ -311,48 +311,24 @@ NSData *_Nullable FIRDataWithDictionary(NSDictionary *dictionary, NSError **_Nul
   NSString *requestURLString = [NSString
       stringWithFormat:@"%@/%@%@", baseURL, endpointPath, FIRDynamicLinkAPIKeyParameter(_APIKey)];
 
-  FIRNetworkRequestCompletionHandler completeInvitationByDeviceCallback = ^(
-      NSData *data, NSURLResponse *response, NSError *error) {
-    if (error || !data) {
-      dispatch_async(dispatch_get_main_queue(), ^{
-        handler(nil, nil, error);
-      });
-      return;
-    }
-    NSString *matchMessage = nil;
-    NSError *parsingError = nil;
-    NSDictionary *parsedDynamicLinkParameters =
-        parserBlock(requestURLString, data, &matchMessage, &parsingError);
-
-    // If request was made with pasteboard contents, verify if we got a unique match. If we got
-    // a "none" match, we were unable to get a unique match or deduce using fingerprinting.
-    // In this case, resend requests to IPV4 and IPV6 endpoints for fingerprinting. b/79704203
-    if (requestBody[@"uniqueMatchLinkToCheck"] && parsedDynamicLinkParameters &&
-        (!parsedDynamicLinkParameters[kFIRDLParameterMatchType] ||
-         [parsedDynamicLinkParameters[kFIRDLParameterMatchType] isEqualToString:@"none"])) {
-      NSMutableDictionary *requestBodyMutable = [requestBody mutableCopy];
-      [requestBodyMutable removeObjectForKey:@"uniqueMatchLinkToCheck"];
-      NSMutableArray *baseURLs =
-          [@[ kIosPostInstallAttributionRestBaseUrlIPV4, kIosPostInstallAttributionRestBaseUrlIPV6 ]
-              mutableCopy];
-      if (parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]) {
-        if ([parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]
-                isEqualToString:@"IP_V4"]) {
-          [baseURLs removeObject:kIosPostInstallAttributionRestBaseUrlIPV4];
-        } else if ([parsedDynamicLinkParameters[kFIRDLParameterRequestIPVersion]
-                       isEqualToString:@"IP_V6"]) {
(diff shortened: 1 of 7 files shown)


FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -116,33 +117,66 @@ - (void)resolveShortLink:(NSURL *)url
     @"sdk_version" : FDLSDKVersion
   };
 
-  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSError *error) {
```
palexsua, line 119: Whenever a call to `reopenAttribution` endpoint took place response object was disregarded despite the fact that its status code indicated an error. On the other hand, `error` object itself was not populated and was `nil`. As a result, all traces of an issue used to be lost here and the calling code had no clue that the url was not allowed.

Screenshot 2020-04-29 at 17 58 06

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -70,7 +71,7 @@ void FIRMakeHTTPRequest(NSURLRequest *request, FIRNetworkRequestCompletionHandle
       [session dataTaskWithRequest:request
                  completionHandler:^(NSData *_Nullable data, NSURLResponse *_Nullable response,
                                      NSError *_Nullable error) {
-                   completion(data, error);
+                   completion(data, response, error);
```
palexsua, line 74: Whenever a call to `reopenAttribution` endpoint took place response object was disregarded despite the fact that its status code indicated an error. On the other hand, `error` object itself was `nil`. As a result, all traces of an issue used to be lost here and the calling code had no clue that the url was not allowed.

Screenshot 2020-04-29 at 17 58 06

paulb777:
@palexsua Thanks for the comprehensive PR!  We'll do our best to get a review soon, although it may be a challenge in the very near term with WWDC next week.

One initial question: Is there a way to make the asynchronous API additive instead of breaking? Changing the current API will require holding off on merging this until the next major release.

palexsua:
> @palexsua Thanks for the comprehensive PR! We'll do our best to get a review soon, although it may be a challenge in the very near term with WWDC next week.
> 
> One initial question: Is there a way to make the asynchronous API additive instead of breaking? Changing the current API will require holding off on merging this until the next major release.

@paulb777 Thank you! I think I can make the asynchronous API additive. Do you want me to open a new PR or update the existing?

palexsua:
@paulb777 @eldhosembabu any update on these changes?

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -116,33 +117,66 @@ - (void)resolveShortLink:(NSURL *)url
...
+                                                             NSError *error) {
+    NSURL *resolvedURL = nil;
+    NSHTTPURLResponse *resp = nil;
+
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+      NSError *err = [NSError
+          errorWithDomain:kGenericErrorDomain
+                     code:0
+                 userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+      handler(resolvedURL, error ? error : err);
```
ncooke3, line 130: Curious, what is the context here in deciding which error to pass in the completion? 

Also, nit: perhaps rename `err` to `responseTypeError`, `customError`, or something like that

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -116,33 +117,66 @@ - (void)resolveShortLink:(NSURL *)url
...
-        NSString *invitationIDString;
-        if ([invitationIDObject isKindOfClass:[NSDictionary class]]) {
-          NSDictionary *invitationIDDictionary = invitationIDObject;
-          invitationIDString = invitationIDDictionary[@"id"];
-        } else if ([invitationIDObject isKindOfClass:[NSString class]]) {
-          invitationIDString = invitationIDObject;
+  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSURLResponse *response,
+                                                             NSError *error) {
+    NSURL *resolvedURL = nil;
+    NSHTTPURLResponse *resp = nil;
```
ncooke3, line 123: nit: maybe rename to something like `completedResponse`

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -116,33 +117,66 @@ - (void)resolveShortLink:(NSURL *)url
...
+
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+      NSError *err = [NSError
+          errorWithDomain:kGenericErrorDomain
+                     code:0
+                 userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+      handler(resolvedURL, error ? error : err);
+    } else {
+      resp = (NSHTTPURLResponse *)response;
```
ncooke3, line 132: If I'm not mistaken, `resp` is only used within this `else`'s scope, correct? Perhaps its declaration could be moved  to within the `else` block.

palexsua:
@ncooke3 Thanks for the comments! I'll have a look later today. I've also pushed one more commit as @paulb777 requested for additive changes.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: Per `FIRNetworkRequestCompletionHandler` definition `error` object is nullable. I decided to give precedence to the original error over the local one (`err`), because I didn't want to shadow potentially more serious issue, e.g. stemming from server side, given that initially `response` object had been disregarded completely.

palexsua: Renamed `err` -> `responseTypeError`.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: Moved.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: As we're primarily interested in `statusCode` I decided to get rid of `resp` object altogether: `NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];`.

palexsua:
@ncooke3 Addressed your review comments in https://github.com/firebase/firebase-ios-sdk/pull/5853/commits/0c20e3bb6148a8dbd24a4712117ffbd2f4e14c93. Thanks!

palexsua:
@ryanwilson I guess after I've made the changes additive `public-api-change` label is not relevant anymore, isn't it?

paulb777:
@palexsua Thanks for refactoring. It's still an API change. It's now no longer an API break so we can potentially merge this for a minor update release instead of waiting for the next rarer major release.

The API change will still need an internal API review. cc: @christibbs, @ncooke3

palexsua:
@paulb777 Got it. Thanks!

FirebaseDynamicLinks/Sources/Public/FIRDynamicLinks.h
```diff
@@ -75,12 +90,12 @@ NS_SWIFT_NAME(DynamicLinks)
  * @return Dynamic Link object if the URL is valid and has link parameter, otherwise nil.
  */
 - (nullable FIRDynamicLink *)dynamicLinkFromUniversalLinkURL:(NSURL *)url
-    NS_SWIFT_NAME(dynamicLink(fromUniversalLink:));
+    NS_SWIFT_NAME(dynamicLink(fromUniversalLink:))
+        DEPRECATED_MSG_ATTRIBUTE("Use dynamicLinkFromUniversalLinkURL:completion: instead.");
```
ncooke3, line 94: @ryanwilson @paulb777 Are there any implications to adding a deprecated message to a currently public API?

paulb777: @ncooke3 It should be fine if the API council agrees with the transition recommendation.

christibbs:
> @palexsua Thanks for refactoring. It's still an API change. It's now no longer an API break so we can potentially merge this for a minor update release instead of waiting for the next rarer major release.
> 
> The API change will still need an internal API review. cc: @christibbs, @ncooke3

Thanks for the work @palexsua. I'll kick off the internal API review with @eldhosembabu.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -118,33 +119,65 @@ - (void)resolveShortLink:(NSURL *)url
...
+  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSURLResponse *response,
+                                                             NSError *error) {
+    NSURL *resolvedURL = nil;
+
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+      NSError *responseTypeError = [NSError
+          errorWithDomain:kGenericErrorDomain
+                     code:0
+                 userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+      handler(resolvedURL, error ? error : responseTypeError);
```
eldhosembabu, line 131: "responseTypeError" is used only if error is nil; so can we defer the responseTypeError error creation?

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -118,33 +119,65 @@ - (void)resolveShortLink:(NSURL *)url
...
-          NSDictionary *invitationIDDictionary = invitationIDObject;
-          invitationIDString = invitationIDDictionary[@"id"];
-        } else if ([invitationIDObject isKindOfClass:[NSString class]]) {
-          invitationIDString = invitationIDObject;
+  FIRNetworkRequestCompletionHandler resolveLinkCallback = ^(NSData *data, NSURLResponse *response,
+                                                             NSError *error) {
+    NSURL *resolvedURL = nil;
+
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+      NSError *responseTypeError = [NSError
```
eldhosembabu, line 127: The whole method change that I can see is about constructing the error object and passing it through. Can we refactor this with a new method which can take in the NSHTTPURLResponse object and returns an NSError object and move all your code changes there?

Then in this method you can call the new method without have much changes to the existing method. This would make the code more readable and we can test the new method individually. Also avoids duplicated lines of code to call handler(url,error) from multiple if else blocks.

So the new method would be something like (Feel free to change the name and signature of the method) : 

   `+(nullable NSError *) extractErrorFromResponse:(NSURLResponse *)response
                            withError:(nullable NSError *) error{
  if(error){
    return error;
  }

  NSError *customError = nil;

  //Your logic to create the custom error.

  return customError;
}`

In current completion handler, we will just call : 

`error = [FIRDynamicLinkNetworking extractErrorFromResponse:response
             withError:error];`

and the rest of the code will remains same. Let me know if that makes sense.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -458,14 +490,21 @@ - (BOOL)handleUniversalLink:(NSURL *)universalLinkURL
                 }];
     return YES;
   } else {
-    FIRDynamicLink *dynamicLink = [self dynamicLinkFromUniversalLinkURL:universalLinkURL];
-    if (dynamicLink) {
-      completion(dynamicLink, nil);
-      return YES;
+    if ([self respondsToSelector:@selector(dynamicLinkFromUniversalLinkURL:completion:)]) {
```
eldhosembabu, line 493: could you please explain why you think we need to check for the existence of "dynamicLinkFromUniversalLinkURL:completion" here?

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -440,6 +440,38 @@ - (nullable FIRDynamicLink *)dynamicLinkFromUniversalLinkURL:(NSURL *)url {
   return nil;
 }
 
+- (void)dynamicLinkFromUniversalLinkURL:(NSURL *)url
```
eldhosembabu, line 443: dynamicLinkFromUniversalLinkURL:completion and dynamicLinkFromUniversalLinkURL both shares the same code apart from completion block calling part. So can we refactor the common code to a method which can be called from dynamicLinkFromUniversalLinkURL:completion and dynamicLinkFromUniversalLinkURL which would avoid code duplication and would make the code maintainable.

eldhosembabu:
hi @palexsua sorry for reviewing the code late from my side. I've left a few code review comments. Please let me know if you have any questions .

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: Sure. Fixed.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: I got the point.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
palexsua: Well, initially I introduced it for compatibility purposes, but now I see that it's not needed.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
palexsua: `completion` callback is called in multiple (2) places inside `dynamicLinkFromUniversalLinkURL:completion` method. Moreover, these methods differ in return value - `void` vs `FIRDynamicLink`. The idea is to deprecate `dynamicLinkFromUniversalLinkURL:` now and remove completely in the upcoming releases. Do you still think it makes sense to refactor it? If you insist though, I can do it. No problem.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
palexsua: Please, see the implementation and if it's ok with you I can add a couple of unit tests.

palexsua:
@eldhosembabu Thank you for your review comments! I appreciate. I've pushed the changes and replied to your comments. Please, have a look.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -101,6 +102,45 @@ - (instancetype)initWithAPIKey:(NSString *)APIKey
...
+  NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];
+  NSError *customError = nil;
+
+  if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+    customError =
+        [NSError errorWithDomain:kGenericErrorDomain
+                            code:0
+                        userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+  } else if (statusCode < 200 || statusCode >= 300) {
+    NSDictionary *result = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
```
eldhosembabu, line 122: Don't we need to check data before using?

palexsua: Right. Added the check here: https://github.com/firebase/firebase-ios-sdk/pull/5853/commits/73f3bb291fbd0b37c7e99fe95033c58e0c03fbd2

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: I agree with the fact that the other function is getting deprecated and will get eventually removed but we are not sure when we are going to do it. Since removing the other method would be a breaking change, I'm not sure that we can do it sooner. So until then, if we need to make any fixes in dynamicLinkFromUniversalLinkURL logic, we will end up having it in two places.

My thought was to have a common internal method which can take the completion handler as well returns the dynamic link so that when called from `dynamicLinkFromUniversalLinkURL:` we will return based on the return of this internal method otherwise if called from `dynamicLinkFromUniversalLinkURL:completion`,we will ignore the return and rely on the completion block. So my proposal would be something like: 

    (nullable FIRDynamicLink *)dynamicLinkInternalFromUniversalLinkURL:(NSURL *)url
                                  completion:(FIRDynamicLinkUniversalLinkHandler)completion {
     //Have existing logic here and make call to completion block only if its not nil.
     ....
     if(completion){
     completion(dynamicLink,error);
     }
     ....

When calling this internal function from `dynamicLinkFromUniversalLinkURL:` just do :

    return  [self dynamicLinkInternalFromUniversalLinkURL:url completion:nil]

  And when calling from `dynamicLinkFromUniversalLinkURL:completion` just call : 

    [self dynamicLinkInternalFromUniversalLinkURL:url completion:completion]

I'm not sure this is a good coding pattern but this would help us with code repetition. Once we remove the old method completely, we can get rid of this internal method. Since my experience with Obj C is limited, I would prefer to get advice on this from @paulb777

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
eldhosembabu: seems good to me apart from a minor comment regarding checking the variable.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: Thanks for removing.

FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m
```diff
@@ -101,6 +102,50 @@ - (instancetype)initWithAPIKey:(NSString *)APIKey
...
+  NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];
+  NSError *customError = nil;
+
+  if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+    customError =
+        [NSError errorWithDomain:kGenericErrorDomain
+                            code:0
+                        userInfo:@{@"message" : @"Response should be of type NSHTTPURLResponse."}];
+  } else if (statusCode < 200 || statusCode >= 300) {
+    if (data) {
```
eldhosembabu, line 131: Can you move this data check to parent `else if` so that you don't need to repeat 
     
    customError = [FIRDynamicLinkNetworking genericErrorForUrl:url];

palexsua: Addressed in https://github.com/firebase/firebase-ios-sdk/pull/5853/commits/11e614d0292f1369947f2e65e4cf4d4873a5ff65.

christibbs:
Update from my end: we've kicked off the internal Firebase API review process.

palexsua:
@eldhosembabu @christibbs any update? please, let me know, if any changes are required.

eldhosembabu:
hi @palexsua now we are on our way towards the api review where the new proposed api needs to be approved by the api council...it would take O(weeks) time to get the approval..If we are not getting followup questions, we should get it approved in a week or two. After that we should be able to merge this PR if its all good.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
paulb777: Agreed. It's worth refactoring to a common internal method - even if it's only a few months - since this much duplicate code could easily lead to inconsistent behavior or bugs.

palexsua:
@eldhosembabu Got it. Thank you for the update!

palexsua:
@eldhosembabu @christibbs guys, any update?

christibbs:
> @eldhosembabu @christibbs guys, any update?

We're working on the API change approval right now. Thanks for your patience.

palexsua:
@eldhosembabu @christibbs guys, any update?

palexsua:
@eldhosembabu @christibbs guys, any update?

palexs:
@paulb777 @eldhosembabu @ncooke3 guys, let's merge this PR.

sbaruth:
Would also love to see this get merged, if it's ready to go.

christibbs:
Hey @palexsua @sbaruth. Really sorry for the delay. We will prioritize this and get through API review this week.

palexs:
@christibbs Thanks for taking care of it! Looking forward to hearing the API review verdict ;)

christibbs:
One additional thing: could you provide an example of if/how you're using the new `dynamicLinkFromUniversalLinkURL:completion:` method in your app?

Is it similar to #6 [here](https://firebase.google.com/docs/dynamic-links/ios/receive)? Returning synchronously from `application:continueUserActivity:restorationHandler:` but using the completion for a side effect?

palexsua:
@christibbs Actually, we do not call `dynamicLinkFromUniversalLinkURL:completion:` directly in our app, but we rely on code that uses it. The issue I observed was that long dynamic links bypassed any sort of verification and when I traced back the issue I ended up seeing that response from `reopenAttribution` endpoint was not being "waited for". So, the gist of the proposed fix is to wait for the response and pass allowed/blocked status via `completion` callback.

christibbs:
@palexsua We got the API approval! @eldhosembabu had one outstanding issue with the implementation, but we can merge once that's resolved.

palexsua:
@christibbs That's great news! Thanks for keeping me in the loop.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: @palexsua could you please address this prior to merging?

palexsua: @eldhosembabu Addressed in https://github.com/firebase/firebase-ios-sdk/pull/5853/commits/0944814ddf4b11cf59f258439797ce3e8dd965a5

palexsua:
@christibbs I've addressed the change suggested by @eldhosembabu.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: thanks @palexsua

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -427,9 +432,21 @@ - (nullable FIRDynamicLink *)dynamicLinkFromUniversalLinkURL:(NSURL *)url {
       }
     }
   }
+  if (completion) {
```
eldhosembabu, line 435: I'm not sure whether this completion block calling is useful here since it would just invoke the block without giving any clue on which url we have tried and failed and so i believe this can be removed. thoughts @christibbs @paulb777 ?

eldhosembabu (commented):
Looks good to me.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
paulb777: It may be needed to satisfy the API if it expects there to be a callback.

However, as this is written, there could be two callbacks if the nested if is reached above and that doesn't seem right.

paulb777 (changes requested):
Thanks for your perseverance on this nice API improvement!

In addition to resolving the open callback question, please update https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDynamicLinks/CHANGELOG.md with a description of the deprecation and new API. (The next planned release version is v7.3.0)

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
christibbs: @paulb777 We return on line 430, which should prevent 422 and 436 from both being fired.

eldhosembabu: I agree with you @paulb777; but i think the 3p users can't do anything with the callback as they don't know to the minimum what url we have tried to parse.

I could be wrong but I'm not seeing a chance for two callback method invocation since we are returning right after the service call in line 425.

paulb777: Oh right, agreed that there can't be two callbacks.  I missed that return.

On the other question, couldn't 3p users check for the nil parameters to know that the url failed to parse or that `url.query.length` was 0?

palexsua:
@paulb777 Afaiu, the open callback question is resolved, isn't it? So, now it's only a matter of updating the changelog?

paulb777:
@palexsua Agreed. Sorry about the confusion. I'll merge it here and we can follow up on the changelog separately.

paulb777 (approved):

christibbs:
@palexsua Thanks for your patience. And congrats: now you are a Firebase contributor!

palexsua:
@christibbs Thank you! It feels great! 🥇

Issue #5783, opened 2020-06-10
Firestore connection delays when enables WiFi in background


shosokawa:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.5 Build version 11E608c
  * Firebase SDK version: 6.25.0
  * Firebase Component: Firestore
  * Component version: 1.14.0
  * Installation method: `CocoaPods`

### [REQUIRED] Step 2: Describe the problem

After changing a connection from 4G to WiFi in the background, it seems that it spent long time (about 20 seconds) to reconnect to the server at when the app has become active.
Some documents are created while the app is in the background, then makes the app active: 
- `snapshotListener` isn't called until about 20 seconds after
- `query.getDocuments` with `FirestoreSource.server` is failed with the error:
```
Error Domain=FIRFirestoreErrorDomain Code=14 "Failed to get documents from server. (However, these documents may exist in the local cache. Run again without setting source to FirestoreSourceServer to retrieve the cached documents.)"
```
If it is called with `FirestoreSource.default`, it returns results from the cache. 

#### Steps to reproduce:

1. Activates an app with WiFi disabled. (4G is used)
2. Makes the app background.
3. Enables WiFi.
4. Creates some documents on firestore in order to make `snapshotListener` called.
4. Makes the app active
-> After about 20 seconds, `snapshotListener` is called

schmidt-sebastian:
@shosokawa Thanks for reporting this.

We have seen similar reports for Transactions (https://github.com/firebase/firebase-ios-sdk/issues/4905) and believe that its fix (https://github.com/firebase/firebase-ios-sdk/pull/4985 ) should apply to all of our networking logic. 

It will take me a couple of days to look at this in more detail, which should allow us figure out why the code changes for #4905 don't apply in this case.  Please do bear with us.

PR #4985, opened 2020-02-27
Checks for connectivity when App is moving to foreground.

wu-hui:
Attempt to fix https://github.com/firebase/firebase-ios-sdk/issues/4905

PR #5857, opened 2020-06-19
Reset backoff when connectivity status changes

schmidt-sebastian:
This is a follow up to https://github.com/firebase/firebase-ios-sdk/pull/4985 and addresses an additional problem that I saw in the test app. If the SDK was already in the background and we were in a backoff phase, the SDK waited for the backoff to finish before reconnecting, which could take up to a minute. This PR plumbs the connectivity handler through to RemoteStore and resets the backoff when a connectivity change occurred.

cc @wu-hui  

Fixes https://github.com/firebase/firebase-ios-sdk/issues/5783

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+- [fixed] Removed delay that may have prevented Firestore from immediately
+  establishing a network connections if a connectivity change occurred while
+  the app was in the background.
+
 # v1.16.0
 - [fixed] Fixed an issue that may have prevented the client from connecting
   to the backend immediately after a user signed in.
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 - [fixed] Removed delay that may have prevented Firestore from immediately
-  establishing a network connections if a connectivity change occurred while
-  the app was in the background.
+  establishing network connections if a connectivity change occurred while
+  in the background.
 
 # v1.16.0
 - [fixed] Fixed an issue that may have prevented the client from connecting
--- a/Firestore/core/src/remote/datastore.cc
+++ b/Firestore/core/src/remote/datastore.cc
@@ -90,21 +90,14 @@ void LogGrpcCallFinished(absl::string_view rpc_name,
 
 Datastore::Datastore(const DatabaseInfo& database_info,
                      const std::shared_ptr<AsyncQueue>& worker_queue,
-                     std::shared_ptr<CredentialsProvider> credentials)
-    : Datastore{database_info, worker_queue, credentials,
-                ConnectivityMonitor::Create(worker_queue)} {
-}
-
-Datastore::Datastore(const DatabaseInfo& database_info,
-                     const std::shared_ptr<AsyncQueue>& worker_queue,
                      std::shared_ptr<CredentialsProvider> credentials,
-                     std::unique_ptr<ConnectivityMonitor> connectivity_monitor)
+                     ConnectivityMonitor* connectivity_monitor)
     : worker_queue_{NOT_NULL(worker_queue)},
       credentials_{std::move(credentials)},
       rpc_executor_{CreateExecutor()},
       connectivity_monitor_{std::move(connectivity_monitor)},
       grpc_connection_{database_info, worker_queue, &grpc_queue_,
-                       connectivity_monitor_.get()},
+                       connectivity_monitor_},
       datastore_serializer_{database_info} {
   if (!database_info.ssl_enabled()) {
     GrpcConnection::UseInsecureChannel(database_info.host());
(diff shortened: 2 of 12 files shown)


Firestore/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Unreleased
+- [fixed] Removed a delay that may have prevented Firestore from immediately
+  establishing a network connections if a connectivity change occurred while
```
var-const, line 3: Nit: either "a network connection" or "network connections".

Firestore/core/src/remote/datastore.cc
```diff
@@ -88,23 +88,16 @@ void LogGrpcCallFinished(absl::string_view rpc_name,
...
-
 Datastore::Datastore(const DatabaseInfo& database_info,
                      const std::shared_ptr<AsyncQueue>& worker_queue,
                      std::shared_ptr<CredentialsProvider> credentials,
-                     std::unique_ptr<ConnectivityMonitor> connectivity_monitor)
+                     ConnectivityMonitor* connectivity_monitor)
     : worker_queue_{NOT_NULL(worker_queue)},
       credentials_{std::move(credentials)},
       rpc_executor_{CreateExecutor()},
       connectivity_monitor_{std::move(connectivity_monitor)},
```
var-const, line 98: Nit: this move is now extraneous (moving a built-in, such as a pointer, is the same as copying it).

Firestore/core/src/remote/remote_store.cc
```diff
@@ -59,13 +59,15 @@ using util::Status;
 constexpr int kMaxPendingWrites = 10;
 
 RemoteStore::RemoteStore(
-    LocalStore* local_store,
+    local::LocalStore* local_store,
```
var-const, line 62: Nit: qualifying `LocalStore` with its namespace shouldn't be necessary given the `using` directive above.

Firestore/core/src/remote/remote_store.cc
```diff
@@ -77,6 +79,15 @@ void RemoteStore::Start() {
   // For now, all setup is handled by `EnableNetwork`. We might expand on this
   // in the future.
   EnableNetwork();
+
+  connectivity_monitor_->AddCallback(
+      [this](ConnectivityMonitor::NetworkStatus /*ignored*/) {
+        if (CanUseNetwork()) {
+          LOG_DEBUG("RemoteStore %s restarting streams as connectivity changed",
+                    this);
+          RestartNetwork();
```
var-const, line 88: This doesn't create any undesirable interactions with the existing logic on connectivity change, right? (I don't think it should, just checking)

Firestore/core/src/remote/remote_store.cc
```diff
@@ -77,6 +79,15 @@ void RemoteStore::Start() {
   // For now, all setup is handled by `EnableNetwork`. We might expand on this
   // in the future.
   EnableNetwork();
+
+  connectivity_monitor_->AddCallback(
+      [this](ConnectivityMonitor::NetworkStatus /*ignored*/) {
```
var-const, line 84: Is it necessary/desirable to run this logic when the network status is "unavailable"?

Firestore/core/src/remote/remote_store.cc
```diff
@@ -528,16 +539,22 @@ absl::optional<TargetData> RemoteStore::GetTargetDataForTarget(
...
   if (CanUseNetwork()) {
     // Tear down and re-create our network streams. This will ensure we get a
     // fresh auth token for the new user and re-fill the write pipeline with new
     // mutations from the `LocalStore` (since mutations are per-user).
     LOG_DEBUG("RemoteStore %s restarting streams for new credential", this);
-    is_network_enabled_ = false;
-    DisableNetworkInternal();
-    online_state_tracker_.UpdateState(OnlineState::Unknown);
-    EnableNetwork();
+    RestartNetwork();
```
var-const, line 557: Looks like this PR could also make reconnection faster upon credentials' change? If correct, perhaps it's worth mentioning in the PR's description?

Firestore/CHANGELOG.md
schmidt-sebastian: Fixed

Firestore/core/src/remote/datastore.cc
schmidt-sebastian: Fixed

Firestore/core/src/remote/remote_store.cc
schmidt-sebastian: Fixed

Firestore/core/src/remote/remote_store.cc
schmidt-sebastian: Probably not - but note that https://github.com/firebase/firebase-ios-sdk/blob/8633f5f6a3ca936085c9ee90680b45660eb19dd0/Firestore/core/src/remote/grpc_connection.cc#L296 kills the GRPC channel no matter what.

Firestore/core/src/remote/remote_store.cc
schmidt-sebastian: The existing logic invalidates the GRPC channel, which speeds up reconnects already and is the foundation this PR is based on.

Firestore/core/src/remote/remote_store.cc
schmidt-sebastian: Only if the credential change happened when while offline, which is probably uncommon (since Auth needs a network connection as well).

schmidt-sebastian:
Verified manually.

PR #5861, opened 2020-06-19
Repo-relative imports for Database

paulb777:
- Run ./scripts/change_headers.swift to update all Database library and tests to be repo-relative
- levelDB headers and SwiftPM are pending Swift Package Manager support for leveldb
- Only run integration tests on iOS to avoid flakiness

#no-changelog - Will be part of a global M74 changelog


.github/workflows/database.yml
```diff
@@ -32,15 +32,13 @@ jobs:
...
-    strategy:
-      matrix:
-        target: [iOS, tvOS, macOS]
     steps:
     - uses: actions/checkout@v2
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: IntegrationTest
-      run: scripts/third_party/travis/retry.sh scripts/build.sh Database ${{ matrix.target }} integration
+      # Only iOS to mitigate flakes.
```
maksymmalyhin, line 40: I wonder may the flakes occur due to concurrent tests execution (modifying the same database keys from tests fro different platforms simultaneously)? If so, we probably can fix it by modifying the tests, but it's probably out of scope of the PR.

paulb777: Yep, I wondered the same thing - but we've also disabled other tests in this suite because of flakiness from a single run. 

In any case until we invest in improving the robustness of these tests, its not worth the cost of running the integration tests for non-iOS platforms.

PR #5862, opened 2020-06-20
Refactor GULSwizzledObject to ARC to unblock SwiftPM support

maksymmalyhin:
- make sure an object is swizzled only once
- call `objc_disposeClassPair(generatedClass);` in `[GULObjectSwizzler dealloc]` only when the swizzled object has been deallocated
- asynchronously call `objc_disposeClassPair(generatedClass);` after a delay only when the swizzled object class matches the generated class


maksymmalyhin:
@visumickey Friendly ping.

GoogleUtilities/ISASwizzler/GULObjectSwizzler.m
```diff
@@ -98,10 +104,9 @@ - (void)copySelector:(SEL)selector fromClass:(Class)aClass isClassSelector:(BOOL
                                   : class_getInstanceMethod(aClass, selector);
   Class targetClass = isClassSelector ? object_getClass(_generatedClass) : _generatedClass;
   IMP implementation = method_getImplementation(method);
+
   const char *typeEncoding = method_getTypeEncoding(method);
-  BOOL success __unused = class_addMethod(targetClass, selector, implementation, typeEncoding);
-  NSAssert(success, @"Unable to add selector %@ to class %@", NSStringFromSelector(selector),
-           NSStringFromClass(targetClass));
+  class_replaceMethod(targetClass, selector, implementation, typeEncoding);
```
visumickey, line 109: Are there any cleanups to the older method that needs to be done before replacing to the new implementation?

With addMethod, we would fail if an implementation already existed. With this change, we might mask the failure and force replace with the new implementation.

GoogleUtilities/ISASwizzler/GULObjectSwizzler.m
```diff
@@ -123,11 +128,20 @@ - (nullable id)getAssociatedObjectForKey:(NSString *)key {
...
+    // The object has been swizzled already.
+    return;
+  }
+
   if (swizzledObject) {
     [GULObjectSwizzler setAssociatedObject:swizzledObject
                                        key:kSwizzlerAssociatedObjectKey
                                      value:self
-                               association:GUL_ASSOCIATION_RETAIN_NONATOMIC];
+                               association:GUL_ASSOCIATION_RETAIN];
```
visumickey, line 144: Why is this atomic? Were there any thread safety issues we had on this earlier?

maksymmalyhin: No, there were no thread safety issues observed by me. I added it because technically the  associated object may be accessed from different threads. E.g. invoking a method on some swizzled objects implies access to the associated object, see [here](https://github.com/firebase/firebase-ios-sdk/blob/296e5504ac6dd27baf7d04a4f7093fb53092e4b0/GoogleUtilities/ISASwizzler/GULSwizzledObject.m#L60). Though I recognize that atomic may introduce additional performance cost,  I would prefer to keep it atomic since we don't have control over the swizzled object life cycle and it's better to be slower than e.g. crashing. WDYT?

GoogleUtilities/ISASwizzler/GULObjectSwizzler.m
maksymmalyhin: My assumption here was that replacing the method implementation is a preferable action. 

What type of handling we would prefer to have here? We can reintroduce the debug assertion in the case the method was replaced or we can return previous behaviour - silently don't replace the implementation and assert under DEBUG only.

visumickey (approved):

PR #5874, opened 2020-06-22
Remove prioritizers, upload packages, data futures, and backwards compatibility

mikehaney24:
This PR will complete the initial version of the new flat file storage system. All persisted state should be saved using storage interface.

Changes:
- All singletons restore their state from the libraryData API. They don't do weird things with NSCoding anymore
- All unique ID generation should be done using the libraryData API
- Prioritizers are no longer necessary because uploaders create event selectors to find events for uploading
- GDTCOREvents aren't backwards compatible with their previous versions (or the old GDTCORStoredEvent class), so the coding keys and coder methods have been updated.
- GDTCOREvents don't track the location to which they've been written. All state is managed by the storage implementation.
- `-transportBytes` is called async as soon as the dataObject is set on the GDTCOREvent
- Batches are stored to the filesystem as well, so the upload coordinator's job is greatly simplified

Files to pay the most attention to:
- GDTCOREvent.m
- GDTCCTUploader.m
- GDTCORFlatFileStorage.m

#no-changelog because this is part of a feature branch


GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -68,15 +58,26 @@ - (nullable instancetype)initWithMappingID:(NSString *)mappingID target:(NSInteg
   if (mappingID == nil || mappingID.length == 0 || target <= 0) {
     return nil;
   }
+  __block NSNumber *eventID;
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  [GDTCOREvent nextEventIDForTarget:target
+                         onComplete:^(NSNumber *_Nullable newEventID) {
+                           eventID = newEventID;
```
maksymmalyhin, line 65: `newEventID` is nullable, but property `eventID` is marked as `nonnull`. Should we fail initialization in this case?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -68,15 +58,26 @@ - (nullable instancetype)initWithMappingID:(NSString *)mappingID target:(NSInteg
...
+                           dispatch_semaphore_signal(sema);
+                         }];
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 5 * NSEC_PER_SEC)) != 0) {
```
maksymmalyhin, line 68: I think we will need to find a way to avoid blocking the thread, but it doesn't look like a blocker to me.

Just to have a better context: why assigning `eventID` by storage (as it was implemented originally) doesn't work?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -196,69 +149,44 @@ + (BOOL)supportsSecureCoding {
...
-    _clockSnapshot = [aDecoder decodeObjectOfClass:[GDTCORClock class] forKey:clockSnapshotKey];
-    NSURL *fileURL = [aDecoder decodeObjectOfClass:[NSURL class] forKey:fileURLKey];
-    if (fileURL) {
-      _GDTFilePath = [fileURL lastPathComponent];
-    } else {
-      _GDTFilePath = [aDecoder decodeObjectOfClass:[NSString class] forKey:kGDTFilePathKey];
+      dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+      [GDTCOREvent nextEventIDForTarget:_target
+                             onComplete:^(NSNumber *_Nullable eventID) {
+                               self->_eventID = eventID;
```
maksymmalyhin, line 161: The same here - `eventID` is nullable.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: Fixed.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: Fixed, good catch.

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
mikehaney24: The previous mechanism worked, but I'd rather the saving of data be consolidated to one place. That way, it can also be reset or deleted when necessary.

Also, the previous mechanism was blocking as well, so this isn't different than what is currently shipping. We have gotten some complaints about high cpu usage here, so it might be worth making it async in the future (I'm not sure I want to tackle that this PR, since there's no degradation in performance at least).

PR #5877, opened 2020-06-23
Completely reimplement storage in GDT

mikehaney24:
This merges the new storage implementation feature to master. Most of these changes have been reviewed already (I pushed changes to the CHANGELOG unilaterally).

Changes:
- All library state should be written to files
- All singletons restore their state from the libraryData API. They don't do weird things with NSCoding anymore
- All unique ID generation should be done using the libraryData API
- Prioritizers are no longer necessary because uploaders create event selectors to find events for uploading
- GDTCOREvents aren't backwards compatible with their previous versions (or the old GDTCORStoredEvent class), so the coding keys and coder methods have been updated.
- GDTCOREvents don't track the location to which they've been written. All state is managed by the storage implementation.
- -transportBytes is called async as soon as the dataObject is set on the GDTCOREvent
- Batches are stored to the filesystem as well, so the upload coordinator's job is greatly simplified

cc: @samedson @visumickey


mikehaney24:
> Will podspec versions be updated in a subsequent PR?

Yes, I was going to update dependencies and versions in a subsequent PR.

maksymmalyhin:
Do you have an idea why tests pass on GHA but fail on Travis?

mikehaney24:
> Do you have an idea why tests pass on GHA but fail on Travis?

I don't, but I did find a bug when I uncommented a test that I meant to have enabled (deleting while enumerating). I'll continue looking into it.

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
```diff
@@ -66,351 +87,155 @@ - (void)tearDown {
...
+    [storage storeEvent:event onComplete:nil];
+  }
+
+  // Generate 50 FLL target events.
+  for (int i = 0; i < 50; i++) {
+    GDTCOREvent *event = [GDTCOREventGenerator generateEventForTarget:kGDTCORTargetFLL
+                                                              qosTier:nil
+                                                            mappingID:nil];
+    [generatedEvents addObject:event];
+    [storage storeEvent:event onComplete:nil];
```
maksymmalyhin, line 117: Should we wait for completion here? I guess we don't have to since storage should enqueue operations and if we have to, then it may be an actual storage issue

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -507,6 +507,9 @@ + (NSString *)pathForTarget:(GDTCORTarget)target
                     qosTier:(NSNumber *)qosTier
              expirationDate:(NSDate *)expirationDate
                   mappingID:(NSString *)mappingID {
+  NSMutableCharacterSet *allowedChars = [[NSCharacterSet alphanumericCharacterSet] mutableCopy];
```
maksymmalyhin, line 510: nit: I think, it can be `[NSMutableCharacterSet alphanumericCharacterSet]`, but the concern is not big enough to restart CI again.

GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m
```diff
@@ -196,6 +196,15 @@ GDTCORNetworkMobileSubtype GDTCORNetworkMobileSubTypeMessage() {
       resultData = [NSKeyedArchiver archivedDataWithRootObject:obj];
 #pragma clang diagnostic pop
       if (archivePath) {
+        BOOL result = [[NSFileManager defaultManager]
```
maksymmalyhin, line 199: The lines 173-188 look mostly identical to 198-213. I wonder, can we avoid duplication somehow?

mikehaney24: Done.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
mikehaney24: Ack.

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
mikehaney24: Wasn't necessary.

mikehaney24:
Travis state showing in the checks section is messed up for some reason, but I can confirm it's green: https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/701450758

PR #5878, opened 2020-06-23
Remove initializer annotations

mikehaney24:
Also update GDTCOREvent target types to GDTCORTarget

#no-changelog because this is a feature review branch

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
+++ b/GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
@@ -80,7 +80,7 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
  * @param target The event's target identifier.
  * @return An instance of this class.
  */
-- (nullable instancetype)initWithMappingID:(NSString *)mappingID target:(NSInteger)target;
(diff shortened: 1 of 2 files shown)


GoogleDataTransport/GDTCORLibrary/Public/GDTCOREvent.h
```diff
@@ -73,17 +74,13 @@ typedef NS_ENUM(NSInteger, GDTCOREventQoS) {
...
-
 /** Initializes an instance using the given mappingID.
  *
  * @param mappingID The mapping identifier.
  * @param target The event's target identifier.
  * @return An instance of this class.
  */
-- (nullable instancetype)initWithMappingID:(NSString *)mappingID
-                                    target:(NSInteger)target NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithMappingID:(NSString *)mappingID target:(GDTCORTarget)target;
```
maksymmalyhin, line 83: Should `mappingID` be nullable as well?

mikehaney24: It shouldn't be nullable in the initializer, but if you call `-init` it'll produce a nil value when you call the getter.

PR #5884, opened 2020-06-23
Add Firebase module for SPM

paulb777:
Update Firebase.h to a scheme that works for both CocoaPods and Swift Package Manager

This works for CocoaPods binary and source distros and SPM source distros.

SPM binary distros still need to be verified.

#no-changelog


Vincent-Loi:
Just needed this today - thank you so much! Possible to add `FirebaseRemoteConfig` and `FirebaseDatabase` though?

paulb777:
@Vincent-Loi We're working on it ...

PR #5887, opened 2020-06-23
Remote Config Testing Console  API

ncooke3:
Adding a testing API (and some tests to showcase it) for interacting with the Remote Config console. Included a README with instructions on how to run the added tests and play with the testing API. With this addition, we could run tests to test remote config's implementation using the fake console API when we don't need to test with a live console, and this new api when we do want to test with a console.


FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -18,10 +18,24 @@ import FirebaseCore
 import XCTest
 
 class APITests: APITestBase {
+  var console: RemoteConfigConsole!
+
   override func setUp() {
     super.setUp()
     if APITests.useFakeConfig {
       fakeConsole.config = ["Key1": "Value1"]
+    } else {
```
paulb777, line 27: Shouldn't there be three cases here?  1. fakeConsole 2. RemoteConfigConsole() 3. Real console.

ncooke3: Yes, you're right!

ncooke3:
Going to open a new PR for this because i messed this one up 🤦‍♂️

ncooke3:
Reopening for fix

ncooke3:
Closing in favor of updated PR

PR #5890, opened 2020-06-24
Remove Protobuf from the ABT SDK

christibbs:
- Removes the generated `ABTExperimentPayload` from the ABT SDK.
- Replaces it with an NSObject subclass that contains the same metadata.
- Fixes up unit tests.

- Updates Remote Config and In-App Messaging to consume "new" API


FirebaseABTesting.podspec
```diff
@@ -45,10 +45,10 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
     'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
   }
   s.dependency 'FirebaseCore', '~> 6.8'
```
paulb777, line 47: Delete `GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS=1 ` above.

scripts/check_imports.swift
```diff
@@ -99,7 +99,9 @@ func checkFile(_ file: String, isPublic: Bool) {
       let importFile = line.components(separatedBy: " ")[1]
       if inSwiftPackageElse {
         if importFile.first != "<" {
-          genError("Import error: \(file):\(lineNum) Import in SWIFT_PACKAGE #else should start with \"<\".")
```
paulb777, line 102: Looks like your local swiftformat might be the wrong version - see https://github.com/firebase/firebase-ios-sdk#code-formatting or just revert this change.

FirebaseABTesting/Sources/Public/ABTExperimentPayload.h
```diff
@@ -0,0 +1,61 @@
...
+  ABTExperimentPayloadExperimentOverflowPolicyUnrecognizedValue = 999,
+  ABTExperimentPayloadExperimentOverflowPolicyUnspecified = 0,
+  ABTExperimentPayloadExperimentOverflowPolicyDiscardOldest = 1,
+  ABTExperimentPayloadExperimentOverflowPolicyIgnoreNewest = 2,
+};
+
+@interface ABTExperimentLite : NSObject
+@property(nonatomic, readonly, copy) NSString *experimentId;
+
+- (instancetype)initWithExperimentId:(NSString *)experimentId;
```
chliangGoogle, line 29: maybe mark this designated initializer?

FirebaseABTesting/Sources/ABTExperimentPayload.m
```diff
@@ -0,0 +1,122 @@
...
+    _triggerEvent = dictionary[kExperimentPayloadKeyTriggerEvent];
+    _setEventToLog = dictionary[kExperimentPayloadKeySetEventToLog];
+    _activateEventToLog = dictionary[kExperimentPayloadKeyActivateEventToLog];
+    _clearEventToLog = dictionary[kExperimentPayloadKeyClearEventToLog];
+    _timeoutEventToLog = dictionary[kExperimentPayloadKeyTimeoutEventToLog];
+    _ttlExpiryEventToLog = dictionary[kExperimentPayloadKeyTTLExpiryEventToLog];
+
+    // Other.
+    NSDate *experimentStartTime = [[[self class] experimentStartTimeFormatter]
+        dateFromString:dictionary[kExperimentPayloadKeyExperimentStartTime]];
```
chliangGoogle, line 82: nit, just want to make sure if dictionary[kExperimentPayloadKeyExperimentStartTime] is nil, it won't crash here.

FirebaseABTesting/Sources/FIRExperimentController.m
```diff
@@ -34,19 +35,19 @@
...
 
 /// Deserialize the experiment payloads.
 ABTExperimentPayload *ABTDeserializeExperimentPayload(NSData *payload) {
+  // Verify that we have a JSON object.
   NSError *error;
-  ABTExperimentPayload *experimentPayload = [ABTExperimentPayload parseFromData:payload
-                                                                          error:&error];
-  if (error) {
+  id JSONObject = [NSJSONSerialization JSONObjectWithData:payload options:kNilOptions error:&error];
+  if (JSONObject == nil) {
```
chliangGoogle, line 46: When there's no experiment running, is it possible that JSON object is just nil without error?

FirebaseABTesting.podspec
christibbs: Done in dade3e0.

FirebaseABTesting/Sources/FIRExperimentController.m
christibbs: Since this method gets called from an enumeration, the payload data should never be nil.  Only way this serialization should fail is if it's a non-JSON object.

FirebaseABTesting/Sources/ABTExperimentPayload.m
christibbs: The changes in c2b6de3 prevent this.

FirebaseABTesting/Sources/Public/ABTExperimentPayload.h
```diff
@@ -0,0 +1,69 @@
+// Copyright 2020 Google
```
paulb777, line 1: If we're not expanding the public API, this file should be put in Sources/Private and added to s.private_header_files in the podspec.

christibbs: Done in 3542804.

FirebaseABTesting/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v3.4.0
+- [changed] Removed Protbuf dependency (#5890).
```
paulb777, line 2: nit: Protobuf spelling

FirebaseABTesting/Sources/ABTExperimentPayload.m
```diff
@@ -0,0 +1,151 @@
+// Copyright 2020 Google
```
paulb777, line 1: Copyright 2020 Google LLC

New copyrights should follow go/copyright

FirebaseABTesting.podspec
```diff
@@ -35,20 +35,19 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
    'FirebaseCore/Sources/Private/*.h',
   ]
   s.requires_arc = base_dir + '*.m'
-  s.public_header_files = base_dir + 'Public/*.h', base_dir + 'Protos/developers/mobile/abt/proto/*.h'
-  s.private_header_files = base_dir + 'Protos/developers/mobile/abt/proto/*.h'
+  s.public_header_files = base_dir + 'Public/*.h'
```
paulb777, line 38: It's weird, but the Private headers should be in the `public_header_files` line additionally.  Full details in the podspec docs.

christibbs: Fixed in 8e836ca.

FirebaseABTesting/CHANGELOG.md
christibbs: Fixed in c75c41b.

FirebaseABTesting/Sources/ABTExperimentPayload.m
christibbs: Fixed in 511e0b3.

paulb777:
The travis issues reproduce locally with Xcode 10.3 targeting an iPhone5 simulator. Likely there is a 32bit/64bit data processing issue.

Screen Shot 2020-06-26 at 9 45 18 AM

FirebaseABTesting/Sources/Public/FIRExperimentController.h
```diff
@@ -17,15 +17,15 @@
 @class ABTExperimentPayload;
 
 // Forward declaration to avoid importing into the module header
-typedef NS_ENUM(int32_t, ABTExperimentPayload_ExperimentOverflowPolicy);
```
paulb777, line 20: Changing the type name will break RemoteConfig users who lock their version and do a `pod update`. Can we keep the type name unchanged until the next major version.

Otherwise, we could consider bumping the major version since we don't really have a public API for ABT.

cc: @ryanwilson

christibbs: I'd prefer bumping the major version since the underscores aren't idiomatic Objective-C. What do you guys think?

christibbs:
> The travis issues reproduce locally with Xcode 10.3 targeting an iPhone5 simulator. Likely there is a 32bit/64bit data processing issue.
> 
> Screen Shot 2020-06-26 at 9 45 18 AM

Thanks for looking into that. I didn't have iPhone 5 on my simulators list. I changed the millisecond properties to `long long` in de42dd0.

paulb777:
I'm not sure about changing the double floating types to int types. I would have thought it would be changing the intValue's to int64Value's - https://developer.apple.com/documentation/foundation/nsnumber/1416870-int64value.

@ryanwilson Will you answer the API question? Are we ok to bump the major version of FirebaseABTesting without a Firebase major version since there are no public ABTesting APIs?

ryanwilson:
After considering it, I think I'm fine with bumping the major version of ABTesting without a Firebase major version.

christibbs:
> After considering it, I think I'm fine with bumping the major version of ABTesting without a Firebase major version.

Great. Bumped major version in 84d865f.

christibbs:
> One more change and then LGTM:
> 
> Bump the FirebaseABTesting version in Firebase.podspec

Done in 8e68e3e.

Issue #5881, opened 2020-06-23
Failed to iterate leveldb directory: Attempting to fetch value instead of handling error Not found (errno 2: No such file or directory)


pradipaub36:
### Step 0: Are you in the right place?

  Yes

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.3.1
  * Firebase SDK version: 6.24.0
  * Firebase 6.24.0 
  * FirebaseABTesting 3.2.0
  * FirebaseAnalytics 6.5.0
 * FirebaseAuth 6.5.3
 * FirebaseCore 6.7.0
 * FirebaseCoreDiagnostics 1.3.0 
 * FirebaseCrashlytics 4.1.0
 * FirebaseDatabase 6.2.1 
 * FirebaseFirestore 1.13.0
 * FirebaseInAppMessaging 0.20.0 
 * FirebaseInstallations 1.2.0 
 * FirebaseInstanceID 4.3.4 
 * FirebaseMessaging 4.4.0 
 * Firebase Component: Auth, Database, Firestore, Messaging, Storage
 * Component version: _____
 * Installation method: CocoaPods

### [REQUIRED] Step 2: Describe the problem
Our app is crashing with this error code. I couldn't find a way to reproduce this error.

![Screenshot 2020-06-23 at 9 32 28 AM](https://user-images.githubusercontent.com/54441132/85359655-81460d00-b534-11ea-89f2-575beaffe0e2.png)

```
Fatal Exception: NSInternalInconsistencyException
FIRESTORE INTERNAL ASSERTION FAILED: Attempting to fetch value instead of handling error Not found: Failed to stat file: /var/mobile/Containers/Data/Application/AC16519C-25FC-4CDA-A36B-FE5B7DC56C3B/Library/Application Support/firestore/__FIRAPP_DEFAULT/cubii-android/main/000057.ldb (errno 2: No such file or directory)
firebase::firestore::util::ObjcThrowHandler(firebase::firestore::util::ExceptionType, char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)

Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x1ac9a096c __exceptionPreprocess
1  libobjc.A.dylib                0x1ac6b9028 objc_exception_throw
2  CoreFoundation                 0x1ac89d4fc +[_CFXNotificationTokenRegistration keyCallbacks]
[...]

PR #5892, opened 2020-06-24
Ignore errors enumerating files during GC

wilhuff:
Prevents symptoms like those in #5881 by logging the error and skipping the current cycle. There might be more work to do next cycle but this isn't a big deal because GC only promises that cache size limitations will be approximate.


Firestore/core/test/unit/local/lru_garbage_collector_test.cc
```diff
@@ -657,7 +660,9 @@ TEST_P(LruGarbageCollectorTest, RemoveTargetsThenGC) {
 TEST_P(LruGarbageCollectorTest, GetsSize) {
   NewTestResources();
 
-  int64_t initial_size = gc_->CalculateByteSize();
+  StatusOr<int64_t> maybe_initial_size = gc_->CalculateByteSize();
```
var-const, line 663: It's probably too much trouble to test the failure case?

Firestore/core/src/local/lru_garbage_collector.cc
```diff
@@ -26,16 +26,18 @@
...
 
-using firebase::firestore::api::Settings;
-using firebase::firestore::model::DocumentKey;
-using firebase::firestore::model::ListenSequenceNumber;
-using firebase::firestore::model::TargetId;
+using api::Settings;
+using model::DocumentKey;
+using model::ListenSequenceNumber;
+using model::TargetId;
+using util::StatusOr;
```
var-const, line 40: Nit: within the PR, some source files qualify `util::StatusOr` and some rely on a `using` declaration like this. Should we pick one and make this consistent? (can certainly be done in a follow-up)

Firestore/core/src/local/leveldb_persistence.cc
```diff
@@ -178,20 +178,30 @@ util::Status LevelDbPersistence::ClearPersistence(
...
 
-  HARD_ASSERT(iter->status().ok(), "Failed to iterate LevelDB directory: %s",
-              iter->status().error_message().c_str());
-  HARD_ASSERT(count >= 0 && count <= std::numeric_limits<int64_t>::max(),
-              "Overflowed counting bytes cached");
+  if (!iter->status().ok()) {
+    return Status::FromCause("Failed to iterate over LevelDB files",
+                             iter->status());
+  }
+  if (count < 0) {
```
var-const, line 201: I'm not sure about this check. `int` overflow is undefined behavior, so it's not guaranteed to work. Also, technically it should be an assertion (the behavior of the whole program is undefined from this point). OTOH, I guess this is likely to work in practice, so feel free to keep as is (it's not like we have a great alternative).

Firestore/core/test/unit/local/lru_garbage_collector_test.cc
wilhuff: Under the current code structure, the test would have to intentionally create a race condition to prompt the failure. Given how quickly you can stat files I'm not sure that's a condition we can create reliably.

I could change this around so that the underlying size calculation takes an iterator as an argument, but this couldn't be expressed in the platform agnostic code this is testing and so it would only be a low-level unit test that errors during iteration are actually propagated into the result. I'm not sure this is terribly useful though.

All told, I'm in the camp of more trouble than it's worth without dramatically altering the dependency structure of all these components.

Firestore/core/src/local/leveldb_persistence.cc
wilhuff: PTAL I've adjusted this to use an unsigned accumulator and then check for overflow on each iteration.

Firestore/core/src/local/lru_garbage_collector.cc
wilhuff: There were a few stray references in `leveldb_persistence.cc` that I've fixed. Otherwise the difference is headers vs source files. In headers we can't have the using statements that shorten this.

Firestore/core/src/local/leveldb_persistence.cc
```diff
@@ -178,10 +178,12 @@ util::Status LevelDbPersistence::ClearPersistence(
   return fs->RecursivelyRemove(leveldb_dir);
 }
 
-util::StatusOr<int64_t> LevelDbPersistence::CalculateByteSize() {
+StatusOr<int64_t> LevelDbPersistence::CalculateByteSize() {
   auto* fs = Filesystem::Default();
 
-  int64_t count = 0;
+  // Accumulate the total size in an unsigned quantity to avoid undefined
```
var-const, line 184: Ultranit: `quantity` seems a little peculiar -- I'd expect `variable` or `integer`.

Firestore/core/test/unit/local/lru_garbage_collector_test.cc
var-const: Sounds good to me.

Firestore/core/src/local/lru_garbage_collector.cc
var-const: Yeah, I only meant source files. Thanks for fixing!

Firestore/core/src/local/leveldb_persistence.cc
var-const: LGTM.

Firestore/core/src/local/leveldb_persistence.cc
wilhuff: Fixed.

PR #5894, opened 2020-06-24
Add crashlytics swift framework testing.

granluo:


granluo:
Can we have this change 
https://github.com/firebase/quickstart-ios/pull/902
applied to CrashlyticsExampleSwift as well? So the run binary can be also be found using zip?

granluo:
> Can we have this change
> [firebase/quickstart-ios#902](https://github.com/firebase/quickstart-ios/pull/902)
> applied to CrashlyticsExampleSwift as well? So the run binary can be also be found using zip?

Done in https://github.com/firebase/quickstart-ios/pull/994
The test was rerun and done in 9b5e0df

PR #5902, opened 2020-06-24
Consolidate GDT and GDTCCT to a single pod

paulb777:
Consolidate GoogleDataTransportCCTSupport into GoogleDataTransport and remove its references from the repo.

Comments out a test assert that also seems broken on master.

Doesn't address the CoreDiagnostics test failures that exist on master.


paulb777:
@mikehaney24 `pod gen` of multiple platforms is problematic for CocoaPods and Xcode. It can generate differences across `pod lint` and Xcode versions.  I updated the `generate_porject.sh` script to generate a single platform on an option with iOS the default.

paulb777:
@mikehaney24 Good catch on the script name. Fixed.

Hmm, the `pod lib lint` tests were failing for me because of the different module name associated with the single platform runs. This change makes `pod gen` and `pod lib lint` consistent.  `./generate_project.sh macos` will generate a mac project that can be tested separate from `pod lib lint`.

Issue #5893, opened 2020-06-24
159818353: Dynamic Links causes a warning banner to appear under iOS 14


rhysforyou:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Version 11.5 (11E608c)
  * Firebase SDK version: 6.10.0
  * Firebase Component: Dynamic Links
  * Component version: 6.10.0
  * Installation method: CocoaPods

### [REQUIRED] Step 2: Describe the problem

When launching my app on a device running the iOS 14 or iPadOS 14 beta, when I've copied something on another device, I get this system generated banner:

![Image from iOS (1)](https://user-images.githubusercontent.com/320910/85512838-af634400-b63d-11ea-91fe-5795c7b8da84.png)

This appears to be caused by Firebase Dynamic Links, which attempts to check for a dynamic link saved in the pasteboard by a previously visited landing page.

Not sure what the best solution here is, but we're definitely going to get complaints when users update to iOS 14, launch our app, and immediately see something that looks like us harvesting their data.

PycKamil:
There are new  iOS 14 Pasteboard API available on, this should probably be used now to avoid getting this banner https://developer.apple.com/documentation/uikit/uipasteboard/detectionpattern/3618951-probableweburl?changes=latest_major

ryanwilson:
Thanks for the report! And thanks @PycKamil for pointing out the new Pasteboard API. We'll look at getting this resolved quickly because you're right - that's not a good user experience at all.

rohandandavati:
Internal tracking: b/159818353

eldhosembabu:
May be we should be able to query the pasteboard with new api (https://developer.apple.com/documentation/uikit/uipasteboard/detectionpattern?changes=latest_minor) before trying to access it.

ncooke3:
Hey everybody, 

While we are still exploring potential solutions, we would appreciate feedback on a current approach we are exploring. Feel free to pull in the below version of 'Firebase/DynamicLinks' and let us know if this reduces the issue in your tests.

```ruby
pod 'Firebase/DynamicLinks', :git => 'https://github.com/firebase/firebase-ios-sdk.git', :branch => 'nc-dl-ios14'
```

We'll keep this thread updated with future developments!

rasod:
@ncooke3 Out of transparency could we have some questions officially answered regarding this feature.

[...]

PR #5905, opened 2020-06-24
Checking the pasteboard for dynamic links in ios14

ncooke3:
In iOS14,  a notification will be be displayed to the user whenever an app reads from the pasteboard. Since dynamic links reads from the pasteboard at launch, it will cause this notification. Using the UIPasteboard's new `detectPatternsForPatterns:` api, we can try detecting if the pasteboard contains a dynamic link as its first item so we don't trigger the notification needlessly.

Draft PR following up on https://github.com/firebase/firebase-ios-sdk/issues/5893


FirebaseDynamicLinks/Sources/FIRDLDefaultRetrievalProcessV2.m
```diff
@@ -296,6 +296,32 @@ - (nullable NSURL *)uniqueMatchLinkToCheck {
...
+                completionHandler:^(NSSet<UIPasteboardDetectionPattern> *patternSet,
+                                    NSError *error) {
+                  if ([patternSet containsObject:UIPasteboardDetectionPatternProbableWebURL]) {
+                    pasteboardContents = [UIPasteboard generalPasteboard].string;
+                  }
+                  dispatch_group_leave(pasteboardGroup);
+                }];
+
+  } else {
+    pasteboardContents = [UIPasteboard generalPasteboard].string;
```
ncooke3, line 319: This line alone will trigger the notification to be displayed!

FirebaseDynamicLinks/Sources/FIRDLDefaultRetrievalProcessV2.m
```diff
@@ -296,6 +296,35 @@ - (nullable NSURL *)uniqueMatchLinkToCheck {
...
+    [[UIPasteboard generalPasteboard]
+        detectPatternsForPatterns:detectionPatterns
+                completionHandler:^(NSSet<UIPasteboardDetectionPattern> *patternSet,
+                                    NSError *error) {
+                  if ([patternSet containsObject:UIPasteboardDetectionPatternProbableWebURL]) {
+                    detectedContents = [UIPasteboard generalPasteboard].string;
+                  }
+                  dispatch_semaphore_signal(semaphore);
+                }];
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
eldhosembabu, line 316: wdyt about using a completion handler for retrievePasteboardContents instead of waiting and returning a string from this method?

ncooke3: I think that is a valid approach too but Im not sure how that would change initial call [here](https://github.com/firebase/firebase-ios-sdk/pull/5905/files#diff-b0b41bd29c92c63bfcb04da66a305f35L273). It might require more refactoring for that method, which im not sure if that is the right move considering this proposal PR is not really a fix. Im not too sharp at handling async methods in objc c though so if you can make it cleaner, feel free! 🙂

eldhosembabu: For proposal PR it should be fine. Otherwise as you can see, this method will get called from app start up and we don't want to freeze the app by waiting here to check the pasteboard.

ryanwilson:
This change LGTM - I tested myself that the `hasURL` call works as expected and this should at the very least reduce the notifications to only when a URL is on the clipboard. We should evaluate how we can do better, but this is an improvement to start.

ryanwilson:
Please wait for @eldhosembabu review before merging @ncooke3

eldhosembabu:
Please add the change log entry before merging.

FirebaseDynamicLinks/Sources/FIRDLDefaultRetrievalProcessV2.m
zolivier: > This line alone will trigger the notification to be displayed!

Looks good to me:
```swift
if (@available(iOS 14.0, *)) {
    // iOS 14+
    ...
} else {
    // iOS 13 and below - no pasteboard notification there 🙈 
    pasteboardContents = [UIPasteboard generalPasteboard].string;
}
```

ignazioc: the line 306 will be executed only on iOS < 10 therefore no popup. 

Still the line 303 will trigger a popup, but I guess the point was to just "reduce" the popups  (check the Changelog

PR #5907, opened 2020-06-24
Use UUID as GDTCOREvent.eventID

maksymmalyhin:
- simplified `GDTCOREvent.eventID` generating logic. There were a couple issues uncovered with the previously implemented logic such as blocking a `[GDTCOREvent init..]` caller thread, implicit dependency of the data object on a storage instance and a couple of unhandled error cases that may lead to generating the same IDs which will cause overriding old events. Fixing these issues leads to increasing of the implementation complexity even more. I propose to use the simplified version of the ID generation logic at least temporary to mitigate the release risks of already major and potentially risky changes. I could not find `eventID` to be sent to the server, so I assume we can postpone implementing the incrementally growing eventIDs until it is supported and required.
- FIRCoreDiagnosticsTests fixed.

#no-changelog Changelog is still relevant.


GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -24,60 +24,32 @@
...
-      }
-      setNewValue:^NSData *_Nullable(void) {
-        if (lastEventID != -1) {
-          int32_t incrementedValue = lastEventID + 1;
-          return [NSData dataWithBytes:&incrementedValue length:sizeof(int32_t)];
-        }
-        return nil;
-      }];
++ (NSString *)nextEventID {
+  // TODO: Consider a way to make the eventIDs incremental without introducing a storage dependency
```
paulb777, line 30: There could be one UUID for each app instance that appends a counter to at least get events from the same app instance incremental.

mikehaney24: Are UUID's incremental? The reason why I did it the way I did was because I needed a robust mechanism (robust against an app crashing, say) that was consecutive because it would make metalogging in the future easier. Crashlytics may want some metalogging done, so knowing how many events were generated, their order, and what events made it or didn't was valuable.

maksymmalyhin: No UUID is not incremental. I suggest to implement the incremental feature separately closer to the time it is required so we don't block the current release.

mikehaney24: sgtm

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
```diff
@@ -24,60 +24,32 @@
...
-  [GDTCOREvent nextEventIDForTarget:target
-                         onComplete:^(NSNumber *_Nullable newEventID) {
-                           eventID = newEventID;
-                           dispatch_semaphore_signal(sema);
-                         }];
-  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 5 * NSEC_PER_SEC)) != 0) {
-    return nil;
-  }
-  if (mappingID == nil || mappingID.length == 0 || target <= 0 || eventID == nil) {
+  if (mappingID == nil || mappingID.length == 0 || target <= 0) {
```
mikehaney24, line 38: wouldn't mappingID.length > 0 accomplish the first two check?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -30,11 +30,8 @@
 /** A library data key this class uses to track batchIDs. */
 static NSString *const gBatchIDCounterKey = @"GDTCORFlatFileStorageBatchIDCounter";
 
-/** A library data key this classes uses to track batches and their associated eventIDs. */
-static NSString *const gBatchesToEventsIDsKey = @"GDTCORFlatFileStorageBatchIDCounter";
-
 /** The separator used between metadata elements in filenames. */
-static NSString *const kMetadataSeparator = @"-";
+static NSString *const kMetadataSeparator = @"|";
```
mikehaney24, line 34: Why change this back?

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
maksymmalyhin: Yeah, good idea, I just removed `eventID == nil`, but I'll update this part as well.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: "-" symbol is a part of UUID, so I had to change the separator. Now I think I can replace "-" in the eventID itself if needed. Do you remember issues with "|"?

mikehaney24: In some of the reading I did, some filesystems were sensitive to use of | as a file or directory name. Also, our % escaping only allowed alphanum and "-", make sure to update that in GDTCOREventSelector.

maksymmalyhin: Ah, ok, I'll revert the separator back and modify eventID itself. Thanks!

maksymmalyhin: Done

GoogleDataTransport/GDTCORLibrary/GDTCOREvent.m
maksymmalyhin: Done

PR #5908, opened 2020-06-24
Cleanups for check_imports.swift

paulb777:
Add main func, add ErrorLog class, and eliminate two globals

### DIFF
--- a/scripts/check_imports.swift
+++ b/scripts/check_imports.swift
@@ -56,19 +56,20 @@ let skipImportPatterns = [
   "FBLPromise",
 ]
 
-var foundError = false
-
-func genError(_ message: String) {
-  print(message)
-  foundError = true
+private class ErrorLogger {
+  static var foundError = false
+  class func log(_ message: String) {
+    print(message)
+    foundError = true
+  }
 }
 
-func checkFile(_ file: String, isPublic: Bool) {
+private func checkFile(_ file: String, inRepo repoURL: URL, isPublic: Bool) {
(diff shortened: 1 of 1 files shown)


scripts/check_imports.swift
```diff
@@ -57,19 +57,24 @@ let skipImportPatterns = [
   "FBLPromise",
 ]
 
-var foundError = false
+private class ErrorLogger {
```
wilhuff, line 60: If you wanted you could also pass the error logger to `checkFile`.

The benefit would be that tests could straightforwardly verify that certain inputs resulted in errors and that certain inputs did not without needing to add a way to clear the global error log.

scripts/check_imports.swift
```diff
@@ -121,59 +127,65 @@ func checkFile(_ file: String, isPublic: Bool) {
...
         }
+        checkFile(fullTransformPath, inRepo: repoURL, isPublic: file.range(of: "/Public/") != nil)
       }
-      checkFile(fullTransformPath, isPublic: file.range(of: "/Public/") != nil)
     }
   }
 }
 
-exit(foundError ? 1 : 0)
+main()
```
wilhuff, line 190: perhaps have main return the status code?

scripts/check_imports.swift
paulb777: Done

scripts/check_imports.swift
paulb777: Done

PR #5916, opened 2020-06-25
Implement unified emulator settings API proposal

morganchen12:
The API proposal in question is "Unified Emulator Connection Settings".

Googlers see b/158072394


FirebaseCore/Sources/Public/FIREmulatorSettings.h
```diff
@@ -0,0 +1,74 @@
...
+
+import Foundation
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A string constant representing a Firebase service with an emulator available.
+typedef NSString *FIREmulatorService NS_TYPED_EXTENSIBLE_ENUM NS_SWIFT_NAME(EmulatorSettings);
+
+/** :nodoc: */
+NSString *const FIREmulatorServiceDatabase;
```
samtstern, line 25: In Android we ended up putting these in the service SDKs rather than in the common one.  So instead of:

```java
enum FirebaseEmulators {
  DATABASE
  FIRESTORE
}
```

We did:
```java
FirebaseFirestore.EMULATOR
FirebaseDatabase.EMULATOR
```

The reason we did this was because we didn't want to have to update the common SDK every time a new emulator was added and then release all the other SDKs to depend on common again.

Not sure if that's a concern on iOS as well, just throwing it out there.

samtstern:
@morganchen12 FWIW storage is not in scope, there's no Storage emulator.

Firestore/Source/API/FIRFirestore.mm
```diff
@@ -329,6 +329,23 @@ + (void)enableLogging:(BOOL)logging {
   util::LogSetLevel(logging ? util::kLogLevelDebug : util::kLogLevelNotice);
 }
 
+- (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
+  if (!host.length) {
+    ThrowInvalidArgument("Host cannot be nil or empty.");
+  }
+  if (!_settings.isUsingDefaultHost) {
```
samtstern, line 336: In Android we decided to make this just a warning and let the `useEmulator` value win over what's in the settings.

FirebaseCore/Sources/Public/FirebaseCore/FirebaseCore.h
```diff
@@ -16,5 +16,6 @@
 
 #import "FIRApp.h"
 #import "FIRConfiguration.h"
+#import "FIREmulatorSettings.h"
```
paulb777, line 19: Did you forget to `git add` FIREmulatorSettings.h?

Firestore/Source/API/FIRFirestore.mm
morganchen12: Done.

FirebaseDatabase/Tests/Integration/FIRDatabaseTests.m
```diff
@@ -94,10 +92,7 @@ - (void)testDifferentInstanceForAppWithURL {
 - (void)testDatabaseForAppWithInvalidCustomURLs {
   id app = [[FIRFakeApp alloc] initWithName:@"testDatabaseForAppWithInvalidCustomURLs"
                                         URL:kFirebaseTestAltNamespace];
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wnonnull"
-  XCTAssertThrows([FIRDatabase databaseForApp:app URL:nil]);
-#pragma clang diagnostic pop
+  XCTAssertThrows([FIRDatabase databaseForApp:app URL:(NSString * _Nonnull) nil]);
```
paulb777, line 95: Nice trick!

paulb777 (commented):
LGTM, deferring iCore approval to @ryanwilson for API change approval

paulb777:
Please fix the Functions travis failure

Firestore/Example/Tests/API/FIRFirestoreTests.mm
```diff
@@ -63,4 +63,16 @@ - (void)testDeleteApp {
...
 
+- (void)testSetEmulatorSettingsSetsHost {
+  // Ensure the app is set appropriately.
+  FIRApp *app = testutil::AppForUnitTesting();
+
+  FIRFirestore *firestore = [FIRFirestore firestoreForApp:app];
+
+  [firestore useEmulatorWithHost:@"localhost" port:1000];
+
+  NSString *host = [firestore valueForKeyPath:@"settings.host"];
```
wilhuff, line 74: Given that these are all properties, this seems like it could be more more concisely spelled as `firestore.settings.host`. This has the benefit of making this show up when when searching for usages.

wilhuff (approved):
Firestore LGTM

Firestore/Source/Public/FirebaseFirestore/FIRFirestore.h
```diff
@@ -163,6 +163,12 @@ NS_SWIFT_NAME(Firestore)
 
 #pragma mark - Network
 
+/**
+ * Configures Firestore to connect to an emulated host instead of the default remote backend. Once
+ * Firestore is initialized, this value cannot be changed.
```
ryanwilson, line 168: What's the startup lifecycle like here? This method must be called after `Firestore.firestore()` is called, which is what I would think means "Firestore is initialized" at that point. FWIW I don't have an issue with the method itself, mostly the docs.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
```diff
@@ -66,7 +66,17 @@ NS_SWIFT_NAME(Functions)
  * See https://firebase.google.com/docs/functions/local-emulator
  * @param origin The origin of the local emulator, such as "http://localhost:5005".
  */
-- (void)useFunctionsEmulatorOrigin:(NSString *)origin NS_SWIFT_NAME(useFunctionsEmulator(origin:));
+- (void)useFunctionsEmulatorOrigin:(NSString *)origin
+    NS_SWIFT_NAME(useFunctionsEmulator(origin:))
+        __attribute__((deprecated("Use setEmulator(host:port:) instead.")));
```
ryanwilson, line 71: `useEmulator` not `setEmulator`

Firestore/Source/Public/FirebaseFirestore/FIRFirestore.h
samtstern: On other platforms we say "after Firestore has been used" or similar.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +173,17 @@ - (void)purgeOutstandingWrites {
     });
 }
 
+- (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
```
samtstern, line 176: (commenting here because I can't comment higher)

Make sure the following sequence works:

1. Get database instance for database `foo.firebaseio.com`
2. Call `useEmulator`
3. Call `refFromUrl('https://foo.firebaseio.com/foo/bar/baz')`

I believe that will currently throw, but it should actually return an emulated ref with path `/foo/bar/baz`

cc @schmidt-sebastian to confirm

FirebaseDatabase/Tests/Integration/FIRDatabaseTests.m
```diff
@@ -469,6 +464,16 @@ - (void)testCallbackQueue {
   [database goOffline];
 }
 
+- (void)testSetEmulatorSettings {
```
samtstern, line 467: Also add a test showing that this doesn't throw in a normal situation and that the `ref` you get back after `useEmulator` points at the emulator.

Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -131,7 +131,13 @@ - (instancetype)initWithProjectID:(NSString *)projectID
 }
 
 - (void)useLocalhost {
-  [self useFunctionsEmulatorOrigin:@"http://localhost:5005"];
+  [self useEmulatorWithHost:@"http://localhost" port:5005];
```
samtstern, line 134: Host should not contain `http://` ... unless I am missing something about this Functions API

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
```diff
@@ -66,7 +66,17 @@ NS_SWIFT_NAME(Functions)
...
+
+/**
+ * Changes this instance to point to a Cloud Functions emulator running locally.
+ * See https://firebase.google.com/docs/functions/local-emulator
+ * @param host The host of the local emulator, such as "http://localhost".
```
samtstern, line 76: In all cases `host` should *not* include the `http://` or any other prefix.  We need to add that internally (and `https://` is not allowed)

Firestore/Example/Tests/API/FIRFirestoreTests.mm
morganchen12: Done.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
morganchen12: Done.

Firestore/Source/Public/FirebaseFirestore/FIRFirestore.h
morganchen12: Amended wording for better clarity.

FirebaseDatabase/Sources/Api/FIRDatabase.m
schmidt-sebastian: We discussed this in the Emulator chat. `refFromUrl` should be updated to only throw for production database URLs (those that end in `.firebaseio.com`).

schmidt-sebastian: After more discussion, we may not validate that the hosts matches at all. Let's wait for Sam to turn this into code before updating this PR.

Functions/FirebaseFunctions/FIRFunctions.m
morganchen12: Removed.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
morganchen12: Removed.

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: SGTM, I'll hold off on new changes to this PR for now.

samtstern: JS is all merged now if you want to copy that implementation.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +173,17 @@ - (void)purgeOutstandingWrites {
...
+    NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
+    NSAssert(self.repo == nil,
+             @"Cannot connect to emulator after database initialization. Call "
+             @"useEmulator(host:port:) before creating a database reference or "
+             @"opening a database connection.");
+    NSString *fullHost =
+        [NSString stringWithFormat:@"%@:%li", host, (long)port];
+    self.repoInfo.host = fullHost;
+}
+
```
schmidt-sebastian, line 186: Can you also port Sam's change to https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDatabase/Sources/Api/FIRDatabase.m#L156? We decided to only validate the host in `referenceFromUrl` if the underlying URL does not point to a custom host. This would allow a user to use production URLs in `referenceFromUrl` even if the use an emulator.

morganchen12: Done.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +173,17 @@ - (void)purgeOutstandingWrites {
     });
 }
 
+- (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
+    NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
```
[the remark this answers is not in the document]
morganchen12: Done.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +173,17 @@ - (void)purgeOutstandingWrites {
...
+    NSAssert(self.repo == nil,
```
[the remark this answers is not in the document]
morganchen12: Done. Also changed the wording

morganchen12:
Removing Ryan since there are no more Core changes.

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: Done. This flow is now described in `testEmulatedDatabaseValidatesOnlyNonCustomURLs` in `FIRDatabaseTests.m`.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -153,13 +153,32 @@ - (FIRDatabaseReference *)referenceFromURL:(NSString *)databaseUrl {
...
-             raise:@"InvalidDatabaseURL"
-            format:
-                @"Invalid URL (%@) passed to getReference(). URL was expected "
-                 "to match configured Database URL: %@",
-                databaseUrl, [self reference].URL];
+
+    // If in an emulated setting, only validate the host if the underlying
+    // URL isn't a custom host. This allows users to create references from
+    // production URLs in an emulated setting.
+    if (self.repoInfo.underlyingHost != nil) {
```
schmidt-sebastian, line 160: What are your thoughts on matching the newest version of the Android implementation (the one that was merged)? This would allow you to get rid of `underlyingHost` and we would just validate hosts if `_repoInfo.isCustomHost` is false.

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +192,26 @@ - (void)purgeOutstandingWrites {
     });
 }
 
+- (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
+    if (host.length == 0) {
+        [NSException raise:NSInvalidArgumentException
+                    format:@"Cannot connect to nil or empty host."];
+    }
+    if (self.repo != nil) {
+        [NSException raise:NSInternalInconsistencyException
```
schmidt-sebastian, line 201: Do you mind re-wrapping this to match the longer line length in this file?

FirebaseDatabase/Sources/Api/FIRDatabase.m
```diff
@@ -173,6 +192,26 @@ - (void)purgeOutstandingWrites {
...
+                    format:@"Cannot connect to emulator after database "
+                           @"initialization. Call "
+                           @"useEmulator(host:port:) before creating a "
+                           @"database reference or "
+                           @"trying to load data."];
+    }
+    NSString *fullHost =
+        [NSString stringWithFormat:@"%@:%li", host, (long)port];
+    FRepoInfo *newInfo = [[FRepoInfo alloc] initWithInfo:self.repoInfo
```
schmidt-sebastian, line 210: s/newInfo/emulatorInfo ?

FirebaseDatabase/Sources/Core/FRepoInfo.h
```diff
@@ -16,25 +16,48 @@
...
+
 @interface FRepoInfo : NSObject <NSCopying>
 
-@property(nonatomic, readonly, strong) NSString *host;
-@property(nonatomic, readonly, strong) NSString *namespace;
-@property(nonatomic, strong) NSString *internalHost;
-@property(nonatomic, readonly) bool secure;
+/// Stores the original host if the host has been set after initialization.
+/// Used to validate references.
+@property(nonatomic, readwrite, copy, nullable) NSString *underlyingHost;
```
schmidt-sebastian, line 25: I think you don't need this. If you do keep this, we may need some API wrapper so that we don't have `underlyingHost`, `host` and `internalHost` all in the same class.

FirebaseDatabase/Sources/Core/FRepoInfo.m
```diff
@@ -25,61 +25,94 @@ @interface FRepoInfo ()
...
+    // Compute the domain based on the original host instead of the emulated
+    // host in an emulated setting.
+    if (self != nil) {
+        _domain =
+            [_underlyingHost containsString:@"."]
+                ? [_underlyingHost
+                      substringFromIndex:[_underlyingHost rangeOfString:@"."]
+                                             .location +
+                                         1]
+                : _underlyingHost;
```
schmidt-sebastian, line 90: This also seems missing in the other ports. Does `localhost` now work for `_domain`?

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: Would `internalHost` be suitable for validation purposes? I added `underlyingHost` because `internalHost` is mutable and I wasn't sure if it would be appropriate for validating reference URLs.

schmidt-sebastian: s/Android/JS (sorry I thought this was done on Android already, but looks like it wasn't yet ported back from JS).

See here: https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/database/src/api/Database.ts;l=160

This removes the need to check the underlying host and reduces error conditions.

morganchen12: Wouldn't this change the behavior by skipping validation for non-emulated environments?

schmidt-sebastian: It should be the other way around. If we are using an emulated URL, we don't do any validation.

FirebaseDatabase/Sources/Core/FRepoInfo.m
morganchen12: I did this because the existing implementation of `isCustomHost` depended on `domain`, which was derived from `host`. `localhost` works fine for `domain`, but setting that in the initializer makes the database instance consider the emulator to be a custom domain.

schmidt-sebastian: After talking to @yuchenshi, I think we want the SDK to consider the emulator URLs as a custom domain. We essentially want to change all internal URLs to "localhost" but still allow production URLs in "refFromUrl". We can talk about this more offline (and still change our plan if need be).

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: Done.

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: Copied the ts implementation.

FirebaseDatabase/Sources/Api/FIRDatabase.m
morganchen12: Done.

FirebaseDatabase/Sources/Core/FRepoInfo.h
morganchen12: Removed.

FirebaseDatabase/Tests/Integration/FIRDatabaseTests.m
```diff
@@ -469,6 +464,65 @@ - (void)testCallbackQueue {
...
+
+  // Test emulated url
+  referenceURLString = @"http://localhost:1111/path";
+  reference = [database referenceFromURL:referenceURLString];
+  XCTAssertNotNil(reference);
+  XCTAssert([reference.URL containsString:@"localhost:1111"]);
+
+  // Test non-custom url with different host throws exception
+  referenceURLString = @"https://test.firebaseio.com/path";
+  XCTAssertThrows([database referenceFromURL:referenceURLString]);
```
schmidt-sebastian, line 523: Sorry, I am confused why this test would still throw here. Can you explain?

morganchen12: The URL is not a custom host, so it gets validated, but it's invalid because it's a different host than the expected host. This should match the behavior here: https://cs.opensource.google/firebase-sdk/firebase-js-sdk/+/master:packages/database/src/api/Database.ts;l=160?q=database.ts

FirebaseDatabase/Tests/Integration/FIRDatabaseTests.m
morganchen12: Done.

FirebaseDatabase/Tests/Integration/FIRDatabaseTests.m
schmidt-sebastian: Sorry, I thought it was the other way around. LG.

schmidt-sebastian (approved):

paulb777 (approved):

morganchen12:
Quickstart test will be fixed by https://github.com/firebase/quickstart-ios/pull/1075. @schmidt-sebastian, do you know why the integration test is failing?

paulb777:
Could be more issues related to the GHA Xcode 12 transition. We should get the full summary from tonight's cron run.

morganchen12:
@paulb777 do you know why the quickstart test is failing with

```
Library not loaded: @rpath/FirebaseABTesting.framework/FirebaseABTesting
```

when ABTesting isn't in the list of dependencies? I wasn't able to reproduce the failure locally.

paulb777:
@morganchen12 It's a result of [this script ](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/setup_quickstart.sh#L42) that makes sure that all Firebase dependency pods are sourced from the repo.

PR #5917, opened 2020-06-25
Fix Import of FirebaseInstallationsInternal to use public header

samedson:
- It seems the Private FirebaseInstallationsInternal.h header is gone
 - I think the methods we use moved to the public header for Crashlytics
 - This updates all references, but I haven't tested the other SDKs that I changed in this PR

#no-changelog

### DIFF
--- a/Crashlytics/Crashlytics/FIRCrashlytics.m
+++ b/Crashlytics/Crashlytics/FIRCrashlytics.m
@@ -42,7 +42,7 @@
 #import "FIRCLSReportManager.h"
 
 #import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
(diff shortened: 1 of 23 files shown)


samedson:
Huh seems this didn't fix the problem... 

I'm not getting the FirebaseInstallations folder in the FirebaseCrashlytics pod:

Screen Shot 2020-06-25 at 4 19 11 PM

samedson:
Ah was missing "Library" in the path. Crashlytics works in my test app now point at this commit `b549554`

samedson:
@paulb777 so I thought that at first but I don't see that folder there.

Edit: ah it's in library too: https://github.com/firebase/firebase-ios-sdk/tree/master/FirebaseInstallations/Source/Library

paulb777:
I made #5919 that should fix

PR #5920, opened 2020-06-25
Avoid deadlock during shutdown on Linux/Windows

wilhuff:
Under some circumstances, the Executor can be accessed by a user task that's running during shutdown. The specific case I observed was if a task that manipulated an idle timer happened to run during shutdown, it would cause a deadlock because the Executor was still holding its lock while trying to join the user threads.


wilhuff:
Fix for the TSAN violation (that's unrelated to this change) is in #5922.

PR #5922, opened 2020-06-25
Fix TSAN violations in FSTTransactionTests

wilhuff:
TSAN uncovered cases where we were reading values in the transaction
thread that were set on the main thread and were unprotected by lock. In
practice this is unlikely to cause an harm, but it's good to keep TSAN
clean for when there are real issues.


Firestore/Example/Tests/Integration/FSTTransactionTests.mm
```diff
@@ -141,17 +142,15 @@ - (instancetype)initWithDb:(FIRFirestore *)db testCase:(FSTTransactionTests *)te
 }
 
 - (FSTTransactionTester *)withExistingDoc {
-  _fromExistingDoc = YES;
-  return self;
+  self.fromExistingDoc = YES;
```
thebrianchen, line 145: add `return self` here and below.

PR #5923, opened 2020-06-25
Fix header style issue caught by copybara

paulb777:
#no-changelog


paulb777:
Verified build fixed in cl/318370096

PR #5933, opened 2020-06-26
Update formatters

paulb777:
Update to clang-format 10.0.0 and swiftformat 0.44.14

brew has stopped supporting installing from git tags - failing with 
```
Warning: Calling Non-checksummed download of clang-format formula file from an arbitrary URL is deprecated! Use 'brew extract' or 'brew create' and 'brew tap-new' to create a formula file in a tap on GitHub instead.
######################################################################## 100.0%
Error: clang-format not installed
```

TBD: find a way to lock versions so that brew version updates don't break CI

### DIFF
--- a/scripts/setup_check.sh
+++ b/scripts/setup_check.sh
@@ -21,8 +21,11 @@ set -euo pipefail
 export HOMEBREW_NO_ANALYTICS=1
 export HOMEBREW_NO_AUTO_UPDATE=1
 
-brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/c6f1cbd/Formula/clang-format.rb
-brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/c13eda8/Formula/swiftformat.rb
+brew install clang-format
+brew install swiftformat
+
+clang-format -version
+swiftformat -version
 
 pip install flake8
 pip install six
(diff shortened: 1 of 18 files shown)


scripts/setup_check.sh
```diff
@@ -21,8 +21,11 @@ set -euo pipefail
 export HOMEBREW_NO_ANALYTICS=1
 export HOMEBREW_NO_AUTO_UPDATE=1
 
-brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/c6f1cbd/Formula/clang-format.rb
-brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/c13eda8/Formula/swiftformat.rb
+brew install clang-format
```
maksymmalyhin, line 24: @paulb777  If we don't specify a particular version, then PRs will have to include formatting changes when a new version of formatter is released. It may be annoying and pollute PRs with irrelevant changes. I would prefer to keep a version number specifier to avoid this. 
WDYT?

paulb777: @maksymmalyhin I share the preference, but the old way to lock versions no longer works and I can't find another solution. If we can come up with a way, I'm happy to update the CI and instructions.

In the meantime, we need to unlock new contributors who don't have a way to install the old version - see the comment in the header.

maksymmalyhin: Ah, and you even explained it in the PR description, sorry for the silly comment.

PR #5944, opened 2020-06-29
Update copyright presubmit and fix 2020 copyrights

paulb777:
Verify that new copyright syntax with "Google LLC" is used.

Googlers see go/copyright


paulb777:
cc: @visumickey

paulb777:
@morganchen12 Sorry I missed the comment before merging.  All impacted files already have changelogs for M75 so I think we're fine to skip noting this one.

PR #5946, opened 2020-06-29
Adding plist property 'FirebaseDeepLinkPasteboardRetrievalEnabled'

eldhosembabu:
Adding plist property"FirebaseDeepLinkPasteBoardRetrievalEnabled" to enable/disable fetching dynamic links from PasteBoard.

With iOS 14, when ever the PasteBoard contents are accessed, a pop up will be displayed. So for FDL clients who doesn't need the deeplinking based on PasteBoard contained dynamic link, can use this property to disable PasteBoard access there by not having a PastBoard access system popup.

### DIFF
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,8 +1,6 @@
-# Unversioned
-- Adding plist property"FirebaseDeepLinkPasteBoardRetrievalEnabled" to enable/disable fetching dynamic links from PasteBoard.
+# v4.2.0
+- [Added] Plist property"FirebaseDeepLinkPasteBoardRetrievalEnabled" to enable/disable fetching dynamic links from PasteBoard.
 - [fixed] Reduce frequency of iOS14 pasteboard notifications by only reading from it when it contains URL(s). (#5905)
-
-# v4.3.0
 - [changed] Functionally neutral updated import references for dependencies. (#5824)
 
 # v4.1.0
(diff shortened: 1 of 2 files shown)


FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unversioned
+- Adding plist property"FirebaseDeepLinkPasteBoardRetrievalEnabled" to enable/disable fetching dynamic links from PasteBoard.
```
paulb777, line 2: Two issues here. Neither directly related to this PR:
1. M74 was deferred to M75 after the CHANGELOG had been updated for it.
2. The current release is 4.1, so the next release should be 4.2.0 and include these top three changes.

eldhosembabu: Done...Please check whether i did it correct.

FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,7 +1,6 @@
-# Unversioned
+# v4.2.0
+- [Added] Plist property"FirebaseDeepLinkPasteBoardRetrievalEnabled" to enable/disable fetching dynamic links from PasteBoard.
```
paulb777, line 2: Organization looks good.

From my read of docs, the "B" in pasteboard should be lowercase here and in the implementation.

Use a  backtick ``` instead of `"` around the property name.

eldhosembabu: Made changes...Please check whether I'm using the right one for "backtick"

PR #5952, opened 2020-06-30
FDL Sample App: Fixing imports and Pod file 

eldhosembabu:
Current state of sample app has Import issues and the pod generation was failing to resolve GoogleUtilities dependency.

- Changed to use proper import
- Added GoogleUtilities dependency in pod file to make the pod install working


eldhosembabu:
Styling is fixed

PR #5957, opened 2020-07-01
Update RemoteConfig workflow.

granluo:
[RemoteConfig quickstart](https://github.com/firebase/quickstart-ios/tree/master/config) is updated and objc quickstart is removed.


granluo:
quickstart framework/quickstart tests are done in 6c14b01.

PR #5958, opened 2020-07-01
Test new add framework tool.

granluo:


granluo:
This is to test the new add_framework tool in the workflow. Will add to quickstart-ios repo after the test passed.
This change is due to some files, e.g. libc++.dylib, not able to copy to a folder and import.

PR #5959, opened 2020-07-01
SPM: Core ObjC unit tests

paulb777:
First iOS unit test runs from Swift Package Manager for FirebaseCore

- Update to swift-tools-version 5.3
  - Add excludes for new 5.3 complete directory structure coverage
- Move FirebaseCore SwiftUnit tests to their own directory to satisfy SPM build constraints
- Update FirebaseCore tests depending on an app's GoogleService-Info.plist to use a mock and skip three of the tests
  - Details about the Swift PM constraints in https://forums.swift.org/t/finding-swift-package-manager-resources/38058
- Updated SPM GHA workflow to use Xcode 12 and run the Core unit tests


paulb777:
@ryanwilson This PR should be latest and greatest for Swift Package Manager source building.

The CoreUnit scheme is found for me locally, but doesn't show up in GHA. 

See the workflow script for what is working locally.

.github/workflows/spm.yml
```diff
@@ -17,7 +17,13 @@ jobs:
...
     - name: Build
       run: swift build
-    - name: Run
-      run: swift test
+    - name: Initialize xcodebuild
+      run: xcodebuild -list
+    - name: Core ObjC Unit Tests macOS
+      run: xcodebuild -scheme CoreUnit test
+    - name: Core ObjC Unit Tests iOS
+      run: xcodebuild -scheme CoreUnit test -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 11'
```
ryanwilson, line 29: We can change this to `xcodebuild -scheme Firebase-Package test ...` and it'll run all `testTarget`s specified in the manifest. Maybe that's enough to unblock it for now and we run less frequently or on changes to the manifest?

Then once we hear more about running individual targets we can split it up accordingly.

paulb777: Done. We can defer figuring out figuring out testing individual targets.

paulb777:
Added a card to https://github.com/firebase/firebase-ios-sdk/projects/1 for updating disabled tests.

PR #5961, opened 2020-07-01
Have ListenerRegistration::Remove block until callbacks are complete

wilhuff:
This fixes a test flake observed very rarely in internal testing, where the callback starts processing, checks to see if it's muted, sees that it's not, and proceeds; meanwhile the test calls remove, and deletes state to which the listener was referring and by the time the callback happens in user code the listener callback now refers to the deleted state.

### DIFF
--- a/Firestore/core/src/api/query_core.cc
+++ b/Firestore/core/src/api/query_core.cc
@@ -163,7 +163,7 @@ std::unique_ptr<ListenerRegistration> Query::AddSnapshotListener(
       QuerySnapshot result(firestore_, query_, std::move(snapshot),
                            std::move(metadata));
 
-      user_listener_->OnEvent(result);
+      user_listener_->OnEvent(std::move(result));
     }
 
    private:
--- a/Firestore/core/src/core/event_listener.h
+++ b/Firestore/core/src/core/event_listener.h
@@ -17,8 +17,8 @@
 #ifndef FIRESTORE_CORE_SRC_CORE_EVENT_LISTENER_H_
 #define FIRESTORE_CORE_SRC_CORE_EVENT_LISTENER_H_
 
-#include <atomic>
 #include <memory>
+#include <mutex>  // NOLINT(build/c++11)
 #include <utility>
 
 #include "Firestore/core/src/util/executor.h"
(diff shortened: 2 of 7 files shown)


Firestore/core/src/api/query_core.cc
```diff
@@ -163,7 +163,7 @@ std::unique_ptr<ListenerRegistration> Query::AddSnapshotListener(
       QuerySnapshot result(firestore_, query_, std::move(snapshot),
                            std::move(metadata));
 
-      user_listener_->OnEvent(result);
+      user_listener_->OnEvent(std::move(result));
```
dconeybe, line 166: Is this actually related to the fix? Or is this just an optimization you discovered while implementing this fix?

Firestore/core/src/core/event_listener.h
```diff
@@ -86,7 +83,19 @@ class AsyncEventListener
...
+  // PORTING NOTE: Android uses a volatile here but that's not enough in C++.
+  //
+  // In C++, the user can call `ListenerRegistration::Remove` (which calls
+  // `Mute`) and then immediately delete the state backing the listener. Using
+  // a mutex here instead of an atomic ensures that `Mute` won't return until
+  // it's safe to delete the state backing a listener. In Java this is safe
+  // because the state backing the listener is garbage collected so it doesn't
+  // matter if the mute is concurrent with a callback.
+  //
+  // Use a recursive mutex instead of `std::mutex` to avoid deadlock in the case
```
dconeybe, line 95: Is there a test case dedicated to verifying no deadlock if Remove() is called by a callback? I didn't see one, but I may have missed it. I realize that you exercise this code path in the new unit test you added but it should have a dedicated unit test being that it would be a "guarantee" that would be easy to break unintentionally.

dconeybe:
I mean to say also, the new test case you added is very elegant. It took a bit of brain power to follow its logic, but it is beautiful once understood!

Firestore/core/src/core/event_listener.h
wilhuff: I've added a test.

PR #5963, opened 2020-07-02
Add Auth Swift Framework Test.

granluo:


granluo:
Test is done in bc507d6

.github/workflows/auth.yml
```diff
@@ -76,6 +77,43 @@ jobs:
       # Only build the unit tests on Catalyst
       run: scripts/test_catalyst.sh FirebaseAuth build FirebaseAuth-Unit-unit
 
+  quickstart_framework:
+    env:
```
paulb777, line 81: Missing

`    # Don't run on private repo.
    if: github.event_name == 'schedule' && github.repository != 'FirebasePrivate/firebase-ios-sdk'`

Also looks like a problem in Storage, DynamicLinks and Database. The scheduled jobs on https://github.com/FirebasePrivate/firebase-ios-sdk/actions should be over in seconds.

granluo: Will add that in the pr.
Looks like I have no access to FirebasePrivate. I cannot get into the link.

granluo: Will update those three in the pr as well.

granluo: Just made another [pr](https://github.com/firebase/firebase-ios-sdk/pull/5968) 😅  to resolve this.

Issue #5969, opened 2020-07-03
Deadlock when app launches in background in GULHeartbeatDateStorage


zacwest:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.5
  * Firebase SDK version: 6.27.0
  * Firebase Component: FirebaseCoreDiagnostics
  * Component version: 1.3?
  * Installation method: CocoaPods

### [REQUIRED] Step 2: Describe the problem

Users are experiencing crashes on startup where it appears a file coordinator around analytics is deadlocking.

Thread 0:

```
Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: Namespace SPRINGBOARD, Code 0x8badf00d
Termination Description: SPRINGBOARD, scene-create watchdog transgression: application<io.robbie.HomeAssistant>:9544 exhausted real (wall clock) time allowance of 29.20 seconds | ProcessVisibility: Background | ProcessState: Suspended | WatchdogEvent: scene-create | WatchdogVisibility: Background | WatchdogCPUStatistics: ( | "Elapsed total CPU time (seconds): 23.390 (user 23.390, system 0.000), 13% CPU", | "Elapsed application CPU time (seconds): 0.171, 0% CPU" | )
Triggered by Thread:  0

Thread 0 name:  Dispatch queue: com.apple.main-thread
Thread 0 Crashed:
0   libsystem_kernel.dylib        	0x00000001a19cc7c0 semaphore_wait_trap + 8
1   libdispatch.dylib             	0x00000001a18a8880 _dispatch_sema4_wait + 28
2   libdispatch.dylib             	0x00000001a18a8f18 _dispatch_semaphore_wait_slow + 132
3   Foundation                    	0x00000001a1f44aa0 -[NSFileCoordinator+ 547488 (NSPrivate) _blockOnAccessClaim:withAccessArbiter:] + 388
4   Foundation                    	0x00000001a1f2ba14 -[NSFileCoordinator+ 444948 (NSPrivate) _coordinateWritingItemAtURL:options:error:byAccessor:] + 632
5   Foundation                    	0x00000001a1f2b774 -[NSFileCoordinator coordinateWritingItemAtURL:options:error:byAccessor:] + 108
[...]

PR #5977, opened 2020-07-06
Access filesystem outside of GULHeartbeatDateStorage initializer

ncooke3:
Currently, when a `GULHeartbeatDateStorage` instance is created, the file system is accessed within the initializer. This might be causing a problem seen in #5969 . Potential fix would be accessing the filesystem somewhere after initialization, in this case, via having a lazy getter for `GULHeartbeatDateStorage`'s `fileURL` property.


GoogleUtilities/Environment/GULHeartbeatDateStorage.m
```diff
@@ -32,11 +36,21 @@ - (instancetype)initWithFileName:(NSString *)fileName {
...
+- (NSURL *)fileURL {
+  if (!_fileURL) {
     NSURL *directoryURL = [[self class] directoryPathURL];
     [[self class] checkAndCreateDirectory:directoryURL fileCoordinator:_fileCoordinator];
-    _fileURL = [directoryURL URLByAppendingPathComponent:fileName];
+    _fileURL = [directoryURL URLByAppendingPathComponent:_fileName];
   }
-  return self;
+  return _fileURL;
 }
```
ncooke3, line 54: I went back and forth with what would be an ideal way to instantiate the `fileURL` property.  What do you think @maksymmalyhin ?

ncooke3:
@maksymmalyhin  For sure! I'll work on some. And do you think there is a better alternative to the lazy getter? I was thinking that it may lead to issues if additions are made to this file and the property is referenced with a `_fileURL` rather than a `self.fileURL`, leading to it never being instantiated. The only other alternative I could think of though was calling a helper method to ensure initialization of the property before it is used... and that might not be very clear

GoogleUtilities/Environment/GULHeartbeatDateStorage.m
```diff
@@ -20,10 +20,14 @@
 @interface GULHeartbeatDateStorage ()
 /** The storage to store the date of the last sent heartbeat. */
 @property(nonatomic, readonly) NSFileCoordinator *fileCoordinator;
+/** The name of the file that stores heartbeat information. */
+@property(nonatomic, readonly) NSString *fileName;
 @end
 
 @implementation GULHeartbeatDateStorage
 
+@synthesize fileURL = _fileURL;
```
maksymmalyhin, line 29: nit: Is it required? We usually use implicit `synthesize` unless it's not possible or lead to errors.

maksymmalyhin:
> @maksymmalyhin For sure! I'll work on some. And do you think there is a better alternative to the lazy getter? I was thinking that it may lead to issues if additions are made to this file and the property is referenced with a `_fileURL` rather than a `self.fileURL`, leading to it never being instantiated. The only other alternative I could think of though was calling a helper method to ensure initialization of the property before it is used... and that might not be very clear

I cannot come up with a better way than the lazy property.  It's much better than calling a helper method. 

To help catching `_fileURL` usage you can update the tests to make sure the directory is removed at the end of each test. Then potential new feature tests using `_fileURL` directly will fail because no directory would be created.

GoogleUtilities/Environment/GULHeartbeatDateStorage.m
ncooke3: Yes, since the property is `readonly` and we implement it's only required accessor (the getter), it isn't auto synthesized so we need to add an implicit `synthesize`.

maksymmalyhin: Is it related to overriding the getter of a readonly property? Interesting, I didn't know it's the case.

ncooke3: Yup, I found this in apple's objc docs about [properties](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html):

>Note: The compiler will automatically synthesize an instance variable in all situations where it’s also synthesizing at least one accessor method. If you implement both a getter and a setter for a readwrite property, or a getter for a readonly property, the compiler will assume that you are taking control over the property implementation and won’t synthesize an instance variable automatically.

I never knew that was the reason but it makes sense I guess!

GoogleUtilities/Example/Tests/Environment/GULHeartbeatDateStorageTest.m
```diff
@@ -18,10 +18,11 @@
 #import "GoogleUtilities/Environment/Private/GULHeartbeatDateStorage.h"
 
 @interface GULHeartbeatDateStorageTest : XCTestCase
-@property(nonatomic) NSURL *fileURL;
 @property(nonatomic) GULHeartbeatDateStorage *storage;
 @end
 
+NSString *const GULStorageHeartbeatTestFileName = @"GULStorageHeartbeatTest";
```
maksymmalyhin, line 24: Please add  prefix `k`. Also if you convert it to `static`,  then the name may be short e.g. 
`static NSString *const kTestFileName`

ncooke3: i like that better. thanks!

PR #5985, opened 2020-07-07
Failure to load zip will fail workflows.

granluo:


granluo:
Failure is expected in 1f090cd. 
The change will fail the workflow in the step "Pull zip from GCS"

PR #5992, opened 2020-07-07

chliangGoogle:
- Move checkin logic from IID to Messaging (FIRMessagingAuth* FIRMessagingCheckin*)
- Move Token logic from IID to Messaging (FIRMessagingToken*, all the token operations files)
- Merge MessagingCode and Utilities between IID and Messaging.
- I have not added unit tests from IID yet to keep this PR not getting bigger.
- All IID code is under token/ folder.
- FIRMessagingTokenManager serves as the central object that holds most of the token information. It is the communication between messaging and previous IID classes.
- All the persistent storage path are kept the same to ensure IID still works with Messaging if users are still using InstanceID.
- Kept all the communications between IID and Messaging in case users still use IID API to handle token and apns info. (APNS used to be in iid)


chliangGoogle:
@googlebot I fixed it.

PR #5993, opened 2020-07-07
Check for Pending Dynamic link guard check logic change

eldhosembabu:
Check for pending DL guard check logic change.

Check for Pending DL check was guarded based on the flag `kFIRDLReadDeepLinkAfterInstallKey` which will get set when developers calls `dynamicLinkFromCustomSchemeURL` method from AppDelegate once they receive callback from FDL SDK with URL. But in some cases where developers will never call this `dynamicLinkFromCustomSchemeURL` method once they receive the URL and they will directly consume it without creating a dynamic link out of it and which internally affects the flag `kFIRDLReadDeepLinkAfterInstallKey` being not set. So in those case when user do a cold restart of the app, since the `kFIRDLReadDeepLinkAfterInstallKey` is not set, we will end up checking for pending dynamic link again.

Once we do check for pending DL, we are setting another flag `kFIRDLOpenURLKey` to mark the attribution api had been called once (regardless of failures). So in this PR, we will check `kFIRDLOpenURLKey` along with `kFIRDLReadDeepLinkAfterInstallKey` to make sure whether we should be needing a pending dl check. We cannot simply relay on `kFIRDLOpenURLKey` for this check since sometimes `kFIRDLReadDeepLinkAfterInstallKey` is set from `setUpWithLaunchOptions` and in that case also we don't need to check for pending DL.


FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -54,7 +54,7 @@
     @"com.google.appinvite.readDeeplinkAfterInstall";
 
 // We should only open url once. We use the following key to store the state in the user defaults.
-static NSString *const kFIRDLOpenURLKey = @"com.google.appinvite.openURL";
```
paulb777, line 57: Why remove the `static`?  There's no reason to pollute the app's namespace unless necessary.

eldhosembabu: reverted

eldhosembabu:
> Please update the Changelog

Done

FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.2.1
+- [fixed]Check for Pending Dynamic link guard check logic
```
paulb777, line 2: Does this fix an existing GitHub issue? If so please add the number.

Casing fixes:
`[fixed] Check for pending dynamic Link guard check logic.`

eldhosembabu: This is an addon fix to PR https://github.com/firebase/firebase-ios-sdk/pull/5665

PR #6009, opened 2020-07-09
Remote Config Console API for Testing

ncooke3:
Adding a testing API (and some tests to showcase it) for interacting with the Remote Config console. Included a README with instructions on how to run the added tests and play with the testing API. With this addition, we could run tests to test remote config's implementation using the fake console API when we don't need to test with a live console, and this new api when we do want to test with a console.

#no-changelog


FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -18,10 +18,24 @@ import FirebaseCore
...
   override func setUp() {
     super.setUp()
     if APITests.useFakeConfig {
       fakeConsole.config = ["Key1": "Value1"]
+    } else {
+      console = RemoteConfigConsole()
+      console.updateRemoteConfigValue("Obi-Wan", for: "Jedi")
+      console.removeRemoteConfigValue(for: "Sith_Lord")
+    }
+  }
```
ncooke3, line 32: @paulb777 Since we have a few cases here (FakeConsole, RemoteConfigConsoleAPI, and regular Remote Config API calls), I was thinking I need to either organize it better so only the tests affiliated with each case run or just separate this file into separate ones based on how we are testing (fake, real, regular API). Happy to do either but I'd appreciate some input

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,244 @@
...
+  public func clearRemoteConfig() {
+    var updatedConfig: [String: Any]! = latestConfig
+    updatedConfig.removeValue(forKey: "parameters")
+    publish(config: updatedConfig)
+  }
+
+  // Other possible API's
+  public func currentRemoteConfigValue(for key: String) -> Any { "" }
+
+  public func validateRemoteConfig(with values: [String: Any]) -> Bool { false }
```
ncooke3, line 130: I was thinking about some other API's that'd be useful for testing. Like the rest, these two would be synchronous. Since remote config can contain JSON values instead of simple primitive key value pairs, I was thinking of a  `validateRemoteConfig` that takes in a dictionary representing a JSON remote config value. Then validates that against the current remote config JSON value stored on the console.

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
paulb777: Maybe it's ok to drop the regular Remote Config API test calls? The RemoteConfigConsoleAPI is a superset that is still talking to a real backend. Right?

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -138,6 +152,159 @@ class APITests: APITestBase {
     waitForExpectations()
   }
 
+  func testFetchAndActivateUnchangedConfig() {
+    guard !APITests.useFakeConfig else { return }
```
paulb777, line 156: @ryanwilson Is `if ... return` preferable to `guard !`?

I find the double negative confusing ...

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,244 @@
...
+
+  public var requestTimeout: TimeInterval = 10
+
+  private var consoleURL: URL {
+    let api = "https://firebaseremoteconfig.googleapis.com"
+    let endpoint = "/v1/projects/\(projectID)/remoteConfig"
+    return URL(string: api + endpoint)!
+  }
+
+  // TODO: Cleanup once the GHA workflow works
```
paulb777, line 27: TODO?

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -138,6 +152,159 @@ class APITests: APITestBase {
...
+
+    let expectation = self.expectation(description: #function)
+
+    XCTAssertEqual(config.settings.minimumFetchInterval, 0)
+
+    let serialQueue = DispatchQueue(label: "\(#function)Queue")
+    let group = DispatchGroup()
+    group.enter()
+    serialQueue.async {
+      // Represents pre-fetch occuring sometime in past
```
paulb777, line 166: Nit: Add periods after comments throughout

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
paulb777: SGTM!

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
```diff
@@ -0,0 +1,66 @@
+# Remote Config Console API
+
+[`RemoteConfigConsole.swift`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift) provides a simple API for interacting with an app's remote config on the Firebase console.
```
paulb777, line 3: Nit: Limit lines in md files to 80 characters so they're any editor friendly

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
```diff
@@ -0,0 +1,66 @@
...
+If your access token wasn't generated, scroll down to the **Troubleshooting** section.
+
+Copy the access token and paste it into [`RemoteConfigConsole.swift`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift). I have included a comment that you can replace with the token. It will be a long `String` but I would avoid trying to reformat it as to not clip any characters or add unneeded spacing.
+
+🚀 Everything is ready to go!
+
+## See it in action
+
+I have included a few tests in [`APITests.swift`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift) showcasing the  `RemoteConfigConsole` in action. Check out the following tests in [`APITests.swift`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift):
+- [`testFetchConfigThenUpdateConsoleThenFetchAgain`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift#L192)
```
paulb777, line 48: Update these to the master branch repo links.

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
```diff
@@ -0,0 +1,66 @@
...
+A `.json` file will be downloaded. Go ahead and rename this file to `ServiceAccount.json` and move it to your `$HOME/.credentials/` directory. You may have to create the `.credentials/` directory.
+
+### Create the Access Token
+We use Google's [Auth Library for Swift](https://github.com/googleapis/google-auth-library-swift) to generate the access token. There are a few example use cases provided. We will use the [`TokenSource`](https://github.com/googleapis/google-auth-library-swift/blob/master/Sources/Examples/TokenSource/main.swift) example.
+
+As you can see below when we set the `GOOGLE_APPLICATION_CREDENTIALS` env variable. This library will use the `ServiceAccount.json` file we generated earlier to create our access token.
+
+#### Terminal Commands
+```bash
+$ git clone git@github.com:googleapis/google-auth-library-swift.git
```
paulb777, line 32: Can this be a script?

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
```diff
@@ -0,0 +1,66 @@
+# Remote Config Console API
+
+[`RemoteConfigConsole.swift`](https://github.com/firebase/firebase-ios-sdk/blob/nc-rc-test-api/FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift) provides a simple API for interacting with an app's remote config on the Firebase console.
+
+## Local Development Setup
+
+You can start by generating the `FirebaseRemoteConfig` project:
```
paulb777, line 7: Think about separating the steps from the explanation.

This document should address two use cases separately:
1. As simple as possible set of steps to start running as quickly as possible without necessarily understanding anything.
2. A separate explanation of why those steps are needed.

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
ncooke3: Yup, I'll make those more clear.

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
ncooke3: Yup, that is correct.

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
ncooke3: Yes, I can add one that does just that for local development. Just wanna add, we can't use the one we use in GitHub actions without knowing the secrets password. Its designed as of now to be used in a single workflow step so we had an easy way to generate tokens for a project in the future (if we add more Admin enabled testing APIs for other Firebase products).

So for local remote config development, the only catch is that you'll need to keep a copy of the unencrypted service account file for the internal rc project (or know the GitHub secrets password to run the script i wrote locally). If someone in the community wants to contribute, they can follow the steps in the readme to creating a demo project and downloading that project's affiliated service account file and so on.

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
ncooke3: I went back and made that initializer a bit cleaner. See [here](https://github.com/firebase/firebase-ios-sdk/blob/8a687101f6c7d68421a63250694fc8b3ae994e86/FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift#L27).

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
ncooke3: See the `generate_access_token.sh` script in this PR. I changed it up to be able to run both in GHA and for local development!

FirebaseRemoteConfig/Tests/SwiftAPI/Resources/README.md
ncooke3: README has been updated

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
ncooke3: I played around with getting these to work but was running into trouble trying to pick out the config value from the remote config HTTPS response object and infer its type (string, number, bool). I think it's probably best to just use the regular Firebase API `fetchConfig` for retrieving the currently active config value and then comparing it to values/dictionaries accordingly since the Firebase unwraps the config value nicely already.

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
ryanwilson: Hmm, I do prefer the `guard` in general for early exits... what do you think about `guard APITests.useFakeConfig == false else { return }`?

I don't have super strong feelings so I'm fine keeping it as `if` if someone does :)

scripts/generate_access_token.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
ryanwilson, line 1: This is excellently documented - thank you for that!

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+  /// `GoogleService-Info.plist` to set `projectID`.
+  convenience init() {
+    let currentBundle = Bundle(for: type(of: self))
+    let projectID = currentBundle.plistValue(for: "PROJECT_ID", from: "GoogleService-Info.plist")
+    self.init(projectID: projectID! as! String)
+  }
+
+  // MARK: - Public API
+
+  public func updateRemoteConfig(parameters: [String: Any]) {
```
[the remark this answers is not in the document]
ncooke3: Ahh fair point. So in the json data that comes back from the REST requests, the key for the dictionary of remote config values is "parameters". I'll make this more clear.

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+
+  private var consoleURL: URL {
+    let api = "https://firebaseremoteconfig.googleapis.com"
+    let endpoint = "/v1/projects/\(projectID)/remoteConfig"
+    return URL(string: api + endpoint)!
+  }
+
+  private lazy var accessToken: String = {
+    guard let fileURL = Bundle(for: type(of: self))
+      .url(forResource: "AccessToken", withExtension: "json") else {
```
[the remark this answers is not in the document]
ncooke3: I've since reran it and it indented it like that. I agree though, an extra two spaces inside the guard's returning clause helps make things look clearer.

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+  /// A more intuitively named setter for `latestConfig`.
+  private func save(_ config: [String: Any]) {
+    latestConfig = config
+  }
+}
+
+// MARK: - Extensions
+
+extension Bundle {
+  func plistValue(for key: String, from plist: String) -> Any? {
```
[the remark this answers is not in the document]
ncooke3: `func plistValue(forKey key: String, fromPlist plist: String) -> Any?`

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+    semaphore.wait()
+  }
+
+  /// Publishes a config object to the live console
+  /// and updates`latestConfig`.
+  private func publish(config: [String: Any]) {
+    if let configData = data(with: config) {
+      perform(configRequest: .put(data: configData))
+    }
+    save(config)
```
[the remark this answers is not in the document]
ncooke3: Agreed!

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+class RemoteConfigConsole {
+  private let projectID: String
+  private var latestConfig: [String: Any]!
```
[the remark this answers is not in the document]
ncooke3: Yup, so my idea was that it should crash if this isn't assigned. Now, when the initializers of this class call `syncWithConsole`, if the sync is not successful, we will throw a fatal error there. So I guess this is okay to leave this as is?

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+                       _ completion: (([String: Any]) -> Void)? = nil) {
+    let request = configRequest.secureRequest(url: consoleURL, with: accessToken, requestTimeout)
+
+    let semaphore = DispatchSemaphore(value: 0)
+
+    let task = URLSession.shared.dataTask(with: request) { data, response, error in
+      guard let data = data else {
+        print(String(describing: error))
+        return
+      }
```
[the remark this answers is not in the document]
ncooke3: Never knew about `defer`. super cool! i read more about it [here](https://nshipster.com/guard-and-defer/#defer)

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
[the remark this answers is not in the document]
ncooke3: Added a comment and changed the method signature

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -0,0 +1,242 @@
...
+        if let response = response as? HTTPURLResponse {
+          if response.statusCode >= 400 {
+            print(String(describing: json["error"]!))
+          }
+        }
+
+        if let completion = completion {
+          completion(json)
+        }
+      }
```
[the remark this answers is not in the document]
ncooke3: Reworked this to always ensure we call completion

ncooke3:
@ryanwilson I believe i've addressed all of your the initial feedback! 👍

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
```diff
@@ -38,60 +38,53 @@ class RemoteConfigConsole {
...
+    for (key, value) in parameters {
+      updatedParameters.updateValue(["defaultValue": ["value": value]], forKey: key)
     }
+    updatedConfig.updateValue(updatedParameters, forKey: "parameters")
 
     publish(config: updatedConfig)
   }
 
-  public func updateRemoteConfigValue(_ value: Any, for key: String) {
+  public func updateRemoteConfigValue(_ value: String, forKey key: String) {
```
ryanwilson, line 87: Did you mean to change the `value` to be a `String` here? Seems like it should stay as `Any`, unless you found a good reason otherwise :)

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
```diff
@@ -177,17 +186,19 @@ class APITests: APITestBase {
...
-        // the same time as `lastFetchTime`.
-        XCTAssertLessThanOrEqual(Double(self.config.settings.lastETagUpdateTime),
-                                 Double(self.config.lastFetchTime?.timeIntervalSince1970 ?? 0))
+        // The `lastETagUpdateTime` should either be older or the same time as `lastFetchTime`.
+        if let lastFetchTime = try? XCTUnwrap(self.config.lastFetchTime) {
+          XCTAssertLessThanOrEqual(Double(self.config.settings.lastETagUpdateTime),
+                                   Double(lastFetchTime.timeIntervalSince1970))
+        } else {
+          XCTFail("Could not unwrap lastFetchTime.")
+        }
```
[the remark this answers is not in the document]
ncooke3: Whenever I use `try` in a Firebase method's closure, I get a compiler error like this:
>`Invalid conversion from throwing function of type '(RemoteConfigFetchAndActivateStatus, Error?) throws -> Void' to non-throwing function type '(RemoteConfigFetchAndActivateStatus, Error?) -> Void'`

^This makes sense, but I can't figure out if there's a way to add the ability to `throw` when you provide the closure. Is there a way to add `throws` somewhere so you can use `try` in the method's closure?

Like `config.fetchAndActivate { **throws** result, error in ...}` <- this doesn't work but just an example of something I tried.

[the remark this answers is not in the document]
ryanwilson: Discussed offline - `XCTUnwrap` is good to go. May need to remove the `RemoteConfig` Xcode 10.3 test from Travis for it, though.

FirebaseRemoteConfig/Tests/SwiftAPI/RemoteConfigConsole.swift
ncooke3: I realized that the value needs to be able to be convertible to a string. This [json object](https://firebase.google.com/docs/remote-config/automate-rc#rest-conditions) is pretty much what the RemoteConfigConsole API is sending to the servers.  Is there a way to check if an object can be used to construct a String like `let stringifiedThing = String(object)`?

ncooke3: I guess the server logic can parse the config object out and infer the proper type for everything 😅

FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift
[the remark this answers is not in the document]
paulb777: travis is already skipping tests so it should be fine.

ncooke3:
Hi @karenyz,

This PR adds a Swift API for use during testing. It allows for synchronously changing Remote Config values through Remote Config's REST API. Since the FirebaseRemoteConfig API for iOS doesn’t have APIs to change Remote Config values, it can be difficult to test fetching dynamic Remote Config values without mocking the values returned from the Firebase console. 

This testing API allows us to run tests where we fetch from Remote Config, then use this test API to update Remote Config values on the console programmatically, and then fetch again to assert that the device’s instance of Remote Config stays up to date.

PR #6010, opened 2020-07-09
GDTCORFlatFileStorage: keep not expired events when expired batch removed

maksymmalyhin:
Fixes the following issue. It it possible that a batch expiration time is shorter than an event expiration time. With the old implementation in this case when the batch expires it will lead to removing all the events in it disregarding the event expiration time. 

The proposed fix:
- check batches for expiration first
- remove expired batches and move events from them back to the main location (unconditionally) 
- check and remove expired events from the main location (expired events from expired batches will also be there)

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -348,7 +348,7 @@ NSString *const kGDTCORBatchComponentsExpirationKey = @"GDTCORBatchComponentsExp
     // TODO: Storage may not have enough context to remove batches because a batch may be being
     // uploaded but the storage has not context of it.
 
-    // Find expired batches and remove them moving the events back to the main storage.
+    // Find expired batches and move their events back to the main storage.
     NSString *batchDataPath = [GDTCORFlatFileStorage batchDataStoragePath];
     NSArray<NSString *> *batchDataPaths = [fileManager contentsOfDirectoryAtPath:batchDataPath
                                                                            error:nil];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -27,6 +27,8 @@
 #import "GoogleDataTransport/GDTCORLibrary/Private/GDTCORRegistrar_Private.h"
 #import "GoogleDataTransport/GDTCORLibrary/Private/GDTCORUploadCoordinator.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 /** A library data key this class uses to track batchIDs. */
 static NSString *const gBatchIDCounterKey = @"GDTCORFlatFileStorageBatchIDCounter";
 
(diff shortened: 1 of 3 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -384,45 +341,46 @@ - (void)hasEventsForTarget:(GDTCORTarget)target onComplete:(void (^)(BOOL hasEve
...
-        [pathsToDelete addObject:[eventDataPath stringByAppendingPathComponent:path]];
-      }
-    }
 
-    // TODO: Events from expired batches with not expired events must be moved back to queue to
-    // avoid data loss.
     // TODO: Storage may not have enough context to remove batches because a batch may be being
     // uploaded but the storage has not context of it.
+
+    // Find expired batches and remove them moving the events back to the main storage.
```
paulb777, line 351: Do you mean 

"Find expired batches and move their events back to the main storage."

?

maksymmalyhin: yes, exactly what I mean, thank you! Sorry, I corrected the comment a couple times and ended up with a weird version.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -539,6 +497,53 @@ - (BOOL)moveContentsOfDirectoryAtPath:(NSString *)sourcePath
...
+    if (deleteEvents) {
+      removeBatchDir(batchDirPath);
+    } else {
+      NSString *batchDirName = [batchDirPath lastPathComponent];
+      NSDictionary<NSString *, id> *components = [self batchComponentsFromFilename:batchDirName];
+      NSNumber *target = components[kGDTCORBatchComponentsTargetKey];
+      NSString *destinationPath = [[GDTCORFlatFileStorage eventDataStoragePath]
+          stringByAppendingPathComponent:target.stringValue];
+
+      // `- [NSFileManager moveItemAtPath:toPath:error:] method fails if an item by the
```
mikehaney24, line 530: Unbalanced backtick

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
```diff
@@ -783,8 +790,27 @@ - (void)testBatchIDWithTarget {
   [self waitForExpectations:@[ expectation ] timeout:10];
 }
 
+- (void)testBatchIDsForTarget {
+  __auto_type expectedBatch = [self generateAndBatchEvents];
```
mikehaney24, line 794: I assume this compiles in Xcode 10?

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
```diff
@@ -823,78 +849,88 @@ - (void)testCheckEventExpiration {
...
+- (void)testCheckForExpirations_WhenBatchWithNotExpiredEventsExpires {
+  NSTimeInterval batchExpiresIn = 0.5;
+  // 0.1. Generate and batch events
+  __auto_type generatedBatch = [self generateAndBatchEventsExpiringIn:1000
+                                                      batchExpiringIn:batchExpiresIn];
+  NSNumber *generatedBatchID = [[generatedBatch allKeys] firstObject];
+  NSSet<GDTCOREvent *> *generatedEvents = generatedBatch[generatedBatchID];
+  // 0.2. Wait for batch expiration.
+  [[NSRunLoop currentRunLoop]
+      runUntilDate:[NSDate dateWithTimeIntervalSinceNow:batchExpiresIn + 0.5]];
```
mikehaney24, line 861: Maybe move the 0.5 to a variable?

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
maksymmalyhin: Yes, fortunately it does.

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
maksymmalyhin: Seems it should work without additional delays, so I removed it.

maksymmalyhin:
@mikehaney24 Thank you for the review.

I like the idea with a block either passed to `dispatch_async` or just executed. If you don't mind I'll consider it for the future changes - I still need to wrap my head around how completion block semantic will look like in this case.

I tried to put the answer to your question to the PR description (I should've done it initially). If I understand the problem correctly - the expired events from the expired batches will be also removed with the proposed implementation.

PR #6011, opened 2020-07-09
SPM - Auth unit tests

paulb777:
- Enable Auth unit tests for Swift Package Manager Builds
- Run SPM tvOS tests in CI
- Disable Auth unit tests needing an app/keychain for SPM via `exclude:` in Package.swift
- Use `#if` to specify platform restricted unit tests

#no-changelog


FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenManagerTests.m
```diff
@@ -14,8 +14,11 @@
  * limitations under the License.
  */
 
-#import <OCMock/OCMock.h>
+#include <TargetConditionals.h>
```
maksymmalyhin, line 17: I curious, why `#include` and not `#import`?

FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenTests.m
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+#include <TargetConditionals.h>
+#if !TARGET_OS_OSX
```
maksymmalyhin, line 18: Is there a reason why some tests use `TARGET_OS_IOS` and some `!TARGET_OS_OSX`?

FirebaseAuth/Tests/Unit/FIRAuthTests.m
```diff
@@ -2145,6 +2145,7 @@ - (void)testIDTokenChanges {
 
 #pragma mark - Automatic Token Refresh Tests.
 
+#if !SWIFT_PACKAGE
```
maksymmalyhin, line 2148: I feel like an explanation why the tests are disabled with a potential TODO may help us in the future to recover the tests if needed.

FirebaseAuth/Tests/Unit/FIROAuthProviderTests.m
```diff
@@ -14,12 +14,15 @@
...
-#import <OCMock/OCMock.h>
+#include <TargetConditionals.h>
+#if TARGET_OS_IOS
+
 #import <XCTest/XCTest.h>
 #import "FirebaseAuth/Sources/Public/FIRAuthErrors.h"
 #import "FirebaseAuth/Sources/Public/FIRAuthUIDelegate.h"
 #import "FirebaseAuth/Sources/Public/FIROAuthProvider.h"
 #import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
+#import "OCMock.h"
```
maksymmalyhin, line 25: nit: I remember I received comments about the header imports order to comply with go/objc-style#order-of-includes. It seems  like `#import "OCMock.h"` and `#import <XCTest/XCTest.h>` should be in a separate group from Auth headers.

FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenManagerTests.m
paulb777: Because I copied from the library source. I fixed all of them. Good catch!

FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenTests.m
paulb777: Only iOS versus iOS and tvOS

FirebaseAuth/Tests/Unit/FIRAuthTests.m
paulb777: Removed here in favor of disabling with comment in Package.swift

FirebaseAuth/Tests/Unit/FIROAuthProviderTests.m
paulb777: Done.

FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenTests.m
maksymmalyhin: Maybe worth commenting for the future. Though I'm fine with leaving it as is.

paulb777: Good point. CocoaPods manages this by explicit excludes in the podspec. A better long term solution might be to organize the unit tests directories corresponding to the platforms on which they're supported. cc: @rosalyntan

PR #6014, opened 2020-07-09
FIS API docs: use term "installation auth token" consistently.

maksymmalyhin:
- b/159732490 - consistent "installation auth token" term usage
- b/168596064 - SHA1 usage in-code explanaion


FirebaseInstallations/Source/Library/InstallationsAPI/FIRInstallationsItem+RegisterInstallationAPI.h
```diff
@@ -26,7 +26,7 @@ NS_ASSUME_NONNULL_BEGIN
  * Parses and validates the Register Installation API response and returns a corresponding
  * `FIRInstallationsItem` instance on success.
  * @param JSONData The data with JSON encoded API response.
- * @param date The Auth Token expiration date will be calculated as `date` +
+ * @param date The Installation auth token expiration date will be calculated as `date` +
```
paulb777, line 29: If Installation is capitalized, all three words should be capitalized - "Installation Auth Token"

maksymmalyhin:
@morganchen12 @andirayo Friendly ping

FirebaseInstallations/Source/Library/InstallationsAPI/FIRInstallationsItem+RegisterInstallationAPI.h
egilmorez: What's the logic behind that Paul?

I'd actually lean toward capitalizing *none* of those words, including "installation."  

This is something to discuss with Rayo, but early on the tech writers urged him to not capitalize a term as common as "installation" (thereby kind of product-izing it), and we even codified that decision in our working doc variable for the service: https://source.corp.google.com/piper///depot/google3/third_party/devsite/firebase/en/_internal/includes/_names.html;rcl=321452486;l=173

paulb777: @egilmorez My understanding is that we're not referring to a generic installation, we're referring to our specific piece of software delivered in the FirebaseInstallations CocoaPod, whether we want to call it a product or not, that we're abbreviating as `Installation` here.

Similarly, we're referring to a specific auth token associated with an instance of `Installation`.  Given that association, my expectation would be that it would also be capitalized.  I'm not 100% certain on that and I'll defer to your expertise.

egilmorez: If we're referring to a literal of some kind that has capitalization, then by all means let's backtick it for code font and keep the caps.

But if we're referring to a thing -- a thing which isn't a product nor otherwise a proper noun -- I think it should not be capitalized.

On the Firebase docs team we have agreed to resist the overall tendency to capitalize and productize common terms used widely as common nouns (vs proper nouns). For instance, "Cloud Functions" is a product you use to deploy "functions" written in JavaScript or TypeScript -- not "Functions." We think that pervasive capitalizing of common nouns is a bad idea that affects the readability of our docs and the readiness with which devs learn about our products.  

For this specific case, I think the Cloud Functions model makes sense.  OK, we have a product that we have called the Firebase Installations service; but an "installation" is still a common lowercase noun, just as "app instance" is lowercase.

morganchen12: Hey Gilmore, FYI I removed your corp link. Please avoid posting corp links on public GitHub. :smile:

morganchen12: I think some amount of capitalization is ok here because it differentiates between a generic app installation and the Firebase concept of an installation that has an auth token attached to it, but am not super strongly opinionated.

egilmorez: We post internally staged links to reference docs pretty much constantly for CF3 and Admin SDKs.  Though I agree it is best to state it is internal as a courtesy to external reviewers who can't view the docs.

The whole point I'm trying to make Morgan is that Firebase doesn't own the concept of an installation, and would be wise to not co-opt it.  If we have truly created something new, let's give it a proper name and capitalize it.  If we haven't, let's stick with lowercase.

morganchen12: Yeah, I think it's an unfortunate side-effect of the abbreviation of FirebaseInstallations being a common word. Leaving the whole thing lowercase suggests that installation auth tokens are a generic concept, and while that's not untrue, it's vague enough to not be meaningful when the doc reader is really looking for information on FIS auth tokens.

egilmorez: If you absolutely insist on that capitalization, then I'd urge prefacing it with "Firebase," as in "Firebase Installation auth token."

Meanwhile, I have let you know where at least three of the Firebase tech writers stand on this capitalization question.  Please note that for future reference.

paulb777: It seems that the disconnect is around co-opting the word `Installation`. Given that we've already released a CocoaPod called `FirebaseInstallation`, hasn't that already happened?

Firebase Installation would clarify, but might get wordy? Perhaps, we should use the acronym FIS instead?

egilmorez: No, to my mind, that doesn't follow.  

For instance, if TestLab has a `testMatrix` object, does it follow that every time a sentence in English in a document has to capitalize "Test Matrix?"  Did the existence of `InstanceID.instanceID().instanceID` mean that we had to capitalize each and every instance of "App Instance" in documentation?  I don't think so; I don't think API naming necessarily forces us to treat the in-text discussion of common nouns as though they were proper nouns.  

Morgan seems to be arguing that an "installation" as we understand it is too unique to be a common noun.  I think *that's* where our question lies.  Is there some other kind of "installation" that developers might confuse this with?  IMO it makes more sense to treat it as a common noun than to clutter our world with more capital letters.  But if you strongly disagree, I will relent.

morganchen12: >Is there some other kind of "installation" that developers might confuse this with?

I think this is unambiguous enough to leave it uncapitalized. Again, not super strongly opinionated.

maksymmalyhin: Updated to all lowercase.

morganchen12 (approved):

paulb777 (approved):

FirebaseInstallations/Source/Library/Public/FirebaseInstallations/FIRInstallations.h
```diff
@@ -85,23 +85,24 @@ NS_SWIFT_NAME(Installations)
...
 
 /**
- * Retrieves (locally or from the server depending on `forceRefresh` value) a valid authorization
- * token. An existing token may be invalidated or expire, so it is recommended to fetch the auth
- * token before each server request. This method should be used with `forceRefresh == YES` when e.g.
- * a request with the previously fetched auth token failed with "Not Authorized" error.
- * @param forceRefresh If `YES` then the locally cached auth token will be ignored and a new one
- * will be requested from the server. If `NO`, then the locally cached auth token will be returned
- * if exists and has not expired yet.
+ * Retrieves (locally or from the server depending on `forceRefresh` value) a valid Installation
```
egilmorez, line 98: Suggest lowercase for consistency with the other instances.

When we're referring to "Installations SDK" generally I think the product init-cap makes sense.

FirebaseInstallations/Source/Library/Public/FirebaseInstallations/FIRInstallations.h
```diff
@@ -85,23 +85,24 @@ NS_SWIFT_NAME(Installations)
...
+ * auth token. An existing token may be invalidated or expire, so it is recommended to fetch the
+ * installation auth token before each server request. This method should be used with `forceRefresh
+ * == YES` when e.g. a request with the previously fetched installation auth token failed with "Not
+ * Authorized" error.
+ * @param forceRefresh If `YES` then the locally cached installation auth token will be ignored and
+ * a new one will be requested from the server. If `NO`, then the locally cached Installation auth
```
egilmorez, line 104: Here too, suggest lowercase.

egilmorez (approved):
Caught a couple of tiny nits, LG!

maksymmalyhin:
cc @jeremydurham

PR #6016, opened 2020-07-10
Turn on ABT unit tests with SPM

paulb777:
Thanks to https://forums.swift.org/t/finding-swift-package-manager-resources/38058/6 handle resources for SPM unit tests

#no-changelog


Istvudlee:
Firebase 6.29.0 not configure - FirebaseApp.configure() don't work - what now work ?

PR #6022, opened 2020-07-10
SPM: Build and unit test RTDB

paulb777:
- Swift Package Manager support added
- Updates Package.swift to build and unit test RTDB
- Update unit tests to get resource from SPM bundle
- Delete unneeded files in repo
- Using leveldb from https://github.com/google/leveldb/pull/818

#no-changelog


paulb777:
Merging now and will create another PR if there are additional comments.

paulb777:
The biggest remaining question here is how best to manage/sustain the leveldb dependency, but we can deal with that later in the context of Firestore and its additional dependencies. cc: @wilhuff

PR #6032, opened 2020-07-13
[GDTCORUploadCoordinator startTimer]: don't restart timer if already started

maksymmalyhin:
A quick fix (mitigation) of b/160969496

### DIFF
--- a/GoogleDataTransport/GDTCORTests/Common/Categories/GDTCORUploadCoordinator+Testing.m
+++ b/GoogleDataTransport/GDTCORTests/Common/Categories/GDTCORUploadCoordinator+Testing.m
@@ -31,12 +31,17 @@
 
 - (void)setTimerInterval:(uint64_t)timerInterval {
   [self setValue:@(timerInterval) forKey:@"_timerInterval"];
-  dispatch_source_set_timer(self.timer, DISPATCH_TIME_NOW, timerInterval, self.timerLeeway);
+  if (self.timer) {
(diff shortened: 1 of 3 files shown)


maksymmalyhin:
> What happens if a subsequent call to `startTimer` is still waiting on the coordinationQueue while its upload events get flushed from a previous `startTimer` call?

With the change `startTimer` will do nothing if there is an existing timer, so this case should be fine if I understand it correctly. Can you see a particular issue?

GoogleDataTransport/GDTCORTests/Common/Categories/GDTCORUploadCoordinator+Testing.m
```diff
@@ -31,12 +31,16 @@ - (void)reset {
 
 - (void)setTimerInterval:(uint64_t)timerInterval {
   [self setValue:@(timerInterval) forKey:@"_timerInterval"];
-  dispatch_source_set_timer(self.timer, DISPATCH_TIME_NOW, timerInterval, self.timerLeeway);
+  if (self.timer) {
```
mikehaney24, line 34: Consider saving self.timer to a local variable to prevent future crashing bugs where it's nil'd between this check and the call to dispatch_source.

maksymmalyhin: Good call, thanks!

visumickey:
Fireperf burst of events testcases pass, now trying with spotty network and will update this PR once done.

visumickey:
Validated that there are no duplicated happening on my sample app with some common no network/spotty network scenario.

PR #6033, opened 2020-07-13
Update zip

granluo:
Turns out the upload/download artifacts are directories without being zipped.


granluo:
test was done in 8466932.

PR #6034, opened 2020-07-13
Enable cron job schedule for Firebase Performance tests on private repository

zijianjoy:
Add yml files for triggering Firebase Performance related tests on `private` repository `perf` branch.

### Explanation

Firebase Performance is currently on private repository only, and `private` repository `master` branch is in sync with `public` repository `master` branch. Any changes to `master` branch needs to happen on `public` repository only.

On private repository, currently Firebase Performance is working on a branch called `perf`, we have scheduled cron job for running tests to validate the most recent changes. 

However, `on.schedule` will run on `master` branch only: See [reference to on.schedule](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule).

In order to trigger jobs on `perf` branch, we need to add yml files to master branch, and enable its reference branch to `perf`. See [GitHub Answer](https://github.community/t/on-schedule-per-branch/17525).

### Process

Change in Github public repository master branch -> Change is synced back to private repository master branch -> Trigger scheduled job based on cron tab -> Run tests on Private repository only -> Checkout code from `perf` branch -> Finish test.

### DIFF
--- a/.github/workflows/performance-integration-tests.yml
+++ b/.github/workflows/performance-integration-tests.yml
@@ -1,0 +1,29 @@
+# Merge the yml file to master branch for the cron job schedule to be effective.
+# Reference: https://github.community/t/on-schedule-per-branch/17525
+name: performance-integration-tests
+
+on:
+  # See cron syntax references:
+  #   - https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule
+  #   - https://crontab.guru/
+  schedule:
+    # Runs every 1 hours.
+    # TODO: It is unclear when the timer starts.
+    - cron:  '0 */1 * * *'
+
+jobs:
+
+  # Build and run the Integration Tests for the Firebase performance E2E Test App.
+  performance-integration-tests:
+    # Firebase Performance lives in private repository only currently.
+    # Remove the check after Firebase Performance is open sourced.
+    if: github.repository == 'FirebasePrivate/firebase-ios-sdk'
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: perf
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/performance.yml
```diff
@@ -0,0 +1,56 @@
+# Merge the yml file to master branch for the cron job schedule to be effective.
```
paulb777, line 1: If there's no compelling reason to move this file also to public, I'd recommend leaving it on private only so that PRs can get immediate feedback.

I'm doubtful the nightly cron is worth the tradeoff, if that's the only reason.

.github/workflows/performance-integration-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+  # Build and run the Integration Tests for the Firebase performance E2E Test App.
+  performance-integration-tests:
+    # Firebase Performance lives in private repository only currently.
+    # Remove the check after Firebase Performance is open sourced.
+    if: github.repository == 'FirebasePrivate/firebase-ios-sdk'
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: perf
```
paulb777, line 25: Might be worth a comment indicating that this is specifying the `perf` branch and should also be removed on open source.

.github/workflows/performance.yml
zijianjoy: Cron job specified in this file guarantees the external dependencies of Fireperf is not breaking Fireperf iOS SDK, giving confidence on the healthiness of the product, and raise alert if breakage happens. Therefore it is needed for including this file for nightly build and test.

.github/workflows/performance-integration-tests.yml
zijianjoy: Added TODO to describe the reason and future removal, thank you for calling it out!

PR #6041, opened 2020-07-14
Add #include for config.h to get the value for HAVE_LIBDISPATCH.

dconeybe:
`FIRFirestore.mm` was updated in #5920 to use a non-libdispatch executor if `HAVE_LIBDISPATCH` is false; however, the `config.h` header file that defines `HAVE_LIBDISPATCH` was neither directly nor indirectly included, causing the `#if HAVE_LIBDISPATCH` statement to unconditionally evaluate to false. The effect of this was that callbacks from Firestore in iOS were changed to run on a non-main thread and should have been run on the main thread.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,4 +1,6 @@
-# Unreleased
+# v1.16.2
+- [fixed] Fixed a configuration issue where listeners were no longer being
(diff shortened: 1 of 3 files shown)


dconeybe:
I will create a test for callbacks occurring on the main thread in a separate PR.

PR #6044, opened 2020-07-14
License was never filled in it seems, figured it warrants a PR

thattridentdude:
License was never filled in it seems, figured it warrants a PR

Shouldn't be an earth shattering change.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 Firebase, Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


thattridentdude:
@googlebot I signed it!

LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 Firebase, Inc.
```
morganchen12, line 190: This should be Google LLC.

thattridentdude: My bad, I double checked and to me it seemed that Firebase, Inc was still in existence as a subsidiary of Google LLC.

morganchen12: No problem, this is Google policy but it's not published publicly so there was no way for you to know what the correct name should be.

thattridentdude: > No problem, this is Google policy but it's not published publicly so there was no way for you to know what the correct name should be.

Ah that makes sense. Either way, glad to be of help. It's a pedantic change, but I noticed it when I wanted to add proper attributions in my app by copying the license. Figured no one noticed after init three years ago.

PR #6045, opened 2020-07-15
Migrate zip test to zip workflow.

granluo:


granluo:
Tests are done in eea629a.

granluo:
> Shouldn't we get remote config working first?
> 
> https://github.com/firebase/firebase-ios-sdk/runs/877141340?check_suite_focus=true

The path issue was fixed before, and this seems to be new last night. I suspect this might be an infra issue in the VM. Will test it out in another pr.

paulb777:
OK, I'll approve this and leave it to your judgement about when is appropriate to merge.

granluo:
https://github.com/firebase/firebase-ios-sdk/pull/6059/checks?check_run_id=879276445
Remote config is able to run if all workflows are able to run without interruptions. Will submit it to see if there is any infra issue again nightly. If so, we might see if we need to retriger the workflow if failed.

PR #6053, opened 2020-07-15
Add #ifndef/#error when expected defines are not defined.

dconeybe:
This change helps to protect against the situation where a `#define` from `config.h` is used as the subject of an `#if` statement but `config.h` was accidentally not included. In this case, the `#if` will evaluate to false even though it would have evaluated to true had the correct header been included. This guard will help prevent bugs like the one fixed by #6041.

#no-changelog

### DIFF
--- a/Firestore/Source/API/FIRFirestore.mm
+++ b/Firestore/Source/API/FIRFirestore.mm
@@ -171,7 +171,9 @@ NS_ASSUME_NONNULL_BEGIN
     _settings = settings;
     _firestore->set_settings([settings internalSettings]);
 
-#if HAVE_LIBDISPATCH
+#ifndef HAVE_LIBDISPATCH
+#error HAVE_LIBDISPATCH must be set to 1 or 0; verify the #include directive for config.h.
+#elif HAVE_LIBDISPATCH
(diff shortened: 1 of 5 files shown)


Firestore/Source/API/FIRFirestore.mm
```diff
@@ -171,7 +171,9 @@ - (void)setSettings:(FIRFirestoreSettings *)settings {
     _settings = settings;
     _firestore->set_settings([settings internalSettings]);
 
-#if HAVE_LIBDISPATCH
+#ifndef HAVE_LIBDISPATCH
+#error HAVE_LIBDISPATCH must be set to 1 or 0; verify the #include for config.h.
```
wilhuff, line 175: This should be a string error message (i.e. enclose in double-quotes).

This is a lot of text. How about

```c
#error "Missing HAVE_LIBDISPATCH; please include config.h"
```

dconeybe:
@wilhuff I really like your idea about a Python script to run as a presubmit to verify that config.h is included when needed. I'm going to look into that and open a separate PR for it, with the plan of abandoning this one.

dconeybe:
I'm abandoning this work since the benefit is marginal and not worth the effort.

PR #6064, opened 2020-07-17

chliangGoogle:
The token branch had a strange 'git rebase' conflicts problem today after I merged from master. Will try to push to a new branch (#6103) so all the changes are not lost and I will start from a new dev branch with smaller number of commits.

Please ignore this one.


FirebaseMessaging.podspec
```diff
@@ -55,7 +58,7 @@ device, and it is completely free.
   s.osx.framework = 'SystemConfiguration'
   s.weak_framework = 'UserNotifications'
   s.dependency 'FirebaseCore', '~> 6.8'
-  s.dependency 'FirebaseInstanceID', '~> 4.3'
+  s.dependency 'FirebaseInstallations', '~> 1.0'
```
paulb777, line 61: Require at least the current 1.5 since we're unlikely to test with anything older.

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -249,6 +246,57 @@ + (void)load {
   return @[ messagingProvider ];
 }
 
+- (void)configureMessagingWithOptions:(FIROptions *)options {
+  NSString *GCMSenderID = options.GCMSenderID;
+  if (!GCMSenderID.length) {
+    FIRMessagingLoggerError(kFIRMessagingMessageCodeFIRApp000,
+                            @"Firebase not set up correctly, nil or empty senderID.");
+    [NSException raise:kFIRMessagingDomain
+                format:@"Could not configure Firebase InstanceID. GCMSenderID must not be nil or "
```
paulb777, line 255: Installations or FCM token manager instead of InstanceID?

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -512,14 +591,18 @@ - (void)setAPNSToken:(NSData *)apnsToken type:(FIRMessagingAPNSTokenType)type {
 
   // Notify InstanceID that APNS Token has been set.
   NSDictionary *userInfo = @{kFIRMessagingAPNSTokenType : @(type)};
+  // TODO(chliang) This is sent to InstanceID in case users are still using the deprecated SDK.
+  // Should be safe to remove once InstanceID is removed.
```
paulb777, line 595: Couldn't users be using InstanceID long after we remove it?

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -804,31 +910,36 @@ - (void)unsubscribeFromTopic:(NSString *)topic
...
-    if (completion) {
-      completion([NSError messagingErrorWithCode:kFIRMessagingErrorCodeInvalidTopicName
-                                   failureReason:failureReason]);
-    }
-  }];
+  [self retrieveFCMTokenForSenderID:self.tokenManager.fcmSenderID
+                         completion:^(NSString *_Nullable FCMToken, NSError *_Nullable error) {
+                           if (error) {
+                             FIRMessagingLoggerError(kFIRMessagingMessageCodeMessaging012,
+                                                     @"The unsubscription operation failed due to "
```
paulb777, line 917: s/unsubscription/unsubscribe/

chliangGoogle: We will have to adjust integration test with it as well. The same as the subscription message. I can do it in a separate PR so we can merge in g3 to fix integration test as this PR might not be checked in to master for a while.

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: No if we remove it completely from firebase-ios-sdk. if they use previous version of IID it's probably fine as messaging still kept those communications.

FirebaseMessaging.podspec
chliangGoogle: Done. Good catch.

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: Done.

PR #6072, opened 2020-07-17
Firestore and Swift Package Manager

paulb777:
Current status is a successful build and run of the Firestore quickstart.  

Next steps are to consider merging to master to facilitate early evaluation and to continue conversations about dependency open questions

Dependency Summary:

General porting notes:

- Using Xcode 12 beta 2
- Started with a master snapshot except boringssl and gRPC which branched from the versions currently depended upon by Firestore via CocoaPods
- Strong preference to organize by directories, along with excludes to avoid [major performance issues from listing files](https://forums.swift.org/t/scaling-swift-package-manager-builds/38535).
- Detailed review of excludes needed. Prototype is a minimally sufficient implementation
- Strategy for keeping Package.swift files up-to-date needed

[boringssl](https://github.com/paulb777/boringssl/tree/spm) - branched from January 22 version 
 - [Package.swift](https://github.com/paulb777/boringssl/blob/spm/Package.swift) 
 - TODO: Renaming strategy to avoid symbol collisions with other openssl libraries.

[grpc](https://github.com/paulb777/grpc/commits/spm) - gRPC-Core 1.28.2 gRPC-C++ 1.28.2 - tagged May 20
 - [Package.swift](https://github.com/paulb777/grpc/blob/spm/Package.swift)
 - Also created [two directories](https://github.com/paulb777/grpc/tree/spm/spm-core-include) to disentangle the public includes for the two libraries from the repo. The directories symbolic link to the public headers.

[absl](https://github.com/paulb777/abseil-cpp/tree/spm) - branched from master July 13
 - [Package.swift](https://github.com/paulb777/abseil-cpp/blob/spm/Package.swift)
 - First tried to match CocoaPods multiple-subspec target approach, but gave up due to circular dependencies and SwiftPM not supporting header-only targets
 - Added `#pragma clang diagnostic ignored "-Wmodules-import-nested-redundant"` to [workaround build error](https://github.com/paulb777/abseil-cpp/commit/6a901b311ba479b388f517f6aa59429c5e48e6f0)

leveldb - [PR](https://github.com/google/leveldb/pull/818) 

Screen Shot 2020-07-29 at 12 05 32 PM


Package.swift
```diff
@@ -390,6 +408,96 @@ let package = Package(
...
+      name: "FirebaseFirestore",
+      dependencies: [
+        "FirebaseCore",
+        "leveldb",
+        .product(name: "nanopb", package: "nanopb"),
+        .product(name: "abseil", package: "abseil"),
+        .product(name: "gRPC-cpp", package: "gRPC"),
+      ],
+      path: "Firestore",
+      exclude: [
```
wilhuff, line 421: Why are all these entries necessary? Shouldn't this only consist of things that would otherwise match the `sources` list?

Package.swift
```diff
@@ -390,6 +408,96 @@ let package = Package(
...
+        "Protos/objc/",
+        "core/include/",
+        "core/src",
+      ],
+      publicHeadersPath: "Source/Public",
+      cSettings: [
+        .headerSearchPath("../"),
+        .headerSearchPath("Source/Public"),
+        .headerSearchPath("Protos/nanopb"),
+        .headerSearchPath("Protos/objc/google/api"),
```
wilhuff, line 464: This is obsolete: we no longer use Objective-C protocol buffers

Package.swift
```diff
@@ -84,14 +96,18 @@ let package = Package(
       url: "https://github.com/paulb777/nanopb.git",
       .revision("849a9b82233aaa5af4286c6c594cd831ca48651c")
     ),
+    .package(
+      name: "abseil",
+      url: "https://github.com/paulb777/abseil-cpp.git",
```
wilhuff, line 101: For the purposes of *internal* experimentation this seems reasonable, but I don't think we should publish this to end users until we've settled how we're going to actually make these dependencies work. The outcome I don't want is that we end up stuck maintaining these ourselves because we've conditioned enough users to use Firestore via SPM that we can't call the experiment a failure.

Package.swift
paulb777: It's everything under Firestore that is not in `sources`. A better directory structure for SPM would be putting all sources in one directory.  I've sorted the list now.

Package.swift
paulb777: Removed here and in the podspec

paulb777:
I'll make a separate PR for the nanopb change after it tests here.

Package.swift
nidegen: Comment aside:  I wouldn't be surprised to see SPM establish itself as a C++ dependecy manager, even if it currently lacks features vs CMake. Looking forward to see how Windows support in 5.3 will push that. I also would love to see Android NDK support Swift/SPM..

PR #6080, opened 2020-07-20
M76 changelog updates - ABT, Auth, FIAM, RC

paulb777:
Changelog and version updates for M76 after adding preliminary Swift Package support for ABTesting and RemoteConfig.


paulb777:
cc: @rosalyntan, @karenyz, @christibbs

paulb777:
Confirmed check.sh locally.

PR #6081, opened 2020-07-20
Add support for dynamic url schemes and fix two crashes

jeremydurham:
* Add support for dynamic url schemes in the format appdistribution-<app id without colons>
  * Fix crash when launching an app directly via a URL scheme
  * Fix a crash on iOS 10 simulator caused by calling NSLogv twice
  * Remove uses of NSLog
  * Fix warnings

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
@@ -56,7 +56,8 @@ SFAuthenticationSession *_safariAuthenticationVC;
 
 - (void)appDistributionRegistrationFlow:(NSURL *)URL
                          withCompletion:(void (^)(NSError *_Nullable error))completion {
-  NSString *callbackURL = [NSString stringWithFormat:@"appdistribution-%@", [[self class] encodedAppId]];
+  NSString *callbackURL =
+      [NSString stringWithFormat:@"appdistribution-%@", [[self class] encodedAppId]];
 
   FIRFADInfoLog(@"Registration URL: %@", URL);
   FIRFADInfoLog(@"Callback URL: %@", callbackURL);
--- a/FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
@@ -49,9 +49,14 @@ SFAuthenticationSession *_safariAuthenticationVC;
   return sharedInstance;
 }
 
++ (NSString *)encodedAppId {
+  return [[[FIRApp defaultApp] options].googleAppID stringByReplacingOccurrencesOfString:@":"
+                                                                              withString:@""];
+}
+
 - (void)appDistributionRegistrationFlow:(NSURL *)URL
                          withCompletion:(void (^)(NSError *_Nullable error))completion {
-  NSString *callbackURL = [NSString stringWithFormat:@"appdistribution-%@", [self encodedAppId]];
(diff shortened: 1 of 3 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
```diff
@@ -71,16 +78,17 @@ - (void)appDistributionRegistrationFlow:(NSURL *)URL
...
-          NSLog(@"Testing: Sign in Complete!");
+          FIRFADInfoLog(@"Sign in complete using SFAuthenticationSession");
           // TODO (b/161538029): Map these errors to AppDistribution error codes
+
           completion(error);
         }];
 
     [_safariAuthenticationVC start];
-  } else {
+  } else if (@available(iOS 9.0, *)) {
```
schnecle, line 91: Should we raise an error in the else case? It should never happen but it may be useful to tell folks that they are out of luck.

FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
```diff
@@ -90,6 +98,11 @@ - (void)appDistributionRegistrationFlow:(NSURL *)URL
   }
 }
 
+- (NSString *)encodedAppId {
```
schnecle, line 101: Should we make this a static method because it is stateless? I'm not sure what the style guide dictates about when this is or isn't a good idea, but my java rat brain tends toward make this a static utility if it does not depend on state.

FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
```diff
@@ -100,7 +113,8 @@ - (void)showUIAlert:(UIAlertController *)alertController {
 - (BOOL)application:(UIApplication *)application
             openURL:(NSURL *)URL
             options:(NSDictionary<NSString *, id> *)options {
-  self.registrationFlowCompletion(nil);
+  FIRFADDebugLog(@"Continuing registration flow: %@", [self registrationFlowCompletion]);
+  if (self.registrationFlowCompletion) self.registrationFlowCompletion(nil);
```
schnecle, line 117: When is this thing nil? Should we instead capture this as an error?

jeremydurham: If you open the app via the URL scheme without going through the registration flow, this can be nil

This definitely can happen in non-error cases, which is why I just decided to log it instead.

FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
jeremydurham: Good call, definitely!

FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
schnecle: 👍  Sounds good!

FirebaseAppDistribution/Sources/FIRAppDistributionAppDelegateInterceptor.m
jeremydurham: Good question. I thought about this too and left it as it to discuss. The podspec will not allow someone to try to build an app for < 9.0, so there is a technical constraint in place to prevent this from happening. We could just log or crash the app in this case, but I wasn't sure if it was worth it.

schnecle: 👍  sounds good!

PR #6084, opened 2020-07-20
fix ordered code unit test for Release build

vimanyu:
Exception is thrown even in Release config. There was nothing catching the expected thrown exception. 
This was causing the "firestore_util_test" to fail, specifically "OrderedCodeInvalidEncodingsTest.NonCanonical".
Catching the expected exception now.


wilhuff:
Ran `check` locally, and the quickstart tests are unaffected. Merging despite GHA hanging.

PR #6087, opened 2020-07-20
Rework config.h handling for Swift Package Manager compatibility

wilhuff:
Swift Package Manager does not support configure-time feature testing
and does not support source transformations of any kind so we can't
cheat by running a sed command to generate config.h before building.

This change makes it so that CMake now generates a config_detected.h
while config.h always exists and contains fallback values if the build
system cannot generate config_detected.h.

Convert CocoaPods to use the same fallbacks since the values are the
same as SPM.

#no-changelog


wilhuff:
One question raised by this work is whether dynamic configuration is worthwhile at all. It certainly would be easier to just add static configuration entries for all supported platforms and be done with it.

The downside of this would be that we'd lose compatibility with any platforms that accidentally work today. For example, it's possible someone might compile Firestore for FreeBSD today and we'd correctly determine that `arc4random` is available there. Moving to a static configuration would prevent that configuration from working at all.

Firestore/core/src/util/config.h
```diff
@@ -0,0 +1,58 @@
...
+#if defined(FIRESTORE_HAVE_CONFIG_DETECTED_H)
+// If FIRESTORE_HAVE_CONFIG_DETECTED_H is defined, assume the configuration is
+// correct. You can define FIRESTORE_HAVE_CONFIG_DETECTED_H=0 to configure the
+// correct HAVE_ values via the compiler command-line.
+#  if FIRESTORE_HAVE_CONFIG_DETECTED_H
+#    include "Firestore/core/src/util/config_detected.h"
+#  endif
+#endif
+
+#elif __APPLE__ && (SWIFT_PACKAGE || COCOAPODS)
```
var-const, line 42: Doesn't the `#endif` at line 40 above close the `#if defined(FIRESTORE_HAVE_CONFIG_DETECTED_H)` branch?

wilhuff: Yes. I'm not sure how that compiled. Fixed.

PR #6088, opened 2020-07-20
Test release testing.

granluo:
Create release testing by updating Podfile to make SpecStaging repo as the source of pods.


granluo:
Test is done in 2e50955.

scripts/localize_podfile.swift
```diff
@@ -65,14 +76,18 @@ for line in lines {
         didImplicits = true
         for implicit in implicitPods {
           let implicitPodspec = repo.appendingPathComponent(implicit + ".podspec").path
-          outBuffer += "pod '\(implicit)', :path => '\(implicitPodspec)'\n"
+          outBuffer +=
+            releaseTesting
+            ? "pod '\(implicit)'\n" : "pod '\(implicit)', :path => '\(implicitPodspec)'\n"
```
paulb777, line 81: Implicit pods should not be used at all for release testing.

If all the podspecs are pushed to staging, they will be found from the `source` directive in the Podfile.

I can do a GVC tomorrow if you want to discuss verbally.

.github/workflows/remoteconfig.yml
```diff
@@ -90,6 +90,34 @@ jobs:
     - name: Test Swift Quickstart
       run: ([ -z $plist_secret ] || scripts/third_party/travis/retry.sh scripts/test_quickstart.sh Config)
 
+  quickstart_release_test:
```
paulb777, line 93: I'm not sure about putting this in remoteconfig.yml versus a separate release.yml for all of them.

scripts/localize_podfile.swift
granluo: Sure we can do the GVC. This change will set specstaging as the source of Podfile only and will not update SpecStaging repo. I am planning to update the Podfile first to let podfile find in the repo first and then we can determine how to update the SpecStaging.
I thought implict pods were involved in the tests as well since they are in the release podspec repo, but we can discuss which pods should be tested in the meeting .

.github/workflows/remoteconfig.yml
granluo: I prefer a separate release.yml since 
1. badges on readme will be very specific to the latest pods in repo. The result is intuitive to developers.
2. If release testings need to be triggered, we don't need to trigger nightly quickstart testing.

scripts/localize_podfile.swift
```diff
@@ -61,18 +72,20 @@ for line in lines {
     }
     let podspec = repo.appendingPathComponent(podName + ".podspec").path
     if FileManager().fileExists(atPath: podspec) {
-      if didImplicits == false {
+      if didImplicits == false, !releaseTesting {
```
granluo, line 75: Implicit pods will not be added if it is release testing.

scripts/localize_podfile.swift
```diff
@@ -45,11 +55,12 @@ while url.path != "/", url.lastPathComponent != "firebase-ios-sdk" {
...
-  "source 'https://cdn.cocoapods.org/'\n"
+var outBuffer =
+  "source 'https://github.com/firebase/SpecsStaging.git'\n"
+    + "source 'https://cdn.cocoapods.org/'\n"
 for line in lines {
-  var newLine = line
-  let tokens = line.components(separatedBy: [" ", ","] as CharacterSet)
+  var newLine = line.trimmingCharacters(in: .whitespacesAndNewlines)
+  let tokens = newLine.components(separatedBy: [" ", ","] as CharacterSet)
   if tokens.first == "pod" {
```
paulb777, line 64: I think this can be `if tokens.first == "pod" && !releaseTesting {` and then undo the changes inside the test.

The only change necessary for testing the staged pods should be adding source `'https://github.com/firebase/SpecsStaging.git'` to the Podfile.

PR #6093, opened 2020-07-21
Add a GitHub Action to check for required #include directives for config.h

dconeybe:
This PR adds the file `scripts/check_cmake_includes.py`. This new Python script, when run via the `scripts/check_cmake_includes.sh` wrapper, reads in `Firestore/core/src/util/config_detected.h.in` and enumerates the `#cmakedefine` declarations. It then scans all of the Firestore source code to verify that any file that uses one of these "defines" has the required `#include "Firestore/core/src/util/config.h"` directive.

The motivation for this check is to prevent future occurrences of bugs like the one fixed in #6041. I had originally implemented this "check" in a much more manual way in #6053, and the reviewer there suggested a more robust, automatic, and and extensible check like this one.

This PR also adds this check to the GitHub Actions Workflows.

#no-changelog


dconeybe:
@wilhuff Just FYI this is not urgent. There is no rush for your feedback.

scripts/check_cmake_includes.py
```diff
@@ -0,0 +1,459 @@
...
+
+  # The string argument to the #include directive that was expected to be
+  # present for the define but was missing.
+  include: str
+
+  # The line number of the file at which the define was used.
+  line_number: int
+
+
+class RequiredIncludesChecker:
```
wilhuff, line 146: If you kept a count of the files processed in here and added a summarize method that just wrote to stdout, you could eliminate the logger.

dconeybe:
I'm abandoning this work since the benefit is marginal and not worth the effort. Thanks @wilhuff for the review.

PR #6096, opened 2020-07-21
Move Core Publics so imports match between SPM and CocoaPods

paulb777:
Also removes modules to enable Objective C++ clients like Firestore

#no-changelog


FirebaseCore/Sources/FIRApp.m
```diff
@@ -22,7 +22,7 @@
 #import <AppKit/AppKit.h>
 #endif
 
-#import "FirebaseCore/Sources/Public/FIRApp.h"
+#import "FirebaseCore/Sources/Public/FirebaseCore/FIRApp.h"
```
wilhuff, line 25: Does this constitute a breaking change? Just making sure, because I suspect it's OK because of CocoaPods flattening behavior, but am not sure.

paulb777: It's my understanding that the CocoaPods flattening makes this non-breaking.  See the FirebaseCore.podspec change.

 I plan to validate with more extensive testing including using this Core with old Firebase 6.x deps ...

paulb777:
All travis and GHA CI tests pass. I also deployed to SpecsStaging and verified that qualified and unqualified FirebaseCore public header imports work as before.

paulb777:
@wilhuff I see how to eliminate the `#if SWIFT_PACKAGE` for accessing other public headers in and out of the repo.  

I suppose for private headers, we could go back to accessing them via the module for CocoaPods and structuring like `include/FirebaseCore/Public.h` and `include/FirebaseCoreInternal/Private.h` so SPM could find both. Is that what you mean and why do you think that's better?

PR #6099, opened 2020-07-21
Error code mapping and DelegateInterceptor refactoring

pranavrajgopal:
* Mapping sign in flow errors to app distribution errors. 
* Also renaming AppDelegateInterceptor to FIRAppDistributionUIService

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -19,9 +19,9 @@
 #import "GoogleUtilities/UserDefaults/Private/GULUserDefaults.h"
 
 #import "FIRAppDistribution+Private.h"
-#import "FIRAppDistributionAppDelegateInterceptor.h"
 #import "FIRAppDistributionMachO+Private.h"
 #import "FIRAppDistributionRelease+Private.h"
+#import "FIRAppDistributionUIService.h"
 #import "FIRFADApiService+Private.h"
 #import "FIRFADLogger+Private.h"
 
@@ -32,7 +32,7 @@
 @interface FIRAppDistribution () <FIRLibrary, FIRAppDistributionInstanceProvider>
 @property(nonatomic) BOOL isTesterSignedIn;
 
-@property(nullable, nonatomic) FIRAppDistributionAppDelegateInterceptor *appDelegateInterceptor;
+@property(nullable, nonatomic) FIRAppDistributionUIService *UIService;
 
 @end
 
@@ -71,8 +71,8 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
 
   if (self) {
     [GULAppDelegateSwizzler proxyOriginalDelegate];
-    self.appDelegateInterceptor = [FIRAppDistributionAppDelegateInterceptor sharedInstance];
-    [GULAppDelegateSwizzler registerAppDelegateInterceptor:self.appDelegateInterceptor];
+    self.UIService = [FIRAppDistributionUIService sharedInstance];
+    [GULAppDelegateSwizzler registerAppDelegateInterceptor:self.UIService];
   }
 
   return self;
@@ -135,7 +135,7 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
     return;
   }
 
-  [self.appDelegateInterceptor initializeUIState];
+  [self.UIService initializeUIState];
   FIRInstallations *installations = [FIRInstallations installations];
 
   // Get a Firebase Installation ID (FID).
@@ -147,6 +147,8 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
                                   : @"Failed to retrieve Installation ID.";
       completion([self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorUnknown
                                              message:description]);
+
+      [self.UIService resetUIState];
       return;
     }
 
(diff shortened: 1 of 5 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -68,10 +68,20 @@ - (void)appDistributionRegistrationFlow:(NSURL *)URL
         callbackURLScheme:callbackURL
         completionHandler:^(NSURL *_Nullable callbackURL, NSError *_Nullable error) {
           [self resetUIState];
-          FIRFADInfoLog(@"Sign in complete using ASWebAuthenticationSession");
-          // TODO (b/161538029): Map these errors to AppDistribution error codes
-
-          completion(error);
+          [self logRegistrationCompletion:error authType:[ASWebAuthenticationSession description]];
+          if (error) {
```
schnecle, line 72: Can we extract this as a method so that we don't have to repeat this code? You could even have the error handler return error or nil and just ` return completion([self getAppDistributionError:appDistributionError])`. Maybe something like:
```
+ (NSError *_Nullable)mapErrorAppDistributionError:(NSError *_Nullable)error{
    if(!error) {
       return nil;
    }
    
    if([error code] == ASWebAuthenticationSessionErrorCodeCanceledLogin || [error code]  == SFAuthenticationErrorCanceledLogin) {
        return [self getAppDistribution: FIRAppDistributionErrorAuthenticationCancelled];
    }
    return [self getAppDistribution: FIRAppDistributionErrorAuthenticationFailure];
}
```

FirebaseAppDistribution/Sources/Private/FIRAppDistributionUIService.h
```diff
@@ -19,31 +19,26 @@
 #import "FIRAppDistribution+Private.h"
 
 NS_ASSUME_NONNULL_BEGIN
-
+/// An instance of this class provides UI services required for the App Distribution sdk
 /// An instance of this class is meant to be registered as an AppDelegate interceptor, and
```
schnecle, line 23: Should we remove or combine the previous comment?

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -71,8 +71,8 @@ - (instancetype)initWithApp:(FIRApp *)app appInfo:(NSDictionary *)appInfo {
 
   if (self) {
     [GULAppDelegateSwizzler proxyOriginalDelegate];
-    self.appDelegateInterceptor = [FIRAppDistributionAppDelegateInterceptor sharedInstance];
-    [GULAppDelegateSwizzler registerAppDelegateInterceptor:self.appDelegateInterceptor];
+    self.UIService = [FIRAppDistributionUIService sharedInstance];
```
schnecle, line 74: nit: uiService - Jeremy and I looked this up, and it looks like you can either do this or do lower case when an acronym begins a variable. My preference is lower case, though it's not a hill I'm willing to die on.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -157,28 +159,33 @@ - (void)signInTesterWithCompletion:(void (^)(NSError *_Nullable error))completio
...
-        }
-        [[GULUserDefaults standardUserDefaults] setBool:YES forKey:kFIRFADSignInStateKey];
-      }];
+- (void)persistTesterSignInStateAndHandleCompletion:(void (^)(NSError *_Nullable error))completion {
+  [FIRFADApiService fetchReleasesWithCompletion:^(NSArray *_Nullable releases,
+                                                  NSError *_Nullable error) {
+    if (error) {
+      FIRFADErrorLog(@"Could not fetch releases with code %ld - %@", [error code],
+                     [error localizedDescription]);
+      completion([self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorAuthenticationFailure
```
schnecle, line 181: hmmm... Do we want this to be an authentication error? I think you want to call the error mapper -https://github.com/firebase/firebase-ios-sdk/blob/app-distribution-eap-reloaded/FirebaseAppDistribution/Sources/FIRAppDistribution.m#L211 - to map the error codes returned from the fetch call appropriately.

pranavrajgopal: I think this is an authentication error and not a fetch releases error. We call fetch releases because it's the only way to validate if the registration is successful. But the end result for the user is that registration failed and the error returned back should reflect that.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
pranavrajgopal: Yeah was not sure about that. Will update.

FirebaseAppDistribution/Sources/Private/FIRAppDistributionUIService.h
pranavrajgopal: That was my intention. But I can make the comments have a better flow.

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
pranavrajgopal: I thought about doing that. But that would mean repeating the ios version availability code again for `ASWebAuthenticationSessionErrorCodeCanceledLogin`  and `SFAuthenticationErrorCanceledLogin `. Neither options felt that great so I decided to go with repeating the logging logic.

schnecle: Why would we need to duplicate that piece of the logic? I think I'm misunderstanding something here.

FirebaseAppDistribution/Sources/Private/FIRAppDistributionUIService.h
schnecle: I think just removing the duplicate `An instance` might do it - it looked like an oversight because of the duplicate structure. I see what you mean here - maybe `An instance of this class provides UI elements required for the App Distribution tester authentication flow as an AppDelegate interceptor.`

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: The error here is really that we failed to persist the sign in state because we failed to fetch releases, but we could have failed to fetch releases because of a network timeout, and that is meaningfully different than a failure to authenticate. In the case where we have successfully registered but failed to persist that state, a call to fetch may subsequently succeed, we just won't know about it session over session. Imo the more specific we can be with the error, the better.

PR #6102, opened 2020-07-21
GDT: fix time units for the request and events uptime

maksymmalyhin:
Fixes b/161747889

- `GDTCORClock.uptime` actual time units aligned with method names
- `GDTCCTNanopbHelpers` - assign actual milliseconds to `..._uptime_ms` proto fields
- tests

### DIFF
--- a/GoogleDataTransport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
+++ b/GoogleDataTransport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
@@ -128,7 +128,7 @@ gdt_cct_LogRequest GDTCCTConstructLogRequest(int32_t logSource,
   GDTCORClock *currentTime = [GDTCORClock snapshot];
   logRequest.request_time_ms = currentTime.timeMillis;
   logRequest.has_request_time_ms = 1;
-  logRequest.request_uptime_ms = currentTime.uptime;
+  logRequest.request_uptime_ms = [currentTime uptimeMilliseconds];
   logRequest.has_request_uptime_ms = 1;
 
   return logRequest;
@@ -138,7 +138,7 @@ gdt_cct_LogEvent GDTCCTConstructLogEvent(GDTCOREvent *event) {
   gdt_cct_LogEvent logEvent = gdt_cct_LogEvent_init_default;
   logEvent.event_time_ms = event.clockSnapshot.timeMillis;
   logEvent.has_event_time_ms = 1;
-  logEvent.event_uptime_ms = event.clockSnapshot.uptime;
+  logEvent.event_uptime_ms = [event.clockSnapshot uptimeMilliseconds];
   logEvent.has_event_uptime_ms = 1;
   logEvent.timezone_offset_seconds = event.clockSnapshot.timezoneOffsetSeconds;
   logEvent.has_timezone_offset_seconds = 1;
--- a/GoogleDataTransport/GDTCCTTests/Unit/GDTCCTNanopbHelpersTest.m
+++ b/GoogleDataTransport/GDTCCTTests/Unit/GDTCCTNanopbHelpersTest.m
@@ -109,10 +109,34 @@
   pb_istream_t istream =
       pb_istream_from_buffer([encodedBatchLogRequest bytes], [encodedBatchLogRequest length]);
   XCTAssertTrue(pb_decode(&istream, gdt_cct_BatchedLogRequest_fields, &decodedBatch));
-  XCTAssert(decodedBatch.log_request_count == batch.log_request_count);
-  XCTAssert(decodedBatch.log_request[0].log_event_count == batch.log_request[0].log_event_count);
-  XCTAssert(decodedBatch.log_request[0].log_event[0].event_time_ms ==
-            batch.log_request[0].log_event[0].event_time_ms);
+  XCTAssertEqual(decodedBatch.log_request_count, batch.log_request_count);
+  XCTAssertEqual(decodedBatch.log_request[0].log_event_count, batch.log_request[0].log_event_count);
+  XCTAssertEqual(decodedBatch.log_request[0].log_event[0].event_time_ms,
+                 batch.log_request[0].log_event[0].event_time_ms);
+  XCTAssertEqual(decodedBatch.log_request[0].log_event[0].event_uptime_ms,
+                 batch.log_request[0].log_event[0].event_uptime_ms);
+  pb_release(gdt_cct_BatchedLogRequest_fields, &batch);
+  pb_release(gdt_cct_BatchedLogRequest_fields, &decodedBatch);
+}
+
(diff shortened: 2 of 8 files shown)


GoogleDataTransport/GDTCORLibrary/Public/GDTCORClock.h
```diff
@@ -27,10 +27,10 @@ NS_ASSUME_NONNULL_BEGIN
 /** The offset from UTC in seconds. */
 @property(nonatomic, readonly) int64_t timezoneOffsetSeconds;
 
-/** The kernel boot time when this clock was created. */
+/** The kernel boot time when this clock was created in nanoseconds. */
 @property(nonatomic, readonly) int64_t kernelBootTime;
 
-/** The device uptime when this clock was created. */
+/** The device uptime when this clock was created in nanoseconds. */
 @property(nonatomic, readonly) int64_t uptime;
```
mikehaney24, line 34: You didn't want to rename the property?

maksymmalyhin: I'm a bit hesitant doing this because it will be another breaking change. Though, I may deprecate it and create another with units in the name.

maksymmalyhin: @paulb777 @ryanwilson Any objections against a new property + deprecation?

paulb777: Fine with me - as long as we avoid a breaking change.

GoogleDataTransport/GDTCCTTests/Unit/GDTCCTNanopbHelpersTest.m
```diff
@@ -109,10 +109,34 @@ - (void)testBytesAreDecodable {
...
+  XCTAssertEqual(decodedBatch.log_request[0].log_event_count, batch.log_request[0].log_event_count);
+  XCTAssertEqual(decodedBatch.log_request[0].log_event[0].event_time_ms,
+                 batch.log_request[0].log_event[0].event_time_ms);
+  XCTAssertEqual(decodedBatch.log_request[0].log_event[0].event_uptime_ms,
+                 batch.log_request[0].log_event[0].event_uptime_ms);
+  pb_release(gdt_cct_BatchedLogRequest_fields, &batch);
+  pb_release(gdt_cct_BatchedLogRequest_fields, &decodedBatch);
+}
+
+- (void)testDecodedEventTimestampMatchToBatchContent {
```
visumickey, line 122: Can we have some tests directly for GDTCCTConstructLogEvent and GDTCCTConstructLogRequest API to ascertain the availability of the values for LogEvent? These are critical fields that the users of Firelog will be dependent on.

GoogleDataTransport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
```diff
@@ -138,7 +138,7 @@ gdt_cct_LogEvent GDTCCTConstructLogEvent(GDTCOREvent *event) {
   gdt_cct_LogEvent logEvent = gdt_cct_LogEvent_init_default;
   logEvent.event_time_ms = event.clockSnapshot.timeMillis;
   logEvent.has_event_time_ms = 1;
-  logEvent.event_uptime_ms = event.clockSnapshot.uptime;
+  logEvent.event_uptime_ms = [event.clockSnapshot uptimeMilliseconds];
```
visumickey, line 141: Is there a way to ascertain if these are milliseconds and not nanoseconds or seconds?

GoogleDataTransport/GDTCORLibrary/Public/GDTCORClock.h
```diff
@@ -27,11 +27,17 @@ NS_ASSUME_NONNULL_BEGIN
...
-/** The device uptime when this clock was created. */
-@property(nonatomic, readonly) int64_t uptime;
+/** The device uptime when this clock was created in nanoseconds. */
+@property(nonatomic, readonly) int64_t uptimeNanoseconds;
+
+@property(nonatomic, readonly) int64_t kernelBootTime DEPRECATED_MSG_ATTRIBUTE(
+    "Please use `kernelBootTimeNanoseconds` instead");
+
+@property(nonatomic, readonly)
+    int64_t uptime DEPRECATED_MSG_ATTRIBUTE("Please use `uptimeNanoseconds` instead");
```
visumickey, line 40: Do you want to qualify that this is buggy? This would enable all team who are depending on the Firelog library would know that they have to update to the newer property?

GoogleDataTransport/GDTCORTests/Unit/GDTCORClockTest.m
```diff
@@ -81,4 +81,36 @@ - (void)testIsAfter {
...
 
+- (void)testUptime {
+  NSTimeInterval timeDiff = 1;
+  GDTCORClock *clock1 = [GDTCORClock snapshot];
+  [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:timeDiff]];
+  GDTCORClock *clock2 = [GDTCORClock snapshot];
+
+  XCTAssertGreaterThan(clock2.uptimeNanoseconds, clock1.uptimeNanoseconds);
+  NSTimeInterval uptimeDiff =
+      (clock2.uptimeNanoseconds - clock1.uptimeNanoseconds) / (NSTimeInterval)NSEC_PER_SEC;
```
visumickey, line 92: I would replace NSTimeInterval with "double" as as it is weird to convert NSEC_PER_SEC to NSTimeInterval.

GoogleDataTransport/GDTCORLibrary/Public/GDTCORClock.h
maksymmalyhin: The current implementation just forwards the call to the new properties so there is no actual harm in using deprecated ones except they are going to be removed in the next major release.

paulb777:
Please add the tests in a separate PR to unblock today's scheduled RC creation.

GoogleDataTransport/GDTCCTLibrary/GDTCCTNanopbHelpers.m
maksymmalyhin: It is asserted [here](https://github.com/firebase/firebase-ios-sdk/blob/505093c87201a1b2314fe9cc193b6ae93ed6bbed/GoogleDataTransport/GDTCCTTests/Unit/GDTCCTNanopbHelpersTest.m#L136).

maksymmalyhin:
FYI @samedson

maksymmalyhin:
Additional tests for GDTCCTConstructLogEvent and GDTCCTConstructLogRequest functions will be added in the following PRs.

PR #6103, opened 2020-07-22

chliangGoogle:
- Move checkin logic from IID to Messaging (FIRMessagingAuth* FIRMessagingCheckin*)
- Move Token logic from IID to Messaging (FIRMessagingToken*, all the token operations files)
(The files above mostly copy over, rename and fix compiling errors and logic issues)
- Merge MessagingCode and Utilities between IID and Messaging.
- I have not added unit tests from IID yet to prevent this PR getting even bigger.
- All IID code is under token/ folder.
- FIRMessagingTokenManager serves as the central object that holds the values of token and checkin info. It is the communication between messaging and previous IID classes.
- All the persistent storage path are kept the same to ensure IID still works with Messaging if users are still using InstanceID.
- Kept all the communications between IID and Messaging in case users still use IID API to handle token and apns info. (APNS used to be in iid)

- [x] Tested with Messaging getToken and deleteToken API.
- [ ] When token is updated from messaging side, need to add a new notification to notify IID to update its cache.
- [ ] Tested interactions between Messaging Token API and IID Token API and FID API (this is blocked by previous task)

#no-changelog

### Testing

### API Changes

    can discuss it together.


chliangGoogle:
The previous PR (#6064) had a strange 'git rebase' conflicts problem today after I merged from master. I suspect it's something with the token branch. Will try to push to a new branch so all the changes are not lost and I will start from a new dev branch with smaller number of commits.

Create this new one and @paulb777 can you approve again. Should be no difference.

PR #6110, opened 2020-07-22
Fix check breakage from when GHA was out

paulb777:
@no-changelog


paulb777:
Manually verified check regression fixed. Merging now to minimize the impact.

PR #6111, opened 2020-07-22
Make the E2E test apps run once every 4 hours.

visumickey:


.github/workflows/performance-integration-tests.yml
```diff
@@ -8,9 +8,8 @@ on:
   #   - https://crontab.guru/
   schedule:
     # Runs every 1 hours.
```
zijianjoy, line 10: NIT: Also change this to `every 4 hours` and remove the TODO below.

visumickey: Done

PR #6113, opened 2020-07-23
Fix RemoteConfig sample

paulb777:
Fix recent build breakages in RemoteConfig sample app from recent repo restructures:

- Replace path imports in private headers with `@class` declarations
- Move Public directory so it works for SPM without a `#if` and still works for CocoaPods
- Add dummy `plist` files to the Sample so it builds without explicit adds
- Use the local FirebaseABTesting in the Podfile so version changes don't break build
- Verify the sample continues to build in GHA CI

#no-changelog Changelog will be updated closer to M77 release in the context of potentially additional structural changes for standardization and SPM.


paulb777:
cc: @erikeldridge

erikeldridge:
Thx @paulb777 !

PR #6117, opened 2020-07-23
Feedback from PR#6099 

pranavrajgopal:
https://github.com/firebase/firebase-ios-sdk/pull/6099 
* Making sure we return granular errors during sign in persistence. 
* Also fixing some naming and comment issues

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -173,19 +173,18 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
 }
 
 - (void)persistTesterSignInStateAndHandleCompletion:(void (^)(NSError *_Nullable error))completion {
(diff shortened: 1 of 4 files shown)


FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -32,7 +32,7 @@ @protocol FIRAppDistributionInstanceProvider <NSObject>
 @interface FIRAppDistribution () <FIRLibrary, FIRAppDistributionInstanceProvider>
 @property(nonatomic) BOOL isTesterSignedIn;
 
-@property(nullable, nonatomic) FIRAppDistributionUIService *UIService;
+@property(nullable, nonatomic) FIRAppDistributionUIService *uiService;
```
jeremydurham, line 35: Looking around, I think we should leave this as UIService

pranavrajgopal: Fixed!

PR #6121, opened 2020-07-24
Add support for deferred OAuth provisioning

rosalyntan:
Add support for deferred OAuth provisioning, to use the GOOGLE_APP_ID in place of the CLIENT_ID for generic IDP and phone auth ReCAPTCHA fallback flows once the CLIENT_ID is no longer provided in the plist file.


FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
```diff
@@ -272,19 +284,25 @@ - (void)getHeadFulLiteURLWithEventID:(NSString *)eventID
...
                                        @"sessionId" : [strongSelf hashforString:sessionID],
                                        @"v" : [FIRAuthBackend authUserAgent],
                                        @"eventId" : eventID,
                                        @"providerId" : strongSelf->_providerID,
                                      } mutableCopy];
+                                     if (clientID) {
+                                       urlArguments[@"clientId"] = clientID;
+                                     } else {
+                                       urlArguments[@"appId"] = firebaseAppID;
+                                     }
```
renkelvin, line 305: I'm not sure if there is a case that someone depends on the "clientId" explicitly. (You should double check FirebaseUI) While this change looks a bit dangerous.

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
```diff
@@ -272,19 +284,25 @@ - (void)getHeadFulLiteURLWithEventID:(NSString *)eventID
                                      }
                                      __strong __typeof__(self) strongSelf = weakSelf;
                                      NSString *bundleID = [NSBundle mainBundle].bundleIdentifier;
-                                     NSString *clienID = strongSelf->_auth.app.options.clientID;
+                                     NSString *clientID = strongSelf->_auth.app.options.clientID;
+                                     NSString *firebaseAppID =
```
renkelvin, line 288: nit: using both firebaseAppID and googleAppID looks confusing. Maybe simply name the variable as `appID`?

FirebaseAuth/Tests/Unit/FIROAuthProviderTests.m
```diff
@@ -91,17 +101,16 @@
     @brief A fake callback URL containing a fake response URL.
  */
 static NSString *const kFakeRedirectURLResponseURL =
-    @"com.googleusercontent.apps.123456://firebase"
-     "auth/"
-     "link?deep_link_id=https%3A%2F%2Fexample.firebaseapp.com%2F__%2Fauth%2Fcallback%3FauthType"
-     "%3DsignInWithRedirect%26link%3D";
+    @"://firebaseauth/"
```
renkelvin, line 104: the scheme is missing

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
rosalyntan: Done.

FirebaseAuth/Tests/Unit/FIROAuthProviderTests.m
rosalyntan: Amended the comment to clarify -- this the URL without the scheme, since the scheme can be either the reverse client ID or the encoded app ID.

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
renkelvin: I actually suggest keeping the key name "clientId". While it's your call.

FirebaseAuth/Tests/Unit/FIROAuthProviderTests.m
renkelvin: Sounds good.

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
rosalyntan: Do you mean for the URL query parameter? On the web widget side, I have a change in progress for adding support to the appId query parameter (as the logic will branch based on whether the appId or the clientId is passed).

Regarding FirebaseUI, it looks like FUIGoogleAuth depends on the clientId. However, it should be noted that for now this change is a no-op (since the client ID is still being provided in the plist). I will update FirebaseUI in another PR.

Does this address your concerns?

renkelvin: Yeah, sounds reasonable. But if we keep using "clientId", you can save the efforts of making the additional changes right?

rosalyntan: Ah I see, so the web widget needs to send a different parameter in the request to the backend for validation based on if it's the client ID or the app ID, so I don't think the query param can be reused.

renkelvin: Got it. Thanks for clarifying!

PR #6126, opened 2020-07-24
Switch App Distribution SDK from AppAuth to FirebaseInstallations

jeremydurham:
* Add GHA for App Distribution SDK
  * Update SDK minimum version to iOS 9.0
  * Update changelog for beta release
  * Change imports to Sources/
  * Switch from AppAuth -> FirebaseInstallations

#no-changelog: We will update the changelog once this is merged and ready to be included in a release.


jeremydurham:
This is a fairly large PR, so we're prepared a guide to help make it easier to review: https://docs.google.com/document/d/1rS4z4igbpPM8AX3NZ88bOQ2bQ5jqQGsM1gvbZpqkvsw/edit

FirebaseAppDistribution.podspec
```diff
@@ -25,14 +25,19 @@ iOS SDK for App Distribution for Firebase.
...
     'GoogleUtilities/AppDelegateSwizzler/Private/*.h',
+    'GoogleUtilities/UserDefaults/Private/*.h',
   ]
   s.public_header_files = base_dir + 'Public/*.h'
   s.private_header_files = base_dir + 'Private/*.h'
 
   s.dependency 'FirebaseCore', '~> 6.8'
-  s.dependency 'AppAuth', '~> 1.2.0'
   s.dependency 'GoogleUtilities/AppDelegateSwizzler', '~> 6.7'
+  s.dependency 'GoogleUtilities/UserDefaults', '~>6.7'
```
paulb777, line 38: nit: space after >

FirebaseAppDistribution.podspec
```diff
@@ -25,14 +25,19 @@ iOS SDK for App Distribution for Firebase.
...
+  s.dependency 'FirebaseInstallations', '~> 1.1'
```
paulb777, line 39: 1.5

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -11,62 +11,70 @@
...
 
-// The OAuth scope needed to authorize the App Distribution Tester API
-NSString *const kOIDScopeTesterAPI = @"https://www.googleapis.com/auth/cloud-platform";
+@implementation FIRAppDistribution
 
 // The App Distribution Tester API endpoint used to retrieve releases
-NSString *const kReleasesEndpointURL =
-    @"https://firebaseapptesters.googleapis.com/v1alpha/devices/-/testerApps/%@/releases";
-NSString *const kTesterAPIClientID =
-    @"319754533822-osu3v3hcci24umq6diathdm0dipds1fb.apps.googleusercontent.com";
```
paulb777, line 42: probably need to invalidate this

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -0,0 +1,215 @@
...
+#import "FirebaseAppDistribution/Sources/Public/FIRAppDistribution.h"
+#import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
+
+#import <AuthenticationServices/AuthenticationServices.h>
+#import <SafariServices/SafariServices.h>
+#import <UIKit/UIKit.h>
+
+@implementation FIRAppDistributionUIService
+
+API_AVAILABLE(ios(9.0))
```
paulb777, line 26: Wondering if ios 9 support is worth it versus starting with 10 or 11?

Firebase is moving to 10 in the fall.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -82,7 +90,7 @@ + (void)load {
   FIRComponentCreationBlock creationBlock =
       ^id _Nullable(FIRComponentContainer *container, BOOL *isCacheable) {
     if (!container.app.isDefaultApp) {
-      // TODO: Implement error handling
+      // TODO: Remove this and log error
```
schnecle, line 93: We will remove this.

ZipBuilder/Sources/ZipBuilder/FirebasePods.swift
```diff
@@ -38,6 +38,7 @@ enum FirebasePods: String, CaseIterable {
   case performance = "FirebasePerformance"
   case remoteConfig = "FirebaseRemoteConfig"
   case storage = "FirebaseStorage"
+  case appdistribution = "FirebaseAppDistribution"
```
paulb777, line 41: alphabetize

FirebaseAppDistribution.podspec
```diff
@@ -25,14 +25,19 @@ iOS SDK for App Distribution for Firebase.
   s.source_files = [
     base_dir + '**/*.{c,h,m,mm}',
     'FirebaseCore/Sources/Private/*.h',
+    'FirebaseInstallations/Source/Library/Private/*.h',
+    'GoogleDataTransport/GDTCORLibrary/Internal/*.h',
     'GoogleUtilities/AppDelegateSwizzler/Private/*.h',
+    'GoogleUtilities/UserDefaults/Private/*.h',
   ]
   s.public_header_files = base_dir + 'Public/*.h'
   s.private_header_files = base_dir + 'Private/*.h'
```
paulb777, line 34: This is being removed separately? `private_header_files` should only be used for intra-pod dependencies.

paulb777:
@maksymmalyhin Would you take a look specifically at the FIS usage?

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
jeremydurham: Supporting 11.x would allow us to drop support for SFSafariViewController, but we would still need to support it with iOS 10. 

Being that our product is generally used by beta testers for older iOS versions and that dropping support for iOS 9 doesn't buy us much, we'd like to ship with iOS 9 support until Firebase officially moves to iOS 10.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -116,181 +123,160 @@ + (instancetype)appDistribution {
...
+
+  FIRFADErrorLog(@"Failed to retrieve releases with unexpected domain %@: %ld", [error domain],
+                 (long)[error code]);
+  return [self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorUnknown
+                                     message:@"Failed to fetch releases for unknown reason."];
 }
+
+- (void)fetchNewLatestRelease:(FIRAppDistributionUpdateCheckCompletion)completion {
+  [FIRFADApiService
+      fetchReleasesWithCompletion:^(NSArray *_Nullable releases, NSError *_Nullable error) {
```
maksymmalyhin, line 245: nit: using static methods and singletons usually limits code testability. Injecting an instance of `FIRFADApiService` usually provides much more room for testing and mocking without touching Objective-C runtime as well as improves the code readability.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -116,181 +123,160 @@ + (instancetype)appDistribution {
...
+  if ([self isTesterSignedIn]) {
+    completion(nil);
+    return;
   }
 
-  self.authState = nil;
-  self.isTesterSignedIn = false;
-}
+  [self.UIService initializeUIState];
+  FIRInstallations *installations = [FIRInstallations installations];
```
maksymmalyhin, line 139: As far as I can see `FIRAppDistribution` supports multiple `FIRApps`. If so, then [installationsWithApp:](https://github.com/firebase/firebase-ios-sdk/blob/0ed26aedb85f3f3fd37b0077133347183db5de7a/FirebaseInstallations/Source/Library/Public/FIRInstallations.h#L74) method should be used to get an instance for the app.

FirebaseAppDistribution/Sources/FIRFADApiService.m
```diff
@@ -0,0 +1,201 @@
...
+    @"https://firebaseapptesters.googleapis.com/v1alpha/devices/"
+    @"-/testerApps/%@/installations/%@/releases";
+NSString *const kInstallationAuthHeader = @"X-Goog-Firebase-Installations-Auth";
+NSString *const kApiHeaderKey = @"X-Goog-Api-Key";
+NSString *const kResponseReleasesKey = @"releases";
+
+@implementation FIRFADApiService
+
++ (void)generateAuthTokenWithCompletion:(FIRFADGenerateAuthTokenCompletion)completion {
+  FIRInstallations *installations = [FIRInstallations installations];
```
maksymmalyhin, line 34: The same here, [installationsWithApp:](https://github.com/firebase/firebase-ios-sdk/blob/0ed26aedb85f3f3fd37b0077133347183db5de7a/FirebaseInstallations/Source/Library/Public/FIRInstallations.h#L74) method should be used if App Distribution supports multiple apps.

FirebaseAppDistribution/Tests/Unit/FIRFADApiServiceTests.m
```diff
@@ -0,0 +1,411 @@
...
+        XCTAssertEqual([error code], FIRFADApiInstallationIdentifierError);
+        [expectation fulfill];
+      }];
+
+  [self waitForExpectations:@[ expectation ] timeout:5.0];
+}
+
+- (void)testFetchReleasesWithCompletionSuccess {
+  NSHTTPURLResponse *fakeResponse = OCMClassMock([NSHTTPURLResponse class]);
+  OCMStub([fakeResponse statusCode]).andReturn(200);
```
maksymmalyhin, line 187: nit: using `OCMExpect` + `OCMVerifyAll` or `OCMVerify` may improve coverage by validating that the expected methods have been actually called.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -116,181 +123,160 @@ + (instancetype)appDistribution {
...
-                                                 NSString *_Nonnull idToken,
-                                                 NSError *_Nullable error) {
+  // Get a Firebase Installation ID (FID).
+  [installations installationIDWithCompletion:^(NSString *__nullable identifier,
+                                                NSError *__nullable error) {
     if (error) {
-      // TODO (schnecle): Add in FIRLogger log statement
-      NSLog(@"Error fetching fresh tokens: %@", [error localizedDescription]);
-      [self signOutTester];
+      NSString *description = error.userInfo[NSLocalizedDescriptionKey]
```
samedson, line 145: One thing to keep in mind is I believe this completion handler is called on the main thread @maksymmalyhin can confirm. We were worried about performance / thread locking up so we put the callback on a background thread: https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m#L108

Might not be necessary here but just a heads up

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -116,181 +123,160 @@ + (instancetype)appDistribution {
...
+                 (long)[error code]);
+  return [self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorUnknown
+                                     message:@"Failed to fetch releases for unknown reason."];
 }
+
+- (void)fetchNewLatestRelease:(FIRAppDistributionUpdateCheckCompletion)completion {
+  [FIRFADApiService
+      fetchReleasesWithCompletion:^(NSArray *_Nullable releases, NSError *_Nullable error) {
+        if (error) {
+          completion(nil, [self mapFetchReleasesError:error]);
```
samedson, line 247: This completion is not wrapped in a dispatch_async, while the one below is. Would recommend always calling the callback on the same thread as it can manifest as super funky bugs down the road.

FirebaseAppDistribution/Sources/FIRFADApiService.m
```diff
@@ -0,0 +1,201 @@
...
+                     code:FIRFADApiTokenGenerationFailure]) {
+      FIRFADErrorLog(@"Error getting fresh auth tokens. Error: %@", [error localizedDescription]);
+
+      completion(nil, nil, error);
+      return;
+    }
+
+    [installations installationIDWithCompletion:^(NSString *__nullable identifier,
+                                                  NSError *__nullable error) {
+      if ([self handleError:&error
```
samedson, line 51: Same comment here re: this is run on main thread. Might be totally fine though, but for our situation we needed to put the block on a background thread.

samedson: Didn't realize these were nested calls, this is probably fine

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: @maksymmalyhin can you expand on that? I have not had issues mocking static methods using OCMock and tend to air on the side of `if it's not holding state, make it a stateless static class`. Certainly a singleton pattern accomplishes the same goal, but I would love to dig in more to the testing impediments you've run into as well as your feelings around consistency and style for my own iOS learning.

maksymmalyhin: There are different points of view on it. Some of them are pretty radical like go/srl-dev/why-fakes#no-ocmock which basically means "never use OCMock". From the other side, from my personal experience, opting out of mocking completely slows down development pretty significantly when fakes used by a single test have to be implemented from scratch. But mocking class methods proved to be worse because it affects entire runtime and not limited to your particular test scope which may lead to unexpected interference with other tests or other unexpected issues.

Couple more things to consider for this particular case:
1. `FIRFADApiService` actually access and modifies different states but it does it via e.g. `NSURLSession` and `FIRInstallations` singletons
2. Using class methods makes the dependency on  `FIRFADApiService` more implicit. One has to look through the implementation in order to spot it
3. If the dependency injected, switching between mock and fake is much easier and would require only test code modifications

I think @samedson may provide a bit more details on the topic since Crashlytics team consciously opted out of using `OCMock`  in the product tests.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: At the moment, I don't believe we support multiple apps: https://github.com/firebase/firebase-ios-sdk/blob/619864e4cb965c0ac07163c7dd6cf080df6f1b48/FirebaseAppDistribution/Sources/FIRAppDistribution.m#L92

Is there somewhere else that indicates that we do support multiple apps?

maksymmalyhin: Ah, I see, sorry, I was confused by the presence of the [initializer accepting an app](https://github.com/firebase/firebase-ios-sdk/blob/619864e4cb965c0ac07163c7dd6cf080df6f1b48/FirebaseAppDistribution/Sources/FIRAppDistribution.m#L68).

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: I spoke with @samedson about this more, and I think I agree with both of you. I think given that this will be a bigger refactor, do you mind if we take this on as a subsequent change @maksymmalyhin?

maksymmalyhin: sure, not an issue from my end

FirebaseAppDistribution/Tests/Unit/FIRFADApiServiceTests.m
schnecle: Added verification to these tests in here - https://github.com/firebase/firebase-ios-sdk/pull/6138 which I will merge to this branch.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: No worries - @samedson  had some good feedback on readability here, and I think we can make this a lot clearer with a little refactoring, but we'll do that in a subsequent PR if that's ok as it will be another large one.

maksymmalyhin: SGTM

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: Sweet! Thank you!

FirebaseAppDistribution.podspec
schnecle: This is done.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: this is done.

PR #6136, opened 2020-07-27
Update recommended public header location

paulb777:
Update documentation to minimize Swift PM `#if`ing for client public header access.

### DIFF
--- a/HeadersImports.md
+++ b/HeadersImports.md
@@ -11,11 +11,11 @@ In Swift Package Manager, it's a library target.
 ## Header File Types and Locations - For Header File Creators
 
 * *Public Headers* - Headers that define the library's API. They should be located in
-  `FirebaseFoo/Sources/Public`. Any additions require a minor version update. Any changes or
-  deletions require a major version update.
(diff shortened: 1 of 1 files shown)


HeadersImports.md
```diff
@@ -86,3 +86,10 @@ inconsistent with expectations of C++ developers. "Private" headers are availabl
 via an explicit import. "Internal" or "Project" headers are only available to their enclosing
 library. Many file names in this repo include "Private" or "Internal" do not comply. Always
 check the build definition to see how the file is used.
+
+### Public Header Location Explanation
+
+CocoaPods flattens all headers specified in the podspec and makes them available via
+`#import "FirebaseFoo/Header.h"`. Swift Package Manager does not flatten. Therefore, the
+directory structure described above allows clients to import headers from either package manager
+with `#if` checks.
```
maksymmalyhin, line 95: Could you please provide a bit more details what `#if` checks are meant.

paulb777: Good catch! It was backwards. Is it clear as `without`?

maksymmalyhin: yep, looks great now!

PR #6139, opened 2020-07-27
fix messaging breakage

chliangGoogle:
Swizzling tests recently start failing after xcode version update. It reported the subclass mock of NotificationCenter was recreated by KVO. Mock singleton method the same way as it mock other class such as FIRMessaging, proxy class seems resolve the issue.

### DIFF
--- a/FirebaseMessaging/Tests/UnitTests/FIRMessagingRemoteNotificationsProxyTest.m
+++ b/FirebaseMessaging/Tests/UnitTests/FIRMessagingRemoteNotificationsProxyTest.m
@@ -159,8 +159,8 @@
   // Update +sharedProxy to always return our test instance
   OCMStub([_mockProxyClass sharedProxy]).andReturn(self.proxy);
   if (@available(macOS 10.14, iOS 10.0, *)) {
-    _mockUserNotificationCenter =
-        OCMPartialMock([UNUserNotificationCenter currentNotificationCenter]);
+    _mockUserNotificationCenter = OCMClassMock([UNUserNotificationCenter class]);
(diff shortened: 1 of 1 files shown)


FirebaseMessaging/Tests/UnitTests/FIRMessagingRemoteNotificationsProxyTest.m
```diff
@@ -159,8 +159,9 @@ - (void)setUp {
   // Update +sharedProxy to always return our test instance
   OCMStub([_mockProxyClass sharedProxy]).andReturn(self.proxy);
   if (@available(macOS 10.14, iOS 10.0, *)) {
-    _mockUserNotificationCenter =
-        OCMPartialMock([UNUserNotificationCenter currentNotificationCenter]);
+    _mockUserNotificationCenter = OCMClassMock([UNUserNotificationCenter class]);
```
maksymmalyhin, line 162: As far as I understand the error started appearing because of KVO implementation changes so now KVO explicitly ISA swizzles target classes which breaks OCMock - KVO compatibility. Probably `OCMock` doesn't care about swizzling (in contrast to `OCMPartialMock`) so the change solves the issue. Does it sound accurate? Just would like to understand the core reason.

chliangGoogle: I notice all singleton methods are stub to return class mock, even though the way OCMPartialMock handles it seems make more sense. Using the same pattern seems work for me, but I didn't know why. 
So I guess KVO now breaks OCPartialMock as it overwrites the subclass again?

maksymmalyhin: Yeah, looks like.

PR #6144, opened 2020-07-28
Initializing UIWindow using initWithWindowScene for ios 13 and removing UI code from main FIRAppDistribution.m

pranavrajgopal:
* Initializing UIWindow using initWithWindowScene for ios 13 
* Also making sure we only reset window in the App delegate if a registration flow is in progress.

#no-changelog this is being merged into another branch

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -278,7 +278,7 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
   if ([self isTesterSignedIn]) {
     [self fetchNewLatestRelease:completion];
   } else {
-    FIRFADUIActionCompletion yesActionCompletion = ^(UIAlertAction *action){
+    FIRFADUIActionCompletion yesActionCompletion = ^(UIAlertAction *action) {
       [self signInTesterWithCompletion:^(NSError *_Nullable error) {
         if (error) {
           completion(nil, error);
@@ -289,11 +289,12 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
       }];
     };
 
-    FIRFADUIActionCompletion noActionCompletion = ^(UIAlertAction *action){
+    FIRFADUIActionCompletion noActionCompletion = ^(UIAlertAction *action) {
       completion(nil, nil);
     };
 
-    [[self uiService] showUIAlertWithYesCompletion:yesActionCompletion withNoCompletion:noActionCompletion];
+    [[self uiService] showUIAlertWithYesCompletion:yesActionCompletion
+                                  withNoCompletion:noActionCompletion];
   }
 }
 @end
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -278,23 +278,24 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
   if ([self isTesterSignedIn]) {
     [self fetchNewLatestRelease:completion];
   } else {
-    FIRFADUIActionCompletion yesActionCompletion = ^(UIAlertAction *action) {
-      [self signInTesterWithCompletion:^(NSError *_Nullable error) {
-        if (error) {
-          completion(nil, error);
-          return;
-        }
+    FIRFADUIActionCompletion actionCompletion = ^(BOOL continued) {
+      if (continued) {
+        [self signInTesterWithCompletion:^(NSError *_Nullable error) {
+          if (error) {
+            completion(nil, error);
+            return;
+          }
 
-        [self fetchNewLatestRelease:completion];
-      }];
+          [self fetchNewLatestRelease:completion];
+        }];
+      } else {
+        completion(
+            nil, [self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorAuthenticationCancelled
+                                             message:@"Tester cancelled authentication flow."]);
+      }
     };
 
-    FIRFADUIActionCompletion noActionCompletion = ^(UIAlertAction *action) {
-      completion(nil, nil);
-    };
-
-    [[self uiService] showUIAlertWithYesCompletion:yesActionCompletion
-                                  withNoCompletion:noActionCompletion];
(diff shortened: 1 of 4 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -168,15 +168,37 @@ - (void)initializeUIState {
...
+#endif
   self.window.rootViewController = self.safariHostingViewController;
 
   // Place it at the highest level within the stack.
   self.window.windowLevel = +CGFLOAT_MAX;
 
   // Run it.
   [self.window makeKeyAndVisible];
+
+  if (self.window) {
```
jeremydurham, line 200: Do we need this?

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -168,15 +168,37 @@ - (void)initializeUIState {
   if (self.window) {
     return;
   }
-  // Create an empty window + viewController to host the Safari UI.
+
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
```
jeremydurham, line 172: is if (@available(iOS 13.0, *)) sufficient? Do we need this?

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -168,15 +168,37 @@ - (void)initializeUIState {
...
+      if (connectedScene.activationState == UISceneActivationStateForegroundActive) {
+        foregroundedScene = connectedScene;
+        break;
+      }
+    }
+
+    if (foregroundedScene) {
+      self.window = [[UIWindow alloc] initWithWindowScene:foregroundedScene];
+    } else {
+      // TODO: figure out if foregroundScene can be nil. If yes, should we return error?
```
jeremydurham, line 185: Maybe we can log here for now?

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
pranavrajgopal: Nope. sorry was trying something else and didnt delete this.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -71,8 +71,8 @@ - (instancetype)initWithApp:(FIRApp *)app appInfo:(NSDictionary *)appInfo {
 
   if (self) {
     [GULAppDelegateSwizzler proxyOriginalDelegate];
-    self.UIService = [FIRAppDistributionUIService sharedInstance];
-    [GULAppDelegateSwizzler registerAppDelegateInterceptor:self.UIService];
```
pranavrajgopal, line 75: @jeremydurham UIService or uiService? What's the convention? I decided to change it back to UIService earlier based on Jeremy's feedback.

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
schnecle: Removed.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -277,39 +278,23 @@ - (void)checkForUpdateWithCompletion:(FIRAppDistributionUpdateCheckCompletion)co
...
-                                                       // precaution to ensure window gets destroyed
-                                                       [self.UIService resetUIState];
-                                                       completion(nil, nil);
-                                                     }];
-
-    [alert addAction:noButton];
-    [alert addAction:yesButton];
-
-    // Create an empty window + viewController to host the Safari UI.
-    [self.UIService showUIAlert:alert];
```
pranavrajgopal, line 312: Since we don't actually use the UIAlertAction we could simplify this `showUIAlert` to have a single completion block with a `userCancelled` boolean - Something like this - 
```
- (void)showUIAlertWithCompletion:(void (^)(BOOL userCancelled))completion {
  UIAlertController *alert = [UIAlertController
      alertControllerWithTitle:@"Enable in-app alerts"
                       message:@"Sign in with your Firebase App Distribution Google account to "
                               @"turn on in-app alerts for new test releases."
                preferredStyle:UIAlertControllerStyleAlert];

  UIAlertAction *yesButton = [UIAlertAction actionWithTitle:@"Turn on"
                                                      style:UIAlertActionStyleDefault
                                                    handler:^(UIAlertAction *action) {
                                                        completion(false);
                                                    }];

  UIAlertAction *noButton = [UIAlertAction actionWithTitle:@"Not now"
                                                     style:UIAlertActionStyleDefault
                                                   handler:^(UIAlertAction *action) {
                                                     [self resetUIState];
                                                     completion(true);
                                                   }];

  [alert addAction:noButton];
  [alert addAction:yesButton];

  // Create an empty window + viewController to host the Safari UI.
  [self showUIAlert:alert];
}
```

You can call this method like this - 
```
    
    [self showUIAlertWithCompletion:^(BOOL userCancelled) {
        if(!userCancelled) {
            [self signInTesterWithCompletion:^(NSError *_Nullable error) {
              if (error) {
                completion(nil, error);
                return;
              }

              [self fetchNewLatestRelease:completion];
            }];
        } else {
            completion(nil, nil);
        }
    }];
```

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.h
```diff
@@ -44,6 +53,9 @@ typedef void (^AppDistributionRegistrationFlowCompletion)(NSError *_Nullable err
 
 - (void)showUIAlert:(UIAlertController *)alertController;
```
pranavrajgopal, line 55: This could be potentially simplified to use a single completion handler. See comment above.

FirebaseAppDistribution/Sources/FIRAppDistribution.m
schnecle: ooo, yeah, I like that a lot!

FirebaseAppDistribution/Sources/FIRAppDistribution.m
jeremydurham: I think Cleo (and you) are right. uiService makes sense (similar to httpResponse)

PR #6190, opened 2020-08-03
Add localized strings to the UIAlert

schnecle:
* Add NSLocalizedStrings to the UIAlert component

#no-changelog App Distribution is not yet public

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
@@ -142,22 +142,28 @@ SFAuthenticationSession *_safariAuthenticationVC;
 
 - (void)showUIAlertWithCompletion:(FIRFADUIActionCompletion)completion {
   UIAlertController *alert = [UIAlertController
-      alertControllerWithTitle:@"Enable new build alerts"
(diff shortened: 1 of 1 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -142,22 +142,28 @@ - (void)showUIAlert:(UIAlertController *)alertController {
...
-                                                   }];
+  UIAlertAction *yesButton = [UIAlertAction
+      actionWithTitle:NSLocalizedString(@"Turn on", @"Button for turning on new build alerts.")
+                style:UIAlertActionStyleDefault
+              handler:^(UIAlertAction *action) {
+                completion(YES);
+              }];
+
+  UIAlertAction *noButton = [UIAlertAction
+      actionWithTitle:NSLocalizedString(@"Not now", @"Button for dismissing the UIAlert")
```
jeremydurham, line 161: Should the comment be something like "Button for dismissing the new build alerts UIAlert."

schnecle: I like that - more parallel structure.

PR #6192, opened 2020-08-03
Use NSNumberFormatter for number values in remote config

### LINKED ISSUE #6188: FIRRemoteConfigValue doubleValue never returns nil

  * Xcode version: `11.6`
  * Firebase SDK version: `6.6.7`
  * Firebase Component: `RemoteConfig`
  * Component version: `6.23.0`

#### Steps to reproduce:

Get the config value for a non existent key with no default. Read `numberValue` from the config value. I expect this to return `nil`, but it always returns `0`.

From the [source](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/FIRConfigValue.m#L50), this value cannot return `nil`.

If `stringValue` is `nil`, `stringValue.numberValue` will return `nil`, which will be implicitly casted to the double `0`.

If `stringValue` is not `nil` and not a valid number, `stringValue.numberValue` will return the double `0`, as per Apple's documentation.

#### Relevant Code:

```swift
remoteConfig.configValue(forKey: "non-existent-key-with-no-defaults").numberValue == nil // fails
```

morganchen12:
Fixes #6188.

### DIFF
--- a/FirebaseRemoteConfig/CHANGELOG.md
+++ b/FirebaseRemoteConfig/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v4.8.2
+- [fixed] Fixed an issue where numerical values would always be zero instead of nil (#6188).
+
 # v4.8.1
 - [fixed] Fixed `FirebaseApp.delete()` related crash in `RC Config Fetch`. (#6123)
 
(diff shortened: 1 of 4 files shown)


morganchen12:
I think it's a bug because the header annotation specifies the return type is `nullable`. Either it should be able to return nil, or it should be marked as nonnull. I am not sure what Android does. @karenyz do you know?

morganchen12:
Unassigning reviewers based on https://github.com/firebase/firebase-ios-sdk/issues/6188#issuecomment-668785165

PR #6201, opened 2020-08-04
Fix cpplint.py to work under Python 3

dconeybe:
In #6114 the default Python version was upgraded from Python 2 to Python 3 for some GitHub Actions. This appeared to work; however, `cpplint.py` was subtly broken. As a temporary workaround, #6198 was merged to force `cpplint.py` to run under Python 2.7. This PR fixes `cpplint.py` when run under Python 3 and reverts the change to force it to run under Python 2.

The "fix" was to simply grab the latest version of `cpplint.py` from https://github.com/cpplint/cpplint/blob/develop/cpplint.py, which, at the time of writing, was commit 7313f7782cfad9c1a4c41700b5a0d1ff693862b5. Then, the local modifications that were made to `cpplint.py` were applied to this upgraded version.


wilhuff:
I've verified that `python3 ./scripts/check_lint.py --trace Firestore/core` works with this change and is indeed running cpplint with python3.

PR #6204, opened 2020-08-04
update nanopb from 0.3.9.5 to 0.3.9.6

vimanyu:
This fixes issues building firestore SDK with Python3.
Interestingly, latest version  nanopb-0.4.2 still errors with Python3 but 0.3.9.6 works. 
The git commit for this fix in nanopb
https://github.com/nanopb/nanopb/commit/63f8e5e4946a5f8f0211b454bd37187871c3d023

### Discussion
Users ran into issues when using Python3 for building firestore
https://github.com/firebase/firebase-cpp-sdk/issues/48

### Testing

### API Changes

    can discuss it together.


paulb777:
Might be risky to update nanopb only for Firestore CPP without doing a Firebase-wide update?

wilhuff:
Firestore also supports building this code via CocoaPods, so the generated sources have to be checked in. See instructions: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Protos/README.md. When I run it on your branch all I see are version diffs like these:

```diff
diff --git a/Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc b/Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
index 154950aab..04b01a948 100644
--- a/Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
+++ b/Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
@@ -15,7 +15,7 @@
  */
 
 /* Automatically generated nanopb constant definitions */
-/* Generated by nanopb-0.3.9.5 */
+/* Generated by nanopb-0.3.9.6 */
 
 #include "maybe_document.nanopb.h"
```

Given that there's no change to the generated code it seems like it's possible to just upgrade C++, but our preference is to upgrade across Firebase because it's hard to know the knock on effects of version skew. There are a few places to update to make this all work:

  * Update the version in google3/third_party/nanopb
  * Update the [nanopb-podspec](https://github.com/google/nanopb-podspec)
  * Update the versions in pod dependencies

The third_party change is required to make the desktop build of Firestore match what we're testing here. Unfortunately, once you change third_party, any binary Firebase pods will now build/test against that, which means the nanopb pod should be updated. That in turn forces us to upgrade all the open-source Firebase pods to match.

wilhuff:
Since I had already run the source generator I've just pushed that update on this branch. Make sure to pull before you make additional changes.

paulb777:
Here's an example of the version dependencies that need to be updated across this repo and g3: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCrashlytics.podspec#L54.  Any pod that is updated should also have a minor version update for its own version.

Docs for staging podspecs for testing are at https://guides.cocoapods.org/making/private-cocoapods.html and you can use the https://github.com/firebase/SpecsStaging repo.

vimanyu:
There are several merge conflicts. Am going to close this one and reopen a new PR with a clean set of changes.

PR #6206, opened 2020-08-04
Update CHANGELOG for Firestore v1.16.4

wilhuff:


wilhuff:
I note that the release spreadsheet has this change down as 1.17.0 though there's no change that would seem to warrant that. Does this need to be 1.17.0?

Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+# v1.16.4
```
paulb777, line 3: I'd like to be conservative and bump to 1.17.0 for more flexibility

wilhuff: What does "more flexibility" mean exactly? The downside of bumping is that this is the end of patch releases that can be trivially picked up by the C++ SDK that builds on top of this.

paulb777:
We're making a change that is not a fix to facilitate a future feature - so it's not really a patch.

Firestore/CHANGELOG.md
wilhuff: I should clarify: C++ releases are going to be more infrequent than iOS ones so I'd like preserve the flexibility to release fixes there to those users.

Meanwhile, this change is functionally transparent to iOS users. It's not an addition or a removal--and it doesn't matter what our intent is.

paulb777: OK, it should be fine to leave as a patch for Firestore.  I feel more strongly about pods with intra-repo dependencies for flexibility in being able to patch if something does go wrong and minimizing the version exponential surface area with different file structures.

PR #6207, opened 2020-08-04
FirebaseInstallations: API requests backoff

maksymmalyhin:
Context: b/160751425.

- exponential backoff for recoverable API errors.
- constant 24 hours backoff for invalid Firebase configuration errors.
- some tests. Tests for remaining cases will be added in the following PRs.


FirebaseInstallations/Source/Library/InstallationsIDController/FIRCurrentDateProvider.m
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "FirebaseInstallations/Source/Library/InstallationsIDController/FIRCurrentDateProvider.h"
+
+FIRCurrentDateProvider FIRRealCurrentDateProvider(void) {
```
paulb777, line 19: Is the date provider only for testability? Or is there any other reason to have something different than `[NSDate date]`?

maksymmalyhin: Yes, it's for testability only. `[NSDate date]` and similar methods are so widely used by SDKs and system that mocking it often leads to weird flaky tests failures. Ideally we should never mock `NSDate` methods. Let me add explanation to the comments as well.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsBackoffController.h
```diff
@@ -0,0 +1,54 @@
...
+typedef NS_ENUM(NSInteger, FIRInstallationsBackoffEvent) {
+  FIRInstallationsBackoffEventSuccess,
+  FIRInstallationsBackoffEventRecoverableFailure,
+  FIRInstallationsBackoffEventUnrecoverableFailure
+};
+
+/** The protocol defines API for a class that encapsulates backoff logic that prevents the SDK from
+ * sending unnecessary server requests. See API docs for the methods for more details. */
+
+@protocol FIRInstallationsBackoffControllerProtocol <NSObject>
```
chliangGoogle, line 32: I wonder if it's necessary to be a protocol since only class is adopting it, or so that in the future any class wanna do backoff can adopt it.

maksymmalyhin: We've had some discussions around an opinion expressed at go/srl-dev/why-fakes#no-ocmock recently. There are some pros and cons of using either mocks or fakes, so didn't make a final decision for Firebase. Defining protocols for dependencies has the following advantages:
- switching between mocks and fakes in tests will not require the production code changes
- dependency on a protocol allows more clearly separate API from implementation details
- even if we use mocks, a mock for a protocol looks cleaner

Do you think these advantages justify the protocol?

maksymmalyhin:
I'm merging the PR, but feel free to leave comments to be addressed in the following PRs to the feature branch.

FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsBackoffController.h
chliangGoogle: Yeah, I didn't know protocol can also help with separating mocking in tests. That's a new thing to learn.

PR #6209, opened 2020-08-05
Fix topic in watchOS

### LINKED ISSUE #6160: Topics do not work with watchOS
If you attempt to sub to a topic, the callback never calls indicating a success or failure.

The reason is due to `self.pubsub` being undefined, and a message is being sent to `self.pubsub` with the completion without a null check:
https://github.com/firebase/firebase-ios-sdk/blob/ad06c605bf08d1649d266f41116f5b3c46345e99/FirebaseMessaging/Sources/FIRMessaging.m#L781

There's this line:
https://github.com/firebase/firebase-ios-sdk/blob/ad06c605bf08d1649d266f41116f5b3c46345e99/FirebaseMessaging/Sources/FIRMessaging.m#L273
Which blocks initialization of `self.pubsub` - however when this is removed, there appears to be no problems. Topics work fine on a standalone watchOS after removing these preprocessor statements. 

Is there any particular reason as to why? I'm assuming something else is not working properly with notifications is why - I will gladly help update the code if need be to get this working, as topics are something we need on our end.

chliangGoogle:
Fix #6160.
We couldn't enable watchOS for topic directly as topic (the FIRMessagingPubsub class) depends FIRMessagingClient which is mostly for direct channel. But topic doesn't really need direct channel to function so refactor to let FIRMessagingPubsub class independent from FIRmessagingClient. Also remove pubsubRegistrar class that is an unnecessary layer of class of repeat FIRMessagingPubSub methods.

Test: Unit test and manual test on independent watch app.

Also remove cancelAllRequests method that is not called anywhere.


FirebaseMessaging/Sources/FIRMMessageCode.h
```diff
@@ -47,7 +47,7 @@ typedef NS_ENUM(NSInteger, FIRMessagingMessageCode) {
   kFIRMessagingMessageCodeTokenDelegateMethodsNotImplemented = 2023,         // I-FCM002023
   kFIRMessagingMessageCodeTopicFormatIsDeprecated = 2024,
   kFIRMessagingMessageCodeDirectChannelConnectionFailed = 2025,
-  kFIRMessagingMessageCodeInvalidClient = 2026, // no longer used
+  kFIRMessagingMessageCodeInvalidClient = 2026,  // no longer used
```
morganchen12, line 50: nit: this whitespace change looks accidental

FirebaseMessaging/Sources/FIRMessagingPubSub.m
```diff
@@ -89,17 +88,18 @@ - (void)subscribeWithToken:(NSString *)token
...
-                              shouldDelete:NO
-                                   handler:^void(NSError *error) {
-                                     handler(error);
-                                   }];
+                              topic:topic
+                            options:options
+                       shouldDelete:NO
+                            handler:^void(NSError *error) {
+                              handler(error);
+                            }];
```
morganchen12, line 96: nit: Can you pass the handler directly into this parameter instead of wrapping it in another block?

FirebaseMessaging/Sources/FIRMessagingPubSub.m
```diff
@@ -126,11 +126,17 @@ - (void)updateSubscriptionWithToken:(NSString *)token
...
   if ([[FIRInstanceID instanceID] tryToLoadValidCheckinInfo]) {
-    [self updateSubscriptionToTopic:topic
-                                    withToken:token
-                                      options:options
-                                 shouldDelete:shouldDelete
-                                      handler:completion];
+    FIRMessagingTopicAction action = FIRMessagingTopicActionSubscribe;
+    if (shouldDelete) {
+      action = FIRMessagingTopicActionUnsubscribe;
+    }
```
morganchen12, line 132: _optional suggestion_: This can be made into a ternary:

```objc
FIRMessagingTopicAction action = shouldDelete ? 
    FIRMessagingTopicActionUnsubscribe : FIRMessagingTopicActionSubscribe
```

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
```diff
@@ -34,11 +34,19 @@ class ExtensionDelegate: NSObject, WKExtensionDelegate, MessagingDelegate {
   /// MessagingDelegate
   func messaging(_ messaging: Messaging, didReceiveRegistrationToken fcmToken: String) {
     print("token:\n" + fcmToken)
+    Messaging.messaging().subscribe(toTopic: "watch") { error in
+      if error != nil {
+        print("error:" + error.debugDescription)
+      } else {
+        print("Successfully subscribe to topic")
```
paulb777, line 41: subscribed

FirebaseMessaging/Sources/FIRMessagingPubSub.m
chliangGoogle: Done. Good catch.

FirebaseMessaging/Sources/FIRMMessageCode.h
chliangGoogle: Done. Not sure if it's clang-format, I revert the change and will run the script again to see.

FirebaseMessaging/Sources/FIRMessagingPubSub.m
chliangGoogle: Done.

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
chliangGoogle: Done.

FirebaseMessaging/Sources/FIRMessagingPubSub.m
chliangGoogle: Done.

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
chliangGoogle: Done.

PR #6225, opened 2020-08-07
Update podfile to point to live releases for EAP customers

schnecle:
Rollback podfile to point at live releases to facilitate EAP testing off of master.

#no-changelog App Distribution is not yet included in the public release


schnecle:
No apologies @paulb777 ! Thank you for bumping versions and keeping everyone up to date! It's an appreciated thing!

PR #6229, opened 2020-08-07
Update swiftformat and control its version via mint

paulb777:
Use mint to install a specific version of swiftformat
Add GHA caching of swiftformat
Update swiftformat to 0.45.3

#no-changelog

### DIFF
--- a/Firestore/third_party/FirestoreEncoder/FirestoreDecoder.swift
+++ b/Firestore/third_party/FirestoreEncoder/FirestoreDecoder.swift
@@ -388,24 +388,24 @@ private struct _FirestoreKeyedDecodingContainer<K: CodingKey>: KeyedDecodingCont
   public func nestedContainer<NestedKey>(keyedBy _: NestedKey.Type,
                                          forKey key: Key) throws
     -> KeyedDecodingContainer<NestedKey> {
-    decoder.codingPath.append(key)
-    defer { self.decoder.codingPath.removeLast() }
+      decoder.codingPath.append(key)
+      defer { self.decoder.codingPath.removeLast() }
 
-    guard let value = self.container[key.stringValue] else {
-      throw DecodingError.valueNotFound(KeyedDecodingContainer<NestedKey>.self,
-                                        DecodingError.Context(codingPath: codingPath,
-                                                              debugDescription: "Cannot get nested keyed container -- no value found for key \"\(key.stringValue)\""))
+      guard let value = self.container[key.stringValue] else {
+        throw DecodingError.valueNotFound(KeyedDecodingContainer<NestedKey>.self,
+                                          DecodingError.Context(codingPath: codingPath,
+                                                                debugDescription: "Cannot get nested keyed container -- no value found for key \"\(key.stringValue)\""))
+      }
+
+      guard let dictionary = value as? [String: Any] else {
+        throw DecodingError
+          ._typeMismatch(at: codingPath, expectation: [String: Any].self, reality: value)
+      }
+
+      let container = _FirestoreKeyedDecodingContainer<NestedKey>(referencing: decoder,
+                                                                  wrapping: dictionary)
+      return KeyedDecodingContainer(container)
     }
-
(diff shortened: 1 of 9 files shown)


Firestore/Swift/Tests/Codable/FirestoreEncoderTests.swift
```diff
@@ -609,17 +609,20 @@ class FirestoreEncoderTests: XCTestCase {
 private func assertThat(_ dictionary: [String: Any],
                         in document: String? = nil,
                         file: StaticString = #file,
-                        line: UInt = #line) -> DictionarySubject {
+                        line: UInt = #line) -> DictionarySubject
+{
```
wilhuff, line 613: Was there a configuration change associated with this? Common Swift style (and Google's go/swift-style specify that braces should trail rather than start a new line.

paulb777: Nope, it looks like we might need to add an option in that case.

paulb777:
Blocked on a solution or workaround to https://github.com/nicklockwood/SwiftFormat/issues/695

Firestore/Swift/Tests/Codable/FirestoreEncoderTests.swift
paulb777: Resolved via https://github.com/nicklockwood/SwiftFormat/issues/696

Firestore/third_party/FirestoreEncoder/FirestoreDecoder.swift
```diff
@@ -388,24 +388,24 @@ private struct _FirestoreKeyedDecodingContainer<K: CodingKey>: KeyedDecodingCont
...
+      }
 
-    let container = _FirestoreKeyedDecodingContainer<NestedKey>(referencing: decoder,
-                                                                wrapping: dictionary)
-    return KeyedDecodingContainer(container)
-  }
+      let container = _FirestoreKeyedDecodingContainer<NestedKey>(referencing: decoder,
+                                                                  wrapping: dictionary)
+      return KeyedDecodingContainer(container)
+    }
```
wilhuff, line 408: The indentation of this method is now two spaces too far. This seems to be the only case that's broken this way so I'm OK with moving forward if there's no obvious fix.

scripts/setup_check.sh
```diff
@@ -38,3 +32,13 @@ if ! git rev-parse origin/master >& /dev/null; then
   git remote set-branches --add origin master
   git fetch origin
 fi
+
+# Setup mint which installs a versioned swiftformat.
+set -x
+brew upgrade mint || brew install mint || true
```
wilhuff, line 38: Why can't this just be `brew install mint`? If it is installed, does it need to be upgraded? If so, could you add a comment here about what's going on? (Note the difference with `clang-format`, below.)

scripts/setup_check.sh
```diff
@@ -38,3 +32,13 @@ if ! git rev-parse origin/master >& /dev/null; then
...
+mint bootstrap --link
+
+brew install clang-format
```
maksymmalyhin, line 41: AFAIK this command will fail if `clang-format` is already installed. Also I wonder if `clang-format` can be installed with mint as well?

Firestore/third_party/FirestoreEncoder/FirestoreDecoder.swift
paulb777: Filed https://github.com/nicklockwood/SwiftFormat/issues/697

scripts/setup_check.sh
paulb777: I checked. Unfortunately no -but at least clang-format seems more stable ...

scripts/setup_check.sh
paulb777: I made the two brew installs consistent with the old way we were doing it.

paulb777:
Going to merge it here so CI isn't broken over the weekend. If there's anything else, I'll address in a subsequent PR.

PR #6233, opened 2020-08-07
Add SPM support for CoreDiagnostics

paulb777:
- Repo-relative headers
- SPM Diagnostics logging
- SPM build
- Skipping standard directory structure because of upcoming deprecation

Verified deploymentType running the Analytics quickstart:

Screen Shot 2020-08-07 at 2 17 49 PM


paulb777:
Maybe? but as pointed out in the description, we are planning to delete this library.

ryanwilson:
> Maybe? but as pointed out in the description, we are planning to delete this library.

Whoops, good call. SGTM.

PR #6236, opened 2020-08-08
Add Hashable conformance to wrappers

### LINKED ISSUE #6237: Add Hashable conformance to firestoreswift property wrappers
### Feature proposal

* Firebase Component: FirestoreSwift

Add Hashable conformance to property wrappers.

I did it in #6236

lorenzofiamingo:
Hello,
I think would be great to add the Hashable conformance to property wrappers when their wrapped value is Hashable, just like Apple does for arrays, optionals, and ranges
In this way, a struct using these wrappers can easily inherit the Hashable conformance by default.
This is really useful in SwiftUI when you pull from Firestore a collection as an array of these Codable structs and use them in a ForEach view.

Fix #6237

### DIFF
--- a/Firestore/Swift/Source/Codable/DocumentID+Hashable.swift
+++ b/Firestore/Swift/Source/Codable/DocumentID+Hashable.swift
@@ -1,0 +1,19 @@
+//
+//  DocumentID+Hashable.swift
+//  
+//
+//  Created by Lorenzo Fiamingo on 08/08/20.
+//
+
+import FirebaseFirestore
+
+extension DocumentID: Hashable where Value: Hashable {
+    
+    static func == (lhs: DocumentID, rhs: DocumentID) -> Bool {
+        lhs.wrappedValue == rhs.wrappedValue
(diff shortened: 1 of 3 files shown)


paulb777:
Also the indent level should be 2 to pass the static analysis checks.  See https://github.com/firebase/firebase-ios-sdk#code-formatting

wilhuff:
Thanks for the feature request and the PR!

Instead of hand-writing these implementations, the Swift compiler can generate them (we were already using this for `Equatable` in some cases). I've sent out #6248 to correctly document our actual Swift version requirement and #6249 that implements this feature that way and adds tests.

lorenzofiamingo:
Great!
Since I added initially conformance in an extension outside the Package, the implementation of 'Hashable' couldn't be automatically synthesized, consequently, I hadn't noticed that... 🙃

wilhuff:
The changes in #6249 have been merged and will go out with the next release. Thanks again for raising this!

PR #6243, opened 2020-08-10
Use nanopb repo instead of personal fork

paulb777:
#no-changelog

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -98,8 +98,9 @@ let package = Package(
     ),
     .package(
       name: "nanopb",
-      url: "https://github.com/paulb777/nanopb.git",
-      .branch("spm2")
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -98,8 +98,9 @@ let package = Package(
     ),
     .package(
       name: "nanopb",
-      url: "https://github.com/paulb777/nanopb.git",
-      .branch("spm2")
+      url: "https://github.com/nanopb/nanopb.git",
+      // This revision adds SPM enablement to the 0.3.9.6 release tag.
+      .revision("3cfa212")
```
wilhuff, line 103: This reference can eventually become ambiguous. Better to use the full commit hash.

PR #6245, opened 2020-08-10
SPM - DynamicLinks

paulb777:
Adds Swift Package Manager support for Dynamic Links

- Move public headers for usage compatibility between CocoaPods and SPM
- ifdef out source builds for non-iOS platforms. There's no way to selectively choose platforms in SPM

#no-changelog


FirebaseDynamicLinks/Sources/FDLURLComponents/FDLURLComponents.m
```diff
@@ -693,3 +696,5 @@ + (NSURLRequest *)shorteningRequestForLongURL:(NSURL *)url
 }
 
 @end
+
+#endif
```
ryanwilson, line 700: Only nit would be to have `#endif  // TARGET_OS_IOS` throughout

PR #6247, opened 2020-08-10
Remove the extension that makes NSDate ServerTimestampWrappable.

wilhuff:
This turns out not to actually work because `NSDate` does not conform to `Codable`. Users can supply their own extensions that make `NSDate` conform to `Codable` and also supply an extension like this one to make `NSDate` conform to `ServerTimestampWrappable`.

Add a test that verifies ServerTimestamps of Swift `Date` actually work. The equivalent test for `NSDate` does not compile.


wilhuff:
Discovered while adding tests for the PR addressing #6237.

PR #6253, opened 2020-08-11
Switch from personal to Firebase forks

paulb777:
#no-changelog


paulb777:
Good point. Branches aren't locked. They would need to be versioned themselves for it to be sustainable.  Changed back to revisions so that the Firebase `revision` dependencies are not changeable without an update to Firebase's Package.swift.

PR #6255, opened 2020-08-11
[App Distribution] Show new build alert when version (or build) changes

jeremydurham:
When the version (or build) of the current app is different than the version or build of the latest release, show the prompt to update to the latest version.

#no-changelog app distibution is not released yet

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -248,59 +248,39 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
 }
 
 - (void)fetchNewLatestRelease:(FIRAppDistributionUpdateCheckCompletion)completion {
-  [FIRFADApiService fetchReleasesWithCompletion:^(NSArray *_Nullable releases,
-                                                  NSError *_Nullable error) {
-    if (error) {
-      dispatch_async(dispatch_get_main_queue(), ^{
-        completion(nil, [self mapFetchReleasesError:error]);
-      });
-      return;
-    }
-
-    for (NSDictionary *releaseDict in releases) {
-      if ([[releaseDict objectForKey:kLatestReleaseKey] boolValue]) {
-        FIRFADInfoLog(@"Tester API - found latest release in response.");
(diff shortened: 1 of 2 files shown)


FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -238,39 +248,59 @@ - (NSError *_Nullable)mapFetchReleasesError:(NSError *)error {
...
+    }
+
+    for (NSDictionary *releaseDict in releases) {
+      if ([[releaseDict objectForKey:kLatestReleaseKey] boolValue]) {
+        FIRFADInfoLog(@"Tester API - found latest release in response.");
+        NSString *displayVersion = [releaseDict objectForKey:kDisplayVersionKey];
+        NSString *buildVersion = [releaseDict objectForKey:kBuildVersionKey];
+
+        FIRFADInfoLog(@"Checking if version matches");
+        if (![self isCurrentVersion:displayVersion buildVersion:buildVersion]) {
```
schnecle, line 267: readability: I wonder if this would be more readable if we extracted the two checks into separate functions - so that we could have a single if -- we can still add the logging in the individual methods, but then we can have something like 

```
if (![self isCurrentVersion:releaseDict] || ![self codeIsIdentical:releaseDict]){
    FIRAppDistributionRelease *release =
              [[FIRAppDistributionRelease alloc] initWithDictionary:releaseDict];
            FIRFADInfoLog(@"Found new release with version: %@", [release displayVersion]);
            completion(release, nil);
        });
    return;
}
```

FirebaseAppDistribution/Tests/Unit/FIRAppDistributionTests.m
```diff
@@ -307,6 +313,8 @@ - (void)testFetchNewLatestReleaseSuccess {
 - (void)testFetchNewLatestReleaseNoNewRelease {
   [self mockFetchReleasesCompletion:_mockReleases error:nil];
   OCMStub([_mockMachO codeHash]).andReturn(_mockCodeHash);
+  OCMStub([[self appDistribution] getAppVersion]).andReturn(_mockDisplayVersion);
```
schnecle, line 316: You can also stub the main bundle directly with a class mock, but this works too 😄

PR #6257, opened 2020-08-11
SPM - FIAM

paulb777:
Adds Swift Package Manager support for In App Messaging

- Repo-relative headers
- Move public headers for usage compatibility between CocoaPods and SPM
- ifdef out source builds for non-iOS platforms. There's no way to selectively choose platforms in SPM
- remove cruft from podspec


FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/InAppMessagingDisplay-Sample.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 51;
+	objectVersion = 50;
```
ryanwilson, line 6: Hmm, any idea why this changed? Should it have?

Package.swift
```diff
@@ -66,6 +66,10 @@ let package = Package(
       name: "FirebaseFunctions",
       targets: ["FirebaseFunctions"]
     ),
+    .library(
+      name: "FirebaseInAppMessaging-Beta",
```
ryanwilson, line 70: Let's chat a bit more about this offline - I'm concerned about the transition when it goes GA that it'll be a big break requiring migration, and we may have to keep that `beta` around until the next breaking change.

FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessaging.h
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+#import <TargetConditionals.h>
+#if TARGET_OS_IOS
+
```
ryanwilson, line 19: I'm good with this now but I think we should discuss how we can make this a better experience for developers by [marking our APIs](https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/marking_api_availability_in_objective-c) with `API_AVAILABLE(ios(10))` and the like. That way they get nice messages with compiler errors instead of "missing symbol".

paulb777: I'll experiment with this.

FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/InAppMessagingDisplay-Sample.xcodeproj/project.pbxproj
paulb777: I might have had an old version on the command line from the release. Will revert.

Package.swift
paulb777: One option is documenting how to change in the Build Phase UI

FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessaging.h
paulb777: Actually it should be `API_UNAVAILABLE(macos,tvos,watchos)`.  Added a post-beta task to the project to revisit this.

Package.swift
paulb777: Will flush out the mechanics of the beta->GA flow in a subsequent task and doc.

paulb777: Screen Shot 2020-08-12 at 9 34 22 AM

PR #6261, opened 2020-08-12
Run swiftformat via mint run

wilhuff:
This avoids users needing to bootstrap manually and avoids permission problems with installation in /usr/local.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,16 +137,15 @@ See [HeadersImports.md](HeadersImports.md).
 ### Code Formatting
 
 To ensure that the code is formatted consistently, run the script
-[./scripts/style.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/style.sh)
+[./scripts/check.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check.sh)
 before creating a PR.
 
 Travis will verify that any code changes are done in a style compliant way. Install
-`clang-format` and `swiftformat`:
+`clang-format` and `mint`:
 
 ```
 brew install clang-format
 brew install mint
-mint install swiftformat@0.45.5
 ```
 
 ### Running Unit Tests
--- a/README.md
+++ b/README.md
@@ -140,8 +140,8 @@ To ensure that the code is formatted consistently, run the script
 [./scripts/check.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check.sh)
 before creating a PR.
 
-Travis will verify that any code changes are done in a style compliant way. Install
-`clang-format` and `mint`:
+GitHub Actions will verify that any code changes are done in a style compliant
+way. Install `clang-format` and `mint`:
 
 ```
 brew install clang-format
--- a/scripts/setup_check.sh
+++ b/scripts/setup_check.sh
@@ -33,12 +33,6 @@ if ! git rev-parse origin/master >& /dev/null; then
   git fetch origin
 fi
 
-# Setup mint which installs a versioned swiftformat.
-set -x
-brew install mint
-mint bootstrap --link
-
+# mint installs tools from Mintfile on demand.
 brew install clang-format
-
-clang-format -version
-swiftformat -version
+brew install mint
--- a/scripts/setup_check.sh
+++ b/scripts/setup_check.sh
@@ -36,3 +36,6 @@ fi
 # mint installs tools from Mintfile on demand.
 brew install clang-format
 brew install mint
+
+# Explicitly mint bootstrap to show its version in the "Setup check" GHA phase
+mint bootstrap
--- a/scripts/style.sh
+++ b/scripts/style.sh
@@ -55,20 +55,23 @@ case "$version" in
     ;;
 esac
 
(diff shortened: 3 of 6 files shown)


wilhuff:
I expect `check` to fail because I've introduced an intentional formatting error. I'll fix that once I see the failure.

scripts/setup_check.sh
```diff
@@ -33,12 +33,6 @@ if ! git rev-parse origin/master >& /dev/null; then
...
 
-# Setup mint which installs a versioned swiftformat.
-set -x
-brew install mint
-mint bootstrap --link
-
+# mint installs tools from Mintfile on demand.
 brew install clang-format
-
-clang-format -version
```
ryanwilson, line 43: I think having the `-version` command for `clang-format` at least was helpful for debugging purposes - I imagine we can get rid of the `swiftformat` one since we know where it comes from at least. Any reason you removed the `clang-format` one otherwise?

scripts/style.sh
```diff
@@ -55,21 +55,24 @@ case "$version" in
...
+
+function swiftformat() {
+  mint run swiftformat "$@"
+}
+
 system=$(uname -s)
 if [[ "$system" == "Darwin" ]]; then
   version=$(swiftformat --version)
   # Log the version in non-interactive use as it can be useful in travis logs.
+  # A version check is not requried because `mint bootstrap` ensures the
```
ryanwilson, line 71: s/requried/required/`

scripts/style.sh
```diff
@@ -55,21 +55,24 @@ case "$version" in
...
+# /usr/local.
+export MINT_PATH=Mint
+
+function swiftformat() {
+  mint run swiftformat "$@"
+}
+
 system=$(uname -s)
 if [[ "$system" == "Darwin" ]]; then
   version=$(swiftformat --version)
```
morganchen12, line 69: This (and the below) can probably be omitted as well, since the swiftformat version will be enforced by the Mintfile. Maybe instead they can be replaced by a check for a local Mint installation?

scripts/style.sh
paulb777, line 71: typo: required

README.md
```diff
@@ -137,16 +137,15 @@ See [HeadersImports.md](HeadersImports.md).
 ### Code Formatting
 
 To ensure that the code is formatted consistently, run the script
-[./scripts/style.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/style.sh)
+[./scripts/check.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check.sh)
 before creating a PR.
 
 Travis will verify that any code changes are done in a style compliant way. Install
```
paulb777, line 143: s/Travis/Git Hub Actions/

scripts/style.sh
```diff
@@ -55,21 +55,24 @@ case "$version" in
     ;;
 esac
 
+# Ensure that tools in `Mintfile` are installed locally to avoid permissions
+# problems that would otherwise arise from the default of installing in
+# /usr/local.
+export MINT_PATH=Mint
+
+function swiftformat() {
```
ryanwilson, line 63: It looks like we only use this twice, and it may be less obvious if a developer goes to run this locally. Any objection to just changing the `swiftformat` to `mint run swiftformat` in the script below?

README.md
wilhuff: Done (though as GitHub Actions).

scripts/setup_check.sh
wilhuff: `style.sh` already logs the version used. This is duplicative.

scripts/style.sh
wilhuff: No objection. Inlined.

Note that by hand use of `swiftformat` is not really something anyone can reasonably do. You need to pass a bunch of command-line options to control the style we use. Now you also need to set `MINT_PATH`. As a result, if you want to format a file without running all the checks you should just run `style.sh`.

scripts/style.sh
wilhuff: Done.

wilhuff: Done.

scripts/style.sh
wilhuff: I'd prefer to leave the version log in because it makes it easier to directly see in CI logs what happened. The comment about not checking the version is also useful because it contrasts with `clang-format`, which we do rigorously check.

wilhuff: Actually, looking at the check failure this CL induced, the first run of `mint run swiftformat` installs it and that turns out to be this version check. It gives output like this:
```
Found: clang-format version 10.0.1 
Found: 🌱 Cloning SwiftFormat 0.45.5
🌱 Resolving package
🌱 Building package
🌱 Installed SwiftFormat 0.45.5
🌱 Running swiftformat 0.45.5...
0.45.5
```

... which is a little ridiculous. I've removed the whole section now since this will be obvious enough.

scripts/style.sh
ryanwilson: Yep that totally makes sense, I just realized that's exactly what I do.

PR #6262, opened 2020-08-12
Crashlytics watchOS support

### LINKED ISSUE #4558: FR: Crashlytics support on watchOS
Please add Crashlytics support to watchOS, so that we can get crash reports from Firebase Crashlytics for watchOS apps just like we can for iOS apps.

jostster:
This PR resolves #4558 and adds crashlytics support for watchOS 6.0 +


jostster:
@googlebot I signed it!

Crashlytics/Crashlytics/Components/FIRCLSApplication.m
```diff
@@ -44,6 +44,8 @@
   return @"mac";
 #elif TARGET_OS_TV
   return @"tvos";
+#elif TARGET_OS_WATCH
+  return @"ios"; // TODO: temporarily use iOS until Firebase can add watchos to the backend
```
paulb777, line 48: I suspect an extra space is needed before the comment to resolve the clang-format issue.  See https://github.com/firebase/firebase-ios-sdk#code-formatting

.github/workflows/crashlytics.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        target: [ios, tvos, macos] # The macos and tvos tests can hang
+        target: [ios, tvos, macos, watchos --skip-tests] # The macos and tvos tests can hang
```
paulb777, line 80: Same change at line 24 for PRs

Crashlytics/Crashlytics/Components/FIRCLSProcess.c
```diff
@@ -485,6 +485,9 @@ bool FIRCLSProcessRecordAllThreads(FIRCLSProcess *process, FIRCLSFile *file) {
 
     FIRCLSSDKLogInfo("recording thread %d data\n", i);
     if (!FIRCLSProcessRecordThread(process, thread, file)) {
+      FIRCLSFileWriteArrayEnd(file);
```
samedson, line 488: Just one suggestion here: could you add an error SDK Log Error, eg:

`FIRCLSSDKLogError("Failed to record thread state. Closing threads JSON to prevent malformed crash report.")`

jostster: Sure, since watchOS doesn't support thread states, should I wrap that if in a supported macro? Otherwise it will always log that error message for watchOS.

samedson: I think it's ok to leave as-is. This will remind folks in the future that this is something we can improve upon 👍

jostster:
@samedson looks like one check failed and was because of teamid during the build. I assume that is fine?

paulb777:
@jostster That build failure is addressed by #6263. I'll merge this one here. Thanks!

PR #6264, opened 2020-08-12
Make the language reported via headers configurable

var-const:


Firestore/Source/API/FIRFirestore.mm
```diff
@@ -97,6 +97,8 @@ + (void)initialize {
   if (self == [FIRFirestore class]) {
     SetThrowHandler(ObjcThrowHandler);
   }
+
+  Firestore::SetClientLanguageHeader("gl-objc/2.0");
```
var-const, line 101: Feel free to suggest a better call site for this -- I'm not sure this is a particularly great place.

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -119,8 +120,32 @@ class HostConfigMap {
 };
 
 HostConfigMap& Config() {
-  static HostConfigMap config_by_host_;
```
var-const, line 122: I think trailing underscores are only for member variables.

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -119,8 +120,32 @@ class HostConfigMap {
...
-  return config_by_host_;
+  static HostConfigMap config_by_host;
+  return config_by_host;
+}
+
+class ClientLanguageHeader {
```
var-const, line 127: I deliberately made this mirror `HostConfigMap` and `Config`. I'm open to suggestions if you think this is overkill.

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -173,16 +198,7 @@ std::unique_ptr<grpc::ClientContext> GrpcConnection::CreateContext(
     context->AddMetadata(kAuthorizationHeader, absl::StrCat("Bearer ", token));
   }
 
-  // TODO(dimond): This should ideally also include the gRPC version, however,
```
var-const, line 176: The concern that this TODO raises doesn't seem true to me -- can you think of a case where a user would be able to link a different version of gRPC at runtime?

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -193,6 +209,12 @@ std::unique_ptr<grpc::ClientContext> GrpcConnection::CreateContext(
   return context;
 }
 
+void GrpcConnection::AddCloudApiHeaders(grpc::ClientContext& context) {
```
var-const, line 212: Note: there are many tests that create a `RemoteStore` but not `FIRFirestore`; they would have the language header blank. I presume this is okay but pointing it out just in case.

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -307,17 +329,20 @@ void GrpcConnection::Unregister(GrpcCall* call) {
...
+void GrpcConnection::SetClientLanguageHeader(std::string language_header) {
+  LanguageHeader().Set(std::move(language_header));
 }
 
-/*static*/ void GrpcConnection::UseInsecureChannel(const std::string& host) {
+void GrpcConnection::UseInsecureChannel(const std::string& host) {
   Config().UseInsecureChannel(host);
 }
 
+void GrpcConnection::UseTestCertificate(const std::string& host,
```
var-const, line 340: Made the order here to be the same as the header file.

scripts/style.sh
```diff
@@ -53,7 +53,7 @@ case "$version" in
   *)
     echo "Please upgrade to clang-format version 10."
     echo "If it's installed via homebrew you can run:"
-    echo "brew install clang-format"
+    echo "brew upgrade clang-format"
```
var-const, line 56: If a package is already installed, `install` would fail and recommend using `upgrade`.

Firestore/Source/API/FIRFirestore.mm
wilhuff: This is the place to do it. However, it should be inside the `if` block (which is a defense against this running more than once).

Firestore/core/src/remote/grpc_connection.cc
wilhuff: This concern dates back from when we were building Firestore in google3. The issue was that we would build our binary SDK against the gRPC in google3 and ship with undefined symbols for gRPC that would actually be supplied via CocoaPods on the end developer machine.

Now that we're distributed the way are, this concern no longer holds, so we could use the gRPC version.

Firestore/core/src/api/firestore.cc
```diff
@@ -191,6 +191,10 @@ void Firestore::DisableNetwork(util::StatusCallback callback) {
   client_->DisableNetwork(std::move(callback));
 }
 
+void Firestore::SetClientLanguageHeader(std::string language_header) {
+  FirestoreClient::SetClientLanguageHeader(std::move(language_header));
```
wilhuff, line 195: I see little reason for all the additional plumbing of the value of this property. It seems reasonable to have `Firestore` directly call the setter on `GrpcConnection`. I do see value in having this be a part of `Firestore`: this makes the core API better defined and gives us flexibility in how/where this is implemented in here without having to update the public C++ code.

Firestore/core/src/api/firestore.h
```diff
@@ -95,6 +95,12 @@ class Firestore : public std::enable_shared_from_this<Firestore> {
   void EnableNetwork(util::StatusCallback callback);
   void DisableNetwork(util::StatusCallback callback);
 
+  /**
+   * Sets the language of the public API in the format of
+   * "gl-<language>/<version>", e.g. `gl-objc/2.0`.
+   */
+  static void SetClientLanguageHeader(std::string language_header);
```
wilhuff, line 102: This value isn't really a header--it's one of many values contained in a header. However, `SetClientLanguageHeaderValue` seems overblown. What about `SetClientLanguage`?

Firestore/core/src/remote/grpc_connection.cc
wilhuff: Why not default to C++? This would make it similar to the way the throw handler defaults to C++.

Firestore/Source/API/FIRFirestore.mm
wilhuff, line 101: This value was previously `"gl-objc/"` because the version doesn't matter/isn't maintained. Changing this needlessly breaks with existing data. Unless we're specifically going to search for/audit existing usage of this in sawmill I think we're better off leaving the version out (which makes this a no-op as far as downstream effects are concerned).

var-const: Done.

var-const: Ok, makes sense. In the future, I'd like us to consider logging the LLVM version instead because, AFAIU, that's what actually determines features available to the developer.

Firestore/core/src/api/firestore.cc
var-const: Previously `Firestore` was unaware of the "remote" layer, but I agree this is more of an academic concern. Changed.

Firestore/core/src/api/firestore.h
var-const: Looking at the original doc, things like `gl-objc` are sometimes called tokens, but I'm not sure `SetClientLanguageToken` is necessarily an improvement. Changed to `SetClientLanguage`.

Firestore/core/src/remote/grpc_connection.cc
var-const: My concern is that if we ever get a bug where the language isn't set for some reason, we would get incorrect overrepresentation of C++ in the data (whereas the numbers should actually add to Objective-C or Unity). What do you think? Also, do you think there should be an assertion to verify that a language was set?

wilhuff: While I definitely respect where this concern is coming from, I don't see this as a practical concern. By hooking into class loading in Objective-C there's essentially no way you can use `FIRFirestore` without setting this value to something reasonable. I can it see it being straightforwardly the case that we can make similar guarantees in the other bindings.

Given that we have a number of these kinds of language preferences in play (exception handling, logging, version reporting, and probably others), I'd prefer that they all worked the same way just to make things simpler.

Firestore/Source/API/FIRFirestore.mm
wilhuff: Future work for llvm version sg. The token still has a 2.0 in it that should be removed.

Firestore/core/src/remote/grpc_connection.h
```diff
@@ -85,6 +85,8 @@ class GrpcConnection {
   void Register(GrpcCall* call);
   void Unregister(GrpcCall* call);
 
+  static void SetClientLanguage(std::string language_header);
```
wilhuff, line 88: The the variable name be `language_token`?

var-const: Done.

Firestore/Source/API/FIRFirestore.mm
var-const: Oh, sorry about that, fixed now.

Firestore/core/src/remote/grpc_connection.cc
var-const: The consistency argument SGTM. I added simple logic for reporting the C++ version, PTAL. We might want to change it in the future, but hopefully it's good enough for now.

var-const:
The strings I'm seeing are:
```
gl-cpp/2011 fire/1.16.4 grpc/1.28.2
gl-objc/ fire/1.16.4 grpc/1.28.2
```

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -119,8 +120,59 @@ class HostConfigMap {
...
 
 HostConfigMap& Config() {
-  static HostConfigMap config_by_host_;
-  return config_by_host_;
+  static HostConfigMap config_by_host;
+  return config_by_host;
+}
+
+std::string GetCppLanguageToken() {
+  std::string cpp_version = [] {
```
wilhuff, line 128: Since these are all literals, you could make this return `const char*` to save the copy.

var-const: Done.

PR #6281, opened 2020-08-17
[App Distribution] Remove typedefs

jeremydurham:
Stop using typedefs for completion blocks in public APIs so that Xcode is more usable in Swift:

#no-changelog App Distribution is not released yet

### DIFF
--- a/FirebaseAppDistribution/Sources/Public/FIRAppDistribution.h
+++ b/FirebaseAppDistribution/Sources/Public/FIRAppDistribution.h
@@ -13,23 +13,8 @@
 // limitations under the License.
 
 @class FIRAppDistributionRelease;
-#import <Foundation/Foundation.h>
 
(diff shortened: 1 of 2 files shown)


FirebaseAppDistribution/Sources/Public/FIRAppDistribution.h
```diff
@@ -13,24 +13,9 @@
 // limitations under the License.
 
 @class FIRAppDistributionRelease;
-#import <Foundation/Foundation.h>
```
paulb777, line 16: Was this delete intended?  It may still be needed for NSError's declaration.

jeremydurham: I couldn't tell initially why we were importing Foundation because it builds either way, but good point on NSError. I'll add it back.

PR #6282, opened 2020-08-17
[App Distribution] Add isExpired to FIRAppDistributionRelease

jeremydurham:
Add isExpired to FIRAppDistributionRelease to allow the user to verify that the Release's download URL is not expired

#no-changelog App Distribution is not released yet

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistributionRelease.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistributionRelease.m
@@ -19,9 +19,13 @@
 @property(nonatomic, copy) NSString *buildVersion;
 @property(nonatomic, copy) NSString *releaseNotes;
 @property(nonatomic, strong) NSURL *downloadURL;
+@property(nonatomic) NSDate *expirationTime;
 @end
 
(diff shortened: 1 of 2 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionRelease.m
```diff
@@ -19,9 +19,13 @@ @interface FIRAppDistributionRelease ()
 @property(nonatomic, copy) NSString *buildVersion;
 @property(nonatomic, copy) NSString *releaseNotes;
 @property(nonatomic, strong) NSURL *downloadURL;
+@property(nonatomic) NSDate *expirationTime;
 @end
 
 @implementation FIRAppDistributionRelease
+
```
pranavrajgopal, line 26: Would it be worth making it slightly less than an hour to avoid any weird race conditions around the 1 hour mark? Maybe 59 minutes?

jeremydurham: Good point. I'll make it 59 minutes instead.

PR #6285, opened 2020-08-18
Add both IID and FCM APIs in test app to test compatibility

chliangGoogle:
- [x] Add buttons on test app to test add/delete token from both Messaging and InstanceID SDK.
- [x] When developers adopt to the new version of messaging without IID as a dependency, but such developer still uses IID to get/delete token. We need to ensure token is also updated in Messaging side.

Added a method to update the cache in Messaging as IID already update its own cache and storage that is shared between messaging and IID, so we just have to update messaging's cache.

### DIFF
--- a/FirebaseMessaging/Sources/Public/FIRMessaging.h
+++ b/FirebaseMessaging/Sources/Public/FIRMessaging.h
@@ -380,6 +380,22 @@ NS_SWIFT_NAME(Messaging)
 @property(nonatomic, readonly, nullable) NSString *FCMToken NS_SWIFT_NAME(fcmToken);
 
 /**
+ *  Asynchronously getting the default token.
+ *
+ *  @param completion The completion handler to handle the token request.
+ */
+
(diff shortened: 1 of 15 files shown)


FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -379,6 +379,23 @@ NS_SWIFT_NAME(Messaging)
  */
 @property(nonatomic, readonly, nullable) NSString *FCMToken NS_SWIFT_NAME(fcmToken);
 
+/**
+ *  Asynchronously getting the default token.
+ *
+ *  @param completion The completion handler to handle the token request.
+ */
+
+- (void)tokenWithCompletion:
```
paulb777, line 388: Did the new public APIs get reviewed?

chliangGoogle: not yet, I realize this could be pushed to master early than API review. I've removed the new public API and use the existing token API for test app testing.

PR #6286, opened 2020-08-18
Update InstanceID to listen to a notification from Messaging

chliangGoogle:
When we release Messaging that removes InstanceID dependency, Messaging will handle FCM token on its own. However, developers might still depend on both Messaging and InstanceID, each has its own cache for token, but share the same storage(keychain). We need to add a method in InstanceID to listen to a notification from Messaging for token refresh, if that happens, InstanceID will update its cache, but not storage, as Messaging already updated it. 

We want developers to adopt this version before hand as we can't re-enforce iid version once messaging remove it from its dependency. So I updated the new version of messaging to depend on at least this version of IID.

Manual tested in the chen/fm-master branch which has the latest version of Messaging without IID dependency.
Add/delete token from Messaging API and try getToken in InstanceID should return the same token and FID.

Minor: 
Fixed an issue that the local defaultFCMToken is not updated when calling tokenWithAuthorizedEntity:scope:options:handler. This cause developers won't get notified of token update. It is not discovered probably because not so many developers use this API.

### DIFF
--- a/Firebase/InstanceID/CHANGELOG.md
+++ b/Firebase/InstanceID/CHANGELOG.md
@@ -1,6 +1,6 @@
-# 2020-08 -- 4.5.2
+# 2020-08 -- 4.6.0
+- [added] Added a new notification listening token refresh from Messaging and update its cache. (#6286)
 - [fixed] Fixed an issue that token refresh notification is not triggered when use `tokenWithAuthorizedEntity:scope:options:handler` to get token. (#6286)
-- [changed] Added a new notification listening token refresh from Messaging and update its cache. (#6286)
 
 # 2020-07 -- 4.5.1
 - [changed] Remove FIRInstanceIDURLQueryItem in favor of NSURLQueryItem. (#5835)
--- a/Firebase/InstanceID/CHANGELOG.md
+++ b/Firebase/InstanceID/CHANGELOG.md
@@ -1,4 +1,5 @@
 # 2020-08 -- 4.5.2
+- [fixed] Fixed an issue that token refresh notification is not triggered when use `tokenWithAuthorizedEntity:scope:options:handler` to get token. (#6286)
 - [changed] Added a new notification listening token refresh from Messaging and update its cache. (#6286)
 
 # 2020-07 -- 4.5.1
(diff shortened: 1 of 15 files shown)


Firebase/InstanceID/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2020-08 -- 4.5.2
```
paulb777, line 1: Should be 4.6.0 with the new functionality

FirebaseInstanceID.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'FirebaseInstanceID'
-  s.version          = '4.5.1'
+  s.version          = '4.5.2'
```
paulb777, line 3: 4.6.0

FirebaseMessaging.podspec
```diff
@@ -55,7 +55,7 @@ device, and it is completely free.
   s.osx.framework = 'SystemConfiguration'
   s.weak_framework = 'UserNotifications'
   s.dependency 'FirebaseCore', '~> 6.10'
-  s.dependency 'FirebaseInstanceID', '~> 4.3'
+  s.dependency 'FirebaseInstanceID', '~> 4.5', '>= 4.5.2'
```
paulb777, line 58: '~> 4.6'

Firebase/InstanceID/CHANGELOG.md
chliangGoogle: There's no public API change, I guess this is for better version managing?

FirebaseMessaging.podspec
chliangGoogle: Done.

Firebase/InstanceID/CHANGELOG.md
paulb777: Minor version updates aren't just for public API changes - they should be used for any changes that go beyond bug fixes.  See https://semver.org/

Firebase/InstanceID/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2020-08 -- 4.6.0
+- [added] Added a new notification listening token refresh from Messaging and update its cache. (#6286)
```
paulb777, line 2: `its` is ambiguous. Do you mean the IID cache?

chliangGoogle: Done. I updated the changelog to clarify it.

PR #6294, opened 2020-08-18
Delete redundant unused files

paulb777:
Fix `pod spec lint` issue since the deleted header isn't available there.

#no-changelog


paulb777:
Should merge before the M78 code freeze.

PR #6313, opened 2020-08-20
Messaging Token API for review

chliangGoogle:
Adding new getToken and deleteToken API in Messaging so we can migrate users to start using them to handle FCM registration token instead of using InstanceID.
Also add a new deleteWithCompletion: method for GDPR delete to replace InstanceID.delete. This method deletes all the tokens and checkin data that FIS.deleteWithCompletion: doesn't do. Developers should call both to handle GDPR delete for messaging.

minor:
- Fixed an issue that when passing a wildcard "*" sender ID in delete token request, we don't notify users that tokens are deleted.
- Adjust the test app to have buttons control token handling from instanceID, Messaging and Installations for easy testing.


FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -617,6 +626,38 @@ - (void)deleteFCMTokenForSenderID:(nonnull NSString *)senderID
                                            handler:completion];
 }
 
+- (void)deleteWithCompletion:(void (^)(NSError *_Nullable))completion {
+  FIRMessaging_WEAKIFY(self);
+  [self.instanceID
```
maksymmalyhin, line 631: Optional nit: I always want to suggest using [Promises](https://github.com/google/promises) when I see several nested async operations.

chliangGoogle: This is still temporary as Messaging underline just calling InstanceID calls. I think we can start applying Promises once we finish the large refactor of removing IID completely.

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
- *  the FCM token to send notifications to your device.
+ * Once you have an FCM token, you should send it to your application server, so it can use
+ * the FCM token to send notifications to your device.
  */
 @property(nonatomic, readonly, nullable) NSString *FCMToken NS_SWIFT_NAME(fcmToken);
 
+/**
+ * Asynchronously getting the default FCM registration token.
+ *
+ * This creates a Firebase Installations ID, if one does not exist, and sends information
```
paulb777, line 385: How does the Installations ID relate to the FCM registration token?

chliangGoogle:
After running the script/style.sh, the swift interface changes from
 NS_SWIFT_NAME(delete(completion:)); 
to 
 NS_SWIFT_NAME(delete (completion:)); 
with a warning in xcode 'swift_name' attribute has invalid identifier for base name"

If I rename something like deleteBlahBlah(completion:), clang format won't format it with a space after that, which has no warning in xcode.

However, I do prefer to use deleteWithCompletion: to be consistent with FIS: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInstallations/Source/Library/Public/FirebaseInstallations/FIRInstallations.h#L116

I can't ignore the format as it won't pass presubmit.

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -403,6 +433,8 @@ NS_SWIFT_NAME(Messaging)
 /**
  *  Invalidates an FCM token for a particular Sender ID. That Sender ID cannot no longer send
  *  notifications to that FCM token.
+ *  This Does not delete the Firebase Installations ID that may have been created when
```
paulb777, line 436: s/Does/does/

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -504,6 +536,20 @@ NS_SWIFT_NAME(Messaging)
  */
 - (FIRMessagingMessageInfo *)appDidReceiveMessage:(NSDictionary *)message;
 
+#pragma mark - GDPR
+/**
+ * Deletes all the tokens and checkin data of the Firebase project  and related data on the server
```
paulb777, line 541: extra space before `and`

FirebaseMessaging/Sources/Public/FIRMessaging.h
chliangGoogle: Actually I just realize this call doesn't necessarily creates a FID, FID is always existed in the client.  @gsakakihara 
So I don't think we need to state here what FID's impact on user info collecting, but we should mention the registration token's impact on this.
@andirayo @egilmorez

paulb777:
See https://github.com/nicklockwood/SwiftFormat#:~:text=You%20can%20disable%20rules%20for,...%5D%5D for disabling swiftformat in local areas.

Interested in @ryanwilson's  perspective on this.

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -504,6 +536,20 @@ NS_SWIFT_NAME(Messaging)
...
+ * side. A network connection is required for the method to succeed.
+ *
+ * This does not delete the Firebase Installations ID that may have been created when
+ * generating the token. See `FIRInstallations deleteWithCompletion:` for deleting that.
+ *
+ * @param completion A completion handler which is invoked when the operation completes. `error ==
+ * nil` indicates success.
+ */
+- (void)deleteWithCompletion:(void (^)(NSError *__nullable error))completion
+    NS_SWIFT_NAME(delete(completion:));
```
maksymmalyhin, line 551: I think `NS_SWIFT_NAME` is redundant here:
Screen Shot 2020-09-01 at 4 59 35 PM

chliangGoogle: I notice this new swift header file, is it a new thing?
I guess with this, we don't need to define any of the above NS_SWIFT_NAME anymore if it's by default defined here?
also add @paulb777 and @ryanwilson here

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -351,8 +351,8 @@ NS_SWIFT_NAME(Messaging)
  *
  * If this flag is disabled, Firebase Messaging does not generate new tokens automatically for
  * message delivery. If this flag is enabled, FCM generates a registration token on application
- * start when there is no existing valid token. FCM also generates a new token when an existing
- * token is deleted.
+ * start when there is no existing valid token and periodically refreshing the token and sending
```
egilmorez, line 354: To keep the verbs synced, suggest "refreshes the token and sends"

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
 @property(nonatomic, assign, getter=isAutoInitEnabled) BOOL autoInitEnabled;
 
 /**
- *  The FCM token is used to identify this device so that FCM can send notifications to it.
- *  It is associated with your APNS token when the APNS token is supplied, so that sending
- *  messages to the FCM token will be delivered over APNS.
+ * The FCM registration token is used to identify this device so that FCM can send notifications to
+ * it. It is associated with your APNS token when the APNS token is supplied, so that sending
```
egilmorez, line 369: Should we be using "APNs" throughout, with lowercase "s?"

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
+ * messages to the FCM token will be delivered over APNS.
```
egilmorez, line 370: Suggest "when the APNS token is supplied, so
messages sent to the FCM token will be delivered over APNs."

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
  *
- *  The FCM token is sometimes refreshed automatically. In your FIRMessaging delegate, the
- *  delegate method `messaging:didReceiveRegistrationToken:` will be called once a token is
- *  available, or has been refreshed. Typically it should be called once per app start, but
- *  may be called more often, if token is invalidated or updated.
+ * The FCM registration token is sometimes refreshed automatically. In your FIRMessaging delegate,
+ * the delegate method `messaging:didReceiveRegistrationToken:` will be called once a token is
+ * available, or has been refreshed. Typically it should be called once per app start, but
+ * may be called more often, if token is invalidated or updated.
```
egilmorez, line 375: Suggest "often if the token"

(no comma)

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
  *
- *  Once you have an FCM token, you should send it to your application server, so it can use
- *  the FCM token to send notifications to your device.
+ * Once you have an FCM token, you should send it to your application server, so it can use
```
egilmorez, line 377: Suggest "FCM registration token"

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
+ * the FCM token to send notifications to your device.
  */
 @property(nonatomic, readonly, nullable) NSString *FCMToken NS_SWIFT_NAME(fcmToken);
 
+/**
+ * Asynchronously getting the default FCM registration token.
```
egilmorez, line 383: Is it iOS style to use gerunds, or should this be "Asynchronously gets"  ?

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
+ *
+ * A network connection is required for the method to succeed and data is sent
```
egilmorez, line 385: Suggest comma after "succeed"

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
+ *
+ * @param completion The completion handler to handle the token request.
+ */
+
+- (void)tokenWithCompletion:
+    (void (^)(NSString *__nullable token, NSError *__nullable error))completion
+    NS_SWIFT_NAME(token(completion:));
+
+/**
+ * Asynchronously deleting the default FCM registration token.
```
egilmorez, line 397: Is it iOS style to use gerunds, or should this be "Asynchronously deletes" ?

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -365,20 +365,47 @@ NS_SWIFT_NAME(Messaging)
...
+ *
+ * This does not delete all tokens for non-default sender ID, See `FIRMessaging
```
egilmorez, line 399: Suggest either "non-default sender IDs" or "a non-default sender ID"

I *think* it can be plural, correct?

FirebaseMessaging/Sources/Public/FIRMessaging.h
egilmorez, line 541: FCM registration tokens specifically, or all tokens?

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -504,6 +536,20 @@ NS_SWIFT_NAME(Messaging)
  */
 - (FIRMessagingMessageInfo *)appDidReceiveMessage:(NSDictionary *)message;
 
+#pragma mark - GDPR
+/**
+ * Deletes all the tokens and checkin data of the Firebase project  and related data on the server
+ * side. A network connection is required for the method to succeed.
+ *
+ * This does not delete the Firebase Installations ID. See `FIRInstallations deleteWithCompletion:` for deleting that.
```
egilmorez, line 544: Suggest just "See `FIRInstallations deleteWithCompletion:`."

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -403,6 +433,8 @@ NS_SWIFT_NAME(Messaging)
 /**
  *  Invalidates an FCM token for a particular Sender ID. That Sender ID cannot no longer send
  *  notifications to that FCM token.
+ *  This Does not delete the Firebase Installations ID that may have been created when
+ *  generating the token. See `FIRInstallations deleteWithCompletion:` for deleting that.
```
egilmorez, line 437: Suggest just "See `FIRInstallations deleteWithCompletion:`."

FirebaseMessaging/Sources/Public/FIRMessaging.h
maksymmalyhin: Compiler generates Swift API for Objective-C. The result usually looks pretty well for Objective-C API that follows guidelines. [NS_SWIFT_NAME](https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/renaming_objective-c_apis_for_swift) is needed only when renaming is required for some reason. 

You may preview the generated Swift interface for any .h file with Xcode using the editor feature as shown on the screenshot:

Screen Shot 2020-09-01 at 6 21 43 PM

FirebaseMessaging/Sources/Public/FIRMessaging.h
maksymmalyhin, line 544: If I remember correctly, we should prefer Swift variants of API in the doc string because we have more devs using Firebase from Swift and this amount is growing. So it should be `Installations.delete(completion:)` instead of `FIRInstallations deleteWithCompletion:`.

@ryanwilson @paulb777 Could you confirm please.

paulb777: Confirmed. If only one is listed it should be the Swift API.

FirebaseMessaging/Sources/Public/FIRMessaging.h
```diff
@@ -401,11 +427,12 @@ NS_SWIFT_NAME(Messaging)
     NS_SWIFT_NAME(retrieveFCMToken(forSenderID:completion:));
 
 /**
- *  Invalidates an FCM token for a particular Sender ID. That Sender ID cannot no longer send
- *  notifications to that FCM token.
+ * Invalidates an FCM token for a particular Sender ID. That Sender ID cannot no longer send
+ * notifications to that FCM token. This does not delete the Firebase Installations ID that may have
+ * been created when  generating the token. See `Installations.delete(completion:)`.
```
egilmorez, line 432: Nit: extra space.

FirebaseMessaging/Sources/Public/FIRMessaging.h
chliangGoogle: Looks like FIS has changed to not auto generated anymore #5656 to keep consistent behavior between android and iOS. The installations ID is only created if some product or developers request it.

FirebaseMessaging/Apps/Sample/Sample/ContentView.swift
```diff
@@ -221,7 +221,7 @@ struct ContentView: View {
   }
 
   func deleteFCM() {
-    Messaging.messaging().delete { error in
+    Messaging.messaging().deleteData { error in
```
chliangGoogle, line 224: @ryanwilson After renaming, the swift API calls it deleteData, does that look good to you? or I should rename swift API to deleteMessagingData as well?

ryanwilson: This looks good to me - thanks!

PR #6315, opened 2020-08-21
avoid using OCMock on NSOperation

chliangGoogle:
To reduce the flakiness in tests.


Example/InstanceID/Tests/FIRInstanceIDTokenOperationsTest.m
```diff
@@ -52,6 +52,23 @@ @interface FIRInstallationsAuthTokenResult (Tests)
 - (instancetype)initWithToken:(NSString *)token expirationDate:(NSDate *)expirationDate;
 @end
 
+// A Fake operation that we have control over the returned error.
```
dmaclach, line 55: This comment doesn't quite match functionality for this one. Maybe:

A fake operation that we allows us to check that perform was called.

Issue #5975, opened 2020-07-06
FR: Make public API for StackFrame creation from address


segrey-profi:
### Feature proposal

* Firebase Component: Crashlytics

Fabric SDK has public function `CLSStackFrame(address:)` which is needed for correct crash logging of Kotlin Multiplatform frameworks (see https://github.com/touchlab/CrashKiOS)

Firebase Crashlytics contains `FIRCLSSymbolResolver` class with `frameForAddress:` method which seems to do the same job, but it is private API so we cannot even check if it works for Kotlin frameworks.

Please make the same public API as `CLSStackFrame(address:)` in Fabric SDK.

Thomas-Vos:
See also: https://github.com/firebase/firebase-ios-sdk/issues/5182

elenadoty:
We've added this API to a branch that should be going out in the next release.

PR #6329, opened 2020-08-25
Made public a method to create a stack frame from address.

elenadoty:
Added a public method to allow stack frames to be created from addresses, as requested in https://github.com/firebase/firebase-ios-sdk/issues/5975.

### DIFF
--- a/Crashlytics/CHANGELOG.md
+++ b/Crashlytics/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [added] Added stackFrameWithAddress API for recording custom errors that are symbolicated on the backend (#5975).
+
 # v4.5.0
 - [fixed] Fixed a compiler warning and removed unused networking code (#6210).
 - [fixed] Fixed a crash that occurred rarely when trying to restart a URL session task without a valid request (#5984).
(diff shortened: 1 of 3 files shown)


samedson:
Can you add a changelog entry?

Here's the file: https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/CHANGELOG.md

Add an "Unreleased" heading, and put something like: ` - [added] Added a `stackFrameWithAddress` API for recording custom errors that are symbolicated on the backend`. The [added] gets converted into the tags on this page: https://firebase.google.com/support/release-notes/ios, and the release manager will convert `Unreleased` into whatever the next release version of Crashlytics ends up being.

samedson:
In addition, can you reference the issue in the description: https://github.com/firebase/firebase-ios-sdk/issues/5975 ?

Crashlytics/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- [added] Added astackFrameWithAddress API for recording custom errors that are symbolicated on the backend (#5975)
```
samedson, line 2: Small typo

PR #6333, opened 2020-08-25
Add a tool to update podspec repo

granluo:


scripts/create_spec_repo/Sources/SpecRepoBuilder/main.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+    return targetedDeps
+  }
+
+  func getTargetedDeps(of pod: String, from podSpecDict: SpecFiles) -> [String] {
+    let deps = searchDeps(of: pod, from: podSpecDict)
+    return filterTargetDeps(deps, with: podSpecDict)
+  }
+
+  func push_podspec(_ pod: String, from sdk_repo: String, sources: [String],
```
ryanwilson, line 181: Nit: same naming comments as before re: style and ambiguous arguments. A better name would likely be:

```
func pushPodspec(forPod podName: String,
                 sdkRepo: String,
                 sources: [String],
                 flags: [String],
                 shell: Shell = Shell.shared) -> Int32
```

Although the `sdkRepo` is a filepath, so a `URL` could be applicable as well.

scripts/create_spec_repo/Sources/SpecRepoBuilder/main.swift
```diff
@@ -0,0 +1,301 @@
...
+                    flags: [String], shell_cmd: Shell = Shell.shared) -> Int32 {
+    let pod_path = sdk_repo + "/" + pod + ".podspec"
```
ryanwilson, line 183: Naming for variables should all change from `_` case to camel case.

PR #6334, opened 2020-08-25
Add DocumentSnapshot.data(as:with:decoder:)

### LINKED ISSUE #6328: FR: Add ServerTimestampBehavior option to FirebaseFirestoreSwift
### Feature proposal

* Firebase Component: FirebaseFirestoreSwift.

These are the currently available methods to get data from a document, if FirebaseFirestoreSwift is installed:

Screenshot 2020-08-25 at 11 16 03

I would expect these additional methods, so that all use cases are covered: 
```swift
data(as: Decodable.Protocol, with: ServerTimestampBehavior) throws
data(as: Decodable.Protocol, with: ServerTimestampBehavior, decoder: Firestore.Decoder?) throws
``` 

This would also go towards feature parity with the Kotlin Android SDK.

Thank you, 
Cheers

wilhuff:
This adds support for specifying alternative server timestamp behaviors when decoding snapshots.

Fixes #6328.


wilhuff:
@ryanwilson Consider this PR a proposal. Note what usage looks like in the tests. I'm not particularly happy with the way `data(as:with:decoder:)` reads and am open to suggestions.

What I have is:
```swift
try snapshot.data(as: Model.self, with: .estimate)  // Or `.none` or `.previous`
```

It might be better to do something like this:
```swift
try snapshot.data(as: Model.self, serverTimestamps: .estimate)
```

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -38,4 +42,28 @@ extension DocumentSnapshot {
...
+  ///     not yet been set to their final value are returned from the snapshot.
+  ///   - decoder: The decoder to use to convert the document. `nil` to use
+  ///     default decoder.
+  public func data<T: Decodable>(as type: T.Type,
+                                 with serverTimestampBehavior: ServerTimestampBehavior,
+                                 decoder: Firestore.Decoder? = nil) throws -> T? {
+    var d = decoder
+    if d == nil {
+      d = Firestore.Decoder()
+    }
```
morganchen12, line 63: Drive-by nit: You can remove an optional in this code with

```swift
let d = decoder ?? Firestore.Decoder() // can use var instead of let if d must be mutable
```

ryanwilson:
> @ryanwilson Consider this PR a proposal. Note what usage looks like in the tests. I'm not particularly happy with the way `data(as:with:decoder:)` reads and am open to suggestions.
> 
> What I have is:
> 
> ```swift
> try snapshot.data(as: Model.self, with: .estimate)  // Or `.none` or `.previous`
> ```
> 
> It might be better to do something like this:
> 
> ```swift
> try snapshot.data(as: Model.self, serverTimestamps: .estimate)
> ```

I agree - I don't think the standalone `with:` reads super well. I think `serverTimestamps:` is a reasonable tradeoff of readability and verbosity (instead of using the full `serverTimestampsBehavio(u)r`). 

Would you want to make the change [in the base SDK](https://github.com/firebase/firebase-ios-sdk/blob/7609f030c6de8b22b0c43f0ede88ac035d1e90d5/Firestore/Source/Public/FirebaseFirestore/FIRDocumentSnapshot.h#L98) as well at next breaking change?

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -38,4 +42,28 @@ extension DocumentSnapshot {
...
+  /// See `Firestore.Decoder` for more details about the decoding process.
+  ///
+  /// - Parameters
+  ///   - type: The type to convert the document fields to.
+  ///   - serverTimestampBehavior: Configures how server timestamps that have
+  ///     not yet been set to their final value are returned from the snapshot.
+  ///   - decoder: The decoder to use to convert the document. `nil` to use
+  ///     default decoder.
+  public func data<T: Decodable>(as type: T.Type,
+                                 with serverTimestampBehavior: ServerTimestampBehavior,
```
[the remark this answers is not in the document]
wilhuff: Done.

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
wilhuff: Nice! Done.

ryanwilson (approved):

PR #6335, opened 2020-08-26
Update release workflow to update spec repo.

granluo:


scripts/create_spec_repo/firebase_sdk.textproto
```diff
@@ -0,0 +1,9 @@
+# Config file to update versions of Pods which will be pushed to SpecsStaging
+# repo by add_podspec tool.
+# e.g.
+# Podspec versions will be the same as firebase-ios-sdk podspecs if they are not
+# specified in this file.
+sdk {
+  sdk_name: "Firebase"
+  sdk_version: "6.31.0"
```
paulb777, line 8: Why is the Firebase version specified here versus the one from the Firebase.podspec in the repo?

scripts/setup_quickstart.sh
```diff
@@ -40,14 +40,21 @@ if check_secrets || [[ ${SAMPLE} == "installations" ]]; then
 
   git clone https://github.com/firebase/quickstart-ios.git
   $scripts_dir/localize_podfile.swift quickstart-ios/"$SAMPLE"/Podfile "$RELEASE_TESTING"
+  if [ ! -z "$RELEASE_TESTING" ]; then
+    set +x
+    sed -i "" '1i\'$'\n'"source 'https://${BOT_TOKEN}@github.com/FirebasePrivate/SpecsTesting.git'"$'\n' quickstart-ios/"$SAMPLE"/Podfile
+    set -x
+    echo "Podfile is udpated."
```
paulb777, line 47: nit: `updated`

scripts/setup_quickstart.sh
```diff
@@ -40,14 +40,21 @@ if check_secrets || [[ ${SAMPLE} == "installations" ]]; then
...
+  fi
   cd quickstart-ios/"$SAMPLE"
 
   # To test a branch, uncomment the following line
   # git checkout {BRANCH_NAME}
 
   bundle update --bundler
   bundle install
-  bundle exec pod install
+  bundle exec pod install --silent
```
paulb777, line 56: `pod install` before `pod update` is redundant

.github/workflows/release.yml
```diff
@@ -4,23 +4,72 @@ on:
...
+        cd scripts/create_spec_repo/
+        swift build
+        pod repo add --silent "${local_repo}" https://"$botaccess"@github.com/FirebasePrivate/SpecsTesting.git
+        BOT_TOKEN="${botaccess}"  .build/debug/SpecRepoBuilder --sdk_repo "${local_sdk_repo_dir}" --local_spec_repo_name "${local_repo}"
+    - name: Clean Artifacts
+      if: ${{ always() }}
+      run: |
+        pod repo remove "${local_repo}"
+        rm -rf bot-access.txt
+    - uses: actions/upload-artifact@v2
```
paulb777, line 41: How is the uploaded data used?

granluo: The updated repo, with a "test" tag, will be transferred to testing jobs, which will use those repo as a source. Podspecs in the spec repo will have source leading to that repo and pods in quickstart could be tested.

scripts/create_spec_repo/firebase_sdk.textproto
granluo: This is due to a potential nil.
https://github.com/firebase/firebase-ios-sdk/blob/master/ZipBuilder/Sources/firebase-pod-updater/main.swift#L52
Has update the pod-version-updator to allow textproto to be empty.

granluo: 275fddf

scripts/create_spec_repo/firebase_sdk.textproto
```diff
@@ -0,0 +1,5 @@
+# Config file to update versions of Pods which will be pushed to SpecsStaging
```
paulb777, line 1: Can this file be deleted?

granluo: > It looks like there's a failing CI - https://github.com/firebase/firebase-ios-sdk/pull/6335/checks?check_run_id=1038350404

This is fixed by https://github.com/firebase/firebase-ios-sdk/pull/6355

> Can this file be deleted?

This file is for some closed source sdk. e.g. FirebaseAnalytics. This will happen in a scenario like when in Firebase.podspec, FirebaseAnalytics is updated to a newer version but it is not released yet, then when we push to a testing repo, it will show error like FirebaseAnalytics version not found and so Quickstart tests cannot run since repo is failed to build. This file can update the version to let tests run before a closed source gets released.
However, if we consider that failure is expected then we can remove that file. One of my concern is manual effort might be needed to update that file.

PR #6338, opened 2020-08-26
Archive testing on GHA

### LINKED ISSUE #2107: Add archive build testing to CI
We should use travis CI to make sure that Archive builds continue to work and we don't release with issues like #2102.

ryanwilson:
Unfortunately we can't generate the app thinning report via exporting the archive since it requires actual code signing.

Fixed #2107


ryanwilson:
Successful run of all three platforms shown at https://github.com/firebase/firebase-ios-sdk/actions/runs/225813633, moving to a cron job now.

ryanwilson:
Moving the following SDKs to cron jobs since they passed (links to passing job):

- [ABTesting](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887492)
- [AppDistribution](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887497)
- [Auth](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887487)
- [Crashlytics](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887494)
- [Database](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887491)
- [DynamicLinks](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887500)
- [Firestore](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887490)
- [Functions](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887493)
- [GDT](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887486)
- [InAppMessaging](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887501)
- [InstanceID](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887504)
- [Messaging](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887488)
- [Remote Config](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887485)
- [Storage](https://github.com/firebase/firebase-ios-sdk/actions/runs/225887498)

I removed Installations since it had some issues with depending on an older version of IID, and it should be covered in other Archive tests.

.github/workflows/abtesting.yml
```diff
@@ -116,3 +116,23 @@ jobs:
...
+    strategy:
+      matrix:
+        target: [ios, tvos, macos]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Setup project and archive
+      run: scripts/test_archiving.sh FirebaseABTesting ${{ matrix.target }} ArchiveOutputs/${{ matrix.target }}.xcarchive
+    - uses: actions/upload-artifact@v2
```
paulb777, line 134: What's the reason for uploading?  How will it be used?

ryanwilson: Good question! It can be downloaded and used to sign locally to do a size measurement, although that may not be a good enough reason to upload. Happy to remove it, especially if it's eating into a quota I'm unaware of.

We could also potentially just use the `.ipa` inside of the archive to do a rough size guesstimate as well without signing, that could be done on GHA.

paulb777: I believe there is some cost, so my preference would be to remove the upload until we have a definite need.

I like the idea of getting the size from this. At the very least, just adding an `ls -l` to the script could be low hanging fruit - and then we could expand it over time to upload and track ...

PR #6348, opened 2020-08-27
App Distribution patch for restricted api keys

schnecle:
* [fixed] Bug for customers with restricted API keys unable to fetch releases.


schnecle:
This should be merged into 6.31.0

PR #6352, opened 2020-08-27
Bump the podspec version and changelog version for release patch

schnecle:
* Bump podspec version for release patch
* Add release patch to changelog


ryanwilson:
LGTM, no need to wait for CI in this case.

PR #6346, opened 2020-08-27
App distribution add bundle header

schnecle:
### Description
Customers with restricted api keys reported being unable to see new build alerts and saw repeated prompts for login

### Changes
* Added x-ios-bundle-identifier for api keys restricted to ios apps

PR #6353, opened 2020-08-27
App distribution patch for restricted api keys

schnecle:
* [fixed] Bug for customers with restricted API keys unable to fetch releases (#6346).


ryanwilson:
LGTM - please wait for CI to merge it.

PR #6365, opened 2020-08-28
GDT: storage size limit

maksymmalyhin:
GoogleDataTransport simple storage size limit (b/129492246).

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h"
+
+@interface GDTCORDirectorySizeCalculator ()
+
+@property(nonatomic, readonly) NSString *directoryPath;
+
+@property(nonatomic) NSNumber *cachedSize;
+
+@end
+
+@implementation GDTCORDirectorySizeCalculator
+
+- (instancetype)initWithDirectoryPath:(NSString *)path {
+  self = [super init];
+  if (self) {
+    _directoryPath = path;
+  }
+  return self;
+}
+
+- (uint64_t)directoryContentSize {
+  if (self.cachedSize == nil) {
+    self.cachedSize = @([self calculateDirectoryContentSize]);
+  }
+
+  return self.cachedSize.unsignedLongLongValue;
+}
+
+- (void)fileWithSize:(GDTCORFileSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
+  if (![path hasPrefix:self.directoryPath]) {
+    // Ignore because the file is not inside the directory.
+    return;
+  }
+
+  self.cachedSize = @([self directoryContentSize] + fileSize);
+}
+
+- (void)fileWithSize:(GDTCORFileSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
+  if (![path hasPrefix:self.directoryPath]) {
+    // Ignore because the file is not inside the directory.
+    return;
+  }
+
+  self.cachedSize = @([self directoryContentSize] - fileSize);
+}
+
+- (void)resetCachedSize {
+  self.cachedSize = nil;
+}
+
+- (uint64_t)calculateDirectoryContentSize {
+  NSArray *prefetchedProperties = @[ NSURLIsRegularFileKey, NSURLFileSizeKey ];
+  uint64_t totalBytes = 0;
+  NSURL *directoryURL = [NSURL URLWithString:self.directoryPath];
+
+  NSDirectoryEnumerator *enumerator = [[NSFileManager defaultManager]
+                 enumeratorAtURL:directoryURL
+      includingPropertiesForKeys:prefetchedProperties
+                         options:NSDirectoryEnumerationSkipsHiddenFiles
+                    errorHandler:^BOOL(NSURL *_Nonnull url, NSError *_Nonnull error) {
+                      return YES;
+                    }];
+
+  for (NSURL *fileURL in enumerator) {
+    NSNumber *isRegularFile;
+    [fileURL getResourceValue:&isRegularFile forKey:NSURLIsRegularFileKey error:nil];
+    if (isRegularFile.boolValue) {
+      NSNumber *fileSize;
+      [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:nil];
+      totalBytes += fileSize.unsignedLongLongValue;
+    }
+  }
+
+  return totalBytes;
+}
+
+@end
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
@@ -34,7 +34,7 @@
   return self;
 }
 
-- (uint64_t)directoryContentSize {
+- (GDTCORFileSizeBytes)directoryContentSize {
   if (self.cachedSize == nil) {
     self.cachedSize = @([self calculateDirectoryContentSize]);
   }
@@ -64,7 +64,7 @@
   self.cachedSize = nil;
 }
 
-- (uint64_t)calculateDirectoryContentSize {
+- (GDTCORFileSizeBytes)calculateDirectoryContentSize {
   NSArray *prefetchedProperties = @[ NSURLIsRegularFileKey, NSURLFileSizeKey ];
   uint64_t totalBytes = 0;
   NSURL *directoryURL = [NSURL URLWithString:self.directoryPath];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
@@ -18,9 +18,11 @@
 
 @interface GDTCORDirectorySizeCalculator ()
 
+/** The observed directory path. */
 @property(nonatomic, readonly) NSString *directoryPath;
 
-@property(nonatomic) NSNumber *cachedSize;
+/** The cached content size of the observed directory. */
+@property(nonatomic, nullable) NSNumber *cachedSizeBytes;
 
 @end
 
@@ -35,11 +37,11 @@
 }
 
 - (GDTCORStorageSizeBytes)directoryContentSize {
-  if (self.cachedSize == nil) {
-    self.cachedSize = @([self calculateDirectoryContentSize]);
+  if (self.cachedSizeBytes == nil) {
+    self.cachedSizeBytes = @([self calculateDirectoryContentSize]);
   }
 
-  return self.cachedSize.unsignedLongLongValue;
+  return self.cachedSizeBytes.unsignedLongLongValue;
 }
 
 - (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
@@ -48,7 +50,7 @@
     return;
   }
 
-  self.cachedSize = @([self directoryContentSize] + fileSize);
+  self.cachedSizeBytes = @([self directoryContentSize] + fileSize);
 }
 
 - (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
@@ -57,11 +59,11 @@
     return;
   }
 
-  self.cachedSize = @([self directoryContentSize] - fileSize);
+  self.cachedSizeBytes = @([self directoryContentSize] - fileSize);
 }
 
 - (void)resetCachedSize {
-  self.cachedSize = nil;
+  self.cachedSizeBytes = nil;
 }
 
 - (GDTCORStorageSizeBytes)calculateDirectoryContentSize {
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
@@ -34,7 +34,7 @@
   return self;
 }
 
-- (GDTCORFileSizeBytes)directoryContentSize {
+- (GDTCORStorageSizeBytes)directoryContentSize {
   if (self.cachedSize == nil) {
     self.cachedSize = @([self calculateDirectoryContentSize]);
   }
@@ -42,7 +42,7 @@
   return self.cachedSize.unsignedLongLongValue;
 }
 
-- (void)fileWithSize:(GDTCORFileSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
+- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
   if (![path hasPrefix:self.directoryPath]) {
     // Ignore because the file is not inside the directory.
     return;
@@ -51,7 +51,7 @@
   self.cachedSize = @([self directoryContentSize] + fileSize);
 }
 
-- (void)fileWithSize:(GDTCORFileSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
+- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
   if (![path hasPrefix:self.directoryPath]) {
     // Ignore because the file is not inside the directory.
     return;
@@ -64,7 +64,7 @@
   self.cachedSize = nil;
 }
 
-- (GDTCORFileSizeBytes)calculateDirectoryContentSize {
+- (GDTCORStorageSizeBytes)calculateDirectoryContentSize {
   NSArray *prefetchedProperties = @[ NSURLIsRegularFileKey, NSURLFileSizeKey ];
   uint64_t totalBytes = 0;
   NSURL *directoryURL = [NSURL URLWithString:self.directoryPath];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h"
+#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeTracker.h"
 
-@interface GDTCORDirectorySizeCalculator ()
+@interface GDTCORDirectorySizeTracker ()
 
 /** The observed directory path. */
 @property(nonatomic, readonly) NSString *directoryPath;
@@ -26,7 +26,7 @@
 
 @end
 
-@implementation GDTCORDirectorySizeCalculator
+@implementation GDTCORDirectorySizeTracker
 
 - (instancetype)initWithDirectoryPath:(NSString *)path {
   self = [super init];
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
@@ -44,7 +44,7 @@
   return self.cachedSizeBytes.unsignedLongLongValue;
 }
 
-- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
+- (void)fileWasAddedAtPath:(NSString *)path withSize:(GDTCORStorageSizeBytes)fileSize {
   if (![path hasPrefix:self.directoryPath]) {
     // Ignore because the file is not inside the directory.
     return;
@@ -53,7 +53,7 @@
   self.cachedSizeBytes = @([self directoryContentSize] + fileSize);
 }
 
-- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
+- (void)fileWasRemovedAtPath:(NSString *)path withSize:(GDTCORStorageSizeBytes)fileSize {
   if (![path hasPrefix:self.directoryPath]) {
     // Ignore because the file is not inside the directory.
     return;
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
@@ -83,13 +83,17 @@
     NSNumber *isRegularFile;
     [fileURL getResourceValue:&isRegularFile forKey:NSURLIsRegularFileKey error:nil];
     if (isRegularFile.boolValue) {
-      NSNumber *fileSize;
-      [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:nil];
-      totalBytes += fileSize.unsignedLongLongValue;
+      totalBytes += [self fileSizeAtURL:fileURL];
     }
   }
 
   return totalBytes;
+}
+
+- (GDTCORStorageSizeBytes)fileSizeAtURL:(NSURL *)fileURL {
+  NSNumber *fileSize;
+  [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:nil];
+  return fileSize.unsignedLongLongValue;
 }
 
 @end
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
@@ -165,14 +165,10 @@ const uint64_t kGDTCORFlatFileStorageSizeLimit = 20 * 1000 * 1000;  // 20 MB.
     BOOL writeResult = GDTCORWriteDataToFile(encodedEvent, filePath, &error);
     if (writeResult == NO || error) {
       GDTCORLogDebug(@"Attempt to write archive failed: path:%@ error:%@", filePath, error);
-    } else {
-      GDTCORLogDebug(@"Writing archive succeeded: %@", filePath);
-    }
-
-    if (error) {
       completion(NO, error);
       return;
     } else {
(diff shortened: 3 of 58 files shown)


GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,93 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h"
+
+@interface GDTCORDirectorySizeCalculator ()
+
+@property(nonatomic, readonly) NSString *directoryPath;
```
mikehaney24, line 21: docstring

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,93 @@
...
+
+@property(nonatomic) NSNumber *cachedSize;
```
mikehaney24, line 23: docstring

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -49,6 +51,16 @@
 
 NSString *const kGDTCORBatchComponentsExpirationKey = @"GDTCORBatchComponentsExpirationKey";
 
+NSString *const GDTCORFlatFileStorageErrorDomain = @"GDTCORFlatFileStorage";
+
+const uint64_t kGDTCORFlatFileStorageSizeLimit = 20 * 1000 * 1000;  // 20 MB.
+
+@interface GDTCORFlatFileStorage ()
+
+@property(nonatomic, readonly) GDTCORDirectorySizeCalculator *sizeCalculator;
```
mikehaney24, line 60: docstring

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORPlatform.h
```diff
@@ -156,6 +156,8 @@ id<NSSecureCoding> _Nullable GDTCORDecodeArchive(Class archiveClass,
                                                  NSData *_Nullable archiveData,
                                                  NSError *_Nullable *error);
 
+BOOL GDTCORWriteDataToFile(NSData *data, NSString *filePath, NSError **outError);
```
mikehaney24, line 159: docstring

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORStorageProtocol.h
```diff
@@ -25,6 +25,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef uint64_t GDTCORStorageSizeBytes;
```
mikehaney24, line 28: comment

GoogleDataTransport/GDTCORTests/Common/Categories/GDTCORFlatFileStorage+Testing.h
```diff
@@ -28,6 +30,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)reset;
 
+@property(nonatomic, readonly) GDTCORDirectorySizeCalculator *sizeCalculator;
```
mikehaney24, line 33: docstring

GoogleDataTransport/GDTCORTests/Unit/GDTCORFlatFileStorageTest.m
```diff
@@ -1193,4 +1274,30 @@ - (void)storeEvent:(GDTCOREvent *)event inStorage:(GDTCORFlatFileStorage *)stora
   [self waitForExpectations:@[ eventStoredExpectation ] timeout:0.5];
 }
 
+- (uint64_t)storageSize {
```
mikehaney24, line 1277: I realize I didn't do this above, but docstring for these helpers as well

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,93 @@
...
+                 enumeratorAtURL:directoryURL
+      includingPropertiesForKeys:prefetchedProperties
+                         options:NSDirectoryEnumerationSkipsHiddenFiles
+                    errorHandler:^BOOL(NSURL *_Nonnull url, NSError *_Nonnull error) {
+                      return YES;
+                    }];
+
+  for (NSURL *fileURL in enumerator) {
+    NSNumber *isRegularFile;
+    [fileURL getResourceValue:&isRegularFile forKey:NSURLIsRegularFileKey error:nil];
```
mikehaney24, line 82: If there are errors, should we continue?

GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+
+#import "GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORStorageProtocol.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/** The class calculates and caches the specified directory content size and uses add/remove signals
+ * from client the client to keep the size up to date without accessing file system. */
```
mikehaney24, line 24: ending */ on next line

GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+#import "GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORStorageProtocol.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/** The class calculates and caches the specified directory content size and uses add/remove signals
```
mikehaney24, line 23: There are no thread-safety or locking mechanisms of any kind in this class. Can you speak to why that is and how to use this class in the doc?

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORPlatform.h
paulb777, line 159: Why is the public API changing?

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORStorageProtocol.h
paulb777, line 28: Why do clients need this?

maksymmalyhin: @paulb777 The type was introduced to help with the API readability. It stresses the value meaning and units. I think this should help with preventing issues like we had with timestamps #6102.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
maksymmalyhin: My assumption here is that if we get an error here, then it is unlikely a file created by GDT, so it should be fine to skip it in the size calculation. WDYT?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -281,7 +324,12 @@ - (void)libraryDataForKey:(nonnull NSString *)key
...
       if (newValue != nil && [newValue isKindOfClass:[NSData class]] && newValue.length) {
         NSError *newValueError;
-        [newValue writeToFile:dataPath options:NSDataWritingAtomic error:&newValueError];
+        if ([newValue writeToFile:dataPath options:NSDataWritingAtomic error:&newValueError]) {
+          // Update storage size.
+          [self.sizeCalculator fileWithSize:data.length wasRemovedAtPath:dataPath];
+          [self.sizeCalculator fileWithSize:newValue.length wasAddedAtPath:dataPath];
+        }
+
         if (newValueError) {
```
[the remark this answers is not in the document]
maksymmalyhin: The [method at line 327](https://developer.apple.com/documentation/foundation/nsdata/1414800-writetofile?language=objc) should return `NO` in the case of the failure so the size will not be updated. But I agree, it is better to replace this `if` by `else` clause to improve readability. Will do.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
mikehaney24: SGTM

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORPlatform.h
maksymmalyhin: Similar methods were in this file, so placed this one here as well. Can you see any issues with it?

maksymmalyhin:
@mikehaney24 I updated the API docs. Please let me know if you would like to see more details.

mikehaney24:
lgtm

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORPlatform.h
paulb777: It looks like it follows the style. 

I don't understand why the public API surface is so big, but that question is probably out of scope for this PR.

maksymmalyhin:
Will be merged after M79 release branch cut.

maksymmalyhin:
Profiler doesn't indicate any significant performance regressions attributed to the changes. The most resource demanding part is still writing the event data to the disk. 
[GDT Size Limit.trace.zip](https://github.com/firebase/firebase-ios-sdk/files/5158563/GDT.Size.Limit.trace.zip)

GoogleDataTransport/GDTCORLibrary/Public/GoogleDataTransport/GDTCORPlatform.h
maksymmalyhin: Created #6383 to deal with it.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -126,14 +141,43 @@ - (void)storeEvent:(GDTCOREvent *)event
...
-    GDTCOREncodeArchive(event, filePath, &error);
+    NSData *encodedEvent = GDTCOREncodeArchive(event, nil, &error);
+    if (error) {
+      completion(NO, error);
+      return;
+    }
+
+    // Check storage size limit before storing the event.
+    uint64_t resultingStorageSize = self.sizeCalculator.directoryContentSize + encodedEvent.length;
+    if (resultingStorageSize > kGDTCORFlatFileStorageSizeLimit) {
```
visumickey, line 152: Instead of dropping the new events, can we drop the old events? There is a good chance that the stale events sitting on the device can force the recent events to get dropped.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -126,14 +141,43 @@ - (void)storeEvent:(GDTCOREvent *)event
...
+
     if (error) {
       completion(NO, error);
       return;
     } else {
       completion(YES, nil);
     }
 
+    // Notify size calculator.
+    [self.sizeCalculator fileWithSize:encodedEvent.length wasAddedAtPath:filePath];
```
visumickey, line 179: Instead of a two step process of adding the event to the filepath and then updating the size. Can we update the size of the directory in the GDTCORWriteDataToFile() method instead?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -281,8 +325,11 @@ - (void)libraryDataForKey:(nonnull NSString *)key
       // the implicit return value will be the block itself. The compiler doesn't detect this.
       if (newValue != nil && [newValue isKindOfClass:[NSData class]] && newValue.length) {
         NSError *newValueError;
-        [newValue writeToFile:dataPath options:NSDataWritingAtomic error:&newValueError];
-        if (newValueError) {
+        if ([newValue writeToFile:dataPath options:NSDataWritingAtomic error:&newValueError]) {
+          // Update storage size.
+          [self.sizeCalculator fileWithSize:data.length wasRemovedAtPath:dataPath];
```
visumickey, line 330: Instead of calling fileRemoved and added file, should we open up a new method named updateFileSizeAtDirectoryPath:WithDelta:.?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -314,8 +363,12 @@ - (void)removeLibraryDataForKey:(nonnull NSString *)key
   dispatch_async(_storageQueue, ^{
```
visumickey, line 363: Since I/O operations are happening on a serial queue (non main queue), can these be a synchronous operation to provide thread safety? This is outside the scope of this PR, but something to be considered for future.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,95 @@
...
+- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
+  if (![path hasPrefix:self.directoryPath]) {
+    // Ignore because the file is not inside the directory.
+    return;
+  }
+
+  self.cachedSizeBytes = @([self directoryContentSize] - fileSize);
+}
+
+- (void)resetCachedSize {
```
visumickey, line 65: This could just be reset. Managing cache is all internal. For an external consumer, this could just be reset:.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,95 @@
...
+
+#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h"
+
+@interface GDTCORDirectorySizeCalculator ()
+
+/** The observed directory path. */
+@property(nonatomic, readonly) NSString *directoryPath;
+
+/** The cached content size of the observed directory. */
+@property(nonatomic, nullable) NSNumber *cachedSizeBytes;
```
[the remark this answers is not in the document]
maksymmalyhin: `nil` is used to represent `no cache` state. Alternatively an additional variable could be used, but usually managing more variables is more error prone, so I prefer using `NSNumber` in this case.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -49,6 +51,17 @@
 
 NSString *const kGDTCORBatchComponentsExpirationKey = @"GDTCORBatchComponentsExpirationKey";
 
+NSString *const GDTCORFlatFileStorageErrorDomain = @"GDTCORFlatFileStorage";
+
+const uint64_t kGDTCORFlatFileStorageSizeLimit = 20 * 1000 * 1000;  // 20 MB.
```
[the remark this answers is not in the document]
maksymmalyhin: There is a design doc with discussions attached to the ticket mentioned in the PR description - b/129492246.

The doc TL;DR - we are replicating the currently implemented Android behaviour so far and will iterate on improvements on both platforms synchronously.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,95 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "GoogleDataTransport/GDTCORLibrary/Internal/GDTCORDirectorySizeCalculator.h"
+
+@interface GDTCORDirectorySizeCalculator ()
```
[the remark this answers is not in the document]
maksymmalyhin: I like it! Will rename.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -83,6 +96,8 @@ - (instancetype)init {
     _storageQueue =
         dispatch_queue_create("com.google.GDTCORFlatFileStorage", DISPATCH_QUEUE_SERIAL);
     _uploadCoordinator = [GDTCORUploadCoordinator sharedInstance];
+    _sizeCalculator =
+        [[GDTCORDirectorySizeCalculator alloc] initWithDirectoryPath:GDTCORRootDirectory().path];
```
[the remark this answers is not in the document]
maksymmalyhin: Good catch! Calling `GDTCORRootDirectory()` will trigger I/O operation. I'll implement lazy initialization of `sizeCalculator` property to defer the I/O until it's really required.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: This is current behaviour on Android and Clearcut SDK. DD attached to b/129492246 describes several alternatives which we are going to iterate on simultaneously on Firelog iOS and Android.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: Currently `GDTCORWriteDataToFile()` is a generic and independent function that has no context on the file it stores, so updating the size tracker in it doesn't seem appropriate. 

But I like the idea in general. I've considered implementing something like `GDTFileManager` which will be responsible for file system operations and keeping track of directory sizes as needed. But this change will increase the scope of the PR, so I would prefer to defer it and implement later. WDYT?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: I considered this approach but decided to use the two existing methods because of the following:
1. A method such `updateFileSizeAtDirectoryPath:WithDelta:` will require calculating the diff from the caller, so actually will be more difficult to use. Though the calculation will most likely be trivial it will have to be repeated on the caller side.
2. Benefits of the methods don't justify maintenance cost of additional API. 

Does it look reasonable to you?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: This particular method is a part of the API, so it can be called from outside the storage and require explicit synchronization. But I agree, there are couple places where similar methods can be separated to sync unprotected part for internal use and async protected for external use. We already use this approach for methods like [this](https://cs.opensource.google/firebase-sdk/firebase-ios-sdk/+/master:GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m;l=503;drc=802cb6a71678084bdff6d7b381cb474eae6b27ef). I would prefer to make such modifications as required.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
```diff
@@ -0,0 +1,95 @@
...
+- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasAddedAtPath:(NSString *)path {
+  if (![path hasPrefix:self.directoryPath]) {
+    // Ignore because the file is not inside the directory.
+    return;
+  }
+
+  self.cachedSizeBytes = @([self directoryContentSize] + fileSize);
+}
+
+- (void)fileWithSize:(GDTCORStorageSizeBytes)fileSize wasRemovedAtPath:(NSString *)path {
```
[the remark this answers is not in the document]
maksymmalyhin: Both options (existing and proposed) read a bit awkward to me, but the current version reads a bit more like a sentence to me. @paulb777 @ryanwilson do you have an opinion on it?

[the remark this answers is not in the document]
paulb777: I like @visumickey's suggestion. It clears up the confusion I had when I first saw those two methods.  However, I might be biased from my C programming background and expecting the key function description information to be in the name and not the parameters ....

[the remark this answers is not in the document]
ryanwilson: How about swapping the path and size:

```
- (void)fileWasRemovedAtPath:(NSString *)path withSize:(GDTCORStorageSizeBytes)fileSize;
- (void)fileWasAddedAtPath:(NSString *)path withSize:(GDTCORStorageSizeBytes)fileSize;
```

`withSize` could also just be `size` as well.

Side note - should we be using file URLs here instead of NSStrings?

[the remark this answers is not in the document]
maksymmalyhin: I like Ryan's suggestion the most so far. 

As for URLs - NSString with paths are uses all across the storage now. I think it will be better to replace them all in one shot to avoid path-to-URL conversions. Another concern with URLs is that URL is a path with URL scheme, parameters, etc. so we will probably need to make more validation on it. I would defer URLs change.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
```diff
@@ -0,0 +1,95 @@
...
+
+- (GDTCORStorageSizeBytes)calculateDirectoryContentSize {
+  NSArray *prefetchedProperties = @[ NSURLIsRegularFileKey, NSURLFileSizeKey ];
+  uint64_t totalBytes = 0;
+  NSURL *directoryURL = [NSURL URLWithString:self.directoryPath];
+
+  NSDirectoryEnumerator *enumerator = [[NSFileManager defaultManager]
+                 enumeratorAtURL:directoryURL
+      includingPropertiesForKeys:prefetchedProperties
+                         options:NSDirectoryEnumerationSkipsHiddenFiles
```
visumickey, line 77: Curious: Why skip hidden files?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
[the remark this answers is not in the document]
visumickey: Fair enough. We can circle on this more in deep in the design doc.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -126,14 +149,43 @@ - (void)storeEvent:(GDTCOREvent *)event
...
+                 userInfo:@{
+                   NSLocalizedFailureReasonErrorKey : @"Storage size limit has been reached."
+                 }];
+      completion(NO, error);
+      return;
+    }
+
+    // Write the encoded event to the file.
+    BOOL writeResult = GDTCORWriteDataToFile(encodedEvent, filePath, &error);
+    if (writeResult == NO || error) {
```
visumickey, line 173: Is there an opportunity to merge these together?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
visumickey: Fair enough to punt on this. can you track this refactor internally (as a bug)?

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
```diff
@@ -545,6 +581,14 @@ - (void)syncThreadUnsafeRemoveBatchWithID:(nonnull NSNumber *)batchID
       removeBatchDir(batchDirPath);
     }
   }
+
+  [self.sizeTracker resetCachedSize];
+}
+
+- (GDTCORStorageSizeBytes)fileSizeAtURL:(NSURL *)fileURL {
+  NSNumber *fileSize;
+  [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:nil];
```
visumickey, line 590: Can we add this as a util method (may be even static) to the size tracker? Not too much sold on it. But to unify the size tracking/calculation logic to a single place.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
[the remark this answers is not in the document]
visumickey: SGTM

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
visumickey: I agree, that goes outside the scope of this PR. How are we managing TODOs for Firelog? Consider creating an internal bug to track this.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
visumickey: Ha.. That is true. Makes sense with this change.

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeCalculator.m
visumickey: SGTM.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: Created b/168098576

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: b/168100199

GoogleDataTransport/GDTCORLibrary/GDTCORDirectorySizeTracker.m
maksymmalyhin: The assumption is that we don't create hidden files, so we don't want to take them into account.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: Good catch! Merged.

GoogleDataTransport/GDTCORLibrary/GDTCORFlatFileStorage.m
maksymmalyhin: Done

visumickey (approved):

Issue #6341, opened 2020-08-27
import Firebase is not sufficient to define the FirebaseApp symbol in my build environment.


canfieldtim:
### [REQUIRED] Step 1: Describe your environment
  * Xcode version: 11.6 (11E708)
  * Firebase SDK version: 6.31.0
  * Firebase Component: Analytics
  * Component version: 6.8.0
  * Installation method: CocoaPods
  * macOS version: 10.15.6 (19G2021)

### [REQUIRED] Step 2: Describe the problem

'import Firebase' is no longer sufficient to define the FirebaseApp symbol in my build environment.  Using Xcode 11.6, with Firebase 6.31.0, I have to import FirebaseCore to avoid getting: "error: use of unresolved identifier 'FirebaseApp'".  I was previously using Firebase 6.27.0 on this same build machine without issue when only importing Firebase.

I believe this is the same issue as #6066.  The comments in that issue indicate that 'import Firebase' should be sufficient and that importing the specific component (or firebase core) is not needed.  I have run through the proposed resolution in that issue (quit Xcode, remove pods dir, remove xcworkspace dir, rerun pod install, open newly created xcworkspace).  This does not resolve the issue in my project.

I am able to reproduce the same issue with the [QuickStart sample](https://github.com/firebase/quickstart-ios) for analytics.

This problem does not occur on my other build machine, which has the same firebase environment, but is running Xcode 11.5 (11E608c) on macOS 10.15.4 (19E266).

#### Steps to reproduce:

Reproducible with [QuickStart sample](https://github.com/firebase/quickstart-ios): 
$ git clone https://github.com/firebase/quickstart-ios.git
$ cd quickstart-ios/analytics/
$ pod install --repo-update
$ open AnalyticsExample.xcworkspace

Building with the Debug configuration for physical device running iOS 13 produces:

```
Prepare build
note: Using new build system
note: Building targets in parallel
note: Planning build
note: Constructing build description

Build target nanopb of project Pods with configuration Debug
Build target Pods-AnalyticsExample of project Pods with configuration Debug
Build target FirebaseCoreDiagnostics of project Pods with configuration Debug
Build target PromisesObjC of project Pods with configuration Debug
Build target GoogleDataTransport of project Pods with configuration Debug
Build target FirebaseCore of project Pods with configuration Debug
Build target AnalyticsExample of project AnalyticsExample with configuration Debug

[...]

PR #6368, opened 2020-08-28
Prepare for 6.31.1

paulb777:
Targeting publishing Monday morning to provide a fix for a Xcode race condition build failure. (#6341)

### DIFF
--- a/CoreOnly/Sources/Firebase.h
+++ b/CoreOnly/Sources/Firebase.h
@@ -12,45 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The module qualified imports are for CocoaPods and the simple file names
-// for Swift Package Manager.
+#import <FirebaseCore/FirebaseCore.h>
 
 #if !defined(__has_include)
   #error "Firebase.h won't import anything if your compiler doesn't support __has_include. Please \
           import the headers individually."
 #else
-  #if __has_include(<FirebaseCore/FirebaseCore.h>)
-    #import <FirebaseCore/FirebaseCore.h>
-  #elif __has_include("FirebaseCore.h")
-    #import "FirebaseCore.h"
-  #endif
-
   #if __has_include(<FirebaseAnalytics/FirebaseAnalytics.h>)
     #import <FirebaseAnalytics/FirebaseAnalytics.h>
   #endif
 
-  #if __has_include(<FirebaseAuth/FirebaseAuth.h>)
-    #import <FirebaseAuth/FirebaseAuth.h>
-  #elif __has_include("FirebaseAuth.h")
-    #import "FirebaseAuth.h"
-  #endif
-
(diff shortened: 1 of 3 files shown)


mikehardy:
Hey @paulb777 / @ryanwilson - a release engineering note, somehow the 6.31.1 release wasn't complete in the way previous releases have been? Basis for that assertion is that the "Latest Release" on github is still showing as 6.31.0, and there's no binary zip connected with CocoaPods-6.31.1 tag. This comes up (for me, and likely others) because the pre-compiled firestore-ios-sdk-frameworks keys off the fresh release existing and then pulls from the zips in order to pre-compile firestore and accelerate all the FlutterFire and react-native-firebase users's builds

- What previous release looks like: https://github.com/firebase/firebase-ios-sdk/releases?after=DataTransport-7.3.0
- What current looks like: https://github.com/firebase/firebase-ios-sdk/releases
- Where we fail to find release: https://github.com/invertase/firestore-ios-sdk-frameworks/blob/640fa0a865e5338b5135a2f785b29c6030cb0f44/.github/workflows/scripts/process-latest-version.sh#L79
- Why this is even a thing :-) - https://github.com/invertase/firestore-ios-sdk-frameworks#before--after

paulb777:
@mikehardy sorry about the confusion.

See https://firebase.google.com/support/release-notes/ios

We released 6.31.1 only for CocoaPods since we've had numerous reports of not finding the Firebase module during builds that we were never able to reproduce but turned out to be a change in 6.28.0 exposing an Xcode build race condition. Details in #6341.

Since we had no reports for other distribution mechanisms, we skipped the release engineering associated with them in favor of deferring to 6.32.0.

I would suggest that you ignore any releases that don't regenerate a zip distribution, but happy to discuss other processes.

And thanks for providing such a nice speed-up for Firestore users!

mikehardy:
Well that was all @Salakar to give credit where it is due, you're becoming more familiar with me because I'm more like the Chief Plate Spinner just trying to keep everything working and then diagnosing when it's not :-). The script I referenced will just automatically fetch 6.32.0 whenever it happens, but we can't control if people attempt to use 6.31.1 - it will just fail for them if they try. Is that a problem? I don't know

But I can add to the script some output indicating why it is not working (so this exchange doesn't happen again) and I can add some documentation to the repository so people can understand why the pre-compiled framework might not be available and just to wait a version if so. That is likely good enough (it is for me)

paulb777:
Sounds good! 

If not, there might be an approach around tagging, versioning, and publishing the FirebaseFirestore binary podspec.

mikehardy:
If there was anything related to that one broken out at a sub-spec level (since our framework pre-compiler is already looking at the subspec from cocoapods IIRC) we could probably alter it to hook to that. If you all think on that and come up with anything, tag me or Mike D (salakar) in and we can collaborate, sure

Issue #6345, opened 2020-08-27
FR: Excluding redundant packages


borut-t:
### Feature proposal

I'm currently using a few packages from firebase (using cocoapods), here is the list:
```swift
pod 'Firebase/Auth'
pod 'Firebase/Messaging'
pod 'Firebase/RemoteConfig'
pod 'Firebase/Crashlytics'
pod 'Firebase/Analytics'
pod 'Firebase/Storage'
pod 'FirebaseUI/Email'
```

Reviewing the `Podfile.lock` file reveals that there are multiple top-level packages installed that I didn't want. Just to name a few:
- FirebaseABTesting
- FirebaseInstallations
- FirebaseInstanceID
- GoogleAppMeasurement
- GoogleDataTransport
- ...

Now wondering if there is a way I can exclude some (probably not all) packages from my app? I would like to reduce the size of the app and every redundant package is well, redundant.

I completely understand that some of those packages are utility packages to the main libs. But probably not all of them are required, are they?

E.g. `FirebaseABTesting` is never used in my app so why bother bundling it in?

So, my proposal/feature request here is that you should be transparent on what is included besides the top-level packages and omit the ones that are not "required".

paulb777:
@borut-t CocoaPods and Firebase do not install extra pods. If you search the `Podfile.lock` for any of the questioned pods, you'll see them indented under the pods that require them.

Firebase is architected with extra pods to increase code sharing, enable more granular installations, and decrease overall size.

borut-t:
@paulb777 I wouldn't be so sure about it. Let's check the `Podfile.lock` file:

```
  - Firebase/Analytics (6.27.1):
    - Firebase/Core
  - Firebase/Auth (6.27.1):
    - Firebase/CoreOnly
    - FirebaseAuth (~> 6.6.0)
  - Firebase/Core (6.27.1):
    - Firebase/CoreOnly
    - FirebaseAnalytics (= 6.6.2)
  - Firebase/CoreOnly (6.27.1):
    - FirebaseCore (= 6.8.1)
  - Firebase/Crashlytics (6.27.1):
    - Firebase/CoreOnly
    - FirebaseCrashlytics (~> 4.2.0)
  - Firebase/Messaging (6.27.1):
    - Firebase/CoreOnly
    - FirebaseMessaging (~> 4.5.0)
  - Firebase/RemoteConfig (6.27.1):
    - Firebase/CoreOnly
[...]

PR #6379, opened 2020-08-31
Enable ABT to be a soft dep of RC

paulb777:
Fix one of the issues raised in #6345.

- Propose a new location for Interop headers in the one version world.
- Add two Interop APIs for the RC -> ABT interface
- Tested with ABTesting removed from the RC podspec, but added it back until 7.0
- Added a separate unit test target for combined RC and ABT testing
- Will wait for M79 code freeze branch before merging.

#no-changelog

### DIFF
--- a/FirebaseABTesting/Sources/FIRExperimentController.m
+++ b/FirebaseABTesting/Sources/FIRExperimentController.m
@@ -20,6 +20,7 @@
 #import "FirebaseABTesting/Sources/Public/FirebaseABTesting/FIRLifecycleEvents.h"
 #import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
 
+#import "FirebaseABTesting/Sources/Interop/FIRABTInterop.h"
 #import "Interop/Analytics/Public/FIRAnalyticsInterop.h"
 
 #ifndef FIRABTesting_VERSION
@@ -120,13 +121,7 @@ NSArray *ABTExperimentsToClearFromPayloads(
   return experimentsToClear;
 }
 
-// ABT doesn't provide any functionality to other components,
-// so it provides a private, empty protocol that it conforms to and use it for registration.
-
-@protocol FIRABTInstanceProvider
-@end
-
-@interface FIRExperimentController () <FIRABTInstanceProvider, FIRLibrary>
+@interface FIRExperimentController () <FIRABTInterop, FIRLibrary>
 @property(nonatomic, readwrite, strong) id<FIRAnalyticsInterop> _Nullable analytics;
 @end
 
@@ -148,7 +143,7 @@ NSArray *ABTExperimentsToClearFromPayloads(
     id<FIRAnalyticsInterop> analytics = FIR_COMPONENT(FIRAnalyticsInterop, container);
     return [[FIRExperimentController alloc] initWithAnalytics:analytics];
   };
-  FIRComponent *abtProvider = [FIRComponent componentWithProtocol:@protocol(FIRABTInstanceProvider)
+  FIRComponent *abtProvider = [FIRComponent componentWithProtocol:@protocol(FIRABTInterop)
                                               instantiationTiming:FIRInstantiationTimingLazy
                                                      dependencies:@[ analyticsDep ]
                                                     creationBlock:creationBlock];
@@ -166,7 +161,7 @@ NSArray *ABTExperimentsToClearFromPayloads(
 
 + (FIRExperimentController *)sharedInstance {
   FIRApp *defaultApp = [FIRApp defaultApp];  // Missing configure will be logged here.
-  id<FIRABTInstanceProvider> instance = FIR_COMPONENT(FIRABTInstanceProvider, defaultApp.container);
+  id<FIRABTInterop> instance = FIR_COMPONENT(FIRABTInterop, defaultApp.container);
 
   // We know the instance coming from the container is a FIRExperimentController instance, cast it.
   return (FIRExperimentController *)instance;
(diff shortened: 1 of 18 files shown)


christibbs:
Will the end result be:
1. Pulling in Remote Config will pull in the ABT dependency by default, but developers can remove it, or...
2. Pulling in Remote Config will not pull in the ABT dependency, but developers can add it?

paulb777:
@christibbs It will be #2, but not until the Firebase 7.0.0 release to avoid an earlier breaking change.

FirebaseRemoteConfig/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+- Unreleased
```
morganchen12, line 1: nit: This should be a header, not a list item `# Unreleased`

FirebaseABTesting/Sources/Interop/FIRABTInterop.h
```diff
@@ -0,0 +1,48 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Connector for bridging communication between Firebase SDKs and FirebaseABTesting API.
+@protocol FIRABTInterop
+
+/// Updates the list of experiments. Experiments already
+/// existing in payloads are not affected, whose state and payload is preserved. This method
+/// compares whether the experiments have changed or not by their variant ID. This runs in a
+/// background queue..
```
ryanwilson, line 27: Nit: the first line can fit more words, also double periods at the end of the paragraph.

FirebaseRemoteConfig.podspec
```diff
@@ -78,6 +78,22 @@ app update.
...
+        'FirebaseRemoteConfig/Tests/Unit/RCNConfigExperimentTest.m',
+        'FirebaseRemoteConfig/Tests/Unit/RCNTestUtilities.[hm]',
+        'FirebaseABTesting/Sources/Private/*.h',
+    ]
+    abt.resources = [
+        'FirebaseRemoteConfig/Tests/Unit/TestABTPayload.txt',
+    ]
+    abt.requires_app_host = true
+    abt.dependency 'OCMock'
+    abt.dependency 'FirebaseABTesting', '~> 4.2'
```
ryanwilson, line 93: Should this be 4.3 now instead of 4.2?

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -29,6 +28,9 @@
 #import "FirebaseRemoteConfig/Sources/RCNConfigValue_Internal.h"
 #import "FirebaseRemoteConfig/Sources/RCNDevice.h"
 
+#import "FirebaseABTesting/Sources/Interop/FIRABTInterop.h"
```
ryanwilson, line 31: Nit: Should this not stay at the top where it was before?

FirebaseABTesting/Sources/FIRExperimentController.m
```diff
@@ -191,6 +186,21 @@ - (void)updateExperimentsWithServiceOrigin:(NSString *)origin
...
+- (void)updateExperimentsWithServiceOrigin:(NSString *)origin
+                             lastStartTime:(NSTimeInterval)lastStartTime
+                                  payloads:(NSArray<NSData *> *)payloads {
+  FIRLifecycleEvents *lifecycleEvent = [[FIRLifecycleEvents alloc] init];
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+  [self updateExperimentsWithServiceOrigin:origin
+                                    events:lifecycleEvent
+                                    policy:ABTExperimentPayloadExperimentOverflowPolicyDiscardOldest
```
ryanwilson, line 198: Will this always be the case for interop, or do we want it to be a parameter as well?

ryanwilson (commented):
Overall LGTM, a few small things

FirebaseABTesting/Sources/FIRExperimentController.m
paulb777: Perhaps we want it to be the default for now? If we need to support other policies, we could add an API and expose the policies via the interop as well.

FirebaseABTesting/Sources/Interop/FIRABTInterop.h
paulb777: Done. Here and in the public header this was copied from.

FirebaseRemoteConfig.podspec
paulb777: Yes. Good catch.

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
paulb777: Done.

FirebaseABTesting/Sources/FIRExperimentController.m
ryanwilson: SGTM, thanks

ryanwilson (approved):
LGTM assuming product owners are good with this (the fact that ABT may not be there for all RC users)

paulb777:
@christibbs Would you confirm this is ok from an ABT product perspective that the fact that ABT users will need to explicitly include the pod?

christibbs:
> @christibbs Would you confirm this is ok from an ABT product perspective that the fact that ABT users will need to explicitly include the pod?

Will confirm with the team. At the very least it'll require some extensive documentation updates.

Also concerned about the case where developers `pod update` and miss the message that Firebase/ABTesting is removed. Or will it be removed in that case?

paulb777:
@christibbs Why "extensive" documentation updates? Isn't it one-line to say to add `pod 'Firebase/ABTesting'` to the Podfille?

paulb777:
Closing since it is a product requirement for ABT to be a hard dependency of RC so that RC developers have the flexibility to add in ABT in the future.

PR #6382, opened 2020-09-01
redact UDID information from binary image data

adamvduke:
It's possible that the binary image data loaded from the OS includes the device's UDID. This makes a best effort at redacting that data before the crash report is written to disk.


samedson:
Can you update: https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/CHANGELOG.md

For the [tag], can probably put `[fixed]`, and for the issue you can put this PR number eg. `(#6382)`

PR #6396, opened 2020-09-02
Removing client id usage in api call and respective checks

eldhosembabu:
Removing client id usage in api call and respective checks

FDL SDK was using client id passed in FIROptions to make API calls. From backend side, we no longer needs client id to be sent. So removing the client id usage in FDL SDK along with parameter no null checks.

-Removed the usage of Client id from FIROptions.
-Removed non null parameter checks for client id.
-Deprecated 1p API method setUpWithLaunchOptions::apiKey:clientID:urlScheme:userDefaults: and added setUpWithLaunchOptions::apiKey:urlScheme:userDefaults:
-Removed Client id usage in Unit tests.

### DIFF
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,5 +1,5 @@
 # v4.3.1
-- [removed] Client id usage in api call and respective checks in the code.
+- [changed] Client id usage in api call and respective checks in the code.
 
 # v4.3.0
 - [changed] Functionally neutral public header refactor to enable Swift Package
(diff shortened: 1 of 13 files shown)


paulb777 (commented):
Please fix the CI failure. See other examples in the same file of disabling the deprecated-implementations warning.

Is the change completely transparent to developers? If not, please add a changelog entry.

eldhosembabu:
@paulb777 Fixed the CI warning. Added to changelog entry.

FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.3.1
+- [removed] Client id usage in api call and respective checks in the code.
```
paulb777, line 2: s/removed/changed/ since nothing is removed from a client perspecitve.

paulb777 (approved):

FirebaseDynamicLinks/CHANGELOG.md
eldhosembabu: done

PR #6402, opened 2020-09-04
Auth Emulator support

rosalyntan:
Implement support for the auth emulator. The entry point is in an internal header for now, and will be moved to the public header when ready for release.

### DIFF
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -1468,7 +1468,7 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
     self->_requestConfiguration.emulatorURL =
         [NSString stringWithFormat:@"%@:%ld", host, (long)port];
 #if TARGET_OS_IOS
-    _settings.appVerificationDisabledForTesting = YES;
+    self->_settings.appVerificationDisabledForTesting = YES;
 #endif
   });
 }
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -1465,7 +1465,8 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
   NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
   dispatch_sync(FIRAuthGlobalWorkQueue(), ^{
-    self->_requestConfiguration.emulatorURL = [NSString stringWithFormat:@"%@:%ld", host, port];
+    self->_requestConfiguration.emulatorURL =
+        [NSString stringWithFormat:@"%@:%ld", host, (long)port];
   });
 }
 
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -1467,6 +1467,7 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
   dispatch_sync(FIRAuthGlobalWorkQueue(), ^{
     self->_requestConfiguration.emulatorURL =
         [NSString stringWithFormat:@"%@:%ld", host, (long)port];
+    _settings.appVerificationDisabledForTesting = YES;
   });
 }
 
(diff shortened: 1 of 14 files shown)


rosalyntan:
Test failure looks to be a transient failure, from hitting Facebook's API request limit.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -1462,6 +1462,17 @@ - (void)useAppLanguage {
   });
 }
 
+- (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
+  NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
+  dispatch_sync(FIRAuthGlobalWorkQueue(), ^{
+    self->_requestConfiguration.emulatorURL =
+        [NSString stringWithFormat:@"%@:%ld", host, (long)port];
```
yuchenshi, line 1469: This won't work for IPv6 addresses as host. `::1` + `8080` should be `[::1]:8080` with IP in brackets.

yuchenshi (changes requested):
The terms used are really confusing here, so let's clear them up a bit.

* `http://localhost:9000/foo/bar/baz` <-- URL or URI, whichever is consistent with existing code
* `localhost` <--- Host (a.k.a. "hostname")
* `9000` <--- Port
* `localhost:9000` <--- hostAndPort (a.k.a. as just "host" in RFC for URLs, but that's not consistent with the iOS SDK terms)

Please rename variables accordingly. e.g. `requestConfiguration.emulatorURL` should be renamed to `emulatorHostAndPort`, or just keep two separate variables. This will make templating much much easier to reason about.

FirebaseAuth/Sources/Backend/FIRIdentityToolkitRequest.m
```diff
@@ -80,23 +84,37 @@ - (BOOL)containsPostBody {
...
+  NSString *emulatorURL = _requestConfiguration.emulatorURL;
+
   if (_useIdentityPlatform) {
     apiURLFormat = kIdentityPlatformAPIURLFormat;
-    if (_useStaging) {
+    apiPrefix = kHttpsPrefix;
+    if (emulatorURL) {
+      apiPrefix = kHttpPrefix;
+      apiHost =
+          [NSString stringWithFormat:kEmulatorHostFormat, emulatorURL, kIdentityPlatformAPIHost];
```
yuchenshi, line 98: I'd also suggest that we don't forcibly fit `localhost:9000/www.googleapis.com` into any variable named `host`. It's confusing since it's not a host (it is host + port + path prefix), and it is fragile since things may break if we ever refactor to use `CFURL` for URL creation. It's up to you though.

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: Done.

FirebaseAuth/Sources/Backend/FIRIdentityToolkitRequest.m
rosalyntan: I changed the variable to be apiHostAndPathPrefix for more clarity.

yuchenshi (approved):

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
```diff
@@ -323,9 +328,17 @@ - (void)getHeadFulLiteURLWithEventID:(NSString *)eventID
...
                                          httpArgumentsStringForArgsDictionary:urlArguments];
-                                     NSString *URLString =
-                                         [NSString stringWithFormat:kHeadfulLiteURLStringFormat,
-                                                                    authDomain, argumentsString];
+                                     NSString *URLString;
+                                     if (strongSelf->_auth.requestConfiguration
+                                             .emulatorHostAndPort) {
+                                       URLString = [NSString
+                                           stringWithFormat:kHeadfulLiteEmulatorURLStringFormat,
+                                                            authDomain, argumentsString];
```
yuchenshi, line 336: nit: the logic here doesn't actually need `authDomain`. Consider extracting most of the `completion` block into a function that takes widget URL without parameters, and call it with either the production or emulator URL. In this way, you don't even need to call or stub `fetchAuthDomainWithRequestConfiguration` at all when emulator is used.

renkelvin (approved):

FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m
rosalyntan: IMO I think it seems neater to leave everything in the completion. Since fetchAuthDomainWithRequestConfiguration returns immediately with the emulator host, I don't think there's any downside in the current logic.

PR #6418, opened 2020-09-08
Add != and notIn support

thebrianchen:
Porting from [web](https://github.com/firebase/firebase-js-sdk/pull/3557) and [Android](https://github.com/firebase/firebase-android-sdk/pull/1872).


Firestore/core/src/remote/serializer.cc
```diff
@@ -1102,20 +1102,28 @@ google_firestore_v1_StructuredQuery_Filter Serializer::EncodeSingularFilter(
...
+        google_firestore_v1_StructuredQuery_Filter_unary_filter_tag;
+    result.unary_filter.which_operand_type =
+        google_firestore_v1_StructuredQuery_UnaryFilter_field_tag;
+    result.unary_filter.field.field_path = EncodeFieldPath(filter.field());
+
+    if (filter.value().is_nan()) {
+      auto op =
+          filter.op() == Filter::Operator::Equal
+              ? google_firestore_v1_StructuredQuery_UnaryFilter_Operator_IS_NAN
+              : google_firestore_v1_StructuredQuery_UnaryFilter_Operator_IS_NOT_NAN;
```
thebrianchen, line 1117: This line exceeds the maximum line length (<=80). Should I just move this into a variable to try and squeeze it in?

Firestore/core/src/core/key_field_in_filter.h
```diff
@@ -32,6 +34,11 @@ class KeyFieldInFilter : public FieldFilter {
  public:
   KeyFieldInFilter(model::FieldPath field, model::FieldValue value);
 
+  static bool Contains(const model::FieldValue::Array& array_value,
```
wilhuff, line 37: One way to document that these helpers exist specifically to share code with another class is to make them private and then make that other class a `friend` of this one.

Firestore/core/src/core/key_field_not_in_filter.cc
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2019 Google LLC
```
wilhuff, line 2: 2020

Firestore/Source/Public/FirebaseFirestore/FIRQuery.h
```diff
@@ -308,6 +332,36 @@ NS_SWIFT_NAME(Query)
 - (FIRQuery *)queryWhereFieldPath:(FIRFieldPath *)path
                                in:(NSArray<id> *)values NS_SWIFT_NAME(whereField(_:in:));
 
+/**
+ * Creates and returns a new `FIRQuery` with the additional filter that documents must contain
+ * the specified field and the value does not equal any of the values from the provided array.
+ *
+ * A query can have only one `notIn` filter, and it cannot be combined with an `arrayContains`,
+ * `arrayContainsAny`, `in`, or `notEqual` filter.
```
wilhuff, line 340: Is it worth calling out the behavior of not-in [null] here? It's counter-intuitive because not-equal null automagically makes that work, but not-in has no equivalent.

Firestore/core/src/core/key_field_not_in_filter.cc
```diff
@@ -0,0 +1,64 @@
...
+namespace firebase {
+namespace firestore {
+namespace core {
+
+using model::Document;
+using model::DocumentKey;
+using model::FieldPath;
+using model::FieldValue;
+
+using Operator = Filter::Operator;
```
wilhuff, line 36: It seems like `query_core.cc` could benefit quite a bit from applying this using statement there (it would allow you to shorten all the `Filter::Operator::Foo` to just `Operator::Foo`). Here we're only accessing the shortened `Operator` just the once. This inconsistency seems glaring.

Firestore/core/src/core/not_in_filter.h
```diff
@@ -0,0 +1,49 @@
...
+#define FIRESTORE_CORE_SRC_CORE_NOT_IN_FILTER_H_
+
+#include <string>
+
+#include "Firestore/core/src/core/field_filter.h"
+
+namespace firebase {
+namespace firestore {
+
+namespace model {
```
wilhuff, line 27: You can get these forward declarations by including `Firestore/core/src/model/model_fwd.h`.

Also, this is inconsistent with `key_field_not_in_filter.h` which neither makes the forward declarations nor includes the `model_fwd.h` header.

Firestore/core/src/core/query.h
```diff
@@ -118,16 +118,11 @@ class Query {
...
    */
-  absl::optional<Filter::Operator> FirstArrayOperator() const;
-
-  /**
-   * Returns the first disjunctive operator (IN or array-contains-any) found
-   * on a filter, or absl::nullopt if there are no disjunctive operators.
-   */
-  absl::optional<Filter::Operator> FirstDisjunctiveOperator() const;
+  absl::optional<Filter::Operator> FindOperator(
+      std::vector<Filter::Operator> ops) const;
```
wilhuff, line 125: Since `FindOperator` isn't storing the `ops` value in a field (or otherwise retaining it beyond the lifetime of the call), you should pass `ops` via const reference.

Firestore/core/test/unit/remote/serializer_test.cc
```diff
@@ -369,9 +369,10 @@ class SerializerTest : public ::testing::Test {
     create_time_proto->set_nanos(4321);
   }
 
-  void ExpectUnaryOperator(const FieldValue& value,
+  void ExpectUnaryOperator(std::string opStr,
```
wilhuff, line 372: In C++ variables should be snake_case.

wilhuff (changes requested):

Firestore/Source/Public/FirebaseFirestore/FIRQuery.h
```diff
@@ -109,6 +109,30 @@ NS_SWIFT_NAME(Query)
...
+/**
+ * Creates and returns a new `FIRQuery` with the additional filter that documents must
+ * contain the specified field and the value does not equal the specified value.
+ *
+ * @param path The path of the field to compare.
+ * @param value The value the field must be equal to.
+ *
+ * @return The created `FIRQuery`.
+ */
+- (FIRQuery *)queryWhereFieldPath:(FIRFieldPath *)path
```
wilhuff, line 121: Also, we can't make these public API changes until we're ready to launch. Are you going to hold this PR indefinitely?

In the past we've made these part of the Internal Query category to be able to commit the bulk of the code and tests and just delay the public release.

Firestore/Source/Public/FirebaseFirestore/FIRQuery.h
wilhuff: Yeah, this is tough. Here's my take, which avoids "results in no document matches" which is a fairly indirect way of saying that the filter can't match the field.

> One special case is that `notIn` filters cannot match `nil` values. To query for documents where a field exists and is `nil`, use a `notEqual` filter, which can handle this special case.

Separately, note that Objective-C and Swift spell "null" as `nil`.

Firestore/core/src/api/query_core.cc
```diff
@@ -333,7 +360,8 @@ void Query::ValidateOrderByField(const FieldPath& order_by_field,
   if (order_by_field != inequality_field) {
     ThrowInvalidArgument(
         "Invalid query. You have a where filter with an inequality "
-        "(lessThan, lessThanOrEqual, greaterThan, or greaterThanOrEqual) on "
+        "(notEqual, lessThan, lessThanOrEqual, greaterThan, or "
```
wilhuff, line 363: nit: reflow?

Firestore/core/src/api/query_core.h
```diff
@@ -20,6 +20,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <vector>
```
wilhuff, line 23: No longer needed?

wilhuff (commented):
Very close to LGTM.

Firestore/core/src/api/query_core.cc
thebrianchen: done.

wilhuff (approved):
LGTM

PR #6423, opened 2020-09-09
Test workflows with tag.

granluo:


paulb777 (commented):
It looks like the tag is only used in release.yml.

What are the changes in the other files for?

granluo:
> It looks like the tag is only used in release.yml.
> 
> What are the changes in the other files for?

release.yml

granluo:
> It looks like the tag is only used in release.yml.
> 
> What are the changes in the other files for?

release.yml will upload quickstart artifacts as well after tests failed.

release_testing_setup.sh is to push tag silent git clone to avoid printing out the token, also replace tags in podpecs which will be pushed to a spec repo.

remove_data.sh is a bug before, it is fixed and also set -e to stop execution when the script has errors.

setup_quickstart.sh is to update podfile.lock/pod versions to specstesting repo for release testing.

paulb777 (approved):

PR #6428, opened 2020-09-09
Ipv4 ipv6 fix

eldhosembabu:
Recreating PR https://github.com/firebase/firebase-ios-sdk/pull/5966

Instead of using firebasedynamiclinks-ipv4.googleapis.com and firebasedynamiclinks-ipv6.googleapis.com for doing attribution calls, we are moving towards using firebasedynamiclinks.googleapis.com.

Fixes https://github.com/firebase/firebase-ios-sdk/issues/5032

-changed  ipv4 and ipv6 attribution calls with a single direct url.
-Removed and refactored code to get rid of code to handle consolidated results from ipv4 and ipv6 result.
-Made changes to Test app to show alert dialog when universal linking when app is not in background case and also for fresh start case.


paulb777:
@googlebot I fixed it 

Confirmed that the code was developed with a valid CLA.

paulb777 (approved):
Thanks!

eldhosembabu:
@googlebot I fixed it

paulb777 (approved):
Great to delete so much code!

FirebaseDynamicLinks/Tests/Sample/FDLBuilderTestAppObjC/AppDelegate.m
```diff
@@ -92,7 +92,10 @@ - (void)_showDynamicLinkInfo:(FIRDynamicLink *)dynamicLink {
   [alertVC addAction:[UIAlertAction actionWithTitle:@"Dismiss"
                                               style:UIAlertActionStyleCancel
                                             handler:NULL]];
-  [self.window.rootViewController presentViewController:alertVC animated:YES completion:NULL];
+
+  [[UIApplication sharedApplication].keyWindow.rootViewController presentViewController:alertVC
```
ryanwilson, line 96: Curious why this changed from `self` and if the change was related to the IPV4/6 change.

ryanwilson (approved):
Great, thanks for fixing this! And thanks @ncooke3 for the initial commits 😄

FirebaseDynamicLinks/Tests/Sample/FDLBuilderTestAppObjC/AppDelegate.m
eldhosembabu: No this is not related with the ipv4/6. While testing the changes, I've seen some issues with the test app.
1. After the fresh install, when the app launches and when we try to show the Alert dialog from the AppDelegate when we receive an FDL, the alert was failing with "First responder error: non-key window attempting reload - allowing due to manual keyboard".

2. Alert was failing when universal linking when the app is started as fresh instead of opening from background.

eldhosembabu:
Did a round of testing on the basic functionalities based on this PR and all seems working. Will merge this PR by EOD today if I'm not seeing any issues with this change

PR #6439, opened 2020-09-10
Remove unnecessary Core private headers

### LINKED ISSUE #3777: FirebaseCore Private header cleanup

paulb777:
Fix #3777

#no-changelog


FirebaseCore/Sources/Private/FIRAppInternal.h
```diff
@@ -110,7 +104,7 @@ extern NSString *const FIRAuthStateDidChangeInternalNotificationUIDKey;
  * FIRApp notification listener.
  */
 + (NSError *)errorForSubspecConfigurationFailureWithDomain:(NSString *)domain
```
ryanwilson, line 106: This method is only used once in FDL and once in RC - think it's worth keeping it in Core for a ~5 line implementation? Diff wise it'd probably be less LOC to just inline the implementation at both call sites, and cleans up Core.

paulb777:
Done.

FirebaseCore/Sources/Private/FIRAppInternal.h
paulb777: Done.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -185,11 +185,10 @@ - (void)configureDynamicLinks:(FIRApp *)app {
...
   } else {
-    error =
-        [FIRApp errorForSubspecConfigurationFailureWithDomain:kFirebaseDurableDeepLinkErrorDomain
-                                                    errorCode:FIRErrorCodeDurableDeepLinkFailed
-                                                      service:@"DynamicLinks"
-                                                       reason:errorDescription];
+    error = [self errorForSubspecConfigurationFailureWithDomain:kFirebaseDurableDeepLinkErrorDomain
+                                                      errorCode:FIRErrorCodeDurableDeepLinkFailed
+                                                        service:@"DynamicLinks"
+                                                         reason:errorDescription];
```
ryanwilson, line 191: Since this call is only used once and it's the same number of lines, I think this could even be inlined.

ryanwilson (approved):

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
paulb777: Done. Thanks!

PR #6443, opened 2020-09-11
Add the ability to query the compiler version and the C++ standard being used

var-const:
This is largely adapted from the [Cloud C++ APIs](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/internal/compiler_info.cc) with the following changes:
* added C++20 to the list of possible standards;
* added "latest" as a possible C++ standard for MSVC;
* added the C++ standard library vendor to the compiler info string, primarily to track STLPort.

The string I'm getting when running this code on my Mac is `AppleClang-11.0.3.11030032-ex-2011-libcpp`.

#no-changelog

### DIFF
--- a/Firestore/core/src/api/compiler_info.cc
+++ b/Firestore/core/src/api/compiler_info.cc
@@ -1,0 +1,29 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "Firestore/core/src/api/compiler_info.h"
+
+#include "Firestore/core/src/util/compiler_info.h"
+
+namespace firebase {
+namespace firestore {
+namespace util {
+
+std::string GetFullCompilerInfo() {
+  return api::GetFullCompilerInfo();
+}
+
+}  // namespace util
+}  // namespace firestore
+}  // namespace firebase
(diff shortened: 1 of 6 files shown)


Firestore/core/src/api/compiler_info.cc
```diff
@@ -0,0 +1,29 @@
...
+
+#include "Firestore/core/src/api/compiler_info.h"
+
+#include "Firestore/core/src/util/compiler_info.h"
+
+namespace firebase {
+namespace firestore {
+namespace util {
+
+std::string GetFullCompilerInfo() {
```
var-const, line 23: Let me know if you think this is worth the trouble. The idea was to ensure that the public C++ SDK doesn't need to use anything except `api`, while the internal code in core never needs to access `api` (basically, that `api` is a layer in-between where the hierarchy of calls goes `public -> api -> core`, but never the reverse).

Firestore/core/src/util/compiler_info.cc
```diff
@@ -0,0 +1,168 @@
...
+      // `/std:c++latest` option is specified".
+      if (FIRESTORE__CPLUSPLUS > 202002L) {
+        return "latest";
+      }
+#endif  // _MSC_VER
+      return "unknown";
+  }
+}
+
+std::string StandardLibraryVendor() {
```
var-const, line 146: What do you think? It seemed like a worthwhile addition, though it should also be obtainable via Firebase logging in the future.

Firestore/core/src/util/compiler_info.h
```diff
@@ -0,0 +1,40 @@
...
+namespace firebase {
+namespace firestore {
+namespace util {
+
+// Returns a string describing the compiler version and settings in the
+// following format:
+//
+//   <CompilerId>-<CompilerVersion>-<CompilerFeatures>-<LanguageVersion>-<StandardLibraryVersion>
+//
+// e.g. "AppleClang-11.0.3.11030032-ex-2011-libcpp".
```
var-const, line 29: Let me know if you think this format is too verbose (the full compiler version number seems overkill).

Firestore/core/src/util/exception.h
```diff
@@ -37,6 +37,7 @@
 
 #include "Firestore/core/src/util/string_format.h"
 #include "absl/base/attributes.h"
+#include "absl/base/config.h"
```
var-const, line 40: I noticed that the header was missing -- `ABSL_HAVE_EXCEPTIONS` is defined in it.

var-const:
On a second thought, this should be moved into the public C++ (I forgot Android cannot access `api`).

Issue #6425, opened 2020-09-09
FirebaseAnalytics symbols not found when installed via SPM


zeusent:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: `12.0 beta 6 (12A8189n)`
  * Firebase SDK version: `6.32-spm-beta`
  * Firebase Component: `Analytics`
  * Component version: `6.32.0`
  * Installation method: `Swift Package Manager`

#### Steps to reproduce:

- I've installed Firebase via SPM using the instructions found at [this](https://github.com/firebase/firebase-ios-sdk/blob/master/SwiftPackageManager.md) page.
- I've also added the `-ObjC` flag as mentioned because I am using FirebaseAnalytics.

When compiling my project I get a bunch of errors like the ones bellow:

```
Undefined symbols for architecture arm64:
  "_kFIRParameterShipping", referenced from:
      {edited}
  "_kFIREventAddPaymentInfo", referenced from:
      {edited}
  "_kFIRParameterCheckoutOption", referenced from:
      {edited}
  "_kFIREventCheckoutProgress", referenced from:
      {edited}
  ...
ld: symbol(s) not found for architecture arm64
```

I get the same errors when trying to run on the simulator but the architecture is `x86_64` instead (obviously).

ryanwilson:
Hi @zeusent - sorry for the troubles you're seeing.

Based on your reproduction steps, I'm unable to reproduce this. I created a new project and followed the instructions linked, added the `-ObjC` flag, using `import Firebase` or `import FirebaseAnalytics` and then calling `Analytics.logEvent("Test", [:])` succeeds in a simulator build as well as building for device (I tested via Archiving). 

A few other questions to try to make progress:
- Was it a blank project that you started with or an existing project?
- If it was an existing project, did you have Firebase integrated already? How did you remove it?

zeusent:
Hi @ryanwilson! Thanks for your fast reply.

The project was an existing project and Firebase was added via CocoaPods. I've removed the lines in my Podfile and executed `pod update` to remove it.

I see in your example that you use a hardcoded string value for the event name and no parameters. Can you use one of the pre-defined events and parameters that Firebase provides? Something likes this:

```Analytics.logEvent(AnalyticsEventEcommercePurchase, parameters: [AnalyticsParameterShipping: 10.0])```

It seems that those events and parameters definitions are the issue while compiling...

paulb777:
[...]

PR #6455, opened 2020-09-14
Move AnalyticsWrapper headers to Public dir.

ryanwilson:
This is a test to see if #6425 is fixed by this change.


ryanwilson:
As a follow up: this will break anyone using quote syntax for imports: `#import "FirebaseAnalytics/FirebaseAnalytics.h"` so the right approach would likely be to put headers here with the same name that just import the actual headers. Kind of gross, but it should work.

PR #6457, opened 2020-09-14
GDTTransformer: fix background tasks handling 

### LINKED ISSUE #6258: GDTTransformer Can't endBackgroundTask.
* Xcode version: Xcode 11.4 (11E146)

#### Steps to reproduce:
Don't know which component crash. We used pod to get firebase like this.
  
> pod 'FirebaseInstanceID'
>   pod 'Firebase/Analytics'
>   pod 'Firebase/RemoteConfig'
>   pod 'Firebase/Crashlytics'
>   pod 'Firebase/Performance'
>   pod 'Firebase/InAppMessaging'

When my phone upgrade to iOS 14 Beta,  and app enter background. Later it will crash.  Crash log is 
`[BackgroundTask] Background task still not ended after expiration handlers were called: <_UIBackgroundTaskInfo: 0x2835e06c0>: taskID = 64, taskName = CCTLogWriter, creationTime = 64357 (elapsed = 0). This app will likely be terminated by the system. Call UIApplication.endBackgroundTask(_:) to avoid this.
Message from debugger: Terminated due to signal 9`
And when I use `UIApplicationEndBackgroundTaskError()` to debug where goes wrong,  It breakpoint here.
F7D914F9D38C4B9558FE20619DC154FE

Could you please check this method to see if the backgroundtask is closed?

maksymmalyhin:
Fixes #6258.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
@@ -56,11 +56,6 @@
       withTransformers:(NSArray<id<GDTCOREventTransformer>> *)transformers
             onComplete:(void (^_Nullable)(BOOL wasWritten, NSError *_Nullable error))completion {
   GDTCORAssert(event, @"You can't write a nil event");
-  BOOL hadOriginalCompletion = completion != nil;
-  if (!completion) {
-    completion = ^(BOOL wasWritten, NSError *_Nullable error) {
-    };
-  }
 
   __block GDTCORBackgroundIdentifier bgID = GDTCORBackgroundIdentifierInvalid;
   bgID = [self.application beginBackgroundTaskWithName:@"GDTTransformer"
@@ -68,6 +63,17 @@
                                        [self.application endBackgroundTask:bgID];
                                        bgID = GDTCORBackgroundIdentifierInvalid;
                                      }];
+
+  __auto_type completionWrapper = ^(BOOL wasWritten, NSError *_Nullable error) {
+    if (completion) {
+      completion(wasWritten, error);
+    }
+
+    // The work is done, cancel the background task if it's valid.
+    [self.application endBackgroundTask:bgID];
+    bgID = GDTCORBackgroundIdentifierInvalid;
+  };
+
   dispatch_async(_eventWritingQueue, ^{
     GDTCOREvent *transformedEvent = event;
     for (id<GDTCOREventTransformer> transformer in transformers) {
@@ -75,13 +81,13 @@
         GDTCORLogDebug(@"Applying a transformer to event %@", event);
         transformedEvent = [transformer transform:transformedEvent];
         if (!transformedEvent) {
-          completion(NO, nil);
+          completionWrapper(NO, nil);
           return;
         }
       } else {
         GDTCORLogError(GDTCORMCETransformerDoesntImplementTransform,
                        @"Transformer doesn't implement transform: %@", transformer);
-        completion(NO, nil);
+        completionWrapper(NO, nil);
         return;
       }
     }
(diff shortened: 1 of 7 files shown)


GoogleDataTransport/GDTCORLibrary/Private/GDTCORTransformer_Private.h
```diff
@@ -16,13 +16,22 @@
...
+@protocol GDTCORApplicationProtocol;
+
 NS_ASSUME_NONNULL_BEGIN
 
 @interface GDTCORTransformer ()
 
 /** The queue on which all work will occur. */
 @property(nonatomic) dispatch_queue_t eventWritingQueue;
 
+/** The application instance is used to begin/end background tasks.  */
```
mikehaney24, line 28: instance that is used

GoogleDataTransport/GDTCORTests/Unit/GDTCORTransformerTest.m
```diff
@@ -125,6 +148,41 @@ - (void)testWriteEventWithTransformersThatCreateANewEvent {
...
+/** Sets  GDTCORApplicationFake handlers to expect the begin and the end of a background task with
+ * the specified name.
+ *  @return An array with the task begin and end XCTestExpectation.
+ */
+- (NSArray<XCTestExpectation *> *)expectationsBackgroundTaskBeginAndEndWithName:
+    (NSString *)expectedName {
+  XCTestExpectation *beginExpectation = [self expectationWithDescription:@"Background task begin"];
+  XCTestExpectation *endExpectation = [self expectationWithDescription:@"Background task end"];
+
+  GDTCORBackgroundIdentifier taskID = rand();
```
mikehaney24, line 166: I know it doesn't matter here, but I'd prefer something in the arc4random family just to prevent this code from being flagged for archaic API usage by external analyzers.

mikehaney24 (approved):
Wow, good find!

GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
```diff
@@ -51,45 +56,46 @@ - (void)transformEvent:(GDTCOREvent *)event
...
-                  [[GDTCORApplication sharedApplication] endBackgroundTask:bgID];
-                  bgID = GDTCORBackgroundIdentifierInvalid;
-                }];
+  bgID = [self.application beginBackgroundTaskWithName:@"GDTTransformer"
+                                     expirationHandler:^{
+                                       [self.application endBackgroundTask:bgID];
+                                       bgID = GDTCORBackgroundIdentifierInvalid;
+                                     }];
+
+  __auto_type completionWrapper = ^(BOOL wasWritten, NSError *_Nullable error) {
```
ryanwilson, line 67: Do we use `__auto_type` already in the SDK? Want to make sure it doesn't break some older Xcode versions or anything.

GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
```diff
@@ -39,10 +39,15 @@ + (instancetype)sharedInstance {
 }
 
 - (instancetype)init {
+  return [self initWithApplication:self.application];
```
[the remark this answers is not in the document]
maksymmalyhin: Great catch, I meant passing the shared application!

GoogleDataTransport/GDTCORLibrary/GDTCORTransformer.m
maksymmalyhin: We've used it in tests only so far, but AFAIK it was introduced several years ago (sorry, could not find release notes for it though). Xcode 10.3 tests pass on Travis. I do you think we can proceed with it and change it back if we have a complain?

visumickey (approved):

PR #6469, opened 2020-09-15
GDT tests: Xcode 12 fixes

maksymmalyhin:
- `XCAssert` maros family doesn't use `self` reference in Xcode 12 any more, so using `XCAssert...` in blocks doesn't require  `weakSelf`. Added `__unused ` modifier to silent the Xcode 12 warnings that fail `pod lib lint`. 

#no-changelog


ryanwilson (approved):
This LGTM as long as it doesn't cause warnings to appear when linting with Xcode 11

PR #6476, opened 2020-09-16
Add auth emulator to public API in prerelease branch.

rosalyntan:
Add the auth emulator entrypoint to the public API in a prerelease branch.


renkelvin (approved):
LGTM on CI green

rosalyntan:
I chatted with @ryanwilson and the CI failure related to FirebaseAnalytics versioning can be ignored, as it is because the auth-emulator-prerelease branch is built on top of the current release branch and the new FirebaseAnalytics pod does not exist publicly yet. As this prerelease branch is meant to be used by the FirebaseAuth pod only, this is fine.

renkelvin:
> I chatted with @ryanwilson and the CI failure related to FirebaseAnalytics versioning can be ignored, as it is because the auth-emulator-prerelease branch is built on top of the current release branch and the new FirebaseAnalytics pod does not exist publicly yet. As this prerelease branch is meant to be used by the FirebaseAuth pod only, this is fine.

Sounds good.

PR #6481, opened 2020-09-17
Update release workflow

granluo:
This update is prepared for released candidate nightly testing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Update SpecsTesting repo setup
       run: |
          ossbotaccess=`cat oss-bot-access.txt`
-         BOT_TOKEN="${ossbotaccess}" nightly_test_version="${nightly_version}" scripts/release_testing_setup.sh
+         BOT_TOKEN="${ossbotaccess}" nightly_test_version="${nightly_version}" scripts/release_testing_setup.sh nightly_testing
     - name: Update SpecsTesting repo
       run: |
         botaccess=`cat bot-access.txt`
@@ -44,7 +44,6 @@ jobs:
       run: |
         pod repo remove "${local_repo}"
         rm -rf bot-access.txt
-
   remoteconfig_quickstart:
     # Don't run on private repo unless it is a PR.
     if: github.repository != 'FirebasePrivate/firebase-ios-sdk' || github.event_name == 'pull_request'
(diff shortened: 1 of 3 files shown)


scripts/release_testing_setup.sh
```diff
@@ -14,28 +14,41 @@
...
-# so error is allowed here.
-git push origin --delete "${tag_version}"
-set -e
-git tag -f -a "${tag_version}" -m "release testing"
-git push origin "${tag_version}"
-
-# Update source and tag, e.g.  ":tag => 'CocoaPods-' + s.version.to_s" to
-# ":tag => test"
-sed  -i "" "s/\s*:tag.*/:tag => '${tag_version}'/" *.podspec
+if [ -n "$tag_version" ]; then
```
granluo, line 39: will only push a tag if tag_version is not null.

scripts/setup_quickstart.sh
```diff
@@ -40,11 +40,17 @@ if check_secrets || [[ ${SAMPLE} == "installations" ]]; then
 
   git clone https://github.com/firebase/quickstart-ios.git
   $scripts_dir/localize_podfile.swift quickstart-ios/"$SAMPLE"/Podfile "$RELEASE_TESTING"
-  if [ ! -z "$RELEASE_TESTING" ]; then
+  if [ "$RELEASE_TESTING" == "nightly_release_testing" ]; then
```
granluo, line 43: update the flag of nightly release testing to "nightly_release_testing" and the flag for release candidate to  "prerelease_testing". Prerelease testing will be set up in another pr.

.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Update SpecsTesting repo setup
       run: |
          ossbotaccess=`cat oss-bot-access.txt`
-         BOT_TOKEN="${ossbotaccess}" nightly_test_version="${nightly_version}" scripts/release_testing_setup.sh
+         BOT_TOKEN="${ossbotaccess}" test_version="${nightly_version}" scripts/release_testing_setup.sh nightly_testing
```
paulb777, line 34: Where is test_version used?

paulb777 (approved):

.github/workflows/release.yml
granluo: It is used in release_testing_setup.sh to update a tag.
https://github.com/firebase/firebase-ios-sdk/blob/38af2d564336b06efdd24920de08230364d4957b/scripts/release_testing_setup.sh#L31-L35

granluo: test_version should be updated in the workflow and testing tags need to be removed manually from repo at this moment.

Issue #6483, opened 2020-09-17
No known class method for selector 'archivedDataWithRootObject:requiringSecureCoding:error:'


sryze:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Version 9.2 (9C40b)
  * Firebase SDK version: 6.28.1
  * Firebase Component: GoogleDataTransport
  * Component version: 7.1.0
  * Installation method: `CocoaPods`

#### Steps to reproduce:

Compile on Xcode 9.2

Actual compiler output:

```
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:204:35: error: no known class method for selector 'archivedDataWithRootObject:requiringSecureCoding:error:'
    resultData = [NSKeyedArchiver archivedDataWithRootObject:obj
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:260:45: error: no known class method for selector 'unarchivedObjectOfClass:fromData:error:'
      unarchivedObject = [NSKeyedUnarchiver unarchivedObjectOfClass:archiveClass
                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
```

#### Relevant Code:

```obj-c
NSData *_Nullable GDTCOREncodeArchive(id<NSSecureCoding> obj,
                                      NSString *archivePath,
                                      NSError *_Nullable *error) {
  BOOL result = NO;
  if (archivePath.length > 0) {
    result = [[NSFileManager defaultManager]
[...]

Issue #6482, opened 2020-09-17
GoogleDataTransport pod does not compile in Xcode 9


sryze:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Version 9.2 (9C40b)
  * Firebase SDK version: 6.28.1
  * Firebase Component: GoogleDataTransport
  * Component version: 7.1.0
  * Installation method: `CocoaPods`

#### Steps to reproduce:

Compile on Xcode 9.2

Actual compiler output:

```
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:144:3: error: control reaches end of non-void function [-Werror,-Wreturn-type]
  }
  ^
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:146:3: error: expected identifier or '('
  if (networkCurrentRadioAccessTechnology) {
  ^
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:150:5: error: expected identifier or '('
  } else {
    ^
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:156:1: error: extraneous closing brace ('}')
}
^
myapp/ios/Pods/GoogleDataTransport/GoogleDataTransport/GDTCORLibrary/GDTCORPlatform.m:203:35: error: no known class method for selector 'archivedDataWithRootObject:requiringSecureCoding:error:'
    resultData = [NSKeyedArchiver archivedDataWithRootObject:obj
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #6486, opened 2020-09-17
GDT: remove pre Xcode 10 availability checks

maksymmalyhin:
Related to #6482 and #6483

- Keep Xcode 10 compatibility so far as we still verify it on Travis.


paulb777 (approved):
LGTM on green

PR #6487, opened 2020-09-17
Add release candidate prerelease workflow.

granluo:


granluo:
The release-6.33.0 branch has an unreleased FirebaseAnalytics sdk, which is updated to a released one so the spec repo could be built.

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,468 @@
...
+        cd scripts/create_spec_repo/
+        swift build
+        pod repo add --silent "${local_repo}" https://"$botaccess"@github.com/FirebasePrivate/SpecsReleasing.git
+        BOT_TOKEN="${botaccess}"  .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}" --sdk-repo-name SpecsReleasing --pod-sources "https://${botaccess}@github.com/FirebasePrivate/SpecsReleasing" --pod-sources "https://cdn.cocoapods.org/"
+    - name: Clean Artifacts
+      if: ${{ always() }}
+      run: |
+        pod repo remove "${local_repo}"
+        rm -rf bot-access.txt
+  remoteconfig_quickstart:
```
paulb777, line 46: Sort remoteconfig

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,468 @@
...
+    # Don't run on private repo unless it is a PR.
```
paulb777, line 47: It would require a few special case tests, but wondering if we should to all the jobs via one matrixed job?

granluo: I made some experiments and it turns out  we might need to update scripts to have matrix. E.g. Crashlytics need to have two binaries,run and upload-symbols. Also, some tests have objc while some do not have. I think we might submit this pr to run prerelease test first, and another pr to refactor nightly release and prerelease workflows.

paulb777: SGTM

paulb777 (approved):
LGTM on green

granluo:
Sources of release/prerelease tests are updated to cocoapods repo from cdn and tests are passed. The failed test is caused by cdn. The workflow works fine and will push.

PR #6498, opened 2020-09-18
Remove unnecessary analytics public header copy

paulb777:
It seems the public headers embedded in the xcframework are sufficient. I'm not sure if there was a bug in an earlier Xcode 12 beta or user error for thinking that they were needed.

I added tests to verify the module imports for iOS tests.

I verified the iOS Analytics quickstart

There's a reasonable chance this could be the explanation for #6425


paulb777:
Whether or not this fixes #6425, should be a worthwhile cleanup and I'm proposing it to land it in the 6.33-spm-beta branch.

ryanwilson (approved):
Awesome!

maksymmalyhin (approved):

Issue #4557, opened 2019-12-27
FR: Analytics support on watchOS


jroets:
Please add Firebase Analytics support to SDK on watchOS, so that we can record to Firebase Analytics directly from watch apps in the same way we can on iOS apps.

To my knowledge right now, to record analytics data from a watch app to Firebase Analytics requires proxying through the SDK running on the iPhone. Or else don't use Firebase for analytics and instead record to GA using its web API (there is no GA SDK for watchOS). Or use a non-Google analytics provider that actually does offer a watchOS SDK, like Yahoo's Flurry SDK. Less than ideal for those of us who want to use Firebase Analytics from a watch app.

hannesjung:
+1

nesterman:
+1

jostster:
+1, Is analytics for iOS open source? I'm not seeing the code for doing community support for watchOS.

paulb777:
@jostster FirebaseAnalytics is not open source.

jostster:
@paulb777 Do you know if it is on the roadmap to allow it to work on watchOS? It seems https://github.com/firebase/firebase-ios-sdk/issues/4558 will be held up by the community without the watchOS support on Analytics since it looks like the FirebaseCrashlytics backend activation requires the Analytics to ping the server.

paulb777:
@jostster I can't speak about roadmaps, but I can say that watchOS is a lower priority than Catalyst, tvOS, and macOS.

Also, Crashlytics does have support for Catalyst, even thought Analytics is not available.

kutsin:
<img width="442" alt="Screenshot1" src="https://user-images.githubusercontent.com/14885155/93630755-29b02400-f9f3-11ea-9e2e-35a368b75321.png">

Any updates on this?

paulb777:
@kutsin Nothing new to report about Analytics platform coverage.

We'll eliminate that warning in an upcoming release in the meantime.

SwiftiSwift:
Still no watchOS support for analytics?

paulb777:
Not yet. We introduced macOS, catalyst and tvOS in 8.9.0 and will try to follow up with watchOS.

Please keep adding thumbs-up's to this feature request to help us with prioritization.

plotkin:
+1

curiousdustin:
@paulb777, Thank you for the updates on this.

Are you able to tell us the nature of the hold up on this?

Is it simply a matter of prioritization and dev resources?

[...]

PR #6500, opened 2020-09-18
Generate missing Analytics warning only for iOS

paulb777:
And simplify warning message so it works for any package manager.

Eliminate the warning since it is not currently possible to solve for macOS, tvOS and watchOS and missing Analytics support is documented.

See https://github.com/firebase/firebase-ios-sdk/issues/4557#issuecomment-695011515

For Firebase 7, consider completely eliminating these for iOS as well.

cc: @christibbs @chliangGoogle @eldhosembabu @visumickey @karenyz


maksymmalyhin (approved):
LGTM

PR #6507, opened 2020-09-21
Integrate Firestore with Firebase platform logging

var-const:
The sample user agent string I'm getting in XCode is `apple-platform/ios apple-sdk/17E8258 fire-fst/1.17.1 fire-ios/6.10.2 swift/true xcode/11E503a`.


Firestore/core/src/util/config.h
```diff
@@ -1,57 +0,0 @@
-/*
```
wilhuff, line 1: This file shouldn't be deleted (unless there's a replacement?).

Note that config.h was previously generated, but this was changed for compatibility with Swift Package Manager which does not allow for build-time configuration scripts.

Firestore/core/src/firebase/firestore/util/config.h
```diff
@@ -0,0 +1,33 @@
+/*
```
wilhuff, line 1: This was the old location of the file (before we shortened all the paths). This should be removed.

Firestore/core/src/remote/firebase_platform_logging_apple.h
```diff
@@ -0,0 +1,53 @@
...
+
+@class FIRApp;
+
+namespace firebase {
+namespace firestore {
+namespace remote {
+
+class FirebasePlatformLoggingApple : public FirebasePlatformLogging {
+ public:
+  explicit FirebasePlatformLoggingApple(FIRApp* app);
```
wilhuff, line 36: Objective-C types should have their nullability defined, either by wrapping in `NS_ASSUME_NONNNULL_BEGIN`/`END` or by annotating the pointers directly (`FIRApp* _Nonnull app`) .

Firestore/core/src/remote/firebase_platform_logging_apple.mm
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/remote/firebase_platform_logging_apple.h"
+
+#import "FirebaseCore/Sources/Private/FIRAppInternal.h"
+#import "FirebaseCore/Sources/Private/FIRHeartbeatInfo.h"
+#import "FirebaseCore/Sources/Private/FIROptionsInternal.h"
+//#import "FirebaseCore/Sources/Public/FirebaseCore/FIRApp.h"
```
wilhuff, line 22: Please remove commented-out code.

Firestore/core/src/remote/firebase_platform_logging.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace firebase {
+namespace firestore {
+namespace remote {
+
+/**
+ * Wraps the platform-dependent functionality associated with Firebase platform
+ * logging.
+ */
+class FirebasePlatformLogging {
```
wilhuff, line 30: In the Android implementation, we have a notion of a [`GrpcMetadataProvider`](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcMetadataProvider.java) which does everything required to set the headers. Doing something similar here could make this a much narrower interface.

On the other hand, an interface like this one makes this information independent of the gRPC implementation which has some merit. Do you have any thoughts on how to make these platforms work similarly? This doesn't seem like something that has to be different on the different platforms.

Firestore/core/test/unit/remote/grpc_stream_tester.h
```diff
@@ -248,6 +249,7 @@ class GrpcStreamTester {
 
   grpc::CompletionQueue grpc_queue_;
   FakeGrpcQueue fake_grpc_queue_;
+  std::unique_ptr<FirebasePlatformLogging> firebase_platform_logging_;
```
wilhuff, line 252: `FirebasePlatformLogging` could be forward declared in this header.

Firestore/core/src/remote/grpc_connection.h
```diff
@@ -25,6 +25,7 @@
 #include "Firestore/core/src/auth/token.h"
 #include "Firestore/core/src/core/database_info.h"
 #include "Firestore/core/src/remote/connectivity_monitor.h"
+#include "Firestore/core/src/remote/firebase_platform_logging.h"
```
wilhuff, line 28: Use a forward declaration? Here and in all other headers.

Firestore/core/src/api/firestore.h
```diff
@@ -113,6 +115,8 @@ class Firestore : public std::enable_shared_from_this<Firestore> {
   std::shared_ptr<util::Executor> user_executor_;
   std::shared_ptr<util::AsyncQueue> worker_queue_;
 
+  std::unique_ptr<remote::FirebasePlatformLogging> firebase_platform_logging_;
```
wilhuff, line 118: This should be declared before `FirestoreClient` so that it outlives that instance during destruction. Indeed, it seems like `FirestoreClient` should be among the last things declared in here.

Firestore/core/src/remote/firebase_platform_logging.h
wilhuff, line 30: Consider renaming this to a regular noun rather than a gerund (i.e. "logger" vs "logging").

wilhuff (commented):
.

Firestore/core/src/remote/firebase_platform_logging.h
wilhuff: The Android implementation is also imperfect. My original thought there was that separate concerns could supply separate metadata providers. That is, Firebase headers from one provider, Cloud headers from another, Auth perhaps from yet another. From that point of view, I think what you have here fits that goal quite well. Whether or not you see any benefit is another matter.

wilhuff: In a forest, the person who cuts the trees is a "logger", while the larger practice is "logging". They're already grammatically related without resorting to calling the lumberjack a "tree logging logger".

My general skepticism with the "firebase platform logging" name comes from the fact that the term names the larger effort including something client-side that sends data, server-side collection, logs that are written, and the warehousing scripts to bucket the data. Since this is just a part of that larger effort, it seems misleading to name this smaller part with the name of whole thing.

To be fully descriptive you might name this thing "firebase platform logging metadata provider", since this doesn't really log data so much as attach headers to a request (aka provide request metadata). I think that's probably more of a mouthful than is warranted, given this thing's narrow scope.

If you agree with my idea above about (eventually) partitioning the header-providing bits along the lines of Firebase/Cloud/Auth/etc, possibly it's just sufficient to call this a FirebaseMetadataProvider.

Thoughts?

Firestore/core/src/remote/firebase_platform_logging_apple.mm
```diff
@@ -0,0 +1,68 @@
...
+  context.AddMetadata(kXFirebaseClientHeader, GetUserAgent());
+  context.AddMetadata(kXFirebaseClientLogTypeHeader, GetHeartbeat());
+
+  std::string gmp_app_id = GetGmpAppId();
+  if (!gmp_app_id.empty()) {
+    context.AddMetadata(kXFirebaseGmpIdHeader, gmp_app_id);
+  }
+}
+
+std::string FirebasePlatformLoggingApple::GetUserAgent() const {
```
wilhuff, line 52: Completely optional: with the exception of `GetGmpAppId`, none of these access any state in this instance. They could just be free functions in the anonymous namespace. Even `GetGmpAppId` could be phrased that way (by taking the `FIRApp*` as a parameter).

Alternatively, you could make these two static rather than const members.

Firestore/core/src/remote/grpc_connection.h
```diff
@@ -109,6 +112,8 @@ class GrpcConnection {
 
   void RegisterConnectivityMonitor();
 
+  void AddFirebasePlatformLoggingHeader(grpc::ClientContext& context) const;
```
wilhuff, line 115: This declaration does not seem to have a definition (but perhaps I missed it)? It also seems not to have any uses.

Firestore/core/src/remote/firebase_platform_logging_noop.h
```diff
@@ -0,0 +1,42 @@
...
+#include <memory>
+#include <string>
+
+#include "Firestore/core/src/remote/firebase_platform_logging.h"
+
+namespace firebase {
+namespace firestore {
+namespace remote {
+
+class FirebasePlatformLoggingNoOp : public FirebasePlatformLogging {
```
wilhuff, line 29: Optional: if the base class were to have a default no-op implementation, this class would be unnecessary.

At that point, anyone that wanted an implementation could also just `absl::make_unique<FirebasePlatformLogging>()` for themselves, also removing the need for `CreateNoOpFirebasePlatformLogging`.

Note that I have no strong preference about whether or not there's an explicit no-op type. I can see the case for it but this could also be simpler without it.

Secondarily, I find the disagreement in word ordering between `FirebasePlatformLoggingNoOp` and `CreateNoOpFirebasePlatformLogging` somewhat jarring. I suggest picking just one. Customarily the ordering for refinements of a base type seems to be refinement-base (as in the Create function) so I favor that ordering, but also don't feel strongly. See the differences searching google3 for `\w+ThreadPool\b lang:c++` vs `\bThreadPool\w+ lang:c++`: the former are refinements of thread pool while the latter are mostly either related types (like `ThreadPoolOptions`) or classes where `ThreadPool` is the refinement of some other base type (like `ThreadPoolTaskSystem`).

wilhuff (commented):
Structurally LGTM, though with several additional bits of optional feedback.

Firestore/core/src/remote/firebase_platform_logging_apple.mm
var-const: Good call, done.

Firestore/core/src/remote/firebase_platform_logging.h
var-const: I treat the feature name as an opaque identifier, so I would see it as similar to "LCD display/PIN number/etc.". However, I like the name `FirebaseMetadataProvider`, so I might as well just go with that.

I fully agree with partitioning metadata providers into Cloud/Firebase/etc.

var-const: > My original thought there was that separate concerns could supply separate metadata providers.

That's pretty much how I see it as well.

Firestore/core/src/remote/grpc_connection.h
var-const: Fixed, sorry.

Firestore/core/src/remote/firebase_platform_logging_noop.h
var-const: Made the word order consistent.

I guess I have a strong preference towards the `base-refinement` ordering because it creates natural grouping, though there is a good argument to be made regarding precedent. I'd certainly prefer not to tackle it in this PR.

I'm slightly in favor of having an explicit no-op class because it's not supposed to be used normally. I can move it under `test/` in a follow-up if you prefer.

Firestore/core/src/util/config.h
var-const: Sorry, originally I deleted the wrong `config.h` file. Thanks for explaining!

wilhuff (approved):
LGTM

var-const:
@VinayGuthal @vkryachko @maksymmalyhin Vinay, Vlad, Maksym -- can you please take a look to make sure the integration is correct? The most relevant parts are in the `firebase_metadata_provider*` files and in `grpc_connection.cc` (where the headers actually get sent).

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -0,0 +1,68 @@
...
+namespace {
+
+using util::MakeString;
+
+std::string GetUserAgent() {
+  return MakeString([FIRApp firebaseUserAgent]);
+}
+
+std::string GetHeartbeat() {
+  return std::to_string([FIRHeartbeatInfo heartbeatCodeForTag:@"fire-fst"]);
```
maksymmalyhin, line 40: FYI, currently `[FIRHeartbeatInfo heartbeatCodeForTag:]` method performs file operations to read and update data in the heartbeat storage. Ideally this call should not happen on the main thread. Also it may be worth to check if the current implementation provides performance good enough for Firestore use case. 

Please let me know if you have any questions or concerns.

Firestore/core/src/core/firestore_client.cc
```diff
@@ -107,13 +109,14 @@ static const size_t kMaxConcurrentLimboResolutions = 100;
...
-    std::shared_ptr<util::AsyncQueue> worker_queue) {
+    std::shared_ptr<CredentialsProvider> credentials_provider,
+    std::shared_ptr<Executor> user_executor,
+    std::shared_ptr<AsyncQueue> worker_queue,
+    std::unique_ptr<FirebaseMetadataProvider> firebase_metadata_provider) {
   // Have to use `new` because `make_shared` cannot access private constructor.
-  std::shared_ptr<FirestoreClient> shared_client(
-      new FirestoreClient(database_info, std::move(credentials_provider),
-                          std::move(user_executor), std::move(worker_queue)));
+  std::shared_ptr<FirestoreClient> shared_client(new FirestoreClient(
```
vkryachko, line 117: curious: any reason not to use `make_shared()` here?

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -0,0 +1,68 @@
...
+}  // namespace
+
+FirebaseMetadataProviderApple::FirebaseMetadataProviderApple(FIRApp* app)
+    : app_(app) {
+}
+
+void FirebaseMetadataProviderApple::UpdateMetadata(
+    grpc::ClientContext& context) {
+  context.AddMetadata(kXFirebaseClientHeader, GetUserAgent());
+  context.AddMetadata(kXFirebaseClientLogTypeHeader, GetHeartbeat());
```
vkryachko, line 56: I don't think the headers are sent unconditionally on android:
https://github.com/firebase/firebase-android-sdk/blob/6e286da45f952142d91482219f0ad1e019c2b06d/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java#L55

cc @VinayGuthal to confirm

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -0,0 +1,68 @@
...
+
+  std::string gmp_app_id = GetGmpAppId(app_);
+  if (!gmp_app_id.empty()) {
+    context.AddMetadata(kXFirebaseGmpIdHeader, gmp_app_id);
+  }
```
vkryachko, line 61: android does not send the gmp app id either. @VinayGuthal, curious why that is the case

maksymmalyhin (approved):
LGTM

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
wilhuff: Firestore runs all networking and and I/O operations on a dedicated queue. This can only ever be called from that queue so this won't have a chance to run on the main thread.

Firestore/core/src/core/firestore_client.cc
wilhuff: `make_shared` requires a public constructor but we're forcing callers to use `FirestoreClient::Create` and making the constructor private.

vkryachko: makes sense, thanks for explaining

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
var-const: Great point, thanks. Before I do the change, one thing I need to clarify is whether the GMP App ID should or shouldn't be sent if heartbeat is `0`. @VinayGuthal

maksymmalyhin: I think sending the user agent, heartbeat type and GMP App ID may be omitted when the heartbeat type is `FIRHeartbeatInfoCodeNone` if the headers are meant to be used for the Platform logging pipeline only as it won't affect instance count estimation. Though I see a potential value in sending the user agent anyway as it  may be used for analysis potential SDK issues.

VinayGuthal: hmm the concern at the time was that we might be causing a latency everytime firestore is used and hence the headers were not sent unconditionally. As @maksymmalyhin mentioned it won't affect instance count estimation.

var-const: Thanks for the explanations. Changed to not send anything if the heartbeat is `0` (how often does it happen, by the way?), PTAL.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
var-const: @VinayGuthal Do you know why Android doesn't send the GMP App ID? This is something we're currently planning to add.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -52,8 +52,13 @@
 
 void FirebaseMetadataProviderApple::UpdateMetadata(
     grpc::ClientContext& context) {
+  int heartbeat = GetHeartbeat();
+  if (heartbeat == 0) {
+    return;
```
maksymmalyhin, line 57: optional: the method name looks generic enough for other thing unrelated to platform logs to be added there. This early exist may be very easy to miss, so the additional metadata may be unintentionally dropped. I would either wrap platform logs metadata into `if(heartbeat != FIRHeartbeatInfoCodeNone)` or introduce a separate method with a name like `addPlatformLoggingMetadata()` with the early exist which will be called from `UpdateMetadata()`. WDYT?

maksymmalyhin (approved):

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -36,8 +36,8 @@
   return MakeString([FIRApp firebaseUserAgent]);
 }
 
-std::string GetHeartbeat() {
-  return std::to_string([FIRHeartbeatInfo heartbeatCodeForTag:@"fire-fst"]);
+int GetHeartbeat() {
```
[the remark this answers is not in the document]
var-const: That's a good point, done.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
var-const: This class is intended specifically for the Firebase platform logging-related functionality. We may abstract away other things (e.g. Cloud headers) in a similar class, but it would be a separate one.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
VinayGuthal: the heartbeat is 0 all the time except for the first time in the day.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
VinayGuthal: hmm looks like we do collect for firebase-iid. To estimate the number of devices using firebase we wouldn't need the GMP_APP_ID however. However the platform log proto has GMP_APP_ID as a optional string which can be useful for other metrics. I can change the android implementation to include the GMP_APP_ID as well.

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
maksymmalyhin: Thank you for clarification. IMO, the class and method names look a bit too generic in this case, but I don't feel too strong about it.

maksymmalyhin (approved):
LGTM

VinayGuthal (approved):

PR #6508, opened 2020-09-21
Add IWYU pragma to FIRDynamicLinks+FirstParty/FIRDynamicLinks

ahunt:
Most FirstParty users will still need to use methods from FIRDynamicLinks.h
(shouldHandleDynamicLinkFromCustomSchemeURL:/dynamicLinkFromCustomSchemeURL: etc.).
Arguably, including both headers looks odd, hence add the IWYU pragma so that it's sufficient
to include only the FirstParty header.

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinks+FirstParty.h
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinks+FirstParty.h
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#import "FirebaseDynamicLinks/Sources/Public/FirebaseDynamicLinks/FIRDynamicLinks.h"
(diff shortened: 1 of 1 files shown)


ahunt:
I'm not 100% sure this is the correct approach - it's possible to argue that it's better to include both headers - but most reviewers I've spoken to seem to think it's weird that I'm including both FIRDynamicLinks.h and FIRDynamicLinks+FirstParty.h when using Firebase in our app, suggesting that the pragma is indeed the best way forward.

paulb777 (approved):
Based on https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md, LGTM

Thanks

maksymmalyhin (approved):
LGTM

paulb777:
cc: @eldhosembabu

PR #6515, opened 2020-09-21
Update FIRStackFrame header to generate HTML correctly for Reference Docs

samedson:
This is because Jazzy generates HTML from the header file, and `<missing>` is not valid HTML

#no-changelog


morganchen12 (approved):
Thanks!

PR #6525, opened 2020-09-22
Import Spec tests

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/2000 (that's right - 2000!)

### DIFF
--- a/Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
+++ b/Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
@@ -245,10 +245,6 @@ NS_ASSUME_NONNULL_BEGIN
     _acknowledgedDocs = [NSMutableArray array];
 
     _rejectedDocs = [NSMutableArray array];
-
-    _waitForPendingWritesEvents = 0;
-
-    _snapshotsInSyncEvents = 0;
   }
   return self;
 }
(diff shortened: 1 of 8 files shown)


Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
```diff
@@ -195,6 +195,7 @@ @implementation FSTSyncEngineTestDriver {
   IndexFreeQueryEngine _queryEngine;
 
   int _snapshotsInSyncEvents;
+  int _waitForPendingWritesEvents;
```
var-const, line 198: Looks like this variable should be initialized to `0` in the `init` function, similar to `_snapshotsInSyncEvents`.

schmidt-sebastian: Done (but I didn't see the other initialization, so I added that as well)

var-const: Sorry, I mistakenly thought line 289 was part of the initialization, but it's actually in its own `reset` function. My Objective-C knowledge is poor -- it could be the case that primitives are set to `0` automatically, I'm not sure (it wouldn't be the case in C++). So this initialization could be redundant (though shouldn't hurt).

var-const (approved):

Firestore/Example/Tests/SpecTests/FSTSyncEngineTestDriver.mm
wilhuff: Objective-C defaults all members to zero, so explicitly making that initialization is redundant.

https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html

var-const: Thanks, Gil! Sebastian, sorry for the redundant comment.

schmidt-sebastian: Removed again.

PR #6526, opened 2020-09-22
Add Performance podspec

paulb777:
Enable more shared test and release processes between binary and source pods.

#no-changelog

### DIFF
--- a/FirebasePerformance.podspec.json
+++ b/FirebasePerformance.podspec.json
@@ -1,0 +1,42 @@
+{
+  "authors": "Google, Inc.",
+  "dependencies": {
+    "FirebaseCore": "~> 6.9",
+    "FirebaseInstallations": "~> 1.5",
+    "FirebaseRemoteConfig": "~> 4.7",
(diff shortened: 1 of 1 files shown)


visumickey (commented):
How are we planning to have this updated on a regular basis? Is this planned to be updated as a part of every release?

paulb777:
Yep, the proposed process is that your team updates the source field of the podspec by code freeze if you want to make an update and the iCore team manages the versioning and release from there.

visumickey:
> Yep, the proposed process is that your team updates the source field of the podspec by code freeze if you want to make an update and the iCore team manages the versioning and release from there.

Thanks Paul.

Just to clarify - Perf team, when we plan to make a release, we make changes to the podspec in the "FirebasePrivate/perf" branch and we also make the change to this public podspec? Want to make sure I update the steps for making a release from private Github.

visumickey (approved):

paulb777:
Confirmed.

After the next release, we'll stop managing releases via spreadsheet and the GitHub master branch becomes the mechanism for identifying what is part of a release.

PR #6541, opened 2020-09-23
SwiftPM for FCM

paulb777:
IID is also ported to SwiftPM but not exposed externally because of its upcoming deprecation.


chliangGoogle (approved):

davdroman:
Just to confirm, this will only be available in Firebase 7? I guess there's some technical limitation and can't be on version 6?

ryanwilson:
Confirmed that this will only be available in Firebase 7. The big technical issue here is Messaging currently depends on Protobuf which is not available via Swift Package Manager, but in Firebase 7 Messaging removed the dependency on Protobuf. There are potential workarounds in terms of hosting a binary version of Protobuf and depending on that, but there's technical complexity and many issues we've seen with binary frameworks so far that it's not worth the investment considering  we're able to add it easily as a source based product in Firebase 7.

davdroman:
@ryanwilson thank you for responding so quickly. It's good to know and I understand. Good to hear a few dependencies are being dropped in the next major version. Given the long list of issues left for the 7 milestone, I presume it'll take a while for it to come out. In the meantime, would you deem it sensible to consume `firebase7-main` from SPM if we depend on frameworks like Messaging?

paulb777:
@davdroman While we don't plan to break anything on `firebase7-main`, it is undergoing substantial churn with a bunch of technical debt being addressed and I wouldn't recommend usage in apps beyond development testing.

If despite those risks, you still want to use, we'll be happy to work with you on any issues you find.

davdroman:
@paulb777 okay, makes sense. Thank you Paul and Ryan for your efforts towards SPM support 👏

PR #6557, opened 2020-09-24
Stop requiring pods to be static frameworks

### LINKED ISSUE #2022: Remove "static_framework" from open-source pods
Some podspec files like FirebaseCore.podspec or FirebaseMessaging.podspec have set `s.static_framework = true`, although they are now open sourced and could be used as dynamic framework (use_frameworks!).
Cocoapods says:
`[!] The 'Pods-xxx target has transitive dependencies that include static frameworks: (FirebaseMessaging)`

Setting `static_framework` to false works as expected. Could `static_framework` be removed? 
Then each developer can decide if he wants to use dymanic frameworks or use the newer `use_modular_headers`-feature to build static libraries.

paulb777:
Fix #2022

Remove `static_framework` specifier from all podspecs.  Use the Podfile option `use_frameworks! :linkage => :static` to get 6.x Firebase linkage behavior.

- Lets developers control whether Firebase gets statically or dynamically linked.
- CocoaPods now has finer grain controls for static and dynamic linkage.
- Better supports some app modularization schemes.
- Firebase no longer has any dependencies on closed source static frameworks
- Firebase closed source pods are still static_frameworks

Additionally, this PR makes several changes to support the different bundle configuration of dynamic frameworks and associated tests:

- The only library change is an update to FIAM to find its Display resources in either location
- Additionally, several test changes were made including centralizing a mockFIROption class to support FirebaseApp.configure() for testing without a GoogleService-Info.plist.

I was also sad to delete the excellent doc's on static and dynamic frameworks from @maksymmalyhin, but I'm not sure they're still relevant.

I plan to do a follow-up PR to leverage CocoaPods 1.10 to add additional CI for `pod lib lint` testing with static frameworks.


FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
```diff
@@ -53,14 +53,17 @@ + (void)didReceiveConfigureSDKNotification:(NSNotification *)notification {
...
-    // TODO. This logic of finding the resource bundle may need to change once it's open
-    // sourced
-    NSBundle *containingBundle = [NSBundle mainBundle];
-    // This is assuming the display resource bundle is contained in the main bundle
-    NSURL *bundleURL = [containingBundle URLForResource:@"InAppMessagingDisplayResources"
-                                          withExtension:@"bundle"];
+    NSBundle *containingBundle;
+    NSURL *bundleURL;
+    // The containing bundle is different whether FIAM is statically or dynamically linked.
+    for (containingBundle in @[ [NSBundle mainBundle], [NSBundle bundleForClass:myClass] ]) {
```
maksymmalyhin, line 62: optional: wonder if we can reuse existing logic from Core?

docs/firebase_in_libraries.md
```diff
@@ -1,76 +0,0 @@
-# Using Firebase from a framework or a library
```
maksymmalyhin, line 1: I would prefer adding a note explaining why it is not relevant for Firebase 7 instead of deleting. Also, it seems still to be a relevant concern for zip, Carthage and binary distributions where we keep using static binaries.

maksymmalyhin (approved):
Great, mostly LGTM except a couple questions.

maksymmalyhin:
It's probably outside of the scope of this PR, but I think it may be useful to try to measure the difference of e.g. pre-main startup time between static and dynamic frameworks integration and add numbers to a README for developers to make more conscious decision on which to use in their project.

FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
paulb777: It doesn't seem worth it. Only the one line internal `relevantBundles` function would match as is.

paulb777 (commented):
Restored the docs with a Firebase 7 add. Thanks!

paulb777:
@maksymmalyhin @morganchen12 Good idea to do some measurements to help developers with the dynamic/static trade offs.  Based on the measurements we did a few years ago, there are definite start up costs to using dynamic linking, and we should reevaluate and quantify.

maksymmalyhin (approved):

PR #6558, opened 2020-09-24
Remove iOS 10 check now that Messaging minimum support is 10.0

chliangGoogle:
#no-changelog


chliangGoogle:
Close this one as I will create a new one from firebase7-main latest change.

Issue #6563, opened 2020-09-25
Fatal Exception: NSInvalidArgumentException


slccper:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.1
  * Firebase SDK version: 6.27.0
  * Firebase Component: Core, Performance, Crashlytics, RemoteConfig)
  * Component version: 6.27.0
  * Installation method: `CocoaPods` 

#### Steps to reproduce:

*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]
-[FIRInstallationsAPIService registerRequestWithInstallation:]

```

Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1924dafd8 __exceptionPreprocess
1  libobjc.A.dylib                0x190f3c538 objc_exception_throw
2  CoreFoundation                 0x1923c19a4 -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
3  CoreFoundation                 0x1923c1814 +[NSDictionary dictionaryWithObjects:forKeys:count:]
4  Newchic                        0x1002af368 -[FIRInstallationsAPIService registerRequestWithInstallation:] + 139 (FIRInstallationsAPIService.m:139)
5  Newchic                        0x1002aecd8 -[FIRInstallationsAPIService registerInstallation:] + 94 (FIRInstallationsAPIService.m:94)
6  Newchic                        0x1002b2bb8 -[FIRInstallationsIDController registerInstallationIfNeeded:] + 256 (FIRInstallationsIDController.m:256)
7  Newchic                        0x100387178 __56-[FBLPromise chainOnQueue:chainedFulfill:chainedReject:]_block_invoke.50 + 273 (FBLPromise.m:273)
8  libdispatch.dylib              0x1913929e0 _dispatch_call_block_and_release
9  libdispatch.dylib              0x1913929a0 _dispatch_client_callout
10 libdispatch.dylib              0x1913973c0 _dispatch_main_queue_callback_4CF
11 CoreFoundation                 0x1924890c0 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
12 CoreFoundation                 0x192486cdc __CFRunLoopRun
13 CoreFoundation                 0x1923b6d94 CFRunLoopRunSpecific
14 GraphicsServices               0x193e20074 GSEventRunModal
15 UIKit                          0x19866f130 UIApplicationMain
[...]

PR #6570, opened 2020-09-25
FIS: Additional FIRInstallationsItem validation

maksymmalyhin:
Related to #6563. 

- more detailed validation of `FIRInstallationsItem` read from storage to catch a potential data corruption (e.g. manual modifications of Keychain item on macOS, etc.)
- validation of `FIRInstallationsItem` before sending registration request.

### DIFF
--- a/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
+++ b/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
@@ -158,15 +158,13 @@
 }
 
 - (void)testGetInstallationItem_WhenNoFIDAndNoIID_ThenFIDIsCreatedAndRegistered {
-  // Stub store get installation.
-  [self expectInstallationsStoreGetInstallationCorruptedFID];
+  [self expectInstallationsStoreGetInstallationNotFound];
   [self expectStoredIIDNotFound];
 
   [self assertGetInstallation_NewFIDCreatedAndRegistered];
 }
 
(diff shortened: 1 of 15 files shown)


maksymmalyhin:
cc @andirayo

paulb777 (approved):

FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
```diff
@@ -1504,4 +1446,76 @@ - (void)expectBackoffEvent:(FIRInstallationsBackoffEvent)event
   OCMVerifyAll(self.mockBackoffController);
 }
 
+- (void)assertGetInstallation_NewFIDCreatedAndRegistered {
```
chliangGoogle, line 1449: nit:should method name be camel case without "_"?

chliangGoogle (approved):

FirebaseInstallations/Source/Tests/Unit/FIRInstallationsIDControllerTests.m
maksymmalyhin: I added "_" in some tests to the very long names to break the name to logical part to improve readability. I struggled to come up with anything better.

chliangGoogle: That's probably fine. Just want to call out the style guide. 
I'm okay if it leaves like this.

PR #6572, opened 2020-09-25
Stop using CocoaPods private headers

paulb777:
Stop using CocoaPods private headers to be more consistent with Swift Package Manager.

Instead internal API's between repo pods should be referenced with repo-relative imports.

FirebaseCore Private's are kept for now because of internal non-Firebase client usage via CocoaPods (Flutter for example)


ryanwilson (commented):
Discussed offline, but we should keep Core's private headers to avoid breaking any downstream dependencies like Flutter and RN depending on it.

ryanwilson (approved):

PR #6576, opened 2020-09-27
UI_USER_INTERFACE_IDIOM() was deprecated in iOS 14

daisuke-t-jp:
Hi,

I used Crashlytics with CocoaPods(Deployment Target iOS 13), will happen a warning.

UI_USER_INTERFACE_IDIOM

The `UI_USER_INTERFACE_IDIOM()` was deprecated. And I changed it.

---

Part of the used Podfile.

```Podfile
platform :ios, '13.0'
use_frameworks!

target 'App' do
    pod 'Firebase/Crashlytics'
end

post_install do | pi |
     pi.pods_project.targets.each do | t |
         t.build_configurations.each do | config |
             config.build_settings ['IPHONEOS_DEPLOYMENT_TARGET'] = '13.0'
         end
     end
end
```

Thanks.

### DIFF
--- a/Crashlytics/Shared/FIRCLSFABHost.m
+++ b/Crashlytics/Shared/FIRCLSFABHost.m
@@ -97,7 +97,7 @@ NSString *FIRCLSHostModelInfo(void) {
 #elif TARGET_OS_TV
   model = @"tvOS Simulator";
 #elif TARGET_OS_IPHONE
-  switch (UI_USER_INTERFACE_IDIOM()) {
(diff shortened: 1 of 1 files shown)


paulb777:
Thanks! 

Firebase doesn't always support changing the deployment target version without warning since we support building for older versions. However, this change looks safe since `userInterfaceIdiom` is available back to iOS 3.2.

samedson (approved):
🚢  Thank you for picking this up!

PR #6588, opened 2020-09-28
Make all GoogleUtilities APIs public

paulb777:
- Makes all GoogleUtilities API usage consistent
- Eliminates need for ifdef differences between SwiftPM and CocoaPods
- Enables transition of GoogleUtilities to another repo without a breaking change


GoogleUtilities/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 7.0.0
+- All APIs are now public. All CocoaPods private headers are tranitioned to public.
```
ryanwilson, line 2: transitioned

ryanwilson: Do we want to leave a comment mentioning that it's still not for non-Google usage?

ryanwilson (approved):
LGTM with the small spelling nit, just want to check to validate we're okay with locking in the API and having to make larger coordinated breaks now.

PR #6589, opened 2020-09-28
Make FirestoreException a type defined by Firestore/core

wilhuff:
Move FirestoreInternalError to firebase::firestore to match FirestoreException in Android.

When importing this into google3, we'll have to delete the Android-specific version of this type.

### DIFF
--- a/Firestore/core/src/util/firestore_exceptions.h
+++ b/Firestore/core/src/util/firestore_exceptions.h
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


Firestore/core/src/util/firestore_exceptions.h
```diff
@@ -0,0 +1,94 @@
...
+
+#ifndef FIRESTORE_CORE_SRC_UTIL_FIRESTORE_EXCEPTIONS_H_
+#define FIRESTORE_CORE_SRC_UTIL_FIRESTORE_EXCEPTIONS_H_
+
+// Common C++ exception declarations across iOS and Android. These aren't a
+// public API surface.
+//
+// See exception.h for how to throw exceptions in a platform-agnostic way.
+
+#include <stdexcept>
```
var-const, line 25: Seems like this should be `<exception>`, not `<stdexcept>`.

var-const (approved):

Firestore/core/src/util/firestore_exceptions.h
wilhuff: Done. `<stdexcept>` made sense when the base type was `std::logic_error`.

Issue #6590, opened 2020-09-29
Allow Notification images without proper path extensions


akivab:
### Environment

  * Xcode version: Version 12.0.1 (12A7300)
  * Firebase SDK version: 6.33.0
  * Firebase Component: Messaging
  * Component version: 6.33.0
  * Installation method: `CocoaPods`

### Problem

Firebase Messaging does not seem to allow for image URLs without an extension that matches MIME type, despite the fact that we may be able to produce the right extension from the MIME type.

In the function
```
- (void)loadAttachmentForURL:(NSURL *)attachmentURL
           completionHandler:(void (^)(UNNotificationAttachment *))completionHandler {
```

We create the extension for the local download using 
```
        NSString *suggestedPathExtension = [response.suggestedFilename pathExtension];
```

Which raises issues later in the function, i.e.
```
            FIRMessagingLoggerError(
                                    kFIRMessagingServiceExtensionLocalFileNotCreated,
                                    @"Failed to move the image file to local location: %@, error: %@\n", localURL,
                                    error);

```

#### Steps to reproduce:

What happened? How can we make the problem occur?

Create a notification with a URL that points to an image resource without the proper extension (e.g. a UUID)

#### Relevant Code:

Fixed the issue by changing the logic to

```
        NSString *suggestedPathExtension = [response.suggestedFilename pathExtension];
        NSString *mimeTypePathExtension = [response.MIMEType stringByReplacingOccurrencesOfString:@"image/" withString:@""];
        NSString *pathExtension = ((suggestedPathExtension.length > 0) ? suggestedPathExtension : mimeTypePathExtension);
        NSString *fileExtension = [NSString stringWithFormat:@".%@", pathExtension];
```

PR #6591, opened 2020-09-29
Use MIMEType for image download extension if not present in resource URL

akivab:
Recommend adding so that we can be creative about extension (without forcing the user to explicitly set in the URL provided, as it might be provided in storage metadata)

Addresses the following issue
https://github.com/firebase/firebase-ios-sdk/issues/6590


akivab:
@chliangGoogle I could move some of this logic to another function to add tests, let me know what your preference would be. Thanks!

paulb777:
@akivab Thanks for the issue and PR! Routing to @chliangGoogle for review.

akivab:
I have tried adding some tests but having trouble with automatic signing issues for the AppHost-FirebaseMessaging-Unit-Tests target (`ld: file not found: /Users/akivab/Library/Developer/Xcode/DerivedData/FirebaseMessaging-hkzirhinkydyvnexsqhhikscqmru/Build/Products/Debug-maccatalyst/AppHost-FirebaseMessaging-Unit-Tests.app/Contents/MacOS/AppHost-FirebaseMessaging-Unit-Tests`)

Attaching the diff so that it might serve as an example of the tests I'm planning to add (but that I didn't want to add due to the concern of being unable to run locally)
[changes_with_tests.txt](https://github.com/firebase/firebase-ios-sdk/files/5296104/changes_with_tests.txt)

paulb777:
It might be easier to test on iOS than Catalyst. You can generate an iOS dev environment with:

`pod gen FirebaseMessaging.podspec --auto-open --local-sources=./ --platforms=ios`

akivab:
Thanks so much @paulb777 -- I've updated my PR with some simple unit tests. @chliangGoogle open to any feedback.

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -62,6 +62,21 @@ - (void)populateNotificationContent:(UNMutableNotificationContent *)content
 }
 
 #if TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_WATCH
+- (NSString *)fileExtensionForResponse:(NSURLResponse *)response {
+  NSString *const kDefaultExtension = @".tmp";
```
akivab, line 66: I chose ".tmp" here because I wanted it to fail if there was no storage metadata included and no extension in the filename, but on reflection may make sense to default to ".png" which is allows us to at the very least make a simple assumption on the user's behalf rather than simply fail. Thoughts?

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -62,6 +62,21 @@ - (void)populateNotificationContent:(UNMutableNotificationContent *)content
 }
 
 #if TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_WATCH
+- (NSString *)fileExtensionForResponse:(NSURLResponse *)response {
+  NSString *const kDefaultExtension = @".png";
+  NSString *const kImagePathPrefix = @"image/";
+  NSString *suggestedPathExtension = [response.suggestedFilename pathExtension];
+  if (suggestedPathExtension != nil && suggestedPathExtension.length > 0) {
```
chliangGoogle, line 69: maybe just 
if (suggestedPathExtension.length > 0) {

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -62,6 +62,21 @@ - (void)populateNotificationContent:(UNMutableNotificationContent *)content
 }
 
 #if TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_WATCH
+- (NSString *)fileExtensionForResponse:(NSURLResponse *)response {
+  NSString *const kDefaultExtension = @".png";
```
chliangGoogle, line 66: Can you move both static string on top of this file following the other static parameters?

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -62,6 +62,21 @@ - (void)populateNotificationContent:(UNMutableNotificationContent *)content
...
+  NSString *const kImagePathPrefix = @"image/";
+  NSString *suggestedPathExtension = [response.suggestedFilename pathExtension];
+  if (suggestedPathExtension != nil && suggestedPathExtension.length > 0) {
+    return [NSString stringWithFormat:@".%@", suggestedPathExtension];
+  }
+  if (response.MIMEType != nil && [response.MIMEType containsString:kImagePathPrefix]) {
+    NSString *mimeTypePathExtension =
+        [response.MIMEType stringByReplacingOccurrencesOfString:kImagePathPrefix withString:@""];
+    return [NSString stringWithFormat:@".%@", mimeTypePathExtension];
```
chliangGoogle, line 75: maybe just replace with
return [response.MIMEType stringByReplacingOccurrencesOfString:kImagePathPrefix withString:@"."]

chliangGoogle (changes requested):
Awesome. Thank you for doing this!

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -62,6 +62,21 @@ - (void)populateNotificationContent:(UNMutableNotificationContent *)content
...
+  }
+  return kDefaultExtension;
```
chliangGoogle, line 77: I don't know image converting enough to know that if it's okay to return a default format like "png" if it is not a png image?
maybe if neither format is valid, we should return empty string and let it fail with the error before?
I don't know if it's also secure to download file that might not be image at all. wdyt?

akivab: I figured we'd want to set to ".png" in case the metadata wasn't stored, but agreed we can let it error instead.

akivab:
Thanks @chliangGoogle, addressed comments.

FirebaseMessaging/Tests/UnitTests/FIRMessagingExtensionHelperTest.m
```diff
@@ -48,6 +50,7 @@ - (void)setUp {
   if (@available(macOS 10.14, iOS 10.0, *)) {
     FIRMessagingExtensionHelper *extensionHelper = [FIRMessaging extensionHelper];
     _mockExtensionHelper = OCMPartialMock(extensionHelper);
+    _mockURLResponse = OCMClassMock([NSURLResponse class]);
```
chliangGoogle, line 53: make sure to call stopMocking in tearDown

chliangGoogle (changes requested):
one last nit bit.

Can you also update the changelog in FirebaseMessaging/CHANGELOG?
#unreleased 
[fixed] Fixed the issue that failed to download image if no extension but MIME type is set. (#6590)
- [fixed] Fix issue :3

FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [fixed] Fixed the issue that failed to download image if no extension but MIME type is set. (#6590)
+
```
chliangGoogle, line 3: Fixed an issue that downloading an image failed when there's no extension in the file name but MIME type is set.

akivab:
Thanks @chliangGoogle  -- added to the CHANGELOG.md and calling teardown on the mock now

chliangGoogle (commented):
Sorry, my wording is bad, I corrected mine again. 

Thank you very much doing this. I will approve this when it's okay to merge master. Our team is doing maintenance work at the moment. Once it's cleared, I will approve and you can merge to master.

FirebaseMessaging/CHANGELOG.md
akivab: Sounds good, updated language.

akivab:
Perfect, sounds good. Contributing here has been a great experience, hoping to do more in the future.

chliangGoogle (approved):

akivab:
Hi @chliangGoogle, I don't have write access to this repo so expecting that you might be able to merge in. Thanks so much.

paulb777:
I'll merge it now - to make sure that none of the upcoming major changes introduce a conflict.

Thanks for the PR @akivab !

PR #6594, opened 2020-09-29
Custom model

manjanac:
Hello! I created the API scaffolding for CustomModel and ModelDownloader classes, along with example usage.


FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -0,0 +1,44 @@
...
+  case CoreML
+}
+
+public struct CustomModel {
+
+  public let modelName: String
+  public internal(set) var modelSize: Int?
+  public internal(set) var modelPath: String?
+  public internal(set) var modelHash: String?
+  public internal(set) var modelFormat: CustomModelFormat?
```
maksymmalyhin, line 30: I curious if a model with `modelPath == nil` and `modelSize != nil` is a valid one? To reduce amount of inconsistent states we may want to group properties together into a separate struct e.g.:
```
struct CustomModel {
  struct Metadata {
    var modelSize: Int
    var modelPath: String
    ...
  }

  let name: String
  var metadata: Metadata?
}
```

maksymmalyhin: My main point is that if we define the model so that it cannot be in an invalid state then compiler will catch some dummy issues and unhandled cases for us. It's relevant for both the internal implementation and the API clients.

So if the struct may represent either a loaded or not-loaded-yet model, then a more expressive API could be something like

```
struct CustomModel {
  struct Metadata {
    var size: Int
    ...
  }
  enum State {
    case initial
    case downloaded(metadata: Metadata)
    ...
  }

  let name: String
  var state: State
}
```

From the other side, the more Swift specific features we use, the more it will be different from the Objective-C API. For Objective-C we will have to declare `NSObject` subclasses instead of structs and will have to transfer the enum with associated type to Obj-C types.

Sorry if I'm missing any ML SDK specifics, I'm just trying to take the most from Swift power since we have access to it 😄

manjanac: We want CustomModel (and ModelDownloader) to be visible from objc, so is it better to have them be subclasses of NSObject in Swift so that we can use the @objc decorator?

maksymmalyhin: Alternatively an additional Objective-C specific model/wrapper may be used, so we can fully laverage Swift advantages and provide a nice Swift API. I would prefer to provide nice Swift API as majority of our customers use Firebase from Swift.

@paulb777 @ryanwilson if you have an opinion on it?

paulb777: I agree with the sentiment to optimize for a nice Swift API while providing a sufficient Objective C API.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -0,0 +1,44 @@
...
+
+  init(withName name: String) {
+    modelName = name
+  }
+
+  public func getLatestModel() -> FileHandle? {
```
ryanwilson, line 36: What's the benefit of this function existing vs a developer just checking for themselves:

```
let customModel = someFuncToGetCustomModel()
if let modelPath = customModel.modelPath {
  // Use FileManager to open modelPath...
}
```

FirebaseMLModelDownloader/Sources/Examples.swift
```diff
@@ -0,0 +1,60 @@
...
+        // Pick model(s) for further use
+      }
+    }
+
+    // Delete downloaded model
+    modelDownloader.deleteDownloadedModel(name: "your_model_name") { error in
+      if error == nil {
+        // Apply any other clean up
+      }
+    }
```
ryanwilson, line 58: This seems like a great start to stubs for unit tests instead of landing it in the `Sources` directory!

FirebaseMLModelDownloader/Sources/CustomModel.swift
ryanwilson: It looks like the `CustomModel` type isn't used in any of the APIs requesting things (I believe it was used before) so likely the modelName can be readonly as well and all the other properties can be non-optional?

manjanac: I decided to go with having the model object always filled out. So there are no optional properties anymore.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+public enum CustomModelFormat {
+  case Unknown
+  case TFLite
+  case TorchScript
+  case CoreML
```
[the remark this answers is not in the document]
manjanac: Good catch, done.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -0,0 +1,44 @@
...
+
+public struct CustomModel {
+
+  public let modelName: String
+  public internal(set) var modelSize: Int?
+  public internal(set) var modelPath: String?
+  public internal(set) var modelHash: String?
+  public internal(set) var modelFormat: CustomModelFormat?
+
+  init(withName name: String) {
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/Examples.swift
```diff
@@ -0,0 +1,59 @@
...
```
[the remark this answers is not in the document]
manjanac: Sounds good, I moved this under Tests.

FirebaseMLModelDownloader/Sources/CustomModel.swift
manjanac: I think the reasoning here is that the dev would probably have no use for modelPath itself, and it is easier to just provide them with the modelFile as needed. I've made the modelPath property internal as well. Would that work?

FirebaseMLModelDownloader/Sources/Examples.swift
manjanac: Done!

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -0,0 +1,49 @@
...
+/// An enum of supported model formats.
+public enum CustomModelFormat {
+  case unknown
+  case tfLite
+  case coreML
+}
+
+/// A custom model that is stored remotely on the server and downloaded to the device.
+public struct CustomModel {
+  public let modelName: String
```
maksymmalyhin, line 26: nit: If prefix `model` to the property names looks redundant if the intended usage is like the following:
```
let model = CustomModel(...)
print("Name:\(model.modelName)")
```
I would use just `name`, `size`, etc. WDYT?

manjanac: We had an internal discussion about this and the reasoning was that in the future there may be a "name" field that may not refer to the modelName. Just future proofing here. :)

manjanac: Update - turns out we can remove the `model` prefix so I updated the struct.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -14,36 +14,10 @@
 
 import Foundation
 
-/// An enum of supported model formats.
-public enum CustomModelFormat {
```
manjanac: We are not supporting different model formats anymore.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -14,36 +14,10 @@
...
-
-  init(name: String, size: Int, path: String, hash: String, format: CustomModelFormat) {
-    modelName = name
-    modelSize = size
-    modelPath = path
-    modelHash = hash
-    modelFormat = format
-  }
-
-  public func getLatestModel() -> FileHandle? {
```
manjanac: We only need modelPath not modelFile according to the TFLite docs, so taking this out.

ryanwilson (approved):
This LGTM from a scaffolding standpoint - two small nits and something to follow up on for the future documented here but after those two are modified I'm happy if this lands and we continue on with furthering the implementation.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -0,0 +1,88 @@
...
+  case resourceExhausted // server limit reached
+  case failedPrecondition // conditions not met to perform download
+  case notEnoughSpace // not enough space on device
+  case unauthenticated // request does not have valid credentials
+  case unimplemented // unsupported operation
+  case invalidArgument // malformed model name or path
+  case canceled // operation was canceled
+  case aborted // operation was aborted, ex: connection dropped
+  case dataLoss // unrecoverable data loss or corruption
+  case outOfRange // operation past valid input/output range
```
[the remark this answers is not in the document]
manjanac: Makes sense. I made a change to errors to better reflect this. This may undergo further refinement in a future PR.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -12,14 +12,12 @@
...
-        // Use XCTAssert and related functions to verify your tests produce the correct
-        // results.
-        XCTAssertEqual(MLModelDownloader().text, "Hello, World!")
-    }
+/// A custom model that is stored remotely on the server and downloaded to the device.
+public struct CustomModel: Hashable {
+  public let name: String
+  public var size: Int
+  public var path: String
+  public var hash: String
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/CustomModel.swift
```diff
@@ -12,14 +12,12 @@
...
-final class MLModelDownloaderTests: XCTestCase {
-    func testExample() {
-        // This is an example of a functional test case.
-        // Use XCTAssert and related functions to verify your tests produce the correct
-        // results.
-        XCTAssertEqual(MLModelDownloader().text, "Hello, World!")
-    }
+/// A custom model that is stored remotely on the server and downloaded to the device.
+public struct CustomModel: Hashable {
+  public let name: String
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -0,0 +1,90 @@
...
+  case notFound
+  /// Caller does not have necessary permissions for this operation.
+  case permissionDenied
+  /// Conditions not met to perform download.
+  case failedPrecondition
+  /// Not enough space for model on device.
+  case notEnoughSpace
+  /// Malformed model name.
+  case invalidArgument
+  /// Other network errors with description.
```
ryanwilson, line 29: Nit: we may want to include non-network errors here, so the `network` in here should likely be removed.

ryanwilson (approved):
LGTM for structure, let's merge whenever you're ready :)

PR #6599, opened 2020-09-29
Add static-framework testing to CI

paulb777:
Now that static framework's are off by default, make sure we still test when it is turned back on.
Remove `--use-modular-headers` testing since it is mostly redundant with `--use-static-frameworks` and no longer a recommended option.

All tests passed before turning cron back on

Note that the `--use-static-framework` pod lib lint option is introduced in CocoaPods 1.10

#no-changelog

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,5 +3,7 @@
 source 'https://rubygems.org'
 
 gem 'cocoapods', '1.10.0.rc.1'
-gem 'cocoapods-generate', git: "https://github.com/square/cocoapods-generate.git", branch: "master"
+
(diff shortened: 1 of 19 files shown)


Gemfile
```diff
@@ -2,6 +2,6 @@
 # commit Gemfile and Gemfile.lock.
 source 'https://rubygems.org'
 
-gem 'cocoapods', "1.9.3"
-gem 'cocoapods-generate', '2.0.0'
+gem 'cocoapods', '1.10.0.rc.1'
+gem 'cocoapods-generate', git: "https://github.com/square/cocoapods-generate.git", branch: "master"
```
ryanwilson, line 6: Can we use a commit instead of pointing at master here?

ryanwilson (approved):

PR #6601, opened 2020-09-29
Update CHANGELOG for Firestore v1.19.0

wilhuff:


wilhuff:
The breaking change for C++ is the movement of the exception types in #6589.

var-const (approved):

Issue #2505, opened 2019-03-09
Migrate Logger Service IDs from FirebaseCore to clients


paulb777:
FirebaseCore should not have knowledge about its clients. The FIRLoggerServices declared in FIRLogger.h and defined in FIRLogger.m should be moved to the clients.

Since this is an internal API change, it does not necessarily need to wait for a major release, but I'm marking it for Firebase 6 as a target.

An incremental approach to this issue would be starting with the open source code and creating an internal issue for each of the closed source clients.

paulb777:
Might just do easy clients that don't have lots of FIRLogger.h imports

paulb777:
Partially done for Firebase 6 in #2720.

I'll remove the Firebase 6 milestone from this issue and leave open for the rest.

paulb777:
#6608 removes the strings for AdMob, ABT, and Auth for 7.0.0

Analytics, Crash, MLKit, Perf, and RemoteConfig are deferred beyond Firebase 7 because the complexity of cleaning up in google3 doesn't currently justify the value - and hopefully that will be simplified in the next few months.

paulb777:
MLKit symbol is tracked internally at b/146429345 and should be removed in Firebase 8.0.0 release

paulb777:
#7908 removes Perf and MLKit for 8.0.0. Analytics, Crash and RemoteConfig are deferred due to google3 references.

paulb777:
Analytics, Crash and RemoteConfig are deferred again due to google3 references.

paulb777:
No plans for Firebase 10. On to Firebase 11

paulb777:
This is no longer a priority, since with one-versioning and SPM, it is no longer a requirement for FirebaseCore to avoid knowing about its clients.

These should probably go away when we move to new logging anyway.

PR #6608, opened 2020-09-29
Migrate FIRLoggerServices strings from Core to clients

paulb777:
Remove knowledge from FirebaseCore of its clients for FIRLoggerService strings by migrating string definitions to ABT and Auth.. The unused AdMob string is also deleted.

The other five are left for now because of usages in other source control systems.

Partial fix of #2505


maksymmalyhin (approved):

paulb777:
Merging now. I'll address any additional comments in a subsequent PR.

PR #6619, opened 2020-09-30
FIS: disable IID migration tests by default

### LINKED ISSUE #6600: Remove old IID version dependency from FIS tests
Since it won't work after a major version update.

maksymmalyhin:
Fix #6600

The tests require a specific IID version that may have dependencies conflicting with Firebase 7. The code being tested by the tests is not going to change as IID SDK has been deprecated, so maintaining the tests is not worth the cost. The tests may be enabled if needed in the future. 

#no-changelog


paulb777 (approved):
LGTM on green

Issue #6547, opened 2020-09-23
Firestore: Remove deprecated APIs


paulb777:
Please remove deprecations in public headers in Firebase 7

```
~/gh4/firebase-ios-sdk/Firestore/Source/Public/FirebaseFirestore (master) $ git grep -B5 -i deprecat
FIRFirestoreSettings.h: * @deprecated This setting now defaults to true and will be removed in a future release. If you are
FIRFirestoreSettings.h- * already setting it to true, just remove the setting. If you are setting it to false, you should
FIRFirestoreSettings.h- * update your code to expect FIRTimestamp objects instead of NSDate and then remove the setting.
FIRFirestoreSettings.h- */
FIRFirestoreSettings.h-@property(nonatomic, getter=areTimestampsInSnapshotsEnabled) BOOL timestampsInSnapshotsEnabled
FIRFirestoreSettings.h:    __attribute__((deprecated));
```

PR #6622, opened 2020-09-30
Remove support for setting timestampsInSnapshotsEnabled

wilhuff:
Snapshots will now always include `FIRTimestamp` values. Use `[FIRTimestamp dateValue]` to convert to `NSDate` if required.

`timestampsInSnapshotsEnabled` was deprecated in Firebase 5.16.0 (January 22, 2019) and has defaulted to true since then.

Projects that have not enabled `timestampsInSnapshotsEnabled` have logged a warning since Firebase 4.12.0 (April 10, 2018).

/cc @var-const Note that this is also a breaking change for the C++ SDK, but only because it currently forces timestamps in snapshots to true. We could remove that in advance of importing M82 without a problem (since `timestampsInSnapshotsEnabled` already defaults to true). Existing users will be unaffected because the dependency on this pod only floats on patch levels.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 # v7.0.0
 - [changed] **Breaking change:** Removed the `areTimestampsInSnapshotsEnabled`
-  setting. Timestamp fields that read from a `FIRDocumentSnapshot` are now
-  always returns `FIRTimestamp` objects. Use `FIRTimestamp.dateValue` to
-  convert to `NSDate` if required.
+  setting. Timestamp fields that read from a `FIRDocumentSnapshot` now always
+  return `FIRTimestamp` objects. Use `FIRTimestamp.dateValue` to convert to
+  `NSDate` if required.
 
 # v1.19.0
 - [changed] Internal improvements for future C++ and Unity support. Includes a
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+# v7.0.0
+- [changed] **Breaking change:** Removed the `areTimestampsInSnapshotsEnabled`
+  setting. Timestamp fields that read from a `FIRDocumentSnapshot` are now
+  always returns `FIRTimestamp` objects. Use `FIRTimestamp.dateValue` to
+  convert to `NSDate` if required.
+
 # v1.19.0
 - [changed] Internal improvements for future C++ and Unity support. Includes a
   breaking change for the Firestore C++ Alpha SDK, but does not affect
--- a/Firestore/Example/Tests/Integration/API/FIRFieldsTests.mm
+++ b/Firestore/Example/Tests/Integration/API/FIRFieldsTests.mm
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google
+ * Copyright 2017 Google LLC
  *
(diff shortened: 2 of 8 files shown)


Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Unreleased
 
+# v7.0.0
+- [changed] **Breaking change:** Removed the `areTimestampsInSnapshotsEnabled`
+  setting. Timestamp fields that read from a `FIRDocumentSnapshot` are now
```
paulb777, line 5: s/are//

Firestore/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+  always returns `FIRTimestamp` objects. Use `FIRTimestamp.dateValue` to
```
paulb777, line 6: s/returns/return/

Firestore/core/src/model/field_value_options.h
```diff
@@ -31,23 +31,16 @@ class FieldValueOptions {
    * Creates an FieldValueOptions instance that specifies deserialization
    * behavior for pending server timestamps.
    */
-  FieldValueOptions(ServerTimestampBehavior server_timestamp_behavior,
-                    bool timestamps_in_snapshots_enabled)
-      : server_timestamp_behavior_(server_timestamp_behavior),
-        timestamps_in_snapshots_enabled_(timestamps_in_snapshots_enabled) {
+  explicit FieldValueOptions(ServerTimestampBehavior server_timestamp_behavior)
```
schmidt-sebastian, line 34: What do you think about removing this class? I can see arguments going either way, so I have no actual opinion here.

Firestore/Example/Tests/Integration/API/FIRFieldsTests.mm
```diff
@@ -255,52 +255,7 @@ - (void)testTimestampsInSnapshots {
   XCTAssertEqualObjects(truncatedTimestamp, timestampFromData);
   XCTAssertEqualObjects(timestampFromSnapshot, timestampFromData);
 }
-@end
```
schmidt-sebastian, line 258: In the Android PR that I started working on, I renamed the previous test to `testTimestampsAreTruncated`, mostly just get rid of the 'timestampsInSnapshots" term from the codebase. Let me know what you end up deciding since the test names should end up matching.

schmidt-sebastian (approved):

Firestore/CHANGELOG.md
wilhuff: Done.

Firestore/CHANGELOG.md
wilhuff: Done.

Firestore/core/src/model/field_value_options.h
wilhuff: I'm disinclined toward removing the class. Admittedly it doesn't serve much purpose, but we've repeatedly mooted additional snapshot options and those would potentially feed into here.

Firestore/Example/Tests/Integration/API/FIRFieldsTests.mm
wilhuff: I renamed to match your Android change.

paulb777:
@wilhuff Does this address #6547

wilhuff:
Yes. I've closed that issue.

PR #6624, opened 2020-09-30
Add auth emulator support to public API.

rosalyntan:
Add auth emulator support `useEmulatorWithHost:port` to public API.

Implementation added in #6402


ryanwilson (approved):
LGTM - I assume the Facebook test failures are API limits, I don't think it should be a blocker for submitting (or you can re-run them until they work)

paulb777 (approved):
It would be nice if we could find a way to run the Facebook integration tests without hitting the API limit so often.

PR #6633, opened 2020-10-01
Remove the `FirebaseOptions()` bare initializer.

### LINKED ISSUE #5728: Mark FirebaseOptions empty initializer as NS_UNAVAILABLE
When creating a [`FirebaseOptions`](https://firebase.google.com/docs/reference/swift/firebasecore/api/reference/Classes/FirebaseOptions) instance, there should **not** be the option of using an initializer with no arguments since it's not a valid `FirebaseOptions` instance (from a usage standpoint).

```swift 
let options = FirebaseOptions() // currently allowed but shouldn't be
```

**Proposed change for [`FIROptions.h`](https://github.com/firebase/firebase-ios-sdk/blob/1367084063a7fd10eeeac611ebde8cc7f5bda15a/FirebaseCore/Sources/Public/FIROptions.h)**:
• Mark initializer with no arguments as `NS_UNAVAILABLE`
```objc
// FIROptions.h

- (instancetype)init NS_UNAVAILABLE;
```

This change helps ensure instances of `FirebaseOptions` are created using one of the designated initializers.

ryanwilson:
This was unintentionally surfaced from the ObjC header and should be
removed.

Note: although the existing initializer doesn't work, this is still a
breaking change.

Fixes #5728 

Following recommendations from https://developer.apple.com/videos/play/wwdc2020/10680

### DIFF
--- a/FirebaseCore/Sources/FIROptions.m
+++ b/FirebaseCore/Sources/FIROptions.m
@@ -160,9 +160,8 @@ static dispatch_once_t sDefaultOptionsDictionaryOnceToken;
 }
 
 - (id)copyWithZone:(NSZone *)zone {
-  FIROptions *newOptions = [[[self class] allocWithZone:zone] init];
+  FIROptions *newOptions = [(FIROptions *)[[self class] allocWithZone:zone] initInternalWithOptionsDictionary:self.optionsDictionary];
   if (newOptions) {
-    newOptions.optionsDictionary = self.optionsDictionary;
     newOptions.deepLinkURLScheme = self.deepLinkURLScheme;
     newOptions.appGroupID = self.appGroupID;
     newOptions.editingLocked = self.isEditingLocked;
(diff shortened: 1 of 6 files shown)


FirebaseCore/Sources/FIROptions.m
```diff
@@ -173,6 +173,12 @@ - (id)copyWithZone:(NSZone *)zone {
 
 #pragma mark - Public instance methods
 
+- (instancetype)init {
+  // Unavailable.
+  [self doesNotRecognizeSelector:_cmd];
```
maksymmalyhin, line 178: I wonder if we really need this? The initializer marked as unavailable should lead to a compiler error, so I guess it should be enough. Or there is a specific case we would like to cover?

maksymmalyhin (approved):
Mostly LGTM, one minor optional question.

FirebaseCore/Sources/FIROptions.m
ryanwilson: Ah, forgot to link! I was watching the "Refine Objective-C frameworks for Swift" WWDC session and that was the suggestion: https://developer.apple.com/videos/play/wwdc2020/10680/?time=1159

They suggested this to make sure there's no accidental calls there. It also surfaced the [`copyWithZone:` implementation](https://github.com/firebase/firebase-ios-sdk/pull/6633/files#diff-fb3d49a63a8c4afacada87c02bc8de7bR163) calling the `init` function when it should have likely called one of the designated initializers.

PR #6429, opened 2020-09-09
Firebase user agent: additional fields for platform logging

maksymmalyhin:
Context: b/168108419
LaunchCal: ariane/4027197

PR #6634, opened 2020-10-01
One Firebase version

### LINKED ISSUE #6295: One version for all Firebase pods
With the next major version, we plan to re-version all Firebase pods to 7.0.0 and then keep them in sync from then on.

This will align CocoaPods versioning with the Swift Package Manager, Carthage and zip distributions.

The implementation should replace all individual pod version handling and option processing with a call to the FirebaseCore pod to get the Firebase version.

Please add any comments or concerns to this thread.

paulb777:
- All Firebase pods now share a single version.
- GoogleUtilities and GoogleDataTransport continue to be versioned independently for CocoaPods.
- Instead of a leading 0 in the version, betas are now indicated by a `-beta` suffix in CocoaPods and `Beta` appended to the product name in SwiftPM
- A new [FIRVersion version] internal API is provide by Core for the rest of Firebase to access the version
- The public version APIs of Storage, Auth and Installations are removed
- I'll do a follow up discussion about whether FIRVersion should be public or not.
- The quickstart tests for Database, Firestore, and Functions are disabled pending a compatible version of FirebaseUI for Firebase 7.x (#6646).
- Delete unused version processing code.

Fix #6295


FirebaseCore/Sources/FIRVersion.m
```diff
@@ -14,20 +14,29 @@
...
+@implementation FIRVersion
+
++ (NSString *)version {
+  const char *const FIRVersionString = (const char *const)STR(Firebase_VERSION);
+  static NSString *v;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    v = [NSString stringWithUTF8String:FIRVersionString];
+  });
+  return v;
```
ryanwilson, line 39: Do you think this is necessary vs instantiating the string every time? I can't imagine this is expensive to instantiate each time in the grand scheme of things.

maksymmalyhin (approved):
It may potentially conflict with #6429, but it's fine.

Package.swift
```diff
@@ -22,7 +22,7 @@
 
 import PackageDescription
 
-let firebaseVersion = "6.34.0"
+let firebaseVersion = "7.0.0"
```
ryanwilson, line 25: I can't comment on it since it's existing code but `Package.swift` also needs to remove the `FIRCore_VERSION` and the rest.

FirebaseAuth/Tests/Sample/Sample/SettingsViewController.m
```diff
@@ -29,6 +29,7 @@
 #import <FirebaseAuth/FirebaseAuth.h>
 #import "StaticContentTableViewManager.h"
 #import "UIViewController+Alerts.h"
+#import "../../../../FirebaseCore/Sources/Private/FIRVersion.h"
```
ryanwilson, line 32: Can this go back to non-relative paths now that Core is keeping it's private headers, or did you want to keep it like this?

FirebaseCore/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   option `use_frameworks! :linkage => :static` to get the Firebase 6.x linkage behavior. (#2022)
 - [removed] Removed broken `FirebaseOptions()` initializer. Use `init(contentsOfFile:)` or
   `init(googleAppID:gcmSenderID:)` instead. (#6633)
+- [changed] All Firebase pods now have the same version.
+- [changed] In CocoaPods, Firebase betas are now indicated in the version tag. In SwiftPM, beta
+  is appended to the product name.
+- [changed] The version must now be specified for the two Swift-only Firebase CocoaPods in the
+  Podfile like `pod 'FirebaseFirestoreSwift', '~> 7.0-beta'`.
```
ryanwilson, line 14: This should this include the patch version, right?

FirebaseInstallations.podspec
```diff
@@ -36,12 +36,12 @@ Pod::Spec.new do |s|
...
-  s.dependency 'FirebaseCore', '~> 6.10'
+  s.dependency 'FirebaseCore', '~> 7.0'
   s.dependency 'PromisesObjC', '~> 1.2'
-  s.dependency 'GoogleUtilities/Environment', '~> 6.7'
-  s.dependency 'GoogleUtilities/UserDefaults', '~> 6.7'
+  s.dependency 'GoogleUtilities/Environment', '~> 7.0'
+  s.dependency 'GoogleUtilities/UserDefaults', '~> 7.0'
 
-  preprocessor_definitions = 'FIRInstallations_LIB_VERSION=' + String(s.version)
+  preprocessor_definitions = ''
```
ryanwilson, line 44: Is this second `'` needed?

FirebaseStorageSwift.podspec
```diff
@@ -14,7 +14,7 @@ Firebase Storage provides robust, secure file uploads and downloads from Firebas
 
   s.source                  = {
     :git => 'https://github.com/Firebase/firebase-ios-sdk.git',
-    :tag => 'StorageSwift-' + s.version.to_s
+    :tag => 'CocoaPods-' + s.version.to_s
```
ryanwilson, line 17: Just realized for here and elsewhere - we'll need to make sure we also create the `CocoaPods-7.Y.Z-beta` tag each release too

FirebaseCore/CHANGELOG.md
paulb777: Not necessarily. It's standard semantic versioning. '~> 7.0-beta'` allows minor version updates. '~> 7.0.0-beta'` will allow only patch updates.

FirebaseAuth/Tests/Sample/Sample/SettingsViewController.m
paulb777: Hmm, it's arguable since we shouldn't really depend on private headers in samples. Let's revisit after we decide on whether the API should become public.

FirebaseCore/CHANGELOG.md
ryanwilson: Gotcha, thanks. I wasn't sure if it would pick it up appropriately given it's not a simple `X.Y.Z` string.

FirebaseInstallations.podspec
ryanwilson: Ahh okay, I wasn't sure if we intended to have an empty string here or not.

FirebaseStorageSwift.podspec
ryanwilson: SG, thanks for looking into it!

ryanwilson (approved):

paulb777:
Going to merge this here.

I'll follow up on any additional comments in another PR.

Also, I plan to make FIRVersion public in a subsequent PR.

PR #6635, opened 2020-10-01
FIS: don't log anything on success validation

maksymmalyhin:
Prevent message ` [Firebase/Installations][I-FIS002004] Stored installation validation error: (null)` from being logged when everything is ok.

#no-changelog


paulb777 (approved):
LGTM on green

PR #6651, opened 2020-10-02
Public Version API

paulb777:
- Makes FIRVersion API public
- Adds docs for deleted version APIs in #6634
- Update ObjC and Swift tests

API review document is under review

### DIFF
--- a/FirebaseCore/Sources/Public/FirebaseCore/FIRVersion.h
+++ b/FirebaseCore/Sources/Public/FirebaseCore/FIRVersion.h
@@ -16,9 +16,12 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
  * This class provides internal access to the Firebase version for any Firebase pod.
  */
+NS_SWIFT_NAME(FirebaseVersion)
 @interface FIRVersion : NSObject
 
 /**
@@ -27,3 +30,5 @@
 + (NSString *)version;
 
 @end
+
+NS_ASSUME_NONNULL_END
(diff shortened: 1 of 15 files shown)


FirebaseCore/Sources/Public/FirebaseCore/FIRVersion.h
```diff
@@ -16,9 +16,12 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
  * This class provides internal access to the Firebase version for any Firebase pod.
  */
+NS_SWIFT_NAME(FirebaseVersion)
```
ryanwilson, line 24: Since we're now exposing this we'll need to make the `init` unavailable, otherwise it will be surfaced in Swift.

SwiftPMTests/swift-test/main.swift
```diff
@@ -54,6 +54,9 @@ class importTest: XCTestCase {
     XCTAssertNil(FirebaseApp.app())
     XCTAssertEqual(GULAppEnvironmentUtil.deviceModel(), "x86_64")
 
+    let majorVersion = try XCTUnwrap(FirebaseVersion.version().first)
```
ryanwilson, line 57: I'm a little confused as what this `first` call is doing here - the `version` is a non-optional String. Perhaps doing something like:

- Split by `.`
- Assert there are 3 items in the array.
- Assert each item in that array is a number
- Maybe assert the first number is 7? Seems like the benefit outweighs the cost of having to trivially update it next breaking change.

ryanwilson (approved):

PR #6658, opened 2020-10-05
Fix tvOS storage issues.

### LINKED ISSUE #6612: Firebase Remote Config: Database cannot be created on a tvOS device

  * Xcode version: Xcode 12
  * Firebase SDK version: 6.33.0
  * Component version: 4.9.0
  * Platform: tvOS

tvOS is a more controlled ecosystem than iOS. In particular [local storage use is limited](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppleTV_PG), some locations not being writable in comparison to iOS.

The local Firebase Remote Config database, in particular, is currently [stored within the Application Support folder](https://github.com/firebase/firebase-ios-sdk/blob/CocoaPods-6.33.0/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m#L47-L55). This location is sadly not writable on tvOS, resulting in the [I-RCN000019 error](https://github.com/firebase/firebase-ios-sdk/blob/CocoaPods-6.33.0/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m#L87-L88) being logged to the console. This of course means that the local Firebase Remote Config database cannot be created and that configuration updates will not be persisted correctly.

Note that this behavior is only observed on a physical Apple TV device. The simulator, while in general quite accurately mirroring the behavior of a physical device, namely does not prevent write attempts to these locations.

#### Steps to reproduce:

I attached a very simple example to this issue. It only initializes the `FirebaseApp` and a `RemoteConfig` in its application delegate, which suffices to see the problem:

1. Unzip the [attached zip file](https://github.com/firebase/firebase-ios-sdk/files/5305588/FirebaseExample.zip).
1. Run `pod install` from the project root to integrate Firebase.
1. Open the sample project with Xcode 12.
1. Add a `GoogleService-Info.plist`file  to the project. There is no need for the configuration to be a perfect match, as Firebase will still attempt to create the local remote config database.
1. Edit your target and set an app id as well as a development team.
1. Run the example on an Apple TV device (the project targets tvOS 14).

You will notice a message similar to the one below logged to the console:

```
FirebaseExample[563:42589] 6.33.0 - [Firebase/RemoteConfig][I-RCN000019] Failed to create subdirectory for database file: Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “RemoteConfig” in the folder “Google”." UserInfo={NSFilePath=/var/mobile/Containers/Data/Application/5CBFFB1D-2F67-436E-BE10-8E75F8C1F830/Library/Application Support/Google/RemoteConfig, NSUnderlyingError=0x28320ccc0 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}.
```

#### Relevant Code:

A basic setup of Firebase with remote config suffices to show the issue:

```
import Firebase
import UIKit

@main
class AppDelegate: UIResponder, UIApplicationDelegate {
    var window: UIWindow?
    var remoteConfig: RemoteConfig?
    
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        FirebaseApp.configure()
        remoteConfig = RemoteConfig.remoteConfig()
        return true
    }
}
```

#### Possible solutions:

To solve this issue `Library/Caches` could be used instead. This would of course introduce a difference in behavior since:

- `Library/Application Support` is backed up and never purged.
- `Library/Caches` is not backed up and might be purged if space needs to be made available.

As far as I could see the Firebase database is some kind of cache which can be rebuilt even if cleaned. This is why I think `Library/Caches` could be considered a viable option for tvOS, probably for iOS as well.

#### Impact for other Firebase components:

This issue only discusses problems with Firebase Remote Config on tvOS devices. Though I haven't checked it is likely that some other Firebase components might attempt to write files in similarly forbidden locations on tvOS. 

These components should probably considered separately depending on their requirements so that a suitable location can be applied for each.

Thanks in advance for having a look at this issue!

ryanwilson:
On tvOS devices, the ApplicationSupport directory cannot be used.
Unfortunately this doesn't come up in simulator testing, only on device.
This hasn't been fully tested on device but given that any writes to the
ApplicationSupport directory fail on tvOS, this change can't break much.

Fixes #6612

@defagos can you please give this a test?


ryanwilson:
@diwu-arete I realize that Segmentation doesn't currently support tvOS but figure I'd make the change so if/when support happens it's one less thing to do 😃

GoogleUtilities/Environment/GULHeartbeatDateStorage.m
```diff
@@ -53,12 +53,17 @@ - (NSURL *)fileURL {
...
 
-/** Returns the URL path of the Application Support folder.
- * @return the URL path of Application Support.
+/** Returns the URL path of the directory for heartbeat storage data.
+ * @return the URL path of the directory for heartbeat storage data.
  */
 + (NSURL *)directoryPathURL {
+#if TARGET_OS_TV
+  NSArray<NSString *> *paths =
+      NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
```
maksymmalyhin, line 62: Thank you for the fix. I created #6662 for a specific tvOS implementation to improve reliability of the heartbeat.

FirebaseRemoteConfig/Tests/Unit/RCNTestUtilities.m
```diff
@@ -39,11 +39,15 @@ + (NSString *)generatedTestAppNameForTest:(NSString *)testName {
 
 /// Remote Config database path for test version
 + (NSString *)remoteConfigPathForTestDatabase {
-  NSArray<NSString *> *dirPaths =
+#if TARGET_OS_TV
+  NSArray *dirPaths = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+#else
+  NSArray *dirPaths =
```
maksymmalyhin, line 45: nit: It seems like this condition appears in the code enough times to justify creating a method in core or google utilities to encapsulate it. Not necessary should be a part of this PR.

maksymmalyhin (approved):
Thank you for the fix!

GoogleUtilities/Tests/Unit/Environment/GULHeartbeatDateStorageTest.m
```diff
@@ -26,8 +26,13 @@ @interface GULHeartbeatDateStorageTest : XCTestCase
 @implementation GULHeartbeatDateStorageTest
 
 - (void)setUp {
-  NSString *documentsPath = [NSSearchPathForDirectoriesInDomains(
-      NSApplicationSupportDirectory, NSUserDomainMask, YES) firstObject];
+#if TARGET_OS_TV
+  NSArray *documentsPath =
```
maksymmalyhin, line 30: Maybe rename it to paths?

GoogleUtilities/Tests/Unit/Environment/GULHeartbeatDateStorageTest.m
```diff
@@ -26,8 +26,13 @@ @interface GULHeartbeatDateStorageTest : XCTestCase
...
+      NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES);
+#else
+  NSArray *documentsPath =
+      NSSearchPathForDirectoriesInDomains(NSApplicationSupportDirectory, NSUserDomainMask, YES);
+#endif
   XCTAssertNotNil(documentsPath);
   NSURL *documentsURL = [NSURL fileURLWithPath:documentsPath];
```
maksymmalyhin, line 37: Should get a path from the array first.

maksymmalyhin (approved):

FirebaseRemoteConfig/Tests/Unit/RCNTestUtilities.m
ryanwilson: I think it's a reasonable idea, I do hesitate because I also want to make sure people are making intentional decisions about where they store things and recognize that this could be writing to a cache. Ideally I'd like to see this pop up in danger or something - if you're running tvOS tests and the active target has the `NSApplicationSupportDirectory` symbol, throw a warning/error.

ryanwilson:
@diwu-arete and @karenyz - any issues with landing this?

karenyz (approved):
I'm not familiar with tvOS, but seems like a safe fix since we're not changing non-tvOS behavior

GoogleDataTransport/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 - Legacy pre Xcode 10 compatibility checks removed. (#6486)
 - `GDTCORDirectorySizeTracker` crash fixed. (#6540)
+- Fixed writing heartbeat to disk on tvOS devices. (#6658)
```
paulb777, line 4: @ryanwilson I don't see any changes that go with this CHANGELOG update

PR #6664, opened 2020-10-05
Simplify FirebaseCore imports

paulb777:
Remove unnecessary imports and convert to forward `@class` declarations where possible.

#no-changelog


maksymmalyhin (approved):
nice

Issue #6644, opened 2020-10-02
FIRESTORE INTERNAL ASSERTION FAILED: Dangling query-target reference found


gautier-gdx:
[REQUIRED] Step 1: Describe your environment

Xcode version: 12.0.1
Firebase SDK version: 6.33.0
Firebase Component: Firestore
Component version: 6.33.0
Installation method: CocoaPods

[REQUIRED] Step 2: Describe the problem

So far this issue is only happening in production but it's happening in about 2% of all cases (about 25k crashes so far) So it starts to be a big problem for us

basically this is an internal error, probably related to levelDB
Here is the crash log

<img width="1272" alt="Screenshot 2020-10-02 at 16 50 12" src="https://user-images.githubusercontent.com/6576319/94937175-83bee800-04cf-11eb-9355-0c242c634b98.png">

`
Fatal Exception: NSInternalInconsistencyException
FIRESTORE INTERNAL ASSERTION FAILED: Dangling query-target reference found: [query_target: canonical_id=apps/App|f:|ob:asc target_id=14336] points to [target: target_id=14336]; seeking there found [target: target_id=26618]
`
```
> Fatal Exception: NSInternalInconsistencyException
> 0  CoreFoundation                 0x184a4b4e4 (Missing)
> 1  libobjc.A.dylib                0x18476e26c (Missing)
> 2  CoreFoundation                 0x18495b59c (Missing)
> 3  Foundation                     0x184e0ed98 (Missing)
> 4  Wizz                           0x100d00538 firebase::firestore::util::ObjcThrowHandler(firebase::firestore::util::ExceptionType, char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 59 (exception_apple.mm:59)
> 5  Wizz                           0x100d001e0 firebase::firestore::util::Throw(firebase::firestore::util::ExceptionType, char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 90 (exception.cc:90)
> 6  Wizz                           0x100d50530 firebase::firestore::util::internal::FailAssertion(char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, char const*) + 42 (hard_assert.cc:42)
[...]

PR #6685, opened 2020-10-07
Log warning instead of failing SDK when discovering dangling target.

wu-hui:
Although still not sure of underlying cause, this should make https://github.com/firebase/firebase-ios-sdk/issues/6644 stopping crashing.

### DIFF
--- a/Firestore/core/src/local/leveldb_target_cache.cc
+++ b/Firestore/core/src/local/leveldb_target_cache.cc
@@ -29,6 +29,7 @@
 #include "Firestore/core/src/model/document_key_set.h"
 #include "Firestore/core/src/nanopb/byte_string.h"
 #include "Firestore/core/src/nanopb/reader.h"
+#include "Firestore/core/src/util/log.h"
 #include "Firestore/core/src/util/string_apple.h"
 #include "absl/strings/match.h"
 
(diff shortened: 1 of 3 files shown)


Firestore/core/src/local/leveldb_target_cache.cc
```diff
@@ -168,7 +169,7 @@ absl::optional<TargetData> LevelDbTargetCache::GetTarget(const Target& target) {
     std::string target_key = LevelDbTargetKey::Key(row_key.target_id());
     target_iterator->Seek(target_key);
     if (!target_iterator->Valid() || target_iterator->key() != target_key) {
-      HARD_FAIL(
+      LOG_WARN(
```
wilhuff, line 172: `HARD_FAIL` could not return, which effectively ensured that the loop terminated and that the `target_iterator` would be both valid and point to a record that matched `target_key`. Changing this to `LOG_WARN` allows the code to continue, breaking this condition.

I think the right thing to do is add a `continue` below this to avoid trying to parse `target_iterator->value()`.

It's probably also worth adding a unit test to validate that this works. You can do this by writing a dummy row using `LevelDbQueryTargetKey` with a canonical ID that matches the query but has an invalid target ID.

wu-hui: Done.

wilhuff (approved):
LGTM

PR #6689, opened 2020-10-07
Fix an IID google3 build error

paulb777:
`FIRInstanceIDTokenOperationsTest.m:219:10: error: capturing 'operation' strongly in this block is likely to lead to a retain cycle [-Werror,-Warc-retain-cycles]`

#no-changelog


morganchen12 (approved):

paulb777:
cc: @chliangGoogle

PR #6700, opened 2020-10-08
Support paths with + in List API

schmidt-sebastian:
NSQueryParams doesn't encode `+` to `%2B`, which breaks the List API since it treats `+` as space. 

Note that other Storage APIs don't rely on query param encoding. There is already an integration tests that verifies that users can up and download files with `+` signs. 

Fixes b/170259897

### DIFF
--- a/FirebaseStorage/Tests/Unit/FIRStorageListTests.m
+++ b/FirebaseStorage/Tests/Unit/FIRStorageListTests.m
@@ -152,6 +152,42 @@
   [FIRStorageTestHelpers waitForExpectation:self];
 }
 
+- (void)testPercentEncodesPlusToken {
+  XCTestExpectation *expectation = [self expectationWithDescription:@"testPercentEncodesPlusToken"];
+  NSURL *expectedURL = [NSURL URLWithString:@"https://firebasestorage.googleapis.com/v0/b/bucket/"
+                                            @"o?prefix=%2Bfoo/&delimiter=/"];
+
+  self.fetcherService.testBlock =
+      ^(GTMSessionFetcher *fetcher, GTMSessionFetcherTestResponse response) {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-retain-cycles"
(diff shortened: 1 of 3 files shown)


FirebaseStorage/Tests/Unit/FIRStorageListTests.m
```diff
@@ -152,6 +152,42 @@ - (void)testListWithPageSizeAndPageToken {
   [FIRStorageTestHelpers waitForExpectation:self];
 }
 
+- (void)testPercentEncodesPlusToken {
+  XCTestExpectation *expectation = [self expectationWithDescription:@"testPercentEncodesPlusToken"];
+  NSURL *expectedURL = [NSURL URLWithString:@"https://firebasestorage.googleapis.com/v0/b/bucket/"
+                                            @"o?prefix=%2Bfoo/&delimiter=/"];
+
+  self.fetcherService.testBlock =
+      ^(GTMSessionFetcher *fetcher, GTMSessionFetcherTestResponse response) {
```
paulb777, line 161: Can you avoid the pragma with an approach like #6693?

paulb777 (approved):
Please add a changelog

FirebaseStorage/Tests/Unit/FIRStorageListTests.m
schmidt-sebastian: I am getting an unused variable warning, and I don't see how the code you linked to doesn't generate the same warning. Is there any trick I need to employ?

schmidt-sebastian (commented):
Sorry, forgot to push the Changelog. Added.

paulb777:
There's something weird going on with the XCTest macros. In the Messaging test, they reference self but not in the Storage test - I can't see what about the projects would trigger the difference.

I also don't reproduce the warning when I remove the #pragma's in the Storage test.

This probably isn't worth this much time - so I'm fine to merge as is.

schmidt-sebastian:
Thanks @paulb777!

PR #6711, opened 2020-10-09
Use completion handler when updating experiments after activation

### LINKED ISSUE #3687: AB tests not working when sending activation event right after activating remote config

karenyz:
Ensure that activate completion is called after `updateExperiments` finishes executing

Fixes #3687

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
+++ b/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
@@ -33,5 +33,5 @@
 - (void)updateExperimentsWithResponse:(NSArray<NSDictionary<NSString *, id> *> *)response;
 
 /// Update experiments to Firebase Analytics when `activateWithCompletion:` happens.
-- (void)updateExperiments;
+- (void)updateExperimentsWithHandler:(nullable void (^)(NSError *_Nullable error))handler;
 @end
--- a/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
+++ b/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
@@ -23,14 +23,15 @@
 @interface RCNConfigExperiment : NSObject
 
 /// Designated initializer;
-- (instancetype)initWithDBManager:(RCNConfigDBManager *)DBManager
-             experimentController:(FIRExperimentController *)controller NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithDBManager:(RCNConfigDBManager *_Nullable)DBManager
+                      experimentController:(FIRExperimentController *_Nullable)controller
+    NS_DESIGNATED_INITIALIZER;
 
 /// Use `initWithDBManager:` instead.
-- (instancetype)init NS_UNAVAILABLE;
+- (nullable instancetype)init NS_UNAVAILABLE;
 
 /// Update/Persist experiment information from config fetch response.
-- (void)updateExperimentsWithResponse:(NSArray<NSDictionary<NSString *, id> *> *)response;
+- (void)updateExperimentsWithResponse:(NSArray<NSDictionary<NSString *, id> *> *_Nullable)response;
 
 /// Update experiments to Firebase Analytics when `activateWithCompletion:` happens.
 - (void)updateExperimentsWithHandler:(nullable void (^)(NSError *_Nullable error))handler;
--- a/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
+++ b/FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
@@ -23,12 +23,12 @@
 @interface RCNConfigExperiment : NSObject
 
 /// Designated initializer;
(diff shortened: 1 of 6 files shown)


paulb777 (commented):
Change generally looks good.

Please fix the travis issue.

Also, I'm not sure why the GHA tests didn't kick off. Hopefully they'll run with the next commit ...

paulb777:
Looks like there may be a broader GHA problem - https://github.com/CocoaPods/CocoaPods/pull/10139#issuecomment-706419730.

Anyway a green travis run should be sufficient for this change if GHA doesn't get fixed.

FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
```diff
@@ -23,14 +23,15 @@
...
 /// Designated initializer;
-- (instancetype)initWithDBManager:(RCNConfigDBManager *)DBManager
-             experimentController:(FIRExperimentController *)controller NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithDBManager:(RCNConfigDBManager *_Nullable)DBManager
+                      experimentController:(FIRExperimentController *_Nullable)controller
+    NS_DESIGNATED_INITIALIZER;
 
 /// Use `initWithDBManager:` instead.
-- (instancetype)init NS_UNAVAILABLE;
+- (nullable instancetype)init NS_UNAVAILABLE;
```
paulb777, line 31: I don't think there should be a `nullable` on the NS_UNAVAILABLE init.

paulb777 (approved):
Thanks! LGTM with one question to resolve

FirebaseRemoteConfig/Sources/RCNConfigExperiment.h
karenyz: Should `initWithDbManager` be `nonnull` too?

paulb777: Yes. Some explanation at https://stackoverflow.com/a/42354763/556617

christibbs:
+1 for non-nullable initializers. LGTM otherwise.

christibbs (approved):

PR #6712, opened 2020-10-09
Change the Crashlytics version constant to read from FIRCore instead of from a build flag.

tejasd:
#no-changelog

### DIFF
--- a/Crashlytics/Shared/FIRCLSConstants.m
+++ b/Crashlytics/Shared/FIRCLSConstants.m
@@ -13,13 +13,14 @@
 // limitations under the License.
 
 #import "Crashlytics/Shared/FIRCLSConstants.h"
+#import <FirebaseCore/FIRVersion.h>
 
 #define STR_HELPER(x) #x
 #define STR(x) STR_HELPER(x)
 
 NSString* const FIRCLSDeveloperToken = @"77f0789d8e230eccdb4b99b82dccd78d47f9b604";
 
-NSString* const FIRCLSVersion = @STR(DISPLAY_VERSION);
+NSString* const FIRCLSVersion = FIRFirebaseVersion();
 
 // User Messages
 NSString* const FIRCLSMissingConsumerKeyMsg = @"consumer key is nil or zero length";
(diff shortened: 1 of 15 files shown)


Crashlytics/Shared/FIRCLSConstants.m
```diff
@@ -13,13 +13,14 @@
 // limitations under the License.
 
 #import "Crashlytics/Shared/FIRCLSConstants.h"
+#import <FirebaseCore/FIRVersion.h>
 
 #define STR_HELPER(x) #x
```
paulb777, line 18: Delete these two lines

paulb777 (changes requested):
Fix CI issues

tejasd:
> Fix CI issues

I think there's a few more issues - I'll send out changes on Tuesday when I'm back!

paulb777:
@tejasd Please provide an update. Code freeze is today. Can this change wait for 7.1?

tejasd:
> @tejasd Please provide an update. Code freeze is today. Can this change wait for 7.1?

Yes, apologies! This isn't very high priority - and some other stuff has come up! I'll get it done ASAP, but I think this'll go in in 7.1.

Crashlytics/Shared/FIRCLSConstants.m
tejasd: I think these are still needed, but deleted the ones in FIRCLSDefines.h

FirebaseCore/Sources/FIRVersion.m
```diff
@@ -28,5 +28,10 @@
 #define STR_EXPAND(x) #x
 
 NSString* FIRFirebaseVersion(void) {
+#ifdef FIREBASE_INTERNAL
```
tejasd, line 31: Is this acceptable? If not I can move this to a function in FIRCLSConstants.m

paulb777: Hmm, I'd prefer to determine if it would be useful to other teams before putting it here.  Without doing some investigation on the impact to the rest of Firebase, I'd prefer FIRCLSConstants.m.

Also, how about a less nebulous name than `INTERNAL`. How about `FIREBASE_1P`?

tejasd: Makes sense! Removed this change - and moved it to FIRCLSConstants.m.

Also renamed CRASHLYTICS_INTERNAL to CRASHLYTICS_1P!

samedson (approved):
👍

paulb777 (approved):
Thanks!

PR #6714, opened 2020-10-09
s/withMaxResults/maxResults

schmidt-sebastian:
Feel free to push back, but it would be nice to align this name with `data(maxSize)`.

### DIFF
--- a/FirebaseStorage/CHANGELOG.md
+++ b/FirebaseStorage/CHANGELOG.md
@@ -1,10 +1,10 @@
-# Unreleased
-- [fixed] Fixed an issue with the List API that prevented listing of locations
-  that contain the "+" sign.
-
 # 7.0.0
 - [changed] The global variable `FIRStorageVersionString` is deleted.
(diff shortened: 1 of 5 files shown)


paulb777 (commented):
Please fix CI

paulb777:
The change sounds reasonable to me. We'll discuss on the iCore team tomorrow.

ryanwilson:
I'm good with the change, I don't think we'd have the functionality to be able to deprecate the Swift naming anyways so it's acceptable in my eyes to do it at breaking change time. Definitely worth an email to the API council though, but LGTM for iOS.

schmidt-sebastian:
> I'm good with the change, I don't think we'd have the functionality to be able to deprecate the Swift naming anyways so it's acceptable in my eyes to do it at breaking change time. Definitely worth an email to the API council though, but LGTM for iOS.

I will email the council. The naming in this PR is the approved name from the review, so any pushback will likely just be from the breaking change.

paulb777 (approved):

maksymmalyhin (approved):
LGTM

PR #6718, opened 2020-10-12
[App Distribution] Add platform logging

jeremydurham:
Add platform logging for App Distribution

#no-changelog

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRFADApiService.m
+++ b/FirebaseAppDistribution/Sources/FIRFADApiService.m
@@ -19,15 +19,18 @@
 
 NSString *const kFIRFADApiErrorDomain = @"com.firebase.appdistribution.api";
 NSString *const kFIRFADApiErrorDetailsKey = @"details";
-NSString *const kHTTPGet = @"GET";
+NSString *const kFIRFADHTTPGet = @"GET";
 // The App Distribution Tester API endpoint used to retrieve releases
-NSString *const kReleasesEndpointURLTemplate =
+NSString *const kFIRFADReleasesEndpointURLTemplate =
     @"https://firebaseapptesters.googleapis.com/v1alpha/devices/"
     @"-/testerApps/%@/installations/%@/releases";
-NSString *const kInstallationAuthHeader = @"X-Goog-Firebase-Installations-Auth";
-NSString *const kApiHeaderKey = @"X-Goog-Api-Key";
-NSString *const kApiBundleKey = @"X-Ios-Bundle-Identifier";
-NSString *const kResponseReleasesKey = @"releases";
+NSString *const kFIRFADInstallationAuthHeader = @"X-Goog-Firebase-Installations-Auth";
+NSString *const kFIRFADApiHeaderKey = @"X-Goog-Api-Key";
+NSString *const kFIRFADApiBundleKey = @"X-Ios-Bundle-Identifier";
+NSString *const kFIRFADFirebaseUserAgentKey = @"X-firebase-client";
+NSString *const kFIRFADFirebaseHeartbeatKey = @"X-firebase-client-log-type";
+NSString *const kFIRFADHeartbeatTag = @"fire-fad";
+NSString *const kFIRFADResponseReleasesKey = @"releases";
 
 @implementation FIRFADApiService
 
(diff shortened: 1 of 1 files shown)


FirebaseAppDistribution/Sources/FIRFADApiService.m
```diff
@@ -72,9 +75,13 @@ + (NSMutableURLRequest *)createHTTPRequest:(NSString *)method
...
   [request setURL:[NSURL URLWithString:urlString]];
   [request setHTTPMethod:method];
-  [request setValue:authTokenResult.authToken forHTTPHeaderField:kInstallationAuthHeader];
-  [request setValue:[[FIRApp defaultApp] options].APIKey forHTTPHeaderField:kApiHeaderKey];
-  [request setValue:[NSBundle mainBundle].bundleIdentifier forHTTPHeaderField:kApiBundleKey];
+  [request setValue:authTokenResult.authToken forHTTPHeaderField:kFIRFADInstallationAuthHeader];
+  [request setValue:[[FIRApp defaultApp] options].APIKey forHTTPHeaderField:kFIRFADApiHeaderKey];
+  [request setValue:[NSBundle mainBundle].bundleIdentifier forHTTPHeaderField:kFIRFADApiBundleKey];
+  [request setValue:[FIRApp firebaseUserAgent] forHTTPHeaderField:kFIRFADFirebaseUserAgentKey];
+  [request setValue:@([FIRHeartbeatInfo heartbeatCodeForTag:kFIRFADHeartbeatTag]).stringValue
```
maksymmalyhin, line 82: `[FIRHeartbeatInfo heartbeatCodeForTag:]` makes disk operations, so it should not be called on main thread. It's hard to understand from the scope of the PR on what thread this method is called. Could you please check it.

schnecle: At the moment we leave whether this is on the main thread or a different thread to the caller (as I understand it) - if we need to make sure this is on a background thread, perhaps we should do that directly from this class.

schnecle (commented):
LGTM following a resolution on the main thread question

FirebaseAppDistribution/Sources/FIRFADApiService.m
maksymmalyhin: Then yes, I would recommend doing it directly.

FirebaseAppDistribution/Sources/FIRFADApiService.m
```diff
@@ -72,9 +75,15 @@ + (NSMutableURLRequest *)createHTTPRequest:(NSString *)method
...
   [request setHTTPMethod:method];
-  [request setValue:authTokenResult.authToken forHTTPHeaderField:kInstallationAuthHeader];
-  [request setValue:[[FIRApp defaultApp] options].APIKey forHTTPHeaderField:kApiHeaderKey];
-  [request setValue:[NSBundle mainBundle].bundleIdentifier forHTTPHeaderField:kApiBundleKey];
+  [request setValue:authTokenResult.authToken forHTTPHeaderField:kFIRFADInstallationAuthHeader];
+  [request setValue:[[FIRApp defaultApp] options].APIKey forHTTPHeaderField:kFIRFADApiHeaderKey];
+  [request setValue:[NSBundle mainBundle].bundleIdentifier forHTTPHeaderField:kFIRFADApiBundleKey];
+  [request setValue:[FIRApp firebaseUserAgent] forHTTPHeaderField:kFIRFADFirebaseUserAgentKey];
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
+    [request setValue:@([FIRHeartbeatInfo heartbeatCodeForTag:kFIRFADHeartbeatTag]).stringValue
```
maksymmalyhin, line 83: As we figured out recently, the server side integration with platform logging pipeline should be implemented before the SDK release, because otherwise if the SDK reports the global heartbeat, it won't be reported by other SDKs and will be lost which will affect the stats for all SDKs.

Please let me know if you need assistance with it.

maksymmalyhin (changes requested):
Mostly LGTM but a couple points to fix.

PR #6729, opened 2020-10-14
Update GDT version for M82

paulb777:


maksymmalyhin (approved):

visumickey:
Should Fireperf move to 8.0 dependency as well? And should the Fireperf version be 8.0.0 as well?

paulb777:
FirebasePerformance should be 7.0 to be consistent with Firebase. It's GDT dependency should be 8.0

maksymmalyhin:
@visumickey Fireperf should be consistent with the rest of Firebase SDKs, so it should be 7.0.0

Issue #6730, opened 2020-10-14
FR:Possible to make Dynamic Links parameters dictionary public


alihanaktay:
### Feature proposal

* Firebase Component: Firebase Dynamic Links

In our use case we need to send some parameters(e.g. utm_ campaign, utm_medium etc.) to the backend for checking is any discount available for this campaign. In current developer's can't reach the dynamic links parameters even if received by app. Is it possible to make Dynamic links parameters public?

rromanchuk:
Also  +1 for anything that can be set via console/api  link creation to have public accessors via an instance of `DynamicLink`,  just because i enjoy being lazy and would enjoy not parsing urls for already structured data. 

```swift
let handled = DynamicLinks.dynamicLinks().handleUniversalLink(link) { fdl, error in
  guard let fdl = fdl, let url = fdl.url else { return }
   
  // 😍  😍 😍 😍
  let campaign: String? = fdl.utmCampaign
  let medium: String? = fdl.utmMedium
}
```

paulb777:
Fixed by #7505

PR #6731, opened 2020-10-14
Dynamic Links parameters dictionary changed to public variable.

alihanaktay:
### Problem
In our use case we need to send some parameters(e.g. utm_ campaign, utm_medium etc.) to the backend for checking is any discount available for this campaign. In current developer's can't reach the dynamic links parameters even if received by app. Is it possible to make Dynamic links parameters public?
https://github.com/firebase/firebase-ios-sdk/issues/6730

### Solution
In `FIRDynamicLink+Private.h` file I've change parametersDictionary variable readonly to readwrite. Also I've added parametersDictionary to the `FIRDynamicLink.h` file for make it public.

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLink+Private.h
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLink+Private.h
@@ -41,7 +41,7 @@ typedef NS_ENUM(NSUInteger, FIRDynamicLinkMatchConfidence) {
 
 @property(nonatomic, copy, nullable) NSString *matchMessage;
 
-@property(nonatomic, copy, readonly) NSDictionary<NSString *, id> *parametersDictionary;
+@property(nonatomic, copy, readwrite) NSDictionary<NSString *, id> *parametersDictionary;
 
 @property(nonatomic, assign, readwrite) FIRDLMatchType matchType;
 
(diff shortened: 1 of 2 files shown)


paulb777:
@alihanaktay Thanks for the PR and sorry about the delay. We'll try to get a response by the end of the week.

FirebaseDynamicLinks/Sources/FIRDynamicLink+Private.h
```diff
@@ -41,7 +41,7 @@ typedef NS_ENUM(NSUInteger, FIRDynamicLinkMatchConfidence) {
 
 @property(nonatomic, copy, nullable) NSString *matchMessage;
 
-@property(nonatomic, copy, readonly) NSDictionary<NSString *, id> *parametersDictionary;
```
eldhosembabu, line 44: Can we get rid of this property from here since we are adding it in FIRDynamicLink.h so that we don't need to add readwrite?

alihanaktay: Hello for sure we can get rid of this property. As you said since we have moved this variable to FIRDynamicLink.h we don't need in Private header. So I've removed that and pushed.

utkuyildirim:
When this improvement will be merged?

eldhosembabu:
The changes looks good to me but we need product team approval to expose this property. Checking on this now.

eldhosembabu:
Had the talk with the product team and got initial approval. So now i need to get the same api implemented for Android SDK as well and get the whole thing approved by api counsel. This will take some time and will keep you posted on this.

eldhosembabu:
The api review got approved with some changes. Based on that we will be exposing a property "utmParametersDictionary" containing the utm params associated with an FDL. Please refer to PR with the changes : https://github.com/firebase/firebase-ios-sdk/pull/7505

paulb777:
Closing based on the merge in #7505.  Thanks for getting this going!

PR #6733, opened 2020-10-14
Misc podspec fixes for Firebase 7

paulb777:
Merge back 7.0 podspec fixes back to master

#no-changelog

### DIFF
--- a/Firebase.podspec
+++ b/Firebase.podspec
@@ -31,6 +31,7 @@ Simplify your app development, grow your user base, and monetize more effectivel
   s.default_subspec = 'Core'
 
   s.subspec 'Core' do |ss|
+    ss.ios.deployment_target = '9.0'
     ss.ios.dependency 'FirebaseAnalytics', '7.0.0'
     ss.dependency 'Firebase/CoreOnly'
   end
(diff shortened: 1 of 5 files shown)


Firebase.podspec
```diff
@@ -48,8 +49,7 @@ Simplify your app development, grow your user base, and monetize more effectivel
         'HEADER_SEARCH_PATHS' => "$(inherited) ${PODS_ROOT}/Firebase/CoreOnly/Sources"
       }
     end
-    # Standard platforms PLUS watchOS.
-    ss.ios.deployment_target = '10.0'
+    ss.ios.deployment_target = '9.0'
```
visumickey, line 52: Is this 9.0? Should we move the version dependencies of other Firebase SDKs to be 9.0 (Currently we moved Fireperf to 10.0)

visumickey (approved):

Firebase.podspec
paulb777: Yes - Analytics and AdMob are still 9. Everything else should now be at 10.

maksymmalyhin (approved):

Firebase.podspec
visumickey: Perfect. Good then!

PR #6669, opened 2020-10-06
Fix an issue notification is not received during app first install

chliangGoogle:
Fixes #6553.

There's a small chance the APNS token is returned from Apple with a delay so the first token request was sent without the apns token info. However, the retry logic didn't really trigger a second request, so fixed that.

PR #6737, opened 2020-10-15
Missing iid handler checking cause crash

chliangGoogle:
It was introduced when we start calling the token call without passing a handler: Fix #6669

Also the changelog for the whole change should be in IID not messaging.


paulb777 (approved):
LGTM on green.

Should this go into Firebase 7?

chliangGoogle:
Yes, is it possible to do a cherry pick on this?

paulb777:
Yes. I'll manage it. I'll merge if master and the release branch are still aligned

morganchen12 (approved):

PR #6747, opened 2020-10-16
Firebase zip contents from FirebaseManifest instead of textproto

paulb777:
Built zip candidate for 7.0.0 and reviewed diffs with previous zips.

#no-changelog

PR #6749, opened 2020-10-16
Address #6747 review comments

paulb777:
#no-changelog

The first code chunk is a bit complicated to continue supporting the `-templateDir` option while adding the new more powerful `-repoDir` option.

The map suggestion is nice and makes the code much cleaner. Thanks @maksymmalyhin

### DIFF
--- a/ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
+++ b/ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
@@ -163,7 +163,7 @@ struct ZipBuilder {
   /// - Throws: One of many errors that could have happened during the build phase.
   func buildAndAssembleFirebaseRelease(inProjectDir projectDir: URL) throws -> ReleaseArtifacts {
     let manifest = FirebaseManifest.shared
-    var podsToInstall = (manifest.pods.filter { $0.zip }).map {
+    var podsToInstall = manifest.pods.filter { $0.zip }.map {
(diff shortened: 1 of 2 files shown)


ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -162,11 +162,9 @@ struct ZipBuilder {
   /// - Returns: Information related to the built artifacts.
   /// - Throws: One of many errors that could have happened during the build phase.
   func buildAndAssembleFirebaseRelease(inProjectDir projectDir: URL) throws -> ReleaseArtifacts {
-    var podsToInstall: [CocoaPodUtils.VersionedPod] = []
     let manifest = FirebaseManifest.shared
-    for pod in (manifest.pods.filter { $0.zip }) {
-      podsToInstall.append(CocoaPodUtils.VersionedPod(name: pod.name,
-                                                      version: manifest.versionString(pod)))
+    var podsToInstall = (manifest.pods.filter { $0.zip }).map {
```
maksymmalyhin, line 166: I curious why parentheses are needed around `manifest.pods.filter { $0.zip }` ?

maksymmalyhin (approved):
Thanks!

ZipBuilder/Sources/ZipBuilder/ZipBuilder.swift
paulb777: Removed. Good catch!

PR #6758, opened 2020-10-16
Finish port of update-versions.py

paulb777:
Port Python `checkVersions` to Swift `verifyTagIsSafeToAdd` and add origin checking in addition to local branch checking.

The `base_branch` option should not be necessary now that the plan is to release patch versions from the existing branch labeled with the minor version.

#no-changelog


ZipBuilder/Sources/FirebaseReleaser/Tags.swift
```diff
@@ -77,4 +79,27 @@ enum Tags {
...
+  private static func verifyTagIsSafeToAdd(_ tag: String, gitRoot: URL) {
+    if checkTagCommand("git tag -l \(tag)", gitRoot: gitRoot) != "" {
+      fatalError("Tag \(tag) already exists locally. Please delete and restart")
+    }
+    if checkTagCommand("git ls-remote origin refs/tags/\(tag)", gitRoot: gitRoot) != "" {
+      fatalError("Tag \(tag) already exists locally. Please delete and restart")
+    }
+  }
+
+  private static func checkTagCommand(_ command: String, gitRoot: URL) -> String {
```
maksymmalyhin, line 92: Could you please add some explanation to what the method does to help us in the future with understanding the code.

maksymmalyhin (approved):
👍

PR #6772, opened 2020-10-19
Rename ReleaseTooling

paulb777:
Rename top-level `ZipBuilder` directory to `ReleaseTooling`
Rename `ReleasePackager` executable to `zip-builder`

The `ReleaseTooling` directory now has two major tools that share underlying building blocks:

- A tool to build a zip distribution for Firebase and for a generic set of CocoaPods
- A tool to automate the Firebase release process

#no-changelog


ryanwilson (approved):
LGTM - `zip-builder` for the tool is alright, I think we originally moved away from it since it also packages Carthage now but I still think it's fine.

PR #6774, opened 2020-10-19
Address two comments from last week's PRs

paulb777:
- `pod repo update` before checking the master spec repo for existing tags
- Comments for `--log-only` option

#no-changelog


ryanwilson (approved):
Thanks!

PR #6779, opened 2020-10-20
Fix a memory leak introduced in #6418 (#6778)

paulb777:


var-const (approved):

paulb777:
The SPM CI failure is a result of #6777. I'll make a fix for that and then restart this.

paulb777:
Merging now. I'll update the tags and rebuild the zip in the morning.

PR #6770, opened 2020-10-19
Only run schedule actions on the main repo

paulb777:
Fix #6769 

#no-changelog

PR #6785, opened 2020-10-20
Fix cron logic for forks

paulb777:
Continue #6770 to fix a second class of inadvertent `schedule` jobs on forked repos

#no-changelog

### DIFF
--- a/.github/workflows/abtesting.yml
+++ b/.github/workflows/abtesting.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   pod-lib-lint:
     # Don't run on private repo unless it is a PR.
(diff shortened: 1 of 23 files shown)


maksymmalyhin (approved):
Thank you Paul. 

I wish we could have a single place to change these conditions. Do you know if we can use a script here or some kind of  action template?

paulb777:
I agree. It would be nice to templatize these. It might be possible to do with [workflow templates](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization), but investigation is needed.

mikehardy:
Thanks for fixing these, I can almost feel the CPUs cooling and the fans spinning down :-)

PR #6799, opened 2020-10-21
Firebase user agent: add iOS on mac platform

maksymmalyhin:


ryanwilson (approved):
I'm good with this change, we should try to confirm that this works as expected (specifically that it falls into the `TARGET_OS_IOS` bucket). I imagine it does, but just in case :P

paulb777 (approved):

GoogleUtilities/Environment/third_party/GULAppEnvironmentUtil.m
```diff
@@ -272,7 +272,11 @@ + (NSString *)applePlatform {
 #if TARGET_OS_MACCATALYST
   applePlatform = @"maccatalyst";
 #elif TARGET_OS_IOS
-  applePlatform = @"ios";
+  if (@available(iOS 14.0, *)) {
```
paulb777, line 275: Should we also add Catalyst at the same time?

maksymmalyhin: Taking into account how ugly are the availability checks I would prefer to keep using `#if TARGET_OS_MACCATALYST` that works just fine.

maksymmalyhin: We have to use the runtime check for iOS on mac because at compile time it is still iOS.

paulb777 (approved):

PR #6807, opened 2020-10-22
Stop failing on warnings

paulb777:
We already check for warnings in our CI and should avoid causing failures if future versions of Xcode introduce warnings.

More context at https://github.com/flutter/flutter/issues/68813


paulb777:
The CI failure is unrelated addressed in #6809

maksymmalyhin (approved):

PR #6808, opened 2020-10-22
Improve performance of SDK on startup

elenadoty:


Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -332,7 +333,7 @@ - (FBLPromise *)deleteUnsentReports {
   }
 
   // Grab existing reports
-  BOOL launchFailure = [self checkForAndCreateLaunchMarker];
+  [self checkForAndCreateLaunchMarker];
```
samedson, line 336: What was the reason we made `launchFailure` a class property? With the current code, I think I'd rather make it a local variable just so we don't attempt to check it before `checkForAndCreateLaunchMarker`.

samedson (commented):
I'm seeing a few errors on Catalyst / tvOS in the tests

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
elenadoty: It was because the checkForAndCreateLaunchMarker method creates a launch marker, so when I try to use that method again later in the code it always returns true even if there wasn't a class.

PR #6817, opened 2020-10-23
Rebrand 3.3.0 as Perf SDK version 3.3.2 to avoid the crashes related to KVO

visumickey:


paulb777 (approved):

FirebasePerformance.podspec.json
```diff
@@ -32,11 +32,11 @@
     "README.md"
   ],
   "source": {
-    "http": "https://dl.google.com/dl/cpdc/ec10178ac6b36ea6/FirebasePerformance-3.3.0.tar.gz"
+    "http": "https://dl.google.com/dl/cpdc/ec10178ac6b36ea6/FirebasePerformance-3.3.2.tar.gz"
```
paulb777, line 35: This one shouldn't change, right? Shouldn't stay the same as 3.3.0?

visumickey: Oopsy! You are right! Fixing it. Thanks for catching it Paul.

raymondlam (approved):

Issue #6721, opened 2020-10-13
Empty orderBy field path when retrieving cached query from LevelDB


gautier-gdx:
[REQUIRED] Step 1: Describe your environment

Xcode version: 12.0.1
Firebase SDK version: 6.34.0
Firebase Component: Firestore
Component version: 6.34.0
Installation method: CocoaPods

[REQUIRED] Step 2: Describe the problem

We had a lot of crash recently with the following error: 
`FIRESTORE INTERNAL ASSERTION FAILED: Invalid field path (). Paths must not be empty, begin with '.', end with '.', or contain '..' (expected !segment.empty())`

After inspecting the code and the stack trace, it's seems that this crash is coming from an assertion when trying to retrieve a query from the levelDB caches specifically on a orderBy. because it's coming from the `FromServerFormatView`, we don't have any crashes coming from  `FromDotSeparatedStringView` so it's not directly related to our codebase

Because all the `FIRESTORE INTERNAL ASSERTION FAILED` are grouped in one single crash in Crashlytics I can't really tell you how much it's impacting our app but it's one of the biggest redundancies.

Here is the full stack trace:

```
FIRESTORE INTERNAL ASSERTION FAILED: Invalid field path (). Paths must not be empty, begin with '.', end with '.', or contain '..' (expected !segment.empty())
Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x19e625114 __exceptionPreprocess
1  libobjc.A.dylib                0x1b1e4bcb4 objc_exception_throw
2  CoreFoundation                 0x19e534308 -[CFPrefsSearchListSource addManagedSourceForIdentifier:user:]
3  Foundation                     0x19f866ca8 -[NSAssertionHandler handleFailureInFunction:file:lineNumber:description:]
4  Wizz                           0x100a9c7b0 firebase::firestore::util::ObjcThrowHandler(firebase::firestore::util::ExceptionType, char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 59 (exception_apple.mm:59)
5  Wizz                           0x100a9c300 firebase::firestore::util::Throw(firebase::firestore::util::ExceptionType, char const*, char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 91 (exception.cc:91)
[...]

PR #6818, opened 2020-10-23
Include QueryTarget proto in error messages when serialization fails

schmidt-sebastian:
To help us debug serialization failures, this PR logs the Proto message for queries that fail serialization.

Addresses https://github.com/firebase/firebase-ios-sdk/issues/6721


Firestore/core/test/unit/local/local_serializer_test.cc
```diff
@@ -352,6 +352,34 @@ TEST_F(LocalSerializerTest, EncodesTargetData) {
...
+  v1::StructuredQuery::Order order;
+  order.mutable_field()->set_field_path(invalid_field_path);
+  order.set_direction(v1::StructuredQuery::ASCENDING);
+  *query_proto->mutable_structured_query()->add_order_by() = std::move(order);
+
+  ByteString bytes = ProtobufSerialize(invalid_target);
+  StringReader reader(bytes);
+
+  auto message = Message<firestore_client_Target>::TryParse(&reader);
+  serializer.DecodeTargetData(&reader, *message);
```
var-const, line 379: Question: what does `message` look like after the call? (I don't suggest checking `message` in the test, just curious)

Firestore/core/test/unit/model/field_path_test.cc
```diff
@@ -26,6 +28,12 @@ namespace firebase {
 namespace firestore {
 namespace model {
 
+namespace {
+FieldPath Parse(const std::string& path) {
+  return FieldPath::FromServerFormat(path).ConsumeValueOrDie();
+}
+}  // anonymous namespace
```
var-const, line 35: Ultranit: the more common Google style convention is to just use a blank "name" for an unnamed namespace; i.e. this should be just `// namespace`, without the `anonymous` part.

Firestore/core/test/unit/model/field_path_test.cc
```diff
@@ -26,6 +28,12 @@ namespace firebase {
 namespace firestore {
 namespace model {
 
+namespace {
+FieldPath Parse(const std::string& path) {
```
var-const, line 32: Optional nit: I'd add blank lines after the opening brace of the namespace and before the closing brace to make it a little easier to see the function separately, but it's just a personal preference -- feel free to ignore.

var-const (commented):
Almost LGTM.

Firestore/core/src/remote/serializer.h
```diff
@@ -171,7 +171,7 @@ class Serializer {
 
   static google_firestore_v1_DocumentMask EncodeFieldMask(
       const model::FieldMask& mask);
-  static model::FieldMask DecodeFieldMask(
+  static util::StatusOr<model::FieldMask> DecodeFieldMask(
```
wilhuff, line 174: This method should take a `Reader` and should just return the `FieldMask`. This will significantly reduce the burden on all the callers.

This will also get rid of all the early returns you've added in the implementation. Once the reader has failed, the result of these values doesn't matter so you'll be able to get rid of the `InvalidOrderBy` method and similar--the result is already going to be invalid.

Firestore/core/src/remote/serializer.h
```diff
@@ -189,7 +189,8 @@ class Serializer {
   EncodeListenRequestLabels(const local::TargetData& target_data) const;
 
   static pb_bytes_array_t* EncodeFieldPath(const model::FieldPath& field_path);
-  static model::FieldPath DecodeFieldPath(const pb_bytes_array_t* field_path);
+  static util::StatusOr<model::FieldPath> DecodeFieldPath(
```
wilhuff, line 192: Similarly, this should take a `Reader` and avoid returning `StatusOr`.

Firestore/core/test/unit/model/field_path_test.cc
schmidt-sebastian: Done

Firestore/core/test/unit/model/field_path_test.cc
schmidt-sebastian: Done. I also fixed the instance I was inspired by.

Firestore/core/src/remote/serializer.h
schmidt-sebastian: Much cleaner. Thanks.

Firestore/core/src/remote/serializer.h
schmidt-sebastian: Thanks

Firestore/core/src/remote/serializer.cc
```diff
@@ -828,21 +851,28 @@ Serializer::EncodeFieldTransform(const FieldTransform& field_transform) const {
 FieldTransform Serializer::DecodeFieldTransform(
     nanopb::Reader* reader,
     const google_firestore_v1_DocumentTransform_FieldTransform& proto) const {
+  StatusOr<FieldPath> field_path = DecodeFieldPath(proto.field_path);
```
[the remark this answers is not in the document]
schmidt-sebastian: This is no longer applicable thanks to Gil's suggestion.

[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/field_path.cc
```diff
@@ -154,23 +166,37 @@ FieldPath FieldPath::FromServerFormatView(absl::string_view path) {
...
+        }
         break;
 
       default:
         segment += c;
         break;
     }
     ++i;
+
+    if (!status.ok()) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/field_path.cc
```diff
@@ -154,23 +166,37 @@ FieldPath FieldPath::FromServerFormatView(absl::string_view path) {
...
+      return status;
+    }
   }
-  finish_segment();
 
-  HARD_ASSERT(!inside_backticks, "Unterminated ` in path %s", path);
+  status = finish_segment();
```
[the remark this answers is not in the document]
schmidt-sebastian: Good catch. Fixed.

Firestore/core/src/model/field_path.cc
```diff
@@ -154,23 +166,37 @@ FieldPath FieldPath::FromServerFormatView(absl::string_view path) {
...
+
+  if (inside_backticks) {
+    status = Status{Error::kErrorInvalidArgument,
+                    StringFormat("Unterminated ` in path %s", path)};
+  }
 
-  return FieldPath{std::move(segments)};
+  return status.ok() ? StatusOr<FieldPath>{FieldPath{std::move(segments)}}
```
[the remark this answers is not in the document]
schmidt-sebastian: I simplified this by using Gil's suggestion and return early in the `inside_backticks` check.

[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/field_path.cc
```diff
@@ -154,23 +166,37 @@ FieldPath FieldPath::FromServerFormatView(absl::string_view path) {
         break;
 
       case '\\':
-        HARD_ASSERT(i + 1 != path.size(),
-                    "Trailing escape characters not allowed in %s", path);
-        ++i;
-        segment += path[i];
+        if (i + 1 == path.size()) {
+          status =
```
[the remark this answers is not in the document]
schmidt-sebastian: The code looks cleaner to me with just one exit point in the loop. Kept as is for now.

Firestore/core/test/unit/local/local_serializer_test.cc
schmidt-sebastian: `Message.ToString()` prints:

```
<Target 0x7ffeedc66090>: {
  target_id: 42
  query {
    parent: "projects/p/databases/d/documents"
    structured_query {
      from {
        collection_id: "room"
      }
      order_by {
        field {
          field_path: "`"
        }
        direction: ASCENDING
      }
    }
  }
```

I hope this bubbles up properly to Crashlytics.

Firestore/core/src/remote/serializer.cc
```diff
@@ -1198,31 +1168,24 @@ Filter Serializer::DecodeUnaryFilter(
...
               unary.which_operand_type);
 
-  StatusOr<FieldPath> field =
-      FieldPath::FromServerFormat(DecodeString(unary.field.field_path));
-
-  if (!field.ok()) {
-    reader->set_status(field.status());
-    return InvalidFilter();
-  }
+  auto field = DecodeFieldPath(reader, unary.field.field_path);
```
var-const, line 1171: Similar nit: similar statements above use an explicit types, please make consistent (I'm fine with either way).

Firestore/core/src/remote/serializer.cc
```diff
@@ -1244,7 +1207,7 @@ FilterList Serializer::DecodeCompositeFilter(
   FilterList result;
   result = result.reserve(composite.filters_count);
 
-  for (pb_size_t i = 0; reader->ok() && i != composite.filters_count; ++i) {
+  for (pb_size_t i = 0; i != composite.filters_count; ++i) {
```
var-const: Question: what is the reason for this change?

Firestore/core/src/model/field_path.cc
```diff
@@ -109,23 +114,30 @@ FieldPath FieldPath::FromDotSeparatedStringView(absl::string_view path) {
...
-                path);
+    if (segment.empty()) {
+      return Status{
+          Error::kErrorInvalidArgument,
+          StringFormat(
+              "Invalid field path (%s). Paths must not be empty, begin with "
+              "'.', end with '.', or contain '..'",
+              path)};
+    }
     // Move operation will clear segment, but capacity will remain the same
```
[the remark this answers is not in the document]
var-const: This is unlikely to lead to an empty `FieldPath` -- if moving doesn't clear the moved-from `segment`, the resulting `FieldPath` may contain extra characters. It probably doesn't happen in practice, but still seems prudent to fix. I can do that in a follow-up PR.

Firestore/core/src/remote/serializer.cc
schmidt-sebastian: I just want to show off my versatile coding skills here. Fixed.

Firestore/core/src/remote/serializer.cc
schmidt-sebastian: It simplifies the loop. The check should no longer be needed since an `InvalidFieldPath` is semantically valid in the consuming types (FieldMask, QueryTarget). I can change this back if exiting early is a better strategy here.

var-const: I presume exiting early is better from the performance perspective, though now that I think about it, it only matters in the error case which should be rare. I'll leave it up to you.

var-const (approved):

PR #6838, opened 2020-10-26
Remove Component version from issue template

paulb777:
Now that Firebase 7.0.0 is a single Firebase version

#no-changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,9 +25,8 @@ template_path=.github/ISSUE_TEMPLATE/bug_report.md
 
   * Xcode version: _____
   * Firebase SDK version: _____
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):

maksymmalyhin:
Sorry for being late to the party. As for me we should remove the component version as we still have a significant amount of developers using previous Firebase version. I would rather mark the field as optional or "if applicable" instead.

paulb777:
My thought was that the usability versus value trade-off isn't justified.  The extra version field will be confusing to 7.x, Carthage, SPM and zip users. And we can almost always determine the component version from the Firebase version, and if we suspect something weird, we need to ask for a Podfile.lock anyway.  I'm happy to chat if you'd like to discuss further.

maksymmalyhin:
Now when you explained it it looks reasonable to me. Thanks.

PR #6839, opened 2020-10-26
should check completion since it can be nullable

chliangGoogle:


morganchen12 (approved):
LGTM, please add a changelog though.

chliangGoogle:
Thanks for the quick response. I was confused between the error type and completion type, [the completion is default nonnull ](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/Public/FirebaseMessaging/FIRMessaging.h#L394)

Close this PR and sorry for the confusion.

PR #6853, opened 2020-10-28
Fix cache for check workflow

### LINKED ISSUE #6848: Suboptimal cache usage for the `check` workflow
  GitHub Actions

Each pull request uses its own separate cache for the check workflow

#### Steps to reproduce:
Open a pull request, it will take a long time to setup for the first time because loading from cache will miss even though the other pull requests are using the same version of swiftformat.
 
#### Relevant Code:
[Check workflow](https://github.com/firebase/firebase-ios-sdk/blob/202df4134d50a92b2149cdbc5f370eb374a284e3/.github/workflows/check.yml)
Other workflows might also be affected.

This is because of the [following restrictions](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).

The easiest solution is to also trigger the action when pushing to the default branch, like [this](https://github.com/Cyberbeni/TypedNotificationCenter/blob/32a5871bdd1c90177cf3f95c6a1add8b2a076adc/.github/workflows/lint.yml), so pull requests will find the cache from the default branch.

paulb777:
Fix #6848 

#no-changelog

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths-ignore:
       - 'Firestore/**'
+  push:
+    branches: master
 
 jobs:
   check:


wilhuff (approved):
LGTM

PR #6858, opened 2020-10-28
Remove explicit MobileCoreServices from podspecs

### LINKED ISSUE #6850: Target Integrity - MobileCoreServices has been renamed

  * Xcode version: 12.1
  * Firebase SDK version: 7.0.0
  * Firebase Component: Firestore, Storage

After updating version 7 and trying to compile these two warnings appear.

Captura de Pantalla 2020-10-27 a la(s) 13 37 19

paulb777:
Fix #6850 

It seems that it's not necessary with modern Xcode versions to explicitly link `MobileCoreServices` and its linkage is problematic because it causes build warnings with CocoaPods (but not Swift Package Manager).

- MobileCoreServices is used in FirebaseStorage and FirebaseDatabase, but not Firestore
- Only the FirebaseStorage and Firestore podspecs were specifying it
- We saw no issues from FirebaseDatabase including MobileCoreServices but not adding an explicit linkage
- With Firebase 7 updating to a minimum iOS version of 10, we're seeing build warnings even though the docs say MobileCoreServices is not deprecated until iOS 12.

### DIFF
--- a/FirebaseFirestore.podspec
+++ b/FirebaseFirestore.podspec
@@ -72,7 +72,7 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   s.dependency 'leveldb-library', '~> 1.22'
   s.dependency 'nanopb', '~> 2.30906.0'
 
-  s.ios.frameworks = 'MobileCoreServices', 'SystemConfiguration', 'UIKit'
+  s.ios.frameworks = 'SystemConfiguration', 'UIKit'
   s.osx.frameworks = 'SystemConfiguration'
   s.tvos.frameworks = 'SystemConfiguration', 'UIKit'
 
--- a/FirebaseStorage.podspec
+++ b/FirebaseStorage.podspec
@@ -31,7 +31,6 @@ Firebase Storage provides robust, secure file uploads and downloads from Firebas
   ]
   s.public_header_files = 'FirebaseStorage/Sources/Public/FirebaseStorage/*.h'
 
-  s.ios.framework = 'MobileCoreServices'
   s.osx.framework = 'CoreServices'
 
   s.dependency 'FirebaseCore', '~> 7.0'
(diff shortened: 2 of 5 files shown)


wilhuff:
The Firestore change seems reasonable since we're not actually using anything from this framework.

The other changes rely on autolinking to get things right. It's not that the dependency isn't there, it's just no longer explicitly there. Moving toward implicit dependencies seems like a bad idea.

Why aren't we treating these warnings as a CocoaPods bug?

paulb777:
As far as I can tell the CocoaPods output is correct, so it's more likely to be an Xcode bug.  Given that we've never had a report of the implicit linking failing in FirebaseDatabase, it seems expedient to do the same in FirebaseStorage.

schmidt-sebastian (approved):
LGTM based on CI and discussions in this PR.

wilhuff:
The CocoaPods bug could be considered to be that it enables warnings by default for dependencies of your dependencies.

I don't feel strongly about this, and you're right that autolinking hasn't bitten us in RTDB, so I'm definitely not going to stand in the way of this. I'm pushing for the other solution only because we see variations on that same core problem repeatedly.

wilhuff:
Also, to be clear: a reasonable interim solution could be to submit this now while working with the CocoaPods folks (assuming this is even feasible/interesting from their side and that there's appetite to push for it from our side).

In other words, LGTM to submit :-).

paulb777:
Thanks :). CocoaPods already does optionally [allow disabling dependency warnings](https://stackoverflow.com/questions/13208202/ignore-xcode-warnings-when-using-cocoapods).  However it seems that `pod lib lint` only errors on compile warnings and not linker warnings.

I could see arguments both ways for which way should be the default, so ultimately I think Firebase should strive to be warning free.

PR #6860, opened 2020-10-28
Add standard SwiftPM versioning support

paulb777:
Add support for installing Firebase with standard Swift Package Manager versioning.

This requires the dependencies for non-tests to also use versioning and requires them to not use unsafeOptions.

gRPC is updated accordingly at  https://github.com/firebase/grpc-SwiftPM/commit/91b62619e6c83bc5f1b99d9d60fe46b2862d3a5a

All the versions match the source of the corresponding CocoaPods version with the exception of leveldb. I hadn't realized before that the Swift Package Manager version was a bit later than 1.22 so I versioned it 1.22.1 for SwiftPM.

#no-changelog


paulb777:
cc: @wilhuff @yulin-liang

Package.swift
```diff
@@ -100,19 +100,17 @@ let package = Package(
     ),
     .package(
       name: "nanopb",
-      url: "https://github.com/nanopb/nanopb.git",
+      url: "https://github.com/firebase/nanopb.git",
       // This revision adds SPM enablement to the 0.3.9.6 release tag.
```
ryanwilson, line 104: Is this comment out of date?

ryanwilson (approved):
Thanks!

PR #6863, opened 2020-10-29
FCM: Fix crash on iOS 14 when multiple app delegate completion methods are called

mikeger:
## Background

Since iOS 7 it is possible to execute the code in the background for iOS apps that receive silent push notifications. Once such push (most likely containing the `content-available` flag) is received, if the application is in the background, it has 30 seconds to complete the background activity and tell iOS that it's done by calling the `completion` block. More info [on application(_:didReceiveRemoteNotification:fetchCompletionHandler:) here](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application)

## Issue we observed

We discovered that on iOS 14 calling the completion block multiple times is leading to a crash. Based on my limited debugging capabilities, it looks like the API is now using dispatch groups to trace the moment when the completion is called, and in the completion block it invokes `dispatch_group_leave`. Once the completion block is called multiple times, the iOS is going to crash since this behavior is unexpected (calling `dispatch_group_leave` more times than `dispatch_group_enter`).

More people complained about the same issue here https://github.com/invertase/react-native-firebase/issues/3944. 

Worth noticing that in our experience the crash is only happening when the app is actually running in the foreground and in case the FCM is used together with the Braze SDK (they are also using `GULAppDelegateSwizzler`).

## Proposed solution

> Similia similibus curantur

It is sufficient to wrap the final completion block provided by iOS into yet another `dispatch_group` that would be entered for every observer of the `application(_:didReceiveRemoteNotification:fetchCompletionHandler:)` and left when the observer is calling to the provided callback.

### DIFF
--- a/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
+++ b/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
@@ -22,6 +22,7 @@
 #import "GoogleUtilities/Network/Public/GoogleUtilities/GULMutableDictionary.h"
 
 #import <objc/runtime.h>
+#import <dispatch/group.h>
 
 // Implementations need to be typed before calling the implementation directly to cast the
 // arguments and the return types correctly. Otherwise, it will crash the app.
@@ -881,18 +882,30 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
   GULRealDidReceiveRemoteNotificationWithCompletionIMP
       didReceiveRemoteNotificationWithCompletionIMP =
           [didReceiveRemoteNotificationWithCompletionIMPPointer pointerValue];
-
+  
+  dispatch_group_t __block callbackGroup = dispatch_group_create();
+  UIBackgroundFetchResult __block latestFetchResult = UIBackgroundFetchResultNoData;
+  dispatch_group_notify(callbackGroup, dispatch_get_main_queue(), ^() {
+    completionHandler(latestFetchResult);
+  });
   // Notify interceptors.
   [GULAppDelegateSwizzler
       notifyInterceptorsWithMethodSelector:methodSelector
                                   callback:^(id<GULApplicationDelegate> interceptor) {
+    
+                                    dispatch_group_enter(callbackGroup);
+                                    void (^localCompletionHandler)(UIBackgroundFetchResult) = ^void (UIBackgroundFetchResult fetchResult){
+                                      latestFetchResult = fetchResult;
+                                      dispatch_group_leave(callbackGroup);
+                                    };
+                                    
                                     NSInvocation *invocation = [GULAppDelegateSwizzler
                                         appDelegateInvocationForSelector:methodSelector];
                                     [invocation setTarget:interceptor];
                                     [invocation setSelector:methodSelector];
                                     [invocation setArgument:(void *)(&application) atIndex:2];
                                     [invocation setArgument:(void *)(&userInfo) atIndex:3];
-                                    [invocation setArgument:(void *)(&completionHandler) atIndex:4];
+                                    [invocation setArgument:(void *)(&localCompletionHandler) atIndex:4];
                                     [invocation invoke];
                                   }];
   // Call the real implementation if the real App Delegate has any.
--- a/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
+++ b/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
@@ -884,20 +884,19 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
           [didReceiveRemoteNotificationWithCompletionIMPPointer pointerValue];
 
   dispatch_group_t __block callbackGroup = dispatch_group_create();
-  UIBackgroundFetchResult __block latestFetchResult = UIBackgroundFetchResultNoData;
-  dispatch_group_notify(callbackGroup, dispatch_get_main_queue(), ^() {
-    completionHandler(latestFetchResult);
-  });
+  NSMutableArray<NSNumber *> *__block fetchResults = [NSMutableArray array];
+
+  void (^localCompletionHandler)(UIBackgroundFetchResult) =
+      ^void(UIBackgroundFetchResult fetchResult) {
+        [fetchResults addObject:[NSNumber numberWithInt:(int)fetchResult]];
+        dispatch_group_leave(callbackGroup);
+      };
+
   // Notify interceptors.
   [GULAppDelegateSwizzler
       notifyInterceptorsWithMethodSelector:methodSelector
                                   callback:^(id<GULApplicationDelegate> interceptor) {
                                     dispatch_group_enter(callbackGroup);
-                                    void (^localCompletionHandler)(UIBackgroundFetchResult) =
-                                        ^void(UIBackgroundFetchResult fetchResult) {
-                                          latestFetchResult = fetchResult;
-                                          dispatch_group_leave(callbackGroup);
-                                        };
 
                                     NSInvocation *invocation = [GULAppDelegateSwizzler
                                         appDelegateInvocationForSelector:methodSelector];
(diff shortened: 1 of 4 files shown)


mikeger:
@googlebot I signed it!

mikeger:
I am afraid there are some CI errors in GitHub Actions, seems like something unrelated to my change. Should I rebase it on your recent `release-xxx` branch?

paulb777:
@mikeger Thanks for the PR! The CI failure is a result of GitHub Actions changing its default Xcode from 11.7 to 12.0.1 yesterday and being addressed in #6865.  You can disregard for now.  master is the correct target for this PR.

Assigning for review to @maksymmalyhin for the GoogleUtilities changes and @chliangGoogle for the FirebaseMessaging changes.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -882,17 +883,30 @@ - (void)application:(GULApplication *)application
       didReceiveRemoteNotificationWithCompletionIMP =
           [didReceiveRemoteNotificationWithCompletionIMPPointer pointerValue];
 
+  dispatch_group_t __block callbackGroup = dispatch_group_create();
+  UIBackgroundFetchResult __block latestFetchResult = UIBackgroundFetchResultNoData;
+  dispatch_group_notify(callbackGroup, dispatch_get_main_queue(), ^() {
```
maksymmalyhin, line 888: It looks like calling `dispatch_group_notify` on an empty group will lead to scheduling the block straight away. I think it should be moved to after line 916.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -882,17 +883,30 @@ - (void)application:(GULApplication *)application
...
                                     [invocation setTarget:interceptor];
                                     [invocation setSelector:methodSelector];
                                     [invocation setArgument:(void *)(&application) atIndex:2];
                                     [invocation setArgument:(void *)(&userInfo) atIndex:3];
-                                    [invocation setArgument:(void *)(&completionHandler) atIndex:4];
+                                    [invocation setArgument:(void *)(&localCompletionHandler)
+                                                    atIndex:4];
                                     [invocation invoke];
                                   }];
   // Call the real implementation if the real App Delegate has any.
```
maksymmalyhin, line 912: The current implementation doesn't add the originally implemented method to the group. I think we should do the same for it: add enter/leave group calls and wrap the completion. Potentially we may use a single completion handler wrapper to be passed to all interceptors and the original implementations.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -882,17 +883,30 @@ - (void)application:(GULApplication *)application
...
+    completionHandler(latestFetchResult);
+  });
   // Notify interceptors.
   [GULAppDelegateSwizzler
       notifyInterceptorsWithMethodSelector:methodSelector
                                   callback:^(id<GULApplicationDelegate> interceptor) {
+                                    dispatch_group_enter(callbackGroup);
+                                    void (^localCompletionHandler)(UIBackgroundFetchResult) =
+                                        ^void(UIBackgroundFetchResult fetchResult) {
+                                          latestFetchResult = fetchResult;
```
maksymmalyhin, line 898: I think we can improve combining the fetch results. Perhaps if there is at least on `UIBackgroundFetchResultNewData` the combined result should be also `UIBackgroundFetchResultNewData`. Also I would pass `UIBackgroundFetchResultFailed` only if all calls failed.

maksymmalyhin (changes requested):
Great idea, thank you for your contribution!

I added a couple comments to adjust the logic a bit. Feel free to ask clarification questions as needed. 

Also we will need to add unit tests for these changes before merging. Please let us know if you help with it.

mikeger:
@maksymmalyhin thanks for the review, all valid comments.

paulb777:
#6865 has merged so rebasing or merging master should now enable a green CI run.

AlexPritchin:
Hello. Are there any updates here? I'm facing exact same issue in Firebase 6.18, 6.31.1, 6.34.0 and even 7.0.0. However I'm not using Braze SDK. According to Crashlytics 80% of such crashes occur on iOS 14.1. Others on other versions of iOS 14. So I'm waiting for this fix in next release version of Firebase. BTW the project I'm supporting now is written on Objective C.

maksymmalyhin:
@mikeger Will you have time to address the comments? We may finish your PR if not. We will make sure your contribution will be visible in Firebase repo. Please let us know.

mikeger:
@maksymmalyhin hey, I would address the comments today. I got stuck since the tests (`./test.sh`) were not running for me (build failed), I will retry rebased on the current master.

maksymmalyhin:
@mikeger Ah, `./test.sh` is an obsolete script that should've been removed. Sorry about that. Currently we mostly use the following ways to run tests:
1. Follow the [development guide](https://github.com/firebase/firebase-ios-sdk#development) and run tests from the generated project. This is the most convenient way for active development.
2. Running `pod lib lint` for a podspec. See CI for examples, e.g. for [FirebaseMessaging](https://github.com/firebase/firebase-ios-sdk/blob/2d39b6c6f6b19c5bf8207d626a581e5d8ac0459e/.github/workflows/messaging.yml#L54) or [GoogleUtilities](https://github.com/firebase/firebase-ios-sdk/blob/2d39b6c6f6b19c5bf8207d626a581e5d8ac0459e/.travis.yml#L176). This is useful to debug CI issues locally.

mikeger:
@maksymmalyhin thanks so much for the testing tip, this allowed me to progress with the tests. Please re-review the changes, I updated it based on your comments.

GoogleUtilities/Tests/Unit/Swizzler/GULAppDelegateSwizzlerTest.m
```diff
@@ -1087,7 +1087,135 @@ - (void)testApplicationDidReceiveRemoteNotificationWithCompletionIsInvokedOnInte
...
+  [GULAppDelegateSwizzler proxyOriginalDelegateIncludingAPNSMethods];
+
+  [GULAppDelegateSwizzler registerAppDelegateInterceptor:interceptor];
+
+  [testAppDelegate application:application
+      didReceiveRemoteNotification:notification
+            fetchCompletionHandler:completion];
+  testAppDelegate.remoteNotificationCompletionHandler(UIBackgroundFetchResultNewData);
+  OCMVerifyAll(interceptor);
+  [self waitForExpectations:@[ completionExpectation ] timeout:0.1];
```
maksymmalyhin, line 1218: nit: Looks like these 3 tests are mostly identical except the background fetch result. Could you please extract the test logic to a helper function like `assertApplicationDidReceiveRemoteNotificationWithCompletionCompletionIsCalledOnceWithResult:(UIBackgroundFetchResult)` to reduce code duplication.

GoogleUtilities/Tests/Unit/Swizzler/GULAppDelegateSwizzlerTest.m
```diff
@@ -1087,7 +1087,135 @@ - (void)testApplicationDidReceiveRemoteNotificationWithCompletionIsInvokedOnInte
...
+  [testAppDelegate application:application
+      didReceiveRemoteNotification:notification
+            fetchCompletionHandler:completion];
+  testAppDelegate.remoteNotificationCompletionHandler(UIBackgroundFetchResultNoData);
+  OCMVerifyAll(interceptor);
+  OCMVerifyAll(interceptor2);
+  [self waitForExpectations:@[ completionExpectation ] timeout:0.1];
+}
+
+- (void)
```
maksymmalyhin, line 1139: It would be nice to add tests for cases with different fetch results or at least add a `TODO` to add them.

maksymmalyhin (commented):
Great, thank you for addressing the comments! It mostly looks good to me. I added a couple minor comments. The comments are mostly optional and I'm fine proceeding with proceeding as is.

@chliangGoogle Could you please take a look at FCM part.

maksymmalyhin (approved):

mikeger:
@maksymmalyhin updated, thanks for the comments!

GoogleUtilities/Tests/Unit/Swizzler/GULAppDelegateSwizzlerTest.m
mikeger: _I thought it's paid by the number of lines_

maksymmalyhin (approved):
👍

maksymmalyhin (changes requested):
I think we've missed on place where we need to call the completion handler after the change: could you please add it to [auth](https://github.com/firebase/firebase-ios-sdk/blob/57123ae0d5578733e4ba2a1abb7c15893651c886/FirebaseAuth/Sources/Auth/FIRAuth.m#L1501) as well.

mikeger:
@maksymmalyhin added, to my understanding there it's sufficient to always indicate `NoData`.

chliangGoogle (approved):
LGTM on messaging

maksymmalyhin (approved):

maksymmalyhin:
All looks good to me, I'm going to merge the PR on CI green.

maksymmalyhin:
@mikeger Thank you again for your contribution! 

We will try to include the fix into 7.1.0 release.

PR #6895, opened 2020-11-04
Remove unnecessary some keywords from Package.swift

mkj-is:
When reading through the Package.swift file I noticed there is inconsistency of platform definitions and there are many unnecessary `.some` optional cases.

I propose this small improvement to unify target definitions and improve readability.


maksymmalyhin (commented):
@mkj-is Thank you for the contribution! 

The change looks good to me, but I defer final approval to @paulb777.

paulb777 (approved):
Thanks!

PR #6897, opened 2020-11-04
GDT: test app fixes

maksymmalyhin:
#no-changelog


GoogleDataTransport.podspec
```diff
@@ -55,7 +55,7 @@ Shared library for iOS SDK data transport needs.
   # Test app specs
   if ENV['GDT_DEV'] && ENV['GDT_DEV'] == '1' then
     s.app_spec 'TestApp' do |app_spec|
-      app_spec.platforms = {:ios => '8.0', :osx => '10.11', :tvos => '10.0'}
+      app_spec.platforms = {:ios => '9.0', :osx => '10.11', :tvos => '10.0'}
```
paulb777, line 58: 10.12?

paulb777 (approved):

GoogleDataTransport.podspec
maksymmalyhin: Good catch. I updated it to use version from the lib spec to avoid such mistakes in the future.

PR #6901, opened 2020-11-05
Add Result support to runTransaction Firestore method

lorenzofiamingo:
This is a strong-typed mapping of runTransaction, that makes usage of generics and Result.
The key difference is that Error is not passed with ErrorPointer.pointee but with `return .failure(myError)`
```swift
extension Firestore {
    
    func runTransaction<T>(_ updateBlock: @escaping (Transaction) -> Result<T, Error>, completion: @escaping (Result<T, Error>) -> Void) {
        runTransaction { transaction, _ in updateBlock(transaction) } completion: { value, error in
            if let result = value as? Result<T, Error> {
                completion(result)
            } else if let error = error {
                completion(.failure(error))
            } else {
                completion(.failure(NSError(domain: "FirebaseFirestoreSwift",
                                            code: -1,
                                            userInfo: [NSLocalizedDescriptionKey:
                                                        "InternalError - Return type and Error code both nil in or updateBlock return type doesn't match completion argument type"
                                            ])))
            }
        }
    }
}
```
So [the first example in docs](https://firebase.google.com/docs/firestore/manage-data/transactions):
```swift
let sfReference = db.collection("cities").document("SF")

db.runTransaction({ (transaction, errorPointer) -> Any? in
    let sfDocument: DocumentSnapshot
    do {
        try sfDocument = transaction.getDocument(sfReference)
    } catch let fetchError as NSError {
        errorPointer?.pointee = fetchError
        return nil
    }

    guard let oldPopulation = sfDocument.data()?["population"] as? Int else {
        let error = NSError(
            domain: "AppErrorDomain",
            code: -1,
            userInfo: [
                NSLocalizedDescriptionKey: "Unable to retrieve population from snapshot \(sfDocument)"
            ]
        )
        errorPointer?.pointee = error
        return nil
    }

    // Note: this could be done without a transaction
    //       by updating the population using FieldValue.increment()
    transaction.updateData(["population": oldPopulation + 1], forDocument: sfReference)
    return nil
}) { (object, error) in
    if let error = error {
        print("Transaction failed: \(error)")
    } else {
        print("Transaction successfully committed!")
    }
}
```
becomes:
```swift
let sfReference = db.collection("cities").document("SF")

db.runTransaction { transaction in
    let sfDocument: DocumentSnapshot
    do {
        try sfDocument = transaction.getDocument(sfReference)
    } catch let fetchError as NSError {
        return  .failure(fetchError)
    }

    guard let oldPopulation = sfDocument.data()?["population"] as? Int else {
        let error = NSError(
            domain: "AppErrorDomain",
            code: -1,
            userInfo: [
                NSLocalizedDescriptionKey: "Unable to retrieve population from snapshot \(sfDocument)"
            ]
        )
        return .failure(error)
    }

    // Note: this could be done without a transaction
    //       by updating the population using FieldValue.increment()
    transaction.updateData(["population": oldPopulation + 1], forDocument: sfReference)
    return .success(())
} completion: { result in
    switch result {
    case .success:
         print("Transaction successfully committed!")
    case .failure(let error):
        print("Transaction failed: \(error)")
    }
}
```
Note that in this example is not evident the advantage of using generics.
Hoping this could be useful!

### DIFF
--- a/Firestore/Swift/Source/Result/Result/Firestore+RunTransaction
+++ b/Firestore/Swift/Source/Result/Result/Firestore+RunTransaction
@@ -13,10 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
- import FirebaseFirestore
- 
- extension Firestore {
+
+#if swift(>=5.0)
+
+extension Firestore {
   /**
    * Executes the given updateBlock and then attempts to commit the changes applied within an atomic
    * transaction.
@@ -49,19 +49,36 @@
    * @param completion The block to call with the result or error of the transaction. This
    *     block will run even if the client is offline, unless the process is killed.
    */
-  func runTransaction<T>(_ updateBlock: @escaping (Transaction) -> Result<T, Error>, completion: @escaping (Result<T, Error>) -> Void) {
-    runTransaction { transaction, _ in updateBlock(transaction) } completion: { value, error in
-      if let result = value as? Result<T, Error> {
-        completion(result)
+  func runTransaction<T>(_ updateBlock: @escaping (Transaction) throws -> T, completion: @escaping (Result<T, Error>) -> Void) {
+    runTransaction { transaction, errorPointer in
+      do {
+        return try updateBlock(transaction)
+      } catch {
+        errorPointer?.pointee = error as NSError
+        return nil
+      }
+    } completion: { value, error in
+      if let value = value {
+        if let value = value as? T {
+          completion(.success(value))
+        } else {
+          completion(.failure(NSError(domain: "FirebaseFirestoreSwift",
+                                      code: -1,
+                                      userInfo: [NSLocalizedDescriptionKey:
+                                                    "InternalError - UpdateBlock return type doesn't match completion argument type"
+                                      ])))
+        }
       } else if let error = error {
         completion(.failure(error))
       } else {
         completion(.failure(NSError(domain: "FirebaseFirestoreSwift",
                                     code: -1,
                                     userInfo: [NSLocalizedDescriptionKey:
-                                                "InternalError - Return type and Error code both nil in or updateBlock return type doesn't match completion argument type"
+                                                "InternalError - Return type and Error code are both nil"
                                     ])))
       }
     }
   }
 }
+
+#endif
--- a/Firestore/Swift/Source/Result/Transaction+Result
+++ b/Firestore/Swift/Source/Result/Transaction+Result
@@ -14,8 +14,41 @@
  * limitations under the License.
  */
  
+ import FirebaseFirestore
+ 
  extension Firestore {
-    
+  /**
+   * Executes the given updateBlock and then attempts to commit the changes applied within an atomic
+   * transaction.
+   *
+   * The maximum number of writes allowed in a single transaction is 500, but note that each usage of
+   * `FieldValue.serverTimestamp()`, `FieldValue.arrayUnion()`, `FieldValue.arrayRemove()`, or
+   * `FieldValue.increment()` inside a transaction counts as an additional write.
+   *
+   * In the updateBlock, a set of reads and writes can be performed atomically using the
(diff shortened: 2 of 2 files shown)


lorenzofiamingo:
another approach would be this:
```swift
func runTransaction<T>(_ updateBlock: @escaping (Transaction) throws -> T, completion: @escaping (Result<T, Error>) -> Void) {
    runTransaction { transaction, errorPointer in
        do {
            return try updateBlock(transaction)
        } catch {
            errorPointer?.pointee = error as NSError
            return nil
        }
    } completion: { value, error in
        if let value = value {
            if let result = value as? Result<T, Error> {
                completion(result)
            } else {
                completion(.failure(NSError(domain: "FirebaseFirestoreSwift",
                                            code: -1,
                                            userInfo: [NSLocalizedDescriptionKey:
                                                        "InternalError - UpdateBlock return type doesn't match completion argument type"
                                            ])))
            }
        } else if let error = error {
            completion(.failure(error))
        } else {
            completion(.failure(NSError(domain: "FirebaseFirestoreSwift",
                                        code: -1,
                                        userInfo: [NSLocalizedDescriptionKey:
                                                    "InternalError - Return type and Error code are both nil"
                                        ])))
        }
    }
}
```

the example turns out even better:
```swift
let sfReference = db.collection("cities").document("SF")

db.runTransaction { transaction -> Void in
    let sfDocument: DocumentSnapshot
    
    try sfDocument = transaction.getDocument(sfReference)
    
    guard let oldPopulation = sfDocument.data()?["population"] as? Int else {
        throw NSError(
            domain: "AppErrorDomain",
            code: -1,
            userInfo: [
                NSLocalizedDescriptionKey: "Unable to retrieve population from snapshot \(sfDocument)"
            ]
        )
    }
    
    // Note: this could be done without a transaction
    //       by updating the population using FieldValue.increment()
    transaction.updateData(["population": oldPopulation + 1], forDocument: sfReference)
} completion: { result in
    switch result {
    case .success:
        print("Transaction successfully committed!")
    case .failure(let error):
        print("Transaction failed: \(error)")
    }
}
```

wilhuff:
Thanks for the PR! We've been meaning to update these APIs for `Result` but haven't had time to tackle this yet.

The second approach yields much better code, since errors can just pass through without catching them, so I think that's the approach to follow. This mirrors the approach we use in Kotlin and Java in the Android SDK.

A few notes:
  * You need to name your new file with a `.swift` extension to get it incorporated into the build. Also the base part of the name should match the class we're extending, so maybe something like `Firestore+RunTransaction.swift` might be a good filename?
  * We still support back to Swift 4.1, and `Result` is part of Swift 5.0, so this feature needs to be wrapped in an `#if swift(>=5.0)` directive.
  * Once you do, the linter and formatter will kick in and validate the PR. You can run its checks locally with `scripts/check.sh` (after performing setup steps in the [README.md](https://github.com/firebase/firebase-ios-sdk/blob/master/README.md#code-formatting).
  * To expedite adoption of this feature, we'd need a test.
      * Extend `FSTIntegrationTestCase` like the [`CodableIntegrationTests`](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Swift/Tests/Integration/CodableIntegrationTests.swift#L21). This gives you a pre-setup SDK that connects to a local emulator for an arbitrary project.
      * You can run that emulator by running `scripts/run_firestore_emulator.sh`.
  * From our side we can't ship a feature like this without going through an API review process. This process usually takes about two weeks including a write-up and reviews, but since we're nearing the holidays this could run longer.

Let me know how you'd like to proceed and if you're willing to push this PR to completion. Even if you can't, this is still a good idea, but it will take us longer to implement this because of other higher priority work we're currently tackling.

lorenzofiamingo:
In one of my projects, all firestore, storage, and auth methods are mapped to Result (like StorageSwift) and then to Combine. I could easily make another PR with those ones, after fixing this.

lorenzofiamingo:
@wilhuff I'm closing this in favor of #6939.

PR #6907, opened 2020-11-05
User defaults

manjanac:

### Discussion

### Testing

### API Changes

    can discuss it together.


FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,165 @@
...
+  let defaultValue: Value
+  let storage: UserDefaults = .firebaseMLDefaults
+
+  var wrappedValue: Value {
+    get {
+      let value = storage.value(forKey: key) as? Value
+      return value ?? defaultValue
+    }
+    set {
+      guard let optional = newValue as Optional?, optional != nil else {
```
manjanac, line 137: What I'm checking here is if `newValue` is nil - is this the right way to do that?

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderTests.swift
```diff
@@ -0,0 +1,90 @@
...
+    FirebaseApp.configure()
+    let testApp = FirebaseApp.app()!
+    let testModelName = "user-defaults-test-model"
+    let modelInfoRetriever = ModelInfoRetriever(
+      app: testApp,
+      modelName: testModelName,
+      defaults: .getTestInstance()
+    )
+    modelInfoRetriever.modelInfo = ModelInfo(app: testApp, name: testModelName)
+    // XCTAssertEqual(modelInfoRetriever.modelInfo?.downloadURL, "")
```
manjanac, line 31: I figured this line would work, given that `getTestInstance()` wipes out user defaults, but it fails. What am I missing?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+    if let info = modelInfo, info.hash.count > 0 {
+      request.setValue(info.hash, forHTTPHeaderField: ModelInfoRetriever.hashMatchHTTPHeader)
+    }
+    return request
+  }
+
+  /// Get model info from server.
+  func downloadModelInfo(completion: @escaping (DownloadError?) -> Void) {
+    /// Get FIS token.
+    installations.authToken { [weak self] tokenResult, error in
```
manjanac, line 133: @maksymmalyhin

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+      guard let result = tokenResult
+      else { completion(.internalError(description: ModelInfoRetriever.tokenErrorDescription))
+        return
+      }
+      /// Get model info fetch URL with appropriate HTTP headers.
+      guard let request = self?.getModelInfoFetchURLRequest(token: result.authToken)
```
maksymmalyhin, line 139: Would it make sense to have a separate `guard` to unwrap `self` at the beginning of the handler?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+
+          guard httpResponse.statusCode == 200 || httpResponse.statusCode == 304 else {
+            // TODO: Improve http status code error handling
+            completion(.notFound)
+            return
+          }
+
+          /// Local model not modified.
+          if httpResponse.statusCode == 304 {
+            return
```
maksymmalyhin, line 164: I wonder why completion is not called in this case? Could you add a comment please?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+          }
+
+          guard let data = data else {
+            completion(.internalError(description: ModelInfoRetriever
+                .invalidHTTPResponseErrorDescription))
+            return
+          }
+
+          self?.saveModelInfo(data: data, modelHash: modelHash)
```
maksymmalyhin, line 180: Should we unwrap `self` at the beginning of the handler?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+        return
+      }
+      /// Get model info fetch URL with appropriate HTTP headers.
+      guard let request = self?.getModelInfoFetchURLRequest(token: result.authToken)
+      else {
+        completion(.internalError(description: ModelInfoRetriever.modelFetchURLErrorDescription))
+        return
+      }
+      /// Download model info.
+      let dataTask = URLSession.shared.dataTask(with: request) { [weak self]
```
maksymmalyhin, line 145: nit: will it be more readable if we move sending the request to a separate method?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+            return
+          }
+
+          /// Local model not modified.
+          if httpResponse.statusCode == 304 {
+            return
+          }
+
+          guard let modelHash = httpResponse
+            .allHeaderFields[ModelInfoRetriever.etagHTTPHeader] as? String else {
```
maksymmalyhin, line 168: Standard caching headers like ETag are usually handled transparently by NSURLSession. I wonder what's the reason of additional manual handling?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+        if let downloadError = error {
+          completion(.internalError(description: downloadError.localizedDescription))
+        } else {
+          guard let httpResponse = response as? HTTPURLResponse else {
+            completion(.internalError(description: ModelInfoRetriever
+                .invalidHTTPResponseErrorDescription))
+            return
+          }
+
+          guard httpResponse.statusCode == 200 || httpResponse.statusCode == 304 else {
```
maksymmalyhin, line 156: nit: It may be better to express this using `switch` with pattern matching, e.g.:

```
switch httpResponse.statusCode {
case 100..<300:
  // Handle success codes
  break
case 304:
  return

default:
...
}
```

In this case it will be easier to spot handled and unhandled cases.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
maksymmalyhin: I wonder if something like the following will work:
```
if let newValue = newValue else { 
  storage.setValue(newValue, forKey: key)
} else {
  storage.removeObject(forKey: key)
}
```

ryanwilson: I don't think this will work. I was playing around in a playground and remember seeing something like this before and came up with a different (somewhat messy?) solution:

```
private protocol OptionalProtocol: ExpressibleByNilLiteral {
  var isNil: Bool { get }
}

extension Optional: OptionalProtocol {
  public var isNil: Bool {
    switch self {
    case .some(_): return false
    case .none: return true
    }
  }
}
```

This is just saying that there's a protocol that `nil` "conforms" to and Optional does as well, then the check can be against the protocol:

```
if let opt = newValue as? OptionalProtocol, opt.isNil {
  // Value is nil, remove it.
} else {
  // Value is not nil.
}
```

Does that make sense? It's not the nicest but it works as far as I can tell :P

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+    // TODO: Save model info to user defaults
+    modelInfo?.hash = modelHash
+  }
+}
+
+/// Named user defaults for FirebaseML.
+extension UserDefaults {
+  static var firebaseMLDefaults: UserDefaults {
+    let suiteName = "com.google.firebase.ml"
+    let defaults = UserDefaults(suiteName: suiteName)!
```
maksymmalyhin, line 198: Please avoid using force unwrap in the production code unless it is 100% safe or required as it potentially leads to unexpected obfuscated crashes.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+    return defaults
+  }
+
+  /// For testing: returns a new cleared instance of user defaults.
+  static func getTestInstance() -> UserDefaults {
```
maksymmalyhin, line 203: I would try to avoid adding test code to production version. Can it be defined in tests instead?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: True. I've added a TODO for myself to do this in a future PR, since this one is primarily about user defaults and FIS.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: I looked into this and it seems like UserDefaults will only fail when you pass it a reserved (therefore invalid) suiteName. So technically this should work 100% of the time, but I've added a TODO here to look into this more and add a test in a future PR.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+    let suiteName = "com.google.firebase.ml.test"
+    let defaults = UserDefaults(suiteName: suiteName)!
+    defaults.removePersistentDomain(forName: suiteName)
+    return defaults
+  }
+}
+
+/// Property initializer for user defaults. Value is always read from or written to a named user defaults store.
+@propertyWrapper struct UserDefaultsBacked<Value> {
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,251 @@
...
+    }
+    return request
+  }
+
+  /// Get model info from server.
+  func downloadModelInfo(completion: @escaping (DownloadError?) -> Void) {
+    /// Get FIS token.
+    installations.authToken { [weak self] tokenResult, error in
+      guard let result = tokenResult
+      else { completion(.internalError(description: ModelInfoRetriever.tokenErrorDescription))
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: My bad, I fixed it.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: I use ETag to extract the model hash returned by the server for the model object.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Makes sense, I updated the code to look more like this.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,207 @@
...
+      hash: info.hash
+    )
+    return model
+  }
+}
+
+/// Extension to handle fetching model info from server.
+extension ModelInfoRetriever {
+  /// HTTP request headers.
+  static let fisTokenHTTPHeader: String = "x-goog-firebase-installations-auth"
```
maksymmalyhin, line 93: nit: the `String` type declaration is redundant when assigned with the literal here and below. It should be just:
```
static let fisTokenHTTPHeader = "x-goog-firebase-installations-auth"
```

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,207 @@
...
+  static let invalidHTTPResponseErrorDescription: String =
+    "Could not get a valid HTTP response from server."
+
+  /// Construct model fetch base URL.
+  var modelInfoFetchURL: URL {
+    var components = URLComponents()
+    components.scheme = "https"
+    components.host = "firebaseml.googleapis.com"
+    components.path = "/v1beta2/projects/\(projectID)/models/\(modelName):download"
+    return components.url!
```
maksymmalyhin, line 113: No force unwrap should be used in production code.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,207 @@
...
+      else {
+        completion(.internalError(description: ModelInfoRetriever.tokenErrorDescription))
+        return
+      }
+      /// Get model info fetch URL with appropriate HTTP headers.
+      let request = self.getModelInfoFetchURLRequest(token: result.authToken)
+
+      /// Download model info.
+      // TODO: Consider moving request to a separate method
+      let dataTask = URLSession.shared.dataTask(with: request) { [weak self]
```
maksymmalyhin, line 148: If I am not mistaken `URLSession.shared` caches request and response into an unencrypted database in the app sandbox, which for some cases may be considered a security vulnerability when request contains data like FIS token API key, etc. You may find more details in b/158644295. 

Could you please add a TODO to reconsider using the shared URL session for the follow up PRs.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
maksymmalyhin: From my experience, we should rely only on the public API definition as the implementation details may change without a notice, so I'm pretty convinced we should handle nullability gracefully here.

maksymmalyhin (approved):
Mostly LGTM for an intermediate PR, but a couple things to be addressed here or later on.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -0,0 +1,207 @@
...
+/// Model info object with details about pending or downloaded model.
+struct ModelInfo {
+  /// Model name.
+  var name: String
+
+  /// User defaults associated with model.
+  var defaults: UserDefaults
+
+  /// Download URL for the model file, as returned by server.
+  @UserDefaultsBacked var downloadURL: String
```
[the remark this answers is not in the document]
manjanac: Added a TODO, and for now ModelInfo is a class.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: I've added a TODO to handle this in a future PR.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Sounds good. I've added a TODO to handle this in a future PR.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done.

Package.swift
```diff
@@ -560,6 +560,53 @@ let package = Package(
       ]
     ),
 
+    .target(
```
paulb777, line 563: This looks problematic and looks to be causing CI failure

PR #6912, opened 2020-11-05
Fix infinite loop exposed in integration tests

paulb777:
Fix for integration test failures in #5916's CI

#no-changelog


morganchen12 (approved):
cc @schmidt-sebastian

PR #6941, opened 2020-11-10
Remove onboarding calls from SDK

elenadoty:


samedson (commented):
So this goes a little too deep. We need to keep Settings around to give ourselves tools to help respond to incidents and such, but we can get rid of the onboarding operation.

In addition, we can remove this method: https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/Crashlytics/Models/FIRCLSSettings.m#L306

And for all cases, assume it is true, and remove any codepaths that depend on it being false.

PR #6945, opened 2020-11-10
replace set-env with github env file workflow

vimanyu:
Suggested workaround for security vulnerability listed at,
https://github.com/actions/toolkit/security/advisories/GHSA-mfwh-5m23-j46w

The following error can be spotted in the github workflow logs (example: https://github.com/firebase/firebase-ios-sdk/runs/1382322482?check_suite_focus=true, remote_config(iOS)->"Generate Access Tokens.."),
```
The `set-env` command is deprecated and will be disabled on November 16th. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```


paulb777 (approved):
Thanks!

PR #6955, opened 2020-11-11
Revert changes in Crashlytics version - as it's causing an issue with the version associated with the crash upload

tejasd:
#no-changelog

b/172929809


paulb777 (approved):
Does this impact 3P?  Do we need a patch for 7.1.0?

tejasd:
It is very likely affecting 3P as well - but I haven't tested it. If you don't mind, I'll merge it in right now and copybara it (so that it unblocks the 1P customer) - and I will confirm the 3P situation so that we can release patch.

samedson:
@paulb777 so far I'm not seeing this affect 3Ps

tejasd:
@paulb777 A patch shouldn't be needed - there's a backend change that'll fix it + revert of this revert + fixes - https://github.com/firebase/firebase-ios-sdk/pull/6958

Issue #6786, opened 2020-10-20
Characters not allowed in a properly URL when execute FIRDLDictionaryFromQuery function  


cjimenezpacho:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.0.1
  * Firebase SDK version: 6.34.0
  * Firebase Component: FirebaseDynamicLinks
  * Component version: 6.34.0
  * Installation method: `CocoaPods`

### [REQUIRED] Step 2: Describe the problem

If the parameter `link` contains diacritical characters deeplink cannot be processed correctly.

#### Steps to reproduce:

Using this sample link: 

https://XXXX.app.goo.gl/?link=https://DEEPLINK?latitude%3D41.404891232066234%26longitude%3D-3.2052324507419998%26last-name%3Dsjim%C3%A9nez

I use this method to process deeplink:

```
[[FIRDynamicLinks dynamicLinks] handleUniversalLink:url
                                             completion:^(FIRDynamicLink *_Nullable dynamicLink, NSError *_Nullable error) {
                                             }];

```

If I debug into `handleUniversalLink: completion` I can see at 414 line breakpoint:

![Captura de pantalla 2020-10-20 a las 16 57 14](https://user-images.githubusercontent.com/2428271/96604394-5a86c000-12f5-11eb-975f-30e0c904d79f.png)

following values:

- `url`: https://XXX/?link=https://ZZZZZZZ?latitude%3D41.404891232066234%26longitude%3D-3.2052324507419998%26last-name%3Dsjim%C3%A9nez
- `parameters`: 
![Captura de pantalla 2020-10-20 a las 16 59 03](https://user-images.githubusercontent.com/2428271/96604794-c8cb8280-12f5-11eb-84c9-1d9e3677aac0.png)

At this moment,`link`  has already removed percentage encoding (see `FIRDLDictionaryFromQuery` from `FDLUtilities.m`)
You can see an almost valid URL. I say almost because it has `\` character.
So, after these two lines:

```
NSString *urlString = parameters[kFIRDLParameterLink];
NSURL *deepLinkURL = [NSURL URLWithString:urlString];

```

`deepLinkURL` is `nil` because `URLWithString`, according to doc:  

> This method expects URLString to contain only characters that are allowed in a properly formed URL. All other characters must be properly percent escaped. The URL string with which to initialize the NSURL object. Must be a URL that conforms to RFC 2396. This method parses URLString according to RFCs 1738 and 1808.

[...]

PR #6956, opened 2020-11-11
Escaping  url string from dynamic link parameter (issue 6786)

cjimenezpacho:
More info here: https://github.com/firebase/firebase-ios-sdk/issues/6786

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
@@ -404,7 +404,8 @@ static const NSInteger FIRErrorCodeDurableDeepLinkFailed = -119;
       if (parameters[kFIRDLParameterLink]) {
         FIRDynamicLink *dynamicLink = [[FIRDynamicLink alloc] init];
         NSString *urlString = parameters[kFIRDLParameterLink];
-        NSURL *deepLinkURL = [NSURL URLWithString:urlString];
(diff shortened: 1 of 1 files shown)


paulb777:
@cjimenezpacho Thanks for the PR! Please resolve the unit test failure and see [here](https://github.com/firebase/firebase-ios-sdk#code-formatting) to run clang-format to fix the formatting.

86Whiskey (approved):

eldhosembabu:
This change looks good to me. Could you please add some tests for the same?

cjimenezpacho:
@eldhosembabu yes, sorry, I have been very busy, as soon as I have time I will add tests. Thank you for the feedback :)

eldhosembabu:
> you

thanks

paulb777:
@eldhosembabu What is the next step on this PR? Can your team add the tests and merge?

eldhosembabu:
> @eldhosembabu What is the next step on this PR? Can your team add the tests and merge?

Hi @paulb777 

I’ve tried creating a new PR with the changes copied and unit tests, but I could see chance for backward compatibility issues here since we are doing string escaping using encoding for the deep link url which is getting passed back to the developer. So the existing developers who expects the regular url as of now will be given encoded url which might break their code.

I think its better not to do this fix at SDK level (until we find a better solution), instead as a workaround the developer can encode the URL at the application level before passing it into the SDK as mentioned here: https://github.com/firebase/firebase-ios-sdk/issues/6786#issuecomment-1301884080

paulb777:
thanks for investigating @eldhosembabu. I'll close based on your conclusions.

PR #6958, opened 2020-11-11
Bring back the changes of reading the version from FIRCore with fixes + unit test.

tejasd:
#no-changelog


Crashlytics/generate_project.sh
```diff
@@ -19,4 +19,4 @@ readonly DIR="$( git rev-parse --show-toplevel )"
 
 "$DIR/Crashlytics/ProtoSupport/generate_crashlytics_protos.sh" || echo "Something went wrong generating protos.";
 
-pod gen "${DIR}/FirebaseCrashlytics.podspec" --auto-open --gen-directory="${DIR}/gen" --local-sources="${DIR}" --platforms=ios,macos,tvos,watchos --clean
+pod gen "${DIR}/FirebaseCrashlytics.podspec" --auto-open --gen-directory="${DIR}/gen" --local-sources="${DIR}" --platforms=ios,macos,tvos --clean
```
samedson, line 22: Can you revert this change?

samedson (approved):
Just one thing 👍

Crashlytics/generate_project.sh
tejasd: Reverted! I need this for generate_project.sh to run locally - I haven't looked deeper into any setup I might've missed, but is that a known issue?

samedson: Yeah I think there's some weird interaction with the podspec

PR #6993, opened 2020-11-18
Download model info

manjanac:
Implementation and tests for getting model info (size, hash, download url) from server and saving it to user defaults.


maksymmalyhin:
@manjanac Could you please update the PR description to reflect the PR content.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -194,17 +179,12 @@ extension ModelInfoRetriever {
 
   /// Save model info to user defaults.
   func saveModelInfo(data: Data, modelHash: String) {
-    // TODO: Save model info to user defaults
+    let decoder = JSONDecoder()
+    guard let modelInfoJSON = try? decoder.decode(ModelInfoResponse.self, from: data)
+    else { return }
+    modelInfo = ModelInfo(app: app, name: modelName)
```
maksymmalyhin, line 185: nit: I would create a local non-option variable to avoid extra optional chains `?` below:
```
let modelInfo = ModelInfo(app: app, name: modelName)
...
self.modelInfo = modelInfo
```

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderTests.swift
```diff
@@ -87,23 +56,51 @@ final class ModelDownloaderTests: XCTestCase {
...
-      projectID: Constants.Options.projectID,
       modelName: testModelName
     )
     let expectation = self.expectation(description: "Wait for model info to download.")
     modelInfoRetriever.downloadModelInfo(completion: { error in
-      guard let downloadError = error else { return }
-      XCTAssertEqual(downloadError, .notFound)
-      print("ERROR: Model not found on server.")
+      XCTAssertNil(error)
+      XCTAssertNotNil(modelInfoRetriever.modelInfo)
```
maksymmalyhin, line 70: nit: You may use [XCTUnwrap](https://developer.apple.com/documentation/xctest/3380195-xctunwrap) instead like e.g. [here](https://github.com/firebase/firebase-ios-sdk/blob/f198987a1aa147a265248d0558722d3f1868624d/FirebaseCore/Tests/SwiftUnit/FirebaseAppTests.swift#L48) to test for `nil` and avoid extra unwraps below. Don't forget to add `throws` to the test method declaration like [here](https://github.com/firebase/firebase-ios-sdk/blob/f198987a1aa147a265248d0558722d3f1868624d/FirebaseCore/Tests/SwiftUnit/FirebaseAppTests.swift#L38).

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderTests.swift
```diff
@@ -87,23 +56,51 @@ final class ModelDownloaderTests: XCTestCase {
     XCTAssertEqual(modelInfoRetriever.modelInfo?.path, nil)
   }
 
+  /// Unit test to download model info - makes an actual network call.
```
maksymmalyhin, line 59: We try to separate tests that require secrets like keys or .plist from pure tests that can be run right after the repo checkout to improve dev experience. You can use [FIS](https://github.com/firebase/firebase-ios-sdk/blob/0842ea46153e1926f2f071d1f317d60eae9cba59/FirebaseInstallations.podspec#L67) or [GDT](https://github.com/firebase/firebase-ios-sdk/blob/0842ea46153e1926f2f071d1f317d60eae9cba59/GoogleDataTransport.podspec#L102) as an example. 
It can be addressed in the following PRs.

maksymmalyhin (approved):
LGTM, a couple minor comments.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderTests.swift
manjanac: I couldn't get `XCTUnwrap` to work with the completion handler, so I've used `guard let` instead. However, the error handling in general is worth revisiting in the future.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderTests.swift
manjanac: The examples are super helpful! I've added TODOs to break these out into integration tests and create pure unit tests.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -129,82 +108,92 @@ extension ModelInfoRetriever {
...
+  }
+
+  /// Get model info from server.
+  func downloadModelInfo(completion: @escaping (DownloadError?) -> Void) {
+    getAuthToken { result in
+      switch result {
+      case let .success(authToken):
+        /// Get model info fetch URL with appropriate HTTP headers.
+        let request = self.getModelInfoFetchURLRequest(token: authToken)
+        let session = URLSession(configuration: .ephemeral)
```
maksymmalyhin, line 131: A point to consider in the future. Ephemeral session doesn't handle ETag headers, so the model will be downloaded each time even if a local copy has already been downloaded. To optimize it we should either send Etag and handle the response ourself (like Remote Config does) or request consultation with security experts to double check if caching FIS token is actually an issue (if not we may put the shared session back). I think it is worth adding a TODO to consider later.

maksymmalyhin (approved):
LGTM. One note to consider in the future.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Great point! I've manually set the caching policy for the request for now. However, I have also added a TODO so we can consult with the experts on what the right approach is.

maksymmalyhin (approved):
LGTM, thank you for addressing the comments!

PR #7005, opened 2020-11-20
Googletest - Xcode integration improvements

TheKuko:
I added case to support tests that are skipped in google test so Xcode does not show them as fails.

I also removed use of deprecated unarchiveObjectWithFile method, however this needs to be tested. 
scripts/check.sh did not show any problems, but I do not know if edited code is being tested anywhere and I was not able to run any google tests which would use the edited code flow (in Xcode GUI, I specified unit tests to run, disabled selected tests, and so on, but each call of the edited method failed because of empty 'filePath' in both code versions).

### DIFF
--- a/Firestore/core/test/unit/FSTGoogleTestTests.mm
+++ b/Firestore/core/test/unit/FSTGoogleTestTests.mm
@@ -84,14 +84,18 @@ NSSet<NSString*>* _Nullable LoadXCTestConfigurationTestsToRun() {
 
 #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 || \
     __MAC_OS_X_VERSION_MAX_ALLOWED >= 101300
-  NSError *error;
-  NSData *data = [NSData dataWithContentsOfFile:filePath options:kNilOptions error:&error];
+  NSError* error;
+  NSData* data = [NSData dataWithContentsOfFile:filePath
+                                        options:kNilOptions
+                                          error:&error];
   if (!data) {
-      NSLog(@"Failed to fill data with contents of file. %@", error);
-      return nil;
+    NSLog(@"Failed to fill data with contents of file. %@", error);
+    return nil;
   }
 
-  id config = [NSKeyedUnarchiver unarchivedObjectOfClass:NSObject.class fromData:data error:&error];
+  id config = [NSKeyedUnarchiver unarchivedObjectOfClass:NSObject.class
+                                                fromData:data
+                                                   error:&error];
 #else
   id config = [NSKeyedUnarchiver unarchiveObjectWithFile:filePath];
 #endif
@@ -227,8 +231,7 @@ void XCTestMethod(XCTestCase* self, SEL _cmd) {
     // Let XCode know that the test ran and succeeded.
     XCTAssertTrue(true);
     return;
-  }
-  else if (result->Skipped()) {
+  } else if (result->Skipped()) {
     // Let XCode know that the test was skipped.
     XCTSkip();
   }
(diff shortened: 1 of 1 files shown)


Firestore/core/test/unit/FSTGoogleTestTests.mm
```diff
@@ -82,10 +82,27 @@ @interface GoogleTests : XCTestCase
     return nil;
   }
 
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 || \
+    __MAC_OS_X_VERSION_MAX_ALLOWED >= 101300
```
morganchen12, line 86: Looks like you'll need to check the iOS SDK version that the test target is being compiled against. 

https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/744997387

paulb777:
@TheKuko Thanks for the PR. It looks like there is a problem with the macos build. PTAL

Also, the travis Xcode 10 build doesn't find `XCSkip`, but we're going to turn those tests off soon since we know longer support Xcode 10.

Firestore/core/test/unit/FSTGoogleTestTests.mm
wilhuff, line 86: Since we no longer support iOS 8, could this can be phrased as an `@available` test?

Also, this needs a case for tvOS 11.

Firestore/core/test/unit/FSTGoogleTestTests.mm
```diff
@@ -214,6 +231,9 @@ void XCTestMethod(XCTestCase* self, SEL _cmd) {
     // Let XCode know that the test ran and succeeded.
     XCTAssertTrue(true);
     return;
+  } else if (result->Skipped()) {
```
wilhuff, line 234: Unlike the above, this is dependent on the version of Xcode used, so unfortunately, it does need the `#if` on `MAX_ALLOWED` properties.

`XCTSkip` was added on Xcode 11.4, the versions should be (based on https://developer.apple.com/documentation/xcode-release-notes/xcode-11_4-release-notes):

```
__IPHONE_OS_VERSION_MAX_ALLOWED >= 130400 ||
__TV_OS_VERSION_MAX_ALLOWED >= 130400 ||
__MAC_OS_X_VERSION_MAX_ALLOWED >= 101504
```

TheKuko: Thank you, I added it.
For some reason, ___MAC_OS_X_VERSION_MAX_ALLOWED_ is on my machine (macOS 11.0 Beta (20A5395g), Xcode 12.1 (12A7403)) defined to ___MAC_10_15_ so I cannot use these guards. However, Availability.h defines ___MAC_10_15_1_ and ___MAC_10_15_4_ as well, and I am building with 10.15 SDK so I do not know where the problem is.

Today I'm going to update the OS so I'll see if it changes anything.

Firestore/core/test/unit/FSTGoogleTestTests.mm
TheKuko: Thank you, I added it and it works locally.

Firestore/core/test/unit/FSTGoogleTestTests.mm
TheKuko: With update to `macOS 11.1 Beta (20C5048l)` and `Xcode 12.2 (12B45b)` the SDK was also updated, so now my `__MAC_OS_X_VERSION_MAX_ALLOWED` is defined to `__MAC_11_0`.

wilhuff (approved):
LGTM

Apologies for the delay over the holiday.

PR #7017, opened 2020-11-23
Adjust check_imports.swift for test files

maksymmalyhin:
- allow `<>` imports in test files

#no-changelog


paulb777 (approved):

wilhuff:
It would be helpful in a PR like this to describe why you're making this change. On its surface this seems like a regression: the rules for what's allowed where are now more complex without any obvious benefit.

maksymmalyhin:
> It would be helpful in a PR like this to describe why you're making this change. On its surface this seems like a regression: the rules for what's allowed where are now more complex without any obvious benefit.

@wilhuff I tried to explain the case in the description. The change was transferred from the private repo where the check blocked a PR. Please let me know if you need more details or have any ideas.

wilhuff:
Description LGTM and makes sense. Thanks!

PR #7018, opened 2020-11-24
Skip loading invalid targets

schmidt-sebastian:
This addresses a crash during LRU by simply ignoring targets that cannot be deserialized. We could possibly be smarter about this as we only need the Target ID and the LastListenSequenceNumber to determine what targets to delete, but this for now is the more targeted fix. 

Fix: https://github.com/firebase/firebase-ios-sdk/issues/6721

### DIFF
--- a/Firestore/core/src/local/leveldb_target_cache.cc
+++ b/Firestore/core/src/local/leveldb_target_cache.cc
@@ -195,7 +195,21 @@ void LevelDbTargetCache::EnumerateTargets(const TargetCallback& callback) {
   it->Seek(target_prefix);
   for (; it->Valid() && absl::StartsWith(it->key(), target_prefix);
        it->Next()) {
-    TargetData target = DecodeTarget(it->value());
+    StringReader reader{it->value()};
+    auto message = Message<firestore_client_Target>::TryParse(&reader);
+    auto target = serializer_->DecodeTargetData(&reader, *message);
+
+    if (!reader.ok()) {
+      // In https://github.com/firebase/firebase-ios-sdk/issues/6721, a customer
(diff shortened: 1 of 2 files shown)


Firestore/core/src/local/leveldb_target_cache.cc
```diff
@@ -195,7 +195,21 @@ void LevelDbTargetCache::EnumerateTargets(const TargetCallback& callback) {
...
-    TargetData target = DecodeTarget(it->value());
+    StringReader reader{it->value()};
+    auto message = Message<firestore_client_Target>::TryParse(&reader);
+    auto target = serializer_->DecodeTargetData(&reader, *message);
+
+    if (!reader.ok()) {
+      // In https://github.com/firebase/firebase-ios-sdk/issues/6721, a customer
+      // reports that their client crashes with an invalid Target proto. Instead
+      // of crashing the client, we skip the target during LRU GC.
+      // TODO: Evaluate whether it makes sense to proceed if we can determine
```
wilhuff, line 206: Use TODO(username) or TODO(bug) to give context.

A way to achieve the stated goal would be to just avoid the call to decoding the target altogether. If the message fails to even parse, we could synthesize a message with an invalid sequence number that the caller could know to delete. That's probably a bigger change than you want to undertake for this specific fix, so feel free to avoid overhauling this interface for now.

schmidt-sebastian: FWIW I will try to fix these targets in a separate PR now.

var-const (approved):

schmidt-sebastian:
Closing in favor of https://github.com/firebase/firebase-ios-sdk/pull/7018

PR #7025, opened 2020-11-24
Swift pods don't need ---use-libraries.

ryanwilson:
This is causing a nightly failure at the moment. cc @manjanac

### DIFF
--- a/.github/workflows/mlmodeldownloader.yml
+++ b/.github/workflows/mlmodeldownloader.yml
@@ -72,7 +72,7 @@ jobs:
 
   mlmodeldownloader-cron-only:
     # Don't run on private repo.
-    if: github.event_name == 'schedule' && github.repository != 'FirebasePrivate/firebase-ios-sdk'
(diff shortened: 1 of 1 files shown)


paulb777 (commented):
Please temporarily comment the cron line to test.

.github/workflows/mlmodeldownloader.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         flags: [
           '--use-modular-headers',
           # Tests are skipped since the Swift tests need modules.
-          '--skip-tests --use-libraries'
+          '--skip-tests'
```
paulb777, line 84: Just delete this line. No reason to repeat without any tests.

paulb777 (approved):
Delete line and restore cron and then LGTM

.github/workflows/mlmodeldownloader.yml
ryanwilson: Good call, I was able to simplify it further since there was only one flag left and just added it directly in the command.

PR #7038, opened 2020-11-26
Remove SimpleQueryEngine

schmidt-sebastian:
This PR addresses a long outstanding TODO and removes SimpleQueryEngine, which existed to make IndexFreeQueryEngine optional during development. This PR makes IndexFreeQueryEngine the only QueryEngine.

#no-changelog

Port of https://github.com/firebase/firebase-js-sdk/pull/4117


dconeybe (approved):

Firestore/core/test/unit/local/query_engine_test.cc
```diff
@@ -151,14 +150,15 @@ class IndexFreeQueryEngineTest : public ::testing::Test {
     });
   }
 
-  DocumentSet ExpectIndexFreeQuery(const std::function<DocumentSet(void)>& f) {
-    local_documents_view_.ExpectIndexFreeExecution(true);
+  DocumentSet ExpectOptimizedCollectionScan(
```
wilhuff, line 153: Give that the next query engine will add indexes, it seems like the old assertion name is actually a better fit for what we're asserting here.

wilhuff (commented):
Since we're planning on implementing other QueryEngines, it seems like a better version of this change would be to leave the QueryEngine interface in place and leave IndexFreeQueryEngine as the only implementation.

In particular, removing all the plumbing for varying the QueryEngine in the test harness is just going to have to be redone later. I'd prefer if we at least leave this machinery in place. Partially reverting this PR in the future will not be easy, especially if dependencies on this PR's state are added between now and when we add the indexing query engine.

PR #7044, opened 2020-11-26
Add arm64 slice for Catalyst builds.

ryanwilson:
Note: this requires Xcode 12.2 to build. Thankfully, if run an older
system it will just ignore the arm64 flag and build it normally.

#no-changelog yet - unsure when we'll be using this.

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
@@ -32,9 +32,8 @@ enum TargetPlatform: CaseIterable {
     case .iOSDevice: return [.armv7, .arm64]
     // Include arm64 slices in the simulator for Apple silicon Macs.
     case .iOSSimulator: return [.i386, .x86_64, .arm64]
-    // TODO: Evaluate arm64 slice for Catalyst Apple silicon Macs, and x86_64h. Previous builds were
-    // limited to x86_64.
-    case .catalyst: return [.x86_64]
+    // TODO: Evaluate x86_64h slice. Previous builds were limited to x86_64.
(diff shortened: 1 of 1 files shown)


ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -32,9 +32,8 @@ enum TargetPlatform: CaseIterable {
     case .iOSDevice: return [.armv7, .arm64]
     // Include arm64 slices in the simulator for Apple silicon Macs.
     case .iOSSimulator: return [.i386, .x86_64, .arm64]
-    // TODO: Evaluate arm64 slice for Catalyst Apple silicon Macs, and x86_64h. Previous builds were
-    // limited to x86_64.
-    case .catalyst: return [.x86_64]
+    // TODO: Evaluate x86_64h slice. Previous builds were limited to x86_64.
```
paulb777, line 35: What's the difference between `x86_64h` and `x86_64` for Catalyst? Are they equivalent?

Does the Xcode 12 ignoring happen silently?

paulb777 (approved):

ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
ryanwilson: I believe `x86_64h` is a sub-architecture of `x86_64` for Haswell+ Intel Macs. We originally identified it but the previous zip builds were still using `x86_64`.

ryanwilson: And yes, the Xcode 12 ignoring is silent. Only appears in the `xcodebuild` output log.

Issue #6882, opened 2020-11-02
Please keep the Firebase deployment target at iOS 9 - lowest supported by Xcode 12


mman:
Beginning with Firebase release 7.0.0 the deployment target is set to minimum iOS 10.

Xcode 11 minimum deployment target is iOS 8, and Xcode 12 minimum deployment target is iOS 9.

I'd love to see Firebase (being a foundation project for others) to stick to the minimum deployment target of the past Xcode, that is iOS 8, or at least to the current shipping Xcode, that is iOS 9.

Especially because there is no apparent change in for example Crashlytics which works great on 6.34-spm-beta.

Is there any reason technical reason why we absolutely need bump to iOS 10? For Crashlytics?

thanks,
Martin

ryanwilson:
Hi @mman,

Thanks for the report. Unfortunately there are quite a few complexities involved and I'm happy to share our reasoning below, and I have a few questions for you if you don't mind!

Before I get into it, we can certainly evaluate if it was a no-op for Crashlytics and see if there are any technical limitations or support issues with moving back to iOS 9 support for Crashlytics using CocoaPods, although it's not preferred due to reasons mentioned below. For SwiftPM, we can only set a single supported version across all SDKs in this repo so it'll have to stay at 10 for other products.

In terms of our approach, here are a few things we take into account:

- Usage. We have some data that shows iOS 9 usage is incredibly small, so we felt it was safe to bump to 10
- Support. We currently support 5 major versions, and that means the testing matrix and support level is high. Supporting an extra major version is just more work, more complex code (conditionals based on supported versions, API availability, deprecations we have to keep around, etc)
- Technical limitations and code complexities across all products
- We can't stay on iOS 8 due to Xcode issues like (#6533) where Xcode 12 adds warnings when a low version is targeted
- Timing. Since our major version breaks are typically out of sync with Xcode releases, we could run into those issues mentioned above and be in a bad state for a while. Never know what Xcode 13 will bring unfortunately
- Consistency across all products and package managers:
  - In SwiftPM we can only specify a single iOS version for all our products, so it makes sense to have the same support floor for all distribution methods
- We try to break as infrequently as possible, so we don't get to update our min targeted version often. Last major release (6.X) we officially had a support version of iOS 9, but Crashlytics stayed on 8 since it was a no-op
[...]

PR #7053, opened 2020-11-30
Crashlytics bump down to iOS 9 Deployment Target

samedson:
Context:
https://github.com/firebase/firebase-ios-sdk/issues/6882

Believe this is a safe change but going to test just in case.

### DIFF
--- a/Crashlytics/CHANGELOG.md
+++ b/Crashlytics/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 - [changed] Removed obsolete API calls from upload-symbols
-- [changed] Decreased Crashlytics minimum deployment target from iOS 10 to iOS 9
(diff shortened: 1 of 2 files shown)


Crashlytics/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 - [changed] Removed obsolete API calls from upload-symbols
+- [changed] Decreased Crashlytics minimum deployment target from iOS 10 to iOS 9
```
paulb777, line 3: "Decreased Crashlytics CocoaPods minimum deployment target from iOS 10 to iOS 9"

since other distribution mechanisms will stay at iOS 10.

paulb777 (approved):
LGTM on green CI

PR #7064, opened 2020-12-01
Expose upload URLs so that perf can stop instrumenting these URLs.

jeremyjiang-dev:
Prototype reference: https://github.com/FirebasePrivate/firebase-ios-sdk/pull/731

### DIFF
--- a/GoogleDataTransport/CHANGELOG.md
+++ b/GoogleDataTransport/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Unreleased (v8.1.0)
 - Expose upload URLs which FirebasePerformance will depend upon.
-
-# v8.0.2
 - Fix out-of-memory crash for a big amount of pending events. (#6995)
 
 # v8.0.1
--- a/GoogleDataTransport/CHANGELOG.md
+++ b/GoogleDataTransport/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased (v8.1.0)
+- Expose upload URLs which FirebasePerformance will depend upon.
+
 # v8.0.2
 - Fix out-of-memory crash for a big amount of pending events. (#6995)
 
--- a/GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
+++ b/GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
@@ -168,7 +168,6 @@ static NSURL *_testServerURL = nil;
 }
 
 + (nullable NSURL *)serverURLForTarget:(GDTCORTarget)target {
-  
 #if !NDEBUG
   if (_testServerURL) {
     return _testServerURL;
--- a/GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
+++ b/GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
@@ -81,6 +81,7 @@ typedef void (^GDTCCTUploaderEventBatchBlock)(NSNumber *_Nullable batchID,
 @implementation GDTCCTUploader
 
 @synthesize uploaderSession = _uploaderSession;
+static NSURL *_testServerURL = nil;
 
 + (void)load {
   GDTCCTUploader *uploader = [GDTCCTUploader sharedInstance];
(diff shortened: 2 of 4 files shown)


maksymmalyhin (approved):
Great, thank you for the PR! Good to be merged on CI green.

GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
```diff
@@ -99,28 +100,24 @@ + (instancetype)sharedInstance {
...
-  return _uploaderSession;
++ (NSURL *_Nullable)testServerURL {
+  return _testServerURL;
 }
 
-/**
- *
- */
-- (nullable NSURL *)serverURLForTarget:(GDTCORTarget)target {
++ (NSDictionary<NSNumber *, NSURL *> *)uploadURLs {
```
visumickey, line 111: I like this idea of having a dictionary. The only concern is replication. If we add a new target and endpoint, we might have to remember to update the endpoint and target mapping in two separate places which might not be the most ideal one.

Can we make one reference the other and have a single source of truth?

maksymmalyhin: This is a good point! I think we can restructure the code so that `serverURLForTarget` uses `uploadURLs` dictionary under the hood. WDYT?

visumickey: I like that idea. @jeremyjiang-dev WDYT?

jeremyjiang-dev: Thanks Visu and Maksym for the great idea! I have updated the changes.

GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
```diff
@@ -171,30 +159,40 @@ - (nullable NSURL *)serverURLForTarget:(GDTCORTarget)target {
     CSHServerURL = [NSURL URLWithString:[NSString stringWithUTF8String:URL]];
   });
 
+  return @{
```
visumickey, line 162: Can we create a static local dictionary and create the mapping just once and return that value everytime we call?With the current implementation, we would be creating a dictionary every time we call this API.

jeremyjiang-dev: Sounds good! Fixed.

maksymmalyhin (approved):
👍

GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
```diff
@@ -170,31 +158,45 @@ - (nullable NSURL *)serverURLForTarget:(GDTCORTarget)target {
...
-  switch (target) {
-    case kGDTCORTargetCCT:
-      return CCTServerURL;
-
-    case kGDTCORTargetFLL:
-      return FLLServerURL;
-
-    case kGDTCORTargetCSH:
-      return CSHServerURL;
+  return [self uploadURLs][@(target)];
```
visumickey, line 181: NIT: Can you fetch the URLs as a local dictionary and return the value from that dictionary?

visumickey (approved):

paulb777:
Ideally there should be a CHANGELOG entry for a minor version update documenting the feature that FirebasePerformance will depend upon

GoogleDataTransport/GDTCCTLibrary/GDTCCTUploader.m
jeremyjiang-dev: Fixed.

jeremyjiang-dev:
> Ideally there should be a CHANGELOG entry for a minor version update documenting the feature that FirebasePerformance will depend upon

Thanks for reminding me on this! Please take a look at the CHANGELOG  and see if it looks good to you.

GoogleDataTransport/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased (v8.1.0)
+- Expose upload URLs which FirebasePerformance will depend upon.
+
 # v8.0.2
```
paulb777, line 4: 8.0.2 is also unreleased, so both of these items should now be under 8.1.0

jeremyjiang-dev: Fixed.

paulb777 (approved):

PR #7078, opened 2020-12-02
add update() to Firestore Codable API

nidegen:
Adds an `update()` function to the Firestore Swift API where updating fields are provided in an array

### DIFF
--- a/Firestore/Swift/Source/Codable/DocumentReference+WriteEncodable.swift
+++ b/Firestore/Swift/Source/Codable/DocumentReference+WriteEncodable.swift
@@ -59,7 +59,7 @@ extension DocumentReference {
                                     completion: ((Error?) -> Void)? = nil) throws {
     setData(try encoder.encode(value), merge: merge, completion: completion)
   }
-
+  
   /// Encodes an instance of `Encodable` and writes the encoded data to the document referred
   /// by this `DocumentReference` by only replacing the fields specified under `mergeFields`.
   /// Any field that is not specified in mergeFields is ignored and remains untouched. If the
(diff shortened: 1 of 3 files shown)


nidegen:
Todo:
- use `[Any]` to mix `String` and `FieldPath`
- use firebase core methods to iterate filter paths?
- make corresponding `update()` method in ObjC to use the internal tools or expose them?

Firestore/Swift/Source/Codable/DocumentReference+WriteEncodable.swift
```diff
@@ -86,4 +86,38 @@ extension DocumentReference {
...
+  ///   - fields: Array of `String` or `FieldPath` elements specifying which fields to
+  ///                  merge. Fields can contain dots to reference nested fields within the
+  ///                  document.
+  ///   - encoder: An encoder instance to use to run the encoding.
+  ///   - completion: A block to execute once the document has been successfully
+  ///                 written to the server. This block will not be called while
+  ///                 the client is offline, though local changes will be visible
+  ///                 immediately.
+  
+  public func update<T: Encodable>(from value: T,
```
wilhuff, line 111: This is semantically invalid. `setData` and `update` are different operations.

`setData` takes document-shaped data, will unconditionally create a document if it doesn't exist, and will optionally merge the document given with the document on the server.

`update` takes pairs of field paths and values, and will fail if a document doesn't exist on the backend.

So: there are two problems with this PR:
  * `update` is taking document-shaped data, but should be taking field-path and value pairs; and
  * It's implemented in terms of `setData` which will unconditionally create a document, but the main point to prefer update over setData is to avoid creating partial documents.

We haven't previously offered `update` in the Codable API for Firestore because Swift does not have a way to interact with `Codable` types except as concrete types. We'd need a way to operate on type-erased `Codable` values to make this work. It's possible to implement something like `AnyEncodable` and `AnyDecodable` using techniques like those described in http://chris.eidhof.nl/post/type-erasers-in-swift/, but this hasn't been a high enough priority because `setData(mergeFields:)` gets you most of the way there (all it's missing is the failure when the document doesn't exist).

So, if you really need the partial document protection of `update`, then the path forward would be to implement type-erased wrappers for `Encodable` and `Decodable`. If you don't need that specific feature then I'd just continue using `setData(mergeFields:)` and drop this PR.

wilhuff (changes requested):

Firestore/Swift/Source/Codable/DocumentReference+WriteEncodable.swift
nidegen: Thanks for already commenting despite the WIP!

It's not only that I want the partial document protection of update, but I also would also appreciate the protection of Codable for not setting non-coded fields. Else, I could mistype a field key and end up with a non-decodable document.

Wouldn't type erasure remove that effect? I haven't fully understood how `AnyEncodable` would work in detail I have to admit.

For my needs, I think that an update with a `Codable` struct and a filtering field list would actually be quite an ideal API that I already do use in my project (with the current limit of ignoring nested fields. There, the core tools would be nice to use, hence the PR)

wilhuff: The general contract for `update` is to take a list of pairs of paths to values. Accepting the main codable type of the document is inconsistent with this general contract and would create confusion where we document the differences between `update` and `set` operations.

We also can't constrain `update` to only fields that exist in a Codable model type since there are use cases we support that rely on this behavior. For example, if you're migrating data from one form to another, you might want to use update to delete a field that's no longer in your Codable model type. Forcing `update` to only operate on fields that are in the Codable type would prevent this kind of usage.

Note that It's not unreasonable to want this kind of guarded update operation, but we're not in a position to accept it into this SDK because it imposes constraints that are inconsistent with the other platforms we support. I suggest either just making this extension private to your own project, or creating a separate library that you can add to multiple projects.

nidegen: OK, I see. (Although I have to say that it would be a purely additive change that only makes sense with Codable in swift ;)

 I'll keep it in my private extension package then:) Thanks for the feedback! 


------------------------------------------
Et cetero censeo Cloud Functions would benefit from supporting Swift!

PR #7080, opened 2020-12-02
Use initWithFrame: to create the in-app message window in non-UIScene apps

### LINKED ISSUE #6803: [InAppMessaging] Firebase in-app message stuck at top and can't be dismissed on iOS 13+

  * Xcode version: 11.6
  * Firebase SDK version: 6.32.2 (https://github.com/firebase/firebase-ios-sdk/releases/tag/CocoaPods-6.32.2)
  * Firebase Component: InAppMessaging
  * Component version: 0.24.0
  * Installation method: Zip file

- UIWindow may get zero frame and misplace FIRIAMCardViewController on iOS 13+.
- This happens on both device and Simulator on iOS 13 and iOS 14, but not on iOS 12 (and below). 
- This problem was introduced with the UIWindow.init(windowScene:) changes for iOS 13+.

#### Steps to reproduce:

1. Make sure that an IAM will trigger when the app launches (in our case it's triggered when an event fires on the main view).
2. Keep the app in foreground but start to put it in the background by swiping up just a little bit from the home indicator and keep your drag gesture when the IAM should be displayed. 
3. This will show the IAM on iOS 12 device/simulator but not work on iOS 13+ device/simulator. For the simulator on iOS 13+ the IAM window will have a zero frame and the FIRIAMCardViewController view frame is zero and on iOS 13+ devices the window will have a zero frame and the FIRIAMCardViewController view will be misplaced (in our case at the top of the screen).
4. You can't interact with the IAM window at all and you can't dismiss the window either. The only way to get rid of it is to restart the app.

This can probably be reproduced in any project under the same basic conditions. 

#### Possible Solution:

There is no reason to use the new UIWindow.init(windowScene:) on iPhone devices and should be reverted back to the old implementation using UIWindow.init(frame:).

IMG_3613  blurred
iOS 14 device (parts of the view removed from screenshot)

Screenshot 2020-10-22 at 15 10 22
iOS 13 simulator view debugger

christibbs:
These UIWindow factory methods aren't going to win any beauty contests but this should fix #6803

### DIFF
--- a/FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMRenderingWindowHelper.m
+++ b/FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMRenderingWindowHelper.m
@@ -29,12 +29,7 @@
   dispatch_once(&onceToken, ^{
 #if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
     if (@available(iOS 13.0, *)) {
-      UIWindowScene *foregroundedScene = [[self class] foregroundedScene];
-      if (foregroundedScene.delegate) {
-        UIWindowForModal = [[UIWindow alloc] initWithWindowScene:foregroundedScene];
-      } else {
-        UIWindowForModal = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-      }
+      UIWindowForModal = [[self class] iOS13PlusWindow];
     } else {
 #endif  // defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
       UIWindowForModal = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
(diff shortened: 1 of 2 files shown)


FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMRenderingWindowHelper.m
```diff
@@ -50,7 +54,11 @@ + (UIWindow *)UIWindowForBannerView {
 #if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
     if (@available(iOS 13.0, *)) {
       UIWindowScene *foregroundedScene = [[self class] foregroundedScene];
-      UIWindowForBanner = [[FIRIAMBannerViewUIWindow alloc] initWithWindowScene:foregroundedScene];
+      if (foregroundedScene.delegate) {
+        UIWindowForBanner = [[UIWindow alloc] initWithWindowScene:foregroundedScene];
```
paulb777, line 58: Should this be `FIRIAMBannerViewUIWindow` instead of `UIWindow`?

Even better would be to refactor to get rid of the duplicated code.

christibbs: Fixed both in 14c31bb.

paulb777 (approved):

PR #7088, opened 2020-12-03
docs(README): note Xcode >=12 / cocoapods >= 1.10 required

mikehardy:
### Discussion

It appears to not be required for every configuration, but probably easier to specify,
as Xcode 12 and cocoapods 1.10 are generally available and will avoid developer
experience surprises 

https://firebase.google.com/support/release-notes/ios#performance-monitoring

### Testing

Docs only change

### API Changes

Docs only change

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,8 @@ Instructions for the experimental Carthage distribution are at
 To develop Firebase software in this repository, ensure that you have at least
 the following software:
 
-  * Xcode 11.0 (or later)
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Thanks! We're still supporting Xcode 11 for and CocoaPods 1.9.0 for CocoaPods users, but I'm fine to remove the ambiguity for Firebase library developers and encourage everyone to move to the most recent major releases.

PR #7089, opened 2020-12-03
Update Functions emulator to fix CI regression

paulb777:
#no-changelog

### DIFF
--- a/Functions/Backend/package.json
+++ b/Functions/Backend/package.json
@@ -2,8 +2,8 @@
   "name": "functions",
   "description": "Cloud Functions for Firebase",
   "dependencies": {
-    "firebase-admin": "^6.0.0",
-    "firebase-functions": "^2.1.0"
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):

paulb777:
Going to merge here to fix CI, but would still appreciate a deeper look @morganchen12 on if the emulator is being set up correctly

morganchen12:
Where is the functions emulator setup located?

paulb777:
It's located in https://github.com/firebase/firebase-ios-sdk/tree/master/Functions/Backend and used like https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/functions.yml#L30

morganchen12:
Functions usage LGTM, but @samtstern can probably give better scrutiny here than I can.

samtstern:
👍

PR #7094, opened 2020-12-03
Standardize support for Firebase products that integrate with Remote Config.

vic-flair:


FirebaseRemoteConfig/Sources/RCNPersonalization.m
```diff
@@ -37,16 +38,31 @@ - (void)logArmActive:(NSString *)key config:(NSDictionary *)config {
   }
 
   NSDictionary *metadata = ids[key];
-  if (!metadata || metadata[kPersonalizationId] == nil) {
+  NSString *personalizationId = metadata[kPersonalizationId];
```
karenyz, line 41: move this to after the `!metadata` check ?

FirebaseRemoteConfig/Sources/RCNPersonalization.m
```diff
@@ -37,16 +38,31 @@ - (void)logArmActive:(NSString *)key config:(NSDictionary *)config {
...
+  if (!metadata || personalizationId == nil) {
     return;
   }
 
+  // This gets dispatched to a serial queue, so this is OK. But even if not, it'll just possibly log
+  // more.
+  if (self->_armsCache[key] == personalizationId) {
```
karenyz, line 48: does this mean we only want to log each arm once?

FirebaseRemoteConfig/Sources/RCNPersonalization.m
vic-flair: Done.

FirebaseRemoteConfig/Sources/RCNPersonalization.m
vic-flair: We want to log at least once. We can do de-duping on the back-end.

karenyz (approved):

FirebaseRemoteConfig/Sources/RCNPersonalization.h
```diff
@@ -19,16 +19,27 @@
 NS_ASSUME_NONNULL_BEGIN
 
 static NSString *const kAnalyticsOriginPersonalization = @"fp";
-static NSString *const kAnalyticsPullEvent = @"_fpc";
-static NSString *const kArmKey = @"_fpid";
-static NSString *const kArmValue = @"_fpct";
+
+static NSString *const kAnalyticsPullEvent = @"personalization_assignment";
```
erikeldridge, line 23: Minor: as I commented on the Android change, it's unclear to me what "pull" means here and below.

Also, given this whole file concerns logging to Analytics, including "Analytics" in some field names, but not others is confusing. I'd vote to just omit it.

The main distinctions from my perspective are internal vs external, and "event" vs "param" (to use Analytics' "feature type" nomenclature from the [event proposal](http://go/p13n-events)).

To summarize, I'd vote to use "internal" and "external" prefixes to differentiate internal from external events, and "event" and "param" suffixes to differentiate event names from param names, eg `kInternalEvent`, `kInternalChoiceIdParam`, etc

Thinking (no action req'd): all these field names are consistent across Android and iOS 👍

FirebaseRemoteConfig/Sources/RCNPersonalization.h
```diff
@@ -19,16 +19,27 @@
...
+static NSString *const kAnalyticsPullEventInternal = @"_fpc";
+static NSString *const kChoiceId = @"choiceId";
+static NSString *const kChoiceIdLogKey = @"_fpid";
 
 @interface RCNPersonalization : NSObject
 
 /// Analytics connector
 @property(nonatomic, strong) id<FIRAnalyticsInterop> _Nullable analytics;
 
+@property(atomic, strong) NSMutableDictionary *armsCache;
```
erikeldridge, line 41: As @danasilver commented on the Android change, something like `loggedChoiceIds` would be more self-descriptive.

FirebaseRemoteConfig/Sources/RCNPersonalization.m
```diff
@@ -37,16 +38,35 @@ - (void)logArmActive:(NSString *)key config:(NSDictionary *)config {
...
   [self->_analytics logEventWithOrigin:kAnalyticsOriginPersonalization
                                   name:kAnalyticsPullEvent
                             parameters:@{
-                              kArmKey : metadata[kPersonalizationId],
-                              kArmValue : values[key].stringValue
+                              kArmKey : key,
+                              kArmValue : values[key].stringValue,
+                              kPersonalizationIdLogKey : metadata[kPersonalizationId],
+                              kArmIndexLogKey : metadata[kArmIndex],
+                              kGroup : metadata[kGroup]
```
erikeldridge, line 64: Thinking: the [design](http://go/sage-p13n-bq) lists five things we want to log:
1. Developer friendly identifier of the personalization (either the rc param, or an id customers can map back to rc param), satisfied by the `arm_key` field
2. The choice that was made (a string value), satisfied by the `arm_value` field
3. The group the user was in (“p13n” or “baseline”), satisfied by the `group` field
4. Personalization ID, satisfied by the `personalization_id` field
5. Arm index, satisfied by the `arm_index` field

Thinking: the external event name is arbitrary, we just need to document what it is. `personalization_assignment` seems self-descriptive to me 👍

Minor: renaming `arm_key` to something like `rc_parameter` would be more self-descriptive given it communicates an RC parameter. Likewise, renaming the `key` arg passed to `logArmActive` to `rcParameter` would clarify its purpose.

FirebaseRemoteConfig/Sources/RCNPersonalization.m
```diff
@@ -37,16 +38,35 @@ - (void)logArmActive:(NSString *)key config:(NSDictionary *)config {
...
                             }];
+
+  [self->_analytics logEventWithOrigin:kAnalyticsOriginPersonalization
+                                  name:kAnalyticsPullEventInternal
+                            parameters:@{kChoiceIdLogKey : choiceId}];
```
erikeldridge, line 69: Thinking: the [design](http://go/p13n-choice-metadata) lists one thing we want to log in the internal event:

1. Choice ID, satisfied by the `_fpid` param

Thinking: the `_fpc` event name is correct per the [events proposal](http://go/p13n-events).

FirebaseRemoteConfig/Tests/Unit/RCNPersonalizationTest.m
```diff
@@ -118,51 +126,97 @@ - (void)testSinglePersonalizationKey {
...
 
   [_personalization logArmActive:@"key1" config:_configContainer];
   [_personalization logArmActive:@"key2" config:_configContainer];
+  [_personalization logArmActive:@"key1" config:_configContainer];
 
-  OCMVerify(times(2),
+  OCMVerify(times(4),
             [_analyticsMock logEventWithOrigin:kAnalyticsOriginPersonalization
-                                          name:kAnalyticsPullEvent
+                                          name:[OCMArg isKindOfClass:[NSString class]]
```
erikeldridge, line 157: IIUC, we no longer assert the event name anywhere. Given a regression in the event name would be bad, I think it's worth asserting.

erikeldridge (commented):
Seems technically correct to me 👍  I dropped a couple comments re event field naming to improve readability, but since they concern naming, which can be contentious, I flagged them as minor. To avoid miscommunication, I'll defer formal approval to Karen, since she's the iOS lead.

danasilver (approved):

FirebaseRemoteConfig/Sources/RCNPersonalization.m
```diff
@@ -25,28 +25,48 @@ - (instancetype)initWithAnalytics:(id<FIRAnalyticsInterop> _Nullable)analytics {
...
+  if (!metadata) {
     return;
   }
 
+  NSString *choiceId = metadata[kChoiceId];
+  if (choiceId == nil) {
+    return;
+  }
+
+  // This gets dispatched to a serial queue, so this is OK. But even if not, it'll just possibly
```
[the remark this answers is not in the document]
vic-flair: "This" refers to this function. It has to do with how iOS handles async stuff. A serial queue means that functions are executed serially, which isn't always the case. Do you have any suggestions on how to make it clearer?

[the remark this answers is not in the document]
danasilver: What do you think of "Listeners like logArmActive are dispatched to a serial queue, so loggedChoiceIds should contain any previously logged RC Parameter/Choice ID pairs."

PR #7097, opened 2020-12-03
GUL: make NSURLSession promise extension public

maksymmalyhin:


paulb777 (approved):
Will you do a minor version bump for this for any clients?

Also, may want to hold off on merging until after the 7.3 release branch is made.

maksymmalyhin:
> Will you do a minor version bump for this for any clients?

This API hasn't been used in any released SDK yet, so I think a minor version bump should be enough. WDYT?

> Also, may want to hold off on merging until after the 7.3 release branch is made.

Sure, I'm fine to hold off though it looks like a safe change.

PR #7100, opened 2020-12-04
Use iPhone 12 Pro for watchOS builds

ryanwilson:
This was causing a nightly failure. Xcode 12.2 is the default as of Nov 30th.


ryanwilson:
Closing in favour of #7101

PR #7102, opened 2020-12-04
Remove Analytics Catalyst tests.

ryanwilson:
By specifying iOS in the package manifest, it includes Analytics but the
Analytics xcframework doesn't have a Catalyst slice.

#no-changelog

### DIFF
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -29,7 +29,7 @@ jobs:
 
   cron-only:
     # Don't run on private repo.
-    #    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
+    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
 
     runs-on: macOS-latest
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/spm.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
   cron-only:
     # Don't run on private repo.
-    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
+    #    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
```
paulb777, line 32: Restore after testing.

paulb777 (approved):
Thanks! LGTM after green and cron restore.

Package.swift
```diff
@@ -801,8 +801,6 @@ let package = Package(
       dependencies: [
         "FirebaseAuth",
         "FirebaseABTesting",
-        .target(name: "FirebaseAnalytics",
```
paulb777, line 804: Same change needed for the swift test

ryanwilson: Good catch, done.

.github/workflows/spm.yml
ryanwilson: Done.

paulb777:
Merging to get cron fixed for tonight.

PR #7118, opened 2020-12-07
Fix Carthage metrics

paulb777:
Fix the toggling of the Carthage flag between the two Core Diagnostics builds.


ryanwilson (approved):
Thanks for this fix!

PR #7122, opened 2020-12-07
Clean up usages of deprecated APIs

rosalyntan:
Clean up usages of deprecated APIs in Auth now that the minimum iOS version is iOS 10.

### DIFF
--- a/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
+++ b/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
@@ -517,20 +517,13 @@ static dispatch_once_t sProxyAppDelegateRemoteNotificationOnceToken;
       NSSelectorFromString(kGULDidReceiveRemoteNotificationWithCompletionSEL);
   SEL didReceiveRemoteNotificationWithCompletionDonorSEL =
       @selector(application:donor_didReceiveRemoteNotification:fetchCompletionHandler:);
-  if ([appDelegate respondsToSelector:didReceiveRemoteNotificationWithCompletionSEL]) {
-    // Only add the application:didReceiveRemoteNotification:fetchCompletionHandler: method if
-    // the original AppDelegate implements it.
-    // This fixes a bug if an app only implements application:didReceiveRemoteNotification:
-    // (if we add the method with completion, iOS sees that one exists and does not call
-    // the method without the completion, which in this case is the only one the app implements).
 
-    [self proxyDestinationSelector:didReceiveRemoteNotificationWithCompletionSEL
-        implementationsFromSourceSelector:didReceiveRemoteNotificationWithCompletionDonorSEL
-                                fromClass:[GULAppDelegateSwizzler class]
-                                  toClass:appDelegateSubClass
-                                realClass:realClass
-         storeDestinationImplementationTo:realImplementationsBySelector];
-  }
+  [self proxyDestinationSelector:didReceiveRemoteNotificationWithCompletionSEL
+      implementationsFromSourceSelector:didReceiveRemoteNotificationWithCompletionDonorSEL
+                              fromClass:[GULAppDelegateSwizzler class]
+                                toClass:appDelegateSubClass
+                              realClass:realClass
+       storeDestinationImplementationTo:realImplementationsBySelector];
 #endif  // !TARGET_OS_WATCH && !TARGET_OS_OSX
 }
 
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -1502,31 +1502,11 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
   completionHandler(UIBackgroundFetchResultNoData);
 }
 
-// iOS 10 deprecation
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-implementations"
-- (void)application:(UIApplication *)application
-    didReceiveRemoteNotification:(NSDictionary *)userInfo {
-  [self canHandleNotification:userInfo];
-}
-#pragma clang diagnostic pop
-
 - (BOOL)application:(UIApplication *)app
             openURL:(NSURL *)url
             options:(NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
   return [self canHandleURL:url];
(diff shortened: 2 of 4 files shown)


paulb777 (approved):

rosalyntan:
Hi @maksymmalyhin, I ran into an issue with Phone Auth in the sample app while testing this PR, and I have a couple of questions:

`application:didReceiveRemoteNotification` was deprecated in iOS 10, and `application:didReceiveRemoteNotification:fetchCompletionHandler` should be used instead. However, `application:didReceiveRemoteNotification:fetchCompletionHandler` is only swizzled [if the selector is present in the application's ApplicationDelegate](https://github.com/firebase/firebase-ios-sdk/blob/289e093f89f5dd4ba44377cf57dbf6062e8354c3/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m#L520). In order to make swizzling work properly, I need to remove that if-statement from GULAppDelegateSwizzler and always swizzle `application:didReceiveRemoteNotification:fetchCompletionHandler`.

Would this be a breaking change? From the comment in GULAppDelegateSwizzler "This fixes a bug if an app only implements `application:didReceiveRemoteNotification:`" it seems like it would break apps in that state.

maksymmalyhin:
> Hi @maksymmalyhin, I ran into an issue with Phone Auth in the sample app while testing this PR, and I have a couple of questions:
> 
> `application:didReceiveRemoteNotification` was deprecated in iOS 10, and `application:didReceiveRemoteNotification:fetchCompletionHandler` should be used instead. However, `application:didReceiveRemoteNotification:fetchCompletionHandler` is only swizzled [if the selector is present in the application's ApplicationDelegate](https://github.com/firebase/firebase-ios-sdk/blob/289e093f89f5dd4ba44377cf57dbf6062e8354c3/GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m#L520). In order to make swizzling work properly, I need to remove that if-statement from GULAppDelegateSwizzler and always swizzle `application:didReceiveRemoteNotification:fetchCompletionHandler`.
> 
> Would this be a breaking change? From the comment in GULAppDelegateSwizzler "This fixes a bug if an app only implements `application:didReceiveRemoteNotification:`" it seems like it would break apps in that state.

I think we can change the condition to add `application:didReceiveRemoteNotification:fetchCompletionHandler` only if `application:didReceiveRemoteNotification:` is not present in the original app delegate. That should fix your issue and still should not break clients that use old API. @rosalyntan WDYT?

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
```diff
@@ -517,20 +517,13 @@ + (void)proxyRemoteNotificationsMethodsWithAppDelegateSubClass:(Class)appDelegat
       NSSelectorFromString(kGULDidReceiveRemoteNotificationWithCompletionSEL);
   SEL didReceiveRemoteNotificationWithCompletionDonorSEL =
       @selector(application:donor_didReceiveRemoteNotification:fetchCompletionHandler:);
-  if ([appDelegate respondsToSelector:didReceiveRemoteNotificationWithCompletionSEL]) {
```
maksymmalyhin, line 520: As you mentioned before, it is actually a breaking change. I think we can change the condition to the following:
```
if (![appDelegate didReceiveRemoteNotificationSEL])
```
It is still may be considered as breaking but, in this case the clients that use Auth will get depreciation warning in their code if they still use `didReceiveRemoteNotification:` so it will be more legitimate for the swizzler to force using the new API.

maksymmalyhin: I would suggest to add a clear changelog item describing the implications of this change to GoogleUtilities and add a reference to it to Auth to increase visibility of it.

maksymmalyhin (changes requested):
LGTM, but let's finalize the swizzler changes approach.

GoogleUtilities/AppDelegateSwizzler/GULAppDelegateSwizzler.m
paulb777: What exactly is breaking? We shouldn't do anything that would break a Firebase 7 app.

maksymmalyhin: If the method [application:didReceiveRemoteNotification:fetchCompletionHandler:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc) is found in the app delegate then the method [application:didReceiveRemoteNotification:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623117-application?language=objc) will never be called.

My suggestion is that we add  [application:didReceiveRemoteNotification:fetchCompletionHandler:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc) only if [application:didReceiveRemoteNotification:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623117-application?language=objc) is not found to avoid accidentally breaking an app that still uses the deprecated method. The downside of this approach is that if `application:didReceiveRemoteNotification:` is still used together with Auth, then Auth will never receive the notification. I guess this downside may be mitigated by the changelog and the depreciation warning form on iOS 10+. 

Probably another option would be just hold off on the change until GoogleUtilities stop supporting iOS 9. 

Any thoughts on how to handle it better? Feel free to schedule a chat.

rosalyntan: It sounds like either way, removing usage of `application:didReceiveRemoteNotification` is a breaking change since some applications may still use it in their own app delegates. I think waiting until the next breaking change release to clean up these usages sounds good.

However, removing the usage of `application:openURL:sourceApplication:annotation:` now should be fine though, correct?

maksymmalyhin: @rosalyntan Well, if I understand correctly, it looks pretty similar, if the old method is still used in the app then Auth won't get notified. If so, I would postpone the change until Firebase 8.

rosalyntan: Gotcha, sounds good to me. I'll wait till Firebase 8 to be on the safe side.

PR #7123, opened 2020-12-08
Info.plist fixes

paulb777:
- Add Info.plist to all static framework builds since SPM needs them in binary xcframeworks
- Fix the version in generated Info.plist's
- Simplify shell command logging to eliminate or make more rare a race condition crash

Tested by verifying zip and Carthage directory diffs

#no-changelog

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/CarthageUtils.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/CarthageUtils.swift
@@ -119,7 +119,10 @@ extension CarthageUtils {
       // Analytics includes all the Core frameworks and Firebase module, do extra work to package
       // it.
       if product == "FirebaseAnalytics" {
-        createFirebaseFramework(inDir: fullPath, rootDir: packagedDir, templateDir: templateDir)
+        createFirebaseFramework(version: firebaseVersion,
+                                inDir: fullPath,
+                                rootDir: packagedDir,
+                                templateDir: templateDir)
 
         // Copy the NOTICES file from FirebaseCore.
         let noticesName = "NOTICES"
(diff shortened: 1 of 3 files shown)


ReleaseTooling/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -286,12 +291,14 @@ extension CarthageUtils {
...
+    generatePlistContents(forName: "Firebase", withVersion: version, to: frameworkDir)
   }
 
-  static func generatePlistContents(forName name: String, to location: URL) {
+  static func generatePlistContents(forName name: String,
+                                    withVersion version: String,
+                                    to location: URL) {
     let plist: [String: String] = ["CFBundleIdentifier": "com.firebase.Firebase-\(name)",
-                                   "CFBundleInfoDictionaryVersion": "6.0",
+                                   "CFBundleInfoDictionaryVersion": version,
```
ryanwilson, line 301: I'm not sure this should be the pod version, this is the version of the Info.plist structure: https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleinfodictionaryversion?language=objc

Instead I think we should change the [`CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion?language=objc) and maybe even [`CFBundleShortVersionString`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring?language=objc)

paulb777: Good catch! The 6.0 threw me off :)

ryanwilson (approved):
Thanks!

PR #7124, opened 2020-12-08
Add singleton-ish instance methods for model downloader (public API).

manjanac:

### Discussion

### Testing

### API Changes

    can discuss it together.


maksymmalyhin (commented):
Could please trying to integrate with Firebase components system and let me once you are blocked so I will help.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -37,7 +37,7 @@ class DownloadHandlers {
 
 /// Manager to handle model downloading device and storing downloaded model info to persistent storage.
 class ModelDownloadTask: NSObject {
-  let app: FirebaseApp
+  private let app: FirebaseApp
```
ryanwilson, line 40: Nit for a future PR (doesn't have to be handled now): looks like the only thing we use here is `app.name` so it's likely easier for testing purposes and everything to just pass in `appName: String` instead.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -49,7 +50,28 @@ public enum ModelDownloadType {
 }
 
 /// Downloader to manage custom model downloads.
-public struct ModelDownloader {
+public class ModelDownloader {
+  /// FirebaseApp associated with this instance of ModelDownloader.
+  private let app: FirebaseApp
+
+  private init(app: FirebaseApp) {
+    self.app = app
```
ryanwilson, line 58: Instead of storing the app, it may make sense to store the properties that we need out of the app. Unless it's quite a few, but would make testing easier likely.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -49,7 +50,28 @@ public enum ModelDownloadType {
...
+  }
+
+  /// Model downloader with default app.
+  static func modelDownloader() throws -> ModelDownloader {
```
ryanwilson, line 62: It pains me to say this, but at the moment we should keep things consistent with the other Firebase SDKs and `fatalError` instead of throwing.

We'll need to re-evaluate, but we treat this in all other SDKs as a programmer error that isn't recoverable.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -49,7 +50,28 @@ public enum ModelDownloadType {
...
+    guard let defaultApp = FirebaseApp.app() else {
+      // TODO: Replace with more appropriate error.
+      throw DownloadError.internalError(description: "Default Firebase app not configured.")
+    }
+    return modelDownloader(app: defaultApp)
+  }
+
+  /// Model Downloader with custom app.
+  static func modelDownloader(app: FirebaseApp) -> ModelDownloader {
+    return ModelDownloader(app: app)
```
ryanwilson, line 72: Right now this instantiates a new instance every team. We need to keep a static dictionary of all the configured instances keyed on FirebaseApp name and return that instance if requested again. The components system does that for us, but since we're not using it yet well have to manually do so.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -49,7 +50,28 @@ public enum ModelDownloadType {
...
+  /// FirebaseApp associated with this instance of ModelDownloader.
+  private let app: FirebaseApp
+
+  private init(app: FirebaseApp) {
+    self.app = app
+  }
+
+  /// Model downloader with default app.
+  static func modelDownloader() throws -> ModelDownloader {
+    guard let defaultApp = FirebaseApp.app() else {
```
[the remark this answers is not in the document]
ryanwilson: Discussed offline: there are challenges consuming the Objective-C API right now and supporting SwiftPM so the Core SDK will need to split to a public and internal API surface. Since that will take some time to design and implement, let's mark a `TODO` here to switch to the component system when ready and implement things manually for now.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
ryanwilson: Also this should be `public`.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -49,7 +50,28 @@ public enum ModelDownloadType {
...
+      // TODO: Replace with more appropriate error.
+      throw DownloadError.internalError(description: "Default Firebase app not configured.")
+    }
+    return modelDownloader(app: defaultApp)
+  }
+
+  /// Model Downloader with custom app.
+  static func modelDownloader(app: FirebaseApp) -> ModelDownloader {
```
ryanwilson, line 71: Nit: `public` as well

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
ryanwilson: During initialization also, since we're not integrating with components yet, can you add a notification listener for the string `FIRAppDeleteNotification`? It gets posted here: https://github.com/firebase/firebase-ios-sdk/blob/d098e5a8b0750254d49fd24758f5931e161c4ef8/FirebaseCore/Sources/FIRApp.m#L305

`userInfo` has a key `FIRAppNameKey` that contains the name of the app being deleted. Then, check the static instances (or the current instance) if that name matches and clear the local instances of that app name. 

Since the symbols for the notification name and key are internal constants, using the strings themselves is fine for now since it's a temporary measure and they haven't changed in years.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -35,7 +35,7 @@ extension ModelInfoResponse {
 /// Model info retriever for a model from local user defaults or server.
 class ModelInfoRetriever: NSObject {
   /// Current Firebase app.
-  var app: FirebaseApp
+  private var app: FirebaseApp
```
ryanwilson, line 38: Same note here for a future PR: options is used and would be easier to pass in. Also taking in an `installations` instance as well may make testing easier.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Makes sense. I made this change in a couple of other classes as well.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: I added a notification observer for app deletion and a shared dictionary to manage instances. 

But I'm not sure how to clean up app instances... Right now when an app is deleted, I'm only removing the instance from the dictionary. What else should I be doing?

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Good catch, updated.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Gotcha, done.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Done!

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Gotcha, updated.

FirebaseMLModelDownloader/Sources/ModelInfo.swift
```diff
@@ -52,20 +58,19 @@ struct ModelInfo {
...
-    self.name = name
+    name = modelName
     self.downloadURL = url
     self.modelHash = modelHash
     self.size = size
     self.path = path
   }
 
-  func writeToDefaults(app: FirebaseApp, defaults: UserDefaults) throws {
+  func save(toDefaults defaults: UserDefaults, appName: String) throws {
```
ryanwilson, line 68: I feel like the old name was better `writeToDefaults`. Either that, or `saveToDefaults(_:appName:)` works as well based on your preference.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -137,7 +139,15 @@ final class ModelDownloaderUnitTests: XCTestCase {
...
     // results.
-    let modelDownloader = ModelDownloader()
+    guard let testApp = FirebaseApp.app() else {
+      XCTFail("Default app was not configured.")
+      return
+    }
+
+    let modelDownloader = ModelDownloader.modelDownloader()
+
+    let modelDownloaderWithApp = ModelDownloader.modelDownloader(app: testApp)
```
ryanwilson, line 149: Do we want to assert that these are the same? Both are using the default app.

ryanwilson (commented):
LGTM after two small nits, thanks!

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
ryanwilson: Whatever local state you think is necessary. Potentially deleting the local models for that name, since it could be configured with different FIROptions on next run.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
manjanac: Great call! I added the test.

FirebaseMLModelDownloader/Sources/ModelInfo.swift
manjanac: Gotcha, updated to `writeToDefaults(_:appName:)`.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: @ryanwilson tagging in case you missed this in the flurry of comments :)

ryanwilson (approved):

PR #7141, opened 2020-12-10
Enabled Perf GitHub workflow for the public OSS.

jeremyjiang-dev:


zijianjoy (approved):

paulb777 (approved):
Thanks!

Two additional items that can be done in a subsequent PR - Add a quickstart test and testing of other Podfile build options. See https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/abtesting.yml#L74 for an example.

jeremyjiang-dev:
> Thanks!
> 
> Two additional items that can be done in a subsequent PR - Add a quickstart test and testing of other Podfile build options. See https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/abtesting.yml#L74 for an example.

Thanks Paul! Will add these in a subsequent PR.

PR #7146, opened 2020-12-10
Mark perf as open source in zip builder and delete podspec.json

paulb777:
#no-changelog


jeremyjiang-dev:
Are we able to release perf from either public or private repo if we make this change ? We might want hold off on this and make sure perf is always releasable from either public or private repo. WDYT?

visumickey (approved):

paulb777:
@jeremyjiang-dev Maybe the master sync should be turned off for the private perf branch so a binary release could still be done from there?  The Zip Builder can only support one build way at a time.

jeremyjiang-dev:
> @jeremyjiang-dev Maybe the master sync should be turned off for the private perf branch so a binary release could still be done from there? The Zip Builder can only support one build way at a time.

Good idea. @visumickey @zijianjoy WDYT? Is there any concerns?

visumickey:
> > @jeremyjiang-dev Maybe the master sync should be turned off for the private perf branch so a binary release could still be done from there? The Zip Builder can only support one build way at a time.
> 
> Good idea. @visumickey @zijianjoy WDYT? Is there any concerns?

That makes sense to me Jeremy. If we have to make a release from Private Github, we can always manually pull in changes if required.

PR #7151, opened 2020-12-12
Fix nightly segmentation and include performance in report

paulb777:
Fix nightly segmentation failure and add performance to the report 

The segmentation failure did not show up in #7149. @granluo, any ideas why?


maksymmalyhin (approved):

granluo:
Sorry for the delay, might need to reset the github notification. It turns out [Github Workflow APIs](https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#list-repository-workflows) has a page parameter which will display 30 workflows by default. We have 37 workflows in total based on the record I get from the API. Segmentation is in the second page, so that is why it is not shown. I will update the tool to fix it then.

PR #7159, opened 2020-12-14
Callback should be on main thread

paulb777:
Fix regression introduced in #5853 discovered with internal testing.  see b/175119564

Internal testing at cl/347068008

Opinions about whether we should publish this in a 7.3.1 for CocoaPods?

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
+++ b/FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
@@ -401,6 +401,16 @@ static const NSInteger FIRErrorCodeDurableDeepLinkFailed = -119;
     dynamicLinkInternalFromUniversalLinkURL:(NSURL *)url
                                  completion:
                                      (nullable FIRDynamicLinkUniversalLinkHandler)completion {
+  // Make sure the completion is always called on the main queue.
+  FIRDynamicLinkUniversalLinkHandler mainQueueCompletion =
+      ^(FIRDynamicLink *_Nullable dynamicLink, NSError *_Nullable error) {
+        if (completion) {
+          dispatch_async(dispatch_get_main_queue(), ^{
+            completion(dynamicLink, error);
+          });
+        }
+      };
+
   if ([self canParseUniversalLinkURL:url]) {
     if (url.query.length > 0) {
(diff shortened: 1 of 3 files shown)


FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -419,7 +419,9 @@ - (nullable FIRDynamicLink *)dynamicLinkFromCustomSchemeURL:(NSURL *)url {
                  FDLSDKVersion:FIRFirebaseVersion()
                     completion:^(NSURL *_Nullable resolverURL, NSError *_Nullable resolverError) {
                       if (completion) {
-                        completion(dynamicLink, resolverError);
+                        dispatch_async(dispatch_get_main_queue(), ^{
```
maksymmalyhin, line 422: nit: The completion is called further in the method on potentially a different queue. Maybe we may wrap the completion at the beginning of the method to help to prevent such an issue from being introduced accidentally in the future. E.g.:
```
FIRDynamicLinkUniversalLinkHandler mainQueueHandler = ^{
  if (completion) {
         completion(dynamicLink, resolverError);
   });
}
```

maksymmalyhin (approved):
LGTM, one nit.

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
paulb777: That might be part of a bigger change, since this is now following the pattern of the existing `handleUniversalLink` below.

maksymmalyhin: I'm not too strong about it but as for me defining the wrapper and using it instead of the raw completion in the method is a small change that reduces amount of the code and improves readability. I'm fine with leaving it as is as well.

christibbs (approved):

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
maksymmalyhin: let me make a PR. Sorry if my explanation looks confusing.

PR #7165, opened 2020-12-15
Download types WIP

manjanac:
Includes changes from #7163


FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -136,3 +155,71 @@ public class ModelDownloader {
...
+                .internalError(description: "Model unavailable due to deleted local model info.")
+              )
+            )
+            return
+          }
+          completion(.success(localModel))
+        }
+      case let .failure(downloadError):
+        completion(.failure(downloadError))
+      }
```
ryanwilson, line 222: Can we wrap these completion handlers to ensure they're on the main thread?

ryanwilson (commented):
I may have missed something here since it looks like this also includes changes from #7163 so I didn't re-review those parts, but otherwise LGTM

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Done!

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -26,34 +27,46 @@ class DownloadHandlers {
   typealias ProgressHandler = (Float) -> Void
   typealias Completion = (Result<CustomModel, DownloadError>) -> Void
 
+  var runsOnMainThread: Bool
```
ryanwilson, line 30: Hmm, is there a reason why we would want the completion handler to ever get called on the background thread? The download itself will always be in the background, I'm not sure if it's needed to have the extra complexity here.

Issue #7171, opened 2020-12-15
Deadlock in Foundation (pre-macOS 11) triggered by GDT code.


bjhomer:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.2
  * Firebase SDK version: 7.3.0
  * Installation method: `CocoaPods`
  * Firebase Component: GoogleDataTransport 

### [REQUIRED] Step 2: Describe the problem
We have users reporting that our macOS app often hangs on launch on pre-macOS 11 systems. We've traced this to a deadlock happening when two threads try to do an initial access of NSUserDefaults at the same time during `+load` calls. See the attached backtrace for details. We're able to reproduce this on ~50% of launches on these older systems. This is probably dependent on the particular ordering and timing of `+load` calls, and so may be very sensitive to the presence of otherwise-unrelated files.

This makes it ultimately an Apple issue, and they seem to have fixed it in macOS 11 (Big Sur). However, our users are still running into it, and it's easy to fix by making one change to GoogleDataTransport.

#### Steps to reproduce:

1. Install `pod Firebase/Crashlytics` in a macOS app that runs on something older than macOS 11. (We've seen it on both 10.13 and 10.15).
2. Launch the app a few times.

What happened? How can we make the problem occur?
The main thread is doing this:
```
 2803 -[GDTCORLifecycle init]  (in GoogleDataTransport) + 84  [0x103530de1]
   2803 standardDefaultCenter  (in Foundation) + 68  [0x7fff34aab6ea]
     2803 -[NSFileManager _registerForUbiquityAccountChangeNotifications]  (in Foundation) + 182  [0x7fff34aab819]
       2803 _dispatch_once_wait  (in libdispatch.dylib) + 76  [0x7fff6c377dbb]
         2803 _dlock_wait  (in libdispatch.dylib) + 44  [0x7fff6c377e29]
           2803 __ulock_wait  (in libsystem_kernel.dylib) + 10  [0x7fff6c51355e]
```

A second thread is doing this:

```
2803 __37-[GDTCORUploadCoordinator startTimer]_block_invoke_2  (in GoogleDataTransport) + 41  [0x1035349f7]
  2803 +[GDTCORApplication sharedApplication]  (in GoogleDataTransport) + 46  [0x103531f34]
    2803 _dispatch_once_callout  (in libdispatch.dylib) + 20  [0x7fff6c3787de]
[...]

PR #7173, opened 2020-12-15
Delay GDTCORUploadCoordinator timer initial run

bjhomer:
If an upload coordinator gets created during `+load` calls, it can end up running its immediately, while +load calls are still running. It's generally a bad idea to be splitting off threads while +load calls are still running.

The upload coordinator checks every 30 seconds, and doesn't really need to run right away. Since it's easy to transitively create an upload coordinator from other calls that happen during +load, we'll just push off the initial run of the timer slightly. That makes it less likely to interfere (or deadlock) with other load calls.

See #7171 for an example of a problem solved by this change.

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/GoogleDataTransport/GDTCORLibrary/GDTCORUploadCoordinator.m
+++ b/GoogleDataTransport/GDTCORLibrary/GDTCORUploadCoordinator.m
@@ -68,10 +68,12 @@
       return;
     }
 
+    // Delay the timer slightly so it doesn't run while +load calls are still running.
+    dispatch_time_t deadline = dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC / 2);
+
     self->_timer =
         dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER, 0, 0, self->_coordinationQueue);
-    dispatch_source_set_timer(self->_timer, DISPATCH_TIME_NOW, self->_timerInterval,
-                              self->_timerLeeway);
+    dispatch_source_set_timer(self->_timer, deadline, self->_timerInterval, self->_timerLeeway);
 
     dispatch_source_set_event_handler(self->_timer, ^{
       if (![[GDTCORApplication sharedApplication] isRunningInBackground]) {


maksymmalyhin:
@bjhomer Could you please sign CLA following the instructions from the [comment](https://github.com/firebase/firebase-ios-sdk/pull/7173#issuecomment-745461084).

bjhomer:
Yep, CLA has been signed. I also pushed up a tweak to make the style checker happy.

maksymmalyhin (approved):
Thank you for the PR! Looks good to be merged on CI green.

maksymmalyhin:
Looks like the change triggered a failure of a test expecting a particular timing of `GDTCORUploadCoordinator`. I assume, the tests have to be updated to reflect the updated timing. I'll take a look what we should do here.

bjhomer:
Ah, yep, it looks like the tests are expecting the timer to kick things off immediately. A shorter delay on the timer _may_ be sufficient, but at this point I think I may defer to your judgment.

bjhomer:
(If you're interested, I can also put up a PR to pull this into `release-7.3`. It appears that the offending test doesn't exist there, so it would probably pass CI. But I'm also okay just using that as my own temporary branch for now.)

maksymmalyhin:
> Ah, yep, it looks like the tests are expecting the timer to kick things off immediately. A shorter delay on the timer _may_ be sufficient, but at this point I think I may defer to your judgment.

Sorry for the delayed response. Yeah, reducing initial delay looks reasonable. Could you please update it to 0.5 sec. To fix the tests feel free to increase slightly increase the waiting time: https://github.com/firebase/firebase-ios-sdk/blob/d5e421c9861b3c79f002f0b438f9c7b33cadc061/GoogleDataTransport/GDTCORTests/Unit/GDTCORUploadCoordinatorTest.m#L133

bjhomer:
Delay is now 0.5 sec. All tests are passing. With the change in delay, no test changes were needed.

maksymmalyhin:
@bjhomer Congratulations with you first contribution to Firebase! Thank you!

PR #7187, opened 2020-12-17
Add service extension, app clips and watch app to Messaging advanced sample app

chliangGoogle:
Also added a function to share fcm token with other apps for easy debugging. I realize this is not a share extension, it's actually using UIActivityViewController to achieve it. Messaging doesn't have a use case for share extension, so not adding it to the test app.

The App Clips shared the same source code with main sample app so no code added.
Tested sending notification to app clips and watch app, also enable image display in service extension. 
#no-changelog


FirebaseMessaging/Apps/AdvancedSample/Podfile
```diff
@@ -3,15 +3,35 @@ use_frameworks!
...
+  shared_pods
+end
 
+target 'NotificationServiceExtension' do
+  platform :ios, '13.0'
+  shared_pods
+end
+
+target 'AppClips' do
+  use_frameworks! :linkage => :static
```
paulb777, line 27: It's kind of weird to put the linkage here versus near the top of the file. I don't think different targets can be linked differently.

paulb777 (commented):
Please fix the style and add a build test to CI

FirebaseMessaging/Apps/AdvancedSample/Podfile
chliangGoogle: I got an crash error "dyld: Library not loaded. Image not found" for Promise library linking when running app clips and  I put this in podfile that fixes the issue.

paulb777: Do you get the dyld error when `use_frameworks! :linkage => :static` is at the top of the file and when it's not there at all?

chliangGoogle: this is only for app clip that is why I put it inside app clip target, originally I had use_framework! on top of file that crashed app clips

paulb777: Hmm, I would think `use_frameworks! :linkage => :static` at the top should be equivalent?

paulb777 (approved):

FirebaseMessaging/Apps/AdvancedSample/Podfile
chliangGoogle: So I guess we can apply this to all targets, I just thought we have to make special case for static linkage for app clips.

paulb777: Unless there's something I don't understand about App Clips, it's not possible to link differently in the same app.

chliangGoogle: mmmm, I put it on top and seems working for all targets. But I don't know why app clips needs this.

christibbs:
Great work @chliangGoogle!

PR #7188, opened 2020-12-17
Fix FirebaseMetadataProvider under VS2015

wilhuff:
VS2015 does not support declaring `static constexpr char[]` members of classes.

#no-changelog


dconeybe (approved):
LGTM

PR #7193, opened 2020-12-17
Disable posting new issues if there are consecutive successful report.

granluo:


paulb777 (approved):
Thanks!

granluo:
The update is to include all workflows, including Segmentation and other workflows not included in the first page, which has 30 workflows by default.

paulb777 (approved):

PR #7204, opened 2020-12-21
Add --pods option

paulb777:
Making it easier to use the zip-builder from the command line. 

Now the common use case should be one option: `--pods <list of pods to build>

#no-changelog


ryanwilson:
Nice!

PR #7220, opened 2020-12-28
Add a flag to skip building for Catalyst

paulb777:
#no-changelog

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/main.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/main.swift
@@ -79,6 +79,12 @@ struct ZipBuilderTool: ParsableCommand {
         help: ArgumentHelp("A flag to indicate keeping (not deleting) the build artifacts."))
   var keepBuildArtifacts: Bool
 
+  /// Flag to skip building the Catalyst slices.
+  @Flag(default: false,
+        inversion: .prefixedNo,
+        help: ArgumentHelp("A flag to indicate keeping (not deleting) the build artifacts."))
+  var skipCatalyst: Bool
+
   /// Flag to run `pod repo update` and `pod cache clean --all`.
   @Flag(default: true,
         inversion: .prefixedNo,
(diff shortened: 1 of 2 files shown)


ReleaseTooling/Sources/ZipBuilder/main.swift
```diff
@@ -79,6 +79,12 @@ struct ZipBuilderTool: ParsableCommand {
         help: ArgumentHelp("A flag to indicate keeping (not deleting) the build artifacts."))
   var keepBuildArtifacts: Bool
 
+  /// Flag to skip building the Catalyst slices.
+  @Flag(default: false,
+        inversion: .prefixedNo,
+        help: ArgumentHelp("A flag to indicate keeping (not deleting) the build artifacts."))
```
ryanwilson, line 85: Nit: argument comment is out of date.

ReleaseTooling/Sources/ZipBuilder/main.swift
```diff
@@ -79,6 +79,12 @@ struct ZipBuilderTool: ParsableCommand {
...
+  var skipCatalyst: Bool
```
ryanwilson, line 86: Should this be something like `includeCatalyst` to avoid a double negative of `no-skip-catalyst`?

ReleaseTooling/Sources/ZipBuilder/main.swift
```diff
@@ -232,6 +238,11 @@ struct ZipBuilderTool: ParsableCommand {
                                macos: minimumMACOSVersion,
                                tvos: minimumTVOSVersion)
 
+    // Update iOS target platforms if `--skip-catalyst` was specified.
+    if skipCatalyst {
+      SkipCatalyst.set()
+    }
+
```
ryanwilson, line 245: Hmm can this be passed in some other way other then having a global flag? This API feels awkward

ReleaseTooling/Sources/ZipBuilder/main.swift
paulb777: Fix in next commit.

ReleaseTooling/Sources/ZipBuilder/main.swift
paulb777: Done.

ReleaseTooling/Sources/ZipBuilder/main.swift
paulb777: Maybe? But I couldn't come up with one. There is a similar API for minimum versions. Any suggestions?

PR #7221, opened 2020-12-28
ZipBuilder: Multi-platform building and usability improvements

paulb777:
This is an umbrella PR for several PR's that add multi-platform support, usability improvements, and some cleanup.

I recommend that the PR associated with each commit be reviewed separately and I'll address any feedback here.

The major open questions are:
1. What should be the default for the `--platforms` option?
2. Should we start releasing Firebase for all three supported platforms?

Pro for all three platforms in number 1: It's the future
Con is that all existing iOS only users will need to add `--platforms ios`

Pro for releasing all three Platforms in Firebase zip: Address requests for more complete zip offering
Con: about 50% slower build time and customers can build it themselves

The zip.yml changes will be reverted before merging.


ryanwilson:
Thanks for doing this, and thanks for the amazing organization! About halfway done the PRs now, but to address some of the open questions:

> The major open questions are:
> 
> 1. What should be the default for the `--platforms` option?

I think all the possible platforms makes sense for the default. That way we're intentional. In an ideal world, it's whatever the pod supports, but I don't think that's a priority.

> 2. Should we start releasing Firebase for all three supported platforms?
> 
> Pro for all three platforms in number 1: It's the future
> Con is that all existing iOS only users will need to add `--platforms ios`

I'm not sure what you mean by that con. iOS only developers can still include the `.xcframework` that includes tvOS and macOS even if their app doesn't, no?

> Pro for releasing all three Platforms in Firebase zip: Address requests for more complete zip offering
> Con: about 50% slower build time and customers can build it themselves

Given we're offloading building to GHA, I think that's a reasonable tradeoff given the benefits.

paulb777:
@ryanwilson Review comments should be addressed. PTAL

ryanwilson (approved):
LGTM after reverting the zip.yml changes

PR #7223, opened 2020-12-28
Check token is different from cache and storage before  updating it 

chliangGoogle:
When messaging updates token and notifying InstanceID, instanceID should first check to see the token from Messaging is different than the one in local cache or in the database. 
This extra check helps reduce storage writing in the future when we remove InstanceID and handle token completely from Messaging.

### DIFF
--- a/Firebase/InstanceID/FIRInstanceID.m
+++ b/Firebase/InstanceID/FIRInstanceID.m
@@ -1138,7 +1138,7 @@ static FIRInstanceID *gInstanceID;
         (self.defaultFCMToken.length && tokenUpdatedFromMessaging.length &&
          ![self.defaultFCMToken isEqualToString:tokenUpdatedFromMessaging]) ||
         (cachedToken.length && tokenUpdatedFromMessaging.length &&
-         ![self.defaultFCMToken isEqualToString:tokenUpdatedFromMessaging])) {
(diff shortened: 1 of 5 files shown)


Firebase/InstanceID/FIRInstanceID.m
```diff
@@ -1127,9 +1127,22 @@ - (void)observeFirebaseMessagingTokenChanges {
...
-    self.defaultFCMToken = tokenUpdatedFromMessaging;
-    [self.tokenManager saveDefaultToken:tokenUpdatedFromMessaging
-                            withOptions:[self defaultTokenOptions]];
+    // Check the token from storage along with local value.
+    FIRInstanceIDTokenInfo *cachedTokenInfo =
+        [self.tokenManager cachedTokenInfoWithAuthorizedEntity:self.fcmSenderID
+                                                         scope:kFIRInstanceIDDefaultTokenScope];
+    NSString *cachedToken = cachedTokenInfo.token;
+
+    if (self.defaultFCMToken.length != tokenUpdatedFromMessaging.length ||
```
paulb777, line 1136: Why are the length checks needed? Aren't two `isEqualToString` checks sufficient?

chliangGoogle: In case one of the string is null, and a [null isEqualtoString: null] in valid.

paulb777 (approved):

PR #7224, opened 2020-12-29
Update to disble blank issue.

granluo:


.github/actions/testing_report_generation/app.rb
```diff
@@ -135,5 +135,5 @@ def get_workflows(client, repo_name)
   client.add_comment(REPO_NAME_WITH_OWNER, last_issue.number,failure_report.get_report)
 # Create a new issue if there exists failed workflows.
 else
```
paulb777, line 137: Shouldn't this else clause be deleted?  Won't the `create_issue` at line 127 cover all cases where a new issue needs t be created?

paulb777:
@granluo See the question above

granluo (commented):
line 127 is under a situation that both failure and success report are nil, and an issue or a comment will be created for no workflows running. The content of the issue is NO_WORKFLOW_RUNNING_INFO in line 127.
The else clause is to create an issue. The content is failure_report.get_report, but we need to make sure is not nil ahead, so the unless there is a guard clause to ensure ``failure_report` is not nil before creating a issue.

.github/actions/testing_report_generation/app.rb
granluo: > @granluo See the question above

The comment was pending and has been updated.

paulb777: Thanks for the clarification.

paulb777 (approved):

granluo:
Will close the issue #7226 and see if it works tonight.

PR #7229, opened 2020-12-30
Update SwiftPackageManager.md with instructions for Crashlytics SPM

elenadoty:
Add instructions for where to find Crashlytics run scripts for Swift Package Manager installation.

### DIFF
--- a/SwiftPackageManager.md
+++ b/SwiftPackageManager.md
@@ -51,7 +51,7 @@ https://github.com/firebase/firebase-ios-sdk/issues/6472#issuecomment-694449182.
 
 <img src="docs/resources/SPMObjC.png">
 
-If you're using FirebaseCrashlytics, use 
+If you're using FirebaseCrashlytics, use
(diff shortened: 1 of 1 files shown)


SwiftPackageManager.md
```diff
@@ -51,6 +51,10 @@ https://github.com/firebase/firebase-ios-sdk/issues/6472#issuecomment-694449182.
 
 <img src="docs/resources/SPMObjC.png">
 
+If you're using FirebaseCrashlytics, use
+`${BUILD_DIR%Build/*}SourcePackages/checkouts/firebase-ios-sdk/Crashlytics/run`
```
paulb777, line 55: Does `%Build` get populated by Xcode for the run script?

paulb777 (approved):

SwiftPackageManager.md
elenadoty: Good point, it does not and we only need the BUILD_DIR part.

PR #7233, opened 2020-12-31
Should only have only one local cache of token 

chliangGoogle:
This is part of the refactor for removing InstanceID dependency.

The local cache of token currently have two copies. Use the one in TokenManager and removed the duplicate from FIRMessaging to avoid complexity of syncing problems.

Also the previous master branch merging with chen/fm-master cause a few conflicts unresolved in unit tests, fixing them as well.
#no-changelog since this is not merging to master yet.


FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -540,7 +544,8 @@ - (void)setAutoInitEnabled:(BOOL)autoInitEnabled {
 
 - (NSString *)FCMToken {
   // Gets the current default token, and requets a new one if it doesn't exist.
-  return [[FIRMessaging messaging].tokenManager tokenAndRequestIfNotExist];
+  NSString *token = [[FIRMessaging messaging].tokenManager tokenAndRequestIfNotExist];
```
maksymmalyhin, line 547: I wonder why `[[FIRMessaging messaging].tokenManager` instead of `self. tokenManager` ?

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -682,13 +686,13 @@ - (void)notifyRefreshedFCMToken {
     return;
   }
   if ([self.delegate respondsToSelector:@selector(messaging:didReceiveRegistrationToken:)]) {
-    [self.delegate messaging:self didReceiveRegistrationToken:self.tokenManager.defaultFCMToken];
+    [self.delegate messaging:self didReceiveRegistrationToken:_tokenManager.defaultFCMToken];
```
maksymmalyhin, line 689: I wonder why use ivar `_tokenManager` here but a property `self.tokenManager` at other places?

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -838,11 +842,12 @@ - (void)defaultInstanceIDTokenWasRefreshed:(NSNotification *)notification {
...
   // Retrieve the Instance ID default token, and should notify delegate and
   // trigger notification as long as the token is different from previous state.
-  NSString *oldToken = self.defaultFcmToken;
-  self.defaultFcmToken = [(NSString *)notification.object copy];
-  if ((self.defaultFcmToken.length && oldToken.length &&
-       ![self.defaultFcmToken isEqualToString:oldToken]) ||
-      self.defaultFcmToken.length != oldToken.length) {
+  NSString *oldToken = _tokenManager.defaultFCMToken;
+  NSString *newToken = [(NSString *)notification.object copy];
+  if ((newToken.length && oldToken.length && ![newToken isEqualToString:oldToken]) ||
```
maksymmalyhin, line 847: optional: I took me some time to understand this check. I wonder would it be more readable if we explicitly check new and old token for `nil` like:
```
(newToken != nil & oldToken != nil & ![newToken isEqualToString:oldToken])
```

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.h
```diff
@@ -173,4 +171,10 @@ typedef NS_OPTIONS(NSUInteger, FIRMessagingInvalidTokenReason) {
  */
 - (void)saveDefaultTokenInfo:(NSString *)defaultFcmToken;
 
+/*
+ * Sets the default FCM token the first time from FIRMessaging.
+ *
+ */
+- (void)setDefaultFCMToken:(NSString *)defaultFcmToken;
```
maksymmalyhin, line 178: This method looks confusingly similar to the one defined above `- (void)saveDefaultTokenInfo:(NSString *)defaultFcmToken;`. What is the difference between them? Do we actually need both?

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -102,23 +102,28 @@ - (NSString *)defaultFCMToken {
...
-                                                        token:defaultFcmToken
-                                                   appVersion:FIRMessagingCurrentAppVersion()
-                                                firebaseAppID:_firebaseAppID];
-  tokenInfo.APNSInfo =
-      [[FIRMessagingAPNSInfo alloc] initWithTokenOptionsDictionary:[self tokenOptions]];
-
-  [self->_tokenStore saveTokenInfoInCache:tokenInfo];
+  if (_defaultFCMToken.length != defaultFcmToken.length ||
+      (_defaultFCMToken.length && defaultFcmToken.length &&
+       ![_defaultFCMToken isEqualToString:defaultFcmToken])) {
```
maksymmalyhin, line 114: nit: I see this non-trivial check again. Would it make sense to move it to a helper to improve readability and reduce code duplication?

FirebaseMessaging/Tests/UnitTests/FIRMessagingTestUtilities.m
```diff
@@ -16,29 +16,18 @@
...
 #import "FirebaseMessaging/Sources/FIRMessagingPubSub.h"
 #import "FirebaseMessaging/Sources/FIRMessagingRmqManager.h"
 #import "FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.h"
+#import "FirebaseMessaging/Tests/UnitTests/FIRMessagingTestUtilities.h"
+#import "FirebaseMessaging/Tests/UnitTests/XCTestCase+FIRMessagingRmqManagerTests.h"
+#import "Interop/Analytics/Public/FIRAnalyticsInterop.h"
 
-NS_ASSUME_NONNULL_BEGIN
-static NSString *const kFIRMessagingDefaultsTestDomain = @"com.messaging.tests";
+NS_ASSUME_NONNULL_BEGIN static NSString *const kFIRMessagingDefaultsTestDomain =
```
maksymmalyhin, line 29: Add a line break after `NS_ASSUME_NONNULL_BEGIN ` ?

maksymmalyhin (approved):
LGTM, a couple minor comments and questions.

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: You are right, self.tokenManager is better here since there's no specific extra action in getTokenManager.

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: Must be missed through my refactor, yeah changed to self.tokenManager

paulb777 (approved):

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: what happen if old token is nil and new token is a string, we want this case to be true.
Or you only mean the first line of the condtion?

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.h
chliangGoogle: They did lost meaning after the refactor. I updated the description and naming of the methods.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTestUtilities.m
chliangGoogle: Done.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
chliangGoogle: Done.

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -688,7 +692,7 @@ - (void)notifyRefreshedFCMToken {
   // Should always trigger the token refresh notification when the delegate method is called
   NSNotificationCenter *center = [NSNotificationCenter defaultCenter];
   [center postNotificationName:FIRMessagingRegistrationTokenRefreshedNotification
-                        object:self.tokenManager.defaultFCMToken];
+                        object:_tokenManager.defaultFCMToken];
```
maksymmalyhin, line 695: `self.tokenManager` here as well?

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.h
maksymmalyhin: It still looks a bit confusing to me, sorry if I'm missing something. As far as I can see the method `- (void)setDefaultFCMToken:(NSString *)defaultFcmToken` only posts as notification about the token update. Maybe a name like `defaultFCMTokenWasUpdate:` or `postDefaultFCMTokenRefreshNotification:` would better reflect the implementation?

FirebaseMessaging/Sources/FIRMessaging.m
maksymmalyhin: Yeah, I mean keeping the same checks but replace `token.length` by `token != nil` which reads a bit better to me. I don't really have a strong opinion here. It looks fine especially after moving the checks to the separate method.

maksymmalyhin (approved):
Thank you for the updates! LGTM, a couple optional comments.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.h
chliangGoogle: Yep, this function eventually trigger a method by the notification to actually set the token, but I see the naming is confusing. Rename it to postDefaultFCMTokenRefreshNotification:.

FirebaseMessaging/Sources/FIRMessaging.m
chliangGoogle: Done.

ryanwilson:
Happy to leave the approvals to Paul and Maksym but let me know if you want me to take a look as well.

maksymmalyhin (approved):
LGTM, thanks!

chliangGoogle:
Thanks @ryanwilson. I think the PR is in a good shape but always welcome any feedback. :)

PR #7245, opened 2021-01-05
Migrate libFuzzer from svn to git

wilhuff:
Now that the LLVM project has completed its migration to git, the svn checkouts are now failing.

Also fix an incidental issue that arose while testing locally--I have homebrew installed in `$HOME/homebrew` and the fuzzer build was failing to find `protoc` as a result.

### DIFF
--- a/Firestore/Example/FuzzTests/FuzzingResources/Serializer/Corpus/ConvertTextToBinary.sh
+++ b/Firestore/Example/FuzzTests/FuzzingResources/Serializer/Corpus/ConvertTextToBinary.sh
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/bash -l
 
-# Copyright 2018 Google
+# Copyright 2018 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Firestore/Example/FuzzTests/FuzzingResources/Serializer/Corpus/ConvertTextToBinary.sh
```diff
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/bash -l
```
dconeybe, line 1: What's the rationale for adding `-l` to the bash command-line arguments?

dconeybe (approved):

Firestore/Example/FuzzTests/FuzzingResources/Serializer/Corpus/ConvertTextToBinary.sh
dconeybe: Nevermind. I forgot to discard this comment and I see that you address this explicitly in the comment added to this file.

Issue #7242, opened 2021-01-03
@DocumentID not populated with custom codable implementation


everlof:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.2
  * Firebase SDK version: 7.3.0
  * Installation method: `Swift Package Manager`
  * Firebase Component: Firestore

### [REQUIRED] Step 2: Describe the problem

I've taken a look into the codable API, which looks really nice! I've watched https://www.youtube.com/watch?v=3-yQeAf3bLE which links to "Encoding and Decoding Custom Types → https://goo.gle/2T1Quic".

However, I haven't found enough resources about how to use `Codable` with:
- Firestore
- Custom types (implement `init(from:Decoder` + `encode(to:Encoder)`)
- `@DocumentID` property wrapper

If that case is supported, I think it would be great to have some examples of how to use it, if it's not supported, perhaps that should be mentioned somewhere.

#### Steps to reproduce:

Create a struct with `@DocumentID`, `CodingKeys` and implement `public init(from decoder: Decoder) throws` as well as `public func encode(to encoder: Encoder) throws`.

How is the property decorated with `@DocumentID` supposed to be encoded/decoded? The only way I've been able to figure it out is by getting it from the `userInfo`, which obviously is wrong.

#### Relevant Code

```swift
struct TestObject: Codable {
    public enum CodingKeys: String, CodingKey {
        case ref = "r"
    }

    @DocumentID
    public var ref: DocumentReference?

    public init() {
        ref = nil
    }

    public init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        ref = try container.decodeIfPresent(DocumentReference.self, forKey: .ref)
    }

    public func encode(to encoder: Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encodeIfPresent(ref, forKey: .ref)
    }
}

class SomeTests: XCTestCase {

    lazy var firestore: Firestore = {
        let path = Bundle(for: SomeTests.self).url(forResource: "GoogleService-Info", withExtension: "plist")!.path
        if FirebaseApp.app() == nil {
            FirebaseApp.configure(options: FirebaseOptions(contentsOfFile: path)!)
        }

        let firestore = Firestore.firestore()
        let settings = firestore.settings
        settings.host = "localhost:8080"
        settings.isPersistenceEnabled = false
        settings.isSSLEnabled = false
        firestore.settings = settings
[...]

PR #7247, opened 2021-01-05
Verify that @DocumentID works with custom Codable methods.

wilhuff:
Prompted by #7242.

This shows that property wrappers like `@DocumentID` can be used in custom Codable methods, and produce results essentially identical to `testSelfDocumentID`.

#no-changelog


thebrianchen (approved):
lgtm, thanks for taking over this one + the detailed explanation!

PR #7259, opened 2021-01-07
Logical checkpoint of ModelDownloader implementation

manjanac:
Includes the following implementations:

1) getModel w/ all download types (#7165)
2) listModels (#7232)
3) deleteModel (#7227)


FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -98,41 +111,203 @@ public class ModelDownloader {
...
+    } catch let error as DownloadedModelError {
+      completion(.failure(error))
+    } catch {
+      completion(.failure(.internalError(description: error.localizedDescription)))
+    }
   }
 
   /// Deletes a custom model from device.
   public func deleteDownloadedModel(name modelName: String,
                                     completion: @escaping (Result<Void, DownloadedModelError>)
```
ryanwilson, line 215: Hmm, I may have missed this in the public API review but it seems like this shouldn't be a result, but should be an optional DownloadModelError as the return type. WDYT?

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
```diff
@@ -176,4 +182,167 @@ final class ModelDownloaderIntegrationTests: XCTestCase {
     waitForExpectations(timeout: 5, handler: nil)
     XCTAssertEqual(modelDownloadManager.downloadStatus, .completed)
   }
+
+  func testGetModel() {
+    let testApp = FirebaseApp.app()!
```
ryanwilson, line 187: I can't remember if we've done it elsewhere already - are you able to use an internal initializer for `FirebaseApp` to instantiate it manually instead of using the shared singleton?

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
```diff
@@ -176,4 +182,167 @@ final class ModelDownloaderIntegrationTests: XCTestCase {
...
+    /// Test download type - latest model.
+    var downloadType: ModelDownloadType = .latestModel
+    let latestModelExpectation = expectation(description: "Get latest model.")
+
+    modelDownloader.getModel(
+      name: testModelName,
+      downloadType: downloadType,
+      conditions: conditions,
+      progressHandler: { progress in
+        XCTAssertNotNil(progress)
```
ryanwilson, line 206: Hmm is progress optional? If so, should it be?

FirebaseMLModelDownloader/Sources/DownloaderUserDefaults.swift
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
manjanac: Ha, good catch! Fixed.

FirebaseMLModelDownloader/Sources/DownloaderUserDefaults.swift
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Protocol to write info to user defaults.
+protocol DownloaderUserDefaults {
```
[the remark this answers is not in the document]
manjanac: I like it. I left it open-ended thinking I may want a method to read from user defaults in the future, but I'll revisit the name when I need to do that.

FirebaseMLModelDownloader/Sources/LocalModelInfo.swift
```diff
@@ -71,8 +65,16 @@ class LocalModelInfo {
...
+
+/// Extension to write local model info to user defaults.
+extension LocalModelInfo: DownloaderUserDefaults {
+  /// Get user defaults key prefix.
+  private static func getUserDefaultsKeyPrefix(appName: String, modelName: String) -> String {
+    let bundleID = Bundle.main.bundleIdentifier ?? ""
+    return "\(bundleID).\(appName).\(modelName)"
+  }
 
-  /// Write local model info to user defaults.
```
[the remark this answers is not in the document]
manjanac: Fixed, thanks!

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -30,7 +30,8 @@ class DownloadHandlers {
   var progressHandler: ProgressHandler?
   var completion: Completion
 
-  init(progressHandler: ProgressHandler?, completion: @escaping Completion) {
+  init(progressHandler: ProgressHandler?,
+       completion: @escaping Completion) {
```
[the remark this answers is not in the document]
manjanac: Done.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -98,41 +111,203 @@ public class ModelDownloader {
...
+  static func modelDownloader(app: FirebaseApp, defaults: UserDefaults) -> ModelDownloader {
+    if let downloader = modelDownloaderDictionary[app.name] {
+      return downloader
+    } else {
+      let downloader = ModelDownloader(app: app, defaults: defaults)
+      modelDownloaderDictionary[app.name] = downloader
+      return downloader
+    }
+  }
+
```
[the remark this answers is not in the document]
manjanac: Good point, I moved it to an extension. Eventually it would be nice to use a protocol for modelDownloader that the tests can implement differently.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Yes we talked about this and went with Result to keep it consistent with all the other public APIs. We could change it to an optional if needed, though.

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
manjanac: I haven't used this before, but added a TODO for a future PR.

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
manjanac: So the handler itself can be optional. But if it exists, it's called within urlSession didWriteData with a non-optional progress.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
ryanwilson: Something to revisit later I think, not a blocker for now obviously

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
ryanwilson: Gotcha, so this could never fail. Maybe a more appropriate check here would be to ensure the value is within the expected range (0-1)?

ryanwilson (commented):
One small nit about the test then LGTM!

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
manjanac: Oh sorry I understand now, so progress can be null because the handler isn't called at all (which happens if the model is already on device). So this is testing that the model is indeed being downloaded. I like the range idea too.

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
```diff
@@ -162,6 +162,8 @@ final class ModelDownloaderIntegrationTests: XCTestCase {
       defaults: .createTestInstance(testName: #function),
       progressHandler: { progress in
         XCTAssertNotNil(progress)
```
ryanwilson, line 164: This `NotNil` line can be removed since the `progress` variable is not an Optional (it's a plain `Float`) so it can never fail.

ryanwilson (approved):
Thanks! One more small touch-up then it's ready for merging (on green)

ryanwilson (approved):

PR #7269, opened 2021-01-07
Remote Config: Revert to previous nil check on experiment JSON

### LINKED ISSUE #7184: RCNConfigExperiment is incorrectly flagging valid experiment JSON payloads as invalid

  * Xcode version: 12.2
  * Firebase SDK version: 7.0.0
  * Installation method: Carthage

Since we upgraded from 6.18.0 to 7.0.0, we're seeing RemoteConfig log "Experiment payload could not be parsed as JSON." a bunch of times. I didn't count, but it's probably being logged once for each of our experiments. The result of this is that our A/B tests seem to have completely stopped getting updated values from the server.

#### Steps to reproduce:

It's easy to reproduce in `RCNConfigExperiement.testLoadExperimentFromTable` if you add an additional assertion to verify that the payloads were handled properly...

```objective-c
XCTAssertEqualObjects(_payloadsData, _configExperiment.experimentPayloads);
```

That assertion will fail before the fix and succeeds after the fix (see below)

#### Relevant Code:

I looked around in the `FirebaseRemoteConfig` source code and pinned down the issue (I have a PR ready to go as well).

As part of the change in this pull request (https://github.com/firebase/firebase-ios-sdk/pull/5890/files)

`RCNConfigExperiment.m` changed how it did JSON validation from... 

```Objective-c
NSError *error;
id experimentPayloadJSON = [NSJSONSerialization JSONObjectWithData:experiment options:kNilOptions error:&error];
if (!experimentPayloadJSON || error) {
```

to...

```Objective-c
if (![NSJSONSerialization isValidJSONObject:experiment]) {
```

`isValidJSONObject` is meant to receive an `NSArray` or `NSDictionary` to test if that collection can be converted into JSON data ([source]( https://developer.apple.com/documentation/foundation/jsonserialization/1418461-isvalidjsonobject)). Passing it data causes it to return false every time. The previous implementation is the correct way to test if `NSData` can be converted to JSON.

christibbs:
Fixes #7184

### DIFF
--- a/FirebaseRemoteConfig/CHANGELOG.md
+++ b/FirebaseRemoteConfig/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v7.5.0
+- [fixed] Fixed bug that was incorrectly flagging ABT experiment payloads as invalid. (#7184)
+
 # v7.1.0
 - [changed] Add support for other Firebase products to integrate with Remote Config. (#6692)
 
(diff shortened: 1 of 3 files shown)


paulb777 (approved):
Add a changelog and lgtm. Thanks!

PR #7281, opened 2021-01-11
Logging on device and to Firelog

manjanac:
Only the scaffolding is in place right now, but creating this PR for feedback on how to handle on-device logging and analytics.


FirebaseMLModelDownloader/Sources/ModelDownloaderDeviceLogger.swift
```diff
@@ -0,0 +1,37 @@
...
+extension OSLog {
+  private static let subsystem: String = {
+    let bundleID = Bundle.main.bundleIdentifier ?? ""
+    return "com.google.firebaseml.\(bundleID)"
+  }()
+
+  /// List of logging categories.
+  static let modelDownload = OSLog(subsystem: subsystem, category: "model-download")
+  static let analytics = OSLog(subsystem: subsystem, category: "analytics")
+}
```
ryanwilson, line 37: Hmm, so other SDKs use [`FIRLogger`](https://github.com/firebase/firebase-ios-sdk/blob/c05801f1d127b1d360693fd7c2a8c75a0dea9374/FirebaseCore/Sources/Private/FIRLogger.h#L110-L119) but it's a private API within Core, and we likely won't be able to re-define these in an extension to make them visible since they're C definitions. 

This'll take some thinking of how we want to do this, we eventually want to switch to `os_log` for logging but this would be inconsistent with the rest of the SDKs.

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
```diff
@@ -0,0 +1,101 @@
...
+
+  /// Encode Firelog event for transport.
+  func transportBytes() -> Data {
+    do {
+      let encoder = JSONEncoder()
+      let data = try encoder.encode(event)
+      return data
+    } catch {
+      DeviceLogger.logEvent(
+        level: .error,
```
maksymmalyhin, line 58: nit: This error doesn't seem to affect the users directly and is not actionable by 3P developers, so I would not assign a level higher than `info` or `debug`. 

Also, as Ryan mentioned, we should try to align logging with the rest of Firebase SDK, to make sure that flags like `-FIRDebugEnabled` are respected.

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
```diff
@@ -0,0 +1,101 @@
...
+      target: GDTCORTarget.FLL
+    ) else {
+      return nil
+    }
+    self.fllTransport = fllTransport
+  }
+}
+
+/// Firelog logger.
+class AnalyticsLogger {
```
maksymmalyhin, line 86: optional: `Analytics` name is easy to mix with the Firebase Analytics product. Should we consider a more distinct name like `TelemetryLogger`?

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
```diff
@@ -0,0 +1,101 @@
...
+      return data
+    } catch {
+      DeviceLogger.logEvent(
+        level: .error,
+        category: .analytics,
+        message: "Unable to encode Firelog event.",
+        messageCode: .analyticsEventEncodeError
+      )
+      return Data()
+    }
```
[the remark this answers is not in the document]
manjanac: Hm `transportBytes()` is a method from the `GDTCOREventDataObject` protocol, so it has to return a `Data` object. Is there another way I can do this?

FirebaseMLModelDownloader/Sources/ModelDownloaderDeviceLogger.swift
manjanac: Ah yes I saw this, but I couldn't figure out how to use this in Swift. Is that possible at this point?

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
manjanac: Great idea, changed!

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
manjanac: Ah sounds good, I updated it to `debug`.

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
```diff
@@ -0,0 +1,101 @@
...
+
+/// Firelog event.
+struct FBMLEvent: Codable {
+  var eventName: String
+  var systemInfo: FBMLSystemInfo
+  var eventDescription: String
+  var errorCode: Int?
+}
+
+/// System info.
```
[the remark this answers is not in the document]
manjanac: Good catch, fixed.

FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift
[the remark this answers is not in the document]
manjanac: Does something like this make sense?

In `FBMLDataObject`, a method like -

```
func isEmpty() -> Bool {
    let data = transportBytes()
    if data.isEmpty {
      return true
    }
    return false
  }
```

And then in `logEvent()` -
```
let eventForTransport = fllTransport.eventForTransport()
let dataObject = FBMLDataObject(event: event)
if !dataObject.isEmpty() {
      eventForTransport.dataObject = FBMLDataObject(event: event)
      fllTransport.sendDataEvent(eventForTransport)
} else {
      DeviceLogger.logEvent(
        level: .debug,
        category: .analytics,
        message: "Unable to encode Firelog event.",
        messageCode: .analyticsEventEncodeError
   )
}
 ```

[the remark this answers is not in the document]
Comment (ryanwilson) [reply FirebaseMLModelDownloader/Sources/ModelDownloaderAnalyticsLogger.swift]: Ahh sorry I was unaware it was a protocol requirement. That seems reasonable to me! I don't think we should send something if it's empty, unless you want to have the remote logging when something is incorrect.

File: FirebaseMLModelDownloader/Sources/ModelDownloaderTelemetryLogger.swift [hunk 6]
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/// Firelog event transporter.
+class FBMLEventTransport {
+  private let logSDKName = "FIREBASE_ML_LOG_SDK"
+  let fllTransport: GDTCORTransport
+
+  init?() {
+    guard let fllTransport = GDTCORTransport(
+      mappingID: logSDKName,
```
maksymmalyhin, line 75: Mapping ID should be an integer value corresponding to the Firelog log source enum value. This is basically the way to let Firelog backend know where to route the events.

FirebaseMLModelDownloader/Sources/DeviceLogger.swift
```diff
@@ -0,0 +1,46 @@
...
+  case modelDownloaded
+  case downloadedModelMovedToURL
+  case analyticsEventEncodeError
+  case telemetryInitError
+}
+
+/// On-device logger.
+class DeviceLogger {
+  /// Log event on device.
+  static func logEvent(level: OSLogType, category: OSLog, message: StaticString,
```
maksymmalyhin, line 30: @ryanwilson I wonder if you've had a chance to look into using FIRLogger from the SDK?

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
```diff
@@ -0,0 +1,163 @@
...
+import Foundation
+import FirebaseCore
+import GoogleDataTransport
+
+/// Extension to set Firebase app info.
+extension SystemInfo {
+  mutating func setAppInfo(apiKey: String?, projectID: String?) {
+    appID = Bundle.main.bundleIdentifier ?? "unknownBundleID"
+    // appVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "unknownAppVersion"
+    appVersion = "1"
```
maksymmalyhin, line 24: I wonder if Firebase SDK version is more important than the app version which don't control and cannot make much sense on most of the cases? 

Firebase version is available via `FirebaseVersion()` func.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
```diff
@@ -0,0 +1,163 @@
...
+
+  init(event: FirebaseMlLogEvent) {
+    self.event = event
+  }
+
+  /// Encode Firelog event for transport.
+  func transportBytes() -> Data {
+    do {
+      // TODO: Should this be binary or json serialized?
+      let data = try event.serializedData()
```
maksymmalyhin, line 86: This is the data that will be sent to the server via Firelog. I'm a bit confused why we add `SwiftProtobuf` dependency if we serialize data with JSON. We either need to serialize data with protobuf here or remove any protobuf related code as not used. Could you please fix or add a TODO.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
```diff
@@ -0,0 +1,163 @@
...
+      return nil
+    }
+    self.fllTransport = fllTransport
+  }
+
+  /// Log events to Firelog.
+  private func logModelEvent(event: FirebaseMlLogEvent) {
+    let eventForTransport: GDTCOREvent = fllTransport.eventForTransport()
+    eventForTransport.dataObject = FBMLDataObject(event: event)
+    eventForTransport.qosTier = .qoSFast
```
maksymmalyhin, line 135: I assume it is needed for debugging only. Could you please add a TODO to update QoS.

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
```diff
@@ -362,4 +362,57 @@ final class ModelDownloaderIntegrationTests: XCTestCase {
...
+      case let .success(model):
+        guard let telemetryLogger = TelemetryLogger(
+          isStatsEnabled: true,
+          apiKey: testApp.options.apiKey,
+          projectID: testApp.options.projectID
+        ) else {
+          XCTFail("Could not initialize logger.")
+          return
+        }
+        telemetryLogger.logModelDownloadEvent(
```
maksymmalyhin, line 405: I assume it is a temporary code for testing/debugging, correct? If yes, could you please add a TODO. If no, could you please explain what's the purpose of calling telemetryLogger here.

maksymmalyhin (approved):
LGTM, a couple comments/questions but no blockers for the PR.

FirebaseMLModelDownloader.podspec
```diff
@@ -33,6 +33,7 @@ Pod::Spec.new do |s|
   s.framework = 'Foundation'
   s.dependency 'FirebaseCore', '~> 7.0'
   s.dependency 'FirebaseInstallations', '~> 7.0'
+  s.dependency 'SwiftProtobuf', '~> 1.0'
```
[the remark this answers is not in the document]
manjanac: Good catch, done!

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
manjanac: On Android, this is the version of the developer's app so I'm logging that at this point for consistency. I'll add a TODO here to revisit logging this.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
manjanac: So at this point, the event is serialized using SwiftProtobuf methods. It can be converted to binary or json (I'm using binary). Ultimately I'll rewrite this to be compatible with whatever replacement we end up using for SwiftProtobuf.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
manjanac: Good catch! Fixed.

FirebaseMLModelDownloader/Tests/Integration/ModelDownloaderIntegrationTests.swift
manjanac: Yep good catch, this was for testing. I'll remove it before we launch.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
maksymmalyhin: Ah, sorry, my bad, I misread the code! All looks good here.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
maksymmalyhin: I'm curious how it is used. I guess, it may be used for A/B testing? If so, then makes sense. The only nit - a constant [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) should be used instead of the string literal.

maksymmalyhin (approved):
LGTM assuming TODOs and Swift PM CI to be fixed in the following PRs.

maksymmalyhin:
> LGTM assuming TODOs and Swift PM CI to be fixed in the following PRs.

Ah, well, actually `spm / swift-build-run` needs to be fixed in the PR as otherwise it will break other PRs. You may either try to update the package definition or disable it temporary.

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
manjanac: Thanks I'll look into this some more. For now I've added a TODO.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -72,6 +74,12 @@ public class ModelDownloader {
     appName = app.name
     options = app.options
     installations = Installations.installations(app: app)
+    /// Respect Firebase-wide data collection setting.
+    telemetryLogger = TelemetryLogger(
+      isStatsEnabled: app.isDataCollectionDefaultEnabled,
```
ryanwilson, line 79: `isDataCollectionDefaultEnabled` is a runtime flag that could be changed at any point. It's likely better to pass the app to `TelemetryLogger` and have it read from the flag each time it's about to send data. Otherwise, a user could opt out of logging and we'd still be sending things.

ryanwilson (commented):
Just one thing then LGTM!

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Great catch! Updated.

FirebaseMLModelDownloader/Sources/DeviceLogger.swift
ryanwilson: Yes, we're going to follow up and try to use `GULLogger` eventually since it's a public API

ryanwilson (approved):
LGTM on Green!

PR #7284, opened 2021-01-11
Firebase pod fix for tvOS and macOS

paulb777:
When the Firebase pod was installed and other Firebase pods installed directly (not via subspec with `Firebase/`, the Firebase module was not getting installed for tvOS and macOS. 

This should be a rare bug in practice, but it was exposed by a crash in the ZipBuilder when Firebase.h was not installed after a tvOS installation.

#no-changelog


ryanwilson (approved):
Seems reasonable to me!

PR #7286, opened 2021-01-11
ZipBuilder workaround

paulb777:
Workaround for crash exposed by different tvOS installation. Fixed in #7284 

Keep the workaround longer term to be conservative about preferring the installation form the first platform (iOS) instead of the last (tvOS).

#no-changelog

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
@@ -244,7 +244,12 @@ struct ZipBuilder {
           }
         }
         // Union all pods built across platforms.
-        podsBuilt[podName] = podInfo
+        // Be conservative and favor iOS if it exists - and workaround
+        // bug where Firebase.h doesn't get installed for tvOS and macOS.
+        // Fixed in #7284.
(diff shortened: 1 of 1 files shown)


ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -244,7 +244,12 @@ struct ZipBuilder {
           }
         }
         // Union all pods built across platforms.
-        podsBuilt[podName] = podInfo
+        // Be conservative and favor iOS if it exists - and workaround
+        // bug where Firebase.h doesn't get installed for tvOS and macOS.
+        // Fixed in #7284.
+        if podsBuilt[podName] == nil {
+          podsBuilt[podName] = podInfo
+        }
```
ryanwilson, line 252: This will only work if iOS is specified first, right?

ryanwilson (approved):

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
paulb777: Right.

PR #7288, opened 2021-01-11
Iterate through all elements of target array for FieldValue::ArrayRemove

thebrianchen:
The previous implementation only deleted the first instance of the ArrayRemove element appearing in the document array. This PR removes all appearances of the ArrayRemove element from the document array.

#no-changelog

### DIFF
--- a/Firestore/core/src/model/transform_operation.cc
+++ b/Firestore/core/src/model/transform_operation.cc
@@ -251,8 +251,8 @@ FieldValue ArrayTransform::Rep::Apply(
       }
     } else {
       HARD_ASSERT(type_ == Type::ArrayRemove);
-      for (int i = 0; i < static_cast<int>(result.size());) {
-        if (type_ == Type::ArrayRemove && element == result.at(i)) {
+      for (size_t i = 0; i < result.size();) {
+        if (element == result.at(i)) {
           result.erase(result.cbegin() + i);
         } else {
           ++i;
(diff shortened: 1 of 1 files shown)


Firestore/core/src/model/transform_operation.cc
```diff
@@ -251,8 +251,12 @@ FieldValue ArrayTransform::Rep::Apply(
       }
     } else {
       HARD_ASSERT(type_ == Type::ArrayRemove);
-      if (pos != result.end()) {
-        result.erase(pos);
+      for (int i = 0; i < static_cast<int>(result.size());) {
```
schmidt-sebastian, line 254:
```suggestion
      for (size_t i = 0; i < result.size();) {
```

Firestore/core/src/model/transform_operation.cc
```diff
@@ -251,8 +251,12 @@ FieldValue ArrayTransform::Rep::Apply(
...
+        if (type_ == Type::ArrayRemove && element == result.at(i)) {
```
schmidt-sebastian, line 255: The first part of this is always true, isn't it?

schmidt-sebastian (approved):

Firestore/core/src/model/transform_operation.cc
thebrianchen: Done.

Firestore/core/src/model/transform_operation.cc
thebrianchen: Oops, forgot to remove.

PR #7294, opened 2021-01-12
[Remove IID] Move some token unit tests from IID to Messaging

chliangGoogle:
Moving a few unit tests from InstanceID to Messaging after the refactor.
- APNSTokenInfoTest
- TokenManagerTest
- AuthKeychainTest

Simplify some init methods as there's no need to inject some classes as they can be initialized inside.
Also since we updated clang-format so there are a few nonmessaging methods are formatted in the branch.


FirebaseMessaging/Sources/Token/FIRMessagingCheckinStore.m
```diff
@@ -17,16 +17,18 @@
 #import "FirebaseMessaging/Sources/Token/FIRMessagingCheckinStore.h"
 
 #import "FirebaseMessaging/Sources/FIRMessagingCode.h"
+#import "FirebaseMessaging/Sources/FIRMessagingConstants.h"
 #import "FirebaseMessaging/Sources/FIRMessagingLogger.h"
 #import "FirebaseMessaging/Sources/FIRMessagingUtilities.h"
 #import "FirebaseMessaging/Sources/NSError+FIRMessaging.h"
-#import "FirebaseMessaging/Sources/Token/FIRMessagingAuthKeyChain.h"
+#import "FirebaseMessaging/Sources/Token/FIRMessagingAuthKeychain.h"
```
maksymmalyhin, line 24: I'm curious if using [GULKeychainUtils](https://github.com/google/GoogleUtilities/blob/main/GoogleUtilities/Environment/Public/GoogleUtilities/GULKeychainUtils.h) instead of `FIRMessagingAuthKeychain` may save some extra code and tests?

maksymmalyhin (approved):
LGTM

FirebaseMessaging/Sources/Token/FIRMessagingCheckinStore.m
chliangGoogle: Good call. I will take a look at.

maksymmalyhin: Let me know if you need any help with that.

chliangGoogle: yeah maybe we should work on separate PR. Feel free to checkin in chen/fm-master while I keep adding more unit tests.

maksymmalyhin: sure

PR #7298, opened 2021-01-14
Twitter dont block main on file ops

jhollida24:
Hello. We see the code referenced here blocking the main thread at startup and causing our watchdog system to kill the app. Synchronously waiting for filesystem operations to complete from the main thread can be safely removed.

image (3)

### DIFF
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
@@ -410,10 +410,16 @@ static void (^reportSentCallback)(void);
   NSOperationQueue *__weak queue = _operationQueue;
   FBLPromise *__weak unsentReportsHandled = _unsentReportsHandled;
   promise = [promise then:^id _Nullable(NSNumber *_Nullable value) {
-    [queue waitUntilAllOperationsAreFinished];
-    // Signal that to callers of processReports that everything is finished.
-    [unsentReportsHandled fulfill:nil];
-    return value;
+      FBLPromise *allOpsFinished = [FBLPromise pendingPromise];
+      [queue addOperationWithBlock:^{
+          [allOpsFinished fulfill:nil];
+      }];
+
+      return [allOpsFinished onQueue:dispatch_get_main_queue() then:^id _Nullable(id  _Nullable allOpsFinishedValue) {
+          // Signal that to callers of processReports that everything is finished.
+          [unsentReportsHandled fulfill:nil];
+          return value;
+      }];
   }];
 
   return promise;
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
@@ -423,7 +423,7 @@ static void (^reportSentCallback)(void);
     }];
   }];
 
-    return promise;
+  return promise;
 }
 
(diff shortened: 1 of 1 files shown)


jhollida24:
@googlebot I signed it!

jhollida24:
Hi! We'd like to get a build of Firebase into the Twitter app release that cuts its branch this week. We can cherry-pick it if necessary, but I thought I would give a friendly nudge here, first. We have another PR here that the same applies to: https://github.com/firebase/firebase-ios-sdk/pull/7302

Thanks!

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -410,10 +410,18 @@ - (FBLPromise *)deleteUnsentReports {
   NSOperationQueue *__weak queue = _operationQueue;
   FBLPromise *__weak unsentReportsHandled = _unsentReportsHandled;
   promise = [promise then:^id _Nullable(NSNumber *_Nullable value) {
```
maksymmalyhin, line 412: optional: I wonder if a version below would be a bit more readable?

```
promise = [promise onQueue:dispatch_get_global_queue(QOS_CLASS_UTILITY, 0) then:^id _Nullable(NSNumber *_Nullable value) {
  [queue waitUntilAllOperationsAreFinished];
  return value;
}
.then(^id _Nullable(NSNumber *_Nullable value) {
  [unsentReportsHandled fulfill:nil];
  return value;
})
```

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -410,10 +410,18 @@ - (FBLPromise *)deleteUnsentReports {
...
-    [queue waitUntilAllOperationsAreFinished];
```
maksymmalyhin, line 413: @samedson @elenadoty Do you remember if blocking UI thread is intentional here or just an accident?

maksymmalyhin (approved):
@jhollida24 Thank you for your contribution! The change looks good to me. 

@samedson @elenadoty Could you please provide review from your end.

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
jhollida24: I'm happy to make changes that you would like, but I think the GCD folks at Apple would discourage this. Dispatching work to the global concurrent queues that blocks causes dispatch worker threads to be created (for the next block coming through), which is then torn down when it's unblocked. An additional operation on an existing (serial) queue doesn't expand the total concurrency.

jhollida24: I do agree with you it's more readable, and it's nice that it doesn't rely on the automatic flatMap behavior of FBLPromise.

maksymmalyhin: Good point on blocking a global queue - agreed it's not the best idea. Current version looks good to me from logic perspective. I'll think how we can improve readability and comment if I have any ideas.

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
samedson: I believe this was an oversight. It shouldn't be a problem to delete in the background

samedson (approved):

PR #7302, opened 2021-01-14
Add Crashlytics bulk key/value logging API

jhollida24:
The key/value logging API atomically writes out changes to
disk. If you set several keys and values consecutively, this
generates a large number of file io operations.

Add an API that accepts an NSDictionary of keys and values
that are written together, reducing the number of file io
ops.

### DIFF
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
@@ -75,6 +75,7 @@ void FIRCLSUserLoggingInit(FIRCLSUserLoggingReadOnlyContext* roContext,
 
 #ifdef __OBJC__
 void FIRCLSUserLoggingRecordUserKeyValue(NSString* key, id value);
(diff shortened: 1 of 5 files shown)


jhollida24:
bump

elenadoty:
Hi @jhollida24,
Thanks for the contribution! I agree that this would be helpful. Since this adds a method to the public API, it would need to go through our internal API review process. I'll run it by our team and let you know if this is something we have the bandwidth to move forward now.

jhollida24:
@elenadoty The Twitter app sets several key/value pairs often, with the most important example being at startup. Our watchdog reporter shows us stack traces where we stall out and end up crashing because of the sync file IO being done here, especially on older devices. I'm happy to elaborate or make any changes you would like to the API.

elenadoty:
@jhollida24 Thanks for the context! I've started the process of review and will let you know if I need any additional information.

samedson (approved):
🚢  it! Thank you @jhollida24 for the contribution!

elenadoty:
+1, thanks @jhollida24! I'll go ahead and merge this.

PR #7303, opened 2021-01-14
Create leveldb "table"s for bundles

wu-hui:


Firestore/core/src/local/leveldb_key.cc
```diff
@@ -596,6 +612,14 @@ class Writer {
     WriteLabeledString(ComponentLabel::DocumentId, document_id);
   }
 
+  void WriteBundleId(absl::string_view bundle_id) {
```
wilhuff, line 615: Preserve relative ordering and put this after `WriteSnapshotVersion`?

Firestore/core/src/local/leveldb_key.h
```diff
@@ -635,6 +643,76 @@ class LevelDbRemoteDocumentReadTimeKey {
   model::SnapshotVersion read_time_;
 };
 
+/**
+ * A key in the bundles table, storing the bundle Id for each entry.
+ */
+class LevelDbBundlesKey {
```
wilhuff, line 649: Even though the table is plural ("bundles"), each individual entry in it is singular ("bundle"). This should be a `LevelDbBundleKey` because the key addresses a specific bundle.

Firestore/core/src/local/leveldb_key.h
```diff
@@ -635,6 +643,76 @@ class LevelDbRemoteDocumentReadTimeKey {
...
+  }
+
+ private:
+  std::string bundle_id_;
+};
+
+/**
+ * A key in the named_queries table, storing the query name for each entry.
+ */
+class LevelDbNamedQueriesKey {
```
wilhuff, line 684: Similar to above. The table is plural because it contains many keys, but each key points to a specific named query. This should be `LevelDbNamedQueryKey`.

Firestore/core/src/local/leveldb_key.h
```diff
@@ -635,6 +643,76 @@ class LevelDbRemoteDocumentReadTimeKey {
...
+ public:
+  /**
+   * Creates a key prefix that points just before the first key of the table.
+   */
+  static std::string KeyPrefix();
+
+  /**
+   * Creates a key that points to the key for the given query name.
+   */
+  static std::string Key(absl::string_view bundle_id);
```
wilhuff, line 694: The parameter name here should be something other than `bundle_id`. Perhaps `query_name`?

Firestore/core/src/local/leveldb_key.h
```diff
@@ -635,6 +643,76 @@ class LevelDbRemoteDocumentReadTimeKey {
...
+  /** The query name for this entry. */
+  const std::string& name() const {
+    return name_;
+  }
+
+ private:
+  std::string name_;
+};
+
 }  // namespace local
```
wilhuff, line 716: Please add a test that verifies that the description works. Something like this:

https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/test/unit/local/leveldb_key_test.cc#L419

What you'll find, I think, is that you're missing changes to `Reader::Describe` for the new components you've added: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/local/leveldb_key.cc#L465.

Firestore/core/src/local/leveldb_key.cc
wu-hui: Done.

Firestore/core/src/local/leveldb_key.h
wu-hui: Done.

Firestore/core/src/local/leveldb_key.h
wu-hui: Done.

Firestore/core/src/local/leveldb_key.h
wu-hui: Done.

Firestore/core/src/local/leveldb_key.h
wu-hui: Done.

wilhuff (approved):
LGTM

Issue #7304, opened 2021-01-14
Symbol not found: _objc_opt_respondsToSelector


jostster:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.2
  * Firebase SDK version: 7.0
  * Installation method: `Swift Package Manager`
  * Firebase Component: Instance, Core, Database, Messaging, Storage, Crashlytics)

### [REQUIRED] Step 2: Describe the problem
When running crashlytics on watchOS 5.0 it crashes with `Symbol not found: _objc_opt_respondsToSelector` even though the selector should exist. While Crashlytics is only supported for watchOS 6.0, if an application also supports 5.0, all 5.0 devices will get this crash.

#### Steps to reproduce:

What happened? How can we make the problem occur?
Run the crashlytics test watchos application on a 5.3 simulator

#### Relevant Code:
Even with checks in place to not run the Firebase configureApp call, the mach-o load will still run all `+(void)load` calls which trigger this error.

#### Fix:
Wrap the `+(void)load` in checks to make sure it is only running on the min version or greater.

paulb777:
@jostster Thanks for the issue and PR. This approach doesn't seem right to me, however.   The Package.swift specifies the minimum versions [here](https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L29) and that should prevent the installation for any apps that require watchOS 5.  

As described, this sounds like a Swift Package Manager issue, and I don't think we should change the Firebase sources for it.

jostster:
Hello @paulb777,
I have a test project that uses cocoapods version and am able to replicate this with cocoapods as well

paulb777:
Hmm, watchOS must handle versioning differently than other platforms which don't allow apps with a higher minimum version to be installed. 

It seems that if that's the case, we should assert in one place in FirebaseCore since there could be other issues with watchOS 6 APIs not supported that were assumed by the build system.

Are you able to share a reproducible example?

jostster:
Attached is my test ios app with a watchos extension. I have removed the GoogleInfo.plist file.
[watch-crashlytics copy.zip](https://github.com/firebase/firebase-ios-sdk/files/5823333/watch-crashlytics.copy.zip)

paulb777:
Thanks! I've reproduced it. I still think this is an Xcode watchOS bug and will chat with more of the team next week to get additional input on the right approach.

paulb777:
[...]

PR #7305, opened 2021-01-14
Fix for #7304

jostster:
This is a fix for issue #7304 . While it fixes the issue, it could cause issues with the individual components since the define is in a crashlytics define file. There is probably a better place to put this. Open to input.


morganchen12 (commented):
It seems unnecessary to gate initialize IID, Installations, GDT, and Storage behind the Crashlytics macro you've defined. Since the crash is in Crashlytics, wouldn't disabling only Crashlytics be sufficient? If not, the macro will likely need to be moved to Core or GoogleUtilities.

Adding Paul and Sam for their perspectives as well.

jostster:
I agree it seems weird, but since all the other components also have a `+(void)load` those are automatically ran. If there is a universal define file I would prefer to move it there.

paulb777:
I'm going to close this for now while the approach and whether or not it should be addressed is done in #7304.

PR #7306, opened 2021-01-14
Track conversions on in-app message campaigns without a clickthrough URL

christibbs:
Conversion tracking wasn't working for campaigns with no click-through URL. This change tracks conversions on these campaigns when the message is shown.


FirebaseInAppMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2021-1 -- v7.5.0
+- [fixed] Fixed conversion tracking for in-app messages without an action URL (#7306).
```
prakhar1989, line 2: I'd say this should be 

```
Fixed conversion tracking for in-app messages with a conversion event but not a button / action URL
```

christibbs: Updated in 2155c7e.

FirebaseInAppMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2021-1 -- v7.5.0
+- [fixed] Fixed conversion tracking for in-app messages with a conversion event but not a button / action URL (#7306).
```
paulb777, line 2: Does this fix an open issue? If so, please use the issue number instead of the PR and also add to the PR description.

paulb777 (approved):

FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -232,13 +236,25 @@ - (void)impressionDetectedForMessage:(FIRInAppMessagingDisplayMessage *)inAppMes
...
+          logConversionTrackingEventForCampaignID:_currentMsgBeingDisplayed.renderData.messageID];
+    }
   } else {
     FIRLogDebug(kFIRLoggerInAppMessaging, @"I-IAM400011",
                 @"A test message. Record the test message impression event.");
     return;
   }
 }
 
+- (BOOL)shouldTrackConversionsOnImpressionForCurrentInAppMessage:
```
prakhar1989, line 251: Shouldn't this also include that the message should have conversion event defined?

If it doesn't we don't need to track it.

FirebaseInAppMessaging/CHANGELOG.md
prakhar1989: here's the bug id: b/176820352

paulb777: I meant GitiHub issue. If there isn't one, the PR is fine here.

FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
christibbs: We weren't checking on a conversion event previously, from [comments](https://osscs.corp.google.com/firebase-sdk/firebase-ios-sdk/+/master:FirebaseInAppMessaging/Sources/Analytics/FIRIAMAnalyticsEventLoggerImpl.m;drc=b41a1273368e8af2af26d1dfcb135fc2307b94b3;l=126) (and piecing together the code) it looks like we currently always track the conversion and then presumably handle reporting on the backend. Do you think we need to check for this now?

christibbs: We're waiting on Analytics to confirm how to proceed.

PR #7307, opened 2021-01-15
Implement model download conditions.

manjanac:
Model download conditions can be set by the developer and influence how the model is downloaded. 
The previous MLKit SDK supported: 1) allowsCellularAccess, and 2) allowsBackgroundDownloading.

In this PR, I've attempted to implement both these conditions (without handling errors). Happy to discuss the PR and any feedback!


FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -43,33 +43,54 @@ class ModelDownloadTask: NSObject {
...
+  /// Optional background completion handler.
+  private var backgroundCompletion: (() -> Void)?
   /// URLSession to handle model downloads.
-  private lazy var downloadSession = URLSession(configuration: .ephemeral,
-                                                delegate: self,
-                                                delegateQueue: nil)
+  private lazy var downloadSession: URLSession = {
+    var configuration = URLSessionConfiguration.ephemeral
+    if conditions.allowsBackgroundDownloading {
+      configuration = URLSessionConfiguration
```
manjanac, line 64: Is this less secure than `.ephemeral`?

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -163,3 +201,22 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
...
+                   handleEventsForBackgroundURLSession identifier: String,
+                   completionHandler: @escaping () -> Void) {
+    assert(identifier == backgroundSessionIdentifier)
+    backgroundCompletion = completionHandler
+  }
+
+  /// Call background completion handler when resumed from background.
+  func urlSessionDidFinishEvents(forBackgroundURLSession session: URLSession) {
+    DispatchQueue.main.async {
+      guard let backgroundCompletionHandler = self.backgroundCompletion else { return }
```
maksymmalyhin, line 218: nit: I think `backgroundCompletionHandler?()` is sufficient here.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
maksymmalyhin: Yes, it may save unencrypted request/response in cache database in the app sandbox. It is not necessarily a bit issue but is safer to avoid if possible.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -43,33 +43,54 @@ class ModelDownloadTask: NSObject {
...
+  private var backgroundCompletion: (() -> Void)?
   /// URLSession to handle model downloads.
-  private lazy var downloadSession = URLSession(configuration: .ephemeral,
-                                                delegate: self,
-                                                delegateQueue: nil)
+  private lazy var downloadSession: URLSession = {
+    var configuration = URLSessionConfiguration.ephemeral
+    if conditions.allowsBackgroundDownloading {
+      configuration = URLSessionConfiguration
+        .background(withIdentifier: backgroundSessionIdentifier)
```
maksymmalyhin, line 65: If I understand the [docs](https://developer.apple.com/documentation/foundation/url_loading_system/downloading_files_in_the_background) correctly, if the app is terminated while an ongoing download, then creating a url session with the same ID will associate the download tasks scheduled before termination with the session (see "Recreate the Session If the App Was Terminated" section). The code looks not ready to handle this case.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -163,3 +201,22 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
     }
   }
 }
+
+/// Extension to handle background downloading.
+extension ModelDownloadTask: UIApplicationDelegate {
```
maksymmalyhin, line 206: These methods will need to be somehow called from the app delegate (e.g. directly by the app dev or via GULAppDelegateSwizzler, etc.).

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -163,3 +201,22 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
...
+  /// Save completion handler for background download.
+  func application(_ application: UIApplication,
+                   handleEventsForBackgroundURLSession identifier: String,
+                   completionHandler: @escaping () -> Void) {
+    assert(identifier == backgroundSessionIdentifier)
```
maksymmalyhin, line 211: If the host app also uses the background downloading, then we will crash (at least in Debug) when handling events for the host app URL session.

maksymmalyhin (commented):
Proper support of the background URL session will require more work and make the code more complicated. IMO, the scope may be big enough to postpone supporting this feature for the 1st SDK version and add it in the following versions.

manjanac:
> Proper support of the background URL session will require more work and make the code more complicated. IMO, the scope may be big enough to postpone supporting this feature for the 1st SDK version and add it in the following versions.

Agreed, I'm removing support for background downloads for now. This can be something we can reintroduce in future versions of the SDK based on customer requests.

FirebaseMLModelDownloader/Sources/ModelDownloadConditions.swift
```diff
@@ -15,8 +15,7 @@
 import Foundation
 
 /// Model download conditions.
-// TODO: Implement model download conditions.
 public struct ModelDownloadConditions {
-  // TODO: Intentionally left blank until implementation.
-  public init() {}
+  /// Allow model downloading on a cellular connection. Default is `YES`.
+  public var allowsCellularAccess = true
```
maksymmalyhin, line 20: Do we have API review for this?

maksymmalyhin (approved):
LGTM, a question on API review for public API change.

FirebaseMLModelDownloader/Sources/ModelDownloadConditions.swift
manjanac: We have a review scheduled next week for this. The original approved API didn't have any conditions (or the init).

PR #7311, opened 2021-01-15
Remove dynamic framework count from diagnostics.

### LINKED ISSUE #7310: Deadlock caused by call to objc_copyImageNames in FIRPopulateProtoWithNumberOfLinkedFrameworks
### Environment

  * Xcode version: 12.1
  * Firebase SDK version: 6.34.0
  * Firebase Component: Crashlytics

### Problem

Firebase's call to `objc_copyImageNames` within `FIRPopulateProtoWithNumberOfLinkedFrameworks` can sometimes trigger a deadlock on launch (if another thread is performing a `dlopen`, e.g. CFNetwork) resulting in the app being terminated by the system watchdog. This has been filed as a bug report with Apple (FB8971497) and discussed on the developer forums [here ](https://developer.apple.com/forums/thread/671497?answerId=657367022&page=1#657354022) however it's worth raising here as well to see if there's any workarounds.

Is it possible to disable diagnostics including the number of linked frameworks to remove the call to `objc_copyImageNames`? 

#### Steps to reproduce:

In an app that makes a network request at launch, with a reasonable number of dynamic libraries linked (~5+), it may trigger a deadlock (timing dependant) resulting in the app being terminated by the system watchdog. Crash report can be found [here](https://gist.github.com/SquaredTiki/11a58e6837029e44da6098b34486658d).

ryanwilson:
This was previously used to help determine if Firebase should switch to
using dynamic frameworks, but now developers have the option themselves.

Fixes #7310

### DIFF
--- a/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
+++ b/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
@@ -454,29 +454,6 @@ void FIRPopulateProtoWithInstalledServices(logs_proto_mobilesdk_ios_ICoreConfigu
   config->sdk_service_installed_count = (int32_t)sdkServiceInstalledArray.count;
 }
 
-/** Populates the proto with the number of linked frameworks.
- *
- * @param config The proto to populate.
- */
-void FIRPopulateProtoWithNumberOfLinkedFrameworks(
-    logs_proto_mobilesdk_ios_ICoreConfiguration *config) {
-  int numFrameworks = -1;  // Subtract the app binary itself.
-  unsigned int numImages;
-  const char **imageNames = objc_copyImageNames(&numImages);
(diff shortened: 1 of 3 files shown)


maksymmalyhin (approved):

paulb777:
Please add a FirebaseCore changelog that references the issue.

Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
```diff
@@ -37,7 +37,7 @@
 
 #ifdef FIREBASE_BUILD_ZIP_FILE
 static BOOL kUsingZipFile = YES;
-#else  // FIREBASE_BUILD_ZIP_FILE
+#else   // FIREBASE_BUILD_ZIP_FILE
```
ryanwilson, line 40: This was unrelated to my change but is part of the clang format 11.0.1 update

PR #7313, opened 2021-01-15
[#7309] Fixed messageDisplaySuppressed Ignored on Slow Networks

### LINKED ISSUE #7309: IAM ignores messageDisplaySuppressed on slow networks

  * Xcode version: 12.3
  * Firebase SDK version: 7.4.0
  * Firebase Component: IAM

On slow networks, the In-App Messaging component appears to ignore the `messageDisplaySuppressed` flag.

Consider the following scenario:

1. An in-app message is configured to display as soon as the user has opened the chat screen;
2. When the on-screen keyboard is displayed, messages must never appear. To implement that behavior, the app receives `keyboardWillShowNotification` and `keyboardDidHideNotification`, and sets IAM’s `messageDisplaySuppressed` property to `true` and `false`, respectively;
3. On slow networks, it takes IAM significant time to fetch a message’s image;
4. In that time, the user is able to open the keyboard, which results in `messageDisplaySuppressed` changed to `true` (`Message display suppress set to 1` is printed to the console);
5. When the image is fetched, IAM ignores the updated `messageDisplaySuppressed` value and displays the message anyway.

These steps result in the message—and most crucially the button used to dismiss it—half-hidden beneath the keyboard. The keyboard itself cannot be dismissed, too, since the message’s semi-transparent background layer does not pass taps through to the chat view controller. This leaves force-quitting the app as the only way to resolve the issue.

The expected / acceptable outcomes are:
* The message is triggered *and displayed* before the user is able to open the keyboard (that’s how it currently works on fast networks);
* If the user opens the keyboard before the message appears, the latter should not display.

The quick fix I see is adding another check on `messageDisplaySuppressed`, performed *just before* the message appears on screen. (I think I can implement and submit such a fix.) An alternative solution, which seems nicer, would be extending `InAppMessagingDisplayDelegate ` with a `shouldDisplayMessage(_:) -> Bool` method called when the message is about to display.

yakovmanshin:
[Closes #7309]

* Added an extra check on `messageDisplaySuppressed`’s value just before the message is displayed on screen
* Added the `testNoRenderingIfMessageDisplayIsSuppressedDuringImageLoading` test method that covers the scenario where `messageDisplaySuppressed` is set to `YES` after the message’s image started loading but before the message is displayed
* Added support for async image loading in `FIRIAMDisplayExecutorTests`, and a new test message `m6` whose image is loaded asynchronously; existing test messages’ behavior is left unchanged

### DIFF
--- a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
+++ b/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
@@ -638,6 +638,12 @@
 #pragma clang diagnostic pop
           }
         }
+      
+        // On slow networks, image loading may take significant time,
+        // in which the value of `suppressMessageDisplay` could change.
+        if (self.suppressMessageDisplay) {
+          return;
+        }
 
         self.impressionRecorded = NO;
         self.isMsgBeingDisplayed = YES;
--- a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
+++ b/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
@@ -642,6 +642,8 @@
         // On slow networks, image loading may take significant time,
         // in which the value of `suppressMessageDisplay` could change.
         if (self.suppressMessageDisplay) {
+          FIRLogDebug(kFIRLoggerInAppMessaging, @"I-IAM400042",
+                      @"Message display suppressed by developer at message display time.");
           return;
         }
 
--- a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
+++ b/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
@@ -638,7 +638,7 @@
 #pragma clang diagnostic pop
           }
         }
-      
+
         // On slow networks, image loading may take significant time,
         // in which the value of `suppressMessageDisplay` could change.
         if (self.suppressMessageDisplay) {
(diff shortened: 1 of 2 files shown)


yakovmanshin:
@googlebot I signed it!

yakovmanshin:
@christibbs Hello there! Could you have a look at this PR?

christibbs:
@yakovmanshin Sorry for the delay. The change looks good-- could you just add this log before the return on line 645 of FIRIAMDisplayExecutor.m?

```
FIRLogDebug(kFIRLoggerInAppMessaging, @"I-IAM400042",
                  @"Message display suppressed by developer at message display time.");
```

christibbs:
@yakovmanshin Make sure to run `scripts/style.sh` again to reformat!

FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
```diff
@@ -639,6 +639,12 @@ - (void)displayForMessage:(FIRIAMMessageDefinition *)message
           }
         }
 
+        // On slow networks, image loading may take significant time,
+        // in which the value of `suppressMessageDisplay` could change.
+        if (self.suppressMessageDisplay) {
+          return;
```
christibbs, line 645: Add log here:

```
FIRLogDebug(kFIRLoggerInAppMessaging, @"I-IAM400042",
                  @"Message display suppressed by developer at message display time.");
```

christibbs (changes requested):

FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m
yakovmanshin: Sure!

christibbs (approved):

christibbs:
@yakovmanshin Thanks for contributing to Firebase.

PR #7315, opened 2021-01-15
Remove GoogleUtilities from repo

paulb777:
Removes GoogleUtilities from the repo now that it has been moved to https://github.com/google/GoogleUtilities

Pre-release integration testing with Firebase should now be done via the SpecsStaging repo or path specifications in Podfiles.

#no-changelog


paulb777:
TODO: figure out cmake rule adjustments to find GoogleUtilities headers and solve failures like https://github.com/firebase/firebase-ios-sdk/runs/1714309783?check_suite_focus=true#step:6:3169

paulb777:
Rebased after #7324 addressed the cmake build transition

Example/watchOSSample/Podfile
```diff
@@ -1,10 +1,12 @@
+source 'https://github.com/firebase/SpecsStaging.git'
```
ryanwilson, line 1: Is this temporary for now or should it stay forever?

ryanwilson (approved):

Example/watchOSSample/Podfile
paulb777: It should stay forever to support GoogleUtilities integration testing

cmake/GoogleUtilities.cmake
```diff
@@ -34,6 +34,12 @@ firebase_ios_add_framework(
   ${headers} ${sources}
 )
 
+# GoogleUtilities
+target_include_directories(
+  GoogleUtilities PRIVATE
+  $<BUILD_INTERFACE:${FIREBASE_EXTERNAL_SOURCE_DIR}/GoogleUtilities>
```
wilhuff, line 40: `PRIVATE` properties only apply at build time anyway, so using `$<BUILD_INTERFACE:...>` is redundant.

PR #7322, opened 2021-01-19
Separate error messages into extensions

manjanac:
Tried to make errors a little more readable, but I'm open to suggestions to do this better!


FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -89,14 +89,34 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
...
+      DispatchQueue.main.async {
+        self.downloadHandlers
+          .completion(.failure(.internalError(description: ModelDownloadTask
+              .invalidServerResponseErrorDescription)))
+      }
+      return
+    }
+
+    guard (200 ..< 299).contains(response.statusCode) else {
+      return
```
maksymmalyhin, line 117: Would we like to handle it as a download failure?

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -320,3 +329,20 @@ extension ModelDownloader {
...
+  private static let parseModelNameErrorDescription = { (path: String) in
+    "Unable to parse model file name at \(path)."
+  }
+
+  private static let retrieveLocalModelInfoErrorDescription =
+    "Failed to get stored model info for model file."
+  private static let outdatedModelPathErrorDescription = "Outdated model paths in local storage."
+  private static let deletedLocalModelInfoErrorDescription =
+    "Model unavailable due to deleted local model info."
+  private static let backgroundModelDownloadErrorDescription: StaticString =
```
maksymmalyhin, line 346: I'm curious whey `StaticString` here?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -199,6 +194,15 @@ extension ModelInfoRetriever {
     }
   }
 
+  /// Date formatter to parse download URL expiry time.
+  private static func getDateFormatter() -> DateFormatter {
+    let dateFormatter = DateFormatter()
+    dateFormatter.locale = Locale(identifier: "en-US_POSIX")
+    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
```
maksymmalyhin, line 201: nit: Looks like it can be parsed with [ISO8601DateFormatter](https://developer.apple.com/documentation/foundation/iso8601dateformatter). If so, it would be preferable to use ISO8601DateFormatter when available and fallback to manual date format. In this case the fallback will be clearly visible to be removed once we bump supported OS versions.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -209,15 +213,49 @@ extension ModelInfoRetriever {
...
+      size: size,
+      urlExpiryTime: expiryTime
     )
   }
 }
+
+/// Possible error messages for model info retrieval.
+extension ModelInfoRetriever {
+  /// Error descriptions.
+  private static let tokenErrorDescription = "Error retrieving FIS token."
```
maksymmalyhin, line 240: optional: It may be a bit nicer to read if we wrap the descriptions into an `enum` for name spacing and remove `Description` part from the constant names, e.g.:
```
extension ModelInfoRetriever {
  private enum Description {
    static let tokenError = "Error retrieving FIS token."
    ...
  }
}
```

Though I don't have a strong opinion here.

maksymmalyhin (approved):
LGTM

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Yes in an upcoming PR I'm handling this as - if the download failed due to expired download url, retry once and then fail. I can add a TODO here to refer to that!

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: So this is a `os_log` requirement which will go away when I replace it with GULLogger hopefully. Is there a better interim solution I could use?

maksymmalyhin: No, all looks good. I just was curious.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -89,14 +89,34 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
     return "fbml_model__\(appName)__\(remoteModelInfo.name)"
   }
 
+  /// Handle client-side errors.
   func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
-    // TODO: Handle model download url expiry and other download errors
+    assert(task == downloadTask)
```
ryanwilson, line 94: We shouldn't assert here, if anything just log an internal error and return. Never want to crash a customer's app.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
ryanwilson: +1, although likely named `ErrorDescription`

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done, I thought this didn't support milliSeconds but it does with iOS 11+.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Good idea, done.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Gotcha! Added a TODO for logging.

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
```diff
@@ -332,17 +335,20 @@ extension ModelDownloader {
...
-  private static let backgroundModelDownloadErrorDescription: StaticString =
-    "Failed to update model in background."
+    static let retrieveLocalModelInfoError =
+      "Failed to get stored model info for model file."
+    static let outdatedModelPathError = "Outdated model paths in local storage."
+    static let deletedLocalModelInfoError =
+      "Model unavailable due to deleted local model info."
+    static let backgroundModelDownloadError: StaticString =
+      "Failed to update model in background."
+  }
```
ryanwilson: Nit: the `Error` appended at the end isn't really necessary anymore since it's prefixed with `ErrorDescription.`

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -188,19 +189,27 @@ extension ModelInfoRetriever {
...
-  /// Date formatter to parse download URL expiry time.
-  private static func getDateFormatter() -> DateFormatter {
-    let dateFormatter = DateFormatter()
-    dateFormatter.locale = Locale(identifier: "en-US_POSIX")
-    dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
-    dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
-    return dateFormatter
+  /// Parse date from string - used to get download URL expiry time.
+  private static func getDateFromString(_ strDate: String) -> Date? {
+    if #available(iOS 11, *) {
```
ryanwilson, line 201: Nit: also please check which other OS's this is available on, specific versions of macOS, tvOS, watchOS

ryanwilson (commented):
LGTM after 2 nits!

FirebaseMLModelDownloader/Sources/ModelDownloader.swift
manjanac: Gotcha, done. Some of them maybe sound a little vague now, e.g. "ModelInfoRetriever.ErrorDescription.serverResponse" (which is for server responses with HTTP error codes). Is that ok or do you have any suggestions on how to rename something like that? Thanks!

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Good catch, done!

PR #7324, opened 2021-01-19
Use external repo for GoogleUtilities cmake build

paulb777:
cmake companion to #7315 
See also https://github.com/google/GoogleUtilities/pull/3

Thanks to @wilhuff for this guide:

https://github.com/firebase/firebase-ios-sdk/pull/4097 is an example of adding an external dependency. If you ignore all the deleted files, it's quite a small change. Unfortunately, it's not a complete example because abseil-cpp has its own CMake configuration and GoogleUtilities probably shouldn't.

Our CMake configuration for GoogleUtilities isn't a production implementation and it depends on utilities within the firebase-ios-sdk repo to function. Likely this means you don't want to include a CMakeLists.txt in the GoogleUtilities repo because it implies support we wouldn't want to continue.

CMake builds proceed in three stages:

1. Download external sources: governed by the configuration in cmake/external
2. Configure the build: governed by the CMakeLists.txt throughout the repository
3. Actually build using make or ninja.

The second and third steps depend on all the sources existing locally.

These are the essential steps in making this work:

- Add a `cmake/external/GoogleUtilities.cmake` file that describes how to download a tarball of the sources
  - Copy `cmake/external/abseil-cpp.cmake` and adjust to fit.
- Include that new download script in `cmake/external/CMakeLists.txt`
  - Add a line consisting of `include(GoogleUtilities)`.
- Move the framework build for GoogleUtilities out of the GoogleUtilities directory into the cmake directory
  - Remove https://github.com/google/GoogleUtilities/blob/main/GoogleUtilities/CMakeLists.txt
  - Within firebase-ios-sdk, move `GoogleUtilities/CMakeLists.txt` to `cmake/GoogleUtilities.cmake`
  - Change all relative path references to absolute ones relative to `${FIREBASE_EXTERNAL_SOURCE_DIR}/GoogleUtilities`
    - e.g. `Environment/*.m` becomes `${FIREBASE_EXTERNAL_SOURCE_DIR}/GoogleUtilities/Environment/*.m`
 - Adjust the top-level `CMakeLists.txt` to `include(GoogleUtilities)` instead of adding an external subdirectory


FirebaseCore/CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ if(NOT APPLE)
   return()
 endif()
 
+# GoogleUtilities is an external dependency without its own cmake configuration.
+include(GoogleUtilities)
```
wilhuff, line 20: This doesn't seem like it belongs here; GoogleUtilities isn't isolated to just FirebaseCore, is it?

A better place would be to put this before line 265 of the top-level CMakeLists.txt.

paulb777: Putting it there causes the duplicate symbol issues I was seeing before 84ccde5

See https://github.com/firebase/firebase-ios-sdk/pull/7324/checks?check_run_id=1730495590#step:6:6581

Is there a way to link GoogleUtilities only once instead of into all of the intermediate libraries if it's in the top level CMakeLists.txt?

paulb777: Internal bug to investigate moving the include back to the top-level CMakeLists.txt at b/178027829

wilhuff: Let's set this issue aside for the moment to unblock the GoogleUtilities move.

wilhuff (approved):
LGTM

PR #6871, opened 2020-10-29
Allow null/NAN values in filters

thebrianchen:
Porting from [Web](https://github.com/firebase/firebase-js-sdk/pull/3960) and [Android](https://github.com/firebase/firebase-android-sdk/pull/2115).

PR #7327, opened 2021-01-19
Fix a duplicate key in test data

var-const:
This triggers the `-Wobjc-dictionary-duplicate-keys` warning, causing the compilation to fail when `-Werror` is used. Introduced in #6871.

#no-changelog


thebrianchen (approved):
Thanks for catching and fixing!

PR #7329, opened 2021-01-20
Handle expired download URL with support for retries

manjanac:

### Discussion

### Testing

### API Changes

    can discuss it together.


FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -116,7 +124,44 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
     }
 
     guard (200 ..< 299).contains(response.statusCode) else {
-      // TODO: Handle download url expiry + retries.
+      /// Possible failure due to download URL expiry.
+      if response.statusCode == 400 {
+        let currentDateTime = Date()
+        if currentDateTime > remoteModelInfo.urlExpiryTime {
```
ryanwilson, line 130: For readability in the else statement and less indenting (I had to carefully measure where this `else` ended), I wonder if a `guard` statement would be better here too? Or can you not have a nested guard? Haven't tried that before.

```
guard remoteModelInfo.urlExpiryTime > currentDateTime else {
  downloadStatus = .failed
  downloadHandlers.completion(.......)
  return
}

// URL has not expired, try the download again.
```

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -116,7 +124,44 @@ extension ModelDownloadTask: URLSessionDownloadDelegate {
...
+          /// Check if ok to retry download.
+          if numberOfRetries > 0 {
```
ryanwilson, line 132: Same here for guard, if that's allowed in order to reduce the nested statements.

manjanac: Yes it is! I made the change.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Good call, changed.

ryanwilson (approved):
Nice! Thanks for making the change. Please wait for CI, but LGTM

Issue #7295, opened 2021-01-13
Implement Combine support for Firebase


peterfriese:
### SDKs to support
- [ ] Authentication
- [ ] Realtime Database
- [ ] Cloud Firestore
- [ ] Cloud Storage
- [ ] Cloud Functions
- [ ] Machine Learning
- [ ] Cloud Messaging
- [ ] Remote Config
- [ ] Dynamic Links

ziadtamim:
@peterfriese I will support Cloud Storage.

peterfriese:
> @peterfriese I will support Cloud Storage.

Great - thanks, Ziad. I penciled you in - once you send a draft PR, I will add the PR to our project tracker.

paulb777:
@peterfriese Should we close and open separate issue for specific items?

peterfriese:
Yes, we should to that. I'll spend some time next week to clean up the tracker and set up new issues.

mortenbekditlevsen:
I still have a PR for initial Combine support in FirebaseDatabase. Let me know if that's of any use:

https://github.com/firebase/firebase-ios-sdk/pull/7544

ziadtamim:
There is a PR for `Decodable QuerySnapshot publisher` https://github.com/firebase/firebase-ios-sdk/pull/7549

ncooke3:
Closing this as the focus has shifted from Combine to Swift Concurrency.

PR #7331, opened 2021-01-20
Provide namespace for the Auth Combine Publishers

ziadtamim:
### Motiviation 

Define a `namespace` for the Auth publishers to avoid overload issues when the compiler cannot decide which overload to favor. For instance, overloading the `signOut` method for  Combine support leads to issues at compile time.

Screenshot 2021-01-19 at 21 38 44

### DIFF
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -14,14 +14,38 @@
 
 #if canImport(Combine) && swift(>=5.0) && canImport(FirebaseAuth)
 
-  import Combine
-  import FirebaseAuth
+import Combine
+import FirebaseAuth
+import FirebaseCore
 
-  @available(swift 5.0)
-  @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
-  extension Auth {
+@available(swift 5.0)
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+extension Auth {
+    
+    /// Returns a publisher that wraps the auth object for the default Firebase app.
+    ///
+    /// The publisher will emit events on the **main** thread.
+    ///
+    /// - Returns: A publisher that wraps the auth object for the default Firebase app.
+    public class func authPublisher() -> AnyPublisher<Auth, Never> {
+        Just(auth()).eraseToAnyPublisher()
+    }
+    
+    /// Returns a publisher that wraps the auth object for the given Firebase app.
+    ///
+    /// The publisher will emit events on the **main** thread.
+    ///
+    /// - Returns: A publisher that wraps the auth object for the given Firebase app.
+    public class func authPublisher(app: FirebaseApp) -> AnyPublisher<Auth, Never> {
+        Just(auth(app: app)).eraseToAnyPublisher()
+    }
+}
+
+@available(swift 5.0)
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+extension Publisher where Output == Auth, Failure == Never {
     // MARK: - Authentication State Management
-
+    
     /// Registers a publisher that publishes authentication state changes.
     ///
     /// The publisher emits values when:
@@ -34,17 +58,20 @@
     ///
     /// - Returns: A publisher emitting a `User` instance (if the user has signed in) or `nil` (if the user has signed out)
     public func authStateDidChangePublisher() -> AnyPublisher<User?, Never> {
-      let subject = PassthroughSubject<User?, Never>()
-      let handle = addStateDidChangeListener { auth, user in
-        subject.send(user)
-      }
-      return subject
-        .handleEvents(receiveCancel: {
-          self.removeStateDidChangeListener(handle)
-        })
-        .eraseToAnyPublisher()
+        flatMap { auth -> AnyPublisher<User?, Never> in
+            let subject = PassthroughSubject<User?, Never>()
+            let handle = auth.addStateDidChangeListener { _, user in
+                subject.send(user)
+            }
+            
+            return subject
+                .handleEvents(receiveCancel: {
+                    auth.removeStateDidChangeListener(handle)
+                }).eraseToAnyPublisher()
+            
+        }.eraseToAnyPublisher()
     }
-
+    
     /// Registers a publisher that publishes ID token state changes.
     ///
     /// The publisher emits values when:
@@ -58,17 +85,20 @@
     ///
     /// - Returns: A publisher emitting a `User` instance (if a different user as signed in or the ID token of the current user has changed) or `nil` (if the user has signed out)
     public func idTokenDidChangePublisher() -> AnyPublisher<User?, Never> {
-      let subject = PassthroughSubject<User?, Never>()
-      let handle = addIDTokenDidChangeListener { auth, user in
-        subject.send(user)
-      }
-      return subject
-        .handleEvents(receiveCancel: {
-          self.removeIDTokenDidChangeListener(handle)
-        })
+        flatMap { auth -> AnyPublisher<User?, Never> in
+            let subject = PassthroughSubject<User?, Never>()
+            let handle = auth.addIDTokenDidChangeListener { _, user in
+                subject.send(user)
+            }
+            
+            return subject
+                .handleEvents(receiveCancel: {
+                    auth.removeIDTokenDidChangeListener(handle)
+                }).eraseToAnyPublisher()
+        }
         .eraseToAnyPublisher()
     }
-
+    
     /// Sets the currentUser on the calling Auth instance to the provided user object.
     ///
     /// The publisher will emit events on the **main** thread.
@@ -77,20 +107,24 @@
     /// - Returns: A publisher that emits as soon as the user of the calling Auth instance has been
     ///   updated or an error was encountered. The publisher will emit on the *main* thread.
     @discardableResult
-    public func updateCurrentUser(_ user: User) -> Future<Void, Error> {
-      Future<Void, Error> { promise in
-        self.updateCurrentUser(user) { error in
-          if let error = error {
-            promise(.failure(error))
-          } else {
-            promise(.success(()))
-          }
-        }
-      }
+    public func updateCurrentUser(_ user: User) -> AnyPublisher<Void, Error> {
+        setFailureType(to: Error.self)
+            .flatMap { auth in
+                Future<Void, Error> { promise in
+                    auth.updateCurrentUser(user) { error in
+                        if let error = error {
+                            promise(.failure(error))
+                        } else {
+                            promise(.success(()))
+                        }
+                    }
+                }
+            }.eraseToAnyPublisher()
     }
-
+    
+    
     // MARK: - Anonymous Authentication
-
+    
     /// Asynchronously creates and becomes an anonymous user.
     ///
     /// If there is already an anonymous user signed in, that user will be returned instead.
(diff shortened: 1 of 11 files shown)


peterfriese:
Does this only happen for `signOut()`, or did you run into any other methods that expose this isse?

ziadtamim:
This also happens with `useAppLanguage()`. Providing a namespace for the authentication publishers can be beneficial to separate concerns for the current and future added methods.

peterfriese:
> This also happens with `useAppLanguage()`. 
So essentially all calls that don't have closures. 

> Providing a namespace for the authentication publishers can be beneficial to separate concerns for the current and future added methods.
While I agree that this approach makes it easier to separate concerns, it also makes these calls less discoverable.

Let's see if we can come up with alternative solutions.

ziadtamim:
> Let's see if we can come up with alternative solutions.

How about instead of overriding the `auth()` and `auth(app :)` factory methods, we define a computed property that returns a publisher that wraps the auth object for the Firebase application. This approach follows Apple convention for supporting Combine as well as making calls discoverable.

```Swift 
 extension Auth {
    
    public struct Publisher {
      var auth: Auth
    }
    
    /// Returns a publisher that wraps the auth object for the Firebase app.
    public var publisher: Publisher {
        Publisher(auth: self)
    }
  }

extension Auth.Publisher {
    
    public func authStateDidChange() -> AnyPublisher<User?, Never> {
        let subject = PassthroughSubject<User?, Never>()
        let handle = auth.addStateDidChangeListener { _, user in
          subject.send(user)
        }

        return subject
          .handleEvents(receiveCancel: {
            auth.removeStateDidChangeListener(handle)
          }).eraseToAnyPublisher()
    }
}
```

lorenzofiamingo:
But why `signOut ` and `useAppLanguage` should be adopted by Combine? They are not asynchronous. 
Syncrounous functions in Combine should not be flatmapped, but used in `.handleEvents(...)`. 
So if the problem only applies to calls that don't have closures, I don't see the problem...

Anyway (in case I'm missing something), in this specific case the solution proposed could work, but what will happen when this will applied to other frameworks? 
For example in Firestore you should be able to call a publisher from `Firestore` (and this is covered in your approach) but also from `DocumentReference`, `Query`, etc instances should be able to call publishers. I **personally** don't like wrapping every class to adopt combine.

> This approach follows Apple convention for supporting Combine ...

The most used approach to adopt Combine is to add Publisher to the end of methods, but as stated [here](https://github.com/firebase/firebase-ios-sdk/blob/combine-main/FirebaseCombineSwift/DECISIONS.md) it applies for stream of events.

I'd prefer to just add Publisher to the end of overloading methods, anyway, `publisher` and `Publisher` should be named with `publishers` and `Publishers` as they acts like containers.

But, as I said, why are synchronous methods converted in Combine?

ziadtamim:
> Why signOut and useAppLanguage should be adopted by Combine? They are not asynchronous.

I understand that synchronous code is intuitive to understand. Although reactive programming supports both synchronous and asynchronous computations, it is especially useful for facilitating asynchronous computations at a relatively high level of expression.

```swift
for number in [1, 2, 3] {
  print(number)
}
```

Can be translated to Combine as:

```swift            
[1, 2, 3].publisher
  .sink(receiveCompletion: { print($0) },
        receiveValue: { print($0) } )
   .store(in: &cancellables)
```

>  Syncrounous functions in Combine should not be flatmapped, but used in .handleEvents(...).

The `handleEvents` operator is about performing side effects that allow you to catch events in a publisher's lifecycle and then take action at every stage. The intention of supporting `signOut` and `useAppLanguage` is to provide full support of Combine for Firebase and not to restrict the consumer part to a specific use. For instance, providing the flexibility to perform such actions can reduce boilerplate code:

```swift
Auth.auth().publisher
    .signOut()
    .retry(1)
    .receive(on: DispatchQueue.global())
    .handleEvents(receiveCompletion: { _ in
      self.flashInternalData()
    })
    .mapError { AuthError.internalError($0) }
    .receive(on: DispatchQueue.main)
    .sink(receiveCompletion: { _ in /* Do something on signout completion */ },
      receiveValue: { /* Do something */ })
    .store(in: &cancellables)
```

> in this specific case the solution proposed could work, but what will happen when this will applied to other frameworks?
> For example in Firestore you should be able to call a publisher from `Firestore` (and this is covered in your approach) but also from `DocumentReference`, `Query`, etc instances should be able to call publishers.

A proposal that could be suggested for the design of such a case would be to build a reactive chain where each publisher suggests underlying related publishers to meet the requirements. For instance, `FirebaseStorage` could be implemented as follows:

```swift
extension FirebaseStorage.Storage {
    
    public struct Publisher { let storage: Storage }
    
    /// Returns a publisher that wraps storage instance.
    public var publisher: Publisher {
        Publisher(storage: self)
    }
}

extension FirebaseStorage.Storage.Publisher {
    
    /// Creates a FIRStorageReference given a gs:// or https:// URL pointing to
    /// a Firebase Storage location.
    ///
    /// - parameters
    ///     - url: A gs:// or https:// URL to initialize the reference with
    ///
    /// - Returns: An instance of `StorageReference` at the given child path.
    public func reference(forURL url: String) -> StorageReference.Publisher {
        storage.reference(forURL: url).publisher
    }
}

extension FirebaseStorage.StorageReference {
    
    public struct Publisher { let storageReference: StorageReference }
    
    // Returns a publisher that wraps the storage reference.
    fileprivate var publisher: Publisher {
        Publisher(storageReference: self)
    }
}

extension FirebaseStorage.StorageReference.Publisher {
    
    /// Asynchronously retrieves a long lived download URL with a revokable token.
    /// This can be used to share the file with others, but can be revoked by a
    /// developer in the Firebase Console if desired.
    ///
    /// The publisher will emit events on the **main** thread.
    ///
    /// - Returns: A  publisher that retrieve a download URL from a storage reference.
    public func downloadURL() -> Future<URL, Error> {
        Future<URL, Error> { promise in
            storageReference.downloadURL { (url, error) in
                if let error = error {
                    promise(.failure(error))
                } else if let url = url {
                    promise(.success(url))
                }
            }
        }
    }
}
```

Sample use:

```swift
Storage.storage().publisher
    .reference(forURL: "gs://fakeimg.pl/300/")
    .downloadURL()
    .sink(receiveCompletion: { _ in /* Do something */ },
          receiveValue: { _ in /* Do something */ } )
    .store(in: &cancellables)
```

> I **personally** don't like wrapping every class to adopt combine.

I'm not a fan of it either 🙂 but this is a common handy / practice where Apple introduced a special annotation just for it `@propertyWrapper`, it is also applied to build` SwiftUI`.

> The most used approach to adopt Combine is to add Publisher to the end of methods, but as stated [here](https://github.com/firebase/firebase-ios-sdk/blob/combine-main/FirebaseCombineSwift/DECISIONS.md) it applies for stream of events.

Agree on the approach to add a publisher suffix at the end of a method if there are only a few affected and not confusing. In our case, there are several `authStateDidChangePublisher`, `idTokenDidChangePublisher`, `useAppLanguagePublisher`, `signOutPublisher` and **other publishers that don't require adding the publisher suffix**, making it inconsistent and confusing for the consumer.  For instance, Apple applies the publisher suffix model to `URLSession` where only two methods affected `dataTaskPublisher(for:)` and the other ones that don't have a publisher suffix are not actually publishers.

Adding a `Publisher` namespace separates the concerns and assures the client that all underlying methods are in fact publishers.

> I'd prefer to just add Publisher to the end of overloading methods, anyway, `publisher` and `Publisher` should be named with `publishers` and `Publishers` as they acts like containers.

Renaming `Publisher` to `Publishers` would make the type act as strict namespace where it is used as a wrapper and namespace.

lorenzofiamingo:
> Syncrounous functions in Combine should not be flatmapped, but used in .handleEvents(...).

You are right, the right counterpart was `.map(...) `.

> I'm not a fan of it either 🙂 but this is a common handy / practice where Apple introduced a special annotation just for it @propertyWrapper, it is also applied to build SwiftUI.

An idea could be make use of `projectedValue`.

> For instance, FirebaseStorage could be implemented as follows:

You should anyway give the possibility to call publishers from any class with this approach. The user could just have a `StorageReference`

> Renaming Publisher to Publishers would make the type act as strict namespace where it is used as a wrapper and namespace.

`Publisher` in my opinion remains misleading, since is not a Publisher. When you call `publisher`, you're expecting a publisher.

> For instance, Apple applies the publisher suffix model to URLSession where only two methods affected dataTaskPublisher(for:) and the other ones that don't have a publisher suffix are not actually publishers.

I think the points is here, why should we do more than Apple itself, which simply mapped just some of the asynchronous functions. Another hint of this is the fact that Combine doesn't have a mapping Publisher for synchronous methods as `Future` is for asynchronous ones.

```swift
Auth.auth().publisher
    .signOut()
    .retry(1)     // <--
    ...
```
Why retry a sync code? Many of Combine methods make sense if in some part of the pipeline you encounter some async code, and in those cases you could simply use .map() to connect the sync code. For the remaining cases you could simply use:
```swift
Just(())
    .tryMap { try Auth.auth().signOut() }
    .retry(1)
    ...
```
The first one is surely cleaner, but I remain doubtful... If we really need a publisher for those, why not just call `.signOutPublisher`?

ziadtamim:
> An idea could be make use of projectedValue

The publisher's wrapper uses the private property for the same purpose as `projectedValue`, to returns the wrapped object.

> You should anyway give the possibility to call publishers from any class with this approach. The user could just have a StorageReference

Could you please elaborate more?

> Publisher in my opinion remains misleading, since is not a Publisher. When you call publisher, you're expecting a publisher.

The current proposed solution is based on separating the concerns, assures the client that all underlying methods are publishers, and making the Firebase calls discoverable and discard non-Firebase publishers from the autocomplete as requested by @peterfriese, 

Screenshot 2021-01-26 at 22 11 28

There is another alternative which is to use the Swift extension feature to introduce Firebase publishers, but this will come to the cost of less discoverable calls.

Screenshot 2021-01-26 at 22 17 59

Another solution which is the one available on the `combine-main` branch, add a `publisher` suffix to ambiguous methods such as `authStateDidChangePublisher`, `idTokenDidChangePublisher`, `useAppLanguagePublisher`, `signOutPublisher` to silence the compiler and **leave other unambiguous publishers without the `publisher` suffix, which makes it inconsistent and confusing to the consumer**. On the other, they are displayed with non-Combine imperative Firebase calls.

Screenshot 2021-01-26 at 22 39 13

> Why retry a sync code? 

`signOut` is throwable.

lorenzofiamingo:
> > You should anyway give the possibility to call publishers from any class with this approach. The user could just have a StorageReference
> 
> Could you please elaborate more?

Example:
``` swift
struct Media {
    let ref: StorageReference = Auth.auth().reference(...)
}
let img = Media()
// Cannot call img.ref publishers, unless you do this:
FirebaseStorage.StorageReference(storageReference: img.ref). downloadURL() 
```
> Publisher in my opinion remains misleading, since is not a Publisher. When you call publisher, you're expecting a publisher.

I'm talking about naming conventions. The struct `Publisher` is not a Publisher, but mostly a _manager_ of publishers. When you type `Auth.auth().publisher`, this is not yet a publisher.
`Timer` has `TimerPublisher`, `URLSession` has `DataTaskPublisher`, etc... This are all publisher. `Auth` should not have a struct `Publisher` that is not a Publisher. For this I'm saying to rename it to `Publishers`, or in an other way.

> ... and makes the Firebase calls discoverable ...

I think is more intuitive to discover call by typing the name of the method you need and then find out the best option to use for your needs. 
Schermata 2021-01-27 alle 01 11 00

>> Why retry a sync code?
>
> signOut is throwable.

The point is that nothing can change between attempts to run a sync code so if it throws an error the first time it will do the same all the others.


In my opinion we should follow what Apple did integrating their frameworks with Combine. Translate sync methods to Combine is really strange, indeed I couldn't find any other Combine implementation doing this. We are stepping out of a standard path the Combine user is used to.

peterfriese:
Thanks all fo the thoughtful discussion.

To summarise, the main reason for introducing a namespace as suggested by @ziadtamim is that some methods (specifically the non-concurrent ones) will have name clashes resulting in the compiler not being able to disambiguate calls. 

We've discussed internally, and feel that wrapping synchronous calls feels out of line with Apple's example. 

For the time being, we would therefor suggest the following:
- Let's finish implementation of the P1 APIs (i.e. Auth, Firestore, Cloud Functions) along the guidelines laid out in [DECISIONS.md](https://github.com/firebase/firebase-ios-sdk/blob/combine-main/FirebaseCombineSwift/DECISIONS.md).
- Let's not wrap synchronous code, but leave it up to the user to decide how to handle making synchronous calls. 
- It will be helpful to provide one or two example apps that showcase some of the functionality. This will also help identify any usability issues with our implementation, and give us a better understanding whether the decision to not wrap synchronous code is valid.

Once we've finished this base (MVP) implementation, we're open to seeing folks put forward alternative approaches. By following this approach, we'll be in a better position to compare competing approaches, as we'll have a more complete API surface to consider.

I've updated the [tracking bug](https://github.com/firebase/firebase-ios-sdk/issues/7295) and included a roadmap.

PR #7333, opened 2021-01-20
Use external repo for GoogleAppMeasurement SPM dependency

paulb777:
Confirmed in Analytics QS and CI.

Will rebase to master after #7315 merges

#no-changelog

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -221,8 +221,8 @@ let package = Package(
       dependencies: [
         .target(name: "FirebaseAnalytics", condition: .when(platforms: [.iOS])),
         .product(name: "GoogleAppMeasurement",
-                package: "GoogleAppMeasurement",
-                condition: .when(platforms: [.iOS])),
+                 package: "GoogleAppMeasurement",
+                 condition: .when(platforms: [.iOS])),
         "FirebaseCore",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -96,6 +96,11 @@ let package = Package(
       url: "https://github.com/apple/swift-protobuf.git",
       from: "1.14.0"
     ),
+    .package(
+      name: "GoogleAppMeasurement",
+      url: "https://github.com/google/GoogleAppMeasurement.git",
+      from: "7.4.0"
```
ryanwilson, line 102: Do we want to do exact versioning here, or are they going to release in a different cadence?

paulb777: There is still a requirement to do exact matching, but the Analytics team will explore loosening in the future. cc: @allenkv @htcgh

ryanwilson: In that case we should use `.package(name: "GoogleAppMeasurement", url: "https://github.com/google/GoogleAppMeasurement.git", .exact("7.4.0")` (if that's the right syntax)

paulb777: Good catch. Thanks! Done.

ryanwilson (approved):

PR #7348, opened 2021-01-21
Device logging w/ GULLogger.

manjanac:


Package.swift
```diff
@@ -597,6 +597,7 @@ let package = Package(
       dependencies: [
         "FirebaseCore",
         "FirebaseInstallations",
+        "GoogleUtilities_Logger",
```
ryanwilson, line 600: This is out of date from master now, so you'll need to update before this can land.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -123,6 +123,13 @@ final class ModelDownloaderUnitTests: XCTestCase {
     }
   }
 
+  /// Test on-device logging.
+  func testDeviceLogging() {
+    DeviceLogger.logEvent(level: .error,
+                          message: "This is a test error logged to device.",
+                          messageCode: .testError)
+  }
```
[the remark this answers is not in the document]
manjanac: No. For now all this test does is log a "error" because nothing in the SDK so far is dire enough to warrant that level. I wasn't sure how to check if something was logged to device. Is there a way?

manjanac:
@ryanwilson Not sure what I'm doing wrong with adding the `GoogleUtilities` dependencies. Any pointers? Thanks!

paulb777:
In Swift PM, the module name is different. See https://github.com/google/GoogleUtilities/blob/main/SwiftPMTests/objc-import-test/objc-module.m#L18.  Perhaps wrap in `#if SWIFT_PACKAGE`?

paulb777:
The test passes with:

```
#if SWIFT_PACKAGE
import GoogleUtilities_Logger
#else
import GoogleUtilities
#endif
```

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -243,7 +242,7 @@ extension ModelDownloadTask {
     static let invalidServerResponse =
       "Could not get server response for model downloading."
     static let unknownDownloadError = "Unable to download model due to unknown error."
-    static let saveModel: StaticString =
+    static let saveModel =
```
paulb777, line 245: nit: one line

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
```diff
@@ -167,9 +161,9 @@ class TelemetryLogger {
 extension TelemetryLogger {
   /// Error descriptions.
   fileprivate enum ErrorDescription {
-    static let encodeEvent: StaticString =
+    static let encodeEvent =
```
paulb777, line 164: one line here and below

paulb777 (approved):

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Done!

FirebaseMLModelDownloader/Sources/TelemetryLogger.swift
manjanac: Done.

PR #7350, opened 2021-01-21
Allow http schemes when using the auth emulator

### LINKED ISSUE #7342: Firebase Auth Emulator Fails Over Local Network

  * Xcode version: 12.0
  * Firebase SDK version:
```
- Firebase/Auth (7.3.0):
    - Firebase/CoreOnly
    - FirebaseAuth (~> 7.3.0)
  - Firebase/CoreOnly (7.3.0):
    - FirebaseCore (= 7.3.0)
  - FirebaseAuth (7.3.0):
    - FirebaseCore (~> 7.0)
    - GoogleUtilities/AppDelegateSwizzler (~> 7.0)
    - GoogleUtilities/Environment (~> 7.0)
    - GTMSessionFetcher/Core (~> 1.4)
```
  * Installation method: `yarn install + react native autolinking (through cocoapods)`
  * Firebase Component: Auth

We are attempting to host our firebase auth emulator on our local network (192.168.1.1xx) so that our physical devices in debug mode (for react native) can connect to it. Normally, on a simulator, this isn't an issue, as we use localhost. We changed our 'host' properties in firebase.json and everything works well in that regard - we are able to hit 192.168.1.1xx:9099 from browsers on other devices, meaning that our auth emulator has bound correctly.

However! When trying to hit this URL, we encountered a bug deep down in this codebase.

When switching from http://localhost:9099 to http://192.168.1.1xx:9099, we started getting the generic failure messages. Please note that our API was perfectly reachable, just not the auth code!

When digging deep into things, we got this error:

`Insecure fetch request has a scheme (http) not found in fetcher allowedInsecureSchemes ( @[] ): http://192.168.1.100:9099/www.googleapis.com/identitytoolkit/v3/relyingparty/signupNewUser?key=(OURKEYHERE)`

It seems evident to me that this is a bug where you guys need to allow insecure schemes (http) for useEmulator, so that we can successfully reach our auth emulator on the local network.

If you need anything else, please ask. We are happy to provide more info.

#### Steps to reproduce:

To reproduce this problem:
1. Bind auth emulator to your local network, and bind functions.
2. Run both emulators.
3. Observe that both the api and auth emulator are reachable from another device on the internal network (say, a phone).
4. Observe that the api works great.
5. Observe that the auth emulator refuses to hit a local IP on the network `192.168.1.1xx`, resulting in failure for auth to be useful.

#### Relevant Code:

It seems likely that the bug lives in [this file.](https://github.com/firebase/firebase-ios-sdk/blob/289e093f89f5dd4ba44377cf57dbf6062e8354c3/FirebaseAuth/Sources/Backend/FIRAuthBackend.m).

Our code was simply calling
```
firebase.auth().useEmulator('http://192.168.1.1xx:9099/');
await auth().createUserWithEmailAndPassword(email, password);
```

It is worth noting we are using rnfirebase to make the calls down to this source.

rosalyntan:
Fix #7342

### DIFF
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Allow http schemes when using the auth emulator. (#7350)
+
 # 7.4.0
 - [fixed] Check if the reverse client ID is configured as a custom URL scheme before setting it as the callback scheme. (#7211)
 - [added] Add ability to sync auth state across devices. (#6924)
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-- [fixed] Allow http schemes when using the auth emulator. (#7350)
+- [fixed] Auth emulator now works across the local network. (#7350)
 
 # 7.4.0
 - [fixed] Check if the reverse client ID is configured as a custom URL scheme before setting it as the callback scheme. (#7211)
--- a/FirebaseAuth/Sources/Backend/FIRAuthBackend.m
+++ b/FirebaseAuth/Sources/Backend/FIRAuthBackend.m
@@ -654,7 +654,7 @@ static id<FIRAuthBackendImplementation> gBackendImplementation;
   NSString *emulatorHostAndPort = requestConfiguration.emulatorHostAndPort;
   if (emulatorHostAndPort) {
     fetcher.allowLocalhostRequest = YES;
-    fetcher.allowedInsecureSchemes = @[@"http"];
+    fetcher.allowedInsecureSchemes = @[ @"http" ];
   }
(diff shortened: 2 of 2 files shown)


FirebaseAuth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Allow http schemes when using the auth emulator. (#7350)
```
ryanwilson, line 2: Nit: This describes what was done but doesn't get into what it enables, maybe something like "Auth emulators now work across the local network"?  I'm fine landing as is but was curious if you like the suggestion.

ryanwilson (approved):
Thanks for the quick fix!

FirebaseAuth/CHANGELOG.md
rosalyntan: Done.

PR #7351, opened 2021-01-21
When building protoc, also check if CMAKE_CROSSCOMPILING is set.

jonsimantov:
CMAKE_CROSSCOMPILING is the standard variable to be set when you are building for a platform that is incompatible with the one you are running on. Checking if this is set allows Firestore to use external protoc in the case of cross-compiling (e.g. arm64 mac builds on an x64 mac) in addition to the current cases of building for mobile.


wilhuff (approved):
LGTM

PR #7352, opened 2021-01-22
Downloader protocol + tests

manjanac:
Downloading a file is now a protocol: `FileDownloader`. There are two classes that conform to it:
1) `ModelFileDownloader` - also conforms to `URLSessionDownloadDelegate` and used by ModelDownloadTask for true downloads
2) `MockModelFileDownloader` - for unit tests.


FirebaseMLModelDownloader/Sources/FileDownloader.swift
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  func urlSession(_ session: URLSession, taskIsWaitingForConnectivity task: URLSessionTask) {
+    // TODO: Handle waiting for connectivity, if needed.
+  }
+
+  func urlSession(_ session: URLSession, didBecomeInvalidWithError error: Error?) {
+    guard let error = error else { return }
+    /// Unable to resolve hostname or connect to host.
+    configurationErrorHandler?(error)
```
maksymmalyhin, line 112: nit: I think this error can be reported to `downloadErrorHandler`

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -94,93 +81,96 @@ class ModelDownloadTask: NSObject {
...
+                              DispatchQueue.main.async {
+                                self.downloadHandlers
+                                  .completion(.failure(.internalError(description: ModelDownloadTask
+                                      .ErrorDescription
+                                      .invalidHostName(error
+                                        .localizedDescription))))
+                              }
+                            }) { response, location in
+      /// Model download completed.
+      guard let response = response as? HTTPURLResponse else {
```
maksymmalyhin, line 115: nit: This validation may be moved to `FileDownloader`

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -39,74 +56,47 @@ enum DownloadModelInfoResult {
...
+      self.session = urlSession
+    } else {
+      self.session = URLSession(configuration: .ephemeral)
     }
-    return request
   }
 
   /// Get installations auth token.
-  private func getAuthToken(completion: @escaping (Result<String, DownloadError>) -> Void) {
+  lazy var authToken = { (completion: @escaping (Result<String, DownloadError>) -> Void) in
```
maksymmalyhin, line 97: As discussed offline, this property probably should be renamed to something like `authTokenProvider` or `authTokenRetriever`. Also we most likely need to remove `installations` property from the class and use the Installations static method in the real `authTokenProvider (we may need to pass FIRApp as a param along with `completion` into the closure to get installation instance for the specific app).

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -148,3 +144,365 @@ final class ModelDownloaderUnitTests: XCTestCase {
...
+extension UserDefaults {
+  /// Returns a new cleared instance of user defaults.
+  static func createTestInstance(testName: String) -> UserDefaults {
+    let suiteName = "com.google.firebase.ml.test.\(testName)"
+    // TODO: reconsider force unwrapping
+    let defaults = UserDefaults(suiteName: suiteName)!
+    defaults.removePersistentDomain(forName: suiteName)
+    return defaults
+  }
+}
```
maksymmalyhin, line 508: I'm going to push a commit with a test example...

maksymmalyhin: Ah, sorry, I missed you have one!

maksymmalyhin: I created #7355 with a couple suggestions.

FirebaseMLModelDownloader/Sources/FileDownloader.swift
manjanac: Ah but what if we want to handle download errors and configuration errors differently in the caller?

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Same as above. What if we'd like to pass a different error to the final completion based on whether it's a config error, or a HTTP response error, or something else? Or do we want the download error to be fairly opaque and logging to be more specific about the error?

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Good catch, I renamed it to `authTokenProvider`. However I would prefer not to pass `FirebaseApp` past `ModelDownloader` if that's possible. Ideally, it would be best if I could pass the entire definition of `authTokenProvider` from `ModelDownloader` but I haven't had a chance to figure out the signature of that parameter. What are your thoughts?

FirebaseMLModelDownloader/Sources/FileDownloader.swift
maksymmalyhin: I would suggest to define an error type to differentiate different types of errors reported by the `FileDownloader`, e.g.

```
enum FileDownloaderError: Error { 
  case sessionInvalidated(Error)
  case unexpectedResponseType
  case networkError(Error)
  ...
}
```

In this case the API will be a bit cleaner but the client has a chance differentiate types of errors if needed.

nit: Additionally, the completion success and error handlers may be merged into one, but it is up to you, e.g.
```
struct FileDownloaderResponse {
    var urlResponse: HTTPURLResponse
    var fileURL: URL
  }

protocol FileDownloader {
  func downloadFile(with url: URL, completion: (Result<FileDownloaderResponse, Error>)->Void)
}

```

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
maksymmalyhin: I think the approach described in the [comment](https://github.com/firebase/firebase-ios-sdk/pull/7352#discussion_r562863883) should be relevant here as well - you may pass a specific error in this case to be handled by ModelDownloadTask as needed.

FirebaseMLModelDownloader/Sources/FileDownloader.swift
manjanac: Great idea! I made this change.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Resolved with the changes in #7355.

FirebaseMLModelDownloader/Sources/FileDownloader.swift
```diff
@@ -0,0 +1,118 @@
...
+      /// Wait for 10 minutes.
+      self.configuration.timeoutIntervalForResource = 600
+    }
+    configuration.allowsCellularAccess = conditions.allowsCellularAccess
+  }
+
+  func downloadFile(with url: URL,
+                    progressHandler: @escaping (Int64, Int64) -> Void,
+                    completion: @escaping (Result<FileDownloaderResponse, Error>) -> Void) {
+    self.completion = completion
```
maksymmalyhin, line 72: Just in case, we probably should check that this method is not called while there is previous download is still going and call completion with an error straight away in this case. It's ok to address it later as we already have this check in the client.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -94,93 +81,96 @@ class ModelDownloadTask: NSObject {
...
-    // TODO: Log this.
-    guard downloadTask == self.downloadTask else { return }
-    guard let response = downloadTask.response as? HTTPURLResponse else {
-      DispatchQueue.main.async {
-        self.downloadHandlers
-          .completion(.failure(.internalError(description: ModelDownloadTask
-              .ErrorDescription.invalidServerResponse)))
-      }
+  /// Fetch model info again and retry download if allowed.
+  func maybeRetryDownload() {
```
maksymmalyhin, line 128: nit: in the future I would suggest to consider moving retry logic up to `ModelDownloader` level. In this case the `ModelDownloadTask` will just complete with an error and `modelInfoRetriever` dependency will not be required here.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -148,3 +144,376 @@ final class ModelDownloaderUnitTests: XCTestCase {
...
+      projectID: "fakeProjectID",
+      apiKey: "fakeAPIKey",
+      authTokenProvider: successAuthTokenProvider,
+      appName: "fakeAppName",
+      localModelInfo: fakeLocalModelInfo,
+      session: fakeSession
+    )
+
+    modelInfoRetriever.downloadModelInfo { result in
+      switch result {
```
maksymmalyhin, line 248: If the completion is not called the test will still succeed. We need to add an expectation like in line 330 check if the completion was called.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -148,3 +144,376 @@ final class ModelDownloaderUnitTests: XCTestCase {
...
+      .completion?(.success(FileDownloaderResponse(urlResponse: fakeResponse,
+                                                   fileURL: tempFileURL)))
+
+    wait(for: [completionExpectation], timeout: 0.5)
+
+    // Cleanup temp file.
+    try? ModelFileManager.removeFile(at: tempFileURL)
+  }
+
+  // TODO: Update similarly to `testModelDownloadWith200`
```
maksymmalyhin, line 376: Would you like to address the todo in the next PR?

FirebaseMLModelDownloader/proto/firebase_ml_log_sdk.pb.swift
```diff
@@ -1,677 +0,0 @@
-// DO NOT EDIT.
```
maksymmalyhin, line 1: Could you please remind why you're deleting the protobuf files?

maksymmalyhin (approved):
LGTM, a couple non-blocking comments (can be addressed in the following PRs).

FirebaseMLModelDownloader/Sources/FileDownloader.swift
manjanac: Gotcha, added a TODO to address this in an upcoming PR.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: I agree. I'll create a PR to do this.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
manjanac: Good catch, done.

FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
manjanac: Ah sorry I missed this. I've added one test that checks for successful retry (i.e., task has expired url at first but retrieves a subsequent valid url). In order to test failed retry, I think it's best to first refactor `maybeRetryDownload()` to be in `ModelDownloader` (upcoming PR).

FirebaseMLModelDownloader/proto/firebase_ml_log_sdk.pb.swift
manjanac: I already have the same files under `Sources/proto` so these are duplicate. Apologies for the confusion!

Issue #6756, opened 2020-10-16
Remove "whitelist" in FDL source


paulb777:
Consider "allowlist" instead

```
$ git grep whitelist
FirebaseDynamicLinks/Sources/FIRDynamicLinks.m:// Custom domains to be whitelisted are optionally added as an array to the info.plist.
FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.h: Add custom domains from the info.plist to the internal whitelist.
```

paulb777:
Internally tracked at b/178199888

eldhosembabu:
Done. Closing the issue.

PR #7358, opened 2021-01-22
Rewording "whitelist" in comments and function names.

eldhosembabu:
Addressing https://github.com/firebase/firebase-ios-sdk/issues/6756


paulb777 (approved):
Thanks!

PR #7372, opened 2021-01-25
Remove watchOS for Crashlytics Pod Gen

samedson:
When I pod gen, it gives me this error. Updating our main script to pod gen to remove watchOS.

```
Generating FirebaseCrashlytics in `gen/FirebaseCrashlytics`
    Creating stub application
    Writing Podfile

Installing...
[!] The platform of the target `App-watchOS` (watchOS 6.0) is not compatible with `FirebaseCrashlytics/unit (7.4.0)`, which does not support `watchOS`.
```


paulb777 (approved):
In general, I recommend using `pod gen` for more than one platform at a time since Xcode can get confused about which libraries go with which platform.

However, if 3 platforms are working for you - that bug may not affect Crashlytics.

samedson:
Sounds good - it seems to be working for us for now but will keep that in mind

PR #7384, opened 2021-01-26
Remove GDT from repo

paulb777:
GDT is moving to https://github.com/google/GoogleDataTransport

Three files are kept for now because of dependencies to the FirebasePerformance unit tests:

- ./GoogleDataTransport/GDTCORTests/Common/Fakes/GDTCORTransportFake.h
- ./GoogleDataTransport/GDTCORTests/Common/Fakes/GDTCORTransportFake.m
- ./GoogleDataTransport/GDTCORLibrary/Private/GDTCORTransport_Private.h

cc: @visumickey 

#no-changelog


maksymmalyhin (approved):
LGTM.

I guess, if `./GoogleDataTransport/GDTCORLibrary/Private/GDTCORTransport_Private.h` is actually used then Perf tests should fail. If Perf tests pass then, I think, it should be fine to use `#import <GoogleDataTransport/GoogleDataTransport.h>` instead of private headers here: https://github.com/firebase/firebase-ios-sdk/blob/3720a72bac7096989f336228c50f8d3e7a407412/FirebasePerformance/Tests/Unit/Loggers/FPRGDTCCLoggerTest.m#L30

paulb777:
`GDTCORTransport_Private.h` is used. Without it:

Screen Shot 2021-01-27 at 7 37 54 AM

paulb777:
Agreed, but will defer the resolution of those three files to a subsequent PR.

PR #7386, opened 2021-01-27
Logging on-device + Firelog

manjanac:
So far I've mostly tackled on-device logging with a couple of Firelog events. 

Still a work in progress so I will keep this PR in draft until all Firelog events are logged, but appreciate feedback in the meantime as well.


FirebaseMLModelDownloader/Sources/ModelFileManager.swift
```diff
@@ -15,7 +15,7 @@
 import Foundation
 
 /// Manager for common file operations.
-enum ModelFileManager {
+class ModelFileManager: FileManager {
```
ryanwilson, line 18: Did you mean to include this here as well?

ryanwilson (commented):
Just one question otherwise LGTM

FirebaseMLModelDownloader/Sources/ModelFileManager.swift
manjanac: Ah good catch! That was by accident, but I was wondering if I should mock the FileManager dependency (so that I can pass a different directory and still test that file operations work). Do you think that may be worth it? If not, I'll change this back to enum.

ryanwilson (approved):
LGTM! Please wait for green to merge.

PR #7397, opened 2021-01-28
More on-device logging and better error handling.

manjanac:
Continuing with more on-device logging and better error handling. Still a WIP since I need to log more Firelog events, but feedback about the rest of it is welcome!


FirebaseMLModelDownloader/Sources/ModelFileManager.swift
```diff
@@ -75,7 +103,12 @@ class ModelFileManager: FileManager {
       }
     }
     do {
-      try FileManager.default.moveItem(at: sourceURL, to: destinationURL)
+      guard let result = try? hasSufficientSpace(size: size), result == true else {
```
maksymmalyhin, line 106: I wonder if additional disk space is required to move a file? Can it depend on OS version? As for me, I would rather let the file manager to fail and try to translate its error to an error we are interested in. WDYT?

maksymmalyhin (approved):
👍 LGTM! One question.

FirebaseMLModelDownloader/Sources/ModelFileManager.swift
manjanac: I liked this idea too, but I couldn't figure out how to translate whatever error FileManager throws to be about available space (vs something else). Since that's probably the only I/O error that the developer can actually do something about. How could we handle this better? Also @ryanwilson

maksymmalyhin: Looks like [this](https://developer.apple.com/documentation/swift/cocoa_design_patterns/handling_cocoa_errors_in_swift) article uses specifically FileManager case as an example. I guess `CocoaError.fileWriteOutOfSpace` is the error you are interested in, though worth double checking.

manjanac: That is an excellent find, thank you! I'll use `CocoaError.fileWriteOutOfSpace` to isolate the space availability error.

PR #7403, opened 2021-01-29
Fix startAfter/endBefore for orderByKeys queries

IanWyszynski:
We should not be passing a fallback key to startAfter if we're querying on a key index. The indexValue is the key in this case.


FirebaseDatabase/Tests/Integration/FIRDatabaseQueryTests.m
```diff
@@ -1390,6 +1390,50 @@ - (void)testStartAtPriorityAndEndAtPriorityWork {
...
 
+- (void)testStartAfterWithOrderByKey {
+  FIRDatabaseReference* ref = [FTestHelpers getRandomNode];
+  FIRDatabaseReference* childOne = [ref childByAutoId];
+  FIRDatabaseReference* childTwo = [ref childByAutoId];
+
+  [childOne setValue:@1];
+  [childTwo setValue:@2];
+
+  __block bool done = NO;
```
schmidt-sebastian, line 1401: s/bool/BOOL

FirebaseDatabase/Tests/Integration/FIRDatabaseQueryTests.m
```diff
@@ -1390,6 +1390,50 @@ - (void)testStartAtPriorityAndEndAtPriorityWork {
...
+
+- (void)testEndBeforeWithOrderByKey {
+  FIRDatabaseReference* ref = [FTestHelpers getRandomNode];
+  FIRDatabaseReference* childOne = [ref childByAutoId];
+  FIRDatabaseReference* childTwo = [ref childByAutoId];
+
+  [childOne setValue:@1];
+  [childTwo setValue:@2];
+
+  __block bool done = NO;
```
schmidt-sebastian, line 1423: s/bool/BOOL

FirebaseDatabase/Sources/Api/FIRDatabaseQuery.m
```diff
@@ -202,20 +202,26 @@ - (FIRDatabaseQuery *)queryStartingAfterValue:(id)startAfterValue {
...
+                initWithName:INVALID_QUERY_PARAM_ERROR
+                      reason:
+                          @"You must use queryStartingAfterValue: instead of "
+                          @"queryStartingAfterValue:childKey: when using "
+                          @"queryOrderedByKey:"
+                    userInfo:nil];
+        }
+        if ([startAfterValue isKindOfClass:[NSString class]]) {
+            startAfterValue = [FNextPushId successor:startAfterValue];
+        }
```
schmidt-sebastian, line 217: What happens when this is not true? Looks like we just call `queryStartingAtInternal` with the original value, which doesn't seem to do any validation.

IanWyszynski: This case is covered by `validateQueryEndpointsForParams`:

https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDatabase/Sources/Api/FIRDatabaseQuery.m#L106-L107

This is called in `queryingStartingAtInternal`.

FirebaseDatabase/Sources/Api/FIRDatabaseQuery.m
```diff
@@ -95,32 +95,36 @@ - (void)validateQueryEndpointsForParams:(FQueryParams *)params {
...
             }
             if (![params.indexStartValue.val isKindOfClass:[NSString class]]) {
                 [NSException
                      raise:INVALID_QUERY_PARAM_ERROR
-                    format:
-                        @"Can't use queryStartingAtValue: with other types "
-                        @"than string in combination with queryOrderedByKey"];
+                    format:@"Can't use queryStartingAtValue: or "
+                           @"queryStartingAfterValue: "
+                           @"with other types than string in combination with "
```
schmidt-sebastian, line 108: Might be easier to understand if it said "with non-string types when used with queryOrderedByKey:"

schmidt-sebastian (approved):
Thanks for updating the error message.

Can you add a Changelog entry before submitting?

FirebaseDatabase/Sources/Core/FRepo.m
```diff
@@ -518,10 +518,8 @@ - (void)getData:(FIRDatabaseQuery *)query
         (void (^_Nonnull)(NSError *__nullable error,
                           FIRDataSnapshot *__nullable snapshot))block {
     FQuerySpec *querySpec = [query querySpec];
-    id<FNode> node =
-        [self.serverSyncTree calcCompleteEventCacheAtPath:querySpec.path
-                                          excludeWriteIds:@[]];
-    if (![node isEmpty]) {
+    id<FNode> node = [self.serverSyncTree getServerValue:[query querySpec]];
+    if (node != nil && ![node isEmpty]) {
```
IanWyszynski, line 522: @schmidt-sebastian I'm stuck here. I can't figure out why `[node isEmpty]` is needed to pass the two tests:

https://github.com/firebase/firebase-ios-sdk/pull/7403/files#diff-3a8cb26bda8f6a6d6700b7c2ee8b1368a1c147ed7bb29743a0f6146c08d7b723R1437-R1495

The equivalent is not needed for the android changes.

schmidt-sebastian (commented):
I still have do debug the [node isEmpty] issue, but it looks like the SDK doesn't read from cache when it should.

PR #7412, opened 2021-01-29
Fix a crash that happens when Firestore is being terminated when the last shared pointer to it is in a listener

### LINKED ISSUE #6909: Crash: NSInternalInconsistencyException', reason: 'FIRESTORE INTERNAL ASSERTION FAILED: Enqueue methods cannot be called when we are already running on target executor

  * Xcode version: 12.1
  * Firebase SDK version: 7.0.0
  * Firebase Component: Auth, Core

#### Steps to reproduce:

Observing a crash when trying to delete a `FirebaseApp` object.

calling 
```
app.delete {
}
```

where app is a FirebaseApp

Seeing this
```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'FIRESTORE INTERNAL ASSERTION FAILED: Enqueue methods cannot be called when we are already running on target executor (this queue's executor: 'com.google.firebase.firestore.httpsinstacartfirebaseiocomv1', current executor: 'com.google.firebase.firestore.httpsinstacartfirebaseiocomv1') (expected !is_operation_in_progress_ || !executor_->IsCurrentExecutor())'
```

var-const:
If a user calls terminate and immediately disposes their reference to Firestore, it's possible that while termination is in process, the last remaining reference (more precisely, shared pointer) to Firestore is in a listener. When that listener is destroyed as part of the termination, it leads to `api::Firestore` being destroyed. Importantly, termination happens on the worker queue. The destructor of `api::Firestore` calls `Dispose` which presumes it's not called from the worker queue and tries to enqueue work, leading to a failing assertion and a crash.

The general problem is that `FirestoreClient::Dispose` may be invoked on the worker queue from `TerminateInternal` and needs a way to handle it gracefully. This PR largely makes `Dispose` a no-op in that case.

Fixes #6909.

### DIFF
--- a/Firestore/core/src/core/firestore_client.cc
+++ b/Firestore/core/src/core/firestore_client.cc
@@ -64,6 +64,7 @@ namespace core {
 using api::DocumentReference;
 using api::DocumentSnapshot;
 using api::DocumentSnapshotListener;
+using api::Firestore;
 using api::ListenerRegistration;
 using api::QuerySnapshot;
 using api::QuerySnapshotListener;
--- a/Firestore/core/src/core/firestore_client.cc
+++ b/Firestore/core/src/core/firestore_client.cc
@@ -113,13 +113,11 @@ std::shared_ptr<FirestoreClient> FirestoreClient::Create(
     std::shared_ptr<CredentialsProvider> credentials_provider,
     std::shared_ptr<Executor> user_executor,
     std::shared_ptr<AsyncQueue> worker_queue,
-    std::unique_ptr<FirebaseMetadataProvider> firebase_metadata_provider,
-    std::shared_ptr<Firestore> firestore) {
+    std::unique_ptr<FirebaseMetadataProvider> firebase_metadata_provider) {
   // Have to use `new` because `make_shared` cannot access private constructor.
   std::shared_ptr<FirestoreClient> shared_client(new FirestoreClient(
       database_info, std::move(credentials_provider), std::move(user_executor),
-      std::move(worker_queue), std::move(firebase_metadata_provider),
-      std::move(firestore)));
+      std::move(worker_queue), std::move(firebase_metadata_provider)));
 
   std::weak_ptr<FirestoreClient> weak_client(shared_client);
   auto credential_change_listener = [weak_client, settings](User user) mutable {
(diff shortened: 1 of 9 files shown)


Firestore/core/src/core/firestore_client.cc
```diff
@@ -263,7 +275,13 @@ void FirestoreClient::Dispose() {
 
 void FirestoreClient::TerminateAsync(StatusCallback callback) {
   worker_queue_->EnterRestrictedMode();
-  worker_queue_->EnqueueEvenWhileRestricted([this, callback] {
+  worker_queue_->EnqueueEvenWhileRestricted([&, this, callback] {
+    // Make sure this `FirestoreClient` is not destroyed during
+    // `TerminateInternal`, so that `user_executor_` stays valid.
+    auto self = shared_from_this();
```
var-const, line 281: This is arguably unnecessary -- `api::Firestore`, locked below, would contain a pointer to the client. I'm not sure there's a case where `weak_firestore_` would return null here (perhaps in unit testing), but it seems like a safer, if perhaps redundant, approach.

var-const:
@wilhuff Any suggestions on different approaches/potential pitfalls/etc. are very welcome.

var-const:
@wilhuff Looks like extending the lifetime of `api::Firestore` isn't really necessary -- `Dispose` works fine when invoked from `TerminateInternal` as long as I use `is_terminated` to check whether enqueuing `TerminateInternal` is necessary. This approach seems simpler -- sorry if you have already started with the review.

var-const:
Hmm, I'm getting an Asan error in CI. I'll investigate and let you know when this PR is ready for another look.

var-const:
Ok, it's ready for review again. The simpler approach doesn't work, and the previous approach needed a fix.

Firestore/core/src/util/async_queue.cc
```diff
@@ -86,6 +86,7 @@ void AsyncQueue::ExecuteBlocking(const Operation& operation) {
               "ExecuteBlocking may not be called "
               "before the previous operation finishes executing");
 
+  auto self = shared_from_this();
```
var-const, line 89: Is there a reason we haven't done this before? `is_operation_in_progress_ = false;` may crash when `api::Firestore` (and, consequently, the worker queue) is destroyed on the queue (in `FirestoreClient::TerminateAsync`).

wilhuff: This seems like an oversight.

I think the idea at the time was that FirestoreClient would keep the AsyncQueue alive as long as it needed to, but this doesn't account for the case where the Firestore destructor is running on the AsyncQueue.

var-const:
Closing in favor of https://github.com/firebase/firebase-ios-sdk/pull/7421.

PR #7419, opened 2021-02-01
Fixes

manjanac:
Fixes:

1. Update model path check to use `standardizedFileURL`.
2. Add a `downloadStatus` property to prevent multiple concurrent downloads.


FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -43,7 +42,7 @@ class ModelDownloadTask: NSObject {
   /// User defaults to which local model info should ultimately be written.
   private let defaults: UserDefaults
   /// Keeps track of download associated with this model download task.
-  private(set) var downloadStatus: ModelDownloadStatus
+  private(set) var downloadStatus: ModelDownloadStatus = .complete
```
maksymmalyhin, line 45: Why `.complete` is the default state?

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
```diff
@@ -59,19 +58,19 @@ class ModelDownloadTask: NSObject {
...
   /// Name for model file stored on device.
   var downloadedModelFileName: String {
-    return "fbml_model__\(appName)__\(remoteModelInfo.name)"
+    return "fbml_model__\(appName)__\(remoteModelInfo.name).tflite"
   }
 
   func download(progressHandler: ProgressHandler?, completion: @escaping Completion) {
     /// Prevent multiple concurrent downloads.
-    guard downloadStatus == .notStarted else {
+    // TODO: Merge if there are multiple same requests.
```
maksymmalyhin, line 72: I think merging should happen on the class user level.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Ah I reduced the possible statuses to be either `complete` or `inProgress`. Earlier, I had a `notStarted` as well but it seemed a bit redundant. I can bring it back for clarity though.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Do you mean at the `ModelDownloader` level?

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
maksymmalyhin: It looks a bit confusing when initial status is `complete` to me. But I'm mostly fine with either way (not a blocker for the PR for me).

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
maksymmalyhin: yes

maksymmalyhin (approved):
LGTM

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Ok makes sense, I brought back `.notStarted` to clear this up.

FirebaseMLModelDownloader/Sources/ModelDownloadTask.swift
manjanac: Moved the TODO, thanks!

PR #7421, opened 2021-02-02
Fix a crash that happens when Firestore is being terminated when the last shared pointer to it is in a listener

### LINKED ISSUE #6909: Crash: NSInternalInconsistencyException', reason: 'FIRESTORE INTERNAL ASSERTION FAILED: Enqueue methods cannot be called when we are already running on target executor

var-const:
If a user calls terminate and immediately disposes their reference to Firestore, it's possible that while termination is in process, the last remaining reference (more precisely, shared pointer) to Firestore is in a listener. When that listener is destroyed as part of the termination, it leads to `api::Firestore` being destroyed. Importantly, termination happens on the worker queue. The destructor of `api::Firestore` calls `Dispose` which presumes it's not called from the worker queue and tries to enqueue work, leading to a failing assertion and a crash.

The solution is simply to remove the sequential order checks when the queue enters/is in restricted mode. There is a legitimate case where the Firestore destructor can run on the worker queue, as the original issue shows. The complexity of #7412 also indicates that a simpler solution is preferable.

Fixes #6909.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+- [fixed] Fixed a crash that would happen when the app is being deleted and
+  immediately disposed of and there's an active listener (#6909).
+
 # v7.5.0
 - [changed] A write to a document that contains FieldValue transforms is no
   longer split up into two separate operations. This reduces the number of
--- a/Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
+++ b/Firestore/Example/Tests/Integration/API/FIRDatabaseTests.mm
@@ -1450,10 +1450,6 @@ using firebase::firestore::util::TimerId;
   // Make sure there is a listener.
   [doc addSnapshotListener:^(FIRDocumentSnapshot *, NSError *){
   }];
-  NSDictionary<NSString *, id> *data =
-      @{@"owner" : @{@"name" : @"Jonny", @"email" : @"abc@xyz.com"}};
-  // Make sure the client is initialized.
-  [self writeDocumentRef:doc data:data];
 
   XCTestExpectation *expectation = [self expectationWithDescription:@"App is deleted"];
   [app deleteApp:^(BOOL) {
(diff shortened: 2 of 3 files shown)


var-const:
@wilhuff I have reworked the code based on your feedback, and it seemed easier to start from scratch.

Firestore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+- [fixed] Fixed a crash that would happen when the app is being deleted and
```
wilhuff, line 2: Is there a way to simplify this phrasing? Both "the app is being deleted" and "immediately disposed of" sound duplicative. Also, should we capitalize "app"?

wilhuff (approved):
LGTM

wilhuff:
The new test is failing:

> /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/build/../Firestore/Example/Tests/Util/XCTestCase+Await.mm:40: error: -[FIRDatabaseTests testDeleteAppWorksWhenLastReferenceToFirestoreIsInListener] : Asynchronous wait failed: Exceeded timeout of 25 seconds, with unfulfilled expectations: "setData".

var-const:
> The new test is failing

Not sure why -- it was timing out on writing the doc to the backend, even though it's pretty much copied without (and succeeds when I run it locally). Anyway, it turns out that writing a doc isn't actually necessary to trigger the bug -- the fact that the client gets initialized when creating a document reference is enough -- so I removed the writing part, which fixed the CI issue. I verified that the shorter version of the test is enough to trigger the crash (without the fix from this PR).

Firestore/CHANGELOG.md
var-const: Sorry, I got caught up in CI failures and somehow missed this comment. Addressed in #7449.

PR #7423, opened 2021-02-02
Fix import for watchOS

peterfriese:
This needs to be TARGET_OS_WATCH (also see the corresponding .m file, where it actually already is TARGET_OS_WATCH)

#no-changelog

### DIFF
--- a/FirebaseAuth/Sources/Utilities/FIRAuthDefaultUIDelegate.h
+++ b/FirebaseAuth/Sources/Utilities/FIRAuthDefaultUIDelegate.h
@@ -15,7 +15,7 @@
  */
 
 #import <TargetConditionals.h>
-#if !TARGET_OS_OSX && !TARGET_OS_WATCHOS
+#if !TARGET_OS_OSX && !TARGET_OS_WATCH
 
 #import <Foundation/Foundation.h>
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):

derkaramma:
I'm having trouble with this error. My app is not and will not be available for Apple Watch. Would anyone know how I go about quitting the error till the update is pushed?

paulb777:
You could either edit the file manually in your installation or keep using Xcode 12.4 or earlier

pistoleta:
> file

This means edit the file on every build? isn't there a file on unity  that could be modified to avoid this ?
Thanks

pistoleta:
> You could either edit the file manually in your installation or keep using Xcode 12.4 or earlier

This means edit the file on every build? isn't there a file on unity that could be modified to avoid this ?
Thanks

paulb777:
Edit the file after any `pod install` or `pod update`.  You may want ask the Unity question in a different forum - there may not be too much Unity expertise here.

pistoleta:
> Edit the file after any `pod install` or `pod update`. You may want ask the Unity question in a different forum - there may not be too much Unity expertise here.

ooh right, I thought this was from the firebase for unity sdk , thanks

PR #7430, opened 2021-02-02
Update coverage workflow

granluo:
Trigger the code coverage workflow when a PR is created. Fix a bug and remove redundant comments.

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -2,9 +2,10 @@ name: test_coverage
 
 on:
   pull_request:
+    # open will be triggered when a pull request is created.
     # synchronize will be triggered when a pull request has new commits.
     # closed will be triggered when a pull request is closed.
-    types: [synchronize, closed]
+    types: [open, synchronize, closed]
 
 jobs:
   check:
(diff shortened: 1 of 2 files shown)


scripts/code_coverage_report/generate_code_coverage_report/Sources/CoverageReportGenerator/CoverageReportParser.swift
```diff
@@ -66,7 +66,7 @@ extension CoverageReportRequestData {
       // FirebaseDatabase-ios-FirebaseCore.framework, a regex pattern will be
       // used to exclude results that are not related in terms of the target names.
       let sdk_name = resultBundle.components(separatedBy: "-")[0]
-      let range = NSRange(location: 0, length: sdk_name.utf16.count)
+      let range = NSRange(location: 0, length: target.name.utf16.count)
```
granluo, line 69: This bug is detected [here](https://github.com/firebase/firebase-ios-sdk/pull/7428/checks?check_run_id=1817150397). This is caused by the range is not aligned with the `target.name` in line 75.

.github/workflows/test_coverage.yml
```diff
@@ -2,9 +2,10 @@ name: test_coverage
 
 on:
   pull_request:
+    # open will be triggered when a pull request is created.
     # synchronize will be triggered when a pull request has new commits.
     # closed will be triggered when a pull request is closed.
-    types: [synchronize, closed]
+    types: [open, synchronize, closed]
```
granluo, line 8: Synchronize will only trigger a workflow when an existing PR is updated.

paulb777 (approved):

PR #7437, opened 2021-02-03
Testing list models

manjanac:
Two outstanding questions here:

1. Is `ModelFileManager.isFileReachable()` doing the right thing? Should I enforce the synchronization, as per [this](https://developer.apple.com/documentation/foundation/url/1779667-checkresourceisreachable)?
2. Should I be using `URL.path` or `URL.absoluteString` when converting file `URL` to `String`? The fact that "Application Support" has a space in it makes this tricky.

I created a draft PR #7439 that doesn't save model path at all, and simply reconstructs it as needed. Maybe that's the right way to go about it?


FirebaseMLModelDownloader/Tests/Unit/ModelDownloaderUnitTests.swift
```diff
@@ -915,6 +995,13 @@ extension UserDefaults {
     defaults.removePersistentDomain(forName: suiteName)
     return defaults
   }
+
+  /// Returns the existing user defaults instance.
+  static func getTestInstance(testName: String) -> UserDefaults {
+    let suiteName = "com.google.firebase.ml.test.\(testName)"
+    // TODO: reconsider force unwrapping
```
maksymmalyhin, line 1002: nit: Force unwrapping is fine for tests.

maksymmalyhin (approved):
LGTM.

manjanac:
Closing this in favor of #7439.

PR #7445, opened 2021-02-03
Better VoiceOver experience for in-app messages

christibbs:
- Announces the in-app message title over VoiceOver at display time
- Improves accessibility labels
- Announces close button as the last accessible VoiceOver element (so it's easier to find)


paulb777 (approved):
Nice! Is it feasible to add unit tests?

paulb777 (changes requested):
Looks like there are test failures on travis with Xcode 11. PTAL

christibbs:
> Nice! Is it feasible to add unit tests?

No, fortunately you get some accessibility test coverage for free since the UI testing framework uses accessibility identifiers. But there's no way to test the posting of a VoiceOver notification at a particular time.

paulb777 (approved):

PR #7446, opened 2021-02-03
Combine Support: getCredentialWith(_:) in OAuthProvider

ziadtamim:
Proposal for Combine support to `getCredentialWith(_:)` in `OAuthProvider`.

```swift 
  provider.getCredentialWith(uiDelegate)
     .flatMap { credential in 
        Auth().signIn(with: credential) 
     }
     .sink { completion in
        switch completion {
        case .finished:
          // Handle completion
        case let .failure(error):
          // Handle error.
        }
     } receiveValue: { authResult in
         // User is signed in.
     }
     .store(in: &cancellables)
```


FirebaseCombineSwift/Tests/Unit/Bundle+Tests.swift
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+extension Bundle {
+  class func mock(with cls: Bundle.Type) {
```
peterfriese, line 18: Can you elaborate a bit on the functionality of this utility?

OAuthProvider+Combine.swift
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+#if canImport(Combine) && swift(>=5.0) && canImport(FirebaseAuth)
+
+  import Combine
+  import FirebaseAuth
+
+  @available(swift 5.0)
+  @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+  extension OAuthProvider {
```
peterfriese, line 22: This file need to be placed in `FirebaseCombineSwift/Sources/Auth` (it's currently in the root and won't be found by the compiler which results in a broken compile).

peterfriese (changes requested):
Thanks for this PR, Ziad!

I wasn't able to compile initially, as AuthProvider.Combine was in the root. (Do you use `pod gen FirebaseCombineSwift.podspec --local-sources=./ --auto-open --platforms=ios` to generate your development environment?)

Other than that, looks pretty good to me.

OAuthProvider+Combine.swift
ziadtamim: Done

FirebaseCombineSwift/Tests/Unit/Bundle+Tests.swift
ziadtamim: `NSBundle` is used under the hood on `OAuthProvider` to access values such as bundleIdentifier and `object(forInfoDictionaryKey:)` which cannot be injected using internal APIs. Instead, I introduced a utility method to mock the `Bundle` with a subclass to full fill the tests expectations.

peterfriese (approved):
LGTM, thanks for your work!

PR #7450, opened 2021-02-03
GHA zip workflow: add manual trigger

maksymmalyhin:
An attempt to add a [manual trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) for zip GHA workflow. 

I was not able to test the trigger itself as the "Run workflow" button didn't appear for me. I guess the change has to land to the default branch to be picked up by GHA.

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -34,7 +34,7 @@ jobs:
 
   package:
     # Don't run on private repo.
-    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
+    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk' || github.event_name == 'pull_request'
     needs: build
     runs-on: macOS-latest
     steps:
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -20,9 +20,6 @@ on:
         required: true
         default: 'https://cpdc-internal.git.corp.google.com/firebase.git'
 
-env:
-  CUSTOM_SPEC_REPOS: ${{ github.event.inputs.output_dir || https://github.com/firebase/SpecsStaging.git }}
-
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


maksymmalyhin:
The updated package step triggered the correct command for non-manual run: https://github.com/firebase/firebase-ios-sdk/runs/1826829858?check_suite_focus=true#step:4:12
Further packaging was cancelled.

.github/workflows/zip.yml
```diff
@@ -13,6 +13,13 @@ on:
     # Run every day at midnight(PST) - cron uses UTC times
     - cron:  '0 8 * * *'
 
+  workflow_dispatch:
+    inputs:
+      custom_spec_repos:
+        description: 'Custom Podspec repos'
+        required: true
+        default: 'https://cpdc-internal.git.corp.google.com/firebase.git'
```
paulb777, line 21: I don't think GHA has access to `https://cpdc-internal.git.corp.google.com`. I believe we need the whole staging process to SpecsStaging or another public repo.

.github/workflows/zip.yml
```diff
@@ -37,7 +44,7 @@ jobs:
     - name: ZipBuildingTest
       run: |
          mkdir -p zip_output_dir
```
paulb777, line 46: Add an `xcode-select` command to Xcode 12.0

.github/workflows/zip.yml
maksymmalyhin: Yeah, I would prefer moving release staging to GitHub to simplify GHA access control. I think we can address it in following PRs. For now I would like to test the manual trigger in general.

paulb777 (approved):

.github/workflows/zip.yml
maksymmalyhin: I would prefer to add it in a separate PR. With the manual trigger testing will be so much easier :).

PR #7454, opened 2021-02-04
BundleSerializer complete.

wu-hui:


Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -98,31 +458,461 @@ BundleMetadata BundleSerializer::DecodeBundleMetadata(
...
+
+BundledDocumentMetadata BundleSerializer::DecodeDocumentMetadata(
+    util::ReadContext& context,
+    const std::string& document_metadata_string) const {
+  const json& document_metadata = Parse(document_metadata_string);
+
+  if (document_metadata.is_discarded()) {
+    context.Fail("Failed to parse string into json: " +
+                 document_metadata_string);
+    return BundledDocumentMetadata{};
```
wilhuff, line 823: This can probably just be `return {};` without needing to spell out the type.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -98,31 +458,461 @@ BundleMetadata BundleSerializer::DecodeBundleMetadata(
...
+  }
+
+  if (!document_metadata.contains("name") ||
```
wilhuff, line 826: You could avoid these separate checks (with generic error messages) if you accessed required field with a helper like `RequireChild` or maybe even just `Require`. Something like this:

```c++
const json& Reader::RequireChild(const char* child_name, const json& parent) {
  if (!parent.contains(child_name) {
    Fail("Missing child '%s'", child_name);
    return parent;
  }
  return parent.at(child_name);
}
```

Now you can do:

```c++
SnapshotVersion read_time =
    DecodeSnapshotVersion(context, context.Require("readTime", document_metadata));
```

It's not too much extra text at the call site, but it completely eliminates the separate `contains` check.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -98,31 +458,461 @@ BundleMetadata BundleSerializer::DecodeBundleMetadata(
...
+      !document_metadata.contains("readTime")) {
+    context.Fail(
+        "One of bundledDocumentMetadata's 'name' or 'readTime' is missing");
+    return BundledDocumentMetadata{};
+  }
+
+  ResourcePath path = DecodeName(context, document_metadata.at("name"));
+  if (!context.ok()) {
```
wilhuff, line 834: This is probably one of the rare cases where you do actually need an explicit check because `DocumentKey` asserts if its input is broken.

Firestore/core/src/bundle/bundled_document_metadata.h
```diff
@@ -0,0 +1,87 @@
...
+  const model::SnapshotVersion& read_time() const {
+    return read_time_;
+  }
+
+  /** Returns whether the document exists. */
+  bool exists() const {
+    return exists_;
+  }
+
+  /** Returns the names of the queries in this bundle that this document matches
```
wilhuff, line 59: Formatting is broken here.

Firestore/core/src/nanopb/reader.h
```diff
@@ -45,6 +45,9 @@ namespace nanopb {
  */
 class Reader {
  public:
+  Reader() = default;
+  explicit Reader(util::ReadContext context) : context_(std::move(context)) {
```
wilhuff, line 49: I'm surprised this is necessary/works. 

The way you're calling this makes a copy of the ReadContext the caller supplies. This suggests that the local `ReadContext` they're keeping is irrelevant because it's disconnected from what's actually being passed down.

Are you sure this actually works?

If you're able to change the Serializer to just use the `ReadContext*` instead for the methods you need, then you can just remove these additions and not solve this mystery :-).

wilhuff (commented):
Preliminary review. I haven't spent any time on tests yet.

wilhuff:
Note that there are several comments that apply throughout the serializer implementation. For example, I focused on `ReadRequiredString` without noting every circumstance where that could be applied. You might consider similar treatment for other types--I haven't suggested those at all.

In general, I think you can avoid this form with helpers:

```c++
if (!parent.contains("child") || !parent.at("child").is_type()) {
  // fail
}
parent.at("child").get<type>();  // or get_ref
```

wu-hui (commented):
Thanks for the review. Applied `Require*` methods and removed a bunch of validation that is not neccessary.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -24,16 +42,50 @@ namespace firestore {
...
+
+  void Read(const pb_field_t fields[], void* dest_struct) override {
+    (void)fields;
+    (void)dest_struct;
+    UNREACHABLE();
+  }
+};
 
 template <typename int_type>
 int_type ToInt(ReadContext& context, const json& value) {
```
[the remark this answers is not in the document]
wu-hui: Above is obsolete, they are moved into `Reader` instead.

Firestore/core/src/remote/serializer.h
```diff
@@ -125,12 +126,8 @@ class Serializer {
   /**
    * @brief Converts from nanopb proto to the model FieldValue format.
    */
-  // TODO(rsgowman): Once the proto is read, the only thing the reader object is
-  // used for is error handling. This seems questionable. We probably need to
-  // rework error handling. Again. But we'll defer that for now and continue
-  // just passing the reader object.
   model::FieldValue DecodeFieldValue(
-      nanopb::Reader* reader, const google_firestore_v1_Value& proto) const;
+      util::ReadContext* context, const google_firestore_v1_Value& proto) const;
```
wilhuff, line 130: Optional: ReadContext can't ever be null so this could be a mutable reference rather than a pointer (this interface was created before that change to the style guide had gone through). Feel free to leave for a follow-up if you don't want to tackle that now.

You could make this less of an intrusive fix by making `Reader::context()` return a mutable reference too.

Firestore/core/src/timestamp_internal.cc
```diff
@@ -16,12 +16,47 @@
...
+  return FromUntrustedSecondsAndNanos(seconds, nanos);
+}
+
+util::StatusOr<Timestamp> TimestampInternal::FromUntrustedSecondsAndNanos(
+    int64_t seconds, int32_t nanos) {
+  // The Timestamp ctor will assert if we provide values outside the valid
+  // range. However, since we're decoding, a single corrupt byte could cause
+  // this to occur, so we'll verify the ranges before passing them in since we'd
+  // rather not abort in these situations.
+  if (seconds < TimestampInternal::Min().seconds()) {
```
wilhuff, line 43: I suspect you don't have to qualify with `TimestampInternal::` here.

Firestore/core/src/timestamp_internal.cc
```diff
@@ -16,12 +16,47 @@
...
+    return util::Status(
+        firestore::Error::kErrorDataLoss,
```
wilhuff, line 45: Data loss seems like the wrong error by default. It seems like this would be an invalid argument here and then possibly translated to data loss in contexts where we're parsing existing data.

Firestore/core/src/util/read_context.h
```diff
@@ -57,6 +57,11 @@ class ReadContext {
     status_.Update(util::Status(Error::kErrorDataLoss, std::move(description)));
   }
 
+  template <typename... FA>
+  void Fail(const char* format, const FA&... args) {
+    Fail(StringFormat(format, args...));
```
wilhuff, line 62: Please include `string_format.h`.

Firestore/core/src/bundle/bundle_serializer.h
```diff
@@ -19,25 +19,81 @@
...
 namespace firestore {
 namespace bundle {
 
+class JsonReader : public util::ReadContext {
+ public:
+  const std::string& RequireString(const char* name,
+                                   const nlohmann::json& json_object);
+  const std::vector<nlohmann::json>& RequireArray(
+      const char* name, const nlohmann::json& json_object);
+  const nlohmann::json& Require(const char* child_name,
```
wilhuff, line 45: Please add a comment for this method. While the name is nice and short, it's not obvious exactly what is required.

Alternatively, you could want to make this `RequireObject` to parallel `RequireArray`, above. This would make it more obvious.

Firestore/core/src/bundle/bundle_serializer.h
```diff
@@ -19,25 +19,81 @@
...
 #include "Firestore/core/src/model/snapshot_version.h"
+#include "Firestore/core/src/remote/serializer.h"
 #include "Firestore/core/src/util/read_context.h"
 #include "nlohmann/json.hpp"
 
 namespace firebase {
 namespace firestore {
 namespace bundle {
 
+class JsonReader : public util::ReadContext {
```
wilhuff, line 39: Consider adding a class comment to this type that indicates generally how these functions work. Ideally these would all have comments, but they're all mostly the same except for types, so that's probably overkill. A single class comment that describes how `Require` methods will `Fail` if the key is not present or has the wrong type and then return a default value would be enough.

Firestore/core/src/bundle/bundle_serializer.h
```diff
@@ -19,25 +19,81 @@
...
+  const std::vector<nlohmann::json>& RequireArray(
+      const char* name, const nlohmann::json& json_object);
+  const nlohmann::json& Require(const char* child_name,
+                                const nlohmann::json& json_object);
+
+  double RequireDouble(const char* name, const nlohmann::json& value);
+  template <typename int_type>
+  int_type RequireInt(const char* name, const nlohmann::json& value);
+
+  static bool OptionalBool(const char* name, const nlohmann::json& json_object);
```
wilhuff, line 52: You may want to have this take a default value as the last argument (`bool default_value = false`). This would only make sense if you chose to add `OptionalString` and other such methods.

Firestore/core/src/bundle/named_query.h
```diff
@@ -22,7 +22,6 @@
 
 #include "Firestore/core/src/bundle/bundled_query.h"
 #include "Firestore/core/src/model/snapshot_version.h"
-
```
wilhuff, line 25: Please revert. There should be a newline between includes and namespace decls.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -26,10 +38,165 @@ namespace bundle {
...
+        local_serializer(remote_serializer),
+        bundle_serializer(remote_serializer) {
+    msg_diff_.ReportDifferencesToString(&message_differences);
+  }
+
+  remote::Serializer remote_serializer;
+  local::LocalSerializer local_serializer;
+  bundle::BundleSerializer bundle_serializer;
+
+  static std::string GetDocumentFullPath(const std::string& path) {
```
wilhuff, line 77: In the interest of reducing clutter in tests, you could call this just `FullPath`.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -26,10 +38,165 @@ namespace bundle {
...
+    std::string json_string;
+    MessageToJsonString(document, &json_string);
+    VerifyJsonStringDecodeFails(std::move(json_string));
+  }
+
+  BundleDocument VerifyJsonStringDecodes(std::string json_string) {
+    JsonReader reader;
+    BundleDocument actual =
+        bundle_serializer.DecodeDocument(reader, json_string);
+    EXPECT_TRUE(reader.ok());
```
wilhuff, line 133: If this fails you'll get a terrible error message. Instead use `EXPECT_OK` or `EXPECT_NOT_OK` on the underlying status, which will show the error messages if applicable.

Applies throughout.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -26,10 +38,165 @@ namespace bundle {
...
+  // 4. Parse the json back to `NamedQuery` object, then compare.
+  void VerifyNamedQueryRoundtrip(const Query& original) const {
+    // `First` and `None` will be encoded as BundledQuery.limit_type = `First`,
+    // as not all SDKs have a `None`.
+    auto limit_type = original.limit_type() == core::LimitType::Last
+                          ? core::LimitType::Last
+                          : core::LimitType::First;
+    BundledQuery bundled_query(original.ToTarget(), limit_type);
+    NamedQuery named_query("query-1", bundled_query,
+                           model::SnapshotVersion(Timestamp::Now()));
```
wilhuff, line 157: `testutil::Version` is more concise than creating an actual snapshot version. It also results in a specific version, which will make diffs between test runs less random.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -72,51 +245,825 @@ TEST(BundleSerializerTest, DecodesBundleMetadata) {
...
-  EXPECT_FALSE(context.ok());
+  EXPECT_FALSE(reader.ok());
 
   // Replace total_bytes to a string unparseable to integer.
   std::string json_copy =
       ReplacedCopy(json_string, "123456789987654321", "xxxyyyzzz");
-  context = ReadContext{};
-  EXPECT_TRUE(context.ok());
-  serializer.DecodeBundleMetadata(context, json_copy);
+  reader = JsonReader{};
```
wilhuff, line 263: This reassignment between cases is something of a smell. There's a bunch of essentially unrelated cases together in this test that are sharing state. You can avoid unintentionally bleeding through from one case to the other by creating separate blocks for each one. e.g.

```c++
{
  JSONReader reader;
  bundle_serializer.DecodeBundleMetadata(reader, invalid);
  EXPECT_NOT_OK(reader.status());
}

{
  std::string json_copy =
      ReplacedCopy(json_string, "123456789987654321", "xxxyyyzzz");
  JsonReader reader;
  bundle_serializer.DecodeBundleMetadata(reader, json_copy);
  EXPECT_NOT_OK(reader.status());
}
```

This keeps any local changes and other objects scoped to each sub-case.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -72,51 +245,825 @@ TEST(BundleSerializerTest, DecodesBundleMetadata) {
...
+  EXPECT_TRUE(reader.ok());
```
wilhuff, line 264: Consider breaking this base case assertion out into its own sub-case. At first I thought this was related to parsing the json_copy and was an error.

wilhuff (commented):
Feel free to push back on some of the optional suggestions in here. This is getting pretty close!

Firestore/core/src/remote/serializer.h
wu-hui: Added a TODO.

Firestore/core/src/timestamp_internal.cc
wu-hui: Done.

Firestore/core/src/timestamp_internal.cc
wu-hui: Done.

Firestore/core/src/util/read_context.h
wu-hui: Done.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Done.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Done.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Done.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Done.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Don't think it's still needed with separate blocks.

wilhuff (commented):
Code pretty much LGTM. There are still some unaddressed comments. I've resolved everything else to make it clearer. Ignore any comments without a "Resolve" button--somehow GitHub won't resolve comments within a thread even if the thread itself is resolved.

You still need CocoaPods and SPM support for this build. You could conceivably do that outside this PR.

Firestore/core/src/bundle/bundled_document_metadata.h
wu-hui: Should be fixed now.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -72,6 +146,292 @@ json Parse(const std::string& s) {
   return json::parse(s, /*callback=*/nullptr, /*allow_exception=*/false);
 }
 
+Timestamp DecodeTimestamp(ReadContext& context, const json& version) {
+  if (version.is_string()) {
+    Time time;
+    std::string err;
+    bool ok = ParseTime(RFC3339_full, version.get_ref<const std::string&>(),
```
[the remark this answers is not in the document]
wu-hui: Fixed.

wilhuff (commented):
LGTM so far. Are you intending to the CocoaPods/SPM stuff in a separate PR?

wu-hui:
> LGTM so far. Are you intending to the CocoaPods/SPM stuff in a separate PR?

Yes. They will be in separate PRs.

wilhuff (approved):
LGTM

PR #7455, opened 2021-02-04
Dequeue limbo resolutions when their respective queries are stopped

dconeybe:
Fix a bug where enqueued limbo resolutions are left in the queue even after all targets that care about their resolutions are stopped. This erroneous behavior was reported in https://github.com/firebase/firebase-android-sdk/issues/2311 against the Android SDK, but the bug is also present in the Web and iOS SDKs. This PR is a port of the equivalent fix in the Web SDK: https://github.com/firebase/firebase-js-sdk/pull/4395. This bug was introduced when limbo resolution throttling was implemented almost a year ago (https://github.com/firebase/firebase-js-sdk/pull/2790).


Firestore/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 - [fixed] Fixed a crash that could happen when the App is being deleted and
   there's an active listener (#6909).
+- [fixed] Fixed a bug where local cache inconsistencies were unnecessarily
+  being resolved (#7455).
```
schmidt-sebastian, line 5: You should explain what the user visible impact is (as you did in Android).

Firestore/core/src/core/sync_engine.h
```diff
@@ -220,6 +219,53 @@ class SyncEngine : public remote::RemoteStoreCallback, public QueryEventSource {
...
+      void cancel() {
+        cancelled_ = true;
+      }
+
+     private:
+      model::DocumentKey key_;
+      bool cancelled_ = false;
+    };
+
+    void PruneLeadingCancelledQueueEntries();
```
schmidt-sebastian, line 262: Can you add a method comment to explain what this does?

Firestore/core/src/core/sync_engine.h
```diff
@@ -220,6 +219,53 @@ class SyncEngine : public remote::RemoteStoreCallback, public QueryEventSource {
     bool document_received = false;
   };
 
+  class LimboResolutionQueue {
```
schmidt-sebastian, line 222: Should this:
- be an outer class? (@var-const)
- have it's own unit tests?

Firestore/core/src/core/sync_engine.h
```diff
@@ -220,6 +219,53 @@ class SyncEngine : public remote::RemoteStoreCallback, public QueryEventSource {
...
+
+    bool empty() const {
+      return queue_.empty();
+    }
+
+    bool contains(const model::DocumentKey& key) const;
+    std::vector<model::DocumentKey> keys() const;
+
+   private:
+    class QueueEntry {
```
schmidt-sebastian, line 240: Since this wrapper only seems to exist to support cancellation, I was wondering if you are up for a challenge. You could advance to the stage of a "true hacker" (tm) if you use `unordered_map` to indicate whether an entry is still active. You can remove inactive entries from the unordered_map and clean up the vector by treating the map as a source of truth.

Firestore/core/src/util/random_access_queue.h
```diff
@@ -0,0 +1,265 @@
...
+
+  /**
+   * Returns the element at the front of the queue.
+   *
+   * The behavior of this method is undefined if the queue is empty.
+   *
+   * This method has constant-time complexity.
+   */
+  const T& front() const {
+    return queue_.front().element();
```
schmidt-sebastian, line 107: Should we assert here that `removed` is set to false?

Firestore/core/test/unit/util/random_access_queue_test.cc
```diff
@@ -0,0 +1,647 @@
+/*
+ * Copyright 2017 Google LLC
```
schmidt-sebastian, line 2: - 2017: Trump but no Covid
- 2021: No Trump but Covid

Not sure which one is better, but I would still lean towards 2021.

Firestore/core/test/unit/util/random_access_queue_test.cc
```diff
@@ -0,0 +1,647 @@
...
+
+void PrintTo(const RandomAccessQueue<int>& queue, std::ostream* out) {
+  *out << ToDebugString(queue);
+}
+
+MATCHER_P(QueueElementsEqual,
+          expected_elements,
+          std::string("contains elements: ") +
+              ToDebugString(expected_elements)) {
+  return expected_elements == arg.elements();
```
schmidt-sebastian, line 63: Can you use a built-in here? https://chromium.googlesource.com/external/gmock/+/master/test/gmock-generated-matchers_test.cc#387

Firestore/core/test/unit/util/random_access_queue_test.cc
```diff
@@ -0,0 +1,647 @@
...
+  queue.push_back(1);
+  queue.push_back(2);
+  queue.remove(2);
+  queue.remove(1);
+  EXPECT_TRUE(queue.empty());
+  queue.push_back(1);
+  EXPECT_FALSE(queue.empty());
+}
+
+TEST(RandomAccessQueueTest, ContainsShouldReturnTrueOnANewlyCreatedQueue) {
```
schmidt-sebastian, line 248:
```suggestion
TEST(RandomAccessQueueTest, ContainsShouldReturnFalseOnANewlyCreatedQueue) {
```

Firestore/core/test/unit/util/random_access_queue_test.cc
```diff
@@ -0,0 +1,647 @@
...
+  EXPECT_THAT(queue, QueueElementsEqual(std::vector<int>({1, 0, 2})));
+  EXPECT_FALSE(queue.push_back(1));
+  EXPECT_THAT(queue, QueueElementsEqual(std::vector<int>({1, 0, 2})));
+  EXPECT_FALSE(queue.push_back(2));
+  EXPECT_THAT(queue, QueueElementsEqual(std::vector<int>({1, 0, 2})));
+}
+
+}  //  namespace util
+}  //  namespace firestore
+}  //  namespace firebase
```
schmidt-sebastian, line 647: I am thoroughly impressed.

schmidt-sebastian (approved):
This looks good to me. If you don't trust mine and your C++ skills then you should send this to Costa (who can probably skip the tests)

Firestore/core/src/util/random_access_queue.h
dconeybe: Done.

Firestore/core/test/unit/util/random_access_queue_test.cc
dconeybe: Tough call. But I tend to agree. Changed.

Firestore/core/test/unit/util/random_access_queue_test.cc
dconeybe: Done. Great suggestion.

Firestore/core/test/unit/util/random_access_queue_test.cc
dconeybe: Done.

Firestore/core/test/unit/util/random_access_queue_test.cc
dconeybe: Thanks!

PR #7457, opened 2021-02-04
CPDC push: register repo in not exists and fail if one of pushes fails

maksymmalyhin:


ReleaseTooling/Sources/FirebaseReleaser/Push.swift
```diff
@@ -80,7 +88,29 @@ enum Push {
...
+  }
+
+  /// @param defaultRepoName The repo name to register if not exists
+  private static func findOrRegisterPrivateCocoaPodsRepo(repo: String, gitRoot: URL,
+                                                         defaultRepoName: String) -> String {
+    if let repoName = findPrivateCocoaPodsRepo(repo: repo, gitRoot: gitRoot) {
+      return repoName
+    }
+
+    let command = "pod repo add \(defaultRepoName) \(repo)"
```
maksymmalyhin, line 103: This command is executed even with `--log-only true`. I assume it is fine but let me know if you think otherwise.

paulb777 (approved):

PR #7458, opened 2021-02-04
Restore master manifest version

paulb777:
Restore manifest version until 7.6.0 release publishes.

The version update needs to be in the release branch only until we make the release process fully public to keep the nightly zip.yml running.


maksymmalyhin:
I see, thank you for the fix.

maksymmalyhin (approved):

PR #7460, opened 2021-02-04
RTDB FSRWebSocket: suppress diagnostic error

maksymmalyhin:
Fix for cl/355486726 - g3 build error:
```
error: variable length array used [-Werror,-Wvla]
            uint8_t buffer[bufferSize];
                           ^~~~~~~~~~
```

### DIFF
--- a/FirebaseDatabase/Sources/third_party/SocketRocket/FSRWebSocket.m
+++ b/FirebaseDatabase/Sources/third_party/SocketRocket/FSRWebSocket.m
@@ -1559,8 +1559,12 @@ static const size_t SRFrameHeaderOverhead = 32;
 
         case NSStreamEventHasBytesAvailable: {
             SRFastLog(@"NSStreamEventHasBytesAvailable %@", aStream);
+
+            #pragma clang diagnostic push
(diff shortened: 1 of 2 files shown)


FirebaseDatabase/Sources/third_party/SocketRocket/FSRWebSocket.m
```diff
@@ -1559,8 +1559,12 @@ - (void)safeHandleEvent:(NSStreamEvent)eventCode stream:(NSStream *)aStream
 
         case NSStreamEventHasBytesAvailable: {
             SRFastLog(@"NSStreamEventHasBytesAvailable %@", aStream);
+
+            #pragma clang diagnostic push
+            #pragma clang diagnostic ignored "-Wvla"
             const NSUInteger bufferSize = 2048;
```
paulb777, line 1565: This looks related to building for Objective-C++ instead of Objective C. Do you understand why that's happening?

Either way, it might be cleaner to address with a #define BUFFER_SIZE instead since all usages are in this file.

paulb777 (approved):

schmidt-sebastian (approved):

Issue #6521, opened 2020-09-22
Firebase ID token has expired.


IgorRosocha:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 11.7
  * Firebase SDK version: 6.31.0
  * Firebase Component: Auth
  * Component version: 6.31.0
  * Installation method: Carthage

#### Steps to reproduce:

In our app, we use Firebase authentication token to identify user on our backend. We are sending this token with every request. We retrieve Firebase token using [getIDToken(completion:)](https://firebase.google.com/docs/reference/swift/firebaseauth/api/reference/Classes/User#getidtokencompletion:) with every request.

The documentation says the method `Retrieves the Firebase authentication token, possibly refreshing it if it has expired.`, but from time to time, we receive error `Firebase ID token has expired. Get a fresh ID token from your client app and try again (auth/id-token-expired). See https://firebase.google.com/docs/auth/admin/verify-id-tokens for details on how to retrieve an ID token.` This is kind of a major issue for our app, as we automatically log out users which receive unauthenticated error (because of user deletion functionality) - not mentioning that all of the requests automatically fail because of expired token.

Am I misunderstanding the concept of retrieving the token using [getIDToken(completion:)](https://firebase.google.com/docs/reference/swift/firebaseauth/api/reference/Classes/User#getidtokencompletion:), or is this the bug in SDK?

samhorlbeck:
Hi, thanks for reaching out--

Where are you seeing the `auth/id-token-expired` error? In your backend code or on the client?

How much time elapses between calling `getIdToken` and your backend doing the verification? Is it possible the token is becoming expired in flight (meaning: order of minutes between getting it from the client and doing verification on the backend)

IgorRosocha:
> Where are you seeing the auth/id-token-expired error? In your backend code or on the client?

In the response from backend.

> How much time elapses between calling getIdToken and your backend doing the verification? Is it possible the token is becoming expired in flight (meaning: order of minutes between getting it from the client and doing verification on the backend)

I believe not. This happens after user launches the app, and sends initial requests with token to BE. Token is retrieved on the client using `getIDToken(completion:)` with every request.

samhorlbeck:
Hmm yeah this is interesting. The relevant bit of code is here: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m#L110

[...]

PR #7461, opened 2021-02-05
Add debug logging around token expiration

rosalyntan:
Add some debug logging to help figure out why an expired token is sometimes being returned in #6521.

### DIFF
--- a/FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
+++ b/FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
@@ -19,10 +19,13 @@
 #import "FirebaseAuth/Sources/Public/FirebaseAuth/FIRAuth.h"
 
 #import "FirebaseAuth/Sources/Auth/FIRAuthSerialTaskQueue.h"
+#import "FirebaseAuth/Sources/Auth/FIRAuth_Internal.h"
 #import "FirebaseAuth/Sources/Backend/FIRAuthBackend.h"
 #import "FirebaseAuth/Sources/Backend/FIRAuthRequestConfiguration.h"
 #import "FirebaseAuth/Sources/Backend/RPC/FIRSecureTokenRequest.h"
 #import "FirebaseAuth/Sources/Backend/RPC/FIRSecureTokenResponse.h"
+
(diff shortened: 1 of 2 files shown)


FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
```diff
@@ -111,6 +111,7 @@ - (void)fetchAccessTokenForcingRefresh:(BOOL)forceRefresh
       complete();
       callback(self->_accessToken, nil, NO);
     } else {
+      NSLog(@"Fetching new token from backend.");
```
maksymmalyhin, line 114: NSLog will log messages unconditionally which may be undesirable for the app developers not interested in the messages. We have a [logger interface](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Sources/Private/FIRLogger.h) with log levels which helps us to avoid too noisy logs produced by Firebase SDKs. Messages logged with `FIRLogDebug` will be visible only if debug logging is explicitly enabled by the developer. See usage example [here](https://github.com/firebase/firebase-ios-sdk/blob/8d9243454b0d141f520ba7495836d4be71b44a3d/FirebaseInstallations/Source/Library/InstallationsIDController/FIRInstallationsIDController.m#L159).

rosalyntan: Done.

FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
```diff
@@ -200,7 +204,19 @@ - (void)requestAccessToken:(FIRFetchAccessTokenCallback)callback {
 }
```
maksymmalyhin, line 204: Would it make sense to add another log at around line 190 to track when the token ivars are updated?

maksymmalyhin (approved):
LGTM

FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
rosalyntan: Done.

PR #7462, opened 2021-02-05
Mostly minor fixes that came out of the bugbash

manjanac:


FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -281,45 +281,107 @@ class ModelInfoRetriever {
...
+                                                               status: .modelInfoRetrievalSucceeded,
+                                                               errorCode: .noError)
               completion(.success(.notModified))
             case 400:
-              let description = ModelInfoRetriever.ErrorDescription.invalidModelName(self.modelName)
-              DeviceLogger.logEvent(level: .debug,
-                                    message: description,
-                                    messageCode: .invalidModelName)
+              if let data = data,
+                let responseJSON = try? JSONSerialization
```
maksymmalyhin, line 290: nit: looks like the response parsing and logging is repeated several times, so worth refactoring to a separate method/helper/closure to reduce duplication.

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
```diff
@@ -281,45 +281,107 @@ class ModelInfoRetriever {
...
+                .jsonObject(with: data, options: []) as? [String: Any],
+                let error = responseJSON["error"] as? [String: Any],
+                let errorMessage = error["message"] as? String {
+                DeviceLogger.logEvent(level: .debug,
+                                      message: errorMessage,
```
maksymmalyhin, line 295: Another thought - would it be easier and more universal to print the response body as a string without any  content parsing.

maksymmalyhin (approved):
LGTM. Thank you for adding tests!

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Hm that would be ideal. But I tried using `String(data:, encoding:)` and could an ugly string with a whole bunch of `\n` characters. Is there a better way?

maksymmalyhin: Ah, I see. Maybe it's because Swift specifics. This approach worked for be pretty well from Objective-C.

I guess you may print entire `responseJSON` object as a compromise.

ryanwilson: You could always use that string and then use `.replacingOccurrences(of: "\n", with: " ")` and see if that works :)

manjanac: Gotcha. I ended up using option 1 above - method to reduce duplication, because I realized we don't want the entire error message to be surfaced, just the "message" in the error message. :)

FirebaseMLModelDownloader/Sources/ModelInfoRetriever.swift
manjanac: Done! I have made this change but make it part of a future PR and not push here so as not to disturb merging the dependent PR #7464.

PR #7463, opened 2021-02-05
Update changed file tool

granluo:


granluo:
The original update_file script is shell, which is not easy to maintain. This update is to decouple the script to detect updated files and file patterns. File patterns are now in the JSON file.  SDK name in the JSON file should be the same as the run_job variable name in the test_coverage workflow. E.g. sdk name `database` in the[ json file](https://github.com/firebase/firebase-ios-sdk/pull/7463/files#diff-9c1ae229ad74af4f23ec5a48b047e66e74497f77af2924c0eabe614306b42138) should be the same as the [steps.check_files.outputs.**database**_run_job](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/test_coverage.yml#L15) in the workflow file.

paulb777 (approved):
Nice usability improvement. Thanks!

PR #7470, opened 2021-02-05
Add multiple workflows for coverage tests.

granluo:


paulb777:
Why is only Storage shown in the report?

Why are there two reports for Storage?

The first report seems to be measuring coverage of the unit test sources instead of the library source coverage.

granluo:
> Why is only Storage shown in the report?
> 
> Why are there two reports for Storage?
> 
> The first report seems to be measuring coverage of the unit test sources instead of the library source coverage.

That is because the json file containing path under the script directory, [code_coverage_file_list.json](https://github.com/firebase/firebase-ios-sdk/pull/7470/files?file-filters%5B%5D=.json#diff-9c1ae229ad74af4f23ec5a48b047e66e74497f77af2924c0eabe614306b42138R127).
I saw the [storage.yml](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/storage.yml#L9) workflow had this path.

Will remove the path in the json file.

paulb777:
Makes sense. I think that answers the first question. What about the other two?

granluo:
> Makes sense. I think that answers the first question. What about the other two?

`FirebaseStorage-Unit-unit.xctest` is the coverage for unit test.  Should this one be removed?

paulb777:
I don't think we need the coverage info for the test sources - only for the library sources.

paulb777 (approved):
I'm not sure it's quite right yet, but fine to merge and do additional fixes as we evaluate the reports.

granluo:
Will test the tool in the pr before merge.

PR #7472, opened 2021-02-06
fix for _userAccessGroup being nil when getStoredUserForAccessGroup is called

joehinkle11:
This is a small change which fixes a significant bug. Basically the app doesn't let users share access groups until second launch

To reproduce:
 1. build 2 apps with a shared access group (useUserAccessGroup)
 2. sign in the first app
 3. open the second app (for the first time! if you already opened it, delete and reinstall)
 4. notice the second app is unable to load the access group
 5. force close the second app
 6. reopen the second app
 7. notice the second app is now able to load the access group

The cause: `_userAccessGroup` is set AFTER `getStoredUserForAccessGroup` is called, and `getStoredUserForAccessGroup` depends on `_userAccessGroup` not being null

Solution: swap the order of when `_userAccessGroup` is set and when `getStoredUserForAccessGroup` is called


paulb777:
@joehinkle11 Thanks for the fix and PR. 

@rosalyntan Please take a look.

Note the CI failure can be disregarded. (#7476)

rosalyntan (commented):
Thanks for finding and fixing this issue! Your fix should work, but I took a look into the code, and I think you'll actually want to make a change in https://github.com/firebase/firebase-ios-sdk/blob/025eecd1fcad2eaf84e73deb923a7f54004981cc/FirebaseAuth/Sources/Auth/FIRAuth.m#L2261 instead -- I believe we should be using the passed-in `accessGroup` rather than `self.userAccessGroup` there.

joehinkle11:
@rosalyntan updated!

rosalyntan (approved):
Thanks! LGTM

mikehardy:
Irony: react-native-firebase was inadvertently relying on the old behavior in our E2E test suite, and now that this is working our test suite broke when I attempted to adopt v7.8.0. To be clear: as far as I can tell our test suite was not using the API correctly, but the way unintended consequences ripple out always fascinates me. Cheers all

joehinkle11:
@mikehardy Glad I was a service in breaking something 😅

peterfriese:
> @mikehardy Glad I was a service in breaking something 😅

Breaking Stuff as a Service ;-)

PR #7480, opened 2021-02-08
Enabling watchOS unit test

chliangGoogle:
Enable watchOS on unit test
- Use Swift Package Manager to test the case
- Currently not using pod gen to test: The unit test requires an app host so set requires_app_host = true. However, we ran into the same issue that pod generate a app host with main.m file.  Reopen this issue: https://github.com/CocoaPods/CocoaPods/issues/8283

While enabling the test, also discovered and fixed two issues
1.  We are checking the wrong folder (Document folder) for storage size limit. 
2.  The duplicate message check logic should be enabled for watchOS.


FirebaseMessaging.podspec
```diff
@@ -57,9 +57,15 @@ device, and it is completely free.
 
   s.test_spec 'unit' do |unit_tests|
     unit_tests.scheme = { :code_coverage => true }
-    unit_tests.platforms = {:ios => '10.0', :osx => '10.12', :tvos => '10.0'}
-    unit_tests.source_files = 'FirebaseMessaging/Tests/UnitTests*/*.{m,h,swift}'
-    unit_tests.requires_app_host = true
+    unit_tests.platforms = {:ios => '10.0', :osx => '10.12', :tvos => '10.0', :watchos => '7.4'}
```
paulb777, line 60: I don't think that any other lines need to be changed than this one, since they replicate for all the platforms.

paulb777 (commented):
Do you mean a different Issue number? The one linked doesn't seem related?

FirebaseMessaging.podspec
chliangGoogle: You are right, I was playing around the podspec and forgot to cleanup.
I also updated the correct issue number.

paulb777:
Thanks. I reopened the CocoaPods bug. It might be a relatively complicated fix for CocoaPods to add support for watchOS test_specs.

We might want to consider going straight to Swift PM for watchos unit tests.  It currently gets some build errors:

Screen Shot 2021-02-08 at 3 37 36 PM

paulb777:
Looks like the yml files are updated in the wrong branch ...

paulb777:
Can a matrix with watchos be added at https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/messaging.yml#L56?

paulb777 (approved):

PR #7481, opened 2021-02-08
Enable watchOS for Remote Config and ABTesting

chliangGoogle:
ABTesting is a hard dependency so enable it as well.
Testing: I'm able to get config value without any changes using the watchOS sample app.

### DIFF
--- a/Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
+++ b/Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
@@ -20,43 +20,43 @@ import FirebaseRemoteConfig
 
 /// Entry point of the watch app.
 class ExtensionDelegate: NSObject, WKExtensionDelegate, MessagingDelegate {
-  /// Initialize Firebase service here.
-  func applicationDidFinishLaunching() {
-    FirebaseApp.configure()
-    let center = UNUserNotificationCenter.current()
-    center.requestAuthorization(options: [.alert, .sound]) { granted, error in
-      if granted {
-        WKExtension.shared().registerForRemoteNotifications()
-      }
+    /// Initialize Firebase service here.
+    func applicationDidFinishLaunching() {
+        FirebaseApp.configure()
+        let center = UNUserNotificationCenter.current()
+        center.requestAuthorization(options: [.alert, .sound]) { granted, _ in
+            if granted {
+                WKExtension.shared().registerForRemoteNotifications()
+            }
+        }
+        Messaging.messaging().delegate = self
+        let remoteConfig = RemoteConfig.remoteConfig()
+        remoteConfig.fetchAndActivate { _, error in
+            guard error == nil else {
+                print("error:" + error.debugDescription)
+                return
+            }
+            let configValue: String = remoteConfig["test"].stringValue ?? "You have not set up a 'test' key in Remote Config console."
+            print("value:\n" + configValue)
+        }
     }
-    Messaging.messaging().delegate = self
-    let remoteConfig = RemoteConfig.remoteConfig()
-    remoteConfig.fetchAndActivate { status, error in
-      guard error == nil else {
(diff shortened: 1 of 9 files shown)


paulb777 (commented):
Thanks!

Please update CI with a build only `pod lib lint` at https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/remoteconfig.yml#L52

README.md
```diff
@@ -261,7 +261,7 @@ pod 'Firebase/Firestore'     # No watchOS support yet
 pod 'Firebase/Functions'     # No watchOS support yet
 pod 'Firebase/Messaging'
 pod 'Firebase/Performance'   # No macOS, tvOS, watchOS, and Catalyst support yet
-pod 'Firebase/RemoteConfig'  # No watchOS support yet
```
paulb777, line 264: Also ABTesting at line 256 and its CI

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
```diff
@@ -29,17 +30,25 @@ class ExtensionDelegate: NSObject, WKExtensionDelegate, MessagingDelegate {
       }
     }
     Messaging.messaging().delegate = self
+    let remoteConfig = RemoteConfig.remoteConfig()
+    remoteConfig.fetchAndActivate { status, error in
+      guard error == nil else {
+        print("error:" + error.debugDescription)
+        return
+      }
+      print("value:\n" + remoteConfig["test"].stringValue!)
```
ryanwilson, line 39: Nit: for those that aren't using a project with a `test` key, can you please change the force unwrap to a safe fallback? Something like `?? "No value for key 'test'"` or just `?? "nil"`

ryanwilson (commented):
Awesome! LGTM, please wait for @karenyz approval.

FirebaseABTesting.podspec
```diff
@@ -24,6 +24,7 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
   s.ios.deployment_target = '10.0'
```
paulb777, line 24: nit: using Ruby variables for the platform target versions is more robust. See https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore.podspec#L21

paulb777 (approved):

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
karenyz, line 39: might be helpful if we also print the fetch source?

karenyz (approved):

karenyz:
Thanks for testing this! 🎉

Example/watchOSSample/SampleWatchAppWatchKitExtension/ExtensionDelegate.swift
chliangGoogle: Done.

chliangGoogle: Done. I specify it in the output that if the config is setup on console.

README.md
chliangGoogle: Done.

FirebaseABTesting.podspec
chliangGoogle: Done. Also updated remote config.

chliangGoogle:
my clang-format somehow is not in sync even tho it's the same version.

paulb777:
The log might be confusing, since it's actually swiftformat that checks the format of Swift files - not clang-format

chliangGoogle:
Right, I also upgraded swiftformat and I ran both swiftformat and ./script/style.sh. None of them is accepted by CI.

paulb777:
swiftformat should be controlled by the Mintfile and running 0.45.5 both locally and in CI

PR #7482, opened 2021-02-09
Add internal documentation to the FirebaseFirestore podspec

wilhuff:


wu-hui (approved):
Thanks!

I think I should also add my json lib here as well?

wilhuff:
> I think I should also add my json lib here as well?

Yes, before you merge you'll need to do that. If you're not seeing CI failures related to Xcode builds on that PR now you'll see them when you try to merge to master.

PR #7483, opened 2021-02-09
Improve internal header documentation

paulb777:
As we made header searching more consistent between CocoaPods and SPM, private headers imported from other Firebase libraries are now specified with the CocoaPods `source_files` attribute.

Clarify that C++ internal headers should still be specified with `preserve_paths` to avoid namespace collisions in the generated client Xcode workspace.

Note that this has the downside of losing navigation to the header by filename in Xcode for which the workaround is using the file type.

### DIFF
--- a/HeadersImports.md
+++ b/HeadersImports.md
@@ -34,6 +34,10 @@ In Swift Package Manager, it's a library target.
   should be located among the source files. [Xcode](https://stackoverflow.com/a/8016333) refers to
   these as "Project Headers".
 
+* *Library C++ Internal Headers* - In CocoaPods, C++ internal headers should not be included
+  in the `source_files` attribute. Instead, they should be defined with the `preserve_paths`
+  attribute to avoid namespace collisions in the generated Xcode workspace.
+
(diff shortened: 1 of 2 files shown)


HeadersImports.md
```diff
@@ -34,6 +34,10 @@ In Swift Package Manager, it's a library target.
   should be located among the source files. [Xcode](https://stackoverflow.com/a/8016333) refers to
   these as "Project Headers".
 
+* *Library C++ Internal Headers* - In CocoaPods, C++ internal headers should not be included
```
wilhuff, line 37: While this specifies the consequence (that C++ headers should be included), it doesn't give a rationale for why that's the case. The issue is that C++ does not assume a global header map, so if filenames are qualified at all, it's generally by directory, not a filename prefix like in Objective-C. Is it worth adding something here to make this clear?

paulb777: Done

paulb777 (commented):
Comments addressed

wilhuff (approved):
LGTM

PR #7484, opened 2021-02-09
Disable code coverage report in PRs for external users

granluo:


granluo:
Check if the secret is nil before running steps.

granluo:
Workflow seems to not have `if` clause for jobs. There is a [discussion](https://github.community/t/how-to-use-env-context/16975/5?u=granluo) about the misleading of the doc. For now, the conditions are added to steps.

paulb777 (approved):

PR #7489, opened 2021-02-09
Fix Downloader nightly

paulb777:
- Use secret
- Static frameworks instead of modular headers
- Move next to similar regular pod lib lint tests
- Enable issue reporting for failures

### DIFF
--- a/.github/workflows/generate_issues.yml
+++ b/.github/workflows/generate_issues.yml
@@ -16,4 +16,4 @@ jobs:
       with:
         access-token: '${{ secrets.GITHUB_TOKEN }}'
         # This is to exclude workflows that will be searched in the nightly report.
-        exclude-workflow-files: 'performance-integration-tests.ymls'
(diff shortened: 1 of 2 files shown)


.github/workflows/generate_issues.yml
```diff
@@ -16,4 +16,4 @@ jobs:
       with:
         access-token: '${{ secrets.GITHUB_TOKEN }}'
         # This is to exclude workflows that will be searched in the nightly report.
-        exclude-workflow-files: 'performance-integration-tests.yml,mlmodeldownloader.yml'
+        exclude-workflow-files: 'performance-integration-tests.ymls'
```
ryanwilson, line 19: Nit: is the `s` at the end of `ymls` intentional?

ryanwilson (approved):

.github/workflows/generate_issues.yml
paulb777: Nope. Good catch. Thanks!

PR #7492, opened 2021-02-09
Firebase ML packaging updates

paulb777:
Remove  FirebaseMLModelInterpreter and FirebaseMLVision
CocoaPods and Zip Packaging for FirebaseMLModelDownloader
Disable Carthage building for any Swift pods since it's not feasible to construct the correct Swift modules from how we're currently piecing together the frameworks for Carthage.
If Carthage adds support for binary xcframeworks, it should become feasible at that point.


ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -469,28 +473,34 @@ struct FrameworkBuilder {
...
+    }
+    // Copy the module map to the destination.
+    let moduleDir = destination.appendingPathComponent("Modules")
+    do {
+      try FileManager.default.createDirectory(at: moduleDir, withIntermediateDirectories: true)
+    } catch {
+      let frameworkName: String = frameworks.first?.lastPathComponent ?? "<UNKNOWN"
+      fatalError("Could not create Modules directory for framework: \(frameworkName). \(error)")
     }
+    let modulemap = moduleDir.appendingPathComponent("module.modulemap")
```
paulb777, line 489: Below here is only reindented

ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -714,9 +726,17 @@ struct FrameworkBuilder {
       slicedBinariesForCarthage(fromFrameworks: slicedFrameworks,
                                 workingDir: platformFrameworksDir)
 
+    // Copy the framework in the appropriate directory structure.
```
paulb777, line 729: Refactored to build the framework in place instead of building, moving and deleting

CoreOnly/Sources/Firebase.h
```diff
@@ -85,18 +85,6 @@ FirebaseAnalytics dependency to your project to ensure Messaging works as intend
...
-  #endif
-
-  #if __has_include(<FirebaseMLModelInterpreter/FirebaseMLModelInterpreter.h>)
-    #import <FirebaseMLModelInterpreter/FirebaseMLModelInterpreter.h>
-  #endif
-
-  #if __has_include(<FirebaseMLVision/FirebaseMLVision.h>)
-    #import <FirebaseMLVision/FirebaseMLVision.h>
-  #endif
-
```
ryanwilson, line 99: This is a breaking change - even though it's deprecated, anyone using the frameworks currently would fail to find them anymore. This portion need to wait until Firebase 8

Firebase.podspec
```diff
@@ -158,19 +163,4 @@ Simplify your app development, grow your user base, and monetize more effectivel
...
-  s.subspec 'MLModelInterpreter' do |ss|
-    ss.dependency 'Firebase/CoreOnly'
-    ss.ios.dependency 'FirebaseMLModelInterpreter', '~> 7.6.0-beta'
-  end
-
-  s.subspec 'MLVision' do |ss|
-    ss.dependency 'Firebase/CoreOnly'
-    ss.ios.dependency 'FirebaseMLVision', '~> 7.6.0-beta'
-  end
-
```
ryanwilson, line 175: Also a breaking change, will need to wait until Firebase 8.

FirebaseCore/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# FirebaseCore 7.7.0
+- [changed] Removed the deprecated FirebaseMLModelInterpreter and FirebaseMLVision.
```
ryanwilson, line 2: This line will need to be removed

FirebaseMLCommon.podspec.json
```diff
@@ -1,44 +0,0 @@
-{
```
ryanwilson, line 1: This file should stay.

FirebaseMLModelInterpreter.podspec.json
```diff
@@ -1,38 +0,0 @@
-{
```
ryanwilson, line 1: This file should also stay.

FirebaseMLVision.podspec.json
```diff
@@ -1,44 +0,0 @@
-{
```
ryanwilson, line 1: This file should also stay.

ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
```diff
@@ -44,10 +44,7 @@ public let shared = Manifest(
     Pod("FirebasePerformance", platforms: ["ios", "tvos"], zip: true),
     Pod("FirebaseStorage", zip: true),
     Pod("FirebaseStorageSwift", isBeta: true),
-    Pod("FirebaseMLCommon", isClosedSource: true, isBeta: true, platforms: ["ios"]),
-    Pod("FirebaseMLModelInterpreter", isClosedSource: true, isBeta: true, platforms: ["ios"],
-        zip: true),
-    Pod("FirebaseMLVision", isClosedSource: true, isBeta: true, platforms: ["ios"], zip: true),
```
ryanwilson, line 50: I think these should stay, although I could be convinced that we should stop publishing "new" versions

ryanwilson (changes requested):

paulb777:
Is it a breaking change to change beta functionality?

There will continue to be the workaround of the pod name without the slash in the Podfile.

ryanwilson:
> Is it a breaking change to change beta functionality?
If it was just an API change it's fine, but it's a little different having the package manager just fail I would say. Is there a big rush to remove it here vs wait until Firebase 8?

> There will continue to be the workaround of the pod name without the slash in the Podfile.
Makes sense, and that's fair.

paulb777:
How about this - 

I restored the Firebase.h and Firebase.podspec to prevent CocoaPods installation breaking changes.
I'll do a follow-up PR to the release tooling to keep the three deprecated subspecs to 7.6.0-beta so we don't need to keep releasing them.
Carthage will just work by installing the latest published version.
Zip users can keep their old version of the ML frameworks.

ryanwilson:
> How about this -
> 
> I restored the Firebase.h and Firebase.podspec to prevent CocoaPods installation breaking changes.
> I'll do a follow-up PR to the release tooling to keep the three deprecated subspecs to 7.6.0-beta so we don't need to keep releasing them.
> Carthage will just work by installing the latest published version.
> Zip users can keep their old version of the ML frameworks.

Sounds like a great plan, thanks!

ryanwilson (approved):

Issue #7476, opened 2021-02-08
test_coverage workflow fails on external PRs


paulb777:
See #7472 

The `test_coverage` workflow should either be disabled or made to work for PRs from forked repos.

granluo:
If secrets are not able to be accessed, then Metrics Service cannot be used and so code coverage report is not able to be created. To avoid failing test workflows, we might need to disable this workflows under certain situations.
I made some experiments and it looks like the sdk repo can run workflows from a forked repo, not sure if secrets can be used or not yet. Will make more experiments.

paulb777:
Reopening. Still seeing failures in #7507

paulb777:
@granluo Still seeing failures in https://github.com/firebase/firebase-ios-sdk/pull/7549/checks?check_run_id=1921393904

Please test in a fork.

granluo:
Have test on a fork repo. The corresponding step is in the step `check files` of the job `Check changed files` 
https://github.com/firebase/firebase-ios-sdk/runs/1929701292?check_suite_focus=true

The failure is probably caused by that fork repo not syncing to the upstream yet.

paulb777:
I see thanks.

paulb777:
Reopening - the issue looks back in #7955

paulb777:
@granluo This is still failing in #8483 and #8484.  Please test on a forked PR. Reopening.

paulb777:
There is still a post-submit issue, but that's not a big deal. I'll close since presubmits are now working.

PR #7507, opened 2021-02-11
Output log error when RCNConfigFetch is already deallocated

funzin:
### Discussion
- If access `remoteConfig` while running` remoteConfig.fetch(withExpirationDuration:)` `completionHandler` might not be called because `RCNConfigFetch.dealloc` is called.
- It happens about once every five times when the app is launched for the first time.
- Changed to output LogError to judge whether it was deallocated

### Referrence
fetch(withExpirationDuration:): https://github.com/firebase/firebase-ios-sdk/blob/95cf1d3508ab2bfd412917722068ac10fe3979ae/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L229-L237
RCNConfigFetch.dealloc: https://github.com/firebase/firebase-ios-sdk/blob/95cf1d3508ab2bfd412917722068ac10fe3979ae/FirebaseRemoteConfig/Sources/RCNConfigFetch.m#L121-L123

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
@@ -134,6 +134,7 @@ static const NSInteger sFIRErrorCodeConfigFailed = -114;
   dispatch_async(_lockQueue, ^{
     RCNConfigFetch *strongSelf = weakSelf;
     if (strongSelf == nil) {
+      FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000076", @"RCNConfigFetch is already deallocated");
       return;
     }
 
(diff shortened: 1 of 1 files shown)


funzin:
@googlebot I signed it!

paulb777:
@funzin Thanks for the PR! The CI failures look to be the result of recent infrastructure changes (see #7476) and can be disregarded here. 

Routing to @karenyz for review.

maksymmalyhin:
@funzin Thank you for your contribution!

I would like to ask a bit more details in the issue you experience before merging the PR:
1. Were you actually able to reproduce the case when the instance of `RCNConfigFetch` was deallocated before the callback was called?
2. Could you provide steps and ideally a sample project to reproduce the issue you experience?

As far as I can see from the code, the instance of `RCNConfigFetch` should be deallocated only when FirebaseApp is deleted. Could you please check if `FirebaseApp.deleteApp()` (in Swift) or `-[FIRApp deleteApp:]` (in Objective-C) is called in your app? If not we probably should look for another reason of the issue.

funzin:
Our use of remoteConfig was not correct, it seems that this fix does not need to be included:pray:

>Were you actually able to reproduce the case when the instance of RCNConfigFetch was deallocated before the callback was called?

This is sample code.

```swift
import FirebaseRemoteConfig

DispatchQueue.global(qos: .default).async {
    _ = RemoteConfig()
}
_ = RemoteConfig()
```
Initialization may not be completed and may be deallocated depending on timing

paulb777:
Glad you figured it out and thanks for sharing the resolution!

PR #7515, opened 2021-02-11
ML model downloader fixes

maksymmalyhin:
- fix call handlers on the main queue
- fix `listDownloadedModels` method for real devices
- use a subfolder named "com.firebase.FirebaseMLModelDownloader" to store models
- `getModel()`: fail fast with a specific error when model name is empty


ryanwilson (approved):
Thanks!

PR #7516, opened 2021-02-11
Disable token test on forks

paulb777:
This fixes the fork CI issue for the remoteconfig/remoteconfig workflow that showed up in #7507.  The failure still exists for the `test_coverage` workflow. See #7476

#no-changelog


maksymmalyhin (approved):
👍

PR #7520, opened 2021-02-12
Update project.pbxproj for bundle_spec_test.json that was added in #7444

dconeybe:
This was done by running scripts/sync_project.rb, and should have been included in #7444 (but it was forgotten).


schmidt-sebastian (approved):
Shame on the author of #7444 :)

PR #7544, opened 2021-02-17
WIP: RTDB Swift Combine

mortenbekditlevsen:
This is basically a merge of two other works-in-progress: #6976 and #6568

Just so that we have a place to start experimenting with RTDB/Combine.


peterfriese:
Hi @mortenbekditlevsen -

sorry for closing this. We were expecting any in-flight PRs to automatically retarget to `master` once we merge `combine-main`.

Can you do this manually, and re-open the PR?

Thanks,
Peter

paulb777:
I reopened and changed the base branch to master. Make sure you resync the local branch.

mortenbekditlevsen:
I force pushed a commit that just contains the additions that were originally made on this branch. It had become too hard to clean up.

FirebaseTestingSupport/Database/Sources/FIRDataSnapshotFake.m
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "FirebaseTestingSupport/Database/Sources/Public/FirebaseDatabaseTestingSupport/FIRDataSnapshotFake.h"
+
+@implementation FIRDataSnapshotFake
+
+- (instancetype)init {
```
paulb777, line 19: If the warning is expected "Designated initializer missing a 'super' call to a designated initializer of the super class", please disable with a pragma. See other examples in the repo `git grep pragma`.

FirebaseTestingSupport/Database/Sources/FIRDatabaseReferenceFake.m
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "FirebaseTestingSupport/Database/Sources/Public/FirebaseDatabaseTestingSupport/FIRDatabaseReferenceFake.h"
+#import "FirebaseTestingSupport/Database/Sources/Public/FirebaseDatabaseTestingSupport/FIRDataSnapshotFake.h"
+
+@implementation FIRDatabaseReferenceFake
+
+- (instancetype)init {
+  // The object is partially initialized. Make sure the methods used during testing are overridden.
```
paulb777, line 21: Disable build warning here too

FirebaseTestingSupport/Database/Tests/DatabaseReferenceFakeTests.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    let completionExpectation = expectation(description: "Completion called")
+    fakeReference.setValue(value, withCompletionBlock: { error, reference in
+      XCTAssertNil(error)
+      XCTAssertEqual(reference, fakeReference)
+
+      completionExpectation.fulfill()
+    })
+
+    wait(for: [completionExpectation], timeout: 0.1)
```
paulb777, line 45: This may be too short for CI systems - but we can see ...

paulb777 (commented):
This is generally looking good!

I just have a few nit-picky comments/questions. My Combine intuition isn't yet developed enough for the naming questions. Maybe @peterfriese has some suggestions?

- I believe `@available` checks are needed to build. See https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCombineSwift/Sources/Auth/Auth%2BCombine.swift#L15

FirebaseCombineSwift/Tests/Unit/Database/DatabaseReferenceTests.swift
```diff
@@ -0,0 +1,143 @@
...
+    let model = Model(a: "hello", b: 42)
+    let future = fake.setValue(from: model)
+
+    let valueExpectation = expectation(description: "Future sends void value")
+    let completionExpectation = expectation(description: "Future completes")
+
+    cancellable = future.sink { completion in
+      completionExpectation.fulfill()
+    } receiveValue: { _ in
+      valueExpectation.fulfill()
```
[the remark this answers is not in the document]
mortenbekditlevsen: I guess that's guaranteed by the compiler too, but could of course be added.

FirebaseTestingSupport/Database/Tests/DatabaseReferenceFakeTests.swift
mortenbekditlevsen: The fake calls the completion synchronously, so there's actually no real waiting except maybe some context switching.

FirebaseCombineSwift/Sources/Database/DatabaseReference+Combine.swift
```diff
@@ -0,0 +1,110 @@
...
+          }
+        }
+      }
+    }
+
+    // NOTE: Should this perhaps be named XXXPublisher(...)?
+    // If so, what should it be? 'modelPublisher', 'decodedPublisher'?
+    // Should this be built upon the snapshotPublisher, or is it an ok
+    // performance improvement to create this 'directly'?
+    func observe<T: Decodable>(_ eventType: DataEventType,
```
peterfriese, line 77: TL;DR:
* keep the current naming
* and add a new method `valuePublisher` or `snapshotPublisher` that just calls `observe`

## Longer explanation:
So far, we've tried to keep the naming as close to the original API as possible, to make it easy for people to discover the Combine methods when invoking code completion.

**Hence, I think it'd be good to keep the naming as-is.**

However, it might also be useful to use XXXPublisher to keep it more in line with the work on Firestore in PR #7623.

Looking at the JavaScript implementation, they use 

* `get` for reading once
* `onValue` for reading with an observer (to read just once, they pass a parameter)

Following this, I'd suggest adding a method named `valuePublisher`, and call `observe` (i.e. this method here)


@ryanwilson and @paulb777 - I'd love to hear your thoughts on this

peterfriese (changes requested):
I added some thoughts to the naming discussion, but would like to hear @ryanwilson's opinion as well.

Also, some doc comments are missing.

FirebaseCombineSwift/Sources/Database/DatabaseReference+Combine.swift
```diff
@@ -0,0 +1,110 @@
...
+              promise(.success(()))
+            }
+          })
+        } catch {
+          // NOTE: This is only correct since we know that if `try setValue` fails
+          // then the completion is not called.
+          // Would it be better/clearer to 'spell out' the setValue and basically
+          // copy what is in DatabaseReference+Decodable in order to make it clear
+          // that the `try` is connected to the encoding and not the setting of the
+          // value?
```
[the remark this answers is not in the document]
paulb777: Agreed

mortenbekditlevsen:
Latest commit addresses the latest batch of comments regarding doc comments and adds an overload of `observe` called `valuePublisher` as suggested.

mortenbekditlevsen:
Question: #6940 uses `Result<Void, Error>` for completions that can fail, but have no other parameters.

I initially modeled `setValue` in FirebaseDatabaseSwift similarly, but I was advised to change this to a completion of `Error?` instead.

Should we unify this behavior? 

@peterfriese @ryanwilson

paulb777:
What are the advantages and disadvantages of the two different API styles?

mortenbekditlevsen:
The most important thing is of course the consistency, since the difference between the two is so small.

If I should argue for my favorite (`Result<Void, Error>`) I'd say that it carries slightly more semantic meaning than `Error?`.
In the former, the success path is the success case of a Result. In the latter it's the absence of an error.
And similarly, in the former the error value is bound to the failure case of the result - so it's very explicitly a failure and not just some 'random' value.

It's subtle and I'd say that most people would be ok with either choice. :-)

I also personally like the equality between:

```
a(completion: (Result<Void, Error>) -> Void)

a() -> Future<Void, Error>

a() async throws -> Void
```
(Even though the returned void in the last case would not be spelled out...)

paulb777:
@morganchen12 Given the minor differences, I'm fine going with your preference.  If something compelling comes up in the future, we can change then.

ryanwilson:
> @mortenbekditlevsen Given the minor differences, I'm fine going with your preference. If something compelling comes up in the future, we can change then.

SGTM as well, apologies for the delay in response.

mortenbekditlevsen:
@paulb777  Is this PR still relevant for the Swift Modernisation efforts?

It's marked as 'Changes requested', but I can't find any concrete suggestions that are not already addressed.

paulb777:
@mortenbekditlevsen Sorry for losing track of this PR. It looks like it needs to be rebased or merged with master. @peterfriese Do you see anything else to resolve before merging?

FirebaseCombineSwift/Sources/Database/DatabaseReference+Combine.swift
```diff
@@ -0,0 +1,185 @@
...
+  import Combine
+  import FirebaseDatabase
+  import FirebaseDatabaseSwift
+
+  @available(swift 5.0)
+  @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+  public extension DatabaseReference {
+    // MARK: - Setting values on a DatabaseReference
+
+    /**
```
peterfriese, line 26: For new Swift code, we've been using triple-slash comments, see https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCombineSwift/Sources/Firestore/CollectionReference%2BCombine.swift#L31, for example.

Can you update all comments in this PR accordingly?

FirebaseDatabaseSwift.podspec
```diff
@@ -5,7 +5,7 @@
 
 Pod::Spec.new do |s|
   s.name                    = 'FirebaseDatabaseSwift'
-  s.version                 = '7.9.0-beta'
+  s.version                 = '8.0.0-beta'
```
peterfriese, line 8: Needs to be updated.

FirebaseDatabaseSwift/Sources/Codable/DatabaseReference+WriteEncodable.swift
```diff
@@ -18,19 +18,24 @@ import Foundation
...
-  /// See `Database.Encoder` for more details about the encoding process.
-  ///
-  /// - Parameters:
-  ///   - value: An instance of `Encodable` to be encoded to a document.
-  ///   - encoder: An encoder instance to use to run the encoding.
-  ///   - completion: A block to execute once the value has been successfully
-  ///                 written to the server. This block will not be called while
-  ///                 the client is offline, though local changes will be visible
-  ///                 immediately.
+  /** Encode an `Encodable` instance and write it to this Firebase Database location.
```
peterfriese, line 21: Comment style

FirebaseDatabaseTestingSupport.podspec
```diff
@@ -0,0 +1,57 @@
+Pod::Spec.new do |s|
+  s.name                    = 'FirebaseDatabaseTestingSupport'
+  s.version                 = '1.0.0'
+  s.summary                 = 'Firebase SDKs testing support types and utilities.'
+
+  s.description      = <<-DESC
+  Type declarations and utilities needed for unit testing of the code dependent on Firebase SDKs
```
peterfriese, line 7: Make sure to add info that the podspecs are just there for testing. See https://github.com/firebase/firebase-ios-sdk/commit/397215b9f304ba320c7d18c9839509535d8624e1#diff-221380d0cac6c2109a93596f5086bc8691024208216ff59d18546a8162162550R7

mortenbekditlevsen:
I agree that that is a good idea. 😊 I could look into extensions for returning async sequences or an Observable model wrapper type. I’m experimenting with a `Live` generic wrapper around a Codable model entity that can be returned from RTDB or firestore.

PR #7545, opened 2021-02-17
Fix FIRFunctions.useEmulatorWithHost scheme

### LINKED ISSUE #7537: Calling Cloud Functions locally from iOS gives an "Insecure fetch request" error

  * Xcode version: 12.4
  * Firebase SDK version: 7.6.0
  * Firebase Component: Functions 

When running the emulator locally and calling a HTTPS function, the console outputs the following message:

```
 Insecure fetch request has a scheme (localhost) not found in fetcher allowedInsecureSchemes ((
    http
)): localhost:5001/****/us-central1/testFunction
```

#### Steps to reproduce:

Create a cloud function:

Run the emulator locally

Setup the iOS app to call your function and call it

#### Relevant Code:

```
exports.testFunction = functions.https.onCall(async (data, context) => {
}
```

```
firebase emulators:start
```

on iOS:

```
            let functions = Functions.functions()
            functions.useEmulator(withHost: "localhost", port: 5001)

            functions.httpsCallable("testFunction").call([ ...
```

### LINKED ISSUE #7538: Firebase Functions Emulator Fails Over Local Network

  * Xcode version: 12.4
  * Firebase SDK version: 7.6.0
  * Installation method: ` Swift Package Manager`
  * Firebase Component: _Emulators_ , _Functions_

#### Steps to reproduce:

1. Create a new Firebase project via the console
2. Create a new local Firebase project using the CLI
    * Enable Functions and Emulators
    * Enable the Functions and Auth Emulators
    * Use the default ports for setting up the emulators
3. Implement a simple callable `helloWorld` function (see below)
4. Launch the emulators (`firebase emulators:start`)
5. Create a new (SwiftUI) project
6. Download `GoogleService-Info.plist` and add to project
6. Enable `App Transport Security` (allow arbitrary loads)
7. Initialise the emulators (see Swift code below)
8. Call `helloWorld` callable function from Swift (see code below)

#### Expected behaviour

* It is possible to call the `helloWorld` callable function from the iOS app

#### Observed behaviour

* `GTMSessionFetcher` issues error message about `http` not being in list of allowed insecure schemes:

```
2021-02-16 16:32:29.570523+0100 FirebaseEmulatorDemo[47823:2419456] Insecure fetch request has a scheme (localhost) not found in fetcher allowedInsecureSchemes ((
    http
)): localhost:5001/peterfriese-friction-emulators/us-central1/helloWorld
Error Domain=com.google.GTMSessionFetcher Code=-5 "(null)" UserInfo={NSErrorFailingURLStringKey=localhost:5001/peterfriese-friction-emulators/us-central1/helloWorld}
```

##### Callable Function:

```javascript
exports.helloWorld = functions.https.onCall((data, context) => {
    return {
        message: "Hello World from Firebase" 
    };  
});
```

#### Swift code for initialising emulators and calling the Function
```swift
class FirebaseEmulatorDemo: ObservableObject {
  
  lazy var functions = Functions.functions()
  
  init() {
    FirebaseApp.configure()
    
    // set up emulators
    Auth.auth().useEmulator(withHost: "localhost", port: 9099)
    functions.useEmulator(withHost: "localhost", port: 5001)
  }
  
  func helloWorld() {
    let helloWorld = functions.httpsCallable("helloWorld")
    helloWorld.call() { result, error in
      if let error = error {
        print(error)
      }
      else {
        if let message = (result?.data as? [String: Any])?["message"] as? String {
          print("The function returned: \(message)")
        }
      }
    }
  }
  
}
```

### Related material
* #7342 mentions a similar issue
* It is interesting to note that these lines of code do suggest that the expected behaviour should indeed work: 

https://github.com/firebase/firebase-ios-sdk/blob/02488e99b74f5da122b169cbc5fc6a154cff28a3/Functions/FirebaseFunctions/FIRFunctions.m#L247-L251

samtstern:

### Discussion

  * Fixes #7537
  * Fixes #7538

### Testing

  * Updated unit tests

### API Changes

  * This technically could be considered a breaking change however the original implementation was a bug. The API is meant to be `useEmulator("locahost", 1234)` but right now it requires `useEmulator("http://localhost", 1234)`

### DIFF
--- a/Functions/FirebaseFunctions/FIRFunctions.m
+++ b/Functions/FirebaseFunctions/FIRFunctions.m
@@ -154,7 +154,7 @@ NSString *const kFUNDefaultRegion = @"us-central1";
 
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
   NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
-  NSString *origin = [NSString stringWithFormat:@"%@:%li", host, (long)port];
+  NSString *origin = [NSString stringWithFormat:@"http://%@:%li", host, (long)port];
   _emulatorOrigin = origin;
 }
 
--- a/Functions/FirebaseFunctions/FIRFunctions.m
+++ b/Functions/FirebaseFunctions/FIRFunctions.m
@@ -154,7 +154,7 @@ NSString *const kFUNDefaultRegion = @"us-central1";
 
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
   NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
-  NSString *prefix = [host hasPrefix:@"http"] ? @"" :  @"http://";
+  NSString *prefix = [host hasPrefix:@"http"] ? @"" : @"http://";
   NSString *origin = [NSString stringWithFormat:@"%@%@:%li", prefix, host, (long)port];
   _emulatorOrigin = origin;
 }
--- a/Functions/FirebaseFunctions/FIRFunctions.m
+++ b/Functions/FirebaseFunctions/FIRFunctions.m
@@ -149,7 +149,7 @@ NSString *const kFUNDefaultRegion = @"us-central1";
 }
 
 - (void)useLocalhost {
-  [self useEmulatorWithHost:@"http://localhost" port:5005];
+  [self useEmulatorWithHost:@"localhost" port:5005];
 }
 
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
--- a/Functions/FirebaseFunctions/FIRFunctions.m
+++ b/Functions/FirebaseFunctions/FIRFunctions.m
@@ -154,7 +154,8 @@ NSString *const kFUNDefaultRegion = @"us-central1";
 
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
   NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
-  NSString *origin = [NSString stringWithFormat:@"http://%@:%li", host, (long)port];
+  NSString *prefix = [host hasPrefix:@"http"] ? @"" :  @"http://";
+  NSString *origin = [NSString stringWithFormat:@"%@%@:%li", prefix, host, (long)port];
   _emulatorOrigin = origin;
 }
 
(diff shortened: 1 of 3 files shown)


samtstern:
cc @peterfriese

Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -149,12 +149,12 @@ - (instancetype)initWithProjectID:(NSString *)projectID
...
 
 - (void)useLocalhost {
-  [self useEmulatorWithHost:@"http://localhost" port:5005];
+  [self useEmulatorWithHost:@"localhost" port:5005];
 }
 
 - (void)useEmulatorWithHost:(NSString *)host port:(NSInteger)port {
   NSAssert(host.length > 0, @"Cannot connect to nil or empty host");
-  NSString *origin = [NSString stringWithFormat:@"%@:%li", host, (long)port];
+  NSString *origin = [NSString stringWithFormat:@"http://%@:%li", host, (long)port];
```
maksymmalyhin, line 157: Yeah, it is a breaking change. I'm a bit hesitant to push this change as is, especially taking into account that the bug is described as intended behaviour in the [docs](https://firebase.google.com/docs/functions/local-emulator#ios---swift). Maybe we can make it non-breaking by adding the http prefix conditionally, e.g. 
```
NSString *prefix = [host hasPrefix:@"http"] ? @"http://" : @"";
NSString *origin = [NSString stringWithFormat:@"%@%@:%li", prefix, host, (long)port];
```
In this case both the intended behaviour and the old workaround would work. We can remove the workaround at Firebase 8 where we intend to do other breaking changes (if decided so please crate a ticket). WDYT?

maksymmalyhin (commented):
Thank you for the fix! I added one comment on the breaking change topic.

Functions/FirebaseFunctions/FIRFunctions.m
samtstern: Those docs reference the old `.useFunctionsEmulator(origin: "http://localhost:5001")`, this PR only affects the newer `useEmulator("localhost", 5001)` function.

But I agree we can make this non-breaking.

maksymmalyhin: Ah, true, sorry, I misread the snippet.

samtstern: Ok I implemented your change, thank you!

maksymmalyhin (approved):
LGTM, thanks!

yuchenshi (approved):

maksymmalyhin:
@samtstern Looks like you need to run `brew upgrade clang-format` to install 11.1.0 version and then run `./scripts/style.sh` to fix the formatting to make `check` happy.

PR #7547, opened 2021-02-17
Combine Support: Cloud Functions

peterfriese:
This PR adds Combine support for Firebase Cloud Functions.

This will allow callers to invoke Firebase Cloud Functions like this:

```swift
let helloWorld = Functions.functions().httpsCallable("helloWorld")
helloWorld.call(")
  .sink { completion in
    switch completion {
      case .finished:
        print("Finished")
      case let .failure(error):
        print("\(error.localizedDescription)")
    }
  } receiveValue: { functionResult in
    if let result = functionResult.data as? String {
      print("The function returned: \(result)")      
    }
  }
  .store(in: &cancellables)
```


FirebaseCombineSwift/Tests/Unit/FirebaseCombine-unit-Bridging-Header.h
```diff
@@ -14,9 +14,25 @@
...
+
+// Firebase Auth
+
+#import <FirebaseAuth/FIREmailAuthProvider.h>
+// ^-- importing FIREmailAuthProvider like this will throw an error from the linter
+
+// #import "FirebaseAuth/Sources/Public/FirebaseAuth/FIREmailAuthProvider.h"
+// ^-- uncommenting this will break the compile
+
+#import "FirebaseAuth/Sources/AuthProvider/Email/FIREmailPasswordAuthCredential.h"
```
peterfriese: Trying to resolve this conundrum...

So either I use angle-bracket imports (compiles fine, so that the tests actually see the Combine extensions I made available in my code, BUT has the linter complain about using angle-bracket imports).

Or, I use double-quote imports (which makes the linter happy, but breaks the compile).

@paulb777 any ideas what I might be doing wrong? This happens both in Xcode 12.4 and Xcode 12.5.

maksymmalyhin: FIREmailAuthProvider.h is a public header so should be imported with #import <FirebaseAuth/FIREmailAuthProvider.h> outside of the FirebaseAuth module. We should fix the linter...

maksymmalyhin: I added a workaround for a similar case at some point #7017. Now we should adjust it again... I think we can try to check for `/Tests/` in the file path.

paulb777: Yep. https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/check_imports.swift#L144 assumes that any non-library code is tests. Bridging headers should also be excluded there.

peterfriese: Ok, good to hear I wasn't completely in the wrong :-)

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
```diff
@@ -0,0 +1,140 @@
...
+import XCTest
+@testable import FirebaseFunctions
+
+class MockFunctions: Functions {
+  var mockCallFunction: () throws -> HTTPSCallableResult?
+  override func callFunction(_ name: String,
+                             with data: Any?,
+                             timeout: TimeInterval,
+                             completion: @escaping (HTTPSCallableResult?, Error?) -> Void) {
+    do {
```
maksymmalyhin, line 26: nit: I would recommend to validate the input parameters as well to make sure the Combine wrapper doesn't miss or mix anything.

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
```diff
@@ -0,0 +1,140 @@
...
+class HTTPSCallableTests: XCTestCase {
+  func testCallWithoutParametersSuccess() {
+    // given
+    var cancellables = Set<AnyCancellable>()
+    let funcationWasCalledExpectation = expectation(description: "Function was called")
+
+    let functions = MockFunctions.functions()
+    let expectedResult = "mockResult w/o parameters"
+    functions.mockCallFunction = {
+      HTTPSCallableResult(data: expectedResult)
```
maksymmalyhin, line 44: nit: I would suggest adding a specific expectation to validate that the method was called and called only once. It may seem obvious and redundant now as the implementation is pretty simple, but may protect us from unintended typos or accident changes in the future.

peterfriese: I'm going to use assertForOverFulfill for this.

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
peterfriese: Added a callback that allows the test to verify the correct argument was sent.

maksymmalyhin: Thanks! Would you like to validate `name` and `timeout`?

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
```diff
@@ -0,0 +1,152 @@
...
+    var cancellables = Set<AnyCancellable>()
+    let functionWasCalledExpectation = expectation(description: "Function was called")
+    functionWasCalledExpectation.assertForOverFulfill = true
+    let callbackWasCalledExpectation = expectation(description: "Callback was called")
+
+    let functions = MockFunctions.functions()
+    let expectedResult = "mockResult w/o parameters"
+
+    functions.mockCallFunction = {
+      functionWasCalledExpectation.fulfill()
```
maksymmalyhin, line 49: Nice! Nit: Would you like to add it to other tests for consistency?

peterfriese:
If folks are happy with this, I'll go ahead and merge into combine-main.

@ryanwilson 
@paulb777 
@maksymmalyhin

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
```diff
@@ -36,12 +38,16 @@ class HTTPSCallableTests: XCTestCase {
   func testCallWithoutParametersSuccess() {
     // given
     var cancellables = Set<AnyCancellable>()
-    let funcationWasCalledExpectation = expectation(description: "Function was called")
+    let functionWasCalledExpectation = expectation(description: "Function was called")
+    functionWasCalledExpectation.assertForOverFulfill = true
```
maksymmalyhin, line 42: If I'm not mistaken the expectation has [assertForOverFulfill == true](https://developer.apple.com/documentation/xctest/xctestexpectation/2806575-assertforoverfulfill) is `true` and [expectedFulfillmentCount == 1](https://developer.apple.com/documentation/xctest/xctestexpectation/2806572-expectedfulfillmentcount) by default here, so the line may be removed.

FirebaseCombineSwift.podspec
```diff
@@ -53,17 +55,18 @@ Combine Publishers for Firebase.
...
+    unit_tests.requires_app_host = true
     unit_tests.pod_target_xcconfig = {
-      'SWIFT_OBJC_BRIDGING_HEADER' => '$(PODS_TARGET_SRCROOT)/FirebaseCombineSwift/Tests/Unit/FirebaseCombineSwift-unit-Bridging-Header.h'
+      'SWIFT_OBJC_BRIDGING_HEADER' => '$(PODS_TARGET_SRCROOT)/FirebaseCombineSwift/Tests/Unit/FirebaseCombine-unit-Bridging-Header.h'
     }
-    unit_tests.requires_app_host = true
     unit_tests.dependency 'OCMock'
   end
-end
+end
```
maksymmalyhin, line 72: please add a new line

maksymmalyhin (approved):
LGTM 👍 , a couple minor comments.

FirebaseCombineSwift/Tests/Unit/Functions/HTTPSCallableTests.swift
peterfriese: You're right, thanks for pointing this out.

maksymmalyhin (approved):
Thank you for the update! LGTM on CI green 👍

FirebaseCombineSwift.podspec
```diff
@@ -53,17 +55,18 @@ Combine Publishers for Firebase.
   }
 
   s.test_spec 'unit' do |unit_tests|
-    unit_tests.platforms = {:ios => '13.0', :osx => '10.11', :tvos => '10.0'}
+    unit_tests.scheme = { :code_coverage => true }
+    unit_tests.platforms = {:ios => '13.0', :osx => '10.12', :tvos => '10.0'}
```
paulb777, line 59: I suspect tvos should also be 13 here and macos 10.15?

PR #7548, opened 2021-02-17
Creates auth credential for Game Center sign in

ziadtamim:
Creates a `AuthCredential` for a Game Center sign in.

```swift
 GameCenterAuthProvider.getCredential()
  .sink { completion in
    /* On completion */
  } receiveValue: { credential in
    /* On received credential */    
  }
  .store(in: &cancellables)
```


GameCenterAuthProvider+Combine.swift
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+#if canImport(Combine) && swift(>=5.0) && canImport(FirebaseAuth)
+
+  import Combine
+  import FirebaseAuth
+
+  @available(swift 5.0)
+  @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+
```
peterfriese, line 22: Remove blank line

GameCenterAuthProvider+Combine.swift
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#if canImport(Combine) && swift(>=5.0) && canImport(FirebaseAuth)
+
+  import Combine
+  import FirebaseAuth
+
+  @available(swift 5.0)
+  @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
```
peterfriese, line 21: According to https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Sources/Public/FirebaseAuth/FIRGameCenterAuthProvider.h#L47, GameCenterAuthProvider is not available on watchOS.

@rosalyntan according to [Apple's docs](https://developer.apple.com/documentation/gamekit/gklocalplayer), GKPlayer is available on watchOS. Can you comment on whether there are other reasons why we specifically exclude watchOS in [this line](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Sources/Public/FirebaseAuth/FIRGameCenterAuthProvider.h#L47)?

peterfriese (changes requested):
Thanks for sending this PR, @ziadtamim!

A couple of notes:
- Please make sure the file is in the correct folder
- Please add some tests (see [FIRSignInWithGameCenterTests.m](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Tests/Unit/FIRSignInWithGameCenterTests.m))

Also, always feel free to use draft PRs when staging your work.

ziadtamim:
> Thanks for sending this PR, @ziadtamim!
> 
> A couple of notes:
> 
> * Please make sure the file is in the correct folder
> * Please add some tests (see [FIRSignInWithGameCenterTests.m](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Tests/Unit/FIRSignInWithGameCenterTests.m))
> 
> Also, always feel free to use draft PRs when staging your work.

@peterfriese Tests added for `SignInWithGameCenter` and `GameCenterAuthProvider`.

GameCenterAuthProvider+Combine.swift
rosalyntan: I believe watchOS is excluded because Apple deprecated the `playerID` property in watchOS 6, but the replacement `teamPlayerID` is not supported on watchOS.

Adding @ryanwilson who reviewed the PR that added watchOS support to auth.

peterfriese (approved):
LGTM, thanks!

GameCenterAuthProvider+Combine.swift
ryanwilson: Good find, I believe that's the reason!

peterfriese: Ok, so should we

1. completely remove watchOS from the list of supported OSes, 
2. or state that is supported up to watchOS 5?

paulb777: Remove. There's no reason to add support only for old versions.

peterfriese: @ziadtamim, can you address this feedback, please?

ziadtamim: Done under https://github.com/firebase/firebase-ios-sdk/pull/7599

PR #7556, opened 2021-02-18
Stage to SpecsStaging instead of cpdc

paulb777:
Change release process to stage releases to a public repo instead of a private one.


maksymmalyhin (approved):
👍 LGTM

PR #7557, opened 2021-02-18
Functions Combine in protocol extension + tests example

maksymmalyhin:


peterfriese:
I think we can close this for now, as we've agreed to make the initiailiser public in https://github.com/firebase/firebase-ios-sdk/pull/7547

PR #7560, opened 2021-02-18
Automate Release zip build

paulb777:
Use Xcode 12.2 to build the zip and built the first real RC via GHA. See https://github.com/firebase/firebase-ios-sdk/actions/runs/579356249. 

I noticed that FirebaseAnalytics hasn't been built for Carthage since 7.4.0 and will follow up on that in a subsequent PR.

#no-changelog

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -27,6 +27,8 @@ jobs:
     runs-on: macOS-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Xcode 12.2
+      run: sudo xcode-select -s /Applications/Xcode_12.2.app/Contents/Developer
     - name: Build
       run: |
         cd ReleaseTooling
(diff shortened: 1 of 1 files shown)


.github/workflows/zip.yml
```diff
@@ -27,18 +27,22 @@ jobs:
...
+      run: sudo xcode-select -s /Applications/Xcode_12.2.app/Contents/Developer
     - name: Build
       run: |
         cd ReleaseTooling
         swift build -v
 
   package:
     # Don't run on private repo.
-    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
+#    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
```
maksymmalyhin, line 39: Is it necessary or just for testing and will be removed?

maksymmalyhin (approved):

.github/workflows/zip.yml
paulb777: Yep - just for full testing. I'll remove after green

paulb777:
FirebaseAnalytics is not built for Carthage because it provides only an xcframework now and Carthage doesn't yet support binary xcframeworks.

PR #7562, opened 2021-02-18
FirebaseAnalytics SwiftUI API for `screen_view`

ryanwilson:
This is a new FirebaseAnalyticsSwift SDK that adds native Swift
extensions for FirebaseAnalytics, starting with a SwiftUI screen view
function.

#no-changelog

**TODO**:
- [x] Add GHA tests for the Podspec

### DIFF
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
     - 'FirebaseAnalytics.podspec.json'
+    - 'FirebaseAnalyticsSwift**'
     - 'GoogleAppMeasurement.podspec.json'
     - 'Gemfile'
   schedule:
@@ -22,6 +23,8 @@ jobs:
       run: scripts/setup_bundler.sh
     - name: GoogleAppMeasurement
       run: scripts/third_party/travis/retry.sh pod spec lint GoogleAppMeasurement.podspec.json
+    - name: FirebaseAnalyticsSwift
+      run: scripts/third_party/travis/retry.sh pod spec lint FirebaseAnalyticsSwift.podspec
 
 # TODO: Consider pushing GoogleAppMeasurement.podspec.json to SpecsStaging to enable similar test
 # for FirebaseAnalytics.podspec.json
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -24,7 +24,7 @@ jobs:
     - name: GoogleAppMeasurement
       run: scripts/third_party/travis/retry.sh pod spec lint GoogleAppMeasurement.podspec.json
     - name: FirebaseAnalyticsSwift
-      run: scripts/third_party/travis/retry.sh pod spec lint FirebaseAnalyticsSwift.podspec
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseAnalyticsSwift.podspec --platforms=ios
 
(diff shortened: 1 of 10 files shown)


FirebaseAnalyticsSwift/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Unreleased
+
+# 7.6.0-Beta
```
paulb777, line 3: Delete - will be at least 7.8.0-Beta

FirebaseAnalyticsSwift/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Unreleased
+- Initial public beta release. Introduces new SwiftUI friendly APIs for
+  screen tracking. To use, add `pod 'FirebaseAnalyticsSwift'` to the Podfile or
```
paulb777, line 3: Will need a version specifier for the beta.  See the 7.0.0 release notes for an example

paulb777 (approved):

.github/workflows/analytics.yml
```diff
@@ -22,6 +23,8 @@ jobs:
       run: scripts/setup_bundler.sh
     - name: GoogleAppMeasurement
       run: scripts/third_party/travis/retry.sh pod spec lint GoogleAppMeasurement.podspec.json
+    - name: FirebaseAnalyticsSwift
+      run: scripts/third_party/travis/retry.sh pod spec lint FirebaseAnalyticsSwift.podspec
```
paulb777, line 27: This should use the script and be a regular pod lib lint since it's a source pod, not a binary pod like GoogleAppMeasurement

ryanwilson:
Failure during linting:

```
 Undefined symbols for architecture arm64:
      "_OBJC_CLASS_$_FIRAnalytics", referenced from:
          objc-class-ref in Analytics+SwiftUI.o
    ld: symbol(s) not found for architecture arm64
```

Any ideas if I'm doing anything wrong Paul? Added it as a dependency already and not sure what else could be up.

FirebaseAnalyticsSwift.podspec
```diff
@@ -0,0 +1,30 @@
...
+    :git => 'https://github.com/Firebase/firebase-ios-sdk.git',
+    :tag => 'CocoaPods-' + s.version.to_s
+  }
+
+  s.swift_version           = '5.0'
+  s.ios.deployment_target   = '13.0'
+
+  s.cocoapods_version       = '>= 1.10.0'
+  s.prefix_header_file      = false
+
```
paulb777, line 24: Missing platforms specification. Presumably, it's iOS only?

ryanwilson: I thought I could get away with `spec.ios.deployment_target` alone, but I'll try to add the platform specification as well.

ryanwilson: No luck, still running into the same issue. I can reproduce locally, except Xcode seems to know what's going on (autocomplete works for both `import FirebaseAnalytics` and Analytics specific symbols).

paulb777:
The podspec needs to be static since it depends on Analytics.

Add `  s.static_framework = true`

ryanwilson:
> The podspec needs to be static since it depends on Analytics.
> 
> Add ` s.static_framework = true`

Ahh yep, that was it. Thank you!! :D

htcgh (approved):
Thanks!

PR #7564, opened 2021-02-19
Remove unnecessary platform check

ryanwilson:
This isn't actually needed since the class won't be available on platforms it doesn't support.

#no-changelog


paulb777 (approved):
LGTM on green

PR #7567, opened 2021-02-19
Add ByteStream to handle both NSInputStream and istream

wu-hui:
This is intended to merge into https://github.com/firebase/firebase-ios-sdk/pull/7527 before that gets merged.


Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -0,0 +1,104 @@
...
+    EXPECT_EQ(actual, source);
+  }
+
+  // ReadUntil
+  {
+    auto stream = stream_factory_->CreateByteStream(source);
+    std::string actual;
+    std::string chars = "}{";
+    int i = 0;
+    while (true) {
```
var-const, line 90: Consider
```cpp
for (int i = 0; i != kMaxIterations; ++i) {
  char delim = i % 2 ? '}' : '{';
  auto result = stream->ReadUntil(delim), 1000);
}
```

var-const (changes requested):

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -0,0 +1,104 @@
...
+ * limitations under the License.
+ */
+
+#include "Firestore/core/test/unit/util/byte_stream_test.h"
+
+namespace firebase {
+namespace firestore {
+namespace util {
+namespace {
+
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/src/util/byte_stream.h
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Returns the result from a read operation.
+   *
+   * Result will be in a std:string instance when read was successful, otherwise
+   * a `Status` instance will be returned, indicating the `ByteStream` instance
+   * has failed.
+   */
+  const StatusOr<std::string>& result() const {
```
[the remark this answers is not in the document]
wu-hui: I think it makes sense.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -0,0 +1,116 @@
...
+
+  if (read > 0) {
+    buffer_.append(result.substr(0, static_cast<unsigned long>(read)));
+  }
+
+  return static_cast<int32_t>(read);
+}
+
+StreamReadResult ByteStreamNSInputStream::EofResult() {
+  return StreamReadResult(buffer_, true);
```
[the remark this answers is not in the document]
var-const: For extra safety, can you clear the buffer?
```cpp
StreamReadResult result(std::move(buffer_), true);
buffer_.clear();
return result;
```

After being moved from, objects are in a "valid but unspecified state". In practice, the string will likely be empty, but it still feels safer to give them a known empty value.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -51,9 +49,8 @@
 
   // Still not found, return the whole `buffer_` and clear it.
   if (found_at == std::string::npos) {
-    std::string buffer_copy(buffer_);
+    auto read_result = StreamReadResult(std::move(buffer_), eof());
```
var-const, line 52: Nit: you could just do
```cpp
StreamReadResult result(std::move(buffer_), eof());
```

(applies below as well)

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -26,23 +26,21 @@
...
+    found_at = buffer_.find(delim, last_pos);
+    last_pos = buffer_.size();
+    if (found_at != std::string::npos || read == 0) {
       break;
     }
-
-  } while (buffer_.size() < max_length);
+  }
 
   // One last try since the loop might break because eof or max_length reached.
```
var-const, line 45: Question: is this check still necessary? If yes, can you add a test that fails without it?

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -64,24 +61,31 @@
   return read_result;
 }
 
-StreamReadResult ByteStreamNSInputStream::Read(size_t max_length) {
-  auto read = ReadToBuffer(max_length);
+StreamReadResult ByteStreamApple::Read(size_t max_length) {
+  // Serve from buffer_
+  if (buffer_.size() >= max_length) {
```
var-const, line 66: Can you add a test that would fail without this addition?

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -64,24 +61,31 @@
...
+  auto read = ReadToBuffer(max_length - buffer_.size());
   if (read < 0) {
     return ErrorResult();
   }
   if (read == 0) {
     return EofResult();
   }
 
-  std::string buffer_copy(buffer_);
+  auto read_result = StreamReadResult(std::move(buffer_), eof());
```
var-const, line 81: Now that this pattern is repeated a few times, what do you think about creating a helper function?
```cpp
StreamReadResult ConsumeBuffer(bool is_eof) {
  StreamReadResult result(std::move(buffer_), is_eof);
  buffer_.clear();
  return result;
}

// ...
return ConsumeBuffer(eof());
```

Firestore/core/src/bundle/bundle_reader.cc
```diff
@@ -98,20 +98,21 @@ std::unique_ptr<BundleElement> BundleReader::ReadNextElement() {
...
 
   // Underlying stream is closed, and there happens to be no more data to
   // process.
-  if (result.eof() && result.result().ValueOrDie().empty()) {
+  if (result.eof() && result.ValueOrDie().empty()) {
     return absl::nullopt;
   }
 
-  return absl::make_optional(std::move(result.result().ValueOrDie()));
+  return absl::make_optional(std::move(result.ValueOrDie()));
```
var-const, line 115: You should move `result`, i.e. `return absl::make_optional(std::move(result).ValueOrDie());`.

Firestore/core/src/bundle/bundle_reader.cc
```diff
@@ -120,11 +121,11 @@ void BundleReader::ReadJsonToBuffer(size_t required_size) {
...
     auto size = std::min(1024ul, required_size - buffer_.size());
     StreamReadResult result = input_->Read(size);
-    if (!result.result().ok()) {
-      reader_status_.Update(result.result().status());
+    if (!result.ok()) {
+      reader_status_.Update(result.status());
       return;
     }
-    buffer_.append(result.result().ValueOrDie());
+    buffer_.append(result.ValueOrDie());
```
var-const, line 128: Can you move `result` here?

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -21,19 +21,93 @@ namespace firestore {
...
+        "{กว่าบรรดาฝูงสัตว์เดรัจฉาน จงฝ่าฟันพัฒนาวิชาการ อย่าล้า"
+        "งผลาญฤๅเข่นฆ่าบีฑาใคร ไม่ถือโทษโกรธแช่งซัดฮึดฮัดด่า "
+        "หัดอภัยเหมือนกีฬาอัชฌาสัย ปฏิบัติประพฤติกฎกำหนดใจ "
+        "พูดจาให้จ๊ะ ๆ จ๋า ๆ น่าฟังเอยฯ}";
+  }
+  return source;
+}
+
+uint64_t kMaxIteration = 10000 * 10000;
```
var-const, line 43: A couple of nits:
* s/kMaxIteration/kMaxIterations/;
* constants are usually declared before any functions, please move this up above `LargeString`.

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -21,19 +21,93 @@ namespace firestore {
...
+  auto result = stream->ReadUntil('o', 10);
+  EXPECT_EQ(result.ValueOrDie(), "");
+  EXPECT_TRUE(result.eof());
+
+  result = stream->Read(10);
+  EXPECT_EQ(result.ValueOrDie(), "");
+  EXPECT_TRUE(result.eof());
+}
+
+TEST_P(ByteStreamTest, ReadUtilReadsDelimString) {
```
var-const, line 73: Nit: s/Util/Until/ (here and below).

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -21,19 +21,93 @@ namespace firestore {
...
+  auto stream = stream_factory_->CreateByteStream("{{{{");
+
+  auto result = stream->ReadUntil('{', 10);
```
var-const, line 76: Can you call `ReadUntil('{', 10);` once more just to make sure it still doesn't read anything by mistake?

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -21,19 +21,93 @@ namespace firestore {
...
+
+TEST_P(ByteStreamTest, ReadUtilReadsStringWithoutDelim) {
+  auto stream = stream_factory_->CreateByteStream("{{{{");
+
+  auto result = stream->ReadUntil('}', 10);
+  EXPECT_EQ(result.ValueOrDie(), "{{{{");
+  EXPECT_TRUE(result.eof());
+}
+
+TEST_P(ByteStreamTest, ReadsNullCharacter) {
```
var-const, line 89: Cool test! Can you also try reading the whole null-containing string in one go? I.e., add a test that creates the same `"10{conten\0t}5{\0}"` string and then just calls `Read(100)` or something like that.

Firestore/core/test/unit/util/byte_stream_test.cc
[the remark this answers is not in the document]
var-const: Thanks for adding these!
> Passing `0` to `Read` and `ReadUntil`;
You added a great test for null characters, but what I meant here was calling `Read(0)` and `ReadUntil('}', 0)`. (to be clear, the test you added is very useful, thanks for doing that)

Some other test ideas:
* `Read(100)` where the actual stream is small (i.e., passing max length much greater than the length of the actual stream -- `100` is arbitrary);
* `Read(n)` where `n` is a) exactly equal to the string size; b) is equal to string size + 1; c) is equal to string size - 1.

var-const (changes requested):

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

Firestore/core/src/util/byte_stream_apple.mm
wu-hui: Done.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -111,6 +184,16 @@ TEST_P(ByteStreamTest, ReadsNullCharacter) {
   EXPECT_TRUE(result.eof());
 }
 
+TEST_P(ByteStreamTest, ReadsFullStringWithNullCharacter) {
+  // Using explicit string(char*, size) constructor to include \0
+  auto stream =
+      stream_factory_->CreateByteStream(std::string("10{conten\0t}5{\0}", 16));
```
var-const, line 190: Nit: please create a variable for this string and reuse it in the `EXPECT_EQ` below.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -42,45 +42,33 @@
...
   buffer_.erase(0, found_at);
-  auto read_result = StreamReadResult(std::move(result), eof());
+  StreamReadResult read_result(std::move(result), eof());
   return read_result;
 }
 
 StreamReadResult ByteStreamApple::Read(size_t max_length) {
   // Serve from buffer_
-  if (buffer_.size() >= max_length) {
+  if (buffer_.size() > max_length) {
```
var-const, line 59: Was this caught by tests?

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -95,8 +83,23 @@
   return static_cast<int32_t>(read);
 }
 
-StreamReadResult ByteStreamApple::EofResult() {
-  return StreamReadResult(std::move(buffer_), true);
+StreamReadResult ByteStreamApple::ConsumeBuffer() {
```
var-const, line 86: To be clear, I suggested passing the value `eof` as a boolean parameter -- I presume it would avoid the need to do the peeking. Is there a reason peeking is preferable?

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -95,8 +83,23 @@
...
+  // NSInputStream does not have consistent behavior with streamStatus, we
+  // perform a "peek" operation here to match with C++ istream implementation,
+  // and make the behavior deterministic.
+  std::string peek_result(1, '\0');
+  auto* data_ptr = reinterpret_cast<uint8_t*>(&peek_result[0]);
+  NSInteger read = [input_ read:data_ptr maxLength:1];
+
+  bool is_eof = (read == 0);
```
var-const, line 94: Optional: the parentheses aren't necessary (but fine to keep if you find they aid readability).

Firestore/core/src/util/byte_stream_cpp.cc
```diff
@@ -54,8 +54,12 @@ StreamReadResult ByteStreamCpp::ToReadResult(const std::string& result) {
     input_->clear();
   }
 
+  // Saving the read count from reading `result`, because we are doing another
+  // read (peek) to determine if it is EOF yet.
+  auto result_read_count = input_->gcount();
+  bool is_eof = input_->eof() || (input_->peek() == EOF);
```
var-const, line 60: Optional: the parentheses around the `EOF` comparison are not necessary (but fine to keep for readability).

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -70,15 +70,88 @@ TEST_P(ByteStreamTest, ReadsEmptyStringStream) {
...
+    auto stream = stream_factory_->CreateByteStream("0123456");
+    auto result = stream->ReadUntil('a', 8);
+    EXPECT_EQ(result.ValueOrDie(), "0123456");
+    EXPECT_TRUE(result.eof());
+  }
+}
+
+TEST_P(ByteStreamTest, ReadsZeroSizes) {
+  auto stream = stream_factory_->CreateByteStream("0123456");
+  auto result = stream->Read(0);
```
var-const, line 114: Nit: also try with `ReadUntil`?

Firestore/core/test/unit/util/byte_stream_test.cc
```diff
@@ -145,18 +228,39 @@ TEST_P(ByteStreamTest, ReadUntilReadsLargeStream) {
...
       break;
     }
   }
-  EXPECT_LE(i, kMaxIteration);
+  EXPECT_LE(i, kMaxIterations);
   EXPECT_EQ(actual, source);
 }
 
+// This is a test designed for the apple implementation's internal buffer usage.
+// It deliberately fill the internal buffer with ReadsUntil, then uses Read to
```
var-const, line 244: Ultranit: s/fill/fills/

var-const (changes requested):
Close to LGTM. Thanks for addressing the feedback so far!

Firestore/core/src/util/byte_stream_apple.mm
var-const: By the way, which test would fail otherwise?

Firestore/core/src/util/byte_stream_apple.mm
wu-hui: Yes: `ReadsFromInternalBuffer_AppleImpl`

Firestore/core/src/util/byte_stream_apple.mm
wu-hui: Yes. NSInputStream seems to return `eof:false` with `Read(0)` on `""`, while `istream` behaves otherwise.

Essentially, the behaviour btw the two are not consistent, explicit peeking makes the them consistent.

Firestore/core/src/util/byte_stream_apple.mm
wu-hui: Yep, I'd like to keep it.

Firestore/core/src/util/byte_stream_cpp.cc
wu-hui: Same as above.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done. This uncovered a bug, as it was assumed `ReadUntil` always start with empty `buffer_`.

This changed after the peek implementation, however.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

Firestore/core/test/unit/util/byte_stream_test.cc
wu-hui: Done.

var-const (approved):
LGTM with a question and a couple of nits.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    found_at = buffer_.find(delim, last_pos);
+    last_pos = buffer_.size();
+    if (found_at != std::string::npos || read == 0) {
+      break;
+    }
+  }
+
+  // Still not found, or the buffer happens to be of the sized as requested
```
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -0,0 +1,121 @@
...
+  // either way, return the whole `buffer_` and clear it.
+  if (found_at == std::string::npos && buffer_.size() == max_length) {
+    return ConsumeBuffer();
+  }
+
+  // Either we found the delim, or the buffer is already larger then requested,
```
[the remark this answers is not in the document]
wu-hui: Both peeking and ReadUntil can pull data into buffer_ without consuming all of them, which makes it possible the next request is serve entered from buffer_.

Yes, there are tests for this.

Firestore/core/src/util/byte_stream_apple.mm
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  return static_cast<int32_t>(read);
+}
+
+StreamReadResult ByteStreamApple::ConsumeBuffer() {
+  // NSInputStream does not have consistent behavior with streamStatus, we
+  // perform a "peek" operation here to match with C++ istream implementation,
+  // and make the behavior deterministic.
+  std::string peek_result(1, '\0');
```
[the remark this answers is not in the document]
wu-hui: Done.

wu-hui (commented):
Thank you!

PR #7570, opened 2021-02-19
Completely stop forcing static frameworks

paulb777:
Follow up fix for #6557 for Performance, MLModelDownloader, and Segmentation, that were not yet on master when the original 7.0.0 update was made.

Also fix exposed Catalyst build problem with mobile connectivity header. 

Fix #7569

### DIFF
--- a/FirebasePerformance.podspec
+++ b/FirebasePerformance.podspec
@@ -24,7 +24,6 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
   s.tvos.deployment_target = tvos_deployment_target
 
   s.cocoapods_version = '>= 1.4.0'
-  s.static_framework = true
   s.prefix_header_file = false
 
   base_dir = "FirebasePerformance/"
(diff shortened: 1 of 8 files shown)


ariel-demarco-peya (approved):

ryanwilson (approved):

FirebasePerformance.podspec
```diff
@@ -55,9 +54,9 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
     'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"',
   }
 
-  s.ios.framework = 'CoreTelephony'
-  s.ios.framework = 'QuartzCore'
-  s.ios.framework = 'SystemConfiguration'
+  s.framework = 'CoreTelephony'
```
visumickey, line 57: Does this framework exist for tvOS? If not, we might have to create a platform based framework inclusion.

paulb777: Good catch! Thanks.

visumickey (approved):

PR #7580, opened 2021-02-22
Combine Support: Disassociate user account from 3P IdP

ziadtamim:
Disassociates a user account from a third-party identity provider with this user.

```swift
authResult.user
  .unlink(fromProvider: PhoneAuthProviderID)
  .sink { completion in
    /* Handle completion */
  } receiveValue: { user in
    /* Handle user */
  }
  .store(in: &cancellables)
```


peterfriese (approved):
LGTM, thanks!

PR #7588, opened 2021-02-23
Treat bridging headers as test-related

peterfriese:
As discussed in https://github.com/firebase/firebase-ios-sdk/pull/7547#discussion_r580442313, this patch makes sure that bridging headers may use angle-bracket imports.

 #no-changelog


maksymmalyhin (approved):
👍

PR #7590, opened 2021-02-23
Publish command updates

paulb777:
- Now that they contain xcframeworks, Analytics pods no longer need to `--skip-import-validation`
- Fix the staging location path

#no-changelog


maksymmalyhin (approved):
👍

PR #7593, opened 2021-02-23
Automatically update SPM version variable

paulb777:
#no-changelog

### DIFF
--- a/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
+++ b/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
@@ -131,6 +131,7 @@ struct InitializeRelease {
   }
 
   private static func updateSwiftPackageVersion(path: URL, version: String) {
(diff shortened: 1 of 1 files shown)


ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
```diff
@@ -128,4 +129,9 @@ struct InitializeRelease {
       "'Firebase',[[:space:]]*'\\).*'#\\1\(version)'#\" Podfile"
     Shell.executeCommand(sedCommand2, workingDir: collisionPodfile)
   }
+
+  private static func updateSwiftPackageVersion(path: URL, version: String) {
+    Shell.executeCommand("sed -i.bak -e \"s/\\(let firebaseVersion.*=[[:space:]]*\\).*/\\1" +
```
ryanwilson, line 134: LGTM - if you don't mind can you add a comment above splitting out each section of the regex for future readers and those unfamiliar?

ryanwilson (approved):

PR #7600, opened 2021-02-24
Remove the warning when Analytics is not included

### LINKED ISSUE #7487: FirebasePerformance SDK warns that FirebaseAnalytics should be included

  * Xcode version: 12.4
  * Firebase SDK version: 7.5
  * Firebase Component: Performance, though I also have Crashlytics installed.

`Firebase.h` generates a warning when `FirebasePerformance` is included but `FirebaseAnalytics` isn't included in the project. This contradicts the SDK's documentation.

The table in [this section](https://firebase.google.com/docs/ios/setup#available-pods) does not list Analytics as required for the Performance SDK. Additionally my experience so far has been that the Performance SDK has been working just fine without Analytics included.

#### Steps to reproduce:

1. Include the Firebase Performance SDK in a project that does not include the Analytics SDK.
2. In case it matters, I installed the SDK using CocoaPods and also have the Crashlytics SDK included in the project.
3. Build the app.
4. Note the warning `"FirebaseAnalytics.framework is not included in your target. Please add the \
FirebaseAnalytics dependency to your project to ensure Firebase Performance works as intended."`

#### Relevant Code:

It seems like these lines are incorrect and should be removed, or the documentation in the SDK is incorrect.
```
  #if __has_include(<FirebasePerformance/FirebasePerformance.h>)
    #import <FirebasePerformance/FirebasePerformance.h>
    #if TARGET_OS_IOS && !__has_include(<FirebaseAnalytics/FirebaseAnalytics.h>)
      #ifndef FIREBASE_ANALYTICS_SUPPRESS_WARNING
        #warning "FirebaseAnalytics.framework is not included in your target. Please add the \
FirebaseAnalytics dependency to your project to ensure Firebase Performance works as intended."
      #endif // #ifndef FIREBASE_ANALYTICS_SUPPRESS_WARNING
    #endif
  #endif
```

visumickey:
Firebase Performance does not require inclusion of Analytics. Fixes #7487

### DIFF
--- a/CoreOnly/Sources/Firebase.h
+++ b/CoreOnly/Sources/Firebase.h
@@ -99,12 +99,6 @@ FirebaseAnalytics dependency to your project to ensure Messaging works as intend
 
   #if __has_include(<FirebasePerformance/FirebasePerformance.h>)
     #import <FirebasePerformance/FirebasePerformance.h>
-    #if TARGET_OS_IOS && !__has_include(<FirebaseAnalytics/FirebaseAnalytics.h>)
-      #ifndef FIREBASE_ANALYTICS_SUPPRESS_WARNING
-        #warning "FirebaseAnalytics.framework is not included in your target. Please add the \
-FirebaseAnalytics dependency to your project to ensure Firebase Performance works as intended."
-      #endif // #ifndef FIREBASE_ANALYTICS_SUPPRESS_WARNING
-    #endif
   #endif
 
   #if __has_include(<FirebaseRemoteConfig/FirebaseRemoteConfig.h>)
--- a/FirebasePerformance/CHANGELOG.md
+++ b/FirebasePerformance/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 * Deprecate Clearcut event transport mechanism.
 * Enable dynamic framework support. (#7569)
+* Remove the warning to include Firebase Analytics as Perf does not depend on Analytics (#7487)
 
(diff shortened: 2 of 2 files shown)


paulb777 (approved):
Please update changelog

visumickey:
> Please update changelog

Done.

jeremyjiang-dev (approved):

CoreOnly/Sources/Firebase.h
```diff
@@ -99,12 +99,6 @@ FirebaseAnalytics dependency to your project to ensure Messaging works as intend
 
   #if __has_include(<FirebasePerformance/FirebasePerformance.h>)
     #import <FirebasePerformance/FirebasePerformance.h>
```
ramanpreetSinghKhinda, line 101: Do we need to keep this? What is the purpose of this?

ramanpreetSinghKhinda (approved):

CoreOnly/Sources/Firebase.h
paulb777: It's needed. That allows clients to access Perf APIs by including Firebase.h or importing the Firebase module.

visumickey: @ramanpreetSinghKhinda These are Umbrella headers. If the including app includes Firebase.h, all the header files of other Firebase products that used by the app will be automatically imported. This is a huge convenience header for the developers. So this is definitely needed.

PR #7601, opened 2021-02-24
Move the dispatch of gauge metric calls to its dedicated thread asynchronously

### LINKED ISSUE #7535: Crash in FPRGaugeManager addGaugeData

  * Xcode version: 12.2
  * Firebase SDK version: 7.3.0
  * Firebase Component: Performance (Auth, Core, Database, Firestore, Messaging, Storage, etc)

#### Steps to reproduce:
Found crash with no known way to reproduce, about 40 crash for 30 days, so far found on iOS 13 and 14
All/most crash seems to happen some time after app returning to foreground

```
Crashed: com.google.perf.gaugeManager.gaugeData
0  libobjc.A.dylib                0x1974e5c90 objc_retain + 16
1  <appname>                      0x100e0e534 __32-[FPRGaugeManager addGaugeData:]_block_invoke + 4369573172
2  libdispatch.dylib              0x182d7a298 _dispatch_call_block_and_release + 24
3  libdispatch.dylib              0x182d7b280 _dispatch_client_callout + 16
4  libdispatch.dylib              0x182d23dcc _dispatch_lane_serial_drain$VARIANT$mp + 612
5  libdispatch.dylib              0x182d248a8 _dispatch_lane_invoke$VARIANT$mp + 424
6  libdispatch.dylib              0x182d2e338 _dispatch_workloop_worker_thread + 712
7  libsystem_pthread.dylib        0x1cb9695a4 _pthread_wqthread + 272
8  libsystem_pthread.dylib        0x1cb96c874 start_wqthread + 8
```

visumickey:
Fixes #7535


FirebasePerformance/Sources/Gauges/FPRGaugeManager.m
```diff
@@ -117,7 +117,9 @@ - (BOOL)gaugeCollectionEnabled {
 }
 
 - (void)startCollectingGauges:(FPRGauges)gauges forSessionId:(NSString *)sessionId {
-  [self prepareAndDispatchGaugeData];
+  dispatch_async(self.gaugeDataProtectionQueue, ^{
+    [self prepareAndDispatchGaugeData];
+  });
 
   self.currentSessionId = sessionId;
```
maksymmalyhin, line 124: Will old gauge data be associated with the new session ID if we set it now before batching? Is it fine if so?

maksymmalyhin (approved):
👍  LGTM, one question.

paulb777 (approved):
LGTM with resolution of @maksymmalyhin's question

FirebasePerformance/Sources/Gauges/FPRGaugeManager.m
visumickey: Very good question and a great catch. I have modified the code to handle that scenario better. Please take a look.

FirebasePerformance/Sources/Gauges/FPRGaugeManager.m
```diff
@@ -117,9 +117,8 @@ - (BOOL)gaugeCollectionEnabled {
 }
 
 - (void)startCollectingGauges:(FPRGauges)gauges forSessionId:(NSString *)sessionId {
-  dispatch_async(self.gaugeDataProtectionQueue, ^{
-    [self prepareAndDispatchGaugeData];
-  });
+  // Dispatch the already available gauge data with old sessionId.
+  [self prepareAndDispatchCollectedGaugeDataWithSessionId:self.currentSessionId];
```
maksymmalyhin, line 121: May it happen that the new session gauge data will be associated with the old session ID? Looks like not, because `addGaugeData` is also protected by the serial queue. But would like to check with you just in case.

maksymmalyhin (approved):
Thank you for the update! LGTM.

FirebasePerformance/Sources/Gauges/FPRGaugeManager.m
visumickey: That is a good question as well. I thought about as well. The ideal solution I had in my mind was to make the prepareAndDispatch method to take both the gauge data and the session ID as inputs (making it more self-contained). But that lead to a lot of repetitive code in reseting the gauge data unprotected. So took this approach which has the risk you mentioned, but since both addGaugeData and prepareAndDispatch happen on the same serial queue, that issue would not happen.

jeremyjiang-dev (approved):
LGTM! This PR has been a great learning experience for me!

PR #7617, opened 2021-02-26
FIRMessagingCheckinServiceTest: use OCMock to stub NSURLSession

maksymmalyhin:


FirebaseMessaging.podspec
```diff
@@ -36,10 +36,14 @@ device, and it is completely free.
     'Firebase/InstanceID/Public/*.h',
     'FirebaseInstallations/Source/Library/Private/*.h',
   ]
-  s.requires_arc = [
-    base_dir + 'Sources/*.m',
-    base_dir + 'Sources/Token/*.m'
-  ]
+
+  # TODO: Find out why `requires_arc` was required.
```
paulb777, line 40: I suspect, it's left over from the when the Protobuf dependency required some files to be built without ARC

maksymmalyhin: If so, looks like safe to be removed.

paulb777: Agreed

maksymmalyhin:
Other test failures seems to be inherited from the parent branch and not relevant to the PR changes.

FirebaseMessaging.podspec
```diff
@@ -36,10 +36,7 @@ device, and it is completely free.
     'Firebase/InstanceID/Public/*.h',
     'FirebaseInstallations/Source/Library/Private/*.h',
   ]
-  s.requires_arc = [
```
chliangGoogle, line 39: curious: this is not needed?

chliangGoogle (approved):
Awesome. thank you for doing this Maksym!

FirebaseMessaging/Tests/UnitTests/FIRMessagingCheckinServiceTest.m
```diff
@@ -47,23 +40,29 @@ @implementation FIRMessagingCheckinServiceTest
 
 - (void)setUp {
   [super setUp];
+
+  // Stub NSURLSession constructor before instantiating FIRMessagingCheckinService to inject
+  // URLSessionMock.
+  // TODO: inject NSURLSession instance via an initializer.
+  self.URLSessionMock = OCMClassMock([NSURLSession class]);
+  OCMStub(ClassMethod([self.URLSessionMock sessionWithConfiguration:[OCMArg any]]))
```
chliangGoogle, line 48: Nice! looks like I don't even have to expose the local session object.

FirebaseMessaging.podspec
maksymmalyhin: If the [reason mentioned by Paul](https://github.com/firebase/firebase-ios-sdk/pull/7617#discussion_r583771492) is the only one, than I think we don't need it any more.

PR #7619, opened 2021-02-26
Limit the number of unsent reports on disk to prevent disk filling

samedson:
If the developer turns automatic collection off, they run the risk of filling up disk space with unsent reports if they never call send/deleteUnsentReports. In addition, having too many unsent reports can cause performance problems when trying to find the newest unsent report.

This change limits the number of unsent reports to 4 to match Android: https://github.com/firebase/firebase-android-sdk/blob/e65f4876ff61b5791986c249b21cb094c37bd29e/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/SettingsJsonConstants.java#L61

### DIFF
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
@@ -29,14 +29,16 @@ const NSUInteger MAX_UNSENT_REPORTS = 4;
 @interface FIRCLSExistingReportManager ()
 
 @property(nonatomic, strong) FIRCLSFileManager *fileManager;
-@property(nonatomic, strong) NSOperationQueue *operationQueue;
 @property(nonatomic, strong) FIRCLSReportUploader *reportUploader;
+@property(nonatomic, strong) NSOperationQueue *operationQueue;
 
 // This list of active reports excludes the brand new active report that will be created this run of
 // the app.
 @property(nonatomic, strong) NSArray *existingUnemptyActiveReportPaths;
 @property(nonatomic, strong) NSArray *processingReportPaths;
 @property(nonatomic, strong) NSArray *preparedReportPaths;
+
+@property(nonatomic, strong) FIRCLSInternalReport *newestInternalReport;
 
 @end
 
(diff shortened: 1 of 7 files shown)


Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
```diff
@@ -17,23 +17,29 @@
...
 #import "Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.h"
 #import "Crashlytics/Crashlytics/DataCollection/FIRCLSDataCollectionToken.h"
+#import "Crashlytics/Crashlytics/Helpers/FIRCLSLogger.h"
 #import "Crashlytics/Crashlytics/Models/FIRCLSFileManager.h"
 #import "Crashlytics/Crashlytics/Models/FIRCLSInternalReport.h"
 #import "Crashlytics/Crashlytics/Private/FIRCrashlyticsReport_Private.h"
 #import "Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlyticsReport.h"
 
+// This value should stay in sync with the Android SDK
+const NSUInteger MAX_UNSENT_REPORTS = 4;
```
adamvduke, line 27: Should this be extern'ed? It's mentioned in the header, but doesn't seem to be defined there.

Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
```diff
@@ -52,15 +58,15 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData
   return self;
 }
 
-NSInteger compareOlder(FIRCLSInternalReport *reportA,
+NSInteger compareNewer(FIRCLSInternalReport *reportA,
```
adamvduke, line 61: Should there be a comment describing which report is newer, and is the type NSComparisonResult more descriptive?

Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
```diff
@@ -52,15 +58,15 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData
...
                        FIRCLSInternalReport *reportB,
                        void *context) {
-  return [reportA.dateCreated compare:reportB.dateCreated];
+  return -1 * [reportA.dateCreated compare:reportB.dateCreated];
```
adamvduke, line 64: Should there be a comment here noting that you are inverting the standard comparison result, or possibly swap A and B instead of multiplying by `-1`?

Crashlytics/UnitTests/FIRCLSExistingReportManagerTests.m
```diff
@@ -0,0 +1,214 @@
...
+  [self createActiveReportWithID:@"report_B" time:12315 withEvents:YES];
+  [self createActiveReportWithID:@"report_C" time:31533 withEvents:YES];
+  [self createActiveReportWithID:@"report_D" time:63263 withEvents:YES];
+
+  [self.existingReportManager collectExistingReports];
+
+  [self.existingReportManager.operationQueue waitUntilAllOperationsAreFinished];
+
+  // Reports with events should be kept if there's less than MAX_UNSENT_REPORTS reports
+  XCTAssertEqual([[self contentsOfActivePath] count], 4);
```
adamvduke, line 176: If `MAX_UNSENT_REPORTS` is extern'ed can we reference it here and below to clarify why the expected value is 4?

Crashlytics/UnitTests/FIRCLSExistingReportManagerTests.m
```diff
@@ -0,0 +1,214 @@
...
+  XCTAssertEqual(self.existingReportManager.unsentReportsCount, 4);
+  XCTAssertEqual(self.existingReportManager.existingUnemptyActiveReportPaths.count, 4);
+
+  // Newest report based on started_at timestamp
+  XCTAssertEqualObjects(self.existingReportManager.newestUnsentReport.reportID, @"report_D");
+}
+
+- (void)testUnsentReportsOverLimit {
```
adamvduke, line 184: Should there be a comment here describing the expected results? A quick scan of the setup code doesn't immediately reveal the expected behavior until you realize that the tests are created with various timestamps.

Crashlytics/Crashlytics/Controllers/FIRCLSExistingReportManager.m
samedson: Ooh good call

adamvduke (approved):

PR #7623, opened 2021-02-27
WIP: Add Firestore support to FirebaseCombine

lorenzofiamingo:
# :construction: Progress
## :sparkles: Implementation
**Collection Reference**
- [x] addDocument

**Document Reference**
- [x] setData
- [x] updateData
- [x] delete
- [x] getDocument
- [x] snapshotPublisher

**Query**
- [x] getDocuments
- [x] snapshotPublisher

**Transaction**
- [x] runTransaction

**WriteBatch**
- [x] commit

## :memo: Documentation
**Collection Reference**
- [x] addDocument

**Document Reference**
- [x] setData
- [x] updateData
- [x] delete
- [x] getDocument
- [x] snapshotPublisher

**Query**
- [x] getDocuments
- [x] snapshotPublisher

**Transaction**
- [x] runTransaction

**WriteBatch**
- [x] commit

## :white_check_mark: Test
**Collection Reference**
- [ ] addDocument

**Document Reference**
- [ ] setData
- [ ] updateData
- [ ] delete
- [ ] getDocument
- [ ] snapshotPublisher

**Query**
- [x] getDocuments
- [ ] snapshotPublisher

**Transaction**
- [ ] runTransaction

**WriteBatch**
- [ ] commit

PR #7632, opened 2021-03-01
WIP: making Firestore / Combine testable

peterfriese:
Hi all,

This is an attempt at making the Firebase / Combine code testable. It is baes on @lorenzofiamingo's PR #7623 - see his original comment about not being able to mock Firestore classes: https://github.com/firebase/firebase-ios-sdk/pull/7623#discussion_r584122727

The main issue (yet again) that the initialisers for many classes are hidden, and only accessible internally. I've tried to add the internal headers to the bridging header, but failed. If someone else would like to have a go at it, please go ahead - I'm no expert in cross-language integrations, so might have missed something when setting up the Podfile.

The current attempt is to use protocols and protocol extensions to make it easier to replace the implementation with a mock. I've implemented this according to @maksymmalyhin's PR #7557. This approach seems to work quite well - except that we need to create dummy results for some of the methods, and (you guessed it) that's not possible because the initialisers are hidden. There is a utility file that should help (`Firestore/Example/Tests/API/FSTAPIHelpers.mm`), but when trying to import it via the bridging header, I get a bunch of error messages as well.

If we cannot mock the Firestore base classes, we'll have to resort to using a similar approach to the original Firestore tests:
1. write integration tests that make use of the Emulator suite
2. use "compile"-tests that verify that the API signatures look like we expect them

Any help/suggestions appreciated.

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/Firebase-Package.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/Firebase-Package.xcscheme
@@ -734,6 +734,118 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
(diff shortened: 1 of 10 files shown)


peterfriese:
Closing this in favour of #7660.

PR #7642, opened 2021-03-02
Add FieldValue helpers

schmidt-sebastian:
This PR adds equal, compareTo and canonicalId helpers that work directly on Protobuf field values.

This is a port of

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/test/java/com/google/firebase/firestore/model/ValuesTest.java

and of  some of the code that we need from https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/ServerTimestamps.java


Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+  Add(equals_group, Wrap(1.0), Wrap(1.0));
+  Add(equals_group, Wrap(1.1), Wrap(1.1));
+  Add(equals_group, Wrap(BlobValue(0, 1, 1)));
+  Add(equals_group, Wrap(BlobValue(0, 1)));
+  Add(equals_group, Wrap("string"), Wrap("string"));
+  Add(equals_group, Wrap("strin"));
+  // latin small letter e + combining acute accent
+  Add(equals_group, Wrap("e\u0301b"));
+  // latin small letter e with acute accent
+  Add(equals_group, Wrap("\u00e9a"));
```
wu-hui, line 164: Other possible case: string with '\0' in the middle of it.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+  Add(equals_group, WrapArray("foo", "bar"), WrapArray("foo", "bar"));
+  Add(equals_group, WrapArray("foo", "bar", "baz"));
+  Add(equals_group, WrapArray("foo"));
+  Add(equals_group, WrapObject("bar", 1, "foo", 2),
+      WrapObject("foo", 2, "bar", 1));
+  Add(equals_group, WrapObject("bar", 2, "foo", 1));
+  Add(equals_group, WrapObject("bar", 1));
+  Add(equals_group, WrapObject("foo", 1));
+
+  for (size_t i = 0; i < equals_group.size(); ++i) {
```
wu-hui, line 188: It took me a while to understand why the tests are written like this..maybe some comment will help to explain what is happening, something like:

Each element of `equals_group` is a vector of values are should equal to each other. Any two values that do no belong in the same `equals_group` element should not equal to each other.

wu-hui: Also consider break the equal/NotEqual check in two separate loop, more readable IMO.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+  Add(comparison_groups, WrapArray("foo", "0"));
+
+  // objects
+  Add(comparison_groups, WrapObject("bar", 0));
+  Add(comparison_groups, WrapObject("bar", 0, "foo", 1));
+  Add(comparison_groups, WrapObject("foo", 1));
+  Add(comparison_groups, WrapObject("foo", 2));
+  Add(comparison_groups, WrapObject("foo", "0"));
+
+  for (size_t i = 0; i < comparison_groups.size(); ++i) {
```
wu-hui, line 286: Again, comment about the test data is ordered, and generally how the tests work should help.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+}
+
+TEST_F(ValuesTest, CanonicalId) {
+  VerifyCanonicalId(Wrap(nullptr), "null");
+  VerifyCanonicalId(Wrap(true), "true");
+  VerifyCanonicalId(Wrap(false), "false");
+  VerifyCanonicalId(Wrap(1), "1");
+  VerifyCanonicalId(Wrap(1.0), "1.000000");
+  VerifyCanonicalId(Wrap(Timestamp(30, 1000)), "time(30,1000)");
+  VerifyCanonicalId(Wrap("a"), "a");
```
wu-hui, line 306: "a\0b" should also work here.

Note you can create above with `std::string("a\0b", 3)`, spelling the length `3` is required.

wu-hui (changes requested):
Mostly around how strings are handled.

Firestore/core/src/model/values.cc
```diff
@@ -0,0 +1,484 @@
...
+
+  std::string result = "{";
+  bool first = true;
+  for (const auto& entry : sorted_keys_to_index) {
+    if (!first) {
+      result += ",";
+    } else {
+      first = false;
+    }
+    result += entry.first + ":" +
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/src/model/values.h
```diff
@@ -0,0 +1,113 @@
...
+enum class ComparisonResult;
+}
+
+namespace model {
+
+/**
+ * The order of types in Firestore. This order is based on the backend's
+ * ordering, but modified to support server timestamps.
+ */
+const int32_t TYPE_ORDER_NULL = 0;
```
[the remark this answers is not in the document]
schmidt-sebastian: Fun. Done.

Firestore/core/src/util/comparison.h
```diff
@@ -72,6 +72,18 @@ constexpr bool Descending(ComparisonResult result) noexcept {
   return result == ComparisonResult::Descending;
 }
 
+/**
```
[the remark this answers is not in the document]
schmidt-sebastian: I am using it from the new values.cc file. I could also just expose it as a header but it might make sense to inline it.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+
+namespace {
+
+double ToDouble(uint64_t value) {
+  return absl::bit_cast<double>(value);
+}
+
+const uint64_t kNanBits = 0x7fff000000000000ULL;
+
+}  // namespace
```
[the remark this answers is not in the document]
schmidt-sebastian: Looking forward to the link :) Fixed

Firestore/core/test/unit/model/values_test.cc
schmidt-sebastian: I added comments and rewrote the loop a bit. Do you think it is more understandable now?

Firestore/core/test/unit/model/values_test.cc
schmidt-sebastian: Addressed (fingers crossed)

schmidt-sebastian (commented):
Thanks for the review! I assigning it to Costa as well for some final thoughts.

Firestore/core/src/model/values.cc
```diff
@@ -0,0 +1,477 @@
...
+
+  // Create a map of field names to index for one of the maps. This is then used
+  // look up the corresponding value for the other map's fields.
+  std::unordered_map<std::string, size_t> key_to_value_index;
+  for (size_t i = 0; i < left_map.fields_count; ++i) {
+    key_to_value_index.emplace(nanopb::MakeString(left_map.fields[i].key), i);
+  }
+
+  for (size_t i = 0; i < right_map.fields_count; ++i) {
+    const std::string& key = nanopb::MakeString(right_map.fields[i].key);
```
wu-hui, line 175: Should also be StringView because it is only used to look up.

Firestore/core/test/unit/model/values_test.cc
wu-hui: Yep!

wu-hui (changes requested):
Almost LGTM..just one more place to use `MakeStringView`, and some more tests around strings containing `\0`.

Firestore/core/test/unit/model/values_test.cc
schmidt-sebastian: There should already be done in the line below (added in that review cycle). I added additional ones for comparison + canoncial ID now.

Firestore/core/src/model/values.cc
schmidt-sebastian: This doesn't seem to compile since I am looking up a std::string in the map. I tried changing the map to absl::string_view, but it did what I expected it to do and failed to compile since there is no hash function. I also think that using a string_view as a map key might be a bit of an antipattern but you or Costa can probably explain better.

```
std::unordered_map<absl::string_view, size_t, absl::Hash<absl::string_view>> key_to_value_index;


Undefined symbols for architecture x86_64:
  "absl::lts_2020_02_25::hash_internal::CityHash64(char const*, unsigned long)", referenced from:
      absl::lts_2020_02_25::hash_internal::CityHashState::CombineContiguousImpl(unsigned long long, unsigned char const*, unsigned long, std::__1::integral_constant<int, 8>) in libfirestore_core.a(values.cc.o)
  "absl::lts_2020_02_25::hash_internal::CityHashState::CombineLargeContiguousImpl64(unsigned long long, unsigned char const*, unsigned long)", referenced from:
      absl::lts_2020_02_25::hash_internal::CityHashState::CombineContiguousImpl(unsigned long long, unsigned char const*, unsigned long, std::__1::integral_constant<int, 8>) in libfirestore_core.a(values.cc.o)
  "absl::lts_2020_02_25::hash_internal::CityHashState::kSeed", referenced from:
      absl::lts_2020_02_25::hash_internal::CityHashState::CityHashState() in libfirestore_core.a(values.cc.o)
```

wu-hui: Ah, my bad, I think you are right. We definitely need someone to review my review :-)

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,331 @@
...
+
+  template <typename T>
+  google_firestore_v1_Value Wrap(T input) {
+    model::FieldValue fv = Value(input);
+    return serializer.EncodeFieldValue(fv);
+  }
+
+  template <typename... Args>
+  google_firestore_v1_Value WrapObject(Args... key_value_pairs) {
+    FieldValue fv = testutil::WrapObject((key_value_pairs)...);
```
var-const, line 68: Nits:
* the extra parentheses around `key_value_pairs` are unnecessary;
* it's not very important because it's a test file, but ideally you would make sure to forward the arguments to `WrapObject` to avoid extra copies (though `WrapObject`'s implementation would then proceed to create those copies regardless -- ideally it should be fixed, but that's out of scope for this PR):
```cpp
  template <typename... Args>
  // Make `Args` a forwarding reference
  google_firestore_v1_Value WrapObject(Args&&... key_value_pairs) {
    FieldValue fv = testutil::WrapObject(std::forward<Args>(key_value_pairs)...);
```
Feel free to disregard the point re. forwarding.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,331 @@
...
+            << "Reverse order check failed for '" << Values::CanonicalId(val1)
+            << "' and '" << Values::CanonicalId(val2) << "' (expected "
+            << static_cast<int>(util::ReverseOrder(expected_result)) << ")";
+      }
+    }
+  }
+
+  void VerifyCanonicalId(const google_firestore_v1_Value& value,
+                         const std::string& expected_canonical_id) {
+    const std::string& actual_canonical_id = Values::CanonicalId(value);
```
var-const, line 135: Does this const reference bind to a temporary? If yes, please replace it with a value.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,331 @@
...
+
+  google_firestore_v1_Value WrapServerTimestamp(
+      const model::FieldValue& input) {
+    // TODO(mrschmidt): Replace with EncodeFieldValue encoding when available
+    return WrapObject("__type__", "server_timestamp", "__local_write_time__",
+                      input.server_timestamp_value().local_write_time());
+  }
+
+  template <typename... Args>
+  void Add(std::vector<std::vector<google_firestore_v1_Value>>& groups,
```
var-const, line 96: Nice!

var-const (changes requested):

Firestore/core/src/model/values.cc
var-const: Are you trying to use `flat_hash_map` (judging by `container_internal::FlatHashMapPolicy`)? Can you try adding `absl::flat_hash_map` to the dependencies?

var-const (changes requested):
Thanks for implementing the feedback!

Firestore/core/src/model/values.cc
schmidt-sebastian: I added both absl_hash and absl_container to the dependencies (which should pull in the hash map), but it still does not seem to be working: https://gist.github.com/schmidt-sebastian/6145cab8862e4a72ce7e949fa4b7ef92

Firestore/core/src/model/server_timestamps.h
```diff
@@ -23,27 +23,20 @@ namespace firebase {
 namespace firestore {
 namespace model {
 
+// Utility methods to handle ServerTimestamps, which are stored using special
+// sentinal fields in MapValues.
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/src/model/value.h
```diff
@@ -0,0 +1,75 @@
...
+enum class ComparisonResult;
+}
+
+namespace model {
+
+/**
+ * The order of types in Firestore. This order is based on the backend's
+ * ordering, but modified to support server timestamps.
+ */
+enum TypeOrder {
```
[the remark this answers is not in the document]
schmidt-sebastian: That matches my understanding of how enums should work. Updated.

Firestore/core/src/model/value.h
```diff
@@ -0,0 +1,75 @@
...
+  kTypeOrderNull = 0,
```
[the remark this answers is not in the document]
schmidt-sebastian: The values here are actually super important, as they don't just distinguish the different types but also define their ordering. For this reason, I would like to explicit.

Firestore/core/src/model/values.h
```diff
@@ -0,0 +1,113 @@
...
+const int32_t kTypeOrderTimestamp = 3;
+const int32_t kTypeOrderServerTimestamp = 4;
+const int32_t kTypeOrderString = 5;
+const int32_t kTypeOrderBlob = 6;
+const int32_t kTypeOrderReference = 7;
+const int32_t kTypeOrderGeoPoint = 8;
+const int32_t kTypeOrderArray = 9;
+const int32_t kTypeOrderMap = 10;
+
+class Values {
```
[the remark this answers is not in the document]
schmidt-sebastian: This file contains equals, compareTo and canonical ID support. That's all we have on Web and Android, but in general we treated this file as a "proto values utility" class on the other platforms.

Firestore/core/test/unit/model/values_test.cc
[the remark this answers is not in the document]
schmidt-sebastian: I am always interested in learning :) I hope I will remember this time.

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+
+  template <typename... Args>
+  google_firestore_v1_Value WrapObject(Args... key_value_pairs) {
+    FieldValue fv = testutil::WrapObject((key_value_pairs)...);
+    return serializer.EncodeFieldValue(fv);
+  }
+
+  template <typename... Args>
+  google_firestore_v1_Value WrapArray(Args... values) {
+    std::vector<model::FieldValue> contents{Value(values)...};
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/test/unit/model/values_test.cc
```diff
@@ -0,0 +1,323 @@
...
+      const model::FieldValue& input) {
+    // TODO(mrschmidt): Replace with EncodeFieldValue encoding when available
+    return WrapObject("__type__", "server_timestamp", "__local_write_time__",
+                      input.server_timestamp_value().local_write_time());
+  }
+
+  template <typename... Args>
+  void Add(std::vector<std::vector<google_firestore_v1_Value>>& groups,
+           Args... values) {
+    std::vector<google_firestore_v1_Value> group{(values)...};
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/values.h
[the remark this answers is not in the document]
var-const: Nit: a file named `value` seems to indicate that it defines a `Value` class. I don't have a better suggestion than the uninspired `value_helpers`, but personally, I feel it would convey the intention better (same for the server timestamp one). If you can think of a better name, that's great.

var-const (approved):

PR #7656, opened 2021-03-04
watchOS support for RTDB

ryanwilson:
The happy-path is tested here, and leaving the app and returning works. Likely at a point it can be merged and tested in real world applications. 

Note: this implementation could be used to get rid of socket rocket eventually, and potentially unblock App Clips as well.


FirebaseDatabase.podspec
```diff
@@ -16,9 +16,16 @@ Simplify your iOS development, grow your user base, and monetize more effectivel
...
   }
   s.social_media_url = 'https://twitter.com/Firebase'
-  s.ios.deployment_target = '10.0'
-  s.osx.deployment_target = '10.12'
-  s.tvos.deployment_target = '10.0'
+
+  ios_deployment_target = '10.0'
+  osx_deployment_target = '10.12'
+  tvos_deployment_target = '10.0'
+  watchos_deployment_target = '7.0'
```
ryanwilson, line 23: Note: for lifecycle changes, watchOS 7 is preferred but the SwiftPM manifest currently states 6. We need to bump it at the next major change.

ryanwilson:
levelDB needs an update as well: https://github.com/firebase/leveldb-library-podspec/pull/11

FirebaseDatabase/Tests/Unit/FUtilitiesTest.m
```diff
@@ -115,7 +115,7 @@ - (void)testUserAgentString {
   XCTAssertEqualObjects(parts[0], @"Firebase");
   XCTAssertEqualObjects(parts[1], kWebsocketProtocolVersion);   // Wire protocol version
   XCTAssertEqualObjects(parts[2], [FIRDatabase buildVersion]);  // Build version
-#if TARGET_OS_IPHONE
+#if TARGET_OS_IPHONE && !TARGET_OS_WATCH
```
maksymmalyhin, line 118: I'm curious if the tests can be run on watchOS with appropriate Xcode version?

maksymmalyhin (approved):
👍 Great to see watchOS support, thank you!

Mostly LGTM, a couple minor comments/questions.

Also it would be nice to have a note/readme/etc. with hints on the development process for watchOS like what application can be used for testing, how to run tests (if possible), etc.

FirebaseDatabase/Sources/Core/FPersistentConnection.m
```diff
@@ -1234,6 +1256,10 @@ - (void)sendConnectStats {
     if (self.config.persistenceEnabled) {
         stats[@"persistence.osx.enabled"] = @1;
     }
+#elif TARGET_OS_WATCH || 1
```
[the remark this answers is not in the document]
ryanwilson: Whoops, yep that was accidentally left for seeing if iOS worked using the new networking code. I'll remove all them, I thought I did but maybe I didn't stage them before committing. Thanks!

FirebaseDatabase/Tests/Unit/FUtilitiesTest.m
ryanwilson: Yes they can, I tested it locally and they all passed (this change was required to make them pass). Once we have it available on GHA I can add it!

maksymmalyhin: Great!

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -255,7 +346,59 @@ - (void)handleIncomingFrame:(NSString *)message {
...
 #pragma mark -
+#pragma mark URLSessionWebSocketDelegate watchOS implementation
+#if TARGET_OS_WATCH
+
+- (void)URLSession:(NSURLSession *)session
+          webSocketTask:(NSURLSessionWebSocketTask *)webSocketTask
+    didOpenWithProtocol:(NSString *)protocol {
+    FFLog(@"I-RDB083008", @"(wsc:%@) webSocketDidOpen", self.connectionId);
+
+    everConnected = YES;
```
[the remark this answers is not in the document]
ryanwilson: Good eye! Moved.

FirebaseDatabase/Sources/Persistence/FLevelDBStorageEngine.m
```diff
@@ -267,7 +267,7 @@ - (void)close {
 }
 
 + (NSString *)firebaseDir {
-#if TARGET_OS_IOS || TARGET_OS_TV
+#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH || 1
```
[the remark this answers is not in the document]
ryanwilson: Added for debugging, nice catch.

FirebaseDatabase.podspec
paulb777: Please open a Firebase 8 issue to track

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -179,21 +230,61 @@ - (void)send:(NSDictionary *)dictionary {
...
     }
 }
 
 - (void)nop:(NSTimer *)timer {
     if (!isClosed) {
         FFLog(@"I-RDB083004", @"(wsc:%@) nop", self.connectionId);
+#if TARGET_OS_WATCH
+        // Note: There's a built in `sendPingWithPongReceiveHandler` call that's
+        // available, but it doesn't match with what the backend is expecting.
+        // Send a string with "0" instead.
```
paulb777, line 275: Looks like this comment applies to the non-watch code also?

ryanwilson: The `sendPingWithPongReceiveHandler` is a `NSURLSessionWebSocketTask` API that's not available with SocketRocket. The "0" being expected from the backend is the same, but I wanted to clarify that this isn't a standard ping/pong that we can use Foundation APIs for.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -179,21 +230,61 @@ - (void)send:(NSDictionary *)dictionary {
     // First send the header so the server knows how many segments are
     // forthcoming
     if (dataSegs.count > 1) {
-        [self.webSocket
-            send:[NSString
-                     stringWithFormat:@"%u", (unsigned int)dataSegs.count]];
+        NSString *formattedData =
+            [NSString stringWithFormat:@"%u", (unsigned int)dataSegs.count];
+#if TARGET_OS_WATCH
```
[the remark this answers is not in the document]
ryanwilson: Done, thanks.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -251,15 +252,15 @@ - (void)send:(NSDictionary *)dictionary {
...
+                         completion:^(NSError *_Nullable error) {
+                           if (error) {
+                               FFWarn(@"I-RDB083018",
+                                      @"Error sending web socket data: %@",
+                                      error);
+                               return;
+                           }
+                         }];
 #else
         [self.webSocket send:segment];
```
maksymmalyhin, line 265: Optional: we probably could have a method like `sendStringToWebSocket:` where we send the string to one of the socket implementations, so the method usage is simpler. But I'm not too strong about it if having different log messages on the error if critical.

maksymmalyhin (approved):
Thank you for the updates! LGTM, one optional comment.

maksymmalyhin (approved):

paulb777 (approved):

FirebaseDatabase.podspec
ryanwilson: Done in https://github.com/firebase/firebase-ios-sdk/issues/7663. Alternatively I could mark up  the API with `@available(watchOS 7, *)` and that may be a better solution.  WDYT?

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: I do like having the log messages there for logging purposes, especially given that I'm curious to see any failures for watchOS in the wild and what the issues are. Definitely something to consider in the future though, thank you!

ryanwilson:
> levelDB needs an update as well: [firebase/leveldb-library-podspec#11](https://github.com/firebase/leveldb-library-podspec/pull/11)

This has been published.

FirebaseDatabase/Sources/Persistence/FLevelDBStorageEngine.m
```diff
@@ -267,11 +267,16 @@ - (void)close {
...
 + (NSString *)firebaseDir {
-#if TARGET_OS_IOS || TARGET_OS_TV
+#if TARGET_OS_IOS || TARGET_OS_WATCH
     NSArray *dirPaths = NSSearchPathForDirectoriesInDomains(
         NSDocumentDirectory, NSUserDomainMask, YES);
     NSString *documentsDir = [dirPaths objectAtIndex:0];
     return [documentsDir stringByAppendingPathComponent:@"firebase"];
+#elif TARGET_OS_TV
+    NSArray *dirPaths = NSSearchPathForDirectoriesInDomains(
+        NSCachesDirectory, NSUserDomainMask, YES);
```
schmidt-sebastian, line 277: Does this change drop the existing cache for Apple TV users?

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -79,12 +90,39 @@ - (id)initWith:(FRepoInfo *)repoInfo
 
         NSURLRequest *req = [[NSURLRequest alloc]
             initWithURL:[[NSURL alloc] initWithString:connectionUrl]];
+#if TARGET_OS_WATCH
```
schmidt-sebastian, line 93: Did you take a look at what it would take to have a Watch-specific implementation of FWebSocketConnection? I am a bit worried about maintaining sane code coverage numbers for this file. 

Another suggestion would be to create an abstraction on top of `NSURLSessionWebSocketTask` and `FSRWebSocket`. That might be more straightforward and would cut most changes to this file.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -255,26 +346,52 @@ - (void)handleIncomingFrame:(NSString *)message {
...
-      self->keepAlive =
-          [NSTimer scheduledTimerWithTimeInterval:kWebsocketKeepaliveInterval
-                                           target:self
-                                         selector:@selector(nop:)
-                                         userInfo:nil
-                                          repeats:YES];
-      FFLog(@"I-RDB083009", @"(wsc:%@) nop timer kicked off",
-            self.connectionId);
-    });
+    [self webSocketOpened];
```
schmidt-sebastian, line 394: We own FSRWebSocketDelegate and could just remove the argument here, which would remove the need for this indirection.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -179,21 +230,61 @@ - (void)send:(NSDictionary *)dictionary {
...
+        NSString *formattedData =
+            [NSString stringWithFormat:@"%u", (unsigned int)dataSegs.count];
+#if TARGET_OS_WATCH
+        [self
+            sendStringToWebSocket:formattedData
+                       completion:^(NSError *_Nullable error) {
+                         if (error) {
+                             FFWarn(
+                                 @"I-RDB083017",
+                                 @"Error sending initial web socket data: %@.",
```
schmidt-sebastian, line 242: Do you think we need these different error messages (especially since we print the data). If not, we could move the logging to `sendStringToWebSocket` as well.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -292,9 +409,42 @@ - (void)webSocket:(FSRWebSocket *)webSocket
...
 }
 
+#endif // TARGET_OS_WATCH
+
 #pragma mark -
 #pragma mark Private methods
 
+#if TARGET_OS_WATCH
+
+/// Sends data to the web socket through native APIs.
```
schmidt-sebastian, line 419: The rest of this code base generally seems to use JavaDoc for method comments.

schmidt-sebastian (commented):
Some comments, feel free to push back on all of them. And sorry for the delay!

FirebaseDatabase/Sources/Persistence/FLevelDBStorageEngine.m
ryanwilson: It actually is broken as is - the `NSDocumentsDirectory` is not writable on tvOS devices. Unfortunately it's writeable in simulators so we didn't catch it before. This should work as expected now though!

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: I did briefly and thought this was a reasonable tradeoff. Options + reasoning below...

1. Selecting files to include/exclude based on package manager.  This doesn't really work well, and would be overly complex.
2. Using separate files to contain each implementation and use `#ifdef` around the whole file. I was worried here about having implementation files with different naming and issues that may come up with it. Also with code duplication.
3. Abstraction on top. Thought about this but didn't seem like it was worth the effort, plus eventually we would want this implementation to take over for SocketRocket so it'd leave more work in the future to revert it.

Re: code coverage, the only reference to this class in tests is https://github.com/firebase/firebase-ios-sdk/blob/14764b8d60a6ad023d8fa5b7f81d42378d92e6fe/FirebaseDatabase/Tests/Unit/FUtilitiesTest.m#L109 so I don't think it's overly tested anyways (for better or for worse). 

The last time this file was modified outside of adding tvOS support or fixing imports (unrelated to functionality) was in June 2017 https://github.com/firebase/firebase-ios-sdk/commit/9917869191ecc00f6ac060fabe3cd4470bebbfab#diff-50e445e5405c9e69557c81053b2ada15120eda819c93ca059d14facffa892951 so I wasn't overly worried about it being modified often and becoming harder to maintain.

If I were to do anything, it would be create an abstraction between the 2 but I'd also worry about modifying the iOS code since it isn't well tested and hasn't been changed in a very long time. Let me know what you think.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: I had it in there for debugging and I'm not sure what to expect when getting real world reports so thought it could be helpful for differentiation since I wasn't sure about the breadth of errors we may get. Happy to consolidate if you feel strongly about it though.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: Will change, thanks

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
schmidt-sebastian: Sounds fair. It didn't occur to me that including the source and not using it would still be challenging.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
schmidt-sebastian: I think we should able to tell the callsites apart just based on the data, and pulling the logging into `sendStringToWebSocket` would remove a couple of redundant lines in this (small) PR.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: Fixed in latest commit.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
ryanwilson: Fixed in latest commit.

ryanwilson:
@schmidt-sebastian thanks for the review, updated based on your comments! PTAL

schmidt-sebastian (approved):

doplumi:
@ryanwilson can't seem to be able to add 'Firebase/Core' and 'Firebase/Database' to a watchOS app yet

```
$ pod install
Analyzing dependencies
[!] The platform of the target `Test WatchKit Extension` (watchOS 7.0) is not compatible with `Firebase/Database (7.9.0)`, which does not support `watchOS`.
```

paulb777:
@doplumi Thanks for the report.  It looks like we didn't make the corresponding update to the Firebase pod.  I'll see if I can get that done for 7.10.0.

In the meantime, a workaround should be to do `pod 'FirebaseDatabase'` (without the slash).

`Firebase/Core` is deprecated in favor of `Firebase/Analytics` and is not supported for watchos.

paulb777:
A fix is underway at #7841 but looks too late for 7.10.0 and will instead is target for 7.11.0

PR #7660, opened 2021-03-05
Open up initialisers to make Firestore more testable

peterfriese:
As discussed, this PR opens up a few initialisers in Firestore to make it more testable from Swift.

It is based on @lorenzofiamingo's PR #7623 - see his original comment about not being able to mock Firestore classes: https://github.com/firebase/firebase-ios-sdk/pull/7623#discussion_r584122727


Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
```diff
@@ -33,8 +33,8 @@ NS_ASSUME_NONNULL_BEGIN
 NS_SWIFT_NAME(Timestamp)
 @interface FIRTimestamp : NSObject <NSCopying>
 
-/** :nodoc: */
-- (instancetype)init NS_UNAVAILABLE;
+/** Initialiser. Only for use in unit tests. */
+- (instancetype)init;
```
maksymmalyhin, line 37: I think we should just make the designated initializer public here:
https://github.com/firebase/firebase-ios-sdk/blob/627ebdcd4575d8f4ca7b7942fe183a54ec0b2ab3/Firestore/Source/API/FIRTimestamp.m#L68

Firestore/Source/Public/FirebaseFirestore/FIRTransaction.h
```diff
@@ -29,8 +29,8 @@ NS_ASSUME_NONNULL_BEGIN
 NS_SWIFT_NAME(Transaction)
 @interface FIRTransaction : NSObject
 
-/** :nodoc: */
-- (id)init __attribute__((unavailable("FIRTransaction cannot be created directly.")));
+/** Initialiser. Only for use in unit tests. */
+- (id)init;
```
maksymmalyhin, line 33: Types like this look a bit different from the simple data types like `FIRTimestamp` or `FIRInAppMessagingDisplayMessage` where exposing the designated initializers looks natural and useful. In this case exposing the init actually may lead to confusion and reduce quality of the API. 

I think for types like this we should take a step forward and provide a fake implementation. We can have a separate module (e.g. `FirebaseTestsSupport`, or multiple like `FirebaseFirestoreTestSupport`, `FirebaseInAppMessagingTestsSupport`, etc.) with Objective-C implementations of the fakes which would subclass the original types and encapsulate the dirty work. In this case this module/podspec can be added as a dependency to tests only, so we can keep the public API clean but unblock testing.

@peterfriese @ryanwilson @paulb777 @ncooke3 WDYT?

FirebaseCombineSwift/Sources/Firestore/CollectionReference+Combine.swift
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2021 Google
```
paulb777, line 2: LLC

In general, please address the GHA check failures

Firestore/Source/Public/FirebaseFirestore/FIRCollectionReference.h
```diff
@@ -29,8 +29,8 @@ NS_ASSUME_NONNULL_BEGIN
 NS_SWIFT_NAME(CollectionReference)
 @interface FIRCollectionReference : FIRQuery
 
-/** :nodoc: */
-- (id)init __attribute__((unavailable("FIRCollectionReference cannot be created directly.")));
+/** Initialiser. Only for use in unit tests. */
```
paulb777, line 32: Please use American English in code and comments. "Initializer". I'm ok with British English in PR descriptions though.

Firestore/Source/Public/FirebaseFirestore/FIRTransaction.h
paulb777: SGTM

ncooke3: this approach SGTM as well!

peterfriese: SG - TBH, I opened up every class that looked remotely interesting, in an attempt to make sure @lorenzofiamingo is unblocked. I realise this might be overly broad, so I might lock the simple types down again and just keep the others open.

Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
peterfriese: We might not even need this. Rolling back for now.

Firestore/Source/Public/FirebaseFirestore/FIRTransaction.h
maksymmalyhin: SGTM to add `init` methods to unblock contribution for now with TODO to refactor it to be available only in a separate testing support target only.

FirebaseCombineSwift/Sources/Firestore/CollectionReference+Combine.swift
```diff
@@ -0,0 +1,89 @@
...
+
+#if canImport(FirebaseFirestoreSwift)
+
+import FirebaseFirestoreSwift
+
+#endif
+
+@available(swift 5.0)
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
+extension CollectionReference {
```
maksymmalyhin: Optional: it may be cleaner to have a separate PR for initializers only.

Firestore/Source/Public/FirebaseFirestore/FIRQuery.h
```diff
@@ -38,8 +38,9 @@ typedef void (^FIRQuerySnapshotBlock)(FIRQuerySnapshot *_Nullable snapshot,
  */
 NS_SWIFT_NAME(Query)
 @interface FIRQuery : NSObject
-/** :nodoc: */
-- (id)init __attribute__((unavailable("FIRQuery cannot be created directly.")));
+
+/** Initialiser. Only for use in unit tests. */
+- (id)init;
```
maksymmalyhin, line 43: Please add a TODO to revisit the approach for such classes.

maksymmalyhin (approved):
@peterfriese Thank you! LGTM, after adding TODOs.

I'll try to send an example with a separate testing support target soon.

peterfriese:
Closing this in favour of #7668

PR #7667, opened 2021-03-05
remove nightly_test version since versions are based on tags.

granluo:


paulb777:
@granluo Close since this was done in another PR?

PR #7668, opened 2021-03-06
Prototype of a module with Firestore fakes for testing dependent code in Swift

maksymmalyhin:
An alternative approach to #7660 to enable testing code dependent on Firestore in Swift.


FirebaseFirestoreTestingSupport.podspec
```diff
@@ -0,0 +1,63 @@
+Pod::Spec.new do |s|
+  s.name                    = 'FirebaseFirestoreTestingSupport'
```
peterfriese, line 2: Should we
a) define a pod for each of the products we support (i.e. FirebaseFirestoreTestingSupport, FirebaseAuthTestingSupport, etc.)
b) define a single pod (FirebaseTestingSupport)

Package.swift
```diff
@@ -837,6 +837,18 @@ let package = Package(
         .define("FIR_VERSION", to: firebaseVersion),
       ]
     ),
+
+    // MARK: Testing support
+
+    .target(
+      name: "FirebaseFirestoreTestingSupport",
```
peterfriese, line 844: Similar to the question about a single vs multiple pods, should we define a single vs. multiple Swift packages?

FirebaseTestingSupport/Firestore/Tests/QueryFakeTests.swift
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+@testable import FirebaseFirestoreTestingSupport
+
+class QueryFakeTests: XCTestCase {
```
peterfriese, line 18: I assume we would put the tests inside FirebaseCombineSwift/Tests

I've tested this, and it works (of course, we need to set a dependency from the FirebaseCombineSwift pod to the testing support pod),

peterfriese (commented):
This looks great, I like this a lot better than all the other approaches we've tried so far!

FirebaseFirestoreTestingSupport.podspec
maksymmalyhin: Good question! Initially I wanted to have a single FirebaseTestingSupport spec but discovered the following disadvantages:
- with a single podspec we will have to add unnecessary dependencies to the test targets (e.g. we probably don't want adding Firestore dependency to Auth tests)
- OS versions requirements may be different for different SDKs. It's harder to express with a single FirebaseTestingSupport spec.

So my proposal is to have separate podspecs/SPM targets for testing support code per SDK.

WDYT?

Package.swift
maksymmalyhin: I think the [response](https://github.com/firebase/firebase-ios-sdk/pull/7668#discussion_r589634538) is relevant here - I think separate targets are better.

FirebaseFirestoreTestingSupport.podspec
peterfriese: SG - happy with the extra flexibility this gives us.

FirebaseTestingSupport/Firestore/Tests/QueryFakeTests.swift
maksymmalyhin: This test is for fake implementation only and may be not 100% required, but it may help in the future debug the test failures and understand if it is the fake implementation issue or its client issue.

maksymmalyhin: Yeah, the testing support targets should be added as dependencies to the test targets which is not demonstrated in the current version of the PR.

maksymmalyhin: Good point on the docs! Will eventually add them, but feel free to start using it if it unblocks the Combine effort.

FirebaseFirestoreTestingSupport.podspec
```diff
@@ -0,0 +1,63 @@
+Pod::Spec.new do |s|
```
paulb777, line 1: If the podspec is only for testing and not for publishing, it might be better to organize in a lower level directory like how Firestore manages LibFuzzer and GoogleBenchmark in Firestore/Example.

FirebaseFirestoreTestingSupport.podspec
```diff
@@ -0,0 +1,63 @@
...
+  s.name                    = 'FirebaseFirestoreTestingSupport'
+  s.version                 = '7.7.0'
```
paulb777, line 3: How about 1.0.0.

I'm not sure we should tie the version to Firebase versioning?

paulb777 (commented):
Let's sort out the directory structure and then LGTM

FirebaseFirestoreTestingSupport.podspec
maksymmalyhin: Good idea, will do.

FirebaseFirestoreTestingSupport.podspec
maksymmalyhin: I like the idea! Let me try to restructure it to see if it works well.

FirebaseFirestoreTestingSupport.podspec
maksymmalyhin: Moved under FirebaseTestingSupport folder.

maksymmalyhin: Hm, just realized that having the podspec in a subfolder will require `scripts/pod_lib_lint.rb` modifications to accommodate it. I'll look into it but we may want to move the podspec back to the root dir temporary to unblock testing.

maksymmalyhin: I'm moving the podspec back to the root dir so far as we do some assumptions in `scripts/pod_lib_lint.rb` that make subfolder support not trivial (not impossible, but just requires a bit of refactoring). 

Another thought, as we may offer the testing support to developer, we may want to keep the spec in the root dir. Let's discuss it a bit more and apply changes if needed in the following PRs.

FirebaseTestingSupport/Firestore/Tests/QueryFakeTests.swift
maksymmalyhin: I added a README file with brief explanations and common steps, PTAL.

FirebaseTestingSupport/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Development
+
+### Generate project
+
+#### Cocoapods
+
+E.g. for `FirebaseFirestoreTestingSupport` run the following command for the root repo directory:
+
+```
+pod gen --auto-open --local-sources=./,./FirbaseTestingSupport/ ./FirebaseTestingSupport/FirebaseFirestoreTestingSupport.podspec --platforms=ios
```
paulb777, line 40: Extra comma?

FirebaseTestingSupport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### Swift Package Manager
+
+- Open the main package definition:
+
+```
+xed Package.swift
+```
+
+- Select (or add if not exists yet) a scheme for the test targte, e.g. `FirestoreTestingSupportTests`
```
paulb777, line 51: s/targte/target/

FirebaseTestingSupport/README.md
```diff
@@ -0,0 +1,53 @@
+# Firebase Testing Support
```
paulb777, line 1: Clarify that these instructions are for test creators not test users.

paulb777 (approved):
LGTM on comment resolution and green

FirebaseTestingSupport/README.md
maksymmalyhin: This is the command for the case when the spec is in the subfolder. The docs are becoming outdated even before the PR merging 😄.

FirebaseTestingSupport/README.md
maksymmalyhin: Updated, PTAL.

FirebaseFirestoreTestingSupport.podspec
```diff
@@ -0,0 +1,63 @@
...
+  s.requires_arc            = true
+
+  base_dir = 'FirebaseTestingSupport/Firestore/'
+
+  s.source_files = [
+    base_dir + 'Sources/**/*.{m,mm,h}',
+    'Firestore/Source/API/*\+Internal.h'
+  ]
+
+  s.preserve_paths = [
```
paulb777, line 40: I don't think we want preserve_paths since they break Xcode completion and we don't need to worry about tests mapping up an end user's header map (the reason they're there in the Firestore pod)

FirebaseFirestoreTestingSupport.podspec
```diff
@@ -0,0 +1,63 @@
...
+    'Firestore/Source/API/*.h',
+  ]
+
+  s.public_header_files = base_dir + '**/*.h'
+
+  s.dependency 'FirebaseFirestore', '~> 7.0'
```
paulb777, line 46: Might as well start with the current 7.7 or 7.8 since we'll never test earlier versions.

FirebaseTestingSupport/README.md
```diff
@@ -0,0 +1,53 @@
+# Firebase Testing Support (for automated test developers)
+
+Firebase Testing Support is a collection of libraries that provide type definitions  and tools required for writing tests for code that uses Firebase, e.g.:
```
paulb777, line 3: Extra space before "and"

paulb777 (approved):

FirebaseFirestoreTestingSupport.podspec
maksymmalyhin: Good catch! It is unnecessary leftovers from the attempts to fix C++ dependent code.

Thank you for the thorough review!

maksymmalyhin:
@peterfriese The PR looks in a good shape. Feel free to merge it or let me know when you would like me to do it.

peterfriese:
Will get to it later today, thanks for putting this together!

peterfriese (approved):
This looks great, thanks @maksymmalyhin!

PR #7563, opened 2021-02-18
Future Apple toolchain fixes

paulb777:
Fix #7552 

The future Apple toolchain is more strict about a few things that impact Firestore and its dependencies:

- Requires usage of stdlib.h instead of cstdlib for some declarations like https://github.com/google/boringssl/commit/061a7f5596c4125fc4bead14ebb37cf51d2ae627
- Requires all headers specified by SPM to build (wasn't the case previously and the packages included some test headers)
- More strict about C++ headers being in extern C - see https://github.com/firebase/leveldb/commit/e4e8ab44a2781a3656b4b5852e71c07fd18fecfa
- Associated PRs
  - https://github.com/firebase/leveldb/pull/2
  - https://github.com/firebase/boringSSL-SwiftPM/pull/1
  - https://github.com/firebase/abseil-cpp-SwiftPM/pull/2
  - https://github.com/firebase/grpc-SwiftPM/pull/1

#no-changelog - will be invisible to clients if this lands before the future toolchain releases

PR #7678, opened 2021-03-08
Json parser as a hard copy

wu-hui:
Introduce the json lib as a third party dep by copying the lib into our repo. The reason is there is no official SPM and cocoa pods build spec for the lib, and the owner does not want to maintain those. There is a cocoa pods spec that does not work for us, the the owner of that pod is not maintaining as well. This makes a hard copy the most hassle free approach.


Firestore/core/src/remote/grpc_root_certificates_generated.h
```diff
@@ -3,7 +3,7 @@
 #ifndef FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 #define FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 
-#include <stddef.h>
+#include <cstdlib>
```
var-const, line 6: Please revert -- this overwrites the change from #7563.

var-const (changes requested):
* Can you please provide a PR description to explain why this is necessary?
* Have you synced with Paul @paulb777 regarding this approach? In particular, I wonder if there's any legal clearance we need to do to add third-party sources to our repo?

paulb777:
@wu-hui and I discussed this approach this morning and agreed the complexity of managing the copy would be substantially less than managing a CocoaPods podspec and a Swift Package - and it's fine to add MIT licensed code as long as it goes to a third_party directory and is acknowledged.

On thing to consider would be if it's necessary to do symbol renaming to avoid symbol collisions in case Firestore clients could have their own copy of `json.hpp`.

A copy of the LICENSE should appended to https://github.com/firebase/firebase-ios-sdk/blob/master/CoreOnly/NOTICES

paulb777:
On 2nd thought, my C++ is relatively weak, but I don't think a pure header could cause symbol collisions.

Firestore/core/src/remote/grpc_root_certificates_generated.h
wu-hui: Done.

wu-hui:
> On 2nd thought, my C++ is relatively weak, but I don't think a pure header could cause symbol collisions.

Yeah, I think we should be fine here.

var-const:
Thanks, Paul! I also don't think a header-only library should cause symbol collisions.

var-const (approved):

var-const:
@wu-hui Andy, can you add some of the rationale for this PR to the PR description?

PR #7683, opened 2021-03-09
adding more token tests

chliangGoogle:
Move the following token class tests from IID to Messaging
TokenInfo
TokenOperation
TokenStore
Utilities
#no-changelog


chliangGoogle:
Not sure the breakage is related, sth to do with Firestore.

paulb777:
I suspect the CI failure may resolve with a merge or rebase of origin/master

chliangGoogle:
That means I will need to rebase for both my dev branch and master branch.

paulb777:
Sorry I missed that this was targeting the feature branch. That makes the git foo a bit more complicated. I believe the diffs might get cleaned up if you rebase and force push `chen/fm-master`.  Let me know if you want me to work on fixing both branches or schedule a GVC if you want to pair on it.

chliangGoogle:
@paulb777 I'm not sure we can fix chen/fm-master, unless we can revert my previous commit. But yeah feel free to do it without me or we can schedule a time when you are free.

PR #7685, opened 2021-03-09
Allows clients to sign in with just a refresh token

joehinkle11:
A common problem I've had with the Firebase SDK is that you can't manage your own refresh token because there's no public API to convert the token to a User object. This PR solves this problem with minimal code by exposing the internal `completeSignInWithAccessToken` with a public `signInWithRefreshToken` method.

Example usage:
```swift
if let refreshToken = getManuallyManagedRefreshToken() {
    Auth.auth().signIn(withRefreshToken: refreshToken) { (result, error) in
        print(result) // -> User?
        print(error)
    }
}
```


rosalyntan:
Thanks for this contribution (and for your previous contributions), @joehinkle11!

Unfortunately, adding a `signInWithRefreshToken` method to the public API is something the auth team has previously considered and decided against. Since the refresh token is a long-lived token, we don't want to encourage passing them around.

There have been discussions around other solutions to the use cases that exposing a `signInWithRefreshToken` method would solve, but these require some backend changes as well.

PR #7686, opened 2021-03-09
fix asan errors

wu-hui:
Fix 3 things came up in ASAN/TSAN CI run

1. `loadBundle` completion callback might call before the last progress observer, leads to tests checking for state too early. Added `ObserveAtLast` to make sure the completion is called when all other observers are called.

2. A subtle case of returning reference to a temp value in bundle_serializer.cc

3. The `LoadBundleTask` destructor needs to be guarded by mutex as well, as user could very well request some change there while it is being destructed.


Firestore/core/src/api/load_bundle_task.h
```diff
@@ -153,13 +154,27 @@ class LoadBundleTask {
...
    * progress update.
    *
+   * For a given progress update, this observer is guaranteed to be called
+   * after all other observers. Calling `ObserveAtLast` a second time will
+   * override the observer registered the first time.
+   *
    * @return A handle that can be used to remove the callback from this task.
    */
-  LoadBundleHandle Observe(ProgressObserver callback);
+  LoadBundleHandle ObserveAtLast(ProgressObserver observer);
```
var-const, line 177: Consider a name like `SetLastObserver`. It won't be symmetrical with `Observe`, but since this function has different semantics (only one last observer can be added), this seems okay.

Firestore/core/src/api/load_bundle_task.cc
```diff
@@ -27,6 +27,10 @@ namespace firebase {
 namespace firestore {
 namespace api {
 
+LoadBundleTask::~LoadBundleTask() {
+  std::lock_guard<std::mutex> lock(mutex_);
```
var-const, line 31: Question: can you describe a scenario where this lock is helpful?

Firestore/core/src/api/load_bundle_task.h
wu-hui: Done.

Firestore/core/src/api/load_bundle_task.cc
wu-hui: In `TEST_F(LoadBundleTaskTest, ProgressObserverCanRemoveObserver)` for example:

for the last call of `queue.push(3)` happen only user thread. The testing thread could start checking `EXPECT_EQ` right after this call and they will all pass.

Now the user thread could move on to `task.RemoveObserver(2)`, holding the lock, but the testing thread will try to deconstruct `task` right after the checks, hence racing to write the underlying fields.

Firestore/core/test/unit/api/load_bundle_task_test.cc
```diff
@@ -320,7 +320,7 @@ TEST_F(LoadBundleTaskTest, NoObserversAlsoWork) {
 
 TEST_F(LoadBundleTaskTest, ObserveAtLastIsHonored) {
```
var-const, line 321: Nit: the test name also needs to be changed for the new function name.

Firestore/core/src/api/load_bundle_task.cc
var-const: Ah, so sounds like it's to make sure the destructor cannot proceed while an observer is being removed, right? And the problem is that the test can finish in-between `queue.push(3)` and `task.RemoveObserver(2)`?

I think we usually solved this in tests via a `std::promise` that is fulfilled by the callback while the test waits for the promise to be fulfilled before it is finished. This approach seems to work as well. One thing to mention is that in C++ it's usually desirable to define all the copy/move operations and the destructor together (rule of 5). To be pedantic, you would probably need to protect the handle and the observers by a mutex while copying and moving as well.

wu-hui: As discussed, made it no copy-able.

Firestore/core/test/unit/api/load_bundle_task_test.cc
wu-hui: Done.

wu-hui (commented):
Sorry, I accidentally merged the spec test PR into this one.

var-const (approved):

PR #7712, opened 2021-03-13
Add ObjectValue

schmidt-sebastian:
This PR adds a mutable, Protobuf-backed ObjectValue similar to https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java

Since nanopb Protos can be directly manipulated, I did not port the `Overlay` approach from Android. Rather, the protos are modified directly. To ensure O(log n) Map lookups, Maps are kept sorted as well (all maps start out sorted when we receive them from the backend, but on Web and Android they are not necessarily sorted after we apply local modifications). The guaranteed sorted order also allows me to simplify some of the code in value_utils as well.

### DIFF
--- a/FirebaseFirestore.podspec
+++ b/FirebaseFirestore.podspec
@@ -89,6 +89,8 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   abseil_version = '0.20200225.0'
   s.dependency 'abseil/algorithm', abseil_version
   s.dependency 'abseil/base', abseil_version
+  s.dependency 'abseil/container/flat_hash_map', abseil_version
+  s.dependency 'abseil/container/flat_hash_set', abseil_version
   s.dependency 'abseil/memory', abseil_version
   s.dependency 'abseil/meta', abseil_version
   s.dependency 'abseil/strings/strings', abseil_version
--- a/FirebaseFirestore.podspec
+++ b/FirebaseFirestore.podspec
@@ -90,7 +90,6 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   s.dependency 'abseil/algorithm', abseil_version
   s.dependency 'abseil/base', abseil_version
   s.dependency 'abseil/container/flat_hash_map', abseil_version
-  s.dependency 'abseil/container/flat_hash_set', abseil_version
   s.dependency 'abseil/memory', abseil_version
   s.dependency 'abseil/meta', abseil_version
   s.dependency 'abseil/strings/strings', abseil_version
--- a/FirebaseFirestore.podspec
+++ b/FirebaseFirestore.podspec
@@ -89,7 +89,6 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   abseil_version = '0.20200225.0'
   s.dependency 'abseil/algorithm', abseil_version
   s.dependency 'abseil/base', abseil_version
-  s.dependency 'abseil/container/flat_hash_map', abseil_version
   s.dependency 'abseil/memory', abseil_version
   s.dependency 'abseil/meta', abseil_version
   s.dependency 'abseil/strings/strings', abseil_version
--- a/Firestore/core/src/model/object_value.cc
+++ b/Firestore/core/src/model/object_value.cc
@@ -17,14 +17,13 @@
 #include "Firestore/core/src/model/object_value.h"
 
 #include <algorithm>
+#include <map>
 #include <set>
 
 #include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
 #include "Firestore/core/src/nanopb/fields_array.h"
 #include "Firestore/core/src/nanopb/message.h"
 #include "Firestore/core/src/nanopb/nanopb_util.h"
-#include "absl/container/flat_hash_map.h"
-#include "absl/container/flat_hash_set.h"
 
 namespace firebase {
 namespace firestore {
@@ -75,8 +74,8 @@ google_firestore_v1_MapValue_FieldsEntry* FindEntry(
 
 size_t CalculateSizeOfUnion(
     google_firestore_v1_MapValue* map_value,
-    const absl::flat_hash_map<std::string, google_firestore_v1_Value>& upserts,
-    const absl::flat_hash_set<std::string>& deletes) {
+    const std::map<std::string, google_firestore_v1_Value>& upserts,
+    const std::set<std::string>& deletes) {
   // Compute the size of the map after applying all mutations. The final size is
   // the number of existing entries, plus the number of new entries
   // minus the number of deleted entries.
@@ -84,7 +83,7 @@ size_t CalculateSizeOfUnion(
          std::count_if(
              map_value->fields, map_value->fields + map_value->fields_count,
              [&](const google_firestore_v1_MapValue_FieldsEntry& entry) {
-               absl::string_view field = MakeStringView(entry.key);
+               std::string field = MakeString(entry.key);
                // Don't count if entry is deleted or if it is a replacement
                // rather than an insert.
                return deletes.find(field) == deletes.end() &&
@@ -98,8 +97,8 @@ size_t CalculateSizeOfUnion(
  */
 void ApplyChanges(
     google_firestore_v1_MapValue* parent,
-    const absl::flat_hash_map<std::string, google_firestore_v1_Value>& upserts,
-    const absl::flat_hash_set<std::string>& deletes) {
+    const std::map<std::string, google_firestore_v1_Value>& upserts,
+    const std::set<std::string>& deletes) {
   // TODO(mrschmidt): Consider using `absl::btree_map` and `absl::btree_set` for
   // potentially better performance.
   auto source_count = parent->fields_count;
@@ -228,7 +227,7 @@ void MutableObjectValue::Set(const FieldPath& path,
   google_firestore_v1_MapValue* parent_map = ParentMap(path.PopLast());
 
   std::string last_segment = path.last_segment();
-  absl::flat_hash_map<std::string, google_firestore_v1_Value> upserts{
+  std::map<std::string, google_firestore_v1_Value> upserts{
       {std::move(last_segment), value}};
 
   ApplyChanges(parent_map, upserts, /*deletes=*/{});
@@ -238,8 +237,8 @@ void MutableObjectValue::SetAll(const FieldMask& field_mask,
                                 const MutableObjectValue& data) {
   FieldPath parent;
 
-  absl::flat_hash_map<std::string, google_firestore_v1_Value> upserts;
-  absl::flat_hash_set<std::string> deletes;
+  std::map<std::string, google_firestore_v1_Value> upserts;
+  std::set<std::string> deletes;
 
   for (const FieldPath& path : field_mask) {
     if (!parent.IsImmediateParentOf(path)) {
@@ -280,7 +279,7 @@ void MutableObjectValue::Delete(const FieldPath& path) {
   // We can only delete a leaf entry if its parent is a map.
   if (nested_value->which_value_type ==
       google_firestore_v1_Value_map_value_tag) {
-    absl::flat_hash_set<std::string> deletes{path.last_segment()};
+    std::set<std::string> deletes{path.last_segment()};
     ApplyChanges(&nested_value->map_value, /*upserts=*/{}, deletes);
   }
 }
@@ -314,7 +313,7 @@ google_firestore_v1_MapValue* MutableObjectValue::ParentMap(
       google_firestore_v1_Value new_entry{};
       new_entry.which_value_type = google_firestore_v1_Value_map_value_tag;
 
-      absl::flat_hash_map<std::string, google_firestore_v1_Value> upserts{
+      std::map<std::string, google_firestore_v1_Value> upserts{
           {segment, new_entry}};
       ApplyChanges(&parent->map_value, upserts, /*deletes=*/{});
 
--- a/Firestore/core/src/model/object_value.cc
+++ b/Firestore/core/src/model/object_value.cc
@@ -16,10 +16,12 @@
 
 #include "Firestore/core/src/model/object_value.h"
 
+#include <algorithm>
 #include <map>
 #include <set>
 
 #include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
+#include "Firestore/core/src/nanopb/fields_array.h"
 #include "Firestore/core/src/nanopb/message.h"
 #include "Firestore/core/src/nanopb/nanopb_util.h"
 
@@ -27,11 +29,55 @@ namespace firebase {
 namespace firestore {
 namespace model {
 
+namespace {
+
+using nanopb::FieldsArray;
+using nanopb::FreeNanopbMessage;
+using nanopb::MakeArray;
+using nanopb::MakeBytesArray;
+using nanopb::MakeString;
+using nanopb::MakeStringView;
+
+struct MapEntryKeyCompare {
+  bool operator()(const google_firestore_v1_MapValue_FieldsEntry& entry,
+                  absl::string_view segment) {
+    return nanopb::MakeStringView(entry.key) < segment;
+  }
+  bool operator()(absl::string_view segment,
+                  const google_firestore_v1_MapValue_FieldsEntry& entry) {
+    return segment < nanopb::MakeStringView(entry.key);
+  }
+};
+
+/**
+ * Finds an entry by key in the provided map value. Returns `nullptr` if the
+ * entry does not exist.
+ */
+static google_firestore_v1_MapValue_FieldsEntry* FindEntry(
+    const google_firestore_v1_Value& value, absl::string_view segment) {
+  if (value.which_value_type != google_firestore_v1_Value_map_value_tag) {
+    return nullptr;
+  }
+  const google_firestore_v1_MapValue& map_value = value.map_value;
+
+  // MapValues in iOS are always stored in sorted order.
+  auto found = std::equal_range(map_value.fields,
+                                map_value.fields + map_value.fields_count,
+                                segment, MapEntryKeyCompare());
+
+  if (found.first == found.second) {
+    return nullptr;
+  }
+
+  return found.first;
+}
+
+}  // namespace
+
 MutableObjectValue::MutableObjectValue() {
   value_->which_value_type = google_firestore_v1_Value_map_value_tag;
   value_->map_value.fields_count = 0;
-  value_->map_value.fields =
-      nanopb::MakeArray<google_firestore_v1_MapValue_FieldsEntry>(0);
+  value_->map_value.fields = nullptr;
 }
 
 model::FieldMask MutableObjectValue::ToFieldMask() const {
@@ -41,58 +87,62 @@ model::FieldMask MutableObjectValue::ToFieldMask() const {
 model::FieldMask MutableObjectValue::ExtractFieldMask(
     const google_firestore_v1_MapValue& value) const {
   std::set<FieldPath> fields;
+
   for (size_t i = 0; i < value.fields_count; ++i) {
-    FieldPath current_path({nanopb::MakeString(value.fields[i].key)});
-    if (value.fields[i].value.which_value_type ==
+    google_firestore_v1_MapValue_FieldsEntry& entry = value.fields[i];
+    FieldPath current_path({MakeString(entry.key)});
+
+    if (entry.value.which_value_type !=
         google_firestore_v1_Value_map_value_tag) {
-      // Recursively extract the nested map
-      model::FieldMask nested_mask =
-          ExtractFieldMask(value.fields[i].value.map_value);
-      if (nested_mask.begin() == nested_mask.end()) {
-        // Preserve the empty map by adding it to the FieldMask.
-        fields.insert(current_path);
-      } else {
-        for (const FieldPath& nested_path : nested_mask) {
-          fields.insert(current_path.Append(nested_path));
-        }
-      }
-    } else {
       fields.insert(current_path);
+      continue;
+    }
+
+    // Recursively extract the nested map
+    FieldMask nested_mask = ExtractFieldMask(entry.value.map_value);
+    if (nested_mask.begin() == nested_mask.end()) {
+      // Preserve the empty map by adding it to the FieldMask.
+      fields.insert(current_path);
+    } else {
+      for (const FieldPath& nested_path : nested_mask) {
+        fields.insert(current_path.Append(nested_path));
+      }
     }
   }
-  return model::FieldMask(fields);
+
+  return FieldMask(std::move(fields));
 }
 
 absl::optional<google_firestore_v1_Value> MutableObjectValue::Get(
-    const firebase::firestore::model::FieldPath& path) const {
+    const FieldPath& path) const {
   if (path.empty()) {
     return *value_;
-  } else {
-    google_firestore_v1_Value nested_value = *value_;
-    for (const std::string& segment : path) {
-      _google_firestore_v1_MapValue_FieldsEntry* entry =
-          FindEntry(nested_value, segment);
-      if (!entry) return {};
-      nested_value = entry->value;
-    }
-    return nested_value;
   }
+
+  google_firestore_v1_Value nested_value = *value_;
+  for (const std::string& segment : path) {
+    google_firestore_v1_MapValue_FieldsEntry* entry =
+        FindEntry(nested_value, segment);
+    if (!entry) return {};
+    nested_value = entry->value;
+  }
+  return nested_value;
 }
 
-void MutableObjectValue::Set(const model::FieldPath& path,
+void MutableObjectValue::Set(const FieldPath& path,
                              const google_firestore_v1_Value& value) {
   HARD_ASSERT(!path.empty(), "Cannot set field for empty path on ObjectValue");
 
-  _google_firestore_v1_MapValue* parent_map = ParentMap(path.PopLast());
+  google_firestore_v1_MapValue* parent_map = ParentMap(path.PopLast());
 
   std::string last_segment = path.last_segment();
   std::map<std::string, google_firestore_v1_Value> upserts{
-      {last_segment, value}};
+      {std::move(last_segment), value}};
 
-  ApplyChanges(parent_map, upserts, /* deletes= */ {});
+  ApplyChanges(parent_map, upserts, /*deletes=*/{});
 }
 
-void MutableObjectValue::SetAll(const model::FieldMask& field_mask,
+void MutableObjectValue::SetAll(const FieldMask& field_mask,
                                 const MutableObjectValue& data) {
   FieldPath parent;
 
@@ -102,7 +152,7 @@ void MutableObjectValue::SetAll(const model::FieldMask& field_mask,
   for (const FieldPath& path : field_mask) {
     if (!parent.IsImmediateParentOf(path)) {
       // Insert the accumulated changes at this parent location
-      _google_firestore_v1_MapValue* parent_map = ParentMap(parent);
+      google_firestore_v1_MapValue* parent_map = ParentMap(parent);
       ApplyChanges(parent_map, upserts, deletes);
       upserts.clear();
       deletes.clear();
@@ -117,17 +167,43 @@ void MutableObjectValue::SetAll(const model::FieldMask& field_mask,
     }
   }
 
-  _google_firestore_v1_MapValue* parent_map = ParentMap(parent);
+  google_firestore_v1_MapValue* parent_map = ParentMap(parent);
   ApplyChanges(parent_map, upserts, deletes);
 }
 
+void MutableObjectValue::Delete(const FieldPath& path) {
+  HARD_ASSERT(!path.empty(), "Cannot set field for empty path on ObjectValue");
+
+  google_firestore_v1_Value* nested_value = value_.get();
+  for (const std::string& segment : path.PopLast()) {
+    google_firestore_v1_MapValue_FieldsEntry* entry =
+        FindEntry(*nested_value, segment);
+    if (!entry) {
+      // If the entry is not found, exit early. There is nothing to delete.
+      return;
+    }
+    nested_value = &entry->value;
+  }
+
+  // We can only delete a leaf entry if its parent is a map.
+  if (nested_value->which_value_type ==
+      google_firestore_v1_Value_map_value_tag) {
+    std::set<std::string> deletes{path.last_segment()};
+    ApplyChanges(&nested_value->map_value, /*upserts=*/{}, deletes);
+  }
+}
+
+/**
+ * Returns the map that contains the leaf element of `path`. If the parent
+ * entry does not yet exist, or if it is not a map, a new map will be created.
+ */
 google_firestore_v1_MapValue* MutableObjectValue::ParentMap(
     const FieldPath& path) {
   google_firestore_v1_Value* parent = value_.get();
 
   // Find a or create a parent map entry for `path`.
   for (const std::string& segment : path) {
-    _google_firestore_v1_MapValue_FieldsEntry* entry =
+    google_firestore_v1_MapValue_FieldsEntry* entry =
         FindEntry(*parent, segment);
 
     if (entry) {
@@ -135,19 +211,20 @@ google_firestore_v1_MapValue* MutableObjectValue::ParentMap(
           google_firestore_v1_Value_map_value_tag) {
         // Since the element is not a map value, free all existing data and
         // change it to a map type
-        nanopb::FreeNanopbMessage(google_firestore_v1_Value_fields,
-                                  &entry->value);
+        FreeNanopbMessage(FieldsArray<google_firestore_v1_Value>(),
+                          &entry->value);
         entry->value.which_value_type = google_firestore_v1_Value_map_value_tag;
       }
+
       parent = &entry->value;
     } else {
       // Create a new map value for the current segment
-      _google_firestore_v1_Value new_entry{};
+      google_firestore_v1_Value new_entry{};
       new_entry.which_value_type = google_firestore_v1_Value_map_value_tag;
 
       std::map<std::string, google_firestore_v1_Value> upserts{
           {segment, new_entry}};
-      ApplyChanges(&(parent->map_value), upserts, {});
+      ApplyChanges(&(parent->map_value), upserts, /*deletes=*/{});
 
       parent = &(FindEntry(*parent, segment)->value);
     }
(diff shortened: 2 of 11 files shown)


Firestore/core/test/unit/model/object_value_test.cc
```diff
@@ -0,0 +1,337 @@
...
+
+namespace {
+
+using absl::nullopt;
+using testutil::DbId;
+using testutil::Field;
+using testutil::Map;
+using testutil::Value;
+
+class ObjectValueTest : public ::testing::Test {
```
schmidt-sebastian, line 39: Note - can be compared to end result of https://github.com/firebase/firebase-android-sdk/pull/2494

schmidt-sebastian:
I think this leaks memory - we need to free the underlying Protobuf structures when MutableObjectValue is destructed. I will work on a fix.

schmidt-sebastian:
> I think this leaks memory - we need to free the underlying Protobuf structures when MutableObjectValue is destructed. I will work on a fix.

Done

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      {last_segment, value}};
+
+  ApplyChanges(parent_map, upserts, /* deletes= */ {});
+}
+
+void MutableObjectValue::SetAll(const model::FieldMask& field_mask,
+                                const MutableObjectValue& data) {
+  FieldPath parent;
+
+  std::map<std::string, google_firestore_v1_Value> upserts;
```
var-const, line 99: I think you can get away with storing pointers to protos in this map -- they would point into `data` which should stay valid for the duration of this function.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      if (entry->value.which_value_type !=
+          google_firestore_v1_Value_map_value_tag) {
+        // Since the element is not a map value, free all existing data and
+        // change it to a map type
+        nanopb::FreeNanopbMessage(google_firestore_v1_Value_fields,
+                                  &entry->value);
+        entry->value.which_value_type = google_firestore_v1_Value_map_value_tag;
+      }
+      parent = &entry->value;
+    } else {
```
var-const, line 143: Nit: can you add a blank line before `else` to visually separate the two branches?

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+  for (const std::string& segment : path) {
+    _google_firestore_v1_MapValue_FieldsEntry* entry =
+        FindEntry(*parent, segment);
+
+    if (entry) {
+      if (entry->value.which_value_type !=
+          google_firestore_v1_Value_map_value_tag) {
+        // Since the element is not a map value, free all existing data and
+        // change it to a map type
+        nanopb::FreeNanopbMessage(google_firestore_v1_Value_fields,
```
var-const, line 138: Optional: rather than using `google_firestore_v1_Value_fields` directly, consider using `FieldsArray<google_firestore_v1_Value>()` (from `nanopb/fields_array.h`). It's a little more verbose, but `google_firestore_v1_Value_fields` seems like a really low-level detail that's nice to avoid.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      }
+      parent = &entry->value;
+    } else {
+      // Create a new map value for the current segment
+      _google_firestore_v1_Value new_entry{};
+      new_entry.which_value_type = google_firestore_v1_Value_map_value_tag;
+
+      std::map<std::string, google_firestore_v1_Value> upserts{
+          {segment, new_entry}};
+      ApplyChanges(&(parent->map_value), upserts, {});
```
var-const, line 150: Nit: add the argument name (in a comment) to the empty braces?

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+
+  // Compute the size of the map after applying all mutations. The final size is
+  // the number of existing entries, plus the number of new entries
+  // minus the number of deleted entries.
+  size_t target_count =
+      upserts.size() +
+      std::count_if(source_fields, source_fields + source_count,
+                    [&](_google_firestore_v1_MapValue_FieldsEntry entry) {
+                      std::string field = nanopb::MakeString(entry.key);
+                      // Check if the entry is deleted or if it is a replacement
```
var-const, line 174: Optional nit: `s/Check if the entry is/Don't count the entry if it is deleted/`.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+                      // rather than an insert.
+                      return deletes.find(field) == deletes.end() &&
+                             upserts.find(field) == upserts.end();
+                    });
+
+  auto target_fields = static_cast<_google_firestore_v1_MapValue_FieldsEntry*>(
```
var-const, line 180: Can you use `nanopb::MakeArray`? Also, this should be `auto*` to make it clear the variable is a pointer.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      malloc(target_count * sizeof(_google_firestore_v1_MapValue_FieldsEntry)));
+
+  auto delete_it = deletes.begin();
+  auto upsert_it = upserts.begin();
+
+  // Merge the existing data with the deletes and updates.
+  for (pb_size_t target_index = 0, source_index = 0;
```
var-const, line 187: Ultranit: swap the variable order? (I'd expect "source" to go before the "target")

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+       target_index < target_count;) {
+    if (source_index < source_count) {
+      std::string key = nanopb::MakeString(source_fields[source_index].key);
+
+      // Check if the source key is deleted
+      if (delete_it != deletes.end() && *delete_it == key) {
+        nanopb::FreeNanopbMessage(
```
var-const, line 194: Consider creating a helper function that calls `nanopb::FreeNanopbMessage` on `google_firestore_v1_MapValue_FieldsEntry_fields` and only takes a pointer as the parameter.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+
+  return &parent->map_value;
+}
+
+void MutableObjectValue::ApplyChanges(
+    google_firestore_v1_MapValue* parent,
+    std::map<std::string, google_firestore_v1_Value> upserts,
+    std::set<std::string> deletes) const {
+  auto source_count = parent->fields_count;
+  auto source_fields = parent->fields;
```
var-const, line 164: Nit: `s/auto/auto*/`.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      parent = &(FindEntry(*parent, segment)->value);
+    }
+  }
+
+  return &parent->map_value;
+}
+
+void MutableObjectValue::ApplyChanges(
+    google_firestore_v1_MapValue* parent,
+    std::map<std::string, google_firestore_v1_Value> upserts,
```
var-const, line 161: It looks like you can pass `upserts` and `deletes` by const reference.

var-const (changes requested):
Did a partial review.

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,147 @@
...
+   */
+  void Set(const FieldPath& path, const google_firestore_v1_Value& value);
+
+  /**
+   * Sets the provided fields to the provided values. Only fields included in
+   * `field_mask` are modified. If a field is included in field_mask, but
+   * missing in `data`, it is deleted.
+   *
+   * @param field_mask The field mask that controls which fields to modify.
+   * @param data An ObjectValue that contains the field values.
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed (but I will probably rename it back once I remove the original `ObjectValue`).

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,147 @@
...
+  void SetAll(const model::FieldMask& field_mask,
+              const MutableObjectValue& data);
+
+  friend bool operator==(const MutableObjectValue& lhs,
+                         const MutableObjectValue& rhs);
+  friend std::ostream& operator<<(std::ostream& out,
+                                  const MutableObjectValue& object_value);
+
+ private:
+  nanopb::Message<google_firestore_v1_Value> value_{};
```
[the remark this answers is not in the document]
schmidt-sebastian: That's an easy fix (and easy to understand reasoning).

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,147 @@
...
+
+  /** Returns the field mask for the provided map value. */
```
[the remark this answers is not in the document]
schmidt-sebastian: Cleaned up.

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,147 @@
...
+  model::FieldMask ExtractFieldMask(
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed everywhere.

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,147 @@
...
+      const google_firestore_v1_MapValue& value) const;
+
+  /**
+   * Finds an entry by key in the provided map value. Returns `nullptr` if the
+   * entry does not exist.
+   */
+  static _google_firestore_v1_MapValue_FieldsEntry* FindEntry(
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/nanopb/message.h
```diff
@@ -65,6 +65,12 @@ class Message {
    */
   Message() = default;
 
+  /**
+   * Creates a `Message` object that wraps `proto`. Takes ownership of proto.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/nanopb/message.h
```diff
@@ -65,6 +65,12 @@ class Message {
...
+   */
+  explicit Message(T proto) : owns_proto_(true), proto_(proto) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Fun fun. Done.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+      ApplyChanges(&(parent->map_value), upserts, /*deletes=*/{});
+
+      parent = &(FindEntry(*parent, segment)->value);
+    }
+  }
+
+  return &parent->map_value;
+}
+
+void MutableObjectValue::ApplyChanges(
```
var-const, line 236: Does this need to be a member function? It doesn't seem to refer to any member variables -- if that's the case, consider making it a free function.

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -132,6 +133,15 @@ class ValueUtilTest : public ::testing::Test {
     EXPECT_EQ(expected_canonical_id, actual_canonical_id);
   }
 
+  void VerifyDeepClone(const google_firestore_v1_Value& value) {
+    // `Message` takes ownership of the proto value and destroys it. We use
```
var-const, line 137: Hmm, I'm not sure if it verifies that values own their own memory -- `clone1` and `clone2` will be valid for the duration of the test. If I understand the intention correctly, consider something like:
```cpp
    nanopb::Message<google_firestore_v1_Value> clone1;
    [&] {
      nanopb::Message<google_firestore_v1_Value> clone2{DeepClone(value)};
      EXPECT_TRUE(value == *clone2);
      clone1 = nanopb::Message<google_firestore_v1_Value>{DeepClone(*clone2)};
    }();
    // clone2 is destroyed at this point, but clone1 should be still valid.
    EXPECT_TRUE(value == *clone1);
```

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -132,6 +133,15 @@ class ValueUtilTest : public ::testing::Test {
...
+    // it to verify that all values own their own memory.
+    nanopb::Message<google_firestore_v1_Value> clone1{DeepClone(value)};
+    nanopb::Message<google_firestore_v1_Value> clone2{DeepClone(*clone1)};
+    EXPECT_TRUE(value == *clone1);
```
var-const, line 141: Question: when these tests fail, what do the failure messages look like? Is it clear which invocation of `VerifyDeepClone` failed?

var-const (changes requested):

Firestore/core/src/model/value_util.h
```diff
@@ -60,6 +60,9 @@ util::ComparisonResult Compare(const google_firestore_v1_Value& left,
  */
 std::string CanonicalId(const google_firestore_v1_Value& value);
 
+/** Creates a copy of the contents of the Value proto. */
+google_firestore_v1_Value DeepClone(google_firestore_v1_Value source);
```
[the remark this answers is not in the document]
var-const: Actually, feel free to disregard. `google_firestore_v1_Value` is 24 bytes -- about the threshold where copying might be reasonable.

FirebaseFirestore.podspec
```diff
@@ -89,7 +89,6 @@ Google Cloud Firestore is a NoSQL document database built for automatic scaling,
   abseil_version = '0.20200225.0'
   s.dependency 'abseil/algorithm', abseil_version
   s.dependency 'abseil/base', abseil_version
-  s.dependency 'abseil/container/flat_hash_map', abseil_version
```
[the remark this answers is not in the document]
schmidt-sebastian: I added it back and took advantage of it in the code below. I wonder though if there is a size impact from linking against these libraries.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+using nanopb::FieldsArray;
+using nanopb::FreeNanopbMessage;
+using nanopb::MakeArray;
+using nanopb::MakeBytesArray;
+using nanopb::MakeString;
+using nanopb::MakeStringView;
+
+struct MapEntryKeyCompare {
+  bool operator()(const google_firestore_v1_MapValue_FieldsEntry& entry,
+                  absl::string_view segment) {
```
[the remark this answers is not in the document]
schmidt-sebastian: I hope you have keyboard shortcuts for this kind of comment :)

Firestore/core/src/model/object_value.cc
```diff
@@ -16,22 +16,68 @@
...
+                  const google_firestore_v1_MapValue_FieldsEntry& entry) {
+    return segment < nanopb::MakeStringView(entry.key);
+  }
+};
+
+/**
+ * Finds an entry by key in the provided map value. Returns `nullptr` if the
+ * entry does not exist.
+ */
+static google_firestore_v1_MapValue_FieldsEntry* FindEntry(
```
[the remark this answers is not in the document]
schmidt-sebastian: Makes sense. Removed.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+
+}  // namespace
+
+MutableObjectValue::MutableObjectValue() {
+  value_->which_value_type = google_firestore_v1_Value_map_value_tag;
+  value_->map_value.fields_count = 0;
+  value_->map_value.fields = nullptr;
+}
+
+model::FieldMask MutableObjectValue::ToFieldMask() const {
```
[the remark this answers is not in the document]
schmidt-sebastian: Removed (twice)

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+  return ExtractFieldMask(value_->map_value);
+}
+
+model::FieldMask MutableObjectValue::ExtractFieldMask(
+    const google_firestore_v1_MapValue& value) const {
+  std::set<FieldPath> fields;
+
+  for (size_t i = 0; i < value.fields_count; ++i) {
+    google_firestore_v1_MapValue_FieldsEntry& entry = value.fields[i];
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+}
+
+model::FieldMask MutableObjectValue::ToFieldMask() const {
+  return ExtractFieldMask(value_->map_value);
+}
+
+model::FieldMask MutableObjectValue::ExtractFieldMask(
+    const google_firestore_v1_MapValue& value) const {
+  std::set<FieldPath> fields;
+  for (size_t i = 0; i < value.fields_count; ++i) {
```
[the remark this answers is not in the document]
schmidt-sebastian: The only place I see where this could make sense is after the `nested_map` declaration. I want to keep it together with the if-statement though since it belongs to the same part of the logic.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+    const google_firestore_v1_MapValue& value) const {
+  std::set<FieldPath> fields;
+
+  for (size_t i = 0; i < value.fields_count; ++i) {
+    google_firestore_v1_MapValue_FieldsEntry& entry = value.fields[i];
+    FieldPath current_path({MakeString(entry.key)});
+
+    if (entry.value.which_value_type !=
+        google_firestore_v1_Value_map_value_tag) {
+      fields.insert(current_path);
```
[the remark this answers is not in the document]
schmidt-sebastian: Moved

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+      continue;
+    }
+
+    // Recursively extract the nested map
+    FieldMask nested_mask = ExtractFieldMask(entry.value.map_value);
+    if (nested_mask.begin() == nested_mask.end()) {
+      // Preserve the empty map by adding it to the FieldMask.
+      fields.insert(current_path);
```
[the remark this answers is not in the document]
schmidt-sebastian: Moved

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+void MutableObjectValue::SetAll(const FieldMask& field_mask,
+                                const MutableObjectValue& data) {
+  FieldPath parent;
+
+  std::map<std::string, google_firestore_v1_Value> upserts;
+  std::set<std::string> deletes;
+
+  for (const FieldPath& path : field_mask) {
+    if (!parent.IsImmediateParentOf(path)) {
+      // Insert the accumulated changes at this parent location
```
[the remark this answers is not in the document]
schmidt-sebastian: I usually only put full stops at the end of proper sentence. I don't think this is one (and the rest of the comments in this file aren't either). It is missing a subject, and IMHO adding the subject makes the comments less readable.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+
+      parent = &entry->value;
+    } else {
+      // Create a new map value for the current segment
+      google_firestore_v1_Value new_entry{};
+      new_entry.which_value_type = google_firestore_v1_Value_map_value_tag;
+
+      std::map<std::string, google_firestore_v1_Value> upserts{
+          {segment, new_entry}};
+      ApplyChanges(&(parent->map_value), upserts, /*deletes=*/{});
```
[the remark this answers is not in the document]
schmidt-sebastian: Removed

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+    google_firestore_v1_MapValue* parent,
+    const std::map<std::string, google_firestore_v1_Value>& upserts,
+    const std::set<std::string>& deletes) const {
+  auto source_count = parent->fields_count;
+  auto* source_fields = parent->fields;
+
+  // Compute the size of the map after applying all mutations. The final size is
+  // the number of existing entries, plus the number of new entries
+  // minus the number of deleted entries.
+  size_t target_count =
```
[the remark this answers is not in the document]
schmidt-sebastian: Added `CalculateUnionSize`.

Firestore/core/src/model/object_value.cc
schmidt-sebastian: It doesn't. I moved it and removed it from the header.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+      upserts.size() +
+      std::count_if(source_fields, source_fields + source_count,
+                    [&](const google_firestore_v1_MapValue_FieldsEntry& entry) {
+                      std::string field = MakeString(entry.key);
```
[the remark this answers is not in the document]
schmidt-sebastian: I used a string_view here and added a TODO to consider btrees. I expect the total size of updates to be pretty low though. While a user might update a lot of field values, I would expect that most users only specify a few field paths.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+      MakeArray<google_firestore_v1_MapValue_FieldsEntry>(target_count);
+
+  auto delete_it = deletes.begin();
+  auto upsert_it = upserts.begin();
+
+  // Merge the existing data with the deletes and updates.
+  for (pb_size_t source_index = 0, target_index = 0;
+       target_index < target_count;) {
+    if (source_index < source_count) {
+      std::string key = MakeString(source_fields[source_index].key);
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, we only use it for comparisons.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+       target_index < target_count;) {
+    if (source_index < source_count) {
+      std::string key = nanopb::MakeString(source_fields[source_index].key);
+
+      // Check if the source key is deleted
+      if (delete_it != deletes.end() && *delete_it == key) {
+        nanopb::FreeNanopbMessage(
+            google_firestore_v1_MapValue_FieldsEntry_fields,
+            source_fields + source_index);
+        ++delete_it;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      }
+
+      // Check if the source key is updated by the next upsert
+      if (upsert_it != upserts.end() && upsert_it->first == key) {
+        nanopb::FreeNanopbMessage(google_firestore_v1_Value_fields,
+                                  &source_fields[source_index].value);
+        target_fields[target_index].key = source_fields[source_index].key;
+        target_fields[target_index].value = DeepClone(upsert_it->second);
+        ++upsert_it;
+        ++target_index;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    // Otherwise, insert the next upsert.
+    target_fields[target_index].key = MakeBytesArray(upsert_it->first);
+    target_fields[target_index].value = DeepClone(upsert_it->second);
+    ++upsert_it;
+    ++target_index;
+  }
+
+  free(source_fields);
```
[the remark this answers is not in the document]
schmidt-sebastian: Make sense (and it provides an even a stronger reason to address your next comment).

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,314 @@
...
+
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,140 @@
...
+namespace model {
+
+/** A structured object value stored in Firestore. */
+// TODO(mutabledocuments): Rename to ObjectValue once other ObjectValue class
+// is removed
+class MutableObjectValue {
+ public:
+  MutableObjectValue();
+
+  explicit MutableObjectValue(const google_firestore_v1_Value& value)
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

Firestore/core/src/model/object_value.h
```diff
@@ -0,0 +1,140 @@
...
+   * @param field_mask The field mask that controls which fields to modify.
+   * @param data A MutableObjectValue that contains the field values.
+   */
+  void SetAll(const FieldMask& field_mask, const MutableObjectValue& data);
+
+  /**
+   * Removes the field at the specified path. If there is no field at the
+   * specified path, nothing is changed.
+   *
+   * The path must not be empty.
```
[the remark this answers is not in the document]
schmidt-sebastian: This only applies to Set and Delete (SetAll takes a field mask and Get() supports empty field paths). Added to both.

Firestore/core/test/unit/model/value_util_test.cc
schmidt-sebastian: The idea was more to cause a segfault since both messages would free the same memory regions. Your code snippet is better though as it would catch these errors during runtime.

Firestore/core/test/unit/model/value_util_test.cc
schmidt-sebastian: No, it requires debugging. I don't know how to solve this though (other than by passing in the index of the callsite). I wish there was a proper stracktrace. 

I added a better error message though, which should help a bit.

schmidt-sebastian (commented):
Next round. Thanks for the feedback.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,329 @@
...
+      MakeArray<google_firestore_v1_MapValue_FieldsEntry>(target_count);
+
+  auto delete_it = deletes.begin();
+  auto upsert_it = upserts.begin();
+
+  // Merge the existing data with the deletes and updates
+  for (pb_size_t source_index = 0, target_index = 0;
+       target_index < target_count;) {
+    if (source_index < source_count) {
+      std::string key = MakeString(source_fields[source_index].key);
```
schmidt-sebastian, line 118: Note: Using a string_view here causes an ASAN failure as the key may get freed using FreeNanopbMessage in the loop.

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -134,12 +134,18 @@ class ValueUtilTest : public ::testing::Test {
...
-    // it to verify that all values own their own memory.
-    nanopb::Message<google_firestore_v1_Value> clone1{DeepClone(value)};
-    nanopb::Message<google_firestore_v1_Value> clone2{DeepClone(*clone1)};
-    EXPECT_TRUE(value == *clone1);
-    EXPECT_TRUE(value == *clone2);
+    nanopb::Message<google_firestore_v1_Value> clone1;
+
+    [&] {
+      nanopb::Message<google_firestore_v1_Value> clone2{DeepClone(value)};
+      EXPECT_TRUE(value == *clone2)
```
var-const, line 141: You can use `EXPECT_EQ`. Also, a failure message should contain the line where it happened, so the before/after message is probably not necessary.

Firestore/core/src/model/object_value.cc
```diff
@@ -72,6 +73,97 @@ static google_firestore_v1_MapValue_FieldsEntry* FindEntry(
   return found.first;
 }
 
+size_t CalculateSizeOfUnion(
+    google_firestore_v1_MapValue* map_value,
```
var-const, line 76: Nit: you can pass the map by const reference.

Firestore/core/src/model/object_value.cc
```diff
@@ -80,29 +172,29 @@ MutableObjectValue::MutableObjectValue() {
...
 
   for (size_t i = 0; i < value.fields_count; ++i) {
-    google_firestore_v1_MapValue_FieldsEntry& entry = value.fields[i];
+    const google_firestore_v1_MapValue_FieldsEntry& entry = value.fields[i];
     FieldPath current_path({MakeString(entry.key)});
 
     if (entry.value.which_value_type !=
         google_firestore_v1_Value_map_value_tag) {
-      fields.insert(current_path);
+      fields.insert(std::move((current_path)));
```
var-const, line 188: Nit: unnecessary parentheses around `current_path`.

FirebaseFirestore.podspec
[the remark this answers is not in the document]
var-const: That's a good question. A large part of both `std::unordered_map` and Abseil maps is a template, so those only get instantiated on demand (and I won't be surprised if Abseil ends up in less code, but that's just a guess). However, I'm sure Abseil also contains some non-templated code (e.g. for hash calculation) which would add some overhead over the standard library. I doubt it's significant, but we can check later.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,329 @@
...
+      deletes.insert(path.last_segment());
+    }
+  }
+
+  google_firestore_v1_MapValue* parent_map = ParentMap(parent);
+  ApplyChanges(parent_map, upserts, deletes);
+}
+
+void MutableObjectValue::Delete(const FieldPath& path) {
+  HARD_ASSERT(!path.empty(), "Cannot set field for empty path on ObjectValue");
```
var-const, line 266: Nit: `s/set field/delete field/`.

Firestore/core/test/unit/model/value_util_test.cc
var-const: Can you check again after replacing `EXPECT_TRUE` with `EXPECT_EQ`? One possible approach is to add [`SCOPED_TRACE`](https://github.com/google/googletest/blob/master/docs/advanced.md#using-assertions-in-sub-routines) before every invocation of this helper, but please check with `EXPECT_EQ` first to see if it's necessary. Another approach, should `EXPECT_EQ` not work out, is to log the stringified form of both values being compared upon failure.

Firestore/core/src/model/object_value.cc
```diff
@@ -0,0 +1,287 @@
...
+      malloc(target_count * sizeof(_google_firestore_v1_MapValue_FieldsEntry)));
+
+  auto delete_it = deletes.begin();
+  auto upsert_it = upserts.begin();
+
+  // Merge the existing data with the deletes and updates.
+  for (pb_size_t target_index = 0, source_index = 0;
+       target_index < target_count;) {
+    if (source_index < source_count) {
+      std::string key = nanopb::MakeString(source_fields[source_index].key);
```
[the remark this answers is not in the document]
var-const: Bump.

Firestore/core/src/model/object_value.cc
var-const: Bump (fine to leave as a TODO).

Firestore/core/src/model/object_value.cc
var-const: Hmm, but `key` doesn't seem to be used after `FreeNanopbMessage`? AFAICS, the memory can only be freed _after_ a comparison and it causes the loop to continue.

var-const (changes requested):
Close to LGTM. Please see a couple of comments I bumped from the last time.

Firestore/core/test/unit/model/value_util_test.cc
schmidt-sebastian: I wasn't able to get this to work:

```
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/cmake-build-debug/external/src/googletest/googletest/include/gtest/gtest.h:1527:11: error: invalid operands to binary expression ('const firebase::firestore::_google_firestore_v1_Value' and 'const firebase::firestore::_google_firestore_v1_Value')
  if (lhs == rhs) {
      ~~~ ^  ~~~
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/cmake-build-debug/external/src/googletest/googletest/include/gtest/gtest.h:1554:12: note: in instantiation of function template specialization 'testing::internal::CmpHelperEQ<firebase::firestore::_google_firestore_v1_Value, firebase::firestore::_google_firestore_v1_Value>' requested here
    return CmpHelperEQ(lhs_expression, rhs_expression, lhs, rhs);
           ^
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/Firestore/core/test/unit/model/value_util_test.cc:141:7: note: in instantiation of function template specialization 'testing::internal::EqHelper::Compare<firebase::firestore::_google_firestore_v1_Value, firebase::firestore::_google_firestore_v1_Value, nullptr>' requested here
      EXPECT_EQ(value, *clone2)
      ^
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/cmake-build-debug/external/src/googletest/googletest/include/gtest/gtest.h:2028:54: note: expanded from macro 'EXPECT_EQ'
  EXPECT_PRED_FORMAT2(::testing::internal::EqHelper::Compare, val1, val2)
                                                     ^
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/Firestore/core/include/firebase/firestore/geo_point.h:113:13: note: candidate function not viable: no known conversion from 'const firebase::firestore::_google_firestore_v1_Value' to 'const firebase::firestore::GeoPoint' for 1st argument
inline bool operator==(const GeoPoint& lhs, const GeoPoint& rhs) {
            ^
/Users/mrschmidt/GitHub/firebase/firebase-ios-sdk/cmake-build-debug/external/src/googletest/googletest/include/gtest/gtest.h:1518:13: note: candidate function not viable: no known conversion from 'const firebase::firestore::_google_firestore_v1_Value' to 'testing::internal::faketype' for 1st argument
inline bool operator==(faketype, faketype) { return true; }
```

I tried to use EXPECT_EQ in my earlier attempts at writing these tests and for some reason the `==` operator is not recognized.

Firestore/core/src/model/object_value.cc
schmidt-sebastian: Fixed

Firestore/core/src/model/object_value.cc
schmidt-sebastian: Done

Firestore/core/src/model/object_value.cc
schmidt-sebastian: Fixed

Firestore/core/test/unit/model/value_util_test.cc
var-const: Ok, what happens is that due to the way `EXPECT_EQ` is implemented, the `operator==` has to be in the same namespace as at least one of the classes being compared. In this case, `google_firestore_v1_Value` is in the `firebase::firestore` namespace, but the comparison operator is in `firebase::firestore::model`. It's probably a good idea to move `operator==` one namespace level up.

Firestore/core/src/nanopb/fields_array.h
```diff
@@ -23,6 +23,7 @@
 #include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
 #include "Firestore/Protos/nanopb/google/firestore/v1/firestore.nanopb.h"
 #include "Firestore/Protos/nanopb/google/type/latlng.nanopb.h"
+#include "Firestore/core/src/nanopb/message.h"
```
var-const, line 26: This include should no longer be necessary.

var-const (approved):
Thanks for addressing all the feedback! LGTM (with one small comment about includes).

PR #7722, opened 2021-03-15
Mutable Documents

schmidt-sebastian:
This ports MutableDocument from Android: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java

### DIFF
--- a/Firestore/core/src/model/mutable_document.cc
+++ b/Firestore/core/src/model/mutable_document.cc
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/model/mutable_document.h"
+
+#include <ostream>
+#include <utility>
+
+namespace firebase {
+namespace firestore {
+namespace model {
+
+/* static */
+MutableDocument MutableDocument::InvalidDocument(
+    const firebase::firestore::model::DocumentKey& document_key) {
+  return MutableDocument{document_key, DocumentType::kInvalid,
+                         SnapshotVersion::None(), ObjectValue{},
+                         DocumentState::kSynced};
+}
+
+/* static */
+MutableDocument MutableDocument::FoundDocument(
+    const firebase::firestore::model::DocumentKey& document_key,
+    const firebase::firestore::model::SnapshotVersion& version,
+    firebase::firestore::model::ObjectValue value) {
+  return std::move(InvalidDocument(document_key)
+                       .ConvertToFoundDocument(version, std::move(value)));
+}
+
+/* static */
+MutableDocument MutableDocument::UnknownDocument(
+    const firebase::firestore::model::DocumentKey& document_key,
+    const firebase::firestore::model::SnapshotVersion& version) {
+  return std::move(
+      InvalidDocument(document_key).ConvertToUnknownDocument(version));
+}
+
+/* static */
+MutableDocument MutableDocument::NoDocument(
+    const firebase::firestore::model::DocumentKey& document_key,
+    const firebase::firestore::model::SnapshotVersion& version) {
+  return std::move(InvalidDocument(document_key).ConvertToNoDocument(version));
+}
+
+MutableDocument& MutableDocument::ConvertToUnknownDocument(
+    const firebase::firestore::model::SnapshotVersion& version) {
+  version_ = version;
+  document_type_ = DocumentType::kUnknownDocument;
+  value_ = {};
+  document_state_ = DocumentState::kHasCommittedMutations;
+  return *this;
+}
+
+MutableDocument& MutableDocument::ConvertToFoundDocument(
+    const firebase::firestore::model::SnapshotVersion& version,
+    firebase::firestore::model::ObjectValue value) {
+  version_ = version;
+  document_type_ = DocumentType::kFoundDocument;
+  value_ = std::move(value);
+  document_state_ = DocumentState::kSynced;
+  return *this;
+}
+
+MutableDocument& MutableDocument::ConvertToNoDocument(
+    const firebase::firestore::model::SnapshotVersion& version) {
+  version_ = version;
+  document_type_ = DocumentType::kNoDocument;
+  value_ = {};
+  document_state_ = DocumentState::kSynced;
+  return *this;
+}
+
+MutableDocument& MutableDocument::SetHasCommittedMutations() {
+  document_state_ = DocumentState::kHasCommittedMutations;
+  return *this;
+}
+
+MutableDocument& MutableDocument::SetHasLocalMutations() {
+  document_state_ = DocumentState::kHasLocalMutations;
+  return *this;
+}
+
+bool operator==(const MutableDocument& lhs, const MutableDocument& rhs) {
+  return lhs.key_ == rhs.key_ && lhs.document_type_ == rhs.document_type_ &&
+         lhs.version_ == rhs.version_ && lhs.value_ == rhs.value_ &&
+         lhs.document_state_ == rhs.document_state_;
+}
+
+std::ostream& operator<<(std::ostream& os,
+                         MutableDocument::DocumentState state) {
+  switch (state) {
+    case MutableDocument::DocumentState::kHasCommittedMutations:
+      return os << "kHasCommittedMutations";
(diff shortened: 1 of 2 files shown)


Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  bool is_unknown_document() const {
+    return document_type_ == DocumentType ::kUnknownDocument;
+  }
+
+  friend bool operator==(const MutableDocument& lhs,
+                         const MutableDocument& rhs);
+
+  friend std::ostream& operator<<(std::ostream& os, const MutableDocument& doc);
```
wu-hui, line 196: Can we have a `ToString`, and implement `<<` using that? That is what `Target` does. `<<` is not very composable.

Firestore/core/src/model/mutable_document.cc
```diff
@@ -0,0 +1,137 @@
...
+#include "Firestore/core/src/model/mutable_document.h"
+
+namespace firebase {
+namespace firestore {
+namespace model {
+
+/* static */
+MutableDocument MutableDocument::InvalidDocument(
+    const firebase::firestore::model::DocumentKey& document_key) {
+  return MutableDocument{document_key, DocumentType::kInvalid,
```
wu-hui, line 26: I *think* it is preferred to remove `MutableDocument` here, that is simply `return {...}` should work.

Firestore/core/src/model/mutable_document.cc
```diff
@@ -0,0 +1,137 @@
...
+                         SnapshotVersion::None(), ObjectValue{},
+                         DocumentState::kSynced};
+}
+
+/* static */
+MutableDocument MutableDocument::FoundDocument(
+    const firebase::firestore::model::DocumentKey& document_key,
+    const firebase::firestore::model::SnapshotVersion& version,
+    firebase::firestore::model::ObjectValue value) {
+  return std::move(InvalidDocument(document_key)
```
wu-hui, line 36: You can remove the `std::move` on return statements, applied elsewhere too. See https://en.cppreference.com/w/cpp/language/copy_elision

Firestore/core/src/model/mutable_document.cc
```diff
@@ -0,0 +1,137 @@
...
+
+#include "Firestore/core/src/model/mutable_document.h"
+
+namespace firebase {
+namespace firestore {
+namespace model {
+
+/* static */
+MutableDocument MutableDocument::InvalidDocument(
+    const firebase::firestore::model::DocumentKey& document_key) {
```
wu-hui, line 25: I think you can remove all of the namespace prefixes here, they are from the same `model`.

Firestore/core/src/model/mutable_document.cc
```diff
@@ -0,0 +1,137 @@
...
+}
+
+MutableDocument& MutableDocument::SetHasLocalMutations() {
+  document_state_ = DocumentState::kHasLocalMutations;
+  return *this;
+}
+
+bool operator==(const MutableDocument& lhs, const MutableDocument& rhs) {
+  return lhs.key_ == rhs.key_ && lhs.document_type_ == rhs.document_type_ &&
+         lhs.version_ == rhs.version_ && lhs.value_ == rhs.value_ &&
```
wu-hui, line 95: Ultranit: value can be the last check, they can be potentially big.

wu-hui (changes requested):
We should also have some tests for this. Simply create documents, perform mutations and check they are in expected state should be good enough.

Firestore/core/src/model/mutable_document.cc
schmidt-sebastian: Fixed. Sorry for not catching this myself.

Firestore/core/src/model/mutable_document.cc
schmidt-sebastian: You are the expert :) Removed.

Firestore/core/src/model/mutable_document.cc
schmidt-sebastian: This doesn't seem to work. The compiler complains here since there is no copy constructor. I suspect that by default it tries to use the copy constructor and it wants to be free to do so, even if it end up using copy elision later on.

Firestore/core/src/model/mutable_document.cc
schmidt-sebastian: Makes sense.

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+
+#include <ostream>
+#include <utility>
+
+#include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
+#include "Firestore/core/src/model/document_key.h"
+#include "Firestore/core/src/model/field_value.h"
+#include "Firestore/core/src/model/snapshot_version.h"
+
+// TODO(mutabledocuments): We might want to call this class Document and replace
```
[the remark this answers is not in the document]
schmidt-sebastian: These PRs are against a feature branch. I am planning on grepping for `mutabledocuments` before merging. I could use a bug here (and probably should for my next project), but I have already used this pattern in the branch.

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Represents an existing document whose data is unknown (e.g. a document
+     * that was updated without a known base document).
+     */
+    kUnknownDocument
+  };
+
+  /** Describes the `hasPendingWrites` state of a document. */
+  enum class DocumentState {
+    /** Local mutations applied via the mutation queue. Document is potentially
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+       potentially inconsistent. */
+    kHasCommittedMutations,
+    /** No mutations applied. Document was sent to us by Watch. */
+    kSynced
+  };
+
+ public:
+  // MutableDocument contain Proto data that cannot be implicitly created or
+  // copied.
+  MutableDocument() = delete;
```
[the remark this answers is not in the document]
schmidt-sebastian: Leaving at is since the main constructor is private.

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+  MutableDocument(const MutableDocument&) = delete;
+  MutableDocument& operator=(const MutableDocument&) = delete;
```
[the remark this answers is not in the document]
schmidt-sebastian: This is based on https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/nanopb/message.h

MutableDocuments contain ObjectValues, and they store Protos based in Messages. Protos cannot be copied cheaply (they have to be deep cloned), but we can transfer ownership using the move operator.

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+
+  MutableDocument(MutableDocument&& other) noexcept
+      : key_{std::move(other.key_)},
+        document_type_{other.document_type_},
+        version_{other.version_},
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (also fixed below).

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+        value_{std::move(other.value_)},
+        document_state_{other.document_state_} {
+  }
+
+  /**
+   * Creates a document with no known version or data, but which can serve as
```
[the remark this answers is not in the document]
schmidt-sebastian: Split into two sentences.

Firestore/core/src/model/mutable_document.h
```diff
@@ -0,0 +1,232 @@
...
+   * base document for mutations.
+   */
+  static MutableDocument InvalidDocument(const DocumentKey& document_key);
+
+  /** Creates a new document that is known to exist with the given data at the
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed here and once more below.

Firestore/core/src/model/mutable_document.h
schmidt-sebastian: Done. I ended up using a `stringstream` so that I can keep the operators for DocumentState and DocumentType (which might be handy later on).

schmidt-sebastian (commented):
Thanks for the review!

I didn't add tests since this will be exhaustively tested by the LocalStoreTests once I plug this into the rest of the system. The existing document types follow the same testing strategy :)

Firestore/core/src/model/mutable_document.cc
wu-hui: Interesting, I learnt something!

Firestore/core/src/model/mutable_document.h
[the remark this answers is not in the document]
wu-hui: SG.

Firestore/core/src/model/mutable_document.h
wu-hui: good call.

wu-hui (approved):
LGTM!

PR #7725, opened 2021-03-16
Throw exception when projectID is missing from FIROptions

karenyz:
If `projectID` is missing the fetch request can't succeed, so we can fail the request before sending it to the server. This is causing some downstream issues (tracked in b/167621250)

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
@@ -218,6 +218,19 @@ static const NSInteger sFIRErrorCodeConfigFailed = -114;
                                                     NSLocalizedDescriptionKey : errorDescription
                                                   }]];
   }
+  // Check for projectID
+  if (!_options.projectID) {
+    NSString *errorDescription = @"Failed to get `projectID`";
+    FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000070", @"%@",
+                [NSString stringWithFormat:@"%@", errorDescription]);
+    NSError *error = [NSError errorWithDomain:FIRRemoteConfigErrorDomain
+                                         code:FIRRemoteConfigErrorInternalError
+                                     userInfo:@{NSLocalizedDescriptionKey : errorDescription}];
+    self->_settings.isFetchInProgress = NO;
+    return [self reportCompletionOnHandler:completionHandler
+                                withStatus:FIRRemoteConfigFetchStatusFailure
+                                 withError:error];
+  }
 
   __weak RCNConfigFetch *weakSelf = self;
   FIRInstallationsTokenHandler installationsTokenHandler = ^(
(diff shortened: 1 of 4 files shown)


FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -218,6 +218,19 @@ - (void)refreshInstallationsTokenWithCompletionHandler:
                                                     NSLocalizedDescriptionKey : errorDescription
                                                   }]];
   }
+  // Check for projectID
+  if (!_options.projectID) {
```
maksymmalyhin, line 222: Why the conditions here and in FIRRemoteConfigComponent.m are different?

paulb777:
FIS, an RC dependency, should already be checking the FIROptions, so it's not clear what this is changing. Also, this might be considered a breaking change, so assigning to @ryanwilson to approve from an API perspective.

maksymmalyhin (approved):
LGTM.

Note, the change could be considered as a breaking, but most like is not. RC has a dependency on FIS which already [throws on the missing project ID](https://github.com/firebase/firebase-ios-sdk/blob/1f2f5f617e6b964b8377d9f2dddf1d5225e9efe2/FirebaseInstallations/Source/Library/FIRInstallations.m#L130), so the exception won't be new for existing RC clients.

Please wait for @ryanwilson for the final approval on the API change.

ryanwilson:
> LGTM.
> 
> Note, the change could be considered as a breaking, but most like is not. RC has a dependency on FIS which already [throws on the missing project ID](https://github.com/firebase/firebase-ios-sdk/blob/1f2f5f617e6b964b8377d9f2dddf1d5225e9efe2/FirebaseInstallations/Source/Library/FIRInstallations.m#L130), so the exception won't be new for existing RC clients.
> 
> Please wait for @ryanwilson for the final approval on the API change.

Looking at the internal bug though, it seems like they're getting through even with the dependency on FIS so I'm concerned this is going to break someone. Do we know if that's an older version sending the data, maybe pre-FIS? @karenyz

maksymmalyhin:
> Looking at the internal bug though, it seems like they're getting through even with the dependency on FIS so I'm concerned this is going to break someone. Do we know if that's an older version sending the data, maybe pre-FIS? @karenyz

Note, FIS started throwing the exception on the missing project ID since last major version bump.

karenyz:
I was hoping to fix-forward so we would stop sending invalid URLs moving forward, but since FIS is already throwing an exception it seems like this change doesn't make much of a difference (maybe makes it clearer for RC users?)

@ryanwilson yeah the malformed requests are from older versions, so this change wouldn't affect those requests

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -516,16 +529,8 @@ - (void)fetchWithUserProperties:(NSDictionary *)userProperties
 - (NSString *)constructServerURL {
   NSString *serverURLStr = [[NSString alloc] initWithString:kServerURLDomain];
   serverURLStr = [serverURLStr stringByAppendingString:kServerURLVersion];
-
-  if (_options.projectID) {
-    serverURLStr = [serverURLStr stringByAppendingString:kServerURLProjects];
-    serverURLStr = [serverURLStr stringByAppendingString:_options.projectID];
-  } else {
-    FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000070",
-                @"Missing `projectID` from `FirebaseOptions`, please ensure the configured "
```
erikeldridge, line 525: I've recommended just asserting project ID as a required input in FIRRemoteConfigComponent.m. Given that, and since project ID is the only thing we're checking here, I think we can simplify this class by removing this logging.

FirebaseRemoteConfig/Tests/Unit/FIRRemoteConfigComponentTest.m
```diff
@@ -167,6 +167,35 @@ - (void)testThrowsWithNilGCMSenderID {
   XCTAssertThrows([component remoteConfigForNamespace:@"some_namespace"]);
 }
 
+- (void)testThrowsWithEmptyProjectID {
```
erikeldridge, line 170: The fact we're only asserting FIRRemoteConfigComponentTest.m throws supports the idea that validating inputs there is sufficient.

erikeldridge (commented):
We chatted in person, and this change may not be necessary, but if we do make it, I've added comments voting in favor of just asserting required configuration in FIRRemoteConfigComponent.m.

ryanwilson:
LGTM from an API standpoint since this is a no-op

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
karenyz: removing the check here since we already check in `FIRRemoteConfigComponent`

erikeldridge (approved):
Shipit! 🚢

PR #7729, opened 2021-03-16
Enable Analytics patch updates with Firebase pod update

paulb777:
This makes Analytics consistent with the other libraries by enabling self-contained CocoaPods patch releases.

#no-changelog


ryanwilson (commented):
LGTM, deferring to @htcgh for approval.

This would prevent us from having to ship a new Firebase pod + zip build any time we want to do a patch update like 7.8.1.

ryanwilson (approved):

PR #7731, opened 2021-03-16
Add test_spec parameters for podspecs with different test_spec name.

granluo:
Update the tools to be applied to podspecs with unit tests with different name, e.g. [GDT podpsecs](https://github.com/google/GoogleDataTransport/blob/main/GoogleDataTransport.podspec#L91)

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -57,7 +57,7 @@ jobs:
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: Build and test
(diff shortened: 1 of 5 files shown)


.github/workflows/test_coverage.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: Build and test
```
paulb777, line 59: nit: alphabetize Database and Functions

paulb777 (approved):

PR #7733, opened 2021-03-17
Bundles Fix: OrderBy should be optional, not required

wu-hui:


Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -924,6 +926,38 @@ TEST_F(BundleSerializerTest, DecodesOrderBys) {
...
 }
 
+// By default, the queries used for testing in this file always have default
+// OrderBy ("__name__") generated. We need to explicitly remove that for this
+// test.
+TEST_F(BundleSerializerTest, DecodeMissingOrderBysWorks) {
+  // This is `NamedQueryToJson(testutil::Query("bundles/docs/colls"))` with
+  // orderBy field manually removed.
+  auto json_string = R"|(
+{"name":"query-1",
```
var-const, line 936: Nit: add a newline between the opening curly brace and `"name":`. Feel free to do in a follow-up if you don't want to retest this PR.

var-const (approved):

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
wu-hui: Done.

PR #7717, opened 2021-03-15
Release Firestore bundles

wu-hui:
The two commits that have not been reviewed:

c50051f07bb436cb6245f19e235d1819eeed4851 Release Note
9c56fc81906cbddcfe8fee1c39718f08c03176a1 API amendment

PR #7735, opened 2021-03-17
Add missing new header to Firestore podspec

paulb777:
Fix `pod spec lint` failure introduced in #7717 and caught in nightly #7726.

`- NOTE | xcodebuild: FirebaseFirestore/Firestore/core/src/bundle/bundle_serializer.h:33:10: fatal error: 'Firestore/third_party/nlohmann_json/json.hpp' file not found`

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -5,6 +5,8 @@ on:
     paths:
     - '.github/workflows/prerelease.yml'
     - 'Gemfile'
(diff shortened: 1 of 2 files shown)


var-const (approved):

paulb777:
The test still failed since it looks like the prerelease workflow uses master and not the PR from the podspec.  I'll merge, since it should fix, and the release process will validate.

PR #7740, opened 2021-03-17
Fix pod lib lint for watchOS + add CI test

ryanwilson:
#no-changelog

### DIFF
--- a/FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
+++ b/FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
@@ -327,11 +327,15 @@
 }
 
 - (void)receiveWebSocketData {
-    __weak FWebSocketConnection *weakSelf = self;
+    __weak __autotype *weakSelf = self;
     [self.webSocketTask receiveMessageWithCompletionHandler:^(
                             NSURLSessionWebSocketMessage *_Nullable message,
                             NSError *_Nullable error) {
-      FWebSocketConnection *strongSelf = weakSelf;
+      __autotype *strongSelf = weakSelf;
+      if (strongSelf == nil ) {
+        return;
+      }
+
       if (message) {
           [strongSelf handleIncomingFrame:message.string];
       } else if (error && !strongSelf->isClosed) {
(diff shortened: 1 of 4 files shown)


FirebaseDatabase/Tests/Helpers/FIRFakeApp.m
```diff
@@ -47,23 +47,6 @@ @interface FIRComponentContainer (TestInternal)
 @property(nonatomic, strong) NSMutableDictionary<NSString *, FIRComponentCreationBlock> *components;
 @end
 
-@interface FIRComponentContainer (TestInternalImplementations)
```
ryanwilson, line 50: These are already defined in FIRComponentTestUtilities.m in SharedTestUtilities

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -327,20 +327,22 @@ - (void)URLSession:(NSURLSession *)session
 }
 
 - (void)receiveWebSocketData {
+    __weak FWebSocketConnection *weakSelf = self;
```
maksymmalyhin, line 330: I find `__weak __autotype weakSelf = self` as it is more compact and is better handled by Xcode refactoring (no need to be changed), but it's definitely up to you.

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -327,20 +327,22 @@ - (void)URLSession:(NSURLSession *)session
...
     [self.webSocketTask receiveMessageWithCompletionHandler:^(
                             NSURLSessionWebSocketMessage *_Nullable message,
                             NSError *_Nullable error) {
+      FWebSocketConnection *strongSelf = weakSelf;
```
maksymmalyhin, line 334: nit: I think the common pattern to exit early if `strongSelf == nil`, but it'll be fine if we don't in this particular case.

maksymmalyhin (approved):

FirebaseDatabase/Sources/Realtime/FWebSocketConnection.m
```diff
@@ -327,20 +327,26 @@ - (void)URLSession:(NSURLSession *)session
 }
 
 - (void)receiveWebSocketData {
+    __weak __autotype *weakSelf = self;
```
maksymmalyhin, line 330: I think there is a typo here and it should be `__weak __auto_type weakSelf = self;`. I can post a commit if it's already late for you.

PR #7741, opened 2021-03-17
Resolve new SPM package install warning

paulb777:
Also update scheme for recent new targets added.


maksymmalyhin (approved):
👍

paulb777:
cc: @var-const @wu-hui

PR #7748, opened 2021-03-18
Consistent podspec minimum version management

paulb777:
Background is that we need a different set of platforms for libraries and tests when the library supports watchOS since tests aren't supported with watchOS and CocoaPods even with Xcode 12.5.

Current plan is for watchOS CI to be SPM only


paulb777:
Confirmed that CI failures are only because of recent Analytics breakage

ryanwilson (approved):
Nice cleanup!

PR #7754, opened 2021-03-19
Fix data races detected by TSAN: HostConfig in grpc_connection.cc and g_log_level in log_stdio.cc.

dconeybe:
#no-changelog


Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -71,10 +72,40 @@ std::shared_ptr<grpc::ChannelCredentials> CreateSslCredentials(
   return grpc::SslCredentials(options);
 }
 
-struct HostConfig {
-  util::Path certificate_path;
-  std::string target_name;
-  bool use_insecure_channel = false;
+class HostConfig {
```
var-const, line 75: To reduce verbosity, consider using an approach like:
```cpp
class HostConfig {
 public:
 struct Data {
    util::Path certificate_path;
    std::string target_name;
    bool use_insecure_channel = false;
  };

  // `Modify` is a strawman name.
  template <typename CallbackT>
  void Modify(const CallbackT& callback) {
    callback(data_);
  }
  // You could also have a const function that invokes the callback
  // with a `const Data&`, but that's probably overkill.

 private:
  std::mutex mutex_;
  Data data_;
};

host_config.Modify([&](HostConfig::Data& data) {
  data.certificate_path = certificate_path;
  // ...
});
```

Firestore/core/src/util/log_stdio.cc
```diff
@@ -26,20 +27,20 @@ namespace firestore {
...
 namespace {
 
-LogLevel g_log_level = kLogLevelNotice;
+std::atomic<LogLevel> g_log_level(kLogLevelNotice);
 
 }  // namespace
 
 void LogSetLevel(LogLevel level) {
-  g_log_level = level;
+  g_log_level.store(level);
```
var-const, line 35: Any reason not to use the overloaded operators? From a glance, it seems like other than `s/LogLevel/std::atomic<LogLevel>/`, the code in this file could be left as it was.

var-const (changes requested):

Firestore/core/src/remote/grpc_connection.cc
dconeybe: I can definitely make that change; however, how are the data members *read* with this new design? Would there be a method that returns a copy of `data_`? Returning a *const reference to* `data_` would still introduce a race condition because the caller would read members of `data_` without holding the lock. Maybe I'm just missing the obvious though!

Firestore/core/src/util/log_stdio.cc
dconeybe: I chose to use the methods instead of the overloaded operators to highlight that they are atomic operations. I watched a CppCon video (I forget which one exactly) about atomics and the speaker recommended doing it this way. Since I hadn't come across any dissenting opinions, I just went with it :) I'm happy to do whatever feels more conventional in our code base. Please advise.

Firestore/core/src/remote/grpc_connection.cc
var-const: I presumed something like:
```cpp
std::string certificate_path;
host_config.Get([&](const HostConfig::Data& data) {
  certificate_path = data.certificate_path;
});
```
(`Modify` as a function name clearly doesn't work here)

Returning a copy would also work though it seems like the callback-invoking function should cover all use cases.

Firestore/core/src/util/log_stdio.cc
var-const: As long as it is a deliberate choice, I think that's fine. Our existing usage somewhat favors the overloaded operators, but there's not much usage. Personally, I think classes should behave like built-in types syntax-wise where appropriate, but it's really a matter of personal preference.

dconeybe: Done. I've changed the load()/store() calls to use the overloaded operators, since I am ambivalent towards the two options.

Firestore/core/src/util/log_stdio.cc
```diff
@@ -32,15 +32,15 @@ std::atomic<LogLevel> g_log_level(kLogLevelNotice);
 }  // namespace
 
 void LogSetLevel(LogLevel level) {
-  g_log_level.store(level);
+  g_log_level = level;
```
var-const: Consider also using assignment-like syntax for initialization (at line 30 above).

Firestore/core/src/remote/grpc_connection.cc
```diff
@@ -71,10 +72,40 @@ std::shared_ptr<grpc::ChannelCredentials> CreateSslCredentials(
...
+  using Guard = std::lock_guard<std::mutex>;
+
+ public:
+  void certificate_path(const util::Path& new_value) {
```
var-const, line 79: Nit: there is a using declaration for `Path` already -- no need to specify the namespace.

var-const, line 79: Nit: setters should be named like `set_foo`.

var-const (approved):

Firestore/core/src/remote/grpc_connection.cc
dconeybe: Done

dconeybe: Done.

PR #7756, opened 2021-03-19
Import Firebase Swift APIs into FirebasePodTest source.

ncooke3:
Follow up to [#7746](https://github.com/firebase/firebase-ios-sdk/pull/7746). Imported Firebase Swift APIs to `FirebasePodTest`'s `AppDelegate`.

#no-changelog


paulb777 (approved):
Thanks! LGTM on green

PR #7769, opened 2021-03-23
Update release test to generate testing repo with Cocoapods tags.

granluo:


granluo:
Release testing will add release tag, e.g. Cocoapods-7.9.0, to podspecs without updating tag commit in the sdk repo.

.github/workflows/release.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
-         BOT_TOKEN="${ossbotaccess}" scripts/release_testing_setup.sh nightly_testing
+         BOT_TOKEN="${ossbotaccess}" scripts/release_testing_setup.sh release_testing
     - name: Update SpecsTesting repo
       run: |
         botaccess=`cat bot-access.txt`
         cd scripts/create_spec_repo/
         swift build
         pod repo add --silent "${local_repo}" https://"$botaccess"@github.com/FirebasePrivate/SpecsTesting.git
-        BOT_TOKEN="${botaccess}"  .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}"
+        BOT_TOKEN="${botaccess}"  .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}" --pod-sources 'https://${BOT_TOKEN}@github.com/FirebasePrivate/SpecsTesting' --pod-sources "https://github.com/firebase/SpecsStaging.git" --pod-sources "https://cdn.cocoapods.org/"
```
paulb777, line 44: Release should not include SpecsTesting. Everything it needs should be in SpecsStaging.

.github/workflows/release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         cd scripts/create_spec_repo/
         swift build
         pod repo add --silent "${local_repo}" https://"$botaccess"@github.com/FirebasePrivate/SpecsTesting.git
-        BOT_TOKEN="${botaccess}"  .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}" --pod-sources 'https://${BOT_TOKEN}@github.com/FirebasePrivate/SpecsTesting' --pod-sources "https://github.com/firebase/SpecsStaging.git" --pod-sources "https://cdn.cocoapods.org/"
+        BOT_TOKEN="${botaccess}"  .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}" --pod-sources "https://github.com/firebase/SpecsStaging.git" --pod-sources "https://cdn.cocoapods.org/"
```
paulb777, line 44: I think my comment was wrong. I was confusing SpecsTesting with SpecsDev.  I think we need SpecsTesting so we can incrementally build it before updating SpecsStaging.

I think you should change back.

Sorry about my confusion.

paulb777:
For another PR: Now that everything we test and stage on GitHub is in the open, there is no longer the need for private repos and key management.

paulb777 (approved):

.github/workflows/release.yml
granluo: No worries, the source should be easy to update once we wanna have any update in the future.

granluo:
> For another PR: Now that everything we test and stage on GitHub is in the open, there is no longer the need for private repos and key management.

Do you refer to the private testing repo?
(for prerelease workflow)https://github.com/FirebasePrivate/SpecsReleasing
(for release workflow)https://github.com/FirebasePrivate/SpecsTesting 
One of reasons to make it private is a public repo requires a project in the repo and to follow go/releasing, and testing repos do not need to release. Do we need to change the repos to be public? One pro I can think of to make them public is this can enable release/prerelease or presubmit(in the future) workflow for external developers.

paulb777:
I was mostly thinking about reproducibility and being able to run the same scripts locally without needing to remove and adjust for the key, but allowing external people to run the workflows is another advantage. This isn't a huge priority and go/releasing isn't too big a hurdle, but I just wanted to raise it while I was thinking about it.

PR #7776, opened 2021-03-24
Add more token tests

chliangGoogle:
Move the following token class tests from IID to Messaging
TokenInfo
TokenOperation
TokenStore
Utilities
#no-changelog


chliangGoogle:
@paulb777 I think chen/fm-master is corrupted, and chen/token-add-test has the right latest version of code.

chliangGoogle:
Looks like github UI now reflected that it's fixed by itself magically.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenInfoTest.m
```diff
@@ -0,0 +1,199 @@
...
+  [[NSUserDefaults standardUserDefaults] setObject:FIRMessagingCurrentLocale()
+                                            forKey:kFIRMessagingInstanceIDUserDefaultsKeyLocale];
+}
+
+- (void)tearDown {
+  [self.mockOptions stopMocking];
+  [super tearDown];
+}
+
+- (void)testTokenInfoCreationWithInvalidArchive {
```
maksymmalyhin, line 75: I'm not sure what we are actually testing by this test? Looks like only behaviour of deprecated `[NSKeyedUnarchiver unarchiveObjectWithData:badData]` method. I would consider removing it.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenInfoTest.m
```diff
@@ -0,0 +1,199 @@
...
+  XCTAssertNil(info);
+}
+
+// Test that archiving a FIRMessagingTokenInfo object and restoring it from the archive
+// yields the same values for all the fields.
+- (void)testTokenInfoEncodingAndDecoding {
+  FIRMessagingTokenInfo *info = self.validTokenInfo;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+  NSData *archive = [NSKeyedArchiver archivedDataWithRootObject:info];
```
maksymmalyhin, line 95: optional: I would suggest using [GULSecureCoding](https://github.com/google/GoogleUtilities/blob/1e0f385577a122bd935881cdcb24dbd1aa642663/GoogleUtilities/Environment/Public/GoogleUtilities/GULSecureCoding.h#L28) to avoid using deprecated API when unnecessary (here and at similar places).

maksymmalyhin (approved):
LGTM on CI green, but I didn't review all the tests in too much details assuming that these are mostly existing tests moved to FCM. Please let me know if you need more detailed review for some part of the changes.

paulb777 (approved):

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenInfoTest.m
chliangGoogle: Done.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenInfoTest.m
chliangGoogle: that's on my todo list next because we did introduce the NSCoding from IID to Messaging that we should migrate to NSSecureCoding.

Issue #7771, opened 2021-03-23
Data Race / Crash in GDTCCTUploadOperation


cltnschlosser:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.4
  * Firebase SDK version: 7.8.1
  * Installation method: `CocoaPods`
  * Firebase Component: Crashlytics

#### Steps to reproduce:

Data Race / associated crash happening while running unit tests

#### Relevant Code:

```
Crashed Thread:        2  Dispatch queue: com.google.GDTCCTUploader

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
==================
WARNING: ThreadSanitizer: data race (pid=63680)
  Write of size 1 at 0x7b440003acf9 by thread T2:
    #0 -[GDTCCTUploadOperation finishOperation] <null>:2 (<App>:x86_64+0x10321265b)
    #1 __53-[GDTCCTUploadOperation uploadTarget:withConditions:]_block_invoke.72 <null>:2 (<App>:x86_64+0x10320d66a)
    #2 __44-[FBLPromise(CatchAdditions) onQueue:catch:]_block_invoke <null>:2 (<App>:x86_64+0x102d50ba9)
    #3 __56-[FBLPromise chainOnQueue:chainedFulfill:chainedReject:]_block_invoke.68 <null>:2 (<App>:x86_64+0x102d6a562)
    #4 __44-[FBLPromise observeOnQueue:fulfill:reject:]_block_invoke_2 <null>:2 (<App>:x86_64+0x102d69190)
    #5 __wrap_dispatch_group_async_block_invoke <null>:2 (libclang_rt.tsan_iossim_dynamic.dylib:x86_64+0x71c37)
    #6 _dispatch_client_callout <null>:2 (libdispatch.dylib:x86_64+0x28ca)
 
  Previous read of size 1 at 0x7b440003acf9 by thread T5:
    #0 -[GDTCCTUploadOperation isFinished] <null>:2 (<App>:x86_64+0x103212bac)
    #1 __NSOQSchedule_f <null>:2 (Foundation:x86_64+0x9c568)
    #2 _dispatch_client_callout <null>:2 (libdispatch.dylib:x86_64+0x28ca)
 
  Location is heap block of size 320 at 0x7b440003ac00 allocated by thread T6:
[...]

PR #7778, opened 2021-03-24
Enable TSAN diagnostics for xcodebuild tests  

maksymmalyhin:
- enable TSAN diagnostics for xcodebuild to detect issues like #7771 earlier
- disable `xcpretty` because it filters out the diagnostics logs

#no-changelog

### DIFF
--- a/FirebaseCore/Tests/Unit/FIRAppTest.m
+++ b/FirebaseCore/Tests/Unit/FIRAppTest.m
@@ -932,23 +932,6 @@ NSString *const kFIRTestAppName2 = @"test-app-name-2";
   OCMVerifyAllWithDelay(self.mockCoreDiagnosticsConnector, 0.5);
 }
 
-// TODO: Remove after testing the TSAN flag.
-- (void)testTSAN {
-  __block BOOL variable = YES;
-  NSInteger count = 10;
-  XCTestExpectation *expectation = [self expectationWithDescription:@"expectation"];
-  expectation.expectedFulfillmentCount = count;
-  for (NSInteger i = 0; i <= count; i++) {
-    dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
-      // Should fail on the thread sanitizer diagnostics.
-      variable = !variable;
-      [expectation fulfill];
-    });
-  }
-
-  [self waitForExpectations:@[ expectation ] timeout:1];
-}
-
 #pragma mark - private
 
 - (XCTestExpectation *)expectNotificationNamed:(NSNotificationName)name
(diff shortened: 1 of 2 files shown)


maksymmalyhin:
Questions:
- [Firestore job on Travis](https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/764249631) timeouted, most likely due to additional diagnostics and extra logs. Should we just disable it?
- The [log was actually overflown for Firestore](https://github.com/firebase/firebase-ios-sdk/pull/7778/checks?check_run_id=2187447106#step:4:19169) but the full log is available for download from the menu. IMO it is good enough as we will be able to see the error reason for any build without any code modifications (in contrast to disabling xcpretty if a failure reason was masked by it).

@paulb777 @ryanwilson @ncooke3 @granluo WDYT?

granluo:
I think that works as far as logs printed in the console would not time out the tests. Also, printing a large log might slow down tests. One way, which I can think of, to handle large log is to switch the log from stdout to a text/log file and upload the file to GHA/Travis. I think the current solution should be fine if the tests can run without time out due to the log size.

paulb777:
Did you consider using the [existing sanitizer options in build.sh](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh#L61)?

Travis definitely failed because of [exceeding log buffer](https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/764249631#L8735).

We should get review from Firestore team because I believe they added both the existing sanitizer infrastructure and xcpretty filtering. I'll add @var-const to reviewers.

var-const:
+1 to Paul's comment -- please try to reuse the existing `tsan` option rather than unconditionally enable Thread Sanitizer. Note that Thread Sanitizer builds are slower and also may fail for reasons that are impractical to fix short-term (e.g. an issue with a dependency).

Unfortunately, I don't have much context on `xcpretty`. What are the diagnostic logs that are filtered out? TSAN-specific logs?

maksymmalyhin:
Thank you for pointing out to the existing diagnostics options! I didn't notice them first and then didn't realize that they [work with xcbuild](https://github.com/firebase/firebase-ios-sdk/blob/881518c82aa5882bc9fff17b60881edef1ccea9e/scripts/build.sh#L251) as well (not only with cmake). 

As for `xcpretty`, now I think a better approach would be to add the raw logs to the build artifacts but keep using `xcpretty` as the regular output. Looks like the raw log should be already available at [xcodebuild.log](https://github.com/firebase/firebase-ios-sdk/blob/881518c82aa5882bc9fff17b60881edef1ccea9e/scripts/build.sh#L120) file. @granluo Any suggestion on how to do this better are appreciated. 

I'm closing this PR. Thank you everyone for the feedback!

PR #7788, opened 2021-03-25
Reorganize FIAM's Resources for FIAM

### LINKED ISSUE #7715: In-App Messaging via SPM missing resource bundle

  * Xcode version: 12.4
  * Firebase SDK version: 7.8.1
  * Installation method: Swift Package Manager
  * Firebase Component: Analytics, Crashlytics, InAppMessaging, Messaging

#### Steps to reproduce:

I have Firebase integrated via SPM. I am trying to send a test message for In-App Messaging. The device receives the message but I get an error when it attempts to display:

`7.8.1 - [Firebase/InAppMessaging][I-IAM400009] Display ran into error for message 592562000580575232: Error Domain=com.firebase.inappmessaging.display Code=1 "Resource bundle is missing." UserInfo={NSLocalizedDescription=Resource bundle is missing.}
`

#### Relevant Code:

N/A

ncooke3:
FIAM was not working on SPM because the resource bundle was not generating.

There was a lot of experimentation here so I wanted to make note of some realizations/learnings:
• In order for SPM to generate a bundle for a target, the target's resources need to be in the target's `Sources` directory.
• The Package manifest will attempt to compile and build all targets (an issue for targets with platform-specific resources).
• FIAM only provides default UI implementations for iOS (so I've removed the blank tvOS storyboard)
• SPM generates target bundles in a single directory. So if `target B` is listed as a dependency of `target A` and both targets defines `resources` in their declarations, then SPM will generate a `Target_B.bundle` and a `Target_A.bundle` at the **same** directory level (this was against my intuition as I thought target B's bundle would be within target A's bundle since A depends on B).

Note on SPM & platform specific resources:
• There currently isn't a clear cut way to define platform specific resources. [See the conversation on the Swift forums here.](https://forums.swift.org/t/platform-specific-resources/43612) The workaround here is to do something similar to what was done in [SPM for Analytics](https://github.com/firebase/firebase-ios-sdk/blob/881518c82aa5882bc9fff17b60881edef1ccea9e/Package.swift#L233) by creating a new target containing FIAM's iOS resources that is conditionally depended on in the primary FIAM target. 

Future work:
Translating FIAM's default UI implementations from storyboards to programmatic UIKit should make this type of problem easier to fix in the future because we can use preprocessor directives (which we can't use in storyboards). Plus, programmatic UI's should be easier to maintain and update. 

Fixes #7715.


Package.swift
```diff
@@ -599,8 +599,12 @@ let package = Package(
       exclude: [
         "DefaultUI/CHANGELOG.md",
         "DefaultUI/README.md",
+        "Resources/tvOS"
```
ncooke3, line 602: Seems like we still have to exclude the tvOS resources (else build failure when building on iOS targets) despite having the condition for this target set for iOS platforms only. Since FIAM only supports iOS, this should be okay?

paulb777: Should be fine until we're ready to restart the tvOS work

christibbs: @ncooke3 @paulb777 FIAM currently supports tvOS-- there's no default UI, but devs can define custom interfaces.

paulb777: @christibbs Oh right. Sorry about my confusion.

paulb777: FIAM support is there for CocoaPods but not Swift Package Manager : https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L585

ncooke3: I referenced https://firebase.google.com/docs/ios/learn-more ...should this table be amended to include a check for FIAM / tvOS?

christibbs: @ncooke3 Yes, reach out to Mark Arndt to update this.

maksymmalyhin:
@ncooke3 One option that seems to be worth trying if other approaches fail:
- define separate targets with resources per platform, e.g. `FirebaseInAppMessagingResources_ios`, `FirebaseInAppMessagingResources_tvos`, etc.
- define the resource targets as conditional dependencies to `FirebaseInAppMessagingTarget` e.g. like here: https://github.com/firebase/firebase-ios-sdk/blob/a7ada99447d7763b498c0f0887455c7890bc269e/Package.swift#L437

ncooke3:
thanks for the comment @maksymmalyhin! I think FIAM's target is conditionally dependent as well though. See here: https://github.com/firebase/firebase-ios-sdk/pull/7788/files#R585

Package.swift
```diff
@@ -594,11 +607,15 @@ let package = Package(
...
         .product(name: "GULEnvironment", package: "GoogleUtilities"),
         .product(name: "nanopb", package: "nanopb"),
+        .target(name: "FirebaseInAppMessaging_iOS", condition: .when(platforms: [.iOS])),
+        .target(name: "FirebaseInAppMessaging_tvOS", condition: .when(platforms: [.tvOS])),
       ],
       path: "FirebaseInAppMessaging/Sources",
       exclude: [
         "DefaultUI/CHANGELOG.md",
         "DefaultUI/README.md",
+        "iOS",
```
maksymmalyhin, line 617: Not sure if it actually matters here but looks like in other target we add `/` to the directory names we are excluding, so we may need to change it to `"iOS/"`.

ncooke3: Thanks for the tip! I can't tell if it makes a difference because the cause of the failures is elsewhere. I will keep this in mind though 🙂 

The failures we're now seeing are because I have added  platform-specific targets `FirebaseInAppMessaging_tvOS` and `FirebaseInAppMessaging_iOS`. AFAIU, while SPM supports targets having conditionally dependent platform-specific targets, it does not support platform-specific targets since all targets in the package manifest are compiled.

Currently, experimenting with rearranging our workflow tests to see if there is better way to test coverage. Will update soon!

maksymmalyhin: My guess was that exclude actually is not working here so the iOS and tvOS specific resource are still compiled for `FirebaseInAppMessaging` which triggers the compilation error. I think it's worth checking. But if the targets like `FirebaseInAppMessaging_tvOS` are still compiled for e.g. macOS even if they are not included as a dependency anywhere, then it's disappointing.

ncooke3: I believe targets like `FirebaseInAppMessaging_tvOS` are still compiled for e.g. macOS even if they are not included as a dependency anywhere. Here's a screenshot of the compile failure. Despite building for iOS, the packahe manifest still tries to compile the `FirebaseInAppMessaging_tvOS` target.

Screen Shot 2021-03-31 at 2 11 42 PM

maksymmalyhin: Disappointing... I would be curious if this an issue for the tests only or for a real project using Firebase as well.

christibbs:
I see 🟢

I'll run tests on this today and approve if we're looking good.

FirebaseInAppMessaging/iOS/FirebaseInAppMessaging_iOS.m
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Google LLC
```
paulb777, line 1: 2021

FirebaseInAppMessaging/iOS/include/FirebaseInAppMessaging_iOS.h
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Google LLC
```
paulb777, line 1: 2021

FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
```diff
@@ -56,14 +53,23 @@ + (NSBundle *)getViewResourceBundle {
...
+#if SWIFT_PACKAGE
+    bundledResource = @"Firebase_FirebaseInAppMessaging_iOS";
+#else
+    bundledResource = @"InAppMessagingDisplayResources";
+#endif  // SWIFT_PACKAGE
+
     NSBundle *containingBundle;
     NSURL *bundleURL;
     // The containing bundle is different whether FIAM is statically or dynamically linked.
     for (containingBundle in @[ [NSBundle mainBundle], [NSBundle bundleForClass:myClass] ]) {
```
paulb777, line 68: What happens for tvOS?  Perhaps add a comment.

paulb777 (commented):
Nice way to eliminate complexity from the problem!

FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
christibbs: tvOS doesn't have a default display implementation right now-- so this class isn't included.

christibbs: +1 for comment, though.

christibbs:
We're looking good. FIAM still works on iOS and tvOS via CocoaPods. FIAM iOS works via Swift Package Manager, not on tvOS. @ncooke3 this is expected, right?

maksymmalyhin:
@ncooke3 Great investigation, thank you for doing it!

I may be worth creating a doc like [this](https://github.com/firebase/firebase-ios-sdk/blob/master/AddNewPod.md) for the platform-specific resource in SPM with a summary, conclusions and recommendations for teams that use or plan to such resources. I can be done as a follow up for the PR.

christibbs:
Tested this on SPM and CocoaPods for iOS and tvOS. Everything's working.

christibbs (approved):

FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
```diff
@@ -17,8 +17,6 @@
 #import <TargetConditionals.h>
 #if TARGET_OS_IOS
 
-#import <Foundation/Foundation.h>
-
```
ncooke3, line 21: This is already imported in the header file.

FirebaseInAppMessaging/Sources/DefaultUI/FIRIAMDefaultDisplayImpl.m
```diff
@@ -30,7 +28,6 @@
 #import "FirebaseInAppMessaging/Sources/DefaultUI/Modal/FIRIAMModalViewController.h"
 #import "FirebaseInAppMessaging/Sources/Private/Util/FIRIAMTimeFetcher.h"
 #import "FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessaging.h"
-#import "FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessagingRendering.h"
```
ncooke3, line 33: Another duplicate import.

maksymmalyhin (approved):
LGTM 👍

PR #7789, opened 2021-03-25
Bump the nanopb version to 0.3.9.8

### LINKED ISSUE #7787: Remediate CVE-2021-21401 by updating to nanopb 0.3.9.8 or higher

  * Xcode version: 12.1
  * Firebase SDK version: 7.4.0
  * Firebase Component: nanopb

[CVE-2021-21401](https://vuln.whitesourcesoftware.com/vulnerability/CVE-2021-21401): "In Nanopb before versions 0.3.9.8 and 0.4.5, decoding a specifically formed message can cause invalid `free()` or `realloc()` calls if the message type contains an `oneof` field, and the `oneof` directly contains both a pointer field and a non-pointer field." Issue was reported on March 23, 2021, and was resolved with Nanopb 0.3.9.8 or 0.4.5

#### Steps to reproduce:

- Install Firebase 7.4.0 or higher
- Observe Nanopb is version 0.3.9.7 based on Google spec of Nanopb 2.03097.0

var-const:
Fixes #7787.

Update to nanopb 0.3.9.8. Protos do not need to be regenerated since this is a library only nanopb patch.


paulb777:
On quick glance, I see two issues:

- The GoogleDataTransport version is wrong at SpecsDev - https://github.com/firebase/SpecsDev/blob/main/GoogleDataTransport/8.3.1/GoogleDataTransport.podspec#L3
- GoogleAppMeasurement needs to be updated as well on SpecsDev - https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleAppMeasurement.podspec.json#L9

Crashlytics/Protogen/nanopb/crashlytics.nanopb.c
```diff
@@ -14,6 +14,22 @@
  * limitations under the License.
  */
```
paulb777, line 16: Something seems strange here that the copyright was added but the Generated version string was not updated.  Here and in subsequent three files.

paulb777 (commented):
Please delete the extra generated copyrights. 

I'm ok if the protos don't get regenerated for 0.3.9.8 since it a library patch only without a corresponding generation change.

The analytics CI failure can be disregarded in this PR since it is a result of the SpecsDev source not being specified at https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/analytics.yml#L25

paulb777:
Merge #7795 to fix the CI failure

paulb777:
Asking @maksymmalyhin for a review since both @var-const and I have contributed to this PR

maksymmalyhin (approved):
LGTM on CI green

paulb777:
Planning to publish the nanopb 2.30908.0 (0.3.9.8) podspec after a day of CI here

PR #7799, opened 2021-03-29
Document how to initialize Crashlytics with Xcode debugger attached

elenadoty:
#no-changelog


samedson (approved):

peterfriese:
Nice - is this documented in the docs as well?

elenadoty:
Do you mean externally? I don't believe so - I can definitely add something if you think it would be helpful. The comment here was mainly to document the process for internal debugging.

PR #7806, opened 2021-03-30
Fix Segmentation sample nightly testing failure

paulb777:
Fix the Segmentation failure in #7797 

It was not seeing the nanopb in SpecsDev required by the podspecs in master.


ryanwilson (approved):
Do we want to permanently keep those `source` lines at the top of the Podfile?

paulb777:
Yes. The sample test should always build in CI and we depend on the other repos if we want to integration dependent pods like nanopb, GDT, and GoogleUtilities.

PR #7807, opened 2021-03-30
Nightly zip should use SpecsDev

paulb777:
Fix zip failure in #7797


ryanwilson (commented):
Hmm, this could affect using the zip for releases, right? Since now `SpecsDev` could be included in the published zip file which we wouldn't expect. WDYT?

paulb777:
Don't we currently use `github.event.inputs.custom_spec_repos` for releases?

We would need to split this out when we release from a nightly build.

ryanwilson (approved):

PR #7821, opened 2021-04-01
Rotate Crashlytics Install UUID during upload instead of startup

samedson:
This ensures we always call the `installationIDWithCompletion` during upload when we are already collecting crash data. Before we did this process at startup.

This moves the install_uuid out of the `.clsrecord` files and into the proto.

#no-changelog

### DIFF
--- a/Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
+++ b/Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
@@ -27,11 +27,15 @@ static NSString *const FIRCLSInstallationIIDHashKey = @"com.crashlytics.install.
 // Legacy key that is automatically removed
 static NSString *const FIRCLSInstallationADIDKey = @"com.crashlytics.install.adid";
 
+static unsigned long long FIRCLSInstallationsWaitTime = 10 * NSEC_PER_SEC;
+
 @interface FIRCLSInstallIdentifierModel ()
 
 @property(nonatomic, copy) NSString *installID;
 
 @property(nonatomic, readonly) FIRInstallations *installations;
+
+@property(nonatomic, assign) dispatch_once_t regenerateOnceToken;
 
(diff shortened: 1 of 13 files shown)


Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
```diff
@@ -96,20 +100,33 @@ - (NSString *)generateInstallationUUID {
...
+        installationIDWithCompletion:^(NSString *_Nullable currentIID, NSError *_Nullable error) {
+          didRotate = [self rotateCrashlyticsInstallUUIDWithIID:currentIID error:error];
+
+          if (didRotate) {
+            FIRCLSInfoLog(@"Rotated Crashlytics Install UUID because Firebase Install ID changed.");
+          }
+          dispatch_semaphore_signal(semaphore);
+        }];
+
+    intptr_t result = dispatch_semaphore_wait(
```
maksymmalyhin, line 123: Just want to double check that this part is never called and never blocks main thread as it will trigger a deadlock (at least for FIRCLSInstallationsWaitTime) because installationIDWithCompletion completion handler is called on the main thread.

maksymmalyhin (approved):
LGTM 👍

maksymmalyhin:
Please fix CI before merging.

elenadoty (approved):

Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
samedson: I guess I can just put this on a background thread to make sure that's the case

samedson: Alright we call this from a background queue so we should be good
Screen Shot 2021-04-21 at 11 55 53 AM

PR #7828, opened 2021-04-02
Update README with Firebase platform support info

ncooke3:
Adding a section in the readme to point to the docs for a source of truth on firebase's platform support.


paulb777:
How about deleting the table on this page too?  The usage information is covered elsewhere.

morganchen12 (approved):

PR #7841, opened 2021-04-05
Firebase pod support for watchOS for RC and RTDB

paulb777:
Add Firebase pod support for watchOS for Remote Config and RTDB.

Address discussion at https://github.com/firebase/firebase-ios-sdk/pull/7656#issuecomment-813437060

Update the sample to test the Firebase pod.

Add sample build test to CI


chliangGoogle (approved):
Nice! Thanks for doing that Paul!

PR #7842, opened 2021-04-05
UserDataWriter

schmidt-sebastian:
This ports UserDataWriter from Android.


Firestore/core/src/model/resource_path.h
```diff
@@ -53,16 +53,14 @@ class ResourcePath : public impl::BasePath<ResourcePath>,
    * Creates and returns a new path from the given resource-path string, where
    * the path segments are separated by a slash "/".
    */
-  static ResourcePath FromString(const std::string& path);
+  static ResourcePath FromString(absl::string_view path);
```
var-const, line 56: Unfortunately, this needs to be reverted.

We previously ran into trouble when using Abseil classes in functions that cross the boundary between the iOS repo and the C++ repo. IIRC, the problem is that Abseil uses inline namespaces so that `absl::string_view` refers to different types between different versions of Abseil. Let's say that a 2019 version of Abseil defines it as `absl::lts_2019_05_01::string_view`, and a 2020 version as `absl::lts_2020_12_12::string_view`.

This becomes a problem in the iOS build because the C++ repo part of our code is compiled with a certain version of Abseil "baked in", whereas the iOS repo part is compiled by our users using the Abseil version that comes from CocoaPods. In all likelihood, these would be two different Abseil versions. So when using `ResourcePath::FromString(absl::string_view)` to communicate, the two parts would see the signature differently; for example:
* the C++ repo part (the public part) would see it as `ResourcePath::FromString(absl::lts_2019_05_01::string_view)`
* the iOS repo part (the core, or internal, part) would see it as `ResourcePath::FromString(absl::lts_2020_12_12::string_view)`.

So the linker would be unable to resolve the dependency.

Firestore/core/src/model/document_key.h
```diff
@@ -57,6 +59,9 @@ class DocumentKey {
   /** Creates and returns a new document key with the given segments. */
   static DocumentKey FromSegments(std::initializer_list<std::string> list);
 
+  /** Returns a DocumentKey from a fully qualified resource name. */
+  static DocumentKey FromName(absl::string_view name);
```
var-const, line 63: Hmm, this has the potential for the same problem as in `ResourcePath`. It should be okay as long as the code in the public part doesn't come to rely on this function, but that seems brittle.

OTOH, looking at how this is used, replacing the argument with a `const std::string&` would force an undesirable conversion. Unfortunately, the only two alternatives I can think of are unatrractive:
* make this function accept a char pointer and a size;
* make this function private and create a class that is a friend of this class and access `FromName` through it.

What do you think?

Firestore/core/src/model/document_key.cc
```diff
@@ -49,14 +49,26 @@ DocumentKey::DocumentKey(ResourcePath&& path)
   AssertValidPath(*path_);
 }
 
+/* static */
```
var-const, line 52: Nit: `/*static*/` comments are [discouraged](go/c-readability-advice#static-comments).

Firestore/core/src/model/database_id.h
```diff
@@ -46,6 +47,9 @@ class DatabaseId : public util::Comparable<DatabaseId> {
   explicit DatabaseId(std::string project_id,
                       std::string database_id = kDefault);
 
+  /** Returns a DatabaseId from a fully qualified resource name. */
+  static DatabaseId FromName(absl::string_view name);
```
var-const, line 51: Same comment here as other files in `model`.

Firestore/core/src/model/database_id.cc
```diff
@@ -34,6 +36,15 @@ DatabaseId::DatabaseId(std::string project_id, std::string database_id) {
   rep_ = std::make_shared<Rep>(std::move(project_id), std::move(database_id));
 }
 
+/* static */
```
var-const, line 39: Nit: remove the comment.

Firestore/core/src/model/database_id.cc
```diff
@@ -34,6 +36,15 @@ DatabaseId::DatabaseId(std::string project_id, std::string database_id) {
...
+DatabaseId DatabaseId::FromName(absl::string_view name) {
+  ResourcePath resource_name = ResourcePath::FromString(name);
```
var-const, line 41: Nit: consider using `auto`.

Firestore/core/src/model/document_key.cc
```diff
@@ -49,14 +49,26 @@ DocumentKey::DocumentKey(ResourcePath&& path)
...
 DocumentKey DocumentKey::FromSegments(std::initializer_list<std::string> list) {
   return DocumentKey{ResourcePath{list}};
 }
 
+/* static */
+DocumentKey DocumentKey::FromName(absl::string_view name) {
+  ResourcePath resource_name = ResourcePath::FromString(name);
+  HARD_ASSERT(resource_name.size() >= 4 && resource_name[0] == "projects" &&
+                  resource_name[2] == "databases" &&
+                  resource_name[4] == "documents",
```
var-const, line 67: Since you're accessing the fifth element (`[4]`), the size check should be `> 4`, not `>= 4`.

Firestore/core/src/model/database_id.cc
```diff
@@ -34,6 +36,15 @@ DatabaseId::DatabaseId(std::string project_id, std::string database_id) {
...
+  HARD_ASSERT(resource_name.size() >= 3 && resource_name[0] == "projects" &&
```
var-const, line 42: Similar to the other file, the check should be `>3`.

var-const (changes requested):

Firestore/Source/API/FSTUserDataWriter.h
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+#include <map>
```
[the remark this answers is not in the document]
schmidt-sebastian: Removed

Firestore/Source/API/FSTUserDataWriter.h
```diff
@@ -0,0 +1,41 @@
...
+@interface FSTUserDataWriter : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+
+- (instancetype)initWithFirestore:(std::shared_ptr<api::Firestore>)firestore
+          serverTimestampBehavior:(FIRServerTimestampBehavior)serverTimestampBehavior;
+
+- (id)convertedValue:(const firebase::firestore::google_firestore_v1_Value&)value;
+
+@end
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/Source/API/FSTUserDataWriter.mm
```diff
@@ -0,0 +1,164 @@
...
+#import <Firestore/core/src/nanopb/nanopb_util.h>
+#import <Firestore/core/src/util/hard_assert.h>
+#import <Foundation/Foundation.h>
+
+#include <string>
+#include <utility>
+
+#include "Firestore/Protos/nanopb/google/firestore/v1/document.nanopb.h"
+#include "Firestore/Source/API/FIRDocumentReference+Internal.h"
+#include "Firestore/Source/API/FSTUserDataWriter.h"
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/Source/API/FSTUserDataWriter.mm
```diff
@@ -0,0 +1,164 @@
...
+namespace util = firebase::firestore::util;
+namespace model = firebase::firestore::model;
+namespace nanopb = firebase::firestore::nanopb;
+
+using firebase::firestore::GeoPoint;
+using firebase::firestore::google_firestore_v1_Value;
+using firebase::firestore::google_firestore_v1_MapValue;
+using firebase::firestore::google_firestore_v1_ArrayValue;
+using firebase::firestore::google_protobuf_Timestamp;
+using nanopb::MakeNSData;
```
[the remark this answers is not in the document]
schmidt-sebastian: It doesn't seem to work in ObjC files. Thanks for catching - I thought it would work.

Firestore/Source/API/FSTUserDataWriter.mm
```diff
@@ -0,0 +1,164 @@
...
+          serverTimestampBehavior:(FIRServerTimestampBehavior)serverTimestampBehavior {
+  self = [super init];
+  if (self) {
+    _firestore = std::move(firestore);
+    _serverTimestampBehavior = serverTimestampBehavior;
+  }
+  return self;
+}
+
+- (id)convertedValue:(const firebase::firestore::google_firestore_v1_Value &)value {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/Source/API/FSTUserDataWriter.mm
```diff
@@ -0,0 +1,164 @@
...
+          GeoPoint(value.geo_point_value.latitude, value.geo_point_value.longitude));
+  }
+
+  UNREACHABLE();
+}
+
+- (NSDictionary<NSString *, id> *)convertedObject:(const google_firestore_v1_MapValue &)mapValue {
+  NSMutableDictionary *result = [NSMutableDictionary dictionary];
+  for (pb_size_t i = 0; i < mapValue.fields_count; ++i) {
+    const std::string &key = nanopb::MakeString(mapValue.fields[i].key);
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed (and thanks for the details)

Firestore/core/src/model/database_id.cc
schmidt-sebastian: Removed (we have these in many places though)

Firestore/core/src/model/database_id.cc
schmidt-sebastian: Fixed. This needs to be fixed on Android too.

Firestore/core/src/model/document_key.cc
schmidt-sebastian: Removed

Firestore/core/src/model/document_key.cc
schmidt-sebastian: Fixed here (Android coming)

Firestore/core/src/model/server_timestamp_util.h
```diff
@@ -35,7 +42,16 @@ bool IsServerTimestamp(const google_firestore_v1_Value& value);
 /**
  * Returns the local time at which the timestamp was written to the document.
  */
-const google_firestore_v1_Value& GetLocalWriteTime(
+const google_protobuf_Timestamp& GetLocalWriteTime(
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

Firestore/core/src/model/server_timestamp_util.h
```diff
@@ -35,7 +42,16 @@ bool IsServerTimestamp(const google_firestore_v1_Value& value);
...
+    const google_firestore_v1_Value& value);
+
+/**
+ * Returns the value of the field before this ServerTimestamp was set.
+ *
+ * Preserving the previous values allows the user to display the last resoled
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/src/remote/serializer.cc
```diff
@@ -476,7 +488,7 @@ pb_bytes_array_t* Serializer::EncodeResourceName(
 
 ResourcePath Serializer::DecodeResourceName(Reader* reader,
                                             absl::string_view encoded) const {
-  ResourcePath resource = ResourcePath::FromStringView(encoded);
+  ResourcePath resource = ResourcePath::FromString(encoded);
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

Firestore/core/src/model/document_key.h
schmidt-sebastian: I updated the method to take a std::string since it is used in the same code path that also uses the DatabaseId parser (which I reverted back to its original code).

Firestore/core/src/model/resource_path.h
schmidt-sebastian: I ... definitely did not know this. Reverted.

Firestore/core/src/model/database_id.h
schmidt-sebastian: Fixed

Firestore/core/src/model/resource_path.h
var-const: This mostly applies to `api` and, to lesser extent, `model` (the two most used namespaces by the public code). Unfortunately, we don't have an enforcement mechanism other than looking for this in code reviews. If you have any ideas, let me know.

Firestore/core/src/model/server_timestamp_util.cc
```diff
@@ -25,8 +25,41 @@ namespace model {
...
+  result.which_value_type = google_firestore_v1_Value_map_value_tag;
+
+  pb_size_t count = previous_value ? 3 : 2;
+
+  auto& map_value = result.map_value;
+  map_value.fields_count = count;
+  map_value.fields =
+      nanopb::MakeArray<google_firestore_v1_MapValue_FieldsEntry>(count);
+
+  auto field = map_value.fields;
```
var-const, line 44: Nit: `s/auto/auto*/`.

var-const (approved):

PR #7843, opened 2021-04-05
UserDataReader

schmidt-sebastian:
Note: This PR is much smaller than it appears.

This adds Android's UserDataReader, which is basically UserDataConverter but instead of converting user data to FieldValues, it directly converts user data to Proto. To make this less intrusive, UserDataReader currently uses UserDataConverter to then convert the Protos to FieldValues (this step will be removed soon).

To make the diff easier to review, UserDataConverter is renamed to UserDataConverter_legacy and a bunch of comment changes have been made to force Git to show the diff between UserDataReader and UserDataConverter (rather than from UserDataConverter to UserDataConverter_legacy).


Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -277,51 +318,70 @@ - (FieldValue)parsedQueryValue:(id)input allowArrays:(bool)allowArrays {
...
+                                                                      context:std::move(context)]);
     }
   }
 }
 
-- (FieldValue)parseDictionary:(NSDictionary<NSString *, id> *)dict
-                      context:(ParseContext &&)context {
+- (google_firestore_v1_Value)parseDictionary:(NSDictionary<NSString *, id> *)dict
+                                     context:(ParseContext &&)context {
+  google_firestore_v1_Value result;
```
var-const, line 332: Nit: consider initializing `result` using empty curly braces (I think the code initializes all the values correctly, but it just feels like a safer default).

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -277,51 +318,70 @@ - (FieldValue)parsedQueryValue:(id)input allowArrays:(bool)allowArrays {
...
       context.AddToFieldMask(*path);
     }
-    return ObjectValue::Empty().AsFieldValue();
   } else {
-    __block ObjectValue result = ObjectValue::Empty();
-
+    // Populate a vector of fields since we don't know the size of the final fields array.
     [dict enumerateKeysAndObjectsUsingBlock:^(NSString *key, id value, BOOL *) {
-      absl::optional<FieldValue> parsedValue =
+      absl::optional<google_firestore_v1_Value> parsedValue =
```
var-const, line 345: Question: what does it take to know whether the value can be parsed? What I'm getting at is whether it could be feasible to do a first pass on `dict` to get the array size, then create the values right in the array and skip the vector.

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -491,26 +557,100 @@ - (void)parseSentinelFieldValue:(FIRFieldValue *)fieldValue context:(ParseContex
...
+    return [self encodeReference:_databaseID key:reference.key];
 
   } else {
     ThrowInvalidArgument("Unsupported type: %s%s", NSStringFromClass([input class]),
                          context.FieldDescription());
   }
 }
 
-- (std::vector<FieldValue>)parseArrayTransformElements:(NSArray<id> *)elements {
+- (google_firestore_v1_Value)encodeNullValue {
```
var-const, line 568: Note: in C++, I'd suggest making these free functions (they don't need to access anything in the class). I'm not sure what is considered best practices in Objective-C in this regard, but it seems like similar reasoning should apply. The advantage of free functions is that they cannot access private data -- this makes them a little easier to reason about and prevents them from coming to rely on the internal state of the class by accident.

Optionally, you could move these functions into a pure C++ file.

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -277,51 +318,70 @@ - (FieldValue)parsedQueryValue:(id)input allowArrays:(bool)allowArrays {
...
   }
+
+  return result;
 }
 
-- (FieldValue)parseArray:(NSArray<id> *)array context:(ParseContext &&)context {
-  __block FieldValue::Array result;
-  result.reserve(array.count);
+- (google_firestore_v1_Value)parseArray:(NSArray<id> *)array context:(ParseContext &&)context {
+  __block google_firestore_v1_Value result;
```
var-const, line 368: Same note re. using `result{}`. The idea is similar to go/totw/182.

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -277,51 +318,70 @@ - (FieldValue)parsedQueryValue:(id)input allowArrays:(bool)allowArrays {
...
 
   [array enumerateObjectsUsingBlock:^(id entry, NSUInteger idx, BOOL *) {
-    absl::optional<FieldValue> parsedEntry = [self parseData:entry
-                                                     context:context.ChildContext(idx)];
+    absl::optional<google_firestore_v1_Value> parsedEntry =
+        [self parseData:entry context:context.ChildContext(idx)];
     if (!parsedEntry) {
       // Just include nulls in the array for fields being replaced with a sentinel.
-      parsedEntry = FieldValue::Null();
+      parsedEntry = absl::optional<google_firestore_v1_Value>{[self encodeNullValue]};
```
var-const, line 379: Nit: explicitly specifying `absl::optional` shouldn't be necessary.

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -361,22 +421,28 @@ - (void)parseSentinelFieldValue:(FIRFieldValue *)fieldValue context:(ParseContex
     context.AddToFieldTransforms(*context.path(), ServerTimestampTransform());
 
   } else if ([fieldValue isKindOfClass:[FSTArrayUnionFieldValue class]]) {
-    std::vector<FieldValue> parsedElements =
+    std::vector<google_firestore_v1_Value> parsedElements =
         [self parseArrayTransformElements:((FSTArrayUnionFieldValue *)fieldValue).elements];
-    ArrayTransform array_union(TransformOperation::Type::ArrayUnion, std::move(parsedElements));
+    ArrayTransform array_union(TransformOperation::Type::ArrayUnion,
```
var-const, line 426: Nit: in Objective-C-land, I think this should be `arrayUnion` (and similarly `arrayRemove` below).

var-const (changes requested):

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -249,10 +289,11 @@ - (FieldValue)parsedQueryValue:(id)input allowArrays:(bool)allowArrays {
  * @return The parsed value, or nil if the value was a FieldValue sentinel that should not be
  *   included in the resulting parsed data.
  */
-- (absl::optional<FieldValue>)parseData:(id)input context:(ParseContext &&)context {
+- (absl::optional<google_firestore_v1_Value>)parseData:(id)input context:(ParseContext &&)context {
   input = self.preConverter(input);
   if ([input isKindOfClass:[NSDictionary class]]) {
-    return [self parseDictionary:(NSDictionary *)input context:std::move(context)];
+    return absl::optional<google_firestore_v1_Value>{[self parseDictionary:(NSDictionary *)input
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed here and everywhere. It shows up as a compile error in CLion, but the build passes.

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: Done

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: We have to check whether it is a FIeldValue sentinel, which would duplicate the existing check in `parseData`. I don't think we can remove the `parseData` check since this function has more than this callsite.

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: Done

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: Removed

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: Fixed both

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: `parseReference` accesses `self.databaseID`, which means `parseReference`, `parseArray` and `parseObject` cannot be free functions. For consistency, I chose to keep the other functions as is.

Firestore/Source/API/FSTUserDataReader.mm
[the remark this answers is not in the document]
var-const: Just out of curiosity, what is the error?

Firestore/Source/API/FSTUserDataReader.mm
var-const: You could pass it as a parameter, but it's probably not worth the trouble.

Firestore/Source/API/FSTUserDataReader.mm
var-const: Ok, if I'm reading the code correctly, it amounts to going through the dictionary and calling `[input isKindOfClass:[FIRFieldValue class]]` on each element. If this is correct, then I strongly suspect that it's faster than creating and filling a vector and then copying vector elements into the array. What do you think about doing this, perhaps in a follow-up?

var-const (commented):
Pretty much LGTM with one question.

Firestore/Source/API/FSTUserDataReader.mm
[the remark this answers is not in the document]
schmidt-sebastian: `Returning google_firestore_v1_Value from a method returning absl::optional<google_firestore_v1_Value>: Class google_firestore_v1_Value is not compatible with absl::optional<google_firestore_v1_Value>`

Firestore/Source/API/FSTUserDataReader.mm
schmidt-sebastian: I updated this in this PR.

var-const (approved):

PR #7849, opened 2021-04-06
FIRSampleAppUtilities: Replace deprecated method.

ncooke3:
[`openURL:`](https://developer.apple.com/documentation/uikit/uiapplication/1622961-openurl?language=objc) has been deprecated since the introduction iOS 10. Replaced its use with [`openURL:options:completionHandler:`](https://developer.apple.com/documentation/uikit/uiapplication/1648685-openurl?language=objc), passing in empty options and a `nil` completion handler.

Noticed this in the GHA logs while working on #7788.

This should eliminate the deprecation warning when testing. 

#no-changelog


maksymmalyhin (approved):
LGTM on CI green. Thank you for the clean up 👍

paulb777 (approved):

PR #7855, opened 2021-04-07
Disable code coverage when secrets cannot be accessed.

granluo:


granluo:
The issue is caused by secrets not able to be accessed and so token cannot be get.
https://github.com/firebase/firebase-ios-sdk/actions/runs/723832228

paulb777 (approved):

PR #7861, opened 2021-04-07
Ensure to archive/unarchive to original FIRInstanceID TokenInfo and APNSInfo

chliangGoogle:
Looks like we can archive and unarchive with an arbitrary class (Thanks Maksym!) so enforce messaging to always use FIRInstanceIDTokenInfo and FIRInstanceIDAPNSInfo to ensure backward compatibility.

### DIFF
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2021-04 -- v7.11.0
-- [changed] Refactor Messaging to internally not depending on InstanceID, but can co-exist. Will remove InstanceID dependency in the next Firebase breaking change. (#7814)
+- [changed] Refactor Messaging to internally not depend on InstanceID, but can co-exist. This makes Messaging's APIs of handling FCM registration token no longer depend on InstanceID to function. Recommend use Messaging to handle FCm registration tokens. (#7814)
 - [changed] Replaced NSCoding with NSSecureCoding. (#7831)
 
 # 2021-02 -- v7.7.0
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2021-04 -- v7.11.0
-- [changed] Refactor Messaging to internally not depend on InstanceID, but can co-exist. This makes Messaging's APIs of handling FCM registration token no longer depend on InstanceID to function. Recommend use Messaging to handle FCm registration tokens. (#7814)
+- [changed] Remove Messaging's dependency on InstanceID while continuing to optionally interoperate with it. This makes Messaging's APIs of handling FCM registration tokens no longer depend on InstanceID. We recommend migrating the handling of FCM registration tokens from InstanceID APIs to FCM APIs. https://firebase.google.com/docs/cloud-messaging/ios/client#access_the_registration_token (#7814)
 - [changed] Replaced NSCoding with NSSecureCoding. (#7831)
 
 # 2021-02 -- v7.7.0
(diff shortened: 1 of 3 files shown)


FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # 2021-04 -- v7.11.0
-- [changed] Refactor Messaging to internally not depending on InstanceID, but can co-exist. Will remove InstanceID dependency in the next Firebase breaking change. (#7814)
+- [changed] Refactor Messaging to internally not depend on InstanceID, but can co-exist. This makes Messaging's APIs of handling FCM registration token no longer depend on InstanceID to function. Recommend use Messaging to handle FCm registration tokens. (#7814)
```
paulb777, line 2: Consider: "Remove Messaging's dependency on InstanceID while continuing to optionally interoperate with it. This makes Messaging's APIs of handling FCM registration tokens no longer depend on InstanceID. We recommend migrating the handling of FCM registration tokens from InstanceID APIs to FCM APIs".

paulb777: Is there a link that demonstrates an example of migrating from IID to FCM?

paulb777:
Please check the SPM unit test failure

maksymmalyhin:
@chliangGoogle A couple suggestions from FIS experience:
- we can set a different class for a class name from an archive to be able to read old IID data. See an [example from FIS](https://github.com/firebase/firebase-ios-sdk/blob/bdac2b87f6d6e604deae914a970a75136b3a62b5/FirebaseInstallations/Source/Library/IIDMigration/FIRInstallationsIIDTokenStore.m#L102).
- I would suggest actually throwing an exception if IID is detected to help developers catching the undefined behaviour earlier.  [Here is how we do it in FIS](https://github.com/firebase/firebase-ios-sdk/blob/4ac113a271f22e156050cb62261606880080b704/FirebaseInstallations/Source/Library/FIRInstallations.m#L258), for FCM it can be even simpler. The log message looks good for the minor release though.

FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
```diff
@@ -168,9 +168,12 @@ - (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
                          fromData:rawAPNSInfo
                             error:&error];
     if (error) {
-      FIRMessagingLoggerDebug(
+      FIRMessagingLoggerInfo(
           kFIRMessagingMessageCodeTokenInfoBadAPNSInfo,
-          @"Could not parse raw APNS Info while parsing unarchived token info: %@", error);
+          @"Could not parse raw APNS Info while parsing unarchived token info: %@. If you are "
+          @"still using the deprecated InstanceID SDK to handle FCM registration token. Please "
```
maksymmalyhin, line 174: nit: I guess, we can detect if IID is actually installed in the app (e.g. like [here](https://github.com/firebase/firebase-ios-sdk/blob/4ac113a271f22e156050cb62261606880080b704/FirebaseInstallations/Source/Library/FIRInstallations.m#L267)) so we can be more precise in our messaging. I'm not too strong about it though.

maksymmalyhin (approved):
LGTM in general but we may handle the case a bit more gracefully if using new FCM with IID can lead to obscure bugs that may slip unnoticed to production apps.

FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
chliangGoogle: I think we should add this error in general for the breaking change when we remove IID dependency.

FirebaseMessaging/CHANGELOG.md
chliangGoogle: The doc is updated with Messaging API instead of IID since Fireconf: https://firebase.google.com/docs/cloud-messaging/ios/client#access_the_registration_token

I will add it in the comment.

PR #7831, opened 2021-04-02
Replace NSCoding with NSSecureCoding

chliangGoogle:
Also use GULSecureCoding to encode/decode. This partially fixes #3686.

Tested with previously installed app and fcm token is persisted correctly.

PR #7863, opened 2021-04-08
Revert "Replace NSCoding with NSSecureCoding"

chliangGoogle:
Reverts firebase/firebase-ios-sdk#7831

Looks like this is the one breaks the Encoding/Decoding as the GULSecureCoding doesn't offer overwriting the class. Considering the code change is significant, it's less risk to revert this change and fix the secure coding on top of the original implementation. (preferably after breaking change and let the large refactor of token code settle for a while)

### DIFF
--- a/FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
+++ b/FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
@@ -16,7 +16,6 @@
 
 #import "FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.h"
 
-#import <GoogleUtilities/GULSecureCoding.h>
 #import "FirebaseMessaging/Sources/FIRMessagingConstants.h"
 #import "FirebaseMessaging/Sources/FIRMessagingLogger.h"
 #import "FirebaseMessaging/Sources/FIRMessagingUtilities.h"
@@ -120,11 +119,7 @@ static const NSTimeInterval kDefaultFetchTokenInterval = 7 * 24 * 60 * 60;  // 7
   return [self.scope isEqualToString:kFIRMessagingDefaultTokenScope];
 }
 
-#pragma mark - NSSecureCoding
-
-+ (BOOL)supportsSecureCoding {
-  return YES;
-}
+#pragma mark - NSCoding
 
 - (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
   // These value cannot be nil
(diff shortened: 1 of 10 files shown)


paulb777 (approved):
Are we missing a test that would have caught the regression?

chliangGoogle:
The unit test didn't capture this. So we change Messaging from read/write to IIDTokenInfo to MessagingTokenInfo. We need a test like IID call deleteToken and then Messaging try to read the token.

maksymmalyhin (approved):
LGTM. BTW, I just realized that if we bump minimum supported macOS version to 10.13 then we don't need to use `GULSecureCoding` as modern `NSKeyedUnarchiver` API will be available for all supported OS (iOS, tvOS and watchOS versions already support the API).

FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
```diff
@@ -162,15 +158,19 @@ - (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
...
-        unarchivedObjectOfClasses:[NSSet setWithObjects:FIRMessagingAPNSInfo.class, nil]
-                         fromData:rawAPNSInfo
-                            error:&error];
-    if (error) {
-      FIRMessagingLoggerDebug(
-          kFIRMessagingMessageCodeTokenInfoBadAPNSInfo,
-          @"Could not parse raw APNS Info while parsing unarchived token info: %@", error);
+    // TODO(chliangGoogle: Use the new API and secureCoding protocol.
+    @try {
+      [NSKeyedUnarchiver setClass:[FIRMessagingAPNSInfo class]
```
maksymmalyhin, line 163: If we override the class globally then:
- we can still use `GULSecureCoding` since the override will affect it as well
- if the FCM version is used with IID then `FIRInstanceIDAPNSInfo` archives will be unarchived to `FIRMessagingAPNSInfo` in IID as well. It will probably not cause any issue as long as the classes are identical due to Objective-C dynamic nature, but it looks a bit risky and not clean.

I think we should modify it either:
1. override the class only for an `NSKeyedUnarchiver` instance, not globally. It's a safer option.
2. if we are fine with the global override, then the revert is not required. Just appending the global `[NSKeyedUnarchiver setClass:...]` will be enough.

maksymmalyhin (changes requested):
Noticed one potential issue(see the comment). In either way I would suggest changes to the PR. Would you mind discussing before proceeding.

FirebaseMessaging/Sources/Token/FIRMessagingTokenInfo.m
chliangGoogle: I suggest we made the change after I/O breaking change because Messaging already had significant amount of code change and at this point we want to ensure no unnecessary changes in so we can fully tested all the complicated scenarios with IID and Messaging.

maksymmalyhin (approved):
Discussed offline:
- the PR is a clean revert to the tested version, so it should be safe to do 
- we will most likely postpone the NSSecureCoding update until after Firebase 8 to reduce risks
- Firebase 8 is going to include code to make sure IID and FCM are not installed in the app together, so setting the class globally will not be an issue
- A version after Firebase 8 will include NSSecureCoding change based on existing GULSecureCoding API

PR #7870, opened 2021-04-08
Combine Storage: Futures are Results

paulb777:
Futures expect a result type, so no reason to deconstruct and reconstruct


peterfriese (approved):

paulb777:
From https://www.donnywals.com/using-promises-and-futures-in-combine/:  

 _**A Promise in Combine is a typealias for a closure that takes a Result as its single argument.**_

PR #7872, opened 2021-04-08
Run unit tests from pod lib lint

paulb777:
Miscellaneous Storage CI cleanups

- Run unit tests from `pod lib lint`
- Remove a duplicate test
- Test FirebaseStorageSwift with `--no-modular-headers` option on cron

#no-changelog

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -496,14 +496,6 @@ case "$product-$platform-$method" in
 
   Storage-*-xcodebuild)
     pod_gen FirebaseStorage.podspec --platforms=ios
-    RunXcodebuild \
-      -workspace 'gen/FirebaseStorage/FirebaseStorage.xcworkspace' \
-      -scheme "FirebaseStorage-Unit-unit" \
-      "${ios_flags[@]}" \
-      "${xcb_flags[@]}" \
-      build \
-      test
-
     if check_secrets; then
       # Integration tests are only run on iOS to minimize flake failures.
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -522,24 +514,6 @@ case "$product-$platform-$method" in
...
-    pod_gen FirebaseStorage.podspec --platforms=macos --clean
-    RunXcodebuild \
-      -workspace 'gen/FirebaseStorage/FirebaseStorage.xcworkspace' \
-      -scheme "FirebaseStorage-Unit-unit" \
-      "${macos_flags[@]}" \
-      "${xcb_flags[@]}" \
-      build \
-      test
-
-    pod_gen FirebaseStorage.podspec --platforms=tvos --clean
```
ncooke3, line 535: Just so I understand better, why are we removing these from the build script? Are your changes to the `pod-lib-lint` task in `storage.yml` effectively replacing these parts of the build script?

ncooke3 (approved):
Nice cleanup!

scripts/build.sh
paulb777: Yes. The unit tests now get run from `pod lib lint`. It's just recently possible because CocoaPods 1.10 introduced the feature of being able to select test_spec's to run instead of the all or none that it was previously.

paulb777:
Going to merge now, and can address any additional follow up in a subsequent PR

PR #7874, opened 2021-04-09
Fix Storage Integration Test CI

paulb777:
The Storage integration tests have been silently failing ever since 7.0.0 after the GoogleServiceInfo.plist does not get found because FirebaseCore is linked dynamically as a separate bundle instead of statically into the app. https://github.com/firebase/firebase-ios-sdk/runs/2294766037?check_suite_focus=true#step:7:475

This fix uses a ruby script to update the Test Container app with the GoogleService-Info.plist so that it gets found at initialization.

Also, stop running Firestore test_coverage CI on script changes.

See successful Storage integration test logs at https://github.com/firebase/firebase-ios-sdk/runs/2308057031?check_suite_focus=true

### DIFF
--- a/scripts/update_xcode_target.rb
+++ b/scripts/update_xcode_target.rb
@@ -1,0 +1,38 @@
+#!/usr/bin/env ruby
+
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


scripts/update_xcode_target.rb
```diff
@@ -0,0 +1,38 @@
...
+file = project.new_file(file_name)
+
+# Add the file to the all targets
+project.targets.each do |t|
+  if t.to_s == target
+    t.add_file_references([file])
+  end
+end
+
+#save project
```
ncooke3, line 37: super nit: `# Save project`

ncooke3 (approved):
Nice catch and great fix! The `update_xcode_target.rb` script makes a lot of sense 👍

PR #7875, opened 2021-04-09
Remove duplicate Foundation import

ncooke3:
See Foundation import in header file [here](https://github.com/firebase/firebase-ios-sdk/blob/724caede52c05cdc00ecc48b3df91b278db43e25/FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessaging.h#L20)

#no-changelog


paulb777 (commented):
I believe this is ok as is. See go/iwyu, which isn't a priority for us, but probably no reason to get less compliant.

ncooke3:
Thanks for the link and make sense. Good to know 👍

PR #7889, opened 2021-04-13
Force the template argument of std::min to prevent Windows build error.

jonsimantov:
Fixes an issue when building for Windows where this std::min line fails to infer the template type.

#no-changelog

### DIFF
--- a/Firestore/core/src/bundle/bundle_reader.cc
+++ b/Firestore/core/src/bundle/bundle_reader.cc
@@ -121,7 +121,7 @@ void BundleReader::ReadJsonToBuffer(size_t required_size) {
   while (buffer_.size() < required_size) {
     // Read at most 1024 bytes every time, to avoid allocating a huge buffer
     // when corruption leads to large `required_size`.
-    auto size = std::min<unsigned long>(1024ul, required_size - buffer_.size());
+    auto size = std::min<size_t>(1024ul, required_size - buffer_.size());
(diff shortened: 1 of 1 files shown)


var-const (approved):

var-const:
Thanks!

Firestore/core/src/bundle/bundle_reader.cc
```diff
@@ -121,7 +121,7 @@ void BundleReader::ReadJsonToBuffer(size_t required_size) {
   while (buffer_.size() < required_size) {
     // Read at most 1024 bytes every time, to avoid allocating a huge buffer
     // when corruption leads to large `required_size`.
-    auto size = std::min(1024ul, required_size - buffer_.size());
+    auto size = std::min<unsigned long>(1024ul, required_size - buffer_.size());
```
var-const, line 124: The linter is not happy:
```
> /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/scripts/check_lint.py 8e79c748219a4fc9969168212d3ae899ce87ac30
Firestore/core/src/bundle/bundle_reader.cc:124:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]
```
Perhaps use `size_t` as the template argument instead?

jonsimantov: Sure, will do.

var-const (approved):

PR #7892, opened 2021-04-13
Return FirebaseVersion as FIRRemoteConfigPodVersion

karenyz:
#no-changelog


FirebaseRemoteConfig/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Return FirebaseVersion as FIRRemoteConfigPodVersion. (#7892)
```
karenyz, line 2: not sure if we need a changelog since this is only used for internal logging

paulb777: Yep, no reason for an internal change release note.

paulb777 (approved):

PR #7893, opened 2021-04-14
Migrate to UserDefaults based heartbeat on tvOS

### LINKED ISSUE #6662: GULHeartbeatStorage: use user defaults instead of a file on tvOS
The cache directory (the only available for apps on tvOS) may be erased at any time, so is not a reliable place to store heartbeats. User defaults should be used instead.

Related PR: https://github.com/firebase/firebase-ios-sdk/pull/6658/files#diff-ba0d6f67c0194dcbcd6a90cfd4ee2a03R61

ryanwilson:
This should mitigate heartbeat files being wiped due to tvOS cache being
invalidated. Fixes #6662

#no-changelog

### DIFF
--- a/FirebaseCore/Sources/FIRHeartbeatInfo.m
+++ b/FirebaseCore/Sources/FIRHeartbeatInfo.m
@@ -17,6 +17,7 @@
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
 #import <GoogleUtilities/GULHeartbeatDateStorageUserDefaults.h>
 #import <GoogleUtilities/GULLogger.h>
+#import "FirebaseCore/Sources/Private/FIRAppInternal.h"
 
 const static long secondsInDay = 86400;
 @implementation FIRHeartbeatInfo : NSObject
@@ -30,8 +31,8 @@ const static long secondsInDay = 86400;
     NSString *const kHeartbeatStorageName = @"HEARTBEAT_INFO_STORAGE";
     id<GULHeartbeatDateStorable> dataStorage;
 #if TARGET_OS_TV
-    NSString *const kFIRCoreSuiteName = @"com.firebase.core";
-    NSUserDefaults *defaults = [[NSUserDefaults alloc] initWithSuiteName:kFIRCoreSuiteName];
+    NSUserDefaults *defaults =
+        [[NSUserDefaults alloc] initWithSuiteName:kFirebaseCoreDefaultsSuiteName];
     dataStorage =
         [[GULHeartbeatDateStorageUserDefaults alloc] initWithDefaults:defaults
                                                                   key:kHeartbeatStorageName];
--- a/FirebaseCore/Sources/FIRHeartbeatInfo.m
+++ b/FirebaseCore/Sources/FIRHeartbeatInfo.m
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #import "FirebaseCore/Sources/Private/FIRHeartbeatInfo.h"
+#import <GoogleUtilities/GULHeartbeatDateStorable.h>
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
+#import <GoogleUtilities/GULHeartbeatDateStorageUserDefaults.h>
 #import <GoogleUtilities/GULLogger.h>
 
 const static long secondsInDay = 86400;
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -125,7 +125,9 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      "7.2.1" ..< "8.0.0"
+      // TODO: Move from `revision` to version below before publishing.
+      .revision("a59fd0181d8f7c2b4c815cd6cddf4c38a4ba35ee")
+//      "7.4.0" ..< "8.0.0"
```
ryanwilson, line 130: This tag won't be available until we publish GoogleUtilities.

FirebaseCore/Sources/FIRHeartbeatInfo.m
```diff
@@ -25,9 +27,17 @@ @implementation FIRHeartbeatInfo : NSObject
  */
 + (BOOL)updateIfNeededHeartbeatDateForTag:(NSString *)heartbeatTag {
   @synchronized(self) {
-    NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
-    GULHeartbeatDateStorage *dataStorage =
-        [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
+    NSString *const kHeartbeatStorageName = @"HEARTBEAT_INFO_STORAGE";
+    id<GULHeartbeatDateStorable> dataStorage;
+#if TARGET_OS_TV
+    NSString *const kFIRCoreSuiteName = @"com.firebase.core";
```
ryanwilson, line 33: I can  stick with `standardUserDefaults` but felt like I should move  away from it. Thoughts on this domain?

Package.swift
ryanwilson: cc @paulb777 - how did we handle this in the past?

FirebaseCore/Sources/FIRHeartbeatInfo.m
maksymmalyhin: A custom domain looks reasonable as there is less probability that it will be modified by developers for their purposes. "com.firebase.core" is pretty generic name, so I would either declare a more specific one for the heartbeat use only or at least define a constant available for the rest of Core if we would like to reuse the same defaults for other purposes. I would prefer the more specific name.

maksymmalyhin (approved):
LGTM

Package.swift
paulb777: We either need to publish GoogleUtilities sooner or update this later in the release cycle.

There are some options to explore from https://forums.swift.org/t/best-practices-for-integration-testing-and-releasing-multiple-packages/47006/11, but nothing too clear or easy.

FirebaseCore/Sources/FIRHeartbeatInfo.m
ryanwilson: I did intentionally name it so Core could use it in the future for other things, since I don't think we need a new suite name for every piece of data we need to store (that's what the keys are for). Breaking it out into a constant would make sense in that case then, and other SDKs could follow suit eventually (`com.firebase.auth`, `.firestore`, etc).

Does that sound reasonable?

maksymmalyhin: SGTM. The only potential issue with reuse may is that the same defaults may potentially be modified from different parts of the code that may be easy to overlook. It seems to me that having extra user defaults is a relatively cheap mitigation of this issue. I'm fine with either for this PR.

Package.swift
ryanwilson: Happy to hold off temporarily and cherry-pick this when it's published. We should also probably set up a presubmit that prevents `.revision` from being checked in to prevent failures in the future

FirebaseCore/Sources/FIRHeartbeatInfo.m
ryanwilson: Yeah, we currently use a single user defaults (`standardUserDefaults`) anyways so this wouldn't be anything new, and we should provide a unique name for each usage point.

maksymmalyhin (approved):

ryanwilson:
Will wait until GoogleUtilities 7.4.0 publishes so we can have the proper `Package.swift` land here, then I'll merge. Thanks for the reviews!

paulb777:
I moved the milestone to 8.0.0 since the update did not make it into 7.11.0.

ncooke3 (approved):

ncooke3:
Kicked off a rerun of CI to test with GoogleUtils 7.4 (that is staged in SpecsStaging atm)

Package.swift
```diff
@@ -150,7 +150,7 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      "7.3.0" ..< "8.0.0"
+      "7.4.0" ..< "8.0.0"
```
ncooke3, line 153: @paulb777 @ryanwilson tagged GoogleUtils 7.4.0 for SwiftPM (on GoogleUtils repo) and am now pointing to GoogleUtils 7.4 in the Package manifest.

ryanwilson (commented):
LGTM - thanks for the update!

paulb777 (approved):

PR #7894, opened 2021-04-14
Merge branch 'combine-main' into master for better ease of development

peterfriese:
As discussed, let's merge combine-main into master to make ongoing development easier.

@lorenzofiamingo, @ziadtamim, and @mortenbekditlevsen FYI. Your working branched should update automatically once we close the `combine-main` branch.

Next steps:
1. Migrate tests to SwiftPM (@paulb777) 
    - Removing CocoaPods support
    - Make sure we're able to run the tests from SwiftPM
2. Need to verify we have test coverage across APIs
3. API review of existing public APIs + proposal to keep adding APIs under "community support" with @ryanwilson reviewing the new additions.


FirebaseCombineSwift/Sources/Auth/MultiFactor+Combine.swift
```diff
@@ -0,0 +1,104 @@
+// Copyright 2021 Google LLC
```
paulb777, line 1: I'd like to revisit the directory structure since I think we might prefer a product specific based directory structure like `FirebaseAuthCombine` to be consistent with the `FirebaseStorageSwift` structure already being used.

I'm fine to defer this refactor until after moving to master.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRHTTPSCallable.h
```diff
@@ -24,6 +24,8 @@ NS_SWIFT_NAME(HTTPSCallableResult)
 
 - (id)init NS_UNAVAILABLE;
 
+- (instancetype)initWithData:(id)result;
```
paulb777, line 27: @ryanwilson Does this need an API review?  Is there a workaround to do in the meantime? Is this part of a larger open question about exposing public initializers for testing?

Package.swift
```diff
@@ -39,6 +39,22 @@ let package = Package(
       name: "FirebaseAppDistribution-Beta",
       targets: ["FirebaseAppDistributionTarget"]
     ),
+    .library(
```
paulb777, line 42: In our meeting, we stated that the new library products weren't needed - but I'm not sure about that.  Maybe we need to add a little tool to find/insert products into the Package.swift for development?

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRHTTPSCallable.h
peterfriese: We might be able to replace this by using the Fakes infrastructure Maksym provided in `FirebaseTestingSupport`

Package.swift
peterfriese: Happy to remove them for now - LMK.

FirebaseCombineSwift/Sources/Auth/MultiFactor+Combine.swift
peterfriese: The reason for keeping everything under one umbrella folder was to make it easier to find the code (and any related documentation) that belongs together.

I can see benefits and drawbacks for both approaches. 

On a related note, would this refactoring be part of our endeavour to reduce the number of top-level folders (to bring the README.md above the fold), as suggested by @ncooke3 ?

paulb777: Understood  thanks. Let's defer this discussion until after the merge and the README.md and top-level folder structure plan is further along that can incorporate this open question.

Package.swift
paulb777: Perhaps, comment them out for now? @ryanwilson Any thoughts here?

paulb777:
LGTM, pending two open questions for @ryanwilson

paulb777:
BTW, I suggest doing the master merge from the command line to avoid squashing the branch history.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRHTTPSCallable.h
ryanwilson: Yes, this would need an API review but preferably it should use the same fakes infrastructure as `FirebaseTestingSupport` provides.

Package.swift
ryanwilson: Hmm, yeah commenting them out sounds good for now.

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRHTTPSCallable.h
peterfriese: Rolled back opening up the API, and used a fake to implement the Cloud Functions tests.

FirebaseCombineSwift.podspec
```diff
@@ -0,0 +1,74 @@
...
+
+  s.source           = {
+    :git => 'https://github.com/firebase/firebase-ios-sdk.git',
+    :tag => 'CocoaPods-' + s.version.to_s
+  }
+
+  s.social_media_url = 'https://twitter.com/Firebase'
+  s.swift_version         = '5.0'
+  s.ios.deployment_target = '13.0'
+  s.osx.deployment_target = '13.0'
```
paulb777, line 22: This is incorrect. Please update to use ruby variables like https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging.podspec#L23

FirebaseCombineSwift/Sources/Auth/MultiFactor+Combine.swift
paulb777: I think the compelling reason to change the organization will be the installation and build time of including a dependency on all Firebase products, especially Firestore. Let's sort out after the merge.

paulb777 (approved):
Thanks, this is close, and I'm fine if we address my remaining comments before or after merge.

Would you a short note to the top-level readme stating something like Combine support is under development and not available for production along with a pointer to the doc pages.

Package.swift
```diff
@@ -340,6 +357,33 @@ let package = Package(
         .headerSearchPath("../../.."),
       ]
     ),
+    .target(
```
paulb777, line 360: I believe these targets should be deleted in favor of the commented ones above.

PR #7896, opened 2021-04-14
Use private header import for FIRVersion.h

ryanwilson:
This was causing a failure in publishing the RemoteConfig pod.

#no-changelog

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNDevice.m
+++ b/FirebaseRemoteConfig/Sources/RCNDevice.m
@@ -19,9 +19,9 @@
 #import <sys/utsname.h>
 
 #import <GoogleUtilities/GULAppEnvironmentUtil.h>
+#import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
 #import "FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h"
 #import "FirebaseRemoteConfig/Sources/RCNConfigConstants.h"
-#import "FIRVersion.h"
 
 #define STR(x) STR_EXPAND(x)
 #define STR_EXPAND(x) #x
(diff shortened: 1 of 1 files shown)


FirebaseRemoteConfig/Sources/RCNDevice.m
```diff
@@ -19,9 +19,9 @@
 #import <sys/utsname.h>
 
 #import <GoogleUtilities/GULAppEnvironmentUtil.h>
-#import "FirebaseCore/Sources/Public/FirebaseCore/FIRVersion.h"
 #import "FirebaseRemoteConfig/Sources/Private/RCNConfigSettings.h"
 #import "FirebaseRemoteConfig/Sources/RCNConfigConstants.h"
+#import "FIRVersion.h"
```
maksymmalyhin, line 24: I think it should be either
```
#import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
```
or
```
#import <FirebaseCore/FirebaseCore.h>
```

paulb777: Yes, either should work, but the normal pattern in the repo is the former.

ryanwilson: I misread the docs! I caught the second half of the exception but not the first. I had the first suggestion you listed before but then thought I followed the right rule... whoops :) nice catch, thanks.

maksymmalyhin (approved):

paulb777 (approved):

ryanwilson:
cc @karenyz as an FYI

karenyz:
oh sorry about that!

ryanwilson:
> oh sorry about that!

No problem! We should have the tooling that catches this - it passed both review and CI so that's definitely not on you 😄

PR #7901, opened 2021-04-14
Only update fetch throttle interval for server errors

### LINKED ISSUE #6628: When the remoteConfig load fails on the first call, the SDK returns the code 8002 (throttled) for the other calls
### **[REQUIRED] Step 1: Describe your environment**

Xcode version: Xcode 12.0.1
Firebase SDK version: 6.33.0
Firebase Component: Remote Config
Component version: 4.9.0
Installation method: CocoaPods
Platform: iOS

### **[REQUIRED] Step 2: Describe the problem**

When I call the fetchAndActivate method and the phone has no network then I reactivate the network to restart the request I get an error 8002 (throttled) while I have no problem when on the first call the network is present.

When the problem occurs I have to wait several minutes for the request to work again.

This is problematic because I want when the minimumFetchInterval is reached the request to the Firebase servers is a success to allow access to my application.

How can I solve this problem?

Here is my code:

```
settings.fetchTimeout = 30
settings.minimumFetchInterval = 60*60
#if DEBUG
settings.minimumFetchInterval = 0
#endif
remoteConfigService.configSettings = settings

remoteConfigService.fetchAndActivate(completionHandler: { (status, error) -> Void in
	if status != .error {
		// Load my config
	} else {
	 	// Manage error
	}
})
```

Thank you in advance.

karenyz:
fixes #6628

### DIFF
--- a/FirebaseRemoteConfig/CHANGELOG.md
+++ b/FirebaseRemoteConfig/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Fixed throttling issue when fetch fails due to no network. (#6628)
+
 # v7.10.0
 - [changed] Throw exception if projectID is missing from FirebaseOptions. (#7725)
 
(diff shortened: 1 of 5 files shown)


paulb777 (approved):
Thanks!

chliangGoogle (approved):

PR #7910, opened 2021-04-16
Fix zip build with new Analytics subspecs.

ryanwilson:
Will need to verify that this includes the correct Analytics subspecs - but it should.

#no-changelog


ncooke3 (approved):
Nice!

PR #7912, opened 2021-04-16
Fix FIAM resources for CocoaPods

ncooke3:
Fix for internal b/185599263


christibbs (approved):

paulb777 (commented):
Can be done after, but is it possible to add a test in CI to catch this?

ncooke3:
@paulb777 Ill add tests to verify the resources are bundled correctly for SPM and CocoaPods in a follow-up PR!

PR #7920, opened 2021-04-19
Update CHANGELOG for m94 relesae

zwu52:

### DIFF
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2021-04 -- v7.11.1
+- [changed] Update to clang-format 12. (#7909)
+
 # 2021-04 -- v7.11.0
 - [changed] Refactor Messaging to internally not depending on InstanceID, but can co-exist. Will remove InstanceID dependency in the next Firebase breaking change. (#7814)
 


FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2021-04 -- v7.11.1
```
chliangGoogle, line 1: I think this should be under v7.11.0. Thanks!

chliangGoogle (changes requested):

FirebaseMessaging/CHANGELOG.md
paulb777: It's a Firebase wide change so it should go to FirebaseCore/CHANGELOG.

Also we typically don't highlight formatting changes in Changelogs.

morganchen12: 👍 feel free to omit this change since 7.11.0 is not empty and clang-format changes aren't noticeable by users unless they go read the source.

paulb777: And it's probably sufficient for those reading the source to check the git logs. 

In summary, thanks for the PR, but recommend closing without merging.

zwu52: got it. Thanks for the inputs here folks :)

PR #7930, opened 2021-04-21
Remove unbalanced clang format pop

ryanwilson:
This *pop*ped up during copybara imports.

#no-changelog


chliangGoogle (approved):

chliangGoogle:
I have an internal cl/369681515 copybara cuz this version might break 1P so I need to re-add iid dependency for them. I will copybara in your change after submission.

PR #7933, opened 2021-04-21
Skip test_coverage when a PR is closed

granluo:
Reference:
pull request actions and merge:
https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#pull_request

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -297,7 +297,7 @@ jobs:
         with:
           path: /Users/runner/test
       - name: Compare Diff and Post a Report
-        if: github.event.pull_request.merged != true && env.METRICS_SERVICE_SECRET
+        if: github.event.pull_request.merged != true && github.event.action != 'closed' && env.METRICS_SERVICE_SECRET
         env:
           base_commit: ${{ needs.check.outputs.base_commit }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test_coverage.yml
```diff
@@ -297,7 +297,7 @@ jobs:
         with:
           path: /Users/runner/test
       - name: Compare Diff and Post a Report
-        if: github.event.pull_request.merged != true && env.METRICS_SERVICE_SECRET
+        if: github.event.pull_request.merged != true && github.event.action != 'closed' && env.METRICS_SERVICE_SECRET
```
paulb777, line 300: Good find! Should the condition be added to an earlier place like line 283 or any of the other jobs, like the `check` job?

paulb777 (approved):

.github/workflows/test_coverage.yml
granluo: `check` will only run when github.event.action == 'opened' || github.event.action == 'synchronize'
pod-lib-lint tests will run if `check` succeed, and so this condition doesn't have to be applied to jobs above.
`Access to Metrics Service` is a preceding job to both `Compare Diff and Post a Report` for presubmit and  `Update New Coverage Data` for postsubmit, so it should be either 1. not closed or 2, closed but merged is true.

PR #7934, opened 2021-04-22
Add podspec presubmit tests.

granluo:


granluo:
This is to set FirebaseDatabase.podspec for presubmit and update post submit to let all changed podspecs from master head can be pushed to [firebase/specstesting ](https://github.com/firebase/SpecsTesting) repo

FirebaseDatabase.podspec
```diff
@@ -88,3 +88,97 @@ Simplify your iOS development, grow your user base, and monetize more effectivel
     int_tests.resources = 'FirebaseDatabase/Tests/Resources/GoogleService-Info.plist'
   end
 end
+
```
paulb777, line 91: Why the changes in this file?

.github/workflows/database.yml
```diff
@@ -137,3 +137,17 @@ jobs:
...
+    runs-on: macOS-latest
+    strategy:
+      matrix:
+        target: [ios, tvos, macos, watchos]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: Build and test
+      run: scripts/third_party/travis/retry.sh pod spec lint FirebaseDatabase.podspec --skip-tests --platforms=${{ matrix.target }} --sources='https://github.com/firebase/SpecsTesting','https://github.com/firebase/SpecsDev.git','https://github.com/firebase/SpecsStaging.git','https://cdn.cocoapods.org/'
```
paulb777, line 153: I don't understand the reason for only `FirebaseDatabase.podspec` here? Is it to verify with FirebaseDatabase first before adding the others?

.github/workflows/prerelease.yml
```diff
@@ -2,9 +2,6 @@ name: prerelease
 
 on:
   pull_request:
-    paths:
-    - '.github/workflows/prerelease.yml'
```
paulb777, line 6: Why are we deleting the paths here? Isn't it still useful to trigger the run when prerelease.yml changes?

granluo: The workflow might  erase the testing podspecs repo and push all podspecs. This process takes three hours and  might fail presubmit tests.

.github/workflows/database.yml
granluo: This is to set up database first and expand to other SDKs in another PR.

FirebaseDatabase.podspec
granluo: Have reverted it

paulb777 (approved):

PR #7937, opened 2021-04-22
Integrate App Check with Firebase umbrella pod 

maksymmalyhin:
- App Check added to Firebase.h
- App Check added Firebase pod
- App Check added to FirebaseManifest.swift
- App Check adoption of Firebase version
- App Check added to Dangerfile 
- disable tests for `pod lib lint` with `--use-libraries` and `--use-modular-headers` options


morganchen12 (approved):
LGTM, but please wait for other approvers before merging.

paulb777 (approved):
LGTM!

Can be done separately - appcheck should also be added to the GHA dashboard in https://github.com/firebase/firebase-ios-sdk/blob/master/README.md

maksymmalyhin:
> LGTM!
> 
> Can be done separately - appcheck should also be added to the GHA dashboard in https://github.com/firebase/firebase-ios-sdk/blob/master/README.md

Will do in a separate PR, thanks!

PR #7942, opened 2021-04-22
Remove instanceID dependency in Messaging

chliangGoogle:
Messaging no longer depends on InstanceID. Remove it for the breaking change release.


paulb777 (approved):

FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [changed] Remove InstanceID dependency from Messaging. This only breaks for Messaging users who are using InstanceID but not specifying InstanceID in their podfile or Swift Package Manager. Please consider migrate to Messaging's token APIs if you are using InstanceID to manage FCM registration token. (#7836)
```
egilmorez, line 2: Can I suggest a big blanket edit?  

"Remove Instance ID dependency from FCM. This only breaks FCM users who are using Instance ID to manage registration tokens without explicitly specifying Instance ID in their podfile or Swift Package Manager. Consider migrating to FCM's token APIs if you are using Instance ID to manage FCM registration tokens.

-- Unless we mean `InstanceID` as a literal, I suggest just "Instance ID"
-- Do you think we should cite the actual FCM API, maybe in parentheses?

egilmorez (commented):
Thanks Chen!  PTAL at my suggestion/questions.

FirebaseMessaging/CHANGELOG.md
egilmorez: Looking again, I wonder if we should also mention the obvious quick workaround as well as the migration recommendation.  Like:

"Remove Instance ID dependency from FCM. This is a breaking change for FCM users who are use Instance ID to manage registration tokens without explicitly specifying Instance ID in their podfile or Swift Package Manager. 

For a short-term resolution, those FCM users can update their app configuration to explicitly specify Instance ID. Longer term, all users should migrate to FCM's token APIs.

chliangGoogle: Updated. Thanks!

chliangGoogle: I'm going to use Messaging instead of FCM here btw.

chliangGoogle:
Submit this PR since it's only change in changelog

PR #7943, opened 2021-04-23
Add missing protocol to fix Xcode 12.5 warning

paulb777:
#no-changelog

### DIFF
--- a/FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
+++ b/FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
@@ -217,9 +217,7 @@ NS_ASSUME_NONNULL_BEGIN
 + (id<FIRDatabaseConnectionContextProvider>)
     contextProviderWithAuth:(nullable id<FIRAuthInterop>)auth
                    appCheck:(nullable id<FIRAppCheckInterop>)appCheck {
-    return (id<FIRDatabaseConnectionContextProvider>)[[self alloc]
-        initWithAuth:auth
-            appCheck:appCheck];
+    return [[self alloc] initWithAuth:auth appCheck:appCheck];
 }
 
 @end
--- a/FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
+++ b/FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
@@ -217,7 +217,9 @@ NS_ASSUME_NONNULL_BEGIN
 + (id<FIRDatabaseConnectionContextProvider>)
     contextProviderWithAuth:(nullable id<FIRAuthInterop>)auth
                    appCheck:(nullable id<FIRAppCheckInterop>)appCheck {
-    return [[self alloc] initWithAuth:auth appCheck:appCheck];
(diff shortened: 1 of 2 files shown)


FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
```diff
@@ -217,7 +217,9 @@ - (void)listenForAppCheckTokenChanges:(fbt_void_nsstring)listener {
 + (id<FIRDatabaseConnectionContextProvider>)
     contextProviderWithAuth:(nullable id<FIRAuthInterop>)auth
                    appCheck:(nullable id<FIRAppCheckInterop>)appCheck {
-    return [[self alloc] initWithAuth:auth appCheck:appCheck];
+    return (id<FIRDatabaseConnectionContextProvider>)[[self alloc]
```
maksymmalyhin, line 220: Thank you for catching this. I think we should add confirmation of `FIRDatabaseConnectionContextProvider` class to `FIRDatabaseConnectionContextProvider` protocol in the header instead of casting. It seems that originally the method could return different classes but it is not the case any more, so an explicit confirmation looks reasonable to me.

cc @schmidt-sebastian

paulb777: @maksymmalyhin It looks like the unit tests depend on the different classes: 

Screen Shot 2021-04-23 at 7 57 28 AM

maksymmalyhin: Yeah, that's correct. Sorry if I was not clear. Let me try to explain my thoughts.

The warning with the original method implementation is actually correct an helpful, because the method is supposed to return an instance that conforms to the `FIRDatabaseConnectionContextProvider` **protocol**, but in fact it returns an instance of `FIRDatabaseConnectionContextProvider` **class** (the same name as the protocol to add more confusion 😄 ) which doesn't declare it's conforming to the protocol. My suggested fix for this is adding the protocol confirmation to the class, e.g. by modifying https://github.com/firebase/firebase-ios-sdk/blob/e4dcb855b534f02a1dcff947af9fcb055ce35af5/FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.h#L59 to the following:
```
@interface FIRDatabaseConnectionContextProvider : NSObject <FIRDatabaseConnectionContextProvider>
```
WDYT?

paulb777: Thanks for explaining. That makes sense and is much cleaner to avoid the casts.

I was confused by the same name for both the protocol and the interface. Is that a typical style?

schmidt-sebastian (approved):

FirebaseDatabase/Sources/Login/FIRDatabaseConnectionContextProvider.m
maksymmalyhin: I usually prefer to add `Protocol` suffix if there is a name conflict with a class, but here I followed the existing name convention and only applied renaming "AuthTokenProvider" -> "ConnectionContextProvider". I actually don't mind renaming the protocol.

maksymmalyhin (approved):
Thank you!

PR #7954, opened 2021-04-23
Update sdk workflows to involve spec tests.

granluo:


granluo:
Should we add `pod spec lint` for `segmentation` podspec, which is excluded in nightly release testing.

paulb777:
Segmentation does not need to be included. We're maintaining the definitive list at https://github.com/firebase/firebase-ios-sdk/blob/master/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift

.github/workflows/core.yml
```diff
@@ -89,3 +89,17 @@ jobs:
...
     - name: PodLibLint Core Cron
       run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseCore.podspec --platforms=${{ matrix.target }} ${{ matrix.flags }}
+
+  podspec-presubmit:
+    # Don't run on private repo unless it is a PR.
+    if: github.repository == 'Firebase/firebase-ios-sdk' && github.event.pull_request.merged != true
+    runs-on: macOS-latest
+    strategy:
+      matrix:
+        target: [ios, tvos, macos, watchos]
```
paulb777, line 99: For `pod spec lint` tests we could skip the matrix and the `--platforms` option and just always do the default of all platforms specified in the podspec.

paulb777 (approved):
LGTM with the simplification of deleting the matrixes and the `--platforms` option.

granluo:
`--allow-warnings` is set in `pod lib lint` and will be set for firestore `pod spec lint` due to the below failure.
https://github.com/firebase/firebase-ios-sdk/runs/2426219724

PR #7956, opened 2021-04-24
Bump zip build minimum version options

paulb777:
The zip builder does not support mixed minimum platform versions. Since AppCheck requires iOS 11.0, bumping the zip and Carthage distribution minimum iOS versions.


paulb777:
I'll explore a less radical fix today.

paulb777:
Closing in favor of #7966

PR #7966, opened 2021-04-26
Disable CocoaPods minimum version install testing for Firebase zip build

paulb777:
This is a more flexible approach to allowing different Firebase pods with different platform version support than #7956.

It's safe because the actual build minimum version is controlled by the podspec. The minimum version that is being disregarded here was only for insuring a full installation had a consistent minimum version.

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/main.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/main.swift
@@ -230,11 +230,6 @@ struct ZipBuilderTool: ParsableCommand {
       fatalError("Missing template inside of the repo. \(templateDir) does not exist.")
     }
 
-    // Set the platform minimum versions.
-    PlatformMinimum.initialize(ios: minimumIOSVersion,
-                               macos: minimumMacOSVersion,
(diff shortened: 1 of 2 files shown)


ReleaseTooling/Sources/ZipBuilder/main.swift
```diff
@@ -300,6 +300,14 @@ struct ZipBuilderTool: ParsableCommand {
       }
     } else {
       // Do a Firebase Zip Release package build.
+
+      // For the Firebase zip distribution, we disable version checking at install time by
+      // setting a high version to install. The minimum versions are controlled by each individual
+      // pod's podspec options.
+      PlatformMinimum.initialize(ios: "14.0",
+                                 macos: "11.0",
+                                 tvos: "14.0")
```
ryanwilson, line 309: Do we want to make a new static method on `PlatformMinimum` like `PlatformMinimum.useLatestMajorVersions()` or something and have those versions defined there instead of here?

ryanwilson (approved):

PR #7967, opened 2021-04-26
Update CHANGELOG.md

elenadoty:


samedson (approved):

Crashlytics/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- [changed] Added a warning to upload-symbols when it detects a dSYM with hidden symbols.
```
paulb777, line 2: Matches the 7.10.0 release note?

elenadoty: This check is for symbols that show up in the dSYM as "__hidden#XX__" (a byproduct of bitcode) while the previous checks for dSYMs without any symbols.

Issue #7962, opened 2021-04-26
Flaky installation of GoogleAppMeasurement in 7.11 with SwiftPM


ryanwilson:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.5
  * Firebase SDK version: 7.11.0
  * Installation method: `Swift Package Manager`
  * Firebase Component: Analytics

#### Steps to reproduce:

From a blank project in Xcode 12.5, attempt to add Firebase via SwiftPM.
Occasionally, this fails to reproduce with the message:

```
artifact of binary target 'GoogleAppMeasurement' failed extraction: The operation couldn't be completed. (TSCBasic.StringError error 1.)
artifact of binary target 'GoogleAppMeasurementWithoutAdIdSupport' failed extraction: The operation couldn't be completed. (TSCBasic.StringError error 1.)
```

#### Relevant Code:

No code needed, error occurs in Xcode during installation.

![Screenshot of error in Xcode](https://user-images.githubusercontent.com/1097316/116090266-ac781c00-a671-11eb-9168-f244a9c35d1e.png)

ryanwilson:
This is likely the underlying implementation of the unzipping: 
https://github.com/apple/swift-tools-support-core/blob/435a2708a6e486d69ea7d7aaa3f4ad243bc3b408/Sources/TSCUtility/Archiver.swift#L65

Will try to work off that and reproduce.

ryanwilson:
Using the `unzip` command an extra directory `__MACOSX` shows up. I suspect this is due to an incorrect compression of the release artifacts, and Xcode gets confused when there are 2 `XCFramework`s nested there. This could also explain the flakiness in that sometimes the proper `XCFramework` gets selected.

![Duplicate folders in finder](https://user-images.githubusercontent.com/1097316/116103047-3a0d3900-a67d-11eb-9187-a9b8a219ef87.png)

ryanwilson:
7.11.1 for SwiftPM was released and this appears to be resolved.

morganchen12:
@ryanwilson does this warrant a release note?

ryanwilson:
> @ryanwilson does this warrant a release note?

Probably - but let's hold off momentarily as I just had the issue happen again.

paulb777:
This seems to have stabilized with some combination of recent Xcodes and Firebase releases. I'll close for now and we can reopen if it reoccurs.

PR #7972, opened 2021-04-26
FirebaseAnalytics 7.11.1 release.

ryanwilson:
This includes the same artifact as 7.11.0 but put into the proper .zip format (with zip on the command line instead of Finder's compress menu option).

#no-changelog

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@
 
 import PackageDescription
 
-let firebaseVersion = "7.11.0"
+let firebaseVersion = "7.11.1"
 
 let package = Package(
   name: "Firebase",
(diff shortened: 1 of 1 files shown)


maksymmalyhin (approved):
LGTM on CI green

paulb777 (approved):

ryanwilson:
I was able to test this via Xcode, depending on the `rw-7.11.1-patch` branch and successfully imported Firebase and compiled Analytics calls.

ryanwilson:
This will resolve #7962 after this lands and is tagged.

ryanwilson:
Cancelled the test_coverage for Firestore since it should be unaffected.

Issue #7648, opened 2021-03-03
Remote Config won't return the correct values based on a set user property value


zeusent:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.4
  * Firebase SDK version: > 7.1.0
  * Installation method: `Swift Package Manager`
  * Firebase Component: RemoteConfig

#### Steps to reproduce:

If I update Firebase to any version bigger than 7.1.0 then RemoteConfig stops working as expected.
My app uses an user property value to grab different values from RemoteConfig.

With any version bigger than 7.1.0 this user property value is ignored and RemoteConfig always returns the same
values no matter what. Whenever I rollback to 7.1.0 then everything works as expected.

If I look in the analytics debug view I can see my user property being set correctly.

#### Relevant Code:

On app start, before grabbing any RemoteConfig values I do this:
```
Analytics.setUserProperty("dummy_value", forName: "dummy_property")
```

Immediately after that I call `fetchAndActivate`. This works as expected on 7.1.0 but fails on any newer versions.

paulb777:
@zeusent Sorry about the trouble and thanks for the report and narrowing down the version.

@karenyz Please take a look. I suspect a regression related to #6692.

jklein24:
Hi folks, I believe this is affecting us still. Was there any progress here?

jklein24:
So It's been a bit unreliable and tricky to reproduce, but what we're seeing is essentially the following setup is inconsistent:

- We have one param which has its value set conditionally to true based on a user property
- We have an experiment whose allocation *excludes* users with that user property
- Somehow we're still seeing users with this user property get the param set to false sometimes (but not always). It also seems to change arbitrarily on occasion.

To make it more concrete with an example:
[...]

PR #7975, opened 2021-04-27
Block calls to get config until DB load completes

### LINKED ISSUE #7424: [Remote Config] Local conf randomly empty even with successful fetch

  * Xcode version: 12.2
  * Firebase SDK version: 7.3.0

We use the 3rd loading strategy (Load new values for next startup) as described here https://firebase.google.com/docs/remote-config/loading
Our application fetches the required configuration on the splash screen.

On the first launch, this fetch succeed and the app behaves normally.
For some reason, on other launch the local configuration sometimes returns empty values.
The Firebase app is configured and the logs doesn't identify any issue even if set to max.

**The problem isn't easily reproducible on all devices, but some of them are failing almost 1 on 3 times.**

We Unfortunately couldn't narrow the devices concerned, neither the OS concerned.

When the content is empty :
- The last fetch status is `succeed`
- `defaultExpiration: TimeInterval = 600`

- When we print the value of :

RemoteConfig.remoteConfig().allKeys(from: .remote)
RemoteConfig.remoteConfig().allKeys(from: .static)
RemoteConfig.remoteConfig().allKeys(from: .default)

It returns 0 (no key was found)

#### Steps to reproduce:

Please find below the link to the sample where you can reproduce : 

https://drive.google.com/file/d/1LG4fbE4FCW4Ts61vGgvzQLUW0l6jSH4s/view?usp=sharing

To reproduce, simply install the sample, kill & launch it multiple times. Once the background is red, it means that the local conf is empty.

karenyz:
`_configLoadFromDBSemaphore = dispatch_semaphore_create(1)` causes the later `dispatch_semaphore_wait` call in `checkAndWaitForInitialDatabaseLoad()` to never be blocked, so it's possible to return a config before we finish loading from the DB.

Removing `dispatch_semaphore_signal` from `loadPersonalization` as its success/failure should not impact RC returning a config

fixes #7424

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigContent.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigContent.m
@@ -122,6 +122,7 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
                    dispatch_semaphore_signal(self->_configLoadFromDBSemaphore);
                  }];
 
+  // TODO(karenzeng): Refactor personalization to be returned in loadMainWithBundleIdentifier above
   [_DBManager loadPersonalizationWithCompletionHandler:^(
                   BOOL success, NSDictionary *fetchedPersonalization,
                   NSDictionary *activePersonalization, NSDictionary *defaultConfig) {
--- a/FirebaseRemoteConfig/Sources/RCNConfigContent.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigContent.m
@@ -86,8 +86,7 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
       _bundleIdentifier = @"";
     }
     _DBManager = DBManager;
-    // Waits for both config and Personalization data to load.
-    _configLoadFromDBSemaphore = dispatch_semaphore_create(1);
+    _configLoadFromDBSemaphore = dispatch_semaphore_create(0);
     [self loadConfigFromMainTable];
   }
   return self;
@@ -128,7 +127,6 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
                   NSDictionary *activePersonalization, NSDictionary *defaultConfig) {
     self->_fetchedPersonalization = [fetchedPersonalization copy];
     self->_activePersonalization = [activePersonalization copy];
-    dispatch_semaphore_signal(self->_configLoadFromDBSemaphore);
   }];
 }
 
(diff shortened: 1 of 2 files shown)


FirebaseRemoteConfig/Sources/RCNConfigContent.m
```diff
@@ -128,7 +127,6 @@ - (void)loadConfigFromMainTable {
                   NSDictionary *activePersonalization, NSDictionary *defaultConfig) {
     self->_fetchedPersonalization = [fetchedPersonalization copy];
     self->_activePersonalization = [activePersonalization copy];
```
chliangGoogle, line 129: Since we only have one dispatch_semaphore_signal blocking inside `[DBManager loadMainWithBundleIdentifier:]` to block the read, should we move this `[DBManager loadPersonalizationWithCompletionHandler:]` inside `[DBManager loadMainWithBundleIdentifier:]` as well?

chliangGoogle (changes requested):

paulb777 (commented):
Is this also a fix for #7648?

FirebaseRemoteConfig/Sources/RCNConfigContent.m
karenyz: Feels a bit weird to chain calls to `DBManager` and add extra latency, especially since the personalization data is separate from the RC template and isn't returned as part of the config.

I think we could refactor `[DBManager loadMainWithBundleIdentifier:]` to return personalization data along with the config (in one call), but that's a larger change and we can tackle it as a follow up. The current change will fix the reported issue, and doesn't change current personalization behavior (we already aren't blocking)

chliangGoogle: my concern is if developers try to read fetchedPersonalization then it might not be ready. I'm not familiar with personalization for but if we are okay that they might not be fully loaded when developers trying to read configs, i'm okay give it good to go.

chliangGoogle: Can you add a todo here on refactor?

chliangGoogle (approved):

FirebaseRemoteConfig/Sources/RCNConfigContent.m
danasilver: Developers don't read the `fetchedPersonalization`s so it should be okay if they're not loaded when developers try to read configs. They're used to report personalization signals to GA when a personalized parameter (stored in the config that's read above and waited on) is read. More signals improve the personalizations but some missed signals don't break the system.

danasilver (approved):

FirebaseRemoteConfig/Sources/RCNConfigContent.m
karenyz: added a TODO for the personalization refactor

chliangGoogle (approved):

paulb777 (approved):

PR #7986, opened 2021-04-27
NanoPB utilities for FieldValue migration

schmidt-sebastian:

This PR adds a couple of new classes and methods that are used in the FieldValue port.  

Example usage:

- `SharedMessage` is a shared pointer around `Message`, which allows a single Protobuf value to be used in multiple classes. This makes it possible for a Query and a Target to contain the same Filter and Bounds. See https://github.com/firebase/firebase-ios-sdk/pull/7904/files?file-filters%5B%5D=.cc&file-filters%5B%5D=.h&file-filters%5B%5D=.mm&file-filters%5B%5D=.txt#diff-ca2b3b9278f6b19636908d4f2da32d8657081510e6ffb60b823d5766992bf263R87

- `SetRepeatedField` simplifies initialization of nanopb repeated fields. See https://github.com/firebase/firebase-ios-sdk/pull/7904/files?file-filters%5B%5D=.cc&file-filters%5B%5D=.h&file-filters%5B%5D=.mm&file-filters%5B%5D=.txt#diff-68ee4b0277a710dcf518d3f8aba6cd6190ae73dc04840aee102c7a6ba5627944R368

- `ReleaseFieldArray` is used to transform ownership of nanopb objects from the Serializer to ObjectValue (and hence the documents): https://github.com/firebase/firebase-ios-sdk/pull/7904/files?file-filters%5B%5D=.cc&file-filters%5B%5D=.h&file-filters%5B%5D=.mm&file-filters%5B%5D=.txt#diff-243c7453f0512236aeb3cb21fdc2ea18b9c280b05bf7f8a105c32b82b823a145R229

- `ResizeFieldArray` is only used in test to allow the tests to build up maps incrementally: https://github.com/firebase/firebase-ios-sdk/pull/7904/files?file-filters%5B%5D=.cc&file-filters%5B%5D=.h&file-filters%5B%5D=.mm&file-filters%5B%5D=.txt#diff-f89aa393ca7ce959593fe159bff272ef460b58a99592b3b8563ca73b446083f0R163
-

### DIFF
--- a/Firestore/Example/Firestore.xcodeproj/project.pbxproj
+++ b/Firestore/Example/Firestore.xcodeproj/project.pbxproj
@@ -28,7 +28,6 @@
 		048A55EED3241ABC28752F86 /* memory_mutation_queue_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = 74FBEFA4FE4B12C435011763 /* memory_mutation_queue_test.cc */; };
 		04D7D9DB95E66FECF2C0A412 /* bundle_cache_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = F7FC06E0A47D393DE1759AE1 /* bundle_cache_test.cc */; };
 		0500A324CEC854C5B0CF364C /* FIRCollectionReferenceTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 5492E045202154AA00B64F25 /* FIRCollectionReferenceTests.mm */; };
-		051D3E20184AF195266EF678 /* no_document_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = AB6B908720322E8800CC290A /* no_document_test.cc */; };
 		0535C1B65DADAE1CE47FA3CA /* string_format_apple_test.mm in Sources */ = {isa = PBXBuildFile; fileRef = 9CFD366B783AE27B9E79EE7A /* string_format_apple_test.mm */; };
 		056542AD1D0F78E29E22EFA9 /* grpc_connection_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = B6D9649021544D4F00EB9CFB /* grpc_connection_test.cc */; };
 		0575F3004B896D94456A74CE /* status_testing.cc in Sources */ = {isa = PBXBuildFile; fileRef = 3CAA33F964042646FDDAF9F9 /* status_testing.cc */; };
@@ -54,12 +53,10 @@
 		0A4E1B5E3E853763AE6ED7AE /* grpc_stream_tester.cc in Sources */ = {isa = PBXBuildFile; fileRef = 87553338E42B8ECA05BA987E /* grpc_stream_tester.cc */; };
 		0A52B47C43B7602EE64F53A7 /* cc_compilation_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = 1B342370EAE3AA02393E33EB /* cc_compilation_test.cc */; };
 		0A6FBE65A7FE048BAD562A15 /* FSTGoogleTestTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 54764FAE1FAA21B90085E60A /* FSTGoogleTestTests.mm */; };
-		0A800CA749750B01E36A6787 /* field_value_benchmark.cc in Sources */ = {isa = PBXBuildFile; fileRef = 6D0EE49C1D5AF75664D0EBE4 /* field_value_benchmark.cc */; };
 		0ABCE06A0D96EA3899B3A259 /* query_engine_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = B8A853940305237AFDA8050B /* query_engine_test.cc */; };
 		0AE084A7886BC11B8C305122 /* string_util_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = AB380CFC201A2EE200D97691 /* string_util_test.cc */; };
 		0B002E2E2012B32EB801C6D5 /* bundle_spec_test.json in Resources */ = {isa = PBXBuildFile; fileRef = 79EAA9F7B1B9592B5F053923 /* bundle_spec_test.json */; };
 		0B071E9044CEEF666D829354 /* field_filter_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = E8551D6C6FB0B1BACE9E5BAD /* field_filter_test.cc */; };
-		0B4D2668C1E81DF6D62BA9BF /* field_value_benchmark.cc in Sources */ = {isa = PBXBuildFile; fileRef = 6D0EE49C1D5AF75664D0EBE4 /* field_value_benchmark.cc */; };
 		0B55CD5CB8DFEBF2D22A2332 /* byte_stream_cpp_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = 01D10113ECC5B446DB35E96D /* byte_stream_cpp_test.cc */; };
 		0B7B24194E2131F5C325FE0E /* async_queue_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = B6FB467B208E9A8200554BA2 /* async_queue_test.cc */; };
 		0B9BD73418289EFF91917934 /* bits_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = AB380D01201BC69F00D97691 /* bits_test.cc */; };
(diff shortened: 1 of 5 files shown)


Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
...
+T* _Nonnull ResizeArray(void* _Nullable ptr, pb_size_t count) {
+  return static_cast<T*>(realloc(ptr, count * sizeof(T)));
+}
+
+/**
+ * Initializes a repeated field with a list of values. Applies `converter` to
+ * each value before assigning.
+ */
+template <typename T, typename Iterator, typename Func>
+void SetRepeatedField(T** fields_array,
```
var-const, line 117: Optional: returning `absl::span` could allow you to get rid of the output parameters. Consider something like:
```cpp
template <typename Iterator, typename Func>
auto SetRepeatedField(Iterator first, Iterator last, const Func& converter)
    -> absl::span<decltype(converter())> {
  using T = decltype(converter());

  auto size = CheckedSize(std::distance(first, last));
  absl::span<T> result{MakeArray<T>(size), size};
  for (auto* e : result) {
    *e = converter(*first);
    ++first;
  }

  return result;
}
```

Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
...
+/**
+ * Initializes a repeated field with a list of values. Applies `converter` to
+ * each value before assigning.
+ */
+template <typename T, typename Container, typename Func>
+void SetRepeatedField(T** fields_array,
+                      pb_size_t* fields_count,
+                      const Container& fields,
+                      Func converter) {
+  *fields_count = CheckedSize(fields.size());
```
var-const, line 141: It seems like you could just delegate to the other overload:
```cpp
  return SetRepeatedField(fields_array, fields_count, fields.begin(), fields.end(), converter);
```

Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
...
+/**
+ * Initializes a repeated field with a list of values. Applies `converter` to
+ * each value before assigning.
+ */
+template <typename T, typename Iterator, typename Func>
+void SetRepeatedField(T** fields_array,
+                      pb_size_t* fields_count,
+                      Iterator first,
+                      Iterator last,
+                      Func converter) {
```
var-const, line 121: Nit: you can pass `converter` by const reference. Since this is a template parameter, it could in theory stand for, say, a lambda with many captures.

Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
...
+}
+
+/**
+ * Zeroes out the memory of `fields`. This can be used if the contents of fields
+ * array are moved to another message that takes on ownership.
+ */
+template <typename T>
+void ReleaseFieldsArray(T* _Nonnull fields, pb_size_t fields_count) {
+  for (pb_size_t i = 0; i < fields_count; ++i) {
+    fields[i] = {};
```
var-const, line 166: Optional: `nullptr` might be a little more obvious here than `{}` (which can stand for a lot of things).

Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
...
+  return SetRepeatedField(fields_array, fields_count, fields,
+                          [](const T& val) { return val; });
+}
+
+/**
+ * Zeroes out the memory of `fields`. This can be used if the contents of fields
+ * array are moved to another message that takes on ownership.
+ */
+template <typename T>
+void ReleaseFieldsArray(T* _Nonnull fields, pb_size_t fields_count) {
```
var-const, line 164: Nit: the name can be understood as releasing (i.e., freeing) the array itself, not its members. Also, "release" can mean "free" rather than "let go of ownership". Consider something like `DisownFieldsArrayMembers`.

Firestore/core/src/nanopb/message.h
```diff
@@ -186,6 +187,49 @@ class Message {
   T proto_{};
 };
 
+/**
+ * A wrapper of Message objects that facilitates shared ownership of non-mutable
```
var-const, line 191: Ultranit: `s/non-mutable/immutable/`.

Firestore/core/src/nanopb/message.h
```diff
@@ -186,6 +187,49 @@ class Message {
...
+   * Creates a valid `SharedMessage` that wraps a value-constructed ("zeroed
+   * out") Nanopb proto.
+   */
+  SharedMessage() = default;
+
+  /**
+   * Creates a `SharedMessage` object that wraps `proto`. Takes ownership of
+   * `proto`.
+   */
+  explicit SharedMessage(const T& proto) : message_{new Message<T>{proto}} {
```
var-const, line 207: Nit: can you use `make_shared` here?

Firestore/core/src/nanopb/message.h
```diff
@@ -186,6 +187,49 @@ class Message {
...
 
+/**
+ * A wrapper of Message objects that facilitates shared ownership of non-mutable
+ * Protobuf data.
+ */
+template <typename T>
+class SharedMessage {
+ public:
+  /**
+   * Creates a valid `SharedMessage` that wraps a value-constructed ("zeroed
```
var-const, line 198: Hmm, this doesn't seem to be the case -- `message_` has no initializer, so it seems like a default-constructed `Message` would have a null pointer to a proto.

Firestore/core/src/nanopb/nanopb_util.h
```diff
@@ -104,6 +104,69 @@ T* _Nonnull MakeArray(pb_size_t count) {
   return static_cast<T*>(calloc(count, sizeof(T)));
 }
 
+template <typename T>
```
var-const, line 107: Consider adding tests for some of the newly-added functionality.

var-const (changes requested):

var-const:
I wonder if we need a helper class to wrap Nanopb arrays, similar to `Message`? (not for this PR)

Firestore/core/src/nanopb/nanopb_util.h
schmidt-sebastian: Tests added for SetRepeatedField

Firestore/Example/Firestore.xcodeproj/project.pbxproj
```diff
@@ -28,7 +28,6 @@
 		048A55EED3241ABC28752F86 /* memory_mutation_queue_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = 74FBEFA4FE4B12C435011763 /* memory_mutation_queue_test.cc */; };
 		04D7D9DB95E66FECF2C0A412 /* bundle_cache_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = F7FC06E0A47D393DE1759AE1 /* bundle_cache_test.cc */; };
 		0500A324CEC854C5B0CF364C /* FIRCollectionReferenceTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 5492E045202154AA00B64F25 /* FIRCollectionReferenceTests.mm */; };
-		051D3E20184AF195266EF678 /* no_document_test.cc in Sources */ = {isa = PBXBuildFile; fileRef = AB6B908720322E8800CC290A /* no_document_test.cc */; };
```
schmidt-sebastian, line 31: This are all the changes from the mrschmidt/migrate branch since this file was edited when I added the test.

schmidt-sebastian:
I am using ArrayValue instead of a vector of values in the implementation which allows me to re-use the existing Message class.

Firestore/core/src/nanopb/nanopb_util.h
var-const: I think that's quite a lot, actually. Additionally, it adds the possibility of passing null arguments, which imposes the extra burden of checking the arguments and finding a way to signal error. Also, see go/cstyle#Inputs_and_Outputs:
> Prefer using return values over output parameters: they improve readability, and often provide the same or better performance

Firestore/core/src/nanopb/message.h
```diff
@@ -188,23 +188,18 @@ class Message {
 };
 
 /**
- * A wrapper of Message objects that facilitates shared ownership of non-mutable
- * Protobuf data.
+ * An wrapper of const Message objects that facilitates shared ownership
```
var-const, line 191: Nit: `s/An/A/`.

Firestore/core/src/nanopb/nanopb_util.h
var-const: Ah, right, I misread that.

Firestore/core/src/nanopb/message.h
schmidt-sebastian: Done

Firestore/core/src/nanopb/nanopb_util.h
schmidt-sebastian: I assume you meant to say "I don't think that's quite a lot, actually."

Changing this would negatively impact all callsites. It would change code such as this https://github.com/firebase/firebase-ios-sdk/pull/8008/files#diff-4446e45495c7ddead4c8e19f3ad9233bd1ba6f0eac229187c20d66d4165f0850R433  from:


``` 
google_firestore_v1_ArrayValue array_value;
SetRepeatedField(&array_value.values, &array_value.values_count, operands);
```

to

``` 
google_firestore_v1_ArrayValue array_value;
auto span = SetRepeatedField(&array_value.values, &array_value.values_count, operands); 
array_value.values = span.begin();
array_value.values_count = static_cast<pb_size_t>(span.size());
```

It creates an extra object and introduces the possibility of not assigning all values, which is one of the main reasons why I created this helper.

var-const: Thanks for the example! I didn't realize the array was nested inside another message (though I guess it should have been obvious).

This could work using `std::tie` and `std::pair` (unfortunately, `span` wouldn't work) to emulate multiple assignment:
```cpp
google_firestore_v1_ArrayValue array_value;
std::tie(array_value.values, array_value.values_count) = SetRepeatedField(operands);
```

If you don't want to deal with `std::tie`, it's absolutely fine to leave as is (but please check the input pointer arguments for null).

var-const (approved):

Firestore/core/src/nanopb/nanopb_util.h
schmidt-sebastian: Null checks added.

PR #7989, opened 2021-04-27
DynamicLinks: Skip retrieving web locale when running targeting below iOS14 (only on simulator & Apple Silicon)

### LINKED ISSUE #7618: Crash on .configure() using M1 chip, iOS 13 simulator and Rosetta

  * Xcode version: 12.4
  * Firebase SDK version: 6.34
  * Firebase Component: Core 

The app crashes when configuring Firebase on simulated iOS 13 devices being run on an M1 chip with Rosetta active (can't build without it, other Pods not supporting arm64 architecture)

#### Steps to reproduce:
Select an iOS 13 device using the aforementioned device and options and the app will crash with a `Thread 1: EXC_BREAKPOINT (code=EXC_I386_BPT, subcode=0x0)` at `FIRDLJavaScriptExecutor.m`

#### Relevant Code:

```
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
    FirebaseApp.configure()
    
    return true
}
```

ncooke3:
Manually tested fix on M1 machine and it successfully ran without crashing. 

Fixes #7618

### DIFF
--- a/FirebaseDynamicLinks/Sources/FIRDLJavaScriptExecutor.m
+++ b/FirebaseDynamicLinks/Sources/FIRDLJavaScriptExecutor.m
@@ -75,18 +75,18 @@ NSString *GINFingerprintJSMethodString(void) {
 
 #pragma mark - Internal methods
 - (void)start {
-  // Initializing a `WKWebView` causes a memory allocation error when the process
-  // is running under Rosetta translation on Apple Silicon.
-  // The issue only occurs on the simulator in apps targeting below iOS 14. (Issue #7618)
-  #if TARGET_OS_SIMULATOR
-  BOOL systemVersionAtLeastiOS14 =
-    [NSProcessInfo.processInfo isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){14,0,0}];
+// Initializing a `WKWebView` causes a memory allocation error when the process
+// is running under Rosetta translation on Apple Silicon.
+// The issue only occurs on the simulator in apps targeting below iOS 14. (Issue #7618)
+#if TARGET_OS_SIMULATOR
+  BOOL systemVersionAtLeastiOS14 = [NSProcessInfo.processInfo
+      isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){14, 0, 0}];
   // Perform an early exit if the process is running under Rosetta translation and targeting
   // under iOS 14.
   if (processIsTranslated() && !systemVersionAtLeastiOS14) {
     return;
   }
-  #endif
+#endif
   NSString *htmlContent =
       [NSString stringWithFormat:@"<html><head><script>%@</script></head></html>", _script];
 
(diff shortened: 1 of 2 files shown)


FirebaseDynamicLinks/Sources/FIRDLJavaScriptExecutor.m
```diff
@@ -73,6 +75,19 @@ - (instancetype)initWithDelegate:(id<FIRDLJavaScriptExecutorDelegate>)delegate
...
+// Initializing a `WKWebView` causes a memory allocation error when the process
+// is running under Rosetta translation on Apple Silicon.
+// The issue only occurs on the simulator in apps targeting below iOS 14. (Issue #7618)
+#if TARGET_OS_SIMULATOR
+  BOOL systemVersionAtLeastiOS14 = [NSProcessInfo.processInfo
+      isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){14, 0, 0}];
+  // Perform an early exit if the process is running under Rosetta translation and targeting
+  // under iOS 14.
+  if (processIsTranslated() && !systemVersionAtLeastiOS14) {
+    [self handleExecutionError:nil];
```
ncooke3, line 87: Calling the delegate method in the early exit case avoids interrupting the dynamic links flow and the locale property in the delegate will be set to empty string. See [here](https://github.com/firebase/firebase-ios-sdk/blob/4662da75142fbc675e294b83415184137ec8bf79/FirebaseDynamicLinks/Sources/FIRDLDefaultRetrievalProcessV2.m#L94)

FirebaseDynamicLinks/Sources/FIRDLJavaScriptExecutor.m
```diff
@@ -135,6 +150,21 @@ - (void)webView:(WKWebView *)webView
   [self handleExecutionError:error];
 }
 
+// Determine whether a process is running under Rosetta translation.
+// Returns 0 for a native process, 1 for a translated process,
+// and -1 when an error occurs.
+// From:
+// https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment
+int processIsTranslated() {
```
paulb777, line 158: Make this C function static so that it does not pollute the global symbol namespace.

paulb777 (commented):
LGTM with static fix.

Deferring approval to @eldhosembabu

FirebaseDynamicLinks/Sources/FIRDLJavaScriptExecutor.m
ncooke3: TIL... thanks!

eldhosembabu (approved):
Thanks for addressing this.

paulb777 (approved):
Please update the changelog

PR #7991, opened 2021-04-28
ValueUtil changes for FieldValue migration

schmidt-sebastian:

This PR adds a couple of new utilities that are used in the FieldValue port (and changes some signatures). They are all based on Android:  https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java  

The PR also adds logic to treat zero-length strings/bytes the same as `null` bytes and strings, as nanopb uses both representations.

### DIFF
--- a/Firestore/core/src/model/value_util.cc
+++ b/Firestore/core/src/model/value_util.cc
@@ -490,7 +490,7 @@ google_firestore_v1_Value NaNValue() {
 
 bool IsNaNValue(const google_firestore_v1_Value& value) {
   return value.which_value_type == google_firestore_v1_Value_double_value_tag &&
-         isnan(value.double_value);
+         std::isnan(value.double_value);
 }
 
 google_firestore_v1_Value RefValue(const model::DatabaseId& database_id,
--- a/Firestore/core/src/model/value_util.cc
+++ b/Firestore/core/src/model/value_util.cc
@@ -17,9 +17,13 @@
 #include "Firestore/core/src/model/value_util.h"
 
 #include <algorithm>
+#include <cmath>
+#include <limits>
 #include <memory>
 #include <vector>
 
+#include "Firestore/core/src/model/database_id.h"
+#include "Firestore/core/src/model/document_key.h"
 #include "Firestore/core/src/model/server_timestamp_util.h"
 #include "Firestore/core/src/nanopb/nanopb_util.h"
 #include "Firestore/core/src/util/comparison.h"
@@ -123,9 +127,10 @@ ComparisonResult CompareBlobs(const google_firestore_v1_Value& left,
                ? util::ComparisonResultFromInt(cmp)
                : util::Compare(left.bytes_value->size, right.bytes_value->size);
   } else {
-    // An empty blob is represented by a nullptr
-    return util::Compare(left.bytes_value != nullptr,
-                         right.bytes_value != nullptr);
+    // An empty blob is represented by a nullptr (or an empty byte array)
+    return util::Compare(
+        !(left.bytes_value == nullptr || left.bytes_value->size == 0),
+        !(right.bytes_value == nullptr || right.bytes_value->size == 0));
   }
 }
 
@@ -247,8 +252,8 @@ ComparisonResult Compare(const google_firestore_v1_Value& left,
   }
 }
 
-bool NumberEquals(const firebase::firestore::google_firestore_v1_Value& left,
-                  const firebase::firestore::google_firestore_v1_Value& right) {
+bool NumberEquals(const google_firestore_v1_Value& left,
+                  const google_firestore_v1_Value& right) {
   if (left.which_value_type == google_firestore_v1_Value_integer_value_tag &&
       right.which_value_type == google_firestore_v1_Value_integer_value_tag) {
     return left.integer_value == right.integer_value;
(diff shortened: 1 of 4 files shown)


Firestore/core/src/model/value_util.cc
```diff
@@ -455,12 +457,61 @@ std::string CanonicalId(const google_firestore_v1_Value& value) {
...
+google_firestore_v1_Value NaNValue() {
+  google_firestore_v1_Value nan_value{};
+  nan_value.which_value_type = google_firestore_v1_Value_double_value_tag;
+  nan_value.double_value = std::numeric_limits<double>::quiet_NaN();
+  return nan_value;
+}
+
+bool IsNaNValue(const google_firestore_v1_Value& value) {
+  return value.which_value_type == google_firestore_v1_Value_double_value_tag &&
+         isnan(value.double_value);
```
wu-hui, line 493: Should be `std::isnan` from `cmath` instead.

`isnan` is also available because `cmath` also include `math.h`, but should not be directly used. Just learnt this today from Costa's code review :-)

Firestore/core/src/model/value_util.h
```diff
@@ -58,14 +63,69 @@ util::ComparisonResult Compare(const google_firestore_v1_Value& left,
...
+            const google_firestore_v1_ArrayValue& right);
+
 /**
  * Generate the canonical ID for the provided field value (as used in Target
  * serialization).
  */
 std::string CanonicalId(const google_firestore_v1_Value& value);
 
+/**
+ * Generate the canonical ID for the provided array value (as used in Target
```
wu-hui, line 76: "Generates"

Firestore/core/src/model/value_util.h
```diff
@@ -58,14 +63,69 @@ util::ComparisonResult Compare(const google_firestore_v1_Value& left,
...
+ * serialization).
+ */
+std::string CanonicalId(const google_firestore_v1_ArrayValue& value);
+
+/** Returns true if the array value contains the specified element. */
+bool Contains(google_firestore_v1_ArrayValue haystack,
+              google_firestore_v1_Value needle);
+
+/** Returns `nullptr` in its Protobuf representation. */
```
wu-hui, line 85: Maybe `Returns a null Protobuf value`? `nullptr` is pretty specifically a pointer to address 0.

Firestore/core/src/model/value_util.h
```diff
@@ -58,14 +63,69 @@ util::ComparisonResult Compare(const google_firestore_v1_Value& left,
...
+google_firestore_v1_Value NullValue();
+
+/** Returns `true` if `value` is `nullptr` in its Protobuf representation. */
```
wu-hui, line 88: `...if given value is a null Protobuf value`?

Firestore/core/src/model/value_util.h
```diff
@@ -58,14 +63,69 @@ util::ComparisonResult Compare(const google_firestore_v1_Value& left,
...
+bool IsNullValue(const google_firestore_v1_Value& value);
+
+/** Returns `NaN` in its Protobuf representation. */
+google_firestore_v1_Value NaNValue();
+
+/** Returns `true` if `value` is `NaN` in its Protobuf representation. */
+bool IsNaNValue(const google_firestore_v1_Value& value);
+
+google_firestore_v1_Value RefValue(const DatabaseId& database_id,
```
wu-hui, line 97: Please add comment.

wu-hui (changes requested):
Some minor changes requested, almost LGTM.

Firestore/core/src/model/value_util.cc
```diff
@@ -123,9 +127,10 @@ ComparisonResult CompareBlobs(const google_firestore_v1_Value& left,
                ? util::ComparisonResultFromInt(cmp)
                : util::Compare(left.bytes_value->size, right.bytes_value->size);
   } else {
-    // An empty blob is represented by a nullptr
-    return util::Compare(left.bytes_value != nullptr,
-                         right.bytes_value != nullptr);
+    // An empty blob is represented by a nullptr (or an empty byte array)
```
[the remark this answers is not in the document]
schmidt-sebastian: This is not meant to be a behavior change but rather a bug fix. When we create an empty bytes array directly with NanoPB, it uses `nullptr` and does not actually allocate memory. When we receive documents from the backend, these empty byte arrays are represented using a zero-length array, but one with an actual address in memory. We need these two representations to yield the same comparison behavior so our users don't see any behavior change.

Firestore/core/src/model/value_util.cc
schmidt-sebastian: Very nice. Hopefully I can use this information in my next code review.

Firestore/core/src/model/value_util.h
schmidt-sebastian: Fixed

Firestore/core/src/model/value_util.h
schmidt-sebastian: Fixed

Firestore/core/src/model/value_util.h
schmidt-sebastian: Done

schmidt-sebastian (commented):
Thanks for the review!

Firestore/core/src/model/value_util.cc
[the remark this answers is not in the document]
wu-hui: Sounds good, thanks.

wu-hui (approved):
There is one place you missed to update a comment. But LGTM still!

Firestore/core/src/model/value_util.h
schmidt-sebastian: Done

schmidt-sebastian: Done

PR #7994, opened 2021-04-28
Firestore cmake: update GoogleUtilities version to 7.4.0

maksymmalyhin:
Fixes [build failure](https://github.com/firebase/firebase-ios-sdk/runs/2456501007?check_suite_focus=true).


ryanwilson (approved):
Nice! Do you think it's worth adding a comment in FirebaseFirestore,.podspec above the GoogleUtilities dependency version to say something like "remember to also update `cmake/external/GoogleUtilities.cmake` or something?

ryanwilson:
cc @var-const

paulb777 (approved):

maksymmalyhin:
The SPM failure is not relevant for the current PR and is being investigated separately. I'll merge the PR on CI green.

Issue #23, opened 2017-05-21
Reorganize public headers


paulb777:
### [READ] Step 1: Are you in the right place?
yes

### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 8.3
  * Firebase SDK version: 4.0
  * Library version: 4.0
  * Firebase Product: all (auth, database, storage, core, messaging, etc)

### [REQUIRED] Step 3: Describe the problem
Consolidate public headers for each component into a single Public folder to make source management and podspec maintenance more robust.

PR #7995, opened 2021-04-28
FirebaseMessaging: Mock FIRHeartbeatInfo in test

ncooke3:
~~Add to test setup to use platform specific heartbeat storage APIs.~~
Revised approach: Mock `FIRHeartbeatInfo` to avoid side effects of using GULHeartbeatDateStorable objects. 

Breakage due to:
GoogleUtilites [#23](https://github.com/google/GoogleUtilities/pull/23) and firebase-ios-sdk #7893 for context.

Fixes FirebaseMessaging Failure in #7993.

cc @chliangGoogle 

#no-changelog

### DIFF
--- a/FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
+++ b/FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
@@ -108,8 +108,6 @@ static NSString *kRegistrationToken = @"token-12345";
   // `FIRMessagingTokenOperation` uses `FIRInstallations` under the hood to get FIS auth token.
   // Stub `FIRInstallations` to avoid using a real object.
   [self stubInstallations];
-
-  [self setupHeartbeatStorageForTesting];
 }
 
 - (void)tearDown {
@@ -119,6 +117,8 @@ static NSString *kRegistrationToken = @"token-12345";
   _checkinService = nil;
   _mockTokenStore = nil;
   [_mockInstallations stopMocking];
+
+  [self tearDownHeartbeatStorage];
 }
 
 - (void)testThatTokenOperationsAuthHeaderStringMatchesCheckin {
@@ -440,7 +440,7 @@ static NSString *kRegistrationToken = @"token-12345";
   OCMStub([_mockInstallations authTokenWithCompletion:authTokenWithCompletionArg]);
 }
 
-- (void)setupHeartbeatStorageForTesting {
+- (void)tearDownHeartbeatStorage {
   NSString *const kHeartbeatStorageName = @"HEARTBEAT_INFO_STORAGE";
   id<GULHeartbeatDateStorable> dataStorage;
 #if TARGET_OS_TV
--- a/FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
+++ b/FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
@@ -18,7 +18,9 @@
 
 #import "OCMock.h"
 
+#import <GoogleUtilities/GULHeartbeatDateStorable.h>
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
+#import <GoogleUtilities/GULHeartbeatDateStorageUserDefaults.h>
 #import "FirebaseCore/Sources/Private/FirebaseCoreInternal.h"
 #import "FirebaseInstallations/Source/Library/Private/FirebaseInstallationsInternal.h"
 #import "FirebaseMessaging/Sources/FIRMessagingConstants.h"
@@ -107,10 +109,7 @@ static NSString *kRegistrationToken = @"token-12345";
   // Stub `FIRInstallations` to avoid using a real object.
   [self stubInstallations];
 
-  NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
-  GULHeartbeatDateStorage *dataStorage =
-      [[GULHeartbeatDateStorage alloc] initWithFileName:kHeartbeatStorageFile];
-  [[NSFileManager defaultManager] removeItemAtURL:[dataStorage fileURL] error:nil];
+  [self setupHeartbeatStorageForTesting];
 }
 
 - (void)tearDown {
@@ -390,7 +389,10 @@ static NSString *kRegistrationToken = @"token-12345";
                     XCTAssertEqualObjects(userAgentValue, [FIRApp firebaseUserAgent]);
                     NSString *heartBeatCode =
                         sentRequest.allHTTPHeaderFields[kFIRMessagingFirebaseHeartbeatKey];
-                    XCTAssertEqualObjects(heartBeatCode, @"3");
+                    // See FirebaseCore/Sources/Private/FIRHeartbeatInfo.h for heartbeat codes.
+                    NSInteger sdkAndGlobalHeartbeatRequestedCode = 3;
+                    XCTAssertEqual(heartBeatCode.integerValue, sdkAndGlobalHeartbeatRequestedCode,
+                                   @"Heartbeat storage info needed to be updated but was not.");
                     [completionExpectation fulfill];
 
                     return YES;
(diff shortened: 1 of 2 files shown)


FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -438,4 +440,23 @@ - (void)stubInstallations {
   OCMStub([_mockInstallations authTokenWithCompletion:authTokenWithCompletionArg]);
 }
 
+- (void)setUpHeartbeatStorageForTesting {
+  NSString *const kHeartbeatStorageName = @"HEARTBEAT_INFO_STORAGE";
+  id<GULHeartbeatDateStorable> dataStorage;
```
maksymmalyhin, line 445: Why is it needed here?

maksymmalyhin (approved):
👍 LGTM, one question.

ryanwilson (approved):

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
ncooke3: Hmm I placed it there because a `dataStorage` is initialized [currently within the test suite](https://github.com/firebase/firebase-ios-sdk/blob/31307cdf7e2e7f0cf931ccf30f54fb409cad037d/FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m#L111) but your question makes me wonder why this is needed (because the tests fail when you don't call this  `setUpHeartbeatStorageForTesting`). Investigating...

maksymmalyhin: Ah, ok, so for file based storage we need an instance to get the file URL. So I guess it can be moved there. I cannot see why we need to instantiate `GULHeartbeatDateStorageUserDefaults` here (but I may miss something).

ncooke3: I understand now: the failing test is written to expect a heartbeat update so we need to clean up the any heartbeat info that may have been generated during a previous test run. Creating a `dataStorage` object is really just a convenient way to remove that data heartbeat data by allowing us to remove the heartbeat file (for non-tvOS) or remove the key in userdefaults (for tvOS).

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -106,11 +108,6 @@ - (void)setUp {
   // `FIRMessagingTokenOperation` uses `FIRInstallations` under the hood to get FIS auth token.
   // Stub `FIRInstallations` to avoid using a real object.
   [self stubInstallations];
-
-  NSString *const kHeartbeatStorageFile = @"HEARTBEAT_INFO_STORAGE";
-  GULHeartbeatDateStorage *dataStorage =
```
chliangGoogle, line 111: I wonder if we can remove this GULHeartbeatDateStorage completely. Is it ever used? I feel like token operation just use a  @"X-firebase-client-log-type" heart beat string that I'm not sure how it is related to GULHeartbeatDataStorage

maksymmalyhin: I think the test was ported from IID. IID accesses `GULHeartbeatDateStorage` via `FIRHeartbeatInfo` to get value for "X-firebase-client-log-type". It looks still relevant to me unless I'm missing something. Yeah, I agree, it's not the cleanest test and ideally we should mock `FIRHeartbeatInfo` here instead of using real objects and modifying underlying storage. I would suggest to make this cleanup as a follow up if the PR is blocking the release, otherwise we probably can do the cleanup here. WDYT?

ncooke3: To add: The `testThatTokenOperationsAuthHeaderStringMatchesCheckin` [expects `FIRHeartbeatInfo` to update the heartbeat](https://github.com/firebase/firebase-ios-sdk/pull/7995/files#diff-d970d32555eab994011dca7012f02f05445f82e8bb1eaf5f49359b06f43b20d4L393). So since we want to trigger a heartbeat update, we need to make sure `FIRHeartbeatInfo` decides to update the storage. If we don't manually remove the storage's artifacts after each test run, then `FIRHeartbeatInfo` [won't properly update the heartbeat because it will read the existing info](https://github.com/firebase/firebase-ios-sdk/blob/fe06273b691d4733be5a862f1f49d2725a8ce120/FirebaseCore/Sources/FIRHeartbeatInfo.m#L46) from a previous run and decide not to call `setHeartbeatForTag:`... which will [cause the assertion in the test to fail](https://github.com/firebase/firebase-ios-sdk/pull/7995/files#diff-d970d32555eab994011dca7012f02f05445f82e8bb1eaf5f49359b06f43b20d4L393) .

chliangGoogle: I believe @"X-firebase-client-log-type" is [a hardcode string](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/Token/FIRMessagingTokenFetchOperation.m#L36) and `GULHeartbeatDateStorage` is not used in the Messaging or instanceID library itself.

I'm guessing FIRURLSessionOCMockStub is the reason that is using GULHeartbeatDataStorage?

Anyway I don't want to block the release, i'm okay submitting it to get it green, but would like to understand more on this.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -438,4 +442,11 @@ - (void)stubInstallations {
   OCMStub([_mockInstallations authTokenWithCompletion:authTokenWithCompletionArg]);
 }
 
+- (void)stubHeartbeatInfo {
+  _mockHeartbeatInfo = OCMClassMock([FIRHeartbeatInfo class]);
+  NSInteger sdkAndGlobalHeartbeatRequestedCode = 3;
+  OCMStub([_mockHeartbeatInfo heartbeatCodeForTag:[OCMArg any]])
```
maksymmalyhin, line 448: We can validate the expected tag value passed to the method, e.g. 
```
OCMStub([_mockHeartbeatInfo heartbeatCodeForTag:@"expected_tag"])...
```

In this case the stub won't be applied to other argument values so the test will fail.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -438,4 +442,11 @@ - (void)stubInstallations {
   OCMStub([_mockInstallations authTokenWithCompletion:authTokenWithCompletionArg]);
 }
 
+- (void)stubHeartbeatInfo {
+  _mockHeartbeatInfo = OCMClassMock([FIRHeartbeatInfo class]);
+  NSInteger sdkAndGlobalHeartbeatRequestedCode = 3;
```
maksymmalyhin, line 447: We can define it as a test property or a const to share between the method and the tests using it.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -18,7 +18,9 @@
 
 #import "OCMock.h"
 
+#import <GoogleUtilities/GULHeartbeatDateStorable.h>
 #import <GoogleUtilities/GULHeartbeatDateStorage.h>
+#import <GoogleUtilities/GULHeartbeatDateStorageUserDefaults.h>
```
maksymmalyhin, line 23: I think, these imports are not needed any more.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -104,13 +107,10 @@ - (void)setUp {
   _instanceID = @"instanceID";
 
   // `FIRMessagingTokenOperation` uses `FIRInstallations` under the hood to get FIS auth token.
-  // Stub `FIRInstallations` to avoid using a real object.
+  // `FIRMessagingTokenFetchOperation` uses `FIRHeartbeatInfo` to retrieve a heartbeat code.
+  // Stub `FIRInstallations` and `FIRHeartbeatInfo` to avoid using real objects.
```
maksymmalyhin, line 111: optional: I would comment on each `[self stubInstallations]` and `[self stubHeartbeatInfo]` separately as it looks a bit more readable to me, but I'm not too strong about it.

maksymmalyhin (approved):
Thank you for the clean up! LGTM with a couple nits.

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
ncooke3: Good spot!

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
ncooke3: While applying this feedback, I realized I could just use `FIRHeartbeatInfoCodeCombined` because the `FIRHeartbeatInfoCode` enum namespace is available in this file. Is this preferred over a test property/const?

maksymmalyhin: Using existing constant SGTM

ncooke3: SG, thanks for all the feedback. I'll merge on green.

maksymmalyhin (approved):

PR #7996, opened 2021-04-28
Document/MutableDocument changes

schmidt-sebastian:

This PR changes MutableDocument to be copy-assignable (since ObjectValue is getting changed to be backed by a SharedMessage). It also introduces Document, as a wrapper for MutableDocuments that can no longer be mutated. These are used once documents become part of a view.


Firestore/core/src/model/document.h
```diff
@@ -17,125 +17,57 @@
 #ifndef FIRESTORE_CORE_SRC_MODEL_DOCUMENT_H_
 #define FIRESTORE_CORE_SRC_MODEL_DOCUMENT_H_
 
-#include <iosfwd>
-#include <memory>
+#include <iostream>
```
var-const, line 20: `<iostream>` is a relatively heavy-weight header, so if `<iosfwd>` suffices, it's better to use `<iosfwd>`. It looks like you can do that if you move the definition of `operator<<` into the source file.

Firestore/core/src/model/document.h
```diff
@@ -17,125 +17,57 @@
...
-};
-
-std::ostream& operator<<(std::ostream& os, DocumentState state);
-
-/**
- * Represents a document in Firestore with a key, version, data and whether the
- * data has local mutations applied to it.
- */
-class Document : public MaybeDocument {
+/** Represents a non-mutable document in Firestore. */
```
var-const, line 30: Ultranit: `s/non-mutable/immutable/`?

Firestore/core/src/model/mutable_document.h
```diff
@@ -211,27 +200,25 @@ class MutableDocument {
...
+        version_{version},
         value_{std::move(value)},
         document_state_{document_state} {
   }
 
   DocumentKey key_;
   DocumentType document_type_ = DocumentType::kInvalid;
   SnapshotVersion version_;
-  ObjectValue value_;
+  std::shared_ptr<const ObjectValue> value_ = std::make_shared<ObjectValue>();
```
var-const, line 215: Nit: `std::make_shared<const ObjectValue>()`?

Firestore/core/src/model/mutable_document.h
```diff
@@ -17,17 +17,19 @@
 #ifndef FIRESTORE_CORE_SRC_MODEL_MUTABLE_DOCUMENT_H_
 #define FIRESTORE_CORE_SRC_MODEL_MUTABLE_DOCUMENT_H_
 
+#include <memory>
 #include <ostream>
```
var-const, line 21: Note: this file could probably replace `<ostream>` with `<iosfwd>` as well.

Firestore/core/src/model/mutable_document.h
var-const, line 215: Can you add a comment to explain the rationale of making this a shared pointer?

Firestore/core/src/model/mutable_document.cc
```diff
@@ -87,18 +83,27 @@ MutableDocument& MutableDocument::SetHasLocalMutations() {
...
 
+size_t MutableDocument::Hash() const {
+  return key_.Hash();
+}
+
 std::string MutableDocument::ToString() const {
   std::stringstream stream;
   stream << "MutableDocument(key=" << key_ << ", type=" << document_type_
-         << ", version=" << version_ << ", value=" << value_
+         << ", version=" << version_ << ", value=" << CanonicalId(value_->Get())
```
var-const, line 93: Question: is it necessary to call `CanonicalId` and `Get` rather than just do `<< *value_`?

Firestore/core/src/model/mutable_document.cc
```diff
@@ -14,36 +14,32 @@
  * limitations under the License.
  */
 
-#include "Firestore/core/src/model/mutable_document.h"
-
 #include <sstream>
+#include "Firestore/core/src/model/mutable_document.h"
```
var-const, line 18: Nits:
* the `mutable_document.h` include should be first (it's the related header), i.e. undo the change;
* there has to be a blank line in-between the last standard header and the first Google header (also a blank line in-between the related header and the first standard header).

See go/cstyle#Names_and_Order_of_Includes.

Firestore/core/src/model/mutable_document.cc
```diff
@@ -63,7 +59,7 @@ MutableDocument& MutableDocument::ConvertToNoDocument(
     const SnapshotVersion& version) {
   version_ = version;
   document_type_ = DocumentType::kNoDocument;
-  value_ = {};
+  value_ = std::make_shared<ObjectValue>();
```
var-const, line 62: You want to make sure `value_` is never null to avoid checking it before each access, right?

Firestore/core/src/model/mutable_document.h
schmidt-sebastian: Done

schmidt-sebastian: Done

Firestore/core/src/model/document.h
schmidt-sebastian: I did not address this yet as this would require me to create `document.cc` and I am not sure if that tradeoff is worth making.

Firestore/core/src/model/mutable_document.h
```diff
@@ -212,7 +212,10 @@ class MutableDocument {
   DocumentKey key_;
   DocumentType document_type_ = DocumentType::kInvalid;
   SnapshotVersion version_;
-  std::shared_ptr<const ObjectValue> value_ = std::make_shared<ObjectValue>();
+  // Using a shared pointer to ObjectValue makes MutableDocument copy-assignable
+  // without having to manually create a deep clone of its Protobuf contents
```
var-const, line 216: Ultranit: please add a full stop at the end of the sentence.

var-const (approved):

Firestore/core/src/model/document.h
var-const: I don't think there's any harm in creating a `.cc` file, even if it's just for one function.

schmidt-sebastian: Done

Firestore/core/src/model/mutable_document.h
schmidt-sebastian: Done

PR #7998, opened 2021-04-28
Update firestore podspec presubmit to ios platform

granluo:
Firestore takes over an hour on presubmit.
https://github.com/firebase/firebase-ios-sdk/runs/2456478584?check_suite_focus=true

This is to limit the platform and intend to shorten the presubmit processing time.


paulb777 (approved):
LGTM on green

PR #8003, opened 2021-04-28
Test Refactor for Protobuf-backed FieldValue

schmidt-sebastian:

This PR contains the test files that only have "mechanical refactor" changes.

- No more DocumentState. Instead we use `SetHasLocalMutations`/`SetHasCommittedMutations`. `
- No more MaybeDocument. Instead we now have MutableDocument.
- No more MaybeDocumentMap. Instead just DocumentMap. 
- A Document is now a const wrapper on top of MutableDocument. Hence, some callsites change to `->`


thebrianchen (approved):
test changes lgtm

PR #8006, opened 2021-04-29
Add iPadOS as split out in Crashlytics Platform

samedson:
We need to split off iPad

#no-changelog

### DIFF
--- a/Crashlytics/Crashlytics/Components/FIRCLSApplication.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSApplication.h
@@ -48,7 +48,7 @@ NSString* FIRCLSApplicationGetSDKBundleID(void);
 NSString* FIRCLSApplicationGetPlatform(void);
 
 /**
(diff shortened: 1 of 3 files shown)


Crashlytics/Crashlytics/Components/FIRCLSApplication.h
```diff
@@ -47,6 +47,11 @@ NSString* FIRCLSApplicationGetSDKBundleID(void);
  */
 NSString* FIRCLSApplicationGetPlatform(void);
 
+/**
+ * Returns the Operating System for filtering. Should be kept consisten with Analytics.
```
paulb777, line 51: typo: consistent

paulb777 (approved):
We could also add the wrapper function to GoogleUtilities, but I'm fine with this for now.

samedson:
@paulb777 totally - I was thinking we go with this for now and then I will add a function for GoogleUtilities in a later release so that Perf / Analytics can possibly read from the same value

PR #8007, opened 2021-04-29
ObjectValue changes

schmidt-sebastian:
This PR is part of the larger series and renames MutableObjectValue to ObjectValue (the original ObjectValue is gone). It also adds support for unsorted MapValues, since the Watch backend does in fact send unsorted maps.

### DIFF
--- a/Firestore/core/src/model/object_value.cc
+++ b/Firestore/core/src/model/object_value.cc
@@ -24,6 +24,7 @@
 #include "Firestore/core/src/nanopb/fields_array.h"
 #include "Firestore/core/src/nanopb/message.h"
 #include "Firestore/core/src/nanopb/nanopb_util.h"
+#include "absl/types/span.h"
 
 namespace firebase {
 namespace firestore {
@@ -31,12 +32,16 @@ namespace model {
 
 namespace {
 
-using nanopb::FieldsArray;
 using nanopb::FreeFieldsArray;
+using nanopb::FreeNanopbMessage;
 using nanopb::MakeArray;
 using nanopb::MakeBytesArray;
 using nanopb::MakeString;
 using nanopb::MakeStringView;
+using nanopb::ReleaseFieldOwnership;
+using nanopb::SetRepeatedField;
+
+void SortFields(google_firestore_v1_Value& value);
 
 struct MapEntryKeyCompare {
   bool operator()(const google_firestore_v1_MapValue_FieldsEntry& entry,
@@ -135,6 +140,7 @@ void ApplyChanges(
 
         target_entry.key = source_entry.key;
         target_entry.value = DeepClone(upsert_it->second);
+        SortFields(target_entry.value);
 
         ++upsert_it;
         ++source_index;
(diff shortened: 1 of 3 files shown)


Firestore/core/test/unit/model/object_value_test.cc
```diff
@@ -85,60 +72,60 @@ TEST_F(ObjectValueTest, ExtractsFieldMask) {
 }
 
 TEST_F(ObjectValueTest, OverwritesExistingFields) {
-  MutableObjectValue object_value = WrapObject("a", "object_value");
+  ObjectValue object_value = WrapObject("a", "object_value");
   EXPECT_EQ(WrapObject("a", "object_value"), object_value);
-  object_value.Set(Field("a"), Wrap("object_value"));
+  object_value.Set(Field("a"), Value("object_value"));
```
schmidt-sebastian, line 77: Note: Value is the function name in `master`, so this actually reduces the diff.

Firestore/core/src/model/object_value.h
```diff
@@ -37,26 +37,22 @@ namespace firestore {
...
-    HARD_ASSERT(
-        value.which_value_type == google_firestore_v1_Value_map_value_tag,
-        "ObjectValues should be backed by a MapValue");
-  }
+  explicit ObjectValue(const google_firestore_v1_Value& value);
 
-  MutableObjectValue(MutableObjectValue&& other) noexcept = default;
-  MutableObjectValue& operator=(MutableObjectValue&& other) = default;
+  ObjectValue(ObjectValue&& other) noexcept = default;
+  ObjectValue& operator=(ObjectValue&& other) = default;
```
var-const, line 48: Nit: `operator=` can be `noexcept` as well.

Firestore/core/src/model/object_value.cc
```diff
@@ -286,12 +350,19 @@ void MutableObjectValue::Delete(const FieldPath& path) {
   }
 }
 
+std::string ObjectValue::ToString() const {
+  return CanonicalId(*value_);
+}
+
+size_t ObjectValue::Hash() const {
+  return std::hash<std::string>()(CanonicalId(*value_));
```
var-const, line 358: Nit: I'd suggest using `util::Hash` (from `util/hasing.h`) instead.

Firestore/core/src/model/object_value.cc
```diff
@@ -49,6 +53,27 @@ struct MapEntryKeyCompare {
   }
 };
 
+void SortFields(google_firestore_v1_Value& value) {
```
var-const, line 56: Optional: perhaps add a comment to note that only map fields are sorted?

Firestore/core/src/model/object_value.h
```diff
@@ -37,26 +37,22 @@ namespace firestore {
...
+  ObjectValue(const ObjectValue& other);
 
-  /** `MutableObjectValue` models unique ownership. */
-  MutableObjectValue(const MutableObjectValue&) = delete;
-  MutableObjectValue& operator=(const MutableObjectValue&) = delete;
+  ObjectValue& operator=(const ObjectValue&) = delete;
+
+  static ObjectValue FromMapValue(google_firestore_v1_MapValue map_value);
+  static ObjectValue FromFieldsEntry(
```
var-const, line 54: Please add a comment to indicate that this function modifies the given `fields_entry`.

Firestore/core/src/model/object_value.h
```diff
@@ -37,26 +37,22 @@ namespace firestore {
...
+  ObjectValue(ObjectValue&& other) noexcept = default;
+  ObjectValue& operator=(ObjectValue&& other) = default;
+  ObjectValue(const ObjectValue& other);
 
-  /** `MutableObjectValue` models unique ownership. */
-  MutableObjectValue(const MutableObjectValue&) = delete;
-  MutableObjectValue& operator=(const MutableObjectValue&) = delete;
+  ObjectValue& operator=(const ObjectValue&) = delete;
+
+  static ObjectValue FromMapValue(google_firestore_v1_MapValue map_value);
```
var-const, line 53: This function effectively transfers ownership of the given `map_value` to the created `ObjectValue`, right? (because it doesn't deep-copy pointers) If so, please add a comment indicating that.

Firestore/core/test/unit/model/object_value_test.cc
```diff
@@ -35,43 +35,30 @@ using testutil::DbId;
...
-      WrapObject("foo", Map("a", 1, "b", true, "c", "string"));
+  ObjectValue value = WrapObject("foo", Map("a", 1, "b", true, "c", "string"));
 
   ASSERT_EQ(google_firestore_v1_Value_map_value_tag,
             value.Get(Field("foo"))->which_value_type);
 
-  EXPECT_TRUE(Wrap(1) == *value.Get(Field("foo.a")));
-  EXPECT_TRUE(Wrap(true) == *value.Get(Field("foo.b")));
-  EXPECT_TRUE(Wrap("string") == *value.Get(Field("foo.c")));
+  EXPECT_TRUE(Value(1) == *value.Get(Field("foo.a")));
```
var-const, line 51: Nit: looks like these checks could use `EXPECT_EQ` instead. See go/c-readability-advice#expect-that-vs-other-expect- (I know it's a preexisting issue)

Firestore/core/src/model/object_value.cc
schmidt-sebastian: Done

Firestore/core/src/model/object_value.cc
schmidt-sebastian: Done

Firestore/core/src/model/object_value.h
schmidt-sebastian: Done

Firestore/core/src/model/object_value.h
schmidt-sebastian: Yes, comment added.

Firestore/core/src/model/object_value.h
schmidt-sebastian: Done

Firestore/core/test/unit/model/object_value_test.cc
schmidt-sebastian: Fixed. I wasn't able to EXPECT_EQ at first because `operator==` was declared in the wrong scope.

Firestore/core/src/model/object_value.h
```diff
@@ -37,26 +37,33 @@ namespace firestore {
...
-  MutableObjectValue(MutableObjectValue&& other) noexcept = default;
-  MutableObjectValue& operator=(MutableObjectValue&& other) = default;
+  ObjectValue(ObjectValue&& other) noexcept = default;
+  ObjectValue& operator=(ObjectValue&& other) noexcept = default;
+  ObjectValue(const ObjectValue& other);
 
-  /** `MutableObjectValue` models unique ownership. */
-  MutableObjectValue(const MutableObjectValue&) = delete;
-  MutableObjectValue& operator=(const MutableObjectValue&) = delete;
+  ObjectValue& operator=(const ObjectValue&) = delete;
```
var-const, line 51: Why is this class copy constructible but not copy assignable?

var-const (commented):
Almost LGTM with one question.

Firestore/core/src/model/object_value.h
schmidt-sebastian: I think I would need this operator: https://github.com/firebase/firebase-ios-sdk/blob/a12b78e395ff6faf08798347ad500d44f856493f/Firestore/core/src/nanopb/message.h#L87

schmidt-sebastian:
@var-const Can you take another look?

Firestore/core/src/model/object_value.h
var-const: Right.

var-const (approved):
Sorry about the delay! LGTM.

PR #8008, opened 2021-04-29
TestUtil methods that create Protobuf

schmidt-sebastian:
These are the test util changes of the larger migration PR (https://github.com/firebase/firebase-ios-sdk/pull/7904)


schmidt-sebastian:
@wu-hui Do you have time to take a look at this?

wu-hui (approved):
LGTM

PR #8014, opened 2021-04-30
Implement app attest key ID storage

ncooke3:
• Implements `FIRAppAttestKeyIDStorage`
• Adds tests

[b/186653653](b/186653653)

### DIFF
--- a/FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
@@ -69,20 +69,16 @@ static NSString *const kKeyIDStorageDefaultsSuiteName = @"com.firebase.FIRAppAtt
 #pragma mark - Helpers
 
 - (void)storeAppAttestKeyID:(nullable NSString *)keyID {
-  @synchronized(self.userDefaults) {
-    if (keyID) {
-      [self.userDefaults setObject:keyID forKey:[self keyIDStorageKey]];
-    } else {
-      [self.userDefaults removeObjectForKey:[self keyIDStorageKey]];
-    }
+  if (keyID) {
+    [self.userDefaults setObject:keyID forKey:[self keyIDStorageKey]];
+  } else {
+    [self.userDefaults removeObjectForKey:[self keyIDStorageKey]];
   }
 }
 
 - (nullable NSString *)appAttestKeyIDFromStorage {
   NSString *appAttestKeyID = nil;
-  @synchronized(self.userDefaults) {
-    appAttestKeyID = [self.userDefaults objectForKey:[self keyIDStorageKey]];
-  }
+  appAttestKeyID = [self.userDefaults objectForKey:[self keyIDStorageKey]];
   return appAttestKeyID;
 }
 
--- a/FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
@@ -25,10 +25,7 @@
 #import "FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckErrorUtil.h"
 
 /// The `NSUserDefaults` suite name for the storage location of the app attest key ID.
-static NSString *const kFIRAppAttestKeyIDStorageDefaultsSuiteName = @"com.firebase.FIRAppAttestKeyIDStorage";
-
-/// The key used to retrieve the app attest key ID from its storage suite.
-static NSString *const kFIRAppAttestKeyIDKey = @"FIRAppAttestKeyID";
+static NSString *const kKeyIDStorageDefaultsSuiteName = @"com.firebase.FIRAppAttestKeyIDStorage";
 
 @interface FIRAppAttestKeyIDStorage ()
 
@@ -47,7 +44,7 @@ static NSString *const kFIRAppAttestKeyIDKey = @"FIRAppAttestKeyID";
   if (self) {
     _appName = [appName copy];
     _appID = [appID copy];
-    _userDefaults = [[NSUserDefaults alloc] initWithSuiteName:kFIRAppAttestKeyIDStorageDefaultsSuiteName];
+    _userDefaults = [[NSUserDefaults alloc] initWithSuiteName:kKeyIDStorageDefaultsSuiteName];
   }
   return self;
 }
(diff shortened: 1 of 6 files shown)


FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
```diff
@@ -36,18 +47,46 @@ - (instancetype)initWithAppName:(NSString *)appName appID:(NSString *)appID {
...
+  }
 }
 
-- (nonnull FBLPromise<NSString *> *)setAppAttestKeyID:(nullable NSString *)keyID {
-  // TODO: Implement.
-  return [FBLPromise resolvedWith:nil];
+#pragma mark - Helpers
+
+- (void)storeAppAttestKeyID:(nullable NSString *)keyID {
+  @synchronized (self.userDefaults) {
```
maksymmalyhin, line 75: optional: Alternatively we may use a private dispatch queue for synchronization  which is a preferable way when we have async API. You may leverage API like `+[FBLPromise onQueue:do:]` to organize to code.

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
```diff
@@ -36,18 +47,46 @@ - (instancetype)initWithAppName:(NSString *)appName appID:(NSString *)appID {
...
+    } else {
+      [self.userDefaults removeObjectForKey:kFIRAppAttestKeyIDKey];
+    }
+  }
+}
+
+- (nullable NSString *)appAttestKeyIDFromStorage {
+  NSString *appAttestKeyID = nil;
+  @synchronized (self.userDefaults) {
+    appAttestKeyID = [self.userDefaults objectForKey:kFIRAppAttestKeyIDKey];
```
maksymmalyhin, line 87: We need to store an App Attest key ID per Firebase app instance to make sure they can operate independently. The app instance is identified by the app name. In addition we need to make sure that if Firebase config changes, the app can still work correctly on user devices after update. For this we need to keep track of the app ID. The easiest way to achieve it in the code I can see is to make the app name and the app ID a part of the key. We use this approach already in the [core part of the SDK](https://github.com/firebase/firebase-ios-sdk/blob/638f5c7cf0628216af28cca96c7c57902217149b/FirebaseAppCheck/Sources/Core/Storage/FIRAppCheckStorage.m#L106) .

maksymmalyhin (changes requested):
Thank you for the implementation! Mostly LGTM, a couple adjustments needed.

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
ncooke3: Done and added a test like `testSetTokenPerApp ` seen [here](https://github.com/firebase/firebase-ios-sdk/blob/638f5c7cf0628216af28cca96c7c57902217149b/FirebaseAppCheck/Tests/Unit/Core/FIRAppCheckStorageTests.m#L76).

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
```diff
@@ -22,11 +22,22 @@
 #import "FBLPromises.h"
 #endif
 
+#import "FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckErrorUtil.h"
+
+/// The `NSUserDefaults` suite name for the storage location of the app attest key ID.
+static NSString *const kFIRAppAttestKeyIDStorageDefaultsSuiteName = @"com.firebase.FIRAppAttestKeyIDStorage";
```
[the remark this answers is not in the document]
ncooke3: Follow-up question: Should "com.firebase.appcheck" be used instead of "com.firebase.FIRAppAttestKeyIDStorage"? I referenced [the conversation in this pull request](https://github.com/firebase/firebase-ios-sdk/pull/7893#discussion_r612891887) and was wondering if "com.firebase.appcheck" is a better name for a suite of appcheck related defaults (and, if so, then  should [`FIRAppCheckSettings` use the "com.firebase.appcheck" suite as well– instead of using the shared user defaults instance](https://github.com/firebase/firebase-ios-sdk/blob/8877fb3554b0e40c7e57e6c078f91067c4343a9f/FirebaseAppCheck/Sources/Core/FIRAppCheck.m#L123)).

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
ncooke3: Discussed offline. Decision is to allow `FIRAppAttestKeyIDStorage`'s sole intended consumer, `FIRAppAttestProvider`, to manage synchronization of `FIRAppAttestKeyIDStorage` on its (`FIRAppAttestProvider`'s) internal serial queue.

A comment of the intended use has been added at the public interface definition of `FIRAppAttestKeyIDStorage`.

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestKeyIDStorage.m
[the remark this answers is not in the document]
maksymmalyhin: I think using a class specific suite name for this case is a good idea, so  "com.firebase.FIRAppAttestKeyIDStorage" LGTM.

maksymmalyhin (approved):
LGTM, one recommendation to add additional test cases.

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestKeyIDStorageTests.m
```diff
@@ -54,38 +55,88 @@ - (void)testSetAndGetAppAttestKeyID {
...
+  XCTAssert(FBLWaitForPromisesWithTimeout(0.5));
+  XCTAssertNil(getPromiseForAppAttestKeyID_3.error);
+
+  XCTAssertNotEqualObjects(getPromiseForAppAttestKeyID.value, getPromiseForAppAttestKeyID_2.value);
+  XCTAssertNotEqualObjects(getPromiseForAppAttestKeyID_2.value,
+                           getPromiseForAppAttestKeyID_3.value);
+
+  // 5. Cleanup other storages.
+  [storage2 setAppAttestKeyID:nil];
+  [storage3 setAppAttestKeyID:nil];
```
maksymmalyhin, line 134: optional: now it looks a bit harder to follow the test. We probably may slightly improve readability here, e.g.:
- create a helper method like `- assertIndependentSetGetForStoragesWithAppName1:appID1:appName2:appID2:` where we assert independent reads and writes
- call this method with all combinations we are interested in

WDYT?

maksymmalyhin (approved):
Great, thank you for the updates. Just one more optional recommendation on improving readability.

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestKeyIDStorageTests.m
ncooke3: Thanks for taking a look. Yeah, I agree. I was thinking about how to make it easier to understand. I like this approach ^ though. 👍

maksymmalyhin (approved):
Looks more readable now, thanks! LGTM.

PR #8015, opened 2021-04-30
Adjust the Messaging changelog after we remove iid from github

chliangGoogle:
Since the Firebase 8 will not have IID as part of its repo, developers will not be able to include InstanceID in the podfile to keep using InstanceID. They will have to migrate to new token APIs. Adjust the changelog to reflect this change.

### DIFF
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2021-04 -- v8.0.0
-- [changed] Remove Instance ID dependency from Messaging. This is a breaking change for Firebase Messaging users who use Instance ID to manage registration tokens without explicitly specifying Instance ID in their podfile or Swift Package Manager. For a short-term resolution, those Messaging users can update their dependencies to explicitly specify Instance ID. Longer term, all users should migrate to Messaging's token APIs following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
+- [changed] Remove Instance ID dependency from Messaging. This is a breaking change for Firebase Messaging users who use Instance ID to manage registration tokens. Please migrate to Messaging's token APIs following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
 
 # 2021-04 -- v7.11.0
 - [changed] Refactor Messaging to internally not depending on InstanceID, but can co-exist. Will remove InstanceID dependency in the next Firebase breaking change. (#7814)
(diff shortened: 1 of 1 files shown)


paulb777 (approved):

FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # 2021-04 -- v8.0.0
-- [changed] Remove Instance ID dependency from Messaging. This is a breaking change for Firebase Messaging users who use Instance ID to manage registration tokens without explicitly specifying Instance ID in their podfile or Swift Package Manager. For a short-term resolution, those Messaging users can update their dependencies to explicitly specify Instance ID. Longer term, all users should migrate to Messaging's token APIs following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
+- [changed] Remove Instance ID dependency from Messaging. This is a breaking change for Firebase Messaging users who use Instance ID to manage registration tokens. Please migrate to Messaging's token APIs following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
```
egilmorez, line 2: The style guide tells us to avoid "please," no matter how polite we are. :)

egilmorez, line 2: As I mentioned in the internal release notes CL, we should use one of the actual standard product names.  Here's we could say "Firebase Cloud Messaging" for the first instance, and then just "FCM" for the next two.

egilmorez (commented):
Thanks for sending this!  Commented on a couple small things.

FirebaseMessaging/CHANGELOG.md
chliangGoogle: I think externally we call it Firebase Messaging, so FCM is no longer making sense here.

egilmorez: Can you point me to the places where we call it "Firebase Messaging" externally?

There are indeed incorrect instances of "Firebase Messaging" in reference documentation.  I think those are mostly based on a conflation of literal type names like "FirebaseMessaging" and actual product names.  

When we refer correctly to Firebase Cloud Messaging as a product, it is definitely one of the three names I mentioned.

chliangGoogle: I mean if you look at this github repo, the SDK is called "Firebase Messaging"

chliangGoogle: I do see our public documentations are using "FCM", I can adjust here, but I feel like we are splitting in places half and half.

chliangGoogle:
@morganchen12 @egilmorez I've updated the changelog here to match internally.

FirebaseMessaging/CHANGELOG.md
egilmorez: The SDK name, as far as I can tell, is a literal, with camel case and no space: `FirebaseMessaging.`

To put a space in there and call it a product is the error that we have in various places.  That an error exists is of course no reason to continue reproducing it :). Thanks for fixing it here!

egilmorez (approved):

PR #8019, opened 2021-04-30
Add a validation check to `Settings`

var-const:
This is for consistency with other platforms (e.g. [Android](https://github.com/firebase/firebase-android-sdk/blob/3f0779ede98e4aa6986844a767c7f0eebc9d19b5/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java#L141-L144)).


var-const:
@wu-hui Friendly ping.

wu-hui (approved):

PR #8020, opened 2021-04-30
Run CI with default GHA Xcode version

paulb777:
To catch issues from new Xcode versions sooner. 

Most of these were added when the default was Xcode 11 and 12 was needed for SPM


ryanwilson (approved):
IIRC we wanted to have at least one build using the minimum supported Xcode version to catch issues - is our zip build sufficient?

paulb777:
I think the zip build should be sufficient.

Issue #7785, opened 2021-03-25
FirebaseAnalyticsSwift-Beta: Failure to Build/Archive with Xcode


cpdunphy:
### Describe your environment

  * Xcode version: 12.4
  * Firebase SDK version: 7.9.1
  * Installation method: `Swift Package Manager`
  * Firebase Component: `FirebaseAnalyticsSwift-Beta`

### Describe the problem

Xcode Archive fails when using the `FirebaseAnalyticsSwift-Beta` component. Building and installing the app with the dependency works so far as expected, the trouble is with archiving.

Note: the archive will succeed if you copy/add the source file (`Analytics+SwiftUI.swift`) to your local project, remove the troubled component (`FirebaseAnalyticsSwift-Beta`), and then attempt an archive. Due to this, I am led to believe this is a configuration issue, not (another) issue with SwiftUI. 

Additionally, as noted by the Firebase SPM documentation, there is a preexisting issue (#6472) that also deals with analytics archive errors. Since the issue I'm detailing seems to be contained to the `FirebaseAnalyticsSwift-Beta` component, I do not currently have reason to believe the other is relevant. 

#### Steps to reproduce:

1. Create a new project
2. Add "https://github.com/firebase/firebase-ios-sdk.git" as a Package Dependency (V 7.9.1)
3. Select the desired components: `FirebaseAnalyticsSwift-Beta` and `FirebaseAnalytics`
4. Archive

#### Relevant Error Messages:
```
FirebaseAnalyticsSwift Group
Swift Compiler Error Group
- Analytics+SwiftUI.swift:35:41: Cannot find type 'View' in scope
- Analytics+SwiftUI.swift:35:24: Cannot find type 'Content' in scope
- Analytics+SwiftUI.swift:25:44: Cannot find type 'ViewModifier' in scope
- Analytics+SwiftUI.swift:53:20: Cannot find type 'View' in scope
Uncategorized Group
- Command CompileSwift failed with a nonzero exit code
```

I'd like to note this additional warning. It does not (always) seem to impede the completion of a build, but does also popup along side the error above and seems contained to the `FirebaseAnalyticsSwift-Beta` component.
```
Skipping duplicate build file in Copy Files build phase: /Users/{user}/Library/Developer/Xcode/DerivedData/{project}/SourcePackages/artifacts/Firebase/FirebaseAnalytics.xcframework/ios-arm64_armv7/FirebaseAnalytics.framework
[...]

PR #8030, opened 2021-05-04
Bump the minimum SwiftPM iOS version to 11.

ryanwilson:
This is to avoid an issue with newer APIs being unable to compile when
building for device while this manifest targets iOS 10.

See https://github.com/firebase/firebase-ios-sdk/issues/7785#issuecomment-831959399
for more details on the issue and a reproduction. This should be
re-evaluated if Xcode ever fixes the issue.

This will need to be cherry-picked onto the `release-8.0` branch.


paulb777 (approved):
LGTM on green

PR #8033, opened 2021-05-04
Implement App Attest `getRandomChallenge`

ncooke3:
[b/187047998](b/187047998)

TODO's
- [x] Implement `getRandomChallenge`
- [x] Add tests

Questions to address during review:
- [x] Difficulties with encoding of challenge.
- [x] Using a dedicated, labeled queue rather than dispatching to `default`.

#no-changelog


FirebaseAppCheck/Sources/Core/APIService/FIRAppCheckAPIService.h
```diff
@@ -29,7 +29,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (FBLPromise<GULURLSessionDataResponse *> *)
     sendRequestWithURL:(NSURL *)requestURL
             HTTPMethod:(NSString *)HTTPMethod
-                  body:(NSData *)body
+                  body:(nullable NSData *)body
```
ncooke3, line 32: Added `nullable` attribute because no body is sent when requesting random challenge.

FirebaseAppCheck/Sources/Core/APIService/FIRAppCheckAPIService.m
```diff
@@ -167,6 +167,10 @@ - (instancetype)initWithURLSession:(NSURLSession *)session
                           }];
 }
 
+- (dispatch_queue_t)defaultQueue {
+  return dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0);
```
ncooke3, line 171: Discussed offline to standardize the QoS if the queues this classes dispatches methods on.

I read that GCD queues that created without specifying a QoS default to the `default` priority. Considering this, would creating internal default QoS queues for `FIRAppCheckAPIService` and `FIRAppAttestAPIService` be better since the queues will have more defined labels (for debugging).

FirebaseAppCheck/Sources/DebugProvider/API/FIRAppCheckDebugProviderAPIService.m
```diff
@@ -82,7 +82,7 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
       });
 }
 
-#pragma mark -
+#pragma mark - Helpers
```
ncooke3, line 85: Not sure if this was intended to be labeled or not.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
```diff
@@ -56,8 +57,30 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
...
+                          }]
+      .then(^id _Nullable(GULURLSessionDataResponse *_Nullable response) {
+        return [self randomChallengeWithAPIResponse:response];
+      });
+}
+
+- (FBLPromise<NSData *> *)randomChallengeWithAPIResponse:(GULURLSessionDataResponse *)response {
+  NSData *randomChallenge = response.HTTPBody;
+  return [FBLPromise resolvedWith:randomChallenge];
+}
```
ncooke3, line 80: Also: I wasn't sure if this helper method should belong in this class or `FIRAppCheckAPIService` since that class does have a similar `tokenWithAPIResponse` helper method. On one hand, it involves manipulating a response from `self.APIService` (a `FIRAppCheckAPIService` instance) but on the other it seems more specific to App Attest than App Check.

Thoughts?

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
```diff
@@ -56,8 +59,83 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
...
+    NSError *error;
+
+    NSData *randomChallenge = [self randomChallengeFromResponseBody:response.HTTPBody error:&error];
+
+    return randomChallenge ?: error;
+  }];
+}
+
+- (NSData *)randomChallengeFromResponseBody:(NSData *)response
+                                      error:(NSError **)outError {
```
ncooke3, line 90: I was trying to follow what was done for App Check token (see [here](https://github.com/firebase/firebase-ios-sdk/blob/f2b77f0dce2def4037806b705c7608cfb4e24298/FirebaseAppCheck/Sources/Core/APIService/FIRAppCheckAPIService.m#L156)). Im unsure if it is a better practice to wrap the random challenge in a custom type (e.g. AppAttestRandomChallenge) so that the parsing done in `randomChallengeFromResponseBody:error:` can be done in the initializer like it is done in [AppCheckToken+Response category](https://github.com/firebase/firebase-ios-sdk/blob/appcheck-appattest-main/FirebaseAppCheck/Sources/Core/APIService/FIRAppCheckToken%2BAPIResponse.m)

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
maksymmalyhin: `tokenWithAPIResponse` was moved up from specific API services to the `FIRAppCheckAPIService` because all FAC token exchange endpoints share the same response format, so it's better to share the implementation and having it in `FIRAppCheckAPIService` was the simplest approach.

As for parsing get challenge response - it should be in `FIRAppAttestAPIService.m` as it is unique for App Attest provider.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
```diff
@@ -56,8 +59,83 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
...
+                                                                                withString:@""];
+  NSTimeInterval secondsToLive = timeToLiveValueString.doubleValue;
+
+  if (secondsToLive == 0) {
+    FIRAppCheckSetErrorToPointer(
+        [FIRAppCheckErrorUtil appCheckTokenResponseErrorWithMissingField:@"ttl"], outError);
+    return nil;
+  }
+
+  NSData *randomChallenge = [challenge dataUsingEncoding:NSUTF8StringEncoding];
```
maksymmalyhin, line 133: I think it's base64URL (don't be confused with just base64). Could you please check and modify if needed.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
```diff
@@ -56,8 +59,83 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
...
+
+  NSString *timeToLiveString = responseDict[@"ttl"];
+  if (![challenge isKindOfClass:[NSString class]] || challenge.length <= 0) {
+    FIRAppCheckSetErrorToPointer(
+        [FIRAppCheckErrorUtil appCheckTokenResponseErrorWithMissingField:@"ttl"], outError);
+    return nil;
+  }
+
+  // Expect a string like "3600s" representing a time interval in seconds.
+  NSString *timeToLiveValueString = [timeToLiveString stringByReplacingOccurrencesOfString:@"s"
```
maksymmalyhin, line 123: I don't think we should parse parameter that we don't actually use. Parsing only `challenge` field should be enough and simplify the code.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
ncooke3: Noted, thanks!

FirebaseAppCheck/Sources/DebugProvider/API/FIRAppCheckDebugProviderAPIService.m
maksymmalyhin: LGTM

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
maksymmalyhin: As we are interested in the challenge data only, not having a specific type is fine. Placing parsing of the response here looks also fine especially since we will simplify it a bit.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
ncooke3: I checked and my understanding is that `getRandomChallenge` returns `NSUTF8StringEncoding` encoded data. See it mocked as such in tests [here](https://github.com/firebase/firebase-ios-sdk/blob/f2b77f0dce2def4037806b705c7608cfb4e24298/FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m#L140)

paulb777 (commented):
LGTM - deferring to @maksymmalyhin for detailed review and approval

FirebaseAppCheck/Tests/Fixture/AppAttestResponseSuccess.json
```diff
@@ -0,0 +1,4 @@
+{
+    "challenge": "random_challenge",
```
maksymmalyhin, line 2: We should use a valid base64 encoded string here, e.g. we can generate something the following Swift code:
```
let str = "random_challenge"
let utf8 = str.data(using: .utf8)
let base64String = utf8?.base64EncodedString() // "cmFuZG9tX2NoYWxsZW5nZQ=="
```

Then after updating the response decoding logic as mentioned in the comment above the tests in FIRAppAttestAPIServiceTests can stay unchanged.

FirebaseAppCheck/Sources/AppAttestProvider/API/FIRAppAttestAPIService.m
```diff
@@ -56,8 +59,68 @@ - (instancetype)initWithAPIService:(id<FIRAppCheckAPIServiceProtocol>)APIService
...
+                                                            HTTPMethod:@"POST"
+                                                                  body:nil
+                                                     additionalHeaders:nil];
+                          }]
+      .then(^id _Nullable(GULURLSessionDataResponse *_Nullable response) {
+        return [self randomChallengeWithAPIResponse:response];
+      });
+}
+
+#pragma mark - Helpers
```
maksymmalyhin, line 79: optional: Seems we can be more specific here, e.g. `#pragma mark - Challenge response parsing`

maksymmalyhin (approved):
LGMT after updating challenge parsing from base64 encoded string from the JSON response.

PR #8036, opened 2021-05-04
FieldFilter changes for Protobuf migration

schmidt-sebastian:
Please compare to https://github.com/firebase/firebase-android-sdk/tree/master/firebase-firestore/src/main/java/com/google/firebase/firestore/core

### DIFF
--- a/Firestore/core/src/core/field_filter.cc
+++ b/Firestore/core/src/core/field_filter.cc
@@ -28,6 +28,7 @@
 #include "Firestore/core/src/core/not_in_filter.h"
 #include "Firestore/core/src/core/operator.h"
 #include "Firestore/core/src/model/document.h"
+#include "Firestore/core/src/model/value_util.h"
 #include "Firestore/core/src/util/exception.h"
 #include "Firestore/core/src/util/hashing.h"
 #include "absl/algorithm/container.h"
@@ -38,8 +39,11 @@ namespace firebase {
 namespace firestore {
 namespace core {
 
+using model::Compare;
 using model::FieldPath;
-using model::FieldValue;
+using model::GetTypeOrder;
+using model::IsArray;
+using model::TypeOrder;
 using util::ComparisonResult;
 using util::ThrowInvalidArgument;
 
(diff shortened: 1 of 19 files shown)


Firestore/core/src/core/field_filter.cc
```diff
@@ -78,40 +82,31 @@ const char* CanonicalName(Filter::Operator op) {
...
-    return ArrayContainsAnyFilter(std::move(path), std::move(value_rhs));
+    return ArrayContainsAnyFilter(path, value_rhs);
   } else if (op == Operator::NotIn) {
-    HARD_ASSERT(value_rhs.type() == FieldValue::Type::Array,
-                "notIn filter has invalid value: %s", value_rhs.type());
-    return NotInFilter(std::move(path), std::move(value_rhs));
+    return NotInFilter(path, value_rhs);
   } else {
-    Rep filter(std::move(path), op, std::move(value_rhs));
+    Rep filter(std::move(path), op, value_rhs);
```
dconeybe, line 109: Should `path` be moved, since it's received by the method as a const reference? This looks on the surface to be a missed occurrence of changing `std::move(path)` to simply `path`.

Firestore/core/src/core/field_filter.cc
```diff
@@ -173,24 +170,25 @@ bool FieldFilter::Rep::MatchesComparison(ComparisonResult comparison) const {
...
 size_t FieldFilter::Rep::Hash() const {
-  return util::Hash(field_, op_, value_rhs_);
+  return util::Hash(field_, op_, model::CanonicalId(*value_rhs_));
 }
 
 bool FieldFilter::Rep::Equals(const Filter::Rep& other) const {
   if (type() != other.type()) return false;
 
-  const auto& other_rep = static_cast<const FieldFilter::Rep&>(other);
+  const auto& other_rep = dynamic_cast<const FieldFilter::Rep&>(other);
```
dconeybe, line 189: Can you explain this change from `static_cast` to `dynamic_cast`? This looks problematic because wouldn't you need RTTI enabled for this to work correctly? Also, this would literally be the only usage of `dynamic_cast` in the entire repository, which looks suspicious.

Firestore/core/src/core/key_field_in_filter.h
```diff
@@ -32,15 +34,15 @@ namespace core {
...
+                   google_firestore_v1_Value value);
 
  private:
   class Rep;
 
-  static bool Contains(const model::FieldValue::Array& array_value,
-                       const model::Document& doc);
-
-  static void ValidateArrayValue(const model::FieldValue& value);
+  static std::set<model::DocumentKey> ExtractDocumentKeysFromValue(
```
dconeybe, line 44: `std::set` generally has poor performance characteristics because it uses a linked structure internally that is not cache friendly on modern CPUs. Is it possible to use `std::unordered_set` instead? Also in `key_field_not_in_filter.h`.

dconeybe (approved):

PR #8041, opened 2021-05-05
FIRAppAttestArtifactStorage implementation and tests

maksymmalyhin:
#no-changelog


FirebaseAppCheck/Tests/Unit/AppAttestProvider/Storage/FIRAppAttestArtifactStorageTests.m
```diff
@@ -0,0 +1,146 @@
...
+  self.appName = @"FIRAppAttestArtifactStorageTests";
+  self.appID = @"1:100000000000:ios:aaaaaaaaaaaaaaaaaaaaaaaa";
+
+  self.storage = [[FIRAppAttestArtifactStorage alloc] initWithAppName:self.appName
+                                                                appID:self.appID
+                                                          accessGroup:nil];
+}
+
+- (void)tearDown {
+  self.storage = nil;
```
ncooke3, line 45: Maybe call `[self.storage setArtifact:nil]` before `self.storage = nil` to clean up left over storage from `testSetAndGetArtifact` (so the left over storage doesnt persist between test runs).

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestArtifactStorage.m
```diff
@@ -22,16 +22,82 @@
...
+
+- (FBLPromise<NSData *> *)getArtifact {
+  return [self.keychainStorage getObjectForKey:[self artifactKey]
+                                   objectClass:[NSData class]
+                                   accessGroup:self.accessGroup]
+      .then(^NSData *(id<NSSecureCoding> storedArtifact) {
+        if ([(NSObject *)storedArtifact isKindOfClass:[NSData class]]) {
+          return (NSData *)storedArtifact;
+        } else {
+          return nil;
```
ncooke3, line 75: Should a promised rejected with some error be returned instead?

ncooke3 (approved):
LGTM after addressing comments!

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestArtifactStorage.h
```diff
@@ -25,9 +25,14 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol FIRAppAttestArtifactStorageProtocol <NSObject>
 
 /// Set the artifact.
+/// @param artifact The artifact data to store.
+/// @return An artifact that is resolved with the artifact data passed into the method in case of
+/// success or is rejected with an error.
```
ncooke3, line 30: optional: mention that passing `nil` will clear any existing artifact data from storage.

ncooke3 (approved):

FirebaseAppCheck/Sources/AppAttestProvider/Storage/FIRAppAttestArtifactStorage.m
maksymmalyhin: Maybe... The method should return `nil` if the artifact was not found. The case when something else was stored in the keychain looks more similar to "not found" case so we eventually should overwrite the value. WDYT?

PR #8043, opened 2021-05-05
RemoteDocumentCache changes

schmidt-sebastian:
RemoteDocumentCache changes for Protobuf migration.

### DIFF
--- a/Firestore/core/src/local/leveldb_remote_document_cache.cc
+++ b/Firestore/core/src/local/leveldb_remote_document_cache.cc
@@ -251,7 +251,7 @@ MutableDocumentMap LevelDbRemoteDocumentCache::GetMatching(
       tasks.Execute([this, &results, document_key, contents] {
         MutableDocument document = DecodeMaybeDocument(contents, document_key);
         if (document.is_found_document()) {
-          results.Insert(MutableDocument(document));
+          results.Insert(document);
         }
       });
     }
--- a/Firestore/core/src/local/leveldb_remote_document_cache.cc
+++ b/Firestore/core/src/local/leveldb_remote_document_cache.cc
@@ -26,7 +26,7 @@
 #include "Firestore/core/src/local/leveldb_persistence.h"
 #include "Firestore/core/src/local/local_serializer.h"
 #include "Firestore/core/src/model/document_key_set.h"
-#include "Firestore/core/src/model/document_map.h"
+#include "Firestore/core/src/model/mutable_document.h"
 #include "Firestore/core/src/nanopb/message.h"
 #include "Firestore/core/src/nanopb/reader.h"
 #include "Firestore/core/src/util/background_queue.h"
@@ -42,13 +42,10 @@ namespace {
 
 using core::Query;
 using leveldb::Status;
-using model::Document;
 using model::DocumentKey;
 using model::DocumentKeySet;
-using model::DocumentMap;
-using model::MaybeDocument;
-using model::MaybeDocumentMap;
-using model::OptionalMaybeDocumentMap;
+using model::MutableDocument;
+using model::MutableDocumentMap;
 using model::ResourcePath;
 using model::SnapshotVersion;
 using nanopb::ByteString;
@@ -106,7 +103,7 @@ LevelDbRemoteDocumentCache::LevelDbRemoteDocumentCache(
 // Out of line because of unique_ptrs to incomplete types.
 LevelDbRemoteDocumentCache::~LevelDbRemoteDocumentCache() = default;
 
-void LevelDbRemoteDocumentCache::Add(const MaybeDocument& document,
+void LevelDbRemoteDocumentCache::Add(const MutableDocument& document,
                                      const SnapshotVersion& read_time) {
   const DocumentKey& key = document.key();
   const ResourcePath& path = key.path();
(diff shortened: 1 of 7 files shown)


Firestore/core/src/local/memory_remote_document_cache.cc
```diff
@@ -55,31 +54,32 @@ void MemoryRemoteDocumentCache::Remove(const DocumentKey& key) {
...
+MutableDocumentMap MemoryRemoteDocumentCache::GetAll(
     const DocumentKeySet& keys) {
-  OptionalMaybeDocumentMap results;
+  MutableDocumentMap results;
   for (const DocumentKey& key : keys) {
     // Make sure each key has a corresponding entry, which is nullopt in case
     // the document is not found.
     // TODO(http://b/32275378): Don't conflate missing / deleted.
-    results = results.insert(key, Get(key));
+    results = results.insert(key, std::move(Get(key)));
```
var-const, line 71: You don't need to move a temporary -- it is an rvalue and will trigger the move constructor anyway.

Firestore/core/src/local/leveldb_remote_document_cache.cc
```diff
@@ -254,29 +249,29 @@ DocumentMap LevelDbRemoteDocumentCache::GetMatching(
 
       const std::string& contents = it->value();
       tasks.Execute([this, &results, document_key, contents] {
-        MaybeDocument maybe_doc = DecodeMaybeDocument(contents, document_key);
-        if (maybe_doc.is_document()) {
-          results.Insert(Document(maybe_doc));
+        MutableDocument document = DecodeMaybeDocument(contents, document_key);
+        if (document.is_found_document()) {
+          results.Insert(MutableDocument(document));
```
var-const, line 254: Nit: you don't need to call the `MutableDocument` constructor here because `document` is already of the right type.

Firestore/core/test/unit/local/remote_document_cache_test.h
```diff
@@ -53,10 +54,18 @@ class RemoteDocumentCacheTest : public ::testing::TestWithParam<FactoryFunc> {
...
+  model::MutableDocument SetTestDocument(absl::string_view path,
+                                         google_firestore_v1_Value data);
+  model::MutableDocument SetTestDocument(absl::string_view path,
+                                         int update_time,
+                                         int read_time);
+  model::MutableDocument SetTestDocument(absl::string_view path,
+                                         google_firestore_v1_Value data,
+                                         int update_time,
+                                         int read_time);
+  void VerifyValue(model::MutableDocument actual_doc,
```
var-const, line 67: Nit: pass `actual_doc` by const reference?

Firestore/core/test/unit/local/remote_document_cache_test.cc
```diff
@@ -230,34 +232,85 @@ TEST_P(RemoteDocumentCacheTest, DocumentsMatchingUsesReadTimeNotUpdateTime) {
...
+  });
+}
 // MARK: - Helpers
 
-Document RemoteDocumentCacheTest::SetTestDocument(const absl::string_view path,
-                                                  int update_time,
-                                                  int read_time) {
-  Document doc = Doc(path, update_time, kDocData);
+MutableDocument RemoteDocumentCacheTest::SetTestDocument(
+    const absl::string_view path,
```
var-const, line 278: Nit: I'd suggest omitting this `const` from `path` -- we generally avoid const parameters, so this one looks like it accidentally sneaked in.

Firestore/core/test/unit/local/remote_document_cache_test.cc
```diff
@@ -230,34 +232,85 @@ TEST_P(RemoteDocumentCacheTest, DocumentsMatchingUsesReadTimeNotUpdateTime) {
...
     const absl::string_view path) {
-  return SetTestDocument(path, kVersion, kVersion);
+  return SetTestDocument(path, DeepClone(kDocData), kVersion, kVersion);
+}
+
+void RemoteDocumentCacheTest::VerifyValue(MutableDocument actual_doc,
+                                          google_firestore_v1_Value data) {
+  MutableDocument expected_doc =
+      Doc(actual_doc.key().ToString(), kVersion, data);
+  ASSERT_EQ(expected_doc, actual_doc);
```
var-const, line 306: Can you use `EXPECT` rather than `ASSERT`? Also, if the failure happens in `VerifyValue`, it won't be obvious which invocation triggered it. You can either use [`SCOPED_TRACE`](https://github.com/google/googletest/blob/master/docs/advanced.md#adding-traces-to-assertions) before every invocation or make `VerifyValue` a matcher.

Firestore/core/test/unit/local/remote_document_cache_test.cc
```diff
@@ -230,34 +232,85 @@ TEST_P(RemoteDocumentCacheTest, DocumentsMatchingUsesReadTimeNotUpdateTime) {
...
 }
 
 void RemoteDocumentCacheTest::SetAndReadTestDocument(
     const absl::string_view path) {
   persistence_->Run("SetAndReadTestDocument", [&] {
-    Document written = SetTestDocument(path);
-    absl::optional<MaybeDocument> read = cache_->Get(testutil::Key(path));
+    MutableDocument written = SetTestDocument(path);
+    absl::optional<MutableDocument> read = cache_->Get(Key(path));
     ASSERT_EQ(*read, written);
```
var-const, line 314: Same comments as `VerifyValue`.

var-const (changes requested):

Firestore/core/test/unit/local/remote_document_cache_test.cc
schmidt-sebastian: I added a customer matcher.

Firestore/core/test/unit/local/remote_document_cache_test.cc
schmidt-sebastian: Used EXPECT_EQ. Leaving the Scoped Trace portion as an exercise to the reader as this is existing code.

Firestore/core/test/unit/local/remote_document_cache_test.h
schmidt-sebastian: Not used anymore

var-const (approved):

PR #8044, opened 2021-05-05
Enable allowLocalhostRequest for Emulator requests

schmidt-sebastian:
For older iOS version, we need to explicitly enable request to `localhost` when we use the Emulator:

https://source.corp.google.com/piper///depot/google3/third_party/objective_c/gtm_session_fetcher/Source/GTMSessionFetcher.h;l=898?q=allowLocalhostRequest    

This should fix the test failure in Google3.


FirebaseStorage/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 8.0.1
```
paulb777, line 1: We only do patches for emergency breakages now. Otherwise all Firebase products version together with minor updates.  This should be 8.1.0 or if it's serious enough we should cherry-pick to 8.0.0 which hasn't released yet.

FirebaseStorage/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+- [fixed] Fixed an issue that prevented the SDK from connecting to the Emulator
+  on iOS 8 or older.
```
paulb777, line 3: We currently support iOS 10 and above only. My understanding is that google3 is testing on iOS 12.

paulb777 (changes requested):
Please investigate the CI failures.

paulb777 (approved):

FirebaseStorage/CHANGELOG.md
schmidt-sebastian: This turned out to be bug in the tests that took forever to hunt down. Thanks for bearing with me :)

PR #8049, opened 2021-05-07
Fix Crashlytics qs in nightly zip test

### LINKED ISSUE #8022: Nightly Testing Report
# Nightly Testing Report
This issue([log](https://github.com/firebase/firebase-ios-sdk/actions/runs/802250734)) is generated at 05/01/2021 06:08 AM, fetching workflow runs triggered in the last 24 hrs.
| Workflow |04/30|
| -------- | -------- |
| [prerelease](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/prerelease.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/801713750) |

paulb777:
Fix #8022 

#no-changelog


ncooke3:
Looks like the source for ReachabilitySwift is only one file. We could try the [manual installation](https://github.com/ashleymills/Reachability.swift#manual) method. Hopefully that'd prevent any 'import' issues.

paulb777:
@ncooke3 Interesting. That would make the test more robust, but at the tradeoff of doing less integration testing and making the quickstart less exemplary.

paulb777:
@granluo Will you take a look to see if you can tell why this fix failed to build?

granluo:
> @granluo Will you take a look to see if you can tell why this fix failed to build?
```
2021-05-07T16:25:10.9392750Z ++ mv -n '/Users/runner/ios_frameworks/Firebase/FirebaseCrashlytics/*' Firebase/
2021-05-07T16:25:10.9395440Z   inflating: /Users/runner/ios_frameworks/Firebase/Binaries/ReachabilitySwift.xcframework/ios-arm64_x86_64-maccatalyst/ReachabilitySwift.framework/ReachabilitySwift  
2021-05-07T16:25:10.9397450Z mv: rename /Users/runner/ios_frameworks/Firebase/FirebaseCrashlytics/* to Firebase/*: No such file or directory
```
It could be caused by copying explicit '/Users/runner/ios_frameworks/Firebase/FirebaseCrashlytics/*'. Will keep investigating.

paulb777:
Interesting. I was following the pattern from the FirebaseAuth example above. I wonder why Crashlytics would be different?

Thanks for investigating!

granluo:
Yep will test it locally since it is building zip which takes a long time in presubmit.

granluo:
Aha, I think that is probably caused by `mv`, `CrashlyticsExample` in the previous job has moved files and so that dir is empty. and there is no more contents in that dir while testing `CrashlyticsSwiftExample`

paulb777:
Good find! I'll rebase this to master and retry the CI in this PR

paulb777:
Another possibility is the ZipBuilder is not properly building the module so that it does not get found.

paulb777:
The `Reachability` Pod is different. We need `ReachabilitySwift` pod with its `Reachability` module.  

At this point, we might want to timebox this investigation, since the perfect fix may be too expensive. I suspect something with the ZipBuilder and Swift modules may not be correct for this case.

Here are a few alternate solutions:

- Use CocoaPods just for ReachabilitySwift and test the generated xcworkspace
- Use CocoaPods to install Reachability.swift and add it to the xcodeproj
- Disable the swift test for now.

Note we still need the `-n` removal change that is in this PR.

granluo:
Could we [add the Reachability.swift file](https://github.com/ashleymills/Reachability.swift#manual) to the project?

paulb777:
If we do that, my preference would be to only to it for the zip test, so that QuickStart still uses CocoaPods.

paulb777:
Clean CI run at https://github.com/firebase/firebase-ios-sdk/runs/2558759039?check_suite_focus=true

PR #8050, opened 2021-05-07
Preserve order in ArrayUnion

schmidt-sebastian:
The elements added in an ArrayUnion should be added in order.


dconeybe (approved):
Is there test coverage for this case? If not, please add a test case.

PR #8070, opened 2021-05-13
Update Query code to use Protobuf

schmidt-sebastian:


Firestore/core/src/core/bound.cc
```diff
@@ -73,15 +80,16 @@ bool Bound::SortsBeforeDocument(const OrderByList& order_by,
 
 std::string Bound::CanonicalId() const {
   std::string result = before_ ? "b:" : "a:";
-  for (const FieldValue& component : position_) {
-    result.append(component.ToString());
+  for (pb_size_t i = 0; i < position_->values_count; ++i) {
+    result.append(model::CanonicalId(position_->values[i]));
```
var-const, line 84: Optional: add a using declaration for `CanonicalId`?

var-const (approved):

Firestore/core/src/core/bound.cc
schmidt-sebastian: Done

PR #8074, opened 2021-05-13
Serializer should guard against invalid input.

ehsannas:


schmidt-sebastian (commented):
:/ I do think this will not work anymore after the FieldValue rewrite as we will just pass the Protobuf through from the backend.

ehsannas:
No worries. I was mainly using this task to get familiar with the Serializer class. Should I close this PR then?

schmidt-sebastian:
There might still be a need for this after the migration has completed, but I would "abandon" in for the next couple of days and get back to it after the merge.

ehsannas:
Sounds good.

Firestore/core/test/unit/remote/serializer_test.cc
```diff
@@ -1742,6 +1742,60 @@ TEST_F(SerializerTest, DecodesVersionWithTargets) {
   ExpectDeserializationRoundTrip(model, proto);
 }
 
+TEST_F(SerializerTest, GracefullyDecodesBadMapValue) {
+  pb_bytes_array_t key{1, {'a'}};
+  google_firestore_v1_MapValue_FieldsEntry fields_entry;
```
var-const, line 1747: Nit: it seems a little inconsistent that some of these C structs are initialized to zeros and some aren't. Note also that for simple structs with few members, you should be able to use aggregate initialization:
```cpp
google_firestore_v1_Value bad_value = {};
google_firestore_v1_MapValue_FieldsEntry fields_entry = {&key, bad_value};
```

Feel free to ignore since this test will likely go away.

var-const (approved):
After syncing with @schmidt-sebastian Sebastian offline, it still seems to me like this PR should be merged. The `DatabaseId` check (which really is the focus of this fix) will still be relevant (though it will change form). The other two checks will go away, but that doesn't seem like a huge deal. Sebastian, please let me know if you agree.

Firestore/core/test/unit/remote/serializer_test.cc
schmidt-sebastian: I have run into a fair number of problems when I didn't initialize the value to zero. I think we should fix this.

schmidt-sebastian (commented):
I am okay with merging this - for extra credit into the mrschmidt/migrate branch :)

Firestore/core/test/unit/remote/serializer_test.cc
ehsannas: Done

ehsannas:
I don't know why this PR fell off my radar. @schmidt-sebastian It wasn't clear to me whether you want me to merge this into `main` or to your `merge` branch?

schmidt-sebastian:
I think this PR will have to look different once we merge `mrschmidt/mutabledocuments`. If you want to be the one to determine its future, then I would recommend rebasing against that branch.

PR #8075, opened 2021-05-14
Remaining Bundle changes

schmidt-sebastian:
Please see Android for any datatype changes.

### DIFF
--- a/Firestore/core/test/unit/bundle/bundle_serializer_test.cc
+++ b/Firestore/core/test/unit/bundle/bundle_serializer_test.cc
@@ -599,6 +599,29 @@ TEST_F(BundleSerializerTest, DecodesReferenceValues) {
   VerifyFieldValueRoundtrip(value);
 }
 
+TEST_F(BundleSerializerTest, DecodeReferenceValuesFromOtherProjectsFails) {
+  ProtoValue value;
+  value.set_reference_value(
+      "projects/p1/databases/new/documents/bundle/test_doc");
+  ProtoDocument document = TestDocument(value);
+
+  std::string json_string;
+  MessageToJsonString(document, &json_string);
+
+  VerifyJsonStringDecodeFails(json_string);
+}
+
+TEST_F(BundleSerializerTest, DecodeInvalidReferenceFails) {
+  ProtoValue value;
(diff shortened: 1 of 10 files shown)


Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -598,29 +599,6 @@ TEST_F(BundleSerializerTest, DecodesReferenceValues) {
   VerifyFieldValueRoundtrip(value);
 }
 
-TEST_F(BundleSerializerTest, DecodeReferenceValuesFromOtherProjectsFails) {
```
schmidt-sebastian, line 601: This now logs a debug message when the reference is accessed.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -598,29 +599,6 @@ TEST_F(BundleSerializerTest, DecodesReferenceValues) {
...
-
-  std::string json_string;
-  MessageToJsonString(document, &json_string);
-
-  VerifyJsonStringDecodeFails(json_string);
-}
-
-TEST_F(BundleSerializerTest, DecodeInvalidReferenceFails) {
-  ProtoValue value;
-  value.set_reference_value("projectxx/p1/datmm/new/documents/bundle/test_doc");
```
schmidt-sebastian, line 615: This now causes an assert in the client...

wu-hui: But does this mean a corrupted bundle can crash the SDK now?

schmidt-sebastian: If the path is not syntactically valid, the yes. If the path is valid but points to a different project, then we log a warning. It's the same on Android (https://github.com/firebase/firebase-android-sdk/blob/7be699f74d6effca46e765d22bd7635d78d17253/firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java#L267) and on Web, where we just pass the entire message through. The alternative is to pre-parse and validate all data, but that defeats the purpose of the FieldValue to Protobuf migration a bit.

wu-hui (approved):
LGTM, thank you!

PR #8076, opened 2021-05-14
Migrate Serializer to Protobuf

schmidt-sebastian:
This migrates the Serializer code to directly use Protobuf.

### DIFF
--- a/Firestore/core/src/core/bound.cc
+++ b/Firestore/core/src/core/bound.cc
@@ -31,7 +31,6 @@ namespace firebase {
 namespace firestore {
 namespace core {
 
-using model::CanonicalId;
 using model::Compare;
 using model::DocumentKey;
 using model::FieldPath;
@@ -82,14 +81,15 @@ bool Bound::SortsBeforeDocument(const OrderByList& order_by,
 std::string Bound::CanonicalId() const {
   std::string result = before_ ? "b:" : "a:";
   for (pb_size_t i = 0; i < position_->values_count; ++i) {
-    result.append(CanonicalId(position_->values[i]));
+    result.append(model::CanonicalId(position_->values[i]));
   }
   return result;
 }
 
 std::string Bound::ToString() const {
   return util::StringFormat("Bound(position=%s, before=%s)",
-                            CanonicalId(*position_), util::ToString(before_));
+                            model::CanonicalId(*position_),
+                            util::ToString(before_));
 }
 
 std::ostream& operator<<(std::ostream& os, const Bound& bound) {
--- a/Firestore/core/src/local/local_serializer.cc
+++ b/Firestore/core/src/local/local_serializer.cc
@@ -31,17 +31,15 @@
 #include "Firestore/core/src/bundle/named_query.h"
 #include "Firestore/core/src/core/query.h"
 #include "Firestore/core/src/local/target_data.h"
-#include "Firestore/core/src/model/document.h"
-#include "Firestore/core/src/model/field_value.h"
+#include "Firestore/core/src/model/mutable_document.h"
 #include "Firestore/core/src/model/mutation_batch.h"
-#include "Firestore/core/src/model/no_document.h"
 #include "Firestore/core/src/model/snapshot_version.h"
-#include "Firestore/core/src/model/unknown_document.h"
 #include "Firestore/core/src/nanopb/byte_string.h"
 #include "Firestore/core/src/nanopb/message.h"
 #include "Firestore/core/src/nanopb/nanopb_util.h"
 #include "Firestore/core/src/util/hard_assert.h"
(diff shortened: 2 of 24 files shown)


Firestore/core/src/core/bound.cc
```diff
@@ -82,14 +81,15 @@ bool Bound::SortsBeforeDocument(const OrderByList& order_by,
 std::string Bound::CanonicalId() const {
   std::string result = before_ ? "b:" : "a:";
   for (pb_size_t i = 0; i < position_->values_count; ++i) {
-    result.append(CanonicalId(position_->values[i]));
+    result.append(model::CanonicalId(position_->values[i]));
```
schmidt-sebastian, line 84: FYI this fix from the last PR broke the build because there is a name conflict between model::CanonicalId and CanonicalId

Firestore/core/test/unit/remote/serializer_test.cc
```diff
@@ -954,35 +914,13 @@ TEST_F(SerializerTest, BadStringValue) {
       Status(Error::kErrorDataLoss, "ignored"), bytes);
 }
 
-TEST_F(SerializerTest, BadTimestampValue_TooLarge) {
```
var-const, line 957: Question: do these no longer apply?

var-const (changes requested):

Firestore/core/test/unit/remote/serializer_test.cc
var-const: And similarly in the C++ public API, right?

Firestore/core/src/remote/serializer.cc
```diff
@@ -1063,25 +802,16 @@ google_firestore_v1_StructuredQuery_Filter Serializer::EncodeFilters(
...
-      MakeArray<google_firestore_v1_StructuredQuery_Filter>(count);
-  pb_size_t i = 0;
-  for (const auto& filter : filters) {
-    if (filter.IsAFieldFilter()) {
-      HARD_ASSERT(i < count, "Index out of bounds");
-      composite.filters[i] = EncodeSingularFilter(FieldFilter{filter});
-      ++i;
-    }
-  }
+  SetRepeatedField(
```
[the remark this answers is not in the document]
var-const: Perhaps file a bug to find ways to improve `SetRepeatedField`? I had some ideas related to that.

Also, very optional: note that you could build it on your Linux workstation (assuming you're building locally now) which is probably more powerful and also won't slow down your local machine. Also, if building with CMake, make sure you use the `-j` argument to use cores efficiently.

Firestore/core/src/local/local_serializer.cc
```diff
@@ -352,6 +350,8 @@ MutationBatch LocalSerializer::DecodeMutationBatch(
           transform_mutation.transform.field_transforms_count;
       new_mutation.update_transforms =
           transform_mutation.transform.field_transforms;
+      // Prevent double-freeing of the write's fields. The field are now owned
```
var-const, line 353: Nit: `s/the field are/the fields are/` (applies throughout).

Firestore/core/src/remote/serializer.cc
```diff
@@ -578,19 +578,23 @@ FieldTransform Serializer::DecodeFieldTransform(
     }
 
     case google_firestore_v1_DocumentTransform_FieldTransform_append_missing_elements_tag: {  // NOLINT
-      auto field_transform = FieldTransform(
+      FieldTransform field_transform(
           std::move(field), ArrayTransform(TransformOperation::Type::ArrayUnion,
                                            proto.append_missing_elements));
-      proto.append_missing_elements = {};  // Release field ownership
+      // Release field ownership to prevent doube-freeing. The values are now
```
var-const, line 584: Nit: `s/doube/double/`.

var-const (approved):

Firestore/core/src/local/local_serializer.cc
schmidt-sebastian: Fixed

Firestore/core/src/remote/serializer.cc
schmidt-sebastian: Fixed

Firestore/core/src/remote/serializer.cc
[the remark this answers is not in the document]
schmidt-sebastian: Filed b/188469292

I am building with 8 cores locally - makes the room nice and hot :)

Firestore/core/test/unit/remote/serializer_test.cc
schmidt-sebastian: Yes - https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/timestamp.cc#L64

Issue #8082, opened 2021-05-14
Fix GHA release zip build process


paulb777:
The zip.yml script is using the `--local-podspec-path` option with master branch. That is useful for daily CI.

However, for releases we should use the release branch and/or only SpecsStaging.

We should run all of the jobs in zip.yml after both a build with `--local-podspec-path` and [without it.](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build_zip.sh#L37) 

This could easily be done by duplicating almost the entire contents of zip.yml, but it would be nice to find a more DRY way.

PR #8084, opened 2021-05-15
Restore Carthage

paulb777:
Restore Carthage distribution now that Carthage 0.38.0 supports binary xcframework distributions.

This PR removes the code that was repackaging build xcframeworks as frameworks for Carthage. Now we distribute exactly the same xcframeworks in the zip distribution as Carthage with the exception of the rebuilt FirebaseCoreDiagnostics.

Successful Carthage 8.0.0 generation in https://github.com/firebase/firebase-ios-sdk/actions/runs/845683211 after generating package on GHA after disabling `--local-podspec-path` from `build_zip.sh`. More context about this in #8082.


ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -354,6 +358,17 @@ struct FrameworkBuilder {
     // Get the framework Headers directory. On macOS, it's a symbolic link.
     let headersDir = archivePath.appendingPathComponent("Headers").resolvingSymlinksInPath()
 
+    // The macOS Headers directory can have a Headers file in it symbolically linked to nowhere.
```
paulb777, line 361: Fixing a zip distribution bug caught by the Carthage hash generation

ReleaseTooling/CarthageJSON/FirebaseMLModelDownloaderBinary.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
ryanwilson, line 2: Hmm we currently don't build this in the zip, do we? I thought we had excluded it

ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift
```diff
@@ -704,172 +706,4 @@ struct FrameworkBuilder {
     }
     return xcframework
   }
-
```
ryanwilson, line 707: Yessssssss so happy all this code is going away 🥳

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -231,30 +230,29 @@ struct ZipBuilder {
...
                                                  logsOutputDir: paths.logsOutputDir,
+                                                 setCarthage: false,
                                                  podInfo: podInfo)
           groupedFrameworks[podName] = (groupedFrameworks[podName] ?? []) + frameworks
-          if platform == .iOS {
-            if let carthageFramework = carthageFramework {
-              carthageToInstall[podName] = [carthageFramework]
-            }
+
+          if podName == "FirebaseCoreDiagnostics" {
```
ryanwilson, line 241: Do we want to also gate this on `includeCarthage`? Otherwise we'll be compiling CD when we don't need to. Same in the below few blocks

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -756,28 +765,17 @@ struct ZipBuilder {
...
       do {
         try fileManager.copyItem(at: framework, to: zipLocation)
-        try fileManager.copyItem(at: framework, to: carthageLocation)
       } catch {
         fatalError("Cannot copy framework at \(framework) while " +
           "attempting to generate frameworks. \(error)")
       }
       frameworks.append(zipLocation)
-
-      CarthageUtils.generatePlistContents(
```
ryanwilson, line 774: Is this not relevant anymore?

ryanwilson:
Thanks for getting this update so quickly!

ReleaseTooling/CarthageJSON/FirebaseMLModelDownloaderBinary.json
paulb777: Doesn't look like it. It's part of 8.0.0. Perhaps we only excluded for Carthage before but now that Carthage matches zip, we shouldn't?

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
paulb777: Any reason, we shouldn't always build Carthage when we're building a Firebase distro. How about removing the `includeCarthage` flag in a subsequent PR?

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
paulb777: It's done in the xcframework build.  Ideally `generatePlistContents` should be moved out of CarthageUtils.  I can do that in a subsequent PR.

ReleaseTooling/CarthageJSON/FirebaseMLModelDownloaderBinary.json
ryanwilson: Discussed offline, SGTM on including it.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
ryanwilson: SGTM to move in a subsequent PR and merging as is.

ryanwilson (approved):

PR #8088, opened 2021-05-16
FirebaseCombineSwift: Docs & API cleanup

ncooke3:
- Reviewed and cleaned up some of the documentation
- Tried to standardize/cleanup some of the`#if` compilation logic
- `FirebaseCombineSwift` now conditionally builds for all platforms (mainly had to get watchOS compiling)
- Removed remaining `weak` references to `self` after reading [this note](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCombineSwift/DECISIONS.md#using-capture-lists) from `FirebaseCombineSwift/DECISIONS.md`

#no-changelog

### DIFF
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -440,7 +440,7 @@
       ///   protocol. This is used for presenting the web context. If `nil`, a default `AuthUIDelegate`
       ///   will be used.
       /// - Returns: A publisher that emits an `AuthDataResult` when the sign-in flow completed
-      ///   successfully, or an error otherwise.
+      ///   successfully, or an error otherwise. The publisher will emit on the *main* thread.
       /// - Remark: Possible error codes:
       ///   - `AuthErrorCodeOperationNotAllowed` - Indicates that email and password accounts are not enabled.
       ///     Enable them in the Auth section of the Firebase console.
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -469,7 +469,7 @@
         }
       }
 
-    #endif  // os(iOS) || targetEnvironment(macCatalyst)
+    #endif // os(iOS) || targetEnvironment(macCatalyst)
 
     /// Asynchronously signs in to Firebase with the given Auth token.
     ///
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -26,7 +26,7 @@
     ///
     /// The publisher emits values when:
     ///
-    /// - A user is registered,
+    /// - It is registered,
     /// - a user with a different UID from the current user has signed in, or
     /// - the current user has signed out.
     ///
@@ -50,7 +50,7 @@
     ///
     /// The publisher emits values when:
     ///
-    /// - A user is registered,
+    /// - It is registered,
     /// - a user with a different UID from the current user has signed in,
     /// - the ID token of the current user has been refreshed, or
     /// - the current user has signed out.
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -26,9 +26,9 @@
     ///
     /// The publisher emits values when:
     ///
-    /// - It is registered,
-    /// - A user with a different UID from the current user has signed in, or
-    /// - The current user has signed out.
+    /// - A user is registered,
+    /// - a user with a different UID from the current user has signed in, or
+    /// - the current user has signed out.
     ///
     /// The publisher will emit events on the **main** thread.
     ///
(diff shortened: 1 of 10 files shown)


FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
```diff
@@ -420,19 +422,19 @@
...
       /// The publisher will emit events on the **main** thread.
       ///
       /// - Parameters:
       ///   - provider: An instance of an auth provider used to initiate the sign-in flow.
-      ///   - uiDelegate: Optionally an instance of a class conforming to the `AuthUIDelegate`
+      ///   - uiDelegate: Optionally, an instance of a class conforming to the `AuthUIDelegate`
       ///   protocol. This is used for presenting the web context. If `nil`, a default `AuthUIDelegate`
       ///   will be used.
       /// - Returns: A publisher that emits an `AuthDataResult` when the sign-in flow completed
-      ///   successfully, or an error otherwise. The publisher will emit on the *main* thread.
```
paulb777, line 435: Is this intended?

FirebaseCombineSwift/Sources/Auth/GameCenterAuthProvider+Combine.swift
```diff
@@ -18,15 +18,15 @@
   import FirebaseAuth
 
   @available(swift 5.0)
+  @available(macOS 10.15, macCatalyst 13, iOS 13, tvOS 13, *)
```
paulb777, line 21: Good spot!

It looks like we're generally inconsistent between 13 and 13.0 across the repo.

paulb777 (approved):
Thanks for the thorough cleanup!

peterfriese (changes requested):
Thanks, Nick!

I left some comments regarding the doc comments. Specifically, I think we should overshare that the publishers return on the main thread.

Thanks for catching the weak selfs!

As for the #if structure - this looks good to me, but I'd love to get @paulb777's or @ryanwilson's opinion on this as well.

FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
ncooke3: It wasn't, I was removing duplicate mentions of  _The publisher will emit on the *main* thread._ and happened to miss this. However, Peter brought up below that it's likely better to make that fact overly explicit for developers reading the docs so I'll add the removed statements back.

FirebaseCombineSwift/Sources/Auth/GameCenterAuthProvider+Combine.swift
ncooke3: I just looked and Apple uses **13.0** in their docs. I'll update this and elsewhere in Combine APIs to follow that convention.

FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
```diff
@@ -70,13 +70,13 @@
         .eraseToAnyPublisher()
     }
 
-    /// Sets the currentUser on the calling Auth instance to the provided user object.
+    /// Sets the `currentUser` on the calling Auth instance to the provided `User` instance.
     ///
     /// The publisher will emit events on the **main** thread.
     ///
     /// - Parameter user: The user object to be set as the current user of the calling Auth instance.
```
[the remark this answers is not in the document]
ncooke3: Good point, I'll just revert to "user object".

FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
```diff
@@ -156,13 +156,13 @@
       }
     }
 
-    /// Signs in using an email address and password.
+    /// Signs in a user with the given email address and password.
```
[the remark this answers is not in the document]
ncooke3: I agree with you. My rationale for this change and the one above was that: an Auth instance is the entity that facilitates the authentication of a user as opposed to being the entity that's getting authenticated. Im thinking this way because the only thing that is changing for the Auth instance is the currently signed in user (currentUser). So, if this is the way to think about it, then it might be more clear to phrase things like: "signs in **a user**", "creates **a user**".

cc: @morganchen12 WDTY about this and the above change?

FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
```diff
@@ -420,19 +422,19 @@
...
       ///
       /// - Parameters:
       ///   - provider: An instance of an auth provider used to initiate the sign-in flow.
-      ///   - uiDelegate: Optionally an instance of a class conforming to the `AuthUIDelegate`
+      ///   - uiDelegate: Optionally, an instance of a class conforming to the `AuthUIDelegate`
       ///   protocol. This is used for presenting the web context. If `nil`, a default `AuthUIDelegate`
       ///   will be used.
       /// - Returns: A publisher that emits an `AuthDataResult` when the sign-in flow completed
-      ///   successfully, or an error otherwise. The publisher will emit on the *main* thread.
+      ///   successfully, or an error otherwise.
```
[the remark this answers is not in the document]
ncooke3: Adding it to all of them to be extra clear

ncooke3 (commented):
Added all feedback and gave it a few passes to confirm I addressed everything. Cleaned up some docs so there were no line breaks shown in Xcode.. everything is looking neat!

ncooke3:
> As for the #if structure - this looks good to me, but I'd love to get @paulb777's or @ryanwilson's opinion on this as well.

cc: @paulb777, @ryanwilson

paulb777:
LGTM I like the consistency, but I'll defer to @ryanwilson's style expertise

ryanwilson (approved):
LGTM, I added a comment here but we should follow up later to do a wider check to see if this is the best way for availability checks (ideally all APIs can be marked with `@available` and not just fail to find the symbols)

peterfriese (approved):

PR #8093, opened 2021-05-17
Build release zip from GHA

### LINKED ISSUE #8082: Fix GHA release zip build process

paulb777:
Adds packaging of tagged release build - either from SpecsStaging or if already published from the CDN.  The version is specified in the FirebaseManifest.

Running the quickstarts probably isn't necessary since presumably they have already been tested with the master build. Also, I don't know of a good way to set that up without a lot of duplication.

Along the way, I found a few other miscellaneous issue. See the embedded review comments.

Confirmed expected diffs after temporarily changing some public headers in https://github.com/firebase/firebase-ios-sdk/actions/runs/857702464.

I also ran into and diagnosed an issue where Firestore stopped `pod installing` locally for me because of a `git` issue. Details at https://stackoverflow.com/a/67550246/556617 and internally at b/188526329.

Companion CL is cl/374304372

Fix #8082 

#no-changelog

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -23,7 +23,7 @@ on:
 jobs:
   package-release:
     # Don't run on private repo.
-#    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
+    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
     runs-on: macOS-latest
     steps:
     - uses: actions/checkout@v2
@@ -59,7 +59,7 @@ jobs:
 
   package-head:
     # Don't run on private repo.
-#    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
+    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
     needs: build
     runs-on: macOS-latest
     steps:
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -21,6 +21,30 @@ on:
         default: 'https://github.com/firebase/SpecsStaging.git'
 
 jobs:
+  package-release:
+    # Don't run on private repo.
+#    if: (github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk') || github.event_name == 'workflow_dispatch'
+    needs: build-head
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Xcode 12.2
+      run: sudo xcode-select -s /Applications/Xcode_12.2.app/Contents/Developer
+    - name: Setup Bundler
+      run: ./scripts/setup_bundler.sh
+    - name: ZipBuildingTest
+      run: |
+         mkdir -p release_zip_dir
+         sh -x scripts/build_zip.sh release_zip_dir \
+           "${{ github.event.inputs.custom_spec_repos || 'https://github.com/firebase/SpecsStaging.git' }}"
+    - uses: actions/upload-artifact@v1
+      with:
+        name: Firebase-release-zip-zip
+        # Zip the entire output directory since the builder adds subdirectories we don't know the
+        # name of.
+        path: release_zip_dir
+
+
   build:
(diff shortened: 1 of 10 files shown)


.github/workflows/zip.yml
```diff
@@ -21,6 +21,29 @@ on:
         default: 'https://github.com/firebase/SpecsStaging.git'
 
 jobs:
+  package-release:
```
paulb777, line 24: Add a 2nd package builder that builds with only staged and trunk pods.

The other package build was building from head.

ReleaseTooling/.gitignore
```diff
@@ -1,4 +0,0 @@
-.DS_Store
```
paulb777, line 1: All .gitignore should be done at the top of the repo.

ReleaseTooling/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -42,31 +42,21 @@ extension CarthageUtils {
   static func packageCarthageRelease(templateDir: URL,
                                      artifacts: ZipBuilder.ReleaseArtifacts,
                                      options: CarthageBuildOptions) -> URL? {
-    guard let zipLocation = artifacts.carthageDir else { return nil }
+    guard let carthagePath = artifacts.carthageDir else { return nil }
 
     do {
       print("Creating Carthage release...")
-      let carthagePath =
-        zipLocation.deletingLastPathComponent().appendingPathComponent("carthage_build")
```
paulb777, line 50: Another copy of the carthage artifacts is not needed since they were already in the dedicated CarthageFirebase directory.

ReleaseTooling/Sources/ZipBuilder/CarthageUtils.swift
```diff
@@ -42,31 +42,21 @@ extension CarthageUtils {
...
-      // Create a copy of the release directory since we'll be modifying it.
-      let fileManager = FileManager.default
-      fileManager.removeIfExists(at: carthagePath)
-      try fileManager.copyItem(at: zipLocation, to: carthagePath)
-
       // Package the Carthage distribution with the current directory structure.
-      let carthageDir = zipLocation.deletingLastPathComponent().appendingPathComponent("carthage")
-      fileManager.removeIfExists(at: carthageDir)
-      let output = carthageDir.appendingPathComponents([artifacts.firebaseVersion, "latest"])
```
paulb777, line 59: Remove redundant `latest` in the path to eliminate a release process step.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -418,9 +418,12 @@ struct ZipBuilder {
     let analyticsPods = analyticsFrameworks.map {
       $0.replacingOccurrences(of: ".framework", with: "")
     }
+    // Skip Analytics and the pods bundled with it.
     let remainingPods = installedPods.filter {
       $0.key != "FirebaseAnalytics" &&
         $0.key != "FirebaseCore" &&
+        $0.key != "FirebaseCoreDiagnostics" &&
```
paulb777, line 425: These were already skipped for the zip build, but not for the Carthage build.

scripts/build_zip.sh
```diff
@@ -22,10 +22,23 @@ EOF
   exit 1
 fi
 
+build_head_option=
```
paulb777, line 25: Two different option sets for building from head versus building release candidates

ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
```diff
@@ -23,9 +23,9 @@ import Foundation
 public let shared = Manifest(
   version: "8.0.0",
   pods: [
-    Pod("FirebaseCoreDiagnostics"),
-    Pod("FirebaseCore"),
-    Pod("FirebaseInstallations"),
+    Pod("FirebaseCoreDiagnostics", zip: true),
```
paulb777, line 26: Otherwise we always build these three from a published podspec and never from head.

ryanwilson (approved):

PR #8098, opened 2021-05-17
Always build Carthage with Firebase zip distributions

paulb777:
Follow up to #8084 to remove the `--enable-carthage-build` option and always build Carthage when building a Firebase zip distribution.

The other open comment in #8084 about the plist generation file requires more thought since it is being used both for xcframework builds and for the special Firebase framework that we're still assembling for Carthage.

Companion CL is cl/374299027


ryanwilson (approved):
Nice!

PR #8113, opened 2021-05-19
Combine: Add a sample app

peterfriese:
In addition to the tests that exercise the new API surface we're creating, we should add an example app that shows how to use the Combine-enabled APIs in context.

This PR adds a SwiftUI app that contains a basic drill-down menu navigation and a simple screen that shows how to use Anonymous Auth and `authStateDidChangePublisher`.

As we add more publishers, their use should be demonstrated in this app as well.

The app currently needs a prod Firebase project, so one of the next steps will be to enable it for local Emulators.

#no-changelog


paulb777 (approved):
This is a great framework for demonstrating Combine usage.

I was able to run after changing the signing and uncommenting the Combine products in the Package.swift.

The anonymous sign in silently failed for me - so it might be worth adding error checking to the app. It looks like my project didn't enable Auth after I stepped in the debugger: 

Screen Shot 2021-05-19 at 4 26 11 PM

peterfriese:
> This is a great framework for demonstrating Combine usage.
:-)

> I was able to run after changing the signing and uncommenting the Combine products in the Package.swift.
I can add a short text to the README for the demo project that says that, for the time being, people need to uncomment the Combine products.

> The anonymous sign in silently failed for me - so it might be worth adding error checking to the app. It looks like my project didn't enable Auth after I stepped in the debugger:
> 
> Screen Shot 2021-05-19 at 4 26 11 PM

Will add this to the README as well.

Example/CombineSample/README.md
```diff
@@ -0,0 +1,64 @@
+# Firebase & Combine Sample
+
+This sample demonstrates how to use Firebase's Combine APIs.
+
+## How to use
+
+### Set up a Firebase project
+
+1. Create a new Firebase project via the [Firebase console](https://console.firebase.google.com/)
+1. Enable the required Firebase services in the Firebase project you created in step 1
```
paulb777, line 10: Two 1's.

Example/CombineSample/README.md
```diff
@@ -0,0 +1,64 @@
...
+   * Firebase Authentication
+      * Enable Anonymous Auth
```
paulb777, line 12: Do you think the sample should also add error handling to give a nicer failure if this is missed?

paulb777 (commented):
Thanks!

Example/CombineSample/README.md
peterfriese: Markdown renders this correctly. Happy to change this to a properly enumerated list if our style guide mandates it.

Example/CombineSample/README.md
paulb777: Looks great - Thanks!

PR #8103, opened 2021-05-18
Potential integer overflow fix

elenadoty:

PR #8115, opened 2021-05-19
Remove unused method

elenadoty:
#no-changelog


samedson (approved):
Had a note in https://github.com/firebase/firebase-ios-sdk/pull/8103#pullrequestreview-663490576 about the changelog but otherwise LGTM!

PR #8117, opened 2021-05-19
App Check: prevent concurrent token requests

maksymmalyhin:
- prevent concurrent App Attest handshake operations
- prevent concurrent token refresh operations for any provider (on the App Check Core level)

#no-changelog

### DIFF
--- a/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
@@ -179,17 +179,18 @@ NS_ASSUME_NONNULL_BEGIN
                             if (self.ongoingGetTokenOperation == nil) {
                               // Kick off a new handshake sequence only when there is no an ongoing
                               // handshake to avoid race conditions.
-                              self.ongoingGetTokenOperation = [self createGetTokenSequencePromise]
+                              self.ongoingGetTokenOperation =
+                                  [self createGetTokenSequencePromise]
 
-                              // Release the ongoing operation promise on completion.
-                              .then(^FIRAppCheckToken *(FIRAppCheckToken *token) {
-                                self.ongoingGetTokenOperation = nil;
-                                return token;
-                              })
-                              .recover(^NSError *(NSError *error) {
-                                self.ongoingGetTokenOperation = nil;
-                                return error;
-                              });
+                                      // Release the ongoing operation promise on completion.
+                                      .then(^FIRAppCheckToken *(FIRAppCheckToken *token) {
+                                        self.ongoingGetTokenOperation = nil;
+                                        return token;
+                                      })
+                                      .recover(^NSError *(NSError *error) {
+                                        self.ongoingGetTokenOperation = nil;
+                                        return error;
+                                      });
                             }
                             return self.ongoingGetTokenOperation;
                           }];
--- a/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
@@ -177,7 +177,7 @@ NS_ASSUME_NONNULL_BEGIN
   return [FBLPromise onQueue:self.queue
                           do:^id _Nullable {
                             if (self.ongoingGetTokenOperation == nil) {
-                              // Kick off a new handshake sequence only when there is no an ongoing
+                              // Kick off a new handshake sequence only when there is not an ongoing
                               // handshake to avoid race conditions.
                               self.ongoingGetTokenOperation =
                                   [self createGetTokenSequencePromise]
--- a/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
@@ -174,13 +174,15 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (FBLPromise<FIRAppCheckToken *> *)getToken {
-  return [FBLPromise onQueue:self.queue do:^id _Nullable{
-    if (self.ongoingGetTokenOperation == nil) {
-      // Kick off a new handshake sequence only when there is no an ongoing handshake to avoid race conditions.
-      self.ongoingGetTokenOperation = [self createGetTokenSequencePromise];
-    }
-    return self.ongoingGetTokenOperation;
-  }];
+  return [FBLPromise onQueue:self.queue
+                          do:^id _Nullable {
+                            if (self.ongoingGetTokenOperation == nil) {
+                              // Kick off a new handshake sequence only when there is no an ongoing
+                              // handshake to avoid race conditions.
+                              self.ongoingGetTokenOperation = [self createGetTokenSequencePromise];
+                            }
+                            return self.ongoingGetTokenOperation;
+                          }];
 }
 
 - (FBLPromise<FIRAppCheckToken *> *)createGetTokenSequencePromise {
--- a/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
@@ -103,7 +103,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, readonly) id<FIRAppAttestKeyIDStorageProtocol> keyIDStorage;
 @property(nonatomic, readonly) id<FIRAppAttestArtifactStorageProtocol> artifactStorage;
 
-@property(nonatomic) FBLPromise<FIRAppCheckToken *> *ongoingGetTokenOperation;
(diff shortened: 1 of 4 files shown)


FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
```diff
@@ -172,6 +174,18 @@ - (void)getTokenWithCompletion:(void (^)(FIRAppCheckToken *_Nullable, NSError *_
 }
 
 - (FBLPromise<FIRAppCheckToken *> *)getToken {
+  return [FBLPromise onQueue:self.queue
+                          do:^id _Nullable {
+                            if (self.ongoingGetTokenOperation == nil) {
+                              // Kick off a new handshake sequence only when there is no an ongoing
```
ncooke3, line 180: typo: no → not

FirebaseAppCheck/Sources/Core/FIRAppCheck.m
```diff
@@ -242,6 +244,18 @@ - (nonnull NSString *)notificationTokenKey {
 #pragma mark - FAA token cache
 
 - (FBLPromise<FIRAppCheckToken *> *)retrieveOrRefreshTokenForcingRefresh:(BOOL)forcingRefresh {
+  return [FBLPromise do:^id _Nullable {
+    if (self.ongoingRetrieveOrRefreshTokenPromise == nil) {
+      // Kick off a new operation only when there is no an ongoing one.
```
ncooke3, line 249: typo: no → not

FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
```diff
@@ -172,6 +174,18 @@ - (void)getTokenWithCompletion:(void (^)(FIRAppCheckToken *_Nullable, NSError *_
...
+                              // handshake to avoid race conditions.
+                              self.ongoingGetTokenOperation = [self createGetTokenSequencePromise];
+                            }
+                            return self.ongoingGetTokenOperation;
+                          }];
```
ncooke3, line 185: Question,  should `ongoingGetTokenOperation` get reset back to `nil` once the promise assigned to it is resolved?

maksymmalyhin: Good catch! yes, it must be reset and tests should be added for it.

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m
```diff
@@ -621,6 +621,75 @@ - (void)testGetToken_WhenKeyRegisteredAndTokenExchangeRequestError {
...
+  OCMExpect([self.mockStorage getAppAttestKeyID])
+      .andReturn([FBLPromise resolvedWith:existingKeyID]);
+
+  // 3. Expect a stored artifact to be requested.
+  NSData *storedArtifact = [@"storedArtifact" dataUsingEncoding:NSUTF8StringEncoding];
+  OCMExpect([self.mockArtifactStorage getArtifactForKey:existingKeyID])
+      .andReturn([FBLPromise resolvedWith:storedArtifact]);
+
+  // 4. Expect random challenge to be requested.
+  // 4.1. Create a pending promise to fulfil later.
```
paulb777, line 641: nit: fulfill

FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
maksymmalyhin: Updated here and in FIRAppCheck, PTAL.

paulb777 (approved):
LGTM, but you may want to wait for @ncooke3 who will likely give a more thorough review

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m
```diff
@@ -621,6 +573,78 @@ - (void)testGetToken_WhenKeyRegisteredAndTokenExchangeRequestError {
...
+
+  // 3. Expect a stored artifact to be requested.
+  NSData *storedArtifact = [@"storedArtifact" dataUsingEncoding:NSUTF8StringEncoding];
+  OCMExpect([self.mockArtifactStorage getArtifactForKey:existingKeyID])
+      .andReturn([FBLPromise resolvedWith:storedArtifact]);
+
+  // 4. Expect random challenge to be requested.
+  // 4.1. Create a pending promise to fulfill later.
+  FBLPromise<NSData *> *challengeRequestPromise = [FBLPromise pendingPromise];
+  // 4.2. Stab getRandomChallenge method.
```
ncooke3, line 595: typo: Stab → Stub

FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
```diff
@@ -172,6 +174,29 @@ - (void)getTokenWithCompletion:(void (^)(FIRAppCheckToken *_Nullable, NSError *_
...
+                                      .then(^FIRAppCheckToken *(FIRAppCheckToken *token) {
+                                        self.ongoingGetTokenOperation = nil;
+                                        return token;
+                                      })
+                                      .recover(^NSError *(NSError *error) {
+                                        self.ongoingGetTokenOperation = nil;
+                                        return error;
+                                      });
+                            }
+                            return self.ongoingGetTokenOperation;
```
ncooke3, line 195: IIUC, we want to return a promise that resolves to a token (or error) and nils out the `ongoingGetTokenOperation` that it was being stored in when it finally resolves? 

While understanding this, I thought of something I'd like your thoughts on:
Could there be a race condition that emerges when the `.then` or `.recover` blocks execute fast enough where its's synchronous (particularly the `.recover` error case)? Because then `ongoingGetTokenOperation` may persist with a value and prevent future calls to `getToken` from spawning a new `getTokenSequencePromise`

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m
```diff
@@ -621,6 +573,78 @@ - (void)testGetToken_WhenKeyRegisteredAndTokenExchangeRequestError {
   [self verifyAllMocks];
 }
 
+#pragma mark - Request merging
+
+- (void)testGetToken_WhenCalledSeveralTimes_ThenThereIsOnlyOneOngoingHandshake {
```
ncooke3, line 578: great test!

FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
maksymmalyhin: Good question.

 Yes, the intention is to return a promise that nils out the `ongoingGetTokenOperation` on completion. 

Currently `ongoingGetTokenOperation` stores such a promise (includes `then` and `recovery` steps). So the intended sequence of events is:
1. the promise created and assigned to `ongoingGetTokenOperation`
2. new promises bound to `ongoingGetTokenOperation` are created in `getTokenWithCompletion:` method to invoke the completion blocks
3. the promise created by `[self createGetTokenSequencePromise]` is either fulfilled or rejected
4. `ongoingGetTokenOperation` is nilled out
  - if another `getTokenWithCompletion:`  method is called at this point the sequence will start from step 1.
5. promises created in `getTokenWithCompletion:`  are fulfilled or rejected correspondingly

The sequence looks correct to me and I cannot see a possible race condition here. Can you see anything missing/off?

ncooke3: Thank you for the detailed explanation. I think it looks good. I'm still trying to understand the ins and outs of how to closely examine promises.

Here I was trying to make sure that `ongoingGetTokenOperation ` is never returned from `getToken:` in a resolved state, because that would prevent other calls to `getToken:` from doing anything.

I think my original doubts were for when the following occurs:
1. `getTokenWithCompletion:` is called but app attest is not supported, so an `attestState.appAttestUnsupportedError` is returned from `createGetTokenSequencePromise:`
2. If this evaluates fast enough (because of the lack of network calls) that it's synchronous, then might the `.recover` block return an error to  `ongoingGetTokenOperation` **before** the process returns from `getToken:`. This means that the `ongoingGetTokenOperation` is returned from `getToken:` in a resolved state which would be bad for future called of `getToken:`.
3. Enabling app attest support and then calling `getTokenWithCompletion` results in nothing because the  `ongoingGetTokenOperation` has persisted.

After taking a closer took, I think we are safe from this because  `attestationState:` executes on another thread so it should prevent the `ongoingGetTokenOperation` from resolving before being returned from `getToken`

ncooke3 (approved):

FirebaseAppCheck/Sources/AppAttestProvider/FIRAppAttestProvider.m
maksymmalyhin: Ah, I see what you mean. Yeah, currently `then` or `recover` blocks will always be executed after the promise is returned. Thank you for the discussion!

PR #8123, opened 2021-05-20
WIP: Make Message passing more explicit

schmidt-sebastian:
This is just a proof of concept, but it might work. Assigning for early comment.


Firestore/core/src/nanopb/message.h
```diff
@@ -187,6 +187,49 @@ class Message {
...
+  OptionalMessage(const OptionalMessage&) = delete;
+  OptionalMessage& operator=(const OptionalMessage&) = delete;
+
+  OptionalMessage(OptionalMessage&& other) noexcept
+      : message_{other.message_ ? std::move(*other.message_) : absl::nullopt} {
+    other.owns_proto_ = false;
+  }
+
+  OptionalMessage& operator=(OptionalMessage&& other) noexcept {
+    if (message_) {
```
var-const, line 207: Why is it impossible to overwrite an empty optional message? Seems like strange semantics.

var-const (commented):
It might be a good idea, but I have a few questions:
* How is using `OptionalMessage` materially better than `optional<Message>`? Does the type being move-only give tangible benefits?
* The use of this class seems limited to the few places where optional values are passed. I don't see how it solves the problem of unclear ownership at large. Do you plan more work on this?
*

Issue #7707, opened 2020-10-09
Open counts from push notifications are not counted (IOS)


chris-o3h:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2019.4.5.f1
  * Firebase Unity SDK version: 6.15.2
  * Source you installed the SDK: Package Manager
  * Problematic Firebase Component: Cloud Messaging
  * Other Firebase Components in use: Auth, Database, Dynamic Linking, Crashalytics, Analytics
  * Additional SDKs you are using: Too many to list, no add mob or any other analytics
  * Platform you are using the Unity editor on: Mac
  * Platform you are targeting: iOS
  * Scripting Runtime: IL2CPP

### [REQUIRED] Please describe the issue here:

Open counts for push notifications are usually zero or low. Even when explicitly testing opening a push not 
We are currently in an alpha stage with our app and we cannot determine why the open rates are so low. Even in our dev environment the open rates are a total of 48 for the last 90 days even though we've opened many more push notes than this and sent over 12,000 push notifications.

We are unsure of how to make the "open count" work. We have tried adding `FIREBASE_ANALYTICS_COLLECTION_ENABLED` to the plist as well as setting the `notification_open` event to a conversion event.

Frankly I'm not sure if this is a bug or simply a detail that is not well documented.

#### Steps to reproduce:

Simply run your app, receive a push note and open the app. The open will not count on the console's open count. And if it does it is not consistent.

khuffie:
@chliangGoogle  - the google-oss-bot is stating this issue is stale.

Is this still being investigated on your end?

We are seeing the same issue in our iOS app. On Android, our push notification rate is about 27%, on iOS, it's less than 1%. We do see an uptick in users on iOS whenever we send a push notifications, so they ARE being received and opened.

paulb777:
Looks like this was marked stale because of the open question above:

@chris-o3h were you able to see notification_open even logged in analytics debug view?

chliangGoogle:
@chris-o3h Can you also let us know if you are using "send now" or "scheduled" message? If you use "scheduled" message, are you using local timezone or specific timezone?

Here's how you can use analytics debug view to see if notificaiton_open is logged: https://firebase.google.com/docs/analytics/debugview

khuffie:
@chliangGoogle  - I can share my experience as well

We have used both "send now" and "scheduled" messages. Both messages were in specific timezones.

[...]

PR #8129, opened 2021-05-21
Fix an issue that notification open is not logged when app is closed

### LINKED ISSUE #8128: open_notification event not being triggered when app is not in background.

  * Xcode version: 12.5
  * Firebase SDK version: [8.0.0](url)
  * Installation method: SPM
  * Firebase Component: Messaging

#### Steps to reproduce:

1. Force quit the app
2. Send a push notification to the app
3. Tap on the push notification
4. Observe Debug View

Expected: notification_open event gets triggered
Actual: No notification_event gets triggered

We are noticing that iOS is severely undercounting the push notification open rate. On Android, our open rate is around 12%. On iOS, our open rate is around 0.29%. However, evidence suggests that users ARE receiving push notifications and opening them on iOS (we see similar surges in traffic as Android), but just that the notification_open event is not getting triggered.

We have confirmed that swizzling is ON.

We have tested the following scenarios, each 5 times, and each reproduced 100%

- App is in the background, notification is received, notification_open is triggered
- App is closed, notification is received, notification_open is NOT triggered
- App is open in foreground, notification is received, notification_foreground is triggered correctly

We are monitoring the events via the DebugView.

Note that we recently updated to v8.0.0. However, we saw the same issue in the previous version of the SDK which was manually integrated.

chliangGoogle:
This fixes #8128 and #7707. Internal bug: b/157964197.

When app is completely closed and user taps on notification to open the app, the client SDK does not correctly capture the state to log notification open.

Also add analytics to sample app.

Added unit test and tested sending when app is closed, now notification_open is logged on analytics debug view console.

### DIFF
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,3 +1,6 @@
+# unreleased
+- [fixed] Fixed an issue that notification open is not logged correctly when app is completely shut off. (#7707, #8128).
+
 # 2021-04 -- v8.0.0
 - [changed] Remove the Instance ID dependency from Firebase Cloud Messaging. This is a breaking change for FCM users who use the deprecated Instance ID API to manage registration tokens. Users should migrate to FCM's token APIs by following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
 
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,4 +1,4 @@
-# unreleased
+# 2021-05 -- v8.1.0
 - [fixed] Fixed an issue that notification open is not logged to Analytics correctly when app is completely shut off. (#7707, #8128).
 
 # 2021-04 -- v8.0.0
--- a/FirebaseMessaging/CHANGELOG.md
+++ b/FirebaseMessaging/CHANGELOG.md
@@ -1,5 +1,5 @@
 # unreleased
-- [fixed] Fixed an issue that notification open is not logged correctly when app is completely shut off. (#7707, #8128).
+- [fixed] Fixed an issue that notification open is not logged to Analytics correctly when app is completely shut off. (#7707, #8128).
 
 # 2021-04 -- v8.0.0
 - [changed] Remove the Instance ID dependency from Firebase Cloud Messaging. This is a breaking change for FCM users who use the deprecated Instance ID API to manage registration tokens. Users should migrate to FCM's token APIs by following the migration guide: https://firebase.google.com/docs/projects/manage-installations#fid-iid. (#7836)
--- a/FirebaseMessaging/Sources/FIRMessagingAnalytics.m
+++ b/FirebaseMessaging/Sources/FIRMessagingAnalytics.m
@@ -189,6 +189,7 @@ static NSString *const kAnalyticsTrackConversions = @"google.c.a."
 + (void)logMessage:(NSDictionary *)notification
        toAnalytics:(id<FIRAnalyticsInterop> _Nullable)analytics {
   // iOS only because Analytics doesn't support other platforms.
+
 #if TARGET_OS_IOS
   if (![self canLogNotification:notification]) {
     return;
@@ -201,9 +202,12 @@ static NSString *const kAnalyticsTrackConversions = @"google.c.a."
   UIApplicationState applicationState = application.applicationState;
   switch (applicationState) {
     case UIApplicationStateInactive:
-      // App was either in background(suspended) or inactive and user tapped on a display
-      // notification.
-      [self logOpenNotification:notification toAnalytics:analytics];
+      // App was in background and in transition to open when user tapped
+      // on a display notification.
+      // Needs to check notification is displayed.
+      if ([[self class] isDisplayNotification:notification]) {
+        [self logOpenNotification:notification toAnalytics:analytics];
+      }
       break;
 
     case UIApplicationStateActive:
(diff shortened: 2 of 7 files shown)


FirebaseMessaging/Sources/FIRMessagingAnalytics.m
```diff
@@ -212,11 +216,34 @@ + (void)logMessage:(NSDictionary *)notification
...
++ (BOOL)isDisplayNotification:(NSDictionary *)notification {
+  NSDictionary *aps = notification[kApsKey];
+  if (!aps || ![aps isKindOfClass:[NSDictionary class]]) {
+    return NO;
+  }
+  NSDictionary *alert = aps[kApsAlertKey];
+  if (!alert) {
+    return NO;
+  }
+  if ([alert isKindOfClass:[NSDictionary class]]) {
```
paulb777, line 239: How is it determined that the next two cases are the only ones that we want to log analytics for?

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
```diff
@@ -201,9 +202,12 @@ + (void)logMessage:(NSDictionary *)notification
...
   switch (applicationState) {
     case UIApplicationStateInactive:
-      // App was either in background(suspended) or inactive and user tapped on a display
-      // notification.
-      [self logOpenNotification:notification toAnalytics:analytics];
+      // App was in background and in transition to open when user tapped
+      // on a display notification.
+      // Needs to check notification is displayed.
+      if ([[self class] isDisplayNotification:notification]) {
+        [self logOpenNotification:notification toAnalytics:analytics];
```
paulb777, line 209: From #8128, it sounds like we're not logging enough notifications to analytics.  Won't adding the conditional make the log count even less?

FirebaseMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [fixed] Fixed an issue that notification open is not logged correctly when app is completely shut off. (#7707, #8128).
```
paulb777, line 2: "not logged to Analytics"

paulb777 (commented):
Thanks!

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
```diff
@@ -212,11 +216,34 @@ + (void)logMessage:(NSDictionary *)notification
...
+      }
       break;
   }
 #endif
 }
 
++ (BOOL)isDisplayNotification:(NSDictionary *)notification {
```
maksymmalyhin, line 230: Would you mind adding a comment to describe the expected dictionary format e.g. a link to docs or with a format like `@{ kApsKey : @{ kApsAlertKey : @"alert" } }`. I think it will be useful for future us.

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
```diff
@@ -212,11 +216,34 @@ + (void)logMessage:(NSDictionary *)notification
       break;
 
     default:
-      // Only a silent notification (i.e. 'content-available' is true) can be received while the app
-      // is in the background. These messages aren't loggable anyway.
+      // App was either in background state or in transition from closed
```
maksymmalyhin, line 219: Would it be possible to add tests for the new logic to guard against regressions in the future?

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
```diff
@@ -212,11 +216,34 @@ + (void)logMessage:(NSDictionary *)notification
...
+      // to open.
+      // Needs to check notification is displayed.
+      if ([[self class] isDisplayNotification:notification]) {
+        [self logOpenNotification:notification toAnalytics:analytics];
+      }
```
maksymmalyhin, line 224: Would it be possible to add tests for the new logic to guard against regressions in the future?

maksymmalyhin (approved):
LGTM, thank you for the fix! 

I'll be happy to help with additional tests if needed.

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
chliangGoogle: We want to differentiate display notifications from hidden ones, so the field "alert" must exist and have at least one value in it.

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
chliangGoogle: The reason that we are not logging enough notification open is we are missing the default condition below:
When app is open from shutdown state, this condition is no longer true but the default condition below, which is the background state. (btw, this is weird cuz we haven't changed the code for a long time)


Because of that, only the notifications open when app is in background is logged, not when it's shutdown state. The Application State names here are confusing but the condition did get hit to default state when app is open from shutdown state.

FirebaseMessaging/CHANGELOG.md
chliangGoogle: Good call.

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
chliangGoogle: I've added unit tests for isDisplayNotifications. 
Btw we have not changed this code for a long time, so I don't know if UIApplicationState has changed because I remember we used to be able to log with UIApplicationStateInactive state. We can mock the UIApplicationState, but for this scenario, we can't prevent regressions if system has some weird shift between these UIApplicationStates, but yeah feel free to add more tests if you see fit!

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
paulb777: Thanks for the explanation!

paulb777 (approved):

FirebaseMessaging/Sources/FIRMessagingAnalytics.m
chliangGoogle: I will add in another PR to ensure we release in time.

PR #8132, opened 2021-05-21
Fix zip qs test build regression

paulb777:
The simplified Carthage directory structure introduced in #8093 caused the `find` command used to extract the zip archive to fail since it now also matched all the Carthage zips.

- Fix the `find` issue
- Always run the quickstart tests on zip builder PRs to catch subtle problems like this on presubmit. Also, the qs tests are the only tests we have of using the zip builder for non-Firebase release builds.

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -3,7 +3,7 @@ name: zip
 on:
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):

paulb777:
cc: @granluo

PR #8143, opened 2021-05-24
Update upload-symbols tool logs

elenadoty:
#no-changelog


adamvduke:
Is the upload-symbols binary intended to be part of this change, and if so, should there be an accompanying version change?

elenadoty:
The change is just a few words in the message that prints once validation is done so I didn't bump the version, but I can if you think it should be.

PR #8146, opened 2021-05-25
Update merge branch in the Metrics Service

granluo:
GITHUB_REF will get a 'merge' branch due to pull_request , instead of a `master` branch.  
`github.head_ref` could fetch the merged branch, e.g. `master`. 
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -292,5 +292,5 @@ jobs:
         run: |
           if [ -d "${{steps.download.outputs.download-path}}" ]; then
           cd scripts/code_coverage_report/generate_code_coverage_report
-          swift run CoverageReportGenerator --merge "firebase/firebase-ios-sdk" --commit "${GITHUB_SHA}" --token $(gcloud auth print-identity-token) --xcresult-dir "/Users/runner/test/codecoverage" --log-link "https://github.com/firebase/firebase-ios-sdk/actions/runs/${GITHUB_RUN_ID}" --branch "${GITHUB_REF##*/}"
+          swift run CoverageReportGenerator --merge "firebase/firebase-ios-sdk" --commit "${GITHUB_SHA}" --token $(gcloud auth print-identity-token) --xcresult-dir "/Users/runner/test/codecoverage" --log-link "https://github.com/firebase/firebase-ios-sdk/actions/runs/${GITHUB_RUN_ID}" --branch "${{ github.head_ref }}"
           fi


.github/workflows/test_coverage.yml
```diff
@@ -292,5 +292,5 @@ jobs:
         run: |
           if [ -d "${{steps.download.outputs.download-path}}" ]; then
           cd scripts/code_coverage_report/generate_code_coverage_report
-          swift run CoverageReportGenerator --merge "firebase/firebase-ios-sdk" --commit "${GITHUB_SHA}" --token $(gcloud auth print-identity-token) --xcresult-dir "/Users/runner/test/codecoverage" --log-link "https://github.com/firebase/firebase-ios-sdk/actions/runs/${GITHUB_RUN_ID}" --branch "${GITHUB_REF##*/}"
+          swift run CoverageReportGenerator --merge "firebase/firebase-ios-sdk" --commit "${GITHUB_SHA}" --token $(gcloud auth print-identity-token) --xcresult-dir "/Users/runner/test/codecoverage" --log-link "https://github.com/firebase/firebase-ios-sdk/actions/runs/${GITHUB_RUN_ID}" --branch "${{ github.head_ref }}"
```
maksymmalyhin, line 295: Just would like to check my understanding: 
- how `commit` and `branch` parameters are related to each other? Do we need to update `commit` as well?
- does it mean that we may have different actual code coverage after merge if we use the top of the source branch instead of the merge?

maksymmalyhin (approved):
LGTM, but a couple questions mostly for me to understand how it is supposed to work.

.github/workflows/test_coverage.yml
granluo: 1. `commit` and `branch` are separated in the API. The reason for the update is that when a pull request is merged, the SHA of the pull_request is the resulting commit that was created from merging the base to the head and that branch is `refs/pull/${prNumber}/merge`. This causes the branch to be `merge` instead of the base branch, e.g. `master`.
2. The code coverage will be compared to the based commit where the branch rooted. This change is related to the update of the database of the Metrics Service. The outcome of this commit will not be displayed, but just stored to database. So next time a new PR rooting from this commit will compare the head of the branch to this base commit.

maksymmalyhin: Looks reasonable, thank you for explanation. I would suggest to consider a bit more elaborative parameter names to reduce the confusion in the future, e.g. `--source-branch`, `--result-commit` or something along these lines.

PR #8156, opened 2021-05-26
Fix code coverage branch

granluo:
According to [`github` context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), `github.head_ref` refers to the head branch which is what we work on. `github.base_ref` is the targeted branch the PR is going to be merged into.


granluo:
This is the update/fix of #8146

maksymmalyhin (approved):
LGTM, but as I mentioned in #8146 I would consider a bit more elaborative parameter names.

PR #8157, opened 2021-05-26
Applied shell wrapper to the current code coverage.

granluo:
This shell wrapper will be applied to the incremental code coverage as well.

### DIFF
--- a/scripts/code_coverage_report/generate_code_coverage_report/Sources/Utils/ShellUtils.swift
+++ b/scripts/code_coverage_report/generate_code_coverage_report/Sources/Utils/ShellUtils.swift
@@ -16,9 +16,7 @@
 
 import Foundation
 
-public struct Shell {
-  static let shared = Shell()
-  public init() {}
+public enum Shell {
   @discardableResult
   public static func run(_ command: String, displayCommand: Bool = true,
                          displayFailureResult: Bool = true) -> Int32 {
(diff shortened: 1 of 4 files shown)


scripts/code_coverage_report/generate_code_coverage_report/Sources/Utils/ShellUtils.swift
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+public struct Shell {
+  static let shared = Shell()
+  public init() {}
```
ryanwilson, line 21: Nit: the `shared` shell and `init` don't really do anything here, and since we're using a static func it's probably better to use a caseless `enum` and treat it as a namespace:

```
public enum Shell {
```

That way we can remove the `static let shared = Shell()` and `public init()` since there's nothing to initialize there.

Happy to chat more about it if you'd like!

ryanwilson (commented):
LGTM after the one change!

scripts/code_coverage_report/generate_code_coverage_report/Sources/Utils/ShellUtils.swift
granluo: Yep it is more reasonable to set it as a caseless `enum`, have updated it.

ryanwilson (approved):

PR #8170, opened 2021-05-31
App Check App Attest: handle attestation rejection

maksymmalyhin:
- reset stored attestation data on attestation rejection from Firebase backend
- outdated TODOs removed
- `kFIRLoggerAppCheckMessageCodeUnknown` replaced by specific unique codes in log messages

### DIFF
--- a/FirebaseAppCheck/Sources/AppAttestProvider/Errors/FIRAppAttestRejectionError.h
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/Errors/FIRAppAttestRejectionError.h
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FIRAppAttestRejectionError : NSError
+
+- (instancetype)init;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
+++ b/FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
@@ -22,19 +22,16 @@
 
 - (instancetype)initWithHTTPResponse:(NSHTTPURLResponse *)HTTPResponse
                                 data:(nullable NSData *)data {
-  NSDictionary *userInfo = [[self class] userInfoWithHTTPResponse:HTTPResponse
-                                                                          data:data];
-  self = [super
-      initWithDomain:kFIRAppCheckErrorDomain
-                code:FIRAppCheckErrorCodeUnknown
-            userInfo:userInfo];
+  NSDictionary *userInfo = [[self class] userInfoWithHTTPResponse:HTTPResponse data:data];
+  self = [super initWithDomain:kFIRAppCheckErrorDomain
+                          code:FIRAppCheckErrorCodeUnknown
+                      userInfo:userInfo];
   if (self) {
     _HTTPResponse = HTTPResponse;
     _data = data;
   }
   return self;
 }
-
 
 + (NSDictionary *)userInfoWithHTTPResponse:(NSHTTPURLResponse *)HTTPResponse
                                       data:(nullable NSData *)data {
--- a/FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
+++ b/FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2021 Google LLC
+ *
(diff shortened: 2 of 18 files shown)


FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
```diff
@@ -0,0 +1,74 @@
...
+
+- (instancetype)initWithHTTPResponse:(NSHTTPURLResponse *)HTTPResponse
+                                data:(nullable NSData *)data {
+  NSDictionary *userInfo = [[self class] userInfoWithHTTPResponse:HTTPResponse data:data];
+  self = [super initWithDomain:kFIRAppCheckErrorDomain
+                          code:FIRAppCheckErrorCodeUnknown
+                      userInfo:userInfo];
+  if (self) {
+    _HTTPResponse = HTTPResponse;
+    _data = data;
```
ryanwilson, line 31: Should we be  using the `userInfo` property of the error and providing a key for `data` instead of having a subclass here? Or is there something I'm missing for why this should be its own class?

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m
```diff
@@ -797,6 +923,48 @@ - (void)assertGetToken_WhenKeyRegistered_Success {
   [self verifyAllMocks];
 }
 
+- (void)expectAppAttestAvailabilityToBeCheckedAndStoredKeyRequested {
```
ryanwilson, line 926: It wasn't immediately obvious to me that this would be setting up a rejection as part of the expectation - is there a way we can encode that in the (already long) name?

ryanwilson (commented):
A couple small questions but otherwise LGTM, thanks!

FirebaseAppCheck/Sources/AppAttestProvider/Errors/FIRAppAttestRejectionError.h
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FIRAppAttestRejectionError : NSError
+
+- (instancetype)init;
```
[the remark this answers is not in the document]
maksymmalyhin: The App Attest provider is already pretty big (close to be bigger than the rest of App Check SDK), so it seems it will make sense to separate it in the future. This error is specific for App Attest, so having it independent from Core will make separation easier in the future. Also FIRAppCheckErrorUtil was intended to be used mostly for errors without specific classes.

FirebaseAppCheck/Sources/Core/Errors/FIRAppCheckHTTPError.m
maksymmalyhin: Working with specific NSError subclasses is so much easier in the client code because we need to check only the class to identify the type of the error and the data associated with it. In contrast we would need to check the error domain, the keys and the value types in the user info dict if we use NSError. Does it look reasonable to you?

ryanwilson: I think that sounds reasonable, thanks!

FirebaseAppCheck/Sources/AppAttestProvider/Errors/FIRAppAttestRejectionError.h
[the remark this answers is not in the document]
ryanwilson: SGTM, thanks

ryanwilson (approved):

FirebaseAppCheck/Sources/Core/FIRAppCheckLogger.m
```diff
@@ -16,6 +16,20 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
 FIRLoggerService kFIRLoggerAppCheck = @"[Firebase/AppCheck]";
 
 NSString *const kFIRLoggerAppCheckMessageCodeUnknown = @"I-FAA001001";
+
+void FIRAppCheckDebugLog(NSString *message, ...) {
+  va_list args_ptr;
+  va_start(args_ptr, message);
+  FIRLogBasic(FIRLoggerLevelDebug, kFIRLoggerAppCheck, kFIRLoggerAppCheckMessageCodeUnknown,
```
ryanwilson, line 30: Nit: If this is landing any time soon, I'd prefer that we pass in the code number inside the function and then pass in `Unknown` everywhere so it's easier to refactor at a later date.  Just want to make sure we don't accidentally ship multiple logs with the same Unknown error code.

ryanwilson (approved):

FirebaseAppCheck/Sources/Core/FIRAppCheckLogger.m
maksymmalyhin: Actually, I intentionally left the default message code here. I don't remember when I actually used the code last time, so it looks like it's has very limited usefulness. I would prefer to keep track of the codes only for logs with higher level and use the default one for debug. WDYT?

ryanwilson: The codes are more for external usage - if  a developer runs into a log statement they can then just search for that code to either find the code itself or other posts referencing that error message, for example:

https://www.google.com/search?q=%22I-COR000003%22

That way if we need to wordsmith the error message at all we can do so and not lose any historical context.

maksymmalyhin: Ah, I didn't think of this particular use case. I'll add codes. Though it still looks much more relevant for non-debug logs, it may be useful for debug as well.

maksymmalyhin: updated codes here and in other log messages.

ryanwilson (approved):
Thanks for updating all the logs!

PR #8171, opened 2021-05-31
Migrate to individual platform flags.

ryanwilson:
This makes it easier to dynamically include platforms in a bash script
since variable expansion with spaces can be a bit dangerous.

#no-changelog


ryanwilson:
This may not be needed - WIP

ryanwilson:
Not necessary any more - bash arrays work instead.

PR #8174, opened 2021-06-01
Set valid format of notification sample

chliangGoogle:
In rare case the test is running when app is not in foreground, falling into the default application state where it checks if notification is a display one. So set the sample notification with valid display format to ensure tests are always passing.
#no-changelog


maksymmalyhin (approved):
LGTM on CI green. Thank you for fixing it!

Issue #8109, opened 2021-05-19
Uploading image to Storage Emulator is uploading file to Firebase Remote Project instead


MarcelBreska:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.5
  * Firebase SDK version: 8.0.0
  * Installation method: `Swift Package Manager` (select one)
  * Firebase Component:  Storage (Auth, Core, Database, Firestore, Messaging, Storage, etc)

### [REQUIRED] Step 2: Describe the problem

I'm trying to upload an image from iOS with the FirebaseSDK 8.0.0 to my Firebase Storage emulator, but the app is instead uploading the file to the Firebase Project Storage. The Authentication and Firestore setup with the emulator ist working without problems, only the Storage is not redirecting to the emulator.

#### Steps to reproduce:

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

If you have a downloadable sample project that reproduces the bug you're reporting, you will
likely receive a faster response on your issue.

#### Relevant Code:

```
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
    
    FirebaseApp.configure()
    
    Auth.auth().useEmulator(withHost: "localhost", port: 9099)
    Storage.storage().useEmulator(withHost: "localhost", port: 9199)
    
    let settings = Firestore.firestore().settings
    settings.host = "localhost:8080"
    settings.isPersistenceEnabled = false
    settings.isSSLEnabled = false
    Firestore.firestore().settings = settings
    
    return true
}

func uploadProfilePicture(imageData: Data, imageId: String) -> AnyPublisher<String, Error> {
    return Future<StorageReference, Error> { promise in
        let metaDataConfig = StorageMetadata()
        metaDataConfig.contentType = "image/jpg"
        
        let storageRef = Storage.storage().reference(withPath: "profiles/\(imageId)")
        
        storageRef.putData(imageData, metadata: metaDataConfig) { (_, error) in
            if let error = error {
                promise(.failure(error))
            } else {
                promise(.success(storageRef))
            }
        }
    }
    .compactMap { [weak self] storageRef in
        self?.donwloadUrl(storageRef: storageRef)
    }
    .switchToLatest()
    .eraseToAnyPublisher()
}
```

morganchen12:
Hey Marcel, is this reproducible in Firebase 7.11?

[...]

PR #8178, opened 2021-06-02
Cache FirebaseStorage instances

schmidt-sebastian:
This re-uses the same logic that RTDB uses to ensure that repeated invocations of the Storage factory function return the same instance for the same bucket.

Fixes https://github.com/firebase/firebase-ios-sdk/issues/8109


morganchen12 (approved):
Sorry for the slow review, I missed this in my inbox.

PR #8181, opened 2021-06-02
Fix warnings generated by Clang-Tidy

ehsannas:
This is the first set of updates. There will be another PR with more fixes coming up later.


Firestore/Source/API/FIRQuery.mm
```diff
@@ -55,6 +55,7 @@
 #include "Firestore/core/src/util/statusor.h"
 #include "Firestore/core/src/util/string_apple.h"
 #include "absl/memory/memory.h"
+#include "absl/strings/str_split.h"
```
var-const, line 58: I think the correct include is `absl/strings/match.h`.

Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
```diff
@@ -24,7 +24,6 @@
 namespace firebase {
 namespace firestore {
 
-using nanopb::PrintEnumField;
```
var-const, line 27: Please revert changes to the `nanopb.cc` files. These are generated files, so the change will get overwritten. The [script in question](https://github.com/firebase/firebase-ios-sdk/blob/acf106a33d50e893f85a8563a19557b3df6110d1/Firestore/Protos/nanopb_cpp_generator.py#L419-L427) simply adds all the using declarations that might be necessary. I think adding extra logic to figure out the minimal set of the necessary using declarations is overkill; simply ignoring warnings in this case (especially considering that this is generated code) seems preferable.

var-const (changes requested):

Firestore/Source/API/FIRQuery.mm
ehsannas: Done

Firestore/Protos/nanopb/firestore/local/maybe_document.nanopb.cc
ehsannas: Ah! Good to know. Done

Firestore/core/src/core/sync_engine.cc
```diff
@@ -168,7 +168,8 @@ void SyncEngine::StopListening(const Query& query) {
 
   TargetId target_id = query_view->target_id();
   auto& queries = queries_by_target_[target_id];
-  queries.erase(std::remove(queries.begin(), queries.end(), query));
+  queries.erase(std::remove(queries.begin(), queries.end(), query),
+                queries.end());
```
ehsannas, line 172: This is the reference for this fix: https://clang.llvm.org/extra/clang-tidy/checks/bugprone-inaccurate-erase.html

ehsannas:
I've gone through all the clang-tidy warnings. This PR should address all of them.

Firestore/core/src/core/sync_engine.cc
var-const: I presume there would normally (perhaps always) only be one query to remove, so this code would work equivalently to the correct code, but this is a very good find regardless.

Firestore/core/src/util/status_apple.mm
```diff
@@ -96,16 +96,17 @@ Status FromFirestoreNSError(NSError* error) {
...
           .WithPlatformError(std::move(original));
     }
 
     error = error.userInfo[NSUnderlyingErrorKey];
   }
 
-  return Status{Error::kErrorUnknown,
-                StringFormat("Unknown error: %s", original->error())}
+  auto description = StringFormat("Unknown error: %s", original->error());
+  return Status{Error::kErrorUnknown, description}
```
var-const: Nit: you can move `description`.

var-const (approved):
Looks good!

Issue #8154, opened 2021-05-26
GoogleSignIn crash


Montura:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: `12.4 (12D4e)`
  * Firebase SDK version: `6.34.0 (the same behavior on 8.0.0)`
  * Installation method: `Zip file`
  * Firebase Component: `Auth`

### [REQUIRED] Step 2: Describe the problem

iPad Air (3rd gen) running iOS 14.0.1

1. App works fine when sign in with _basic_ google account.
2. App crashes when sign in with _google workspace_ account.

`Assertion failure in +[UIAlertAction _actionWithTitle:descriptiveText:image:style:handler:shouldDismissHandler:], UIAlertAction.m:42`
`2021-05-25 17:49:53.950251+0300 GoogleSignIn[16264:9133478] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Actions added to UIAlertController of style UIAlertActionStyleCancel must have a title when running in UIUserInterfaceIdiomPad'
`

Full backtrace is here (https://github.com/Montura/GoogleSignInDemo/blob/main/lldb_bt_all_log.txt)

P.S. It works fine on iOS-simulator

#### Steps to reproduce:

Login with credentials. There is a simple demo with Firebase 6.34.0 (https://github.com/Montura/GoogleSignInDemo)

The same problem was discussed here (https://github.com/firebase/firebase-ios-sdk/issues/6406)

#### Relevant Code:

```
// AppDelegate.m
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    [FIRConfiguration.sharedInstance setLoggerLevel:FIRLoggerLevelWarning];
    [FIRApp configure];
    
    [FIRAnalytics setAnalyticsCollectionEnabled:true];

    GIDSignIn * gidSignIn = [GIDSignIn sharedInstance];
    gidSignIn.clientID = [FIRApp defaultApp].options.clientID;
    gidSignIn.delegate = self;
    
    // Override point for customization after application launch.
    return YES;
}

// ViewController.m
- (void)buttonPressed:(UIButton *)button {
    [GIDSignIn sharedInstance].presentingViewController = self;
    [[GIDSignIn sharedInstance] signIn];
}
```

### [OPTIONAL] Step 3: The expected workflow
   1. To get an alert message when sign in with **google workspace** account
[...]

PR #8182, opened 2021-06-03
Fix Resource generation for Carthage xcframeworks

### LINKED ISSUE #8112: Firestore Carthage installation is missing gRPC-C++ in 8.0.0
Opened to track issue raised at https://github.com/firebase/firebase-ios-sdk/issues/7576#issuecomment-843774872

paulb777:
Fix #8112 

Carthage resources should remain in each `.framework` instead of at the root of `.xcframework`s.

After all xcframework resources are processed and stored at the root of the .xcframework at https://github.com/firebase/firebase-ios-sdk/blob/master/ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift#L697, the new code consolidates both Zip Distro and Carthage resource processing in ZipBuilder.swift. 

It still remains for the future to handle platform-specific Resources.

Successful test at https://github.com/firebase/firebase-ios-sdk/actions/runs/903575515.
I verified the Zip in the artifact matches the 8.1.0 release. The only exception is GoogleSignIn Resources are now in the framework.  See comments for more discussion about GoogleSignIn.  I also verified that the Firestore, GoogleSignIn, and FIAM Carthage zips successfully install on `carthage update` including the xcframeworks that were not installing because of the Resources location.

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/FirebaseBuilder.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/FirebaseBuilder.swift
@@ -46,24 +46,6 @@ struct FirebaseBuilder {
         options: carthageBuildOptions
       )
 
-      // Prepare the release directory for zip packaging.
-      do {
-        // Move the Resources out of each directory in order to maintain the existing Zip structure.
-        let fileManager = FileManager.default
-        let contents = try fileManager.contentsOfDirectory(atPath: location.path)
-        for fileOrFolder in contents {
-          let fullPath = location.appendingPathComponent(fileOrFolder)
-
-          // Ignore any files.
-          guard fileManager.isDirectory(at: fullPath) else { continue }
-
-          // Move all the bundles in the frameworks out to a common "Resources" directory to match the
-          // existing Zip structure.
-          let resourcesDir = fullPath.appendingPathComponent("Resources")
-          _ = try ResourcesManager.moveAllBundles(inDirectory: fullPath, to: resourcesDir)
-        }
-      }
-
       print("Attempting to Zip the directory...")
       let candidateName = "Firebase-\(firebaseVersion)-latest.zip"
       let zipped = Zip.zipContents(ofDir: location, name: candidateName)
--- a/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
@@ -389,15 +389,16 @@ struct ZipBuilder {
                                      frameworksToAssemble: [String: [URL]],
                                      firebasePod: CocoaPodUtils.PodInfo) throws -> URL {
     // Create the directory that will hold all the contents of the Zip file.
-    let zipDir = FileManager.default.temporaryDirectory(withName: packageKind)
+    let fileManager = FileManager.default
+    let zipDir = fileManager.temporaryDirectory(withName: packageKind)
     do {
-      if FileManager.default.directoryExists(at: zipDir) {
-        try FileManager.default.removeItem(at: zipDir)
+      if fileManager.directoryExists(at: zipDir) {
+        try fileManager.removeItem(at: zipDir)
       }
 
-      try FileManager.default.createDirectory(at: zipDir,
-                                              withIntermediateDirectories: true,
-                                              attributes: nil)
+      try fileManager.createDirectory(at: zipDir,
+                                      withIntermediateDirectories: true,
+                                      attributes: nil)
     }
 
     // Copy all required files from the Firebase pod. This will cause a fatalError if anything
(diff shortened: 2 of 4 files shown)


ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -452,28 +453,76 @@ struct ZipBuilder {
                                        rootZipDir: zipDir,
                                        builtFrameworks: frameworksToAssemble,
                                        podsToIgnore: analyticsPods)
-
         // Update the README.
         readmeDeps += dependencyString(for: pod.key, in: productDir, frameworks: podFrameworks)
-
-        // Special case for Crashlytics:
```
paulb777, line 459: The Crashlytics code is only reindented and shows up in green below.

ReleaseTooling/Sources/ZipBuilder/FirebaseBuilder.swift
```diff
@@ -46,24 +46,6 @@ struct FirebaseBuilder {
         options: carthageBuildOptions
       )
 
-      // Prepare the release directory for zip packaging.
```
paulb777, line 49: This chunk is migrated to the `if packageKind == "Firebase" {` code below.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -452,28 +453,76 @@ struct ZipBuilder {
...
+          guard fileManager.isDirectory(at: xcPath),
+            xcPath.lastPathComponent.hasSuffix("xcframework"),
+            fileManager.directoryExists(at: xcResourceDir)
+          else { continue }
+
+          if packageKind == "Firebase" {
+            // Move all the bundles in the frameworks out to a common "Resources" directory to
+            // match the existing Zip structure.
+            let resourcesDir = productPath.appendingPathComponent("Resources")
+            try fileManager.moveItem(at: xcResourceDir, to: resourcesDir)
```
paulb777, line 483: moveItem instead of `ResourcesManager.moveAllBundles` since the resources were already processed when the xcframework was created.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -452,28 +453,76 @@ struct ZipBuilder {
...
-              fatalError("Error copying Crashlytics tools from \(source) to \(target): \(error)")
+      } catch {
+        fatalError("Could not copy frameworks from \(pod) into the zip file: \(error)")
+      }
+      do {
+        // Update Resources: For the zip distribution, they get pulled from the xcframework to the
+        // top-level product directory. For the Carthage distribution, they propagate to each
+        // individual framework.
+        // TODO: Investigate changing the zip distro to also have Resources in the .frameworks to
+        // enable different platform Resources.
```
ryanwilson, line 466: I thought there was an issue in having them embedded in the frameworks but I can't remember, maybe it was an older Xcode version and latest versions would be fine.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -452,28 +453,76 @@ struct ZipBuilder {
...
+            xcPath.lastPathComponent.hasSuffix("xcframework"),
+            fileManager.directoryExists(at: xcResourceDir)
+          else { continue }
+
+          if packageKind == "Firebase" {
+            // Move all the bundles in the frameworks out to a common "Resources" directory to
+            // match the existing Zip structure.
+            let resourcesDir = productPath.appendingPathComponent("Resources")
+            try fileManager.moveItem(at: xcResourceDir, to: resourcesDir)
+
```
ryanwilson, line 484: Nit: extra blank line.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
```diff
@@ -452,28 +453,76 @@ struct ZipBuilder {
...
+          } else {
+            let xcContents = try fileManager.contentsOfDirectory(atPath: xcPath.path)
+            for fileOrFolder in xcContents {
+              let platformPath = xcPath.appendingPathComponent(fileOrFolder)
+              guard fileManager.isDirectory(at: platformPath)
+              else { continue }
```
ryanwilson, line 490: This could probably be on a single line

ryanwilson (approved):
Thanks for the comments, they helped a bunch!

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
paulb777: We might have had another issue in the past, like an invalid Info.plist. My testing shows it to be fine now.

ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
paulb777: Addressing in next commit

paulb777 (commented):
Tests look good. Will fix spaces review commit and summarize testing in PR intro, and then merge

paulb777:
There is an issue with blaze created frameworks 😞
```
~/test-8082/Firebase-actions-dir/carthage/8.1.0/GoogleSignIn  $ swift
Welcome to Apple Swift version 5.4 (swiftlang-1205.0.26.9 clang-1205.0.19.55).
Type :help for assistance.
  1> import Foundation
  2> Bundle(path: "GoogleSignIn.framework")!.infoDictionary
$R0: [String : Any]? = 0 key/value pairs
```
The one diff in the generated zip is the GoogleSignIn resources aren't being moved out of the framework any longer.  I'll restore so that the zip is unchanged.

paulb777:
On additional investigation, the Carthage GoogleSignIn.framework is the same structure it has always been. It turns out that Carthage needs to be able to process Info.plist's for `.xcframework`s , but not for `.framework`s.  Given that GoogleSignIn will soon move to `.xcframework`s and issues like #8154, I'll merge as is.

PR #8184, opened 2021-06-03
Fix AppCheck Catalyst build on Xcode 12.4

### LINKED ISSUE #8175: Nightly Testing Report
# Nightly Testing Report
This issue([log](https://github.com/firebase/firebase-ios-sdk/actions/runs/899519800)) is generated at 06/02/2021 06:29 AM, fetching workflow runs triggered in the last 24 hrs.
| Workflow |06/01|
| -------- | -------- |
| [database](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/database.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/899020434) |
| [prerelease](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/prerelease.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/898599846) |
| [segmentation](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/segmentation.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/898744500) |

paulb777:
Fix #8175

Fix AppCheck Catalyst build and add CI.  This fixes a regression building the zip.

Interestingly, this failure does not repro on Xcode 12.5

### DIFF
--- a/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
@@ -15,7 +15,7 @@
  */
 
 // Currently DCAppAttestService is available on iOS only.
-#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST  // Catalyst should be possible with Xcode 12.5+
 
 #import <DeviceCheck/DeviceCheck.h>
 
--- a/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
@@ -15,7 +15,7 @@
  */
 
 // Currently DCAppAttestService is available on iOS only.
-#if TARGET_OS_IOS
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
 
 #import <DeviceCheck/DeviceCheck.h>
 
--- a/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
+++ b/FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
@@ -31,4 +31,4 @@ API_UNAVAILABLE(macos, tvos, watchos)
 
 NS_ASSUME_NONNULL_END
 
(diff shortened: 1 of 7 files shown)


maksymmalyhin:
Looks like the availability of `DCAppAttestService` changed in Xcode 12.5 from
```
API_AVAILABLE(ios(14.0)) API_UNAVAILABLE(macos, tvos, watchos)
```
to
```
API_AVAILABLE(macos(11.0), ios(14.0)) API_UNAVAILABLE(tvos, watchos)
```

It also reflected in the [docs](https://developer.apple.com/documentation/devicecheck/dcappattestservice?language=objc).

maksymmalyhin:
Keychain dependent logic require additional configuration on Catalyst (enabling Keychain sharing). Probably it will be easier just to disable the Keychain dependent tests for Catalyst.

FirebaseAppCheck/Tests/Unit/AppAttestProvider/FIRAppAttestProviderTests.m
```diff
@@ -92,6 +92,7 @@ - (void)tearDown {
 
 #pragma mark - Init tests
 
+#if !TARGET_OS_MACCATALYST
```
maksymmalyhin, line 95: nit: Could you please add a comment explaining why the tests are disabled.

FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
```diff
@@ -15,7 +15,7 @@
  */
 
 // Currently DCAppAttestService is available on iOS only.
-#if TARGET_OS_IOS
+#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
```
maksymmalyhin, line 18: Maybe not in this PR, but, I think we should allow using App Attest on macOS and Catalyst with Xcode 12.5+. We probably will need to introduce a separate flag like `FIR_APP_ATTEST_SUPPORTED` where we will take into account the Xcode version, platform and OS version (if needed). Then this flag can be used instead of this condition.

maksymmalyhin (approved):
Thank you for fixing it!

FirebaseAppCheck/Sources/AppAttestProvider/DCAppAttestService+FIRAppAttestService.h
paulb777: Adding support in the future sounds good to me. Agreed that it is beyond the scope of this PR that is focused on fixing CI to unblock other work.

PR #8187, opened 2021-06-03
Reduce ObjectValue copies

schmidt-sebastian:
This PR reduces some implicit copies of ObjectValue that can be replaced with std::move. The only remaining callsites of ObjectValue copy-constructor are now in tests (and it is quite difficult to fix there).

### DIFF
--- a/Firestore/core/src/local/local_serializer.cc
+++ b/Firestore/core/src/local/local_serializer.cc
@@ -136,7 +136,7 @@ google_firestore_v1_Document LocalSerializer::EncodeDocument(
   result.name = rpc_serializer_.EncodeKey(doc.key());
 
   // Encode Document.fields (unless it's empty)
-  const google_firestore_v1_MapValue& fields_map = doc.value().map_value;
+  google_firestore_v1_MapValue fields_map = doc.value().map_value;
   SetRepeatedField(
       &result.fields, &result.fields_count,
       absl::Span<google_firestore_v1_MapValue_FieldsEntry>(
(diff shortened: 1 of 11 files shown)


Firestore/core/test/unit/local/remote_document_cache_test.cc
```diff
@@ -98,7 +98,7 @@ MATCHER_P(HasAtLeastDocs,
 }
 
 MATCHER_P(MatchesValue, expected, "") {
-  const google_firestore_v1_Value& actual_value = arg.data().Get();
+  const google_firestore_v1_Value& actual_value = arg.value();
```
var-const, line 101: Nit: it's best not to bind references to temporaries without need (I think we discussed the rationale at some point, so I'll omit it here, but I'm happy to talk about it if you'd like).

Firestore/core/src/local/local_serializer.cc
```diff
@@ -136,7 +136,7 @@ google_firestore_v1_Document LocalSerializer::EncodeDocument(
   result.name = rpc_serializer_.EncodeKey(doc.key());
 
   // Encode Document.fields (unless it's empty)
-  const google_firestore_v1_MapValue& fields_map = doc.data().Get().map_value;
+  const google_firestore_v1_MapValue& fields_map = doc.value().map_value;
```
var-const, line 139: Please make `fields_map` a value instead of a reference.

var-const (approved):

Firestore/core/src/local/local_serializer.cc
schmidt-sebastian: Done

Firestore/core/test/unit/local/remote_document_cache_test.cc
schmidt-sebastian: Done. I do remember the advice (at least for now...).

PR #8194, opened 2021-06-04
Add support for network request completed field.

jeremyjiang-dev:


FirebasePerformance/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Version 8.2.0
 * Update log messages with proper log levels.
+* Fix empty values in `network_info.request_completed_time_us` field from the [data schema](https://firebase.google.com/docs/perf-mon/bigquery-export#detailed_data_schema).
```
paulb777, line 3: While adding the `Completed` implementation, it seems to be removing the `Received` implementation. Does that need to be noted in the Release Notes?

paulb777 (approved):

FirebasePerformance/CHANGELOG.md
jeremyjiang-dev: I removed `Received` implementation from `didSendBodyData:totalBytesWritten:totalBytesExpectedToWrite:` because the response is not actually received at that point. `Received` is still captured in other places. For this reason, I think I don't have to update the Release Notes.

FirebasePerformance/Sources/Instrumentation/Network/Delegates/FPRNSURLSessionDelegate.m
```diff
@@ -39,12 +39,8 @@ - (void)URLSession:(NSURLSession *)session
   @try {
     FPRNetworkTrace *trace = [FPRNetworkTrace networkTraceFromObject:task];
     trace.requestSize = totalBytesSent;
-    if (totalBytesSent >= totalBytesExpectedToSend) {
-      if ([task.response isKindOfClass:[NSHTTPURLResponse class]]) {
-        NSHTTPURLResponse *response = (NSHTTPURLResponse *)task.response;
-        [trace didCompleteRequestWithResponse:response error:task.error];
-        [FPRNetworkTrace removeNetworkTraceFromObject:task];
-      }
+    if (totalBytesSent == totalBytesExpectedToSend) {
```
visumickey, line 42: can we leave this at ">=" instead of "=="? I remember usecases where the bytes sent was higher than expected and it might help in those usecases. Same in other places as well.

jeremyjiang-dev: Updated!

visumickey (approved):

PR #8208, opened 2021-06-07
Fixed analyze issue introduced in Xcode 12.5

paulb777:
Screen Shot 2021-06-07 at 7 12 51 AM


ncooke3 (approved):

maksymmalyhin (approved):
LGTM if we are sure that NULL is never passed as the errorPtr in a way that compiler and analyzer cannot catch. I would prefer to go with nullable pointer and a check for NULL

paulb777:
Thanks. Confirmed that the caller allocates the error pointer - https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDynamicLinks/Sources/FIRDynamicLinkNetworking.m#L325

PR #8210, opened 2021-06-07
Fixed analyze issues introduced in Xcode 12.5

paulb777:
Screen Shot 2021-06-07 at 7 37 01 AM
Screen Shot 2021-06-07 at 7 37 52 AM

### DIFF
--- a/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
+++ b/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
@@ -245,7 +245,7 @@ extern NSString *const FIRPhoneMultiFactorID;
     @param error The error that occurred if any.
     @return The reCAPTCHA token if successful.
  */
-- (NSString *)reCAPTCHATokenForURL:(NSURL *)URL error:(NSError **)error {
+- (nullable NSString *)reCAPTCHATokenForURL:(NSURL *)URL error:(NSError **_Nonnull)error {
   NSURLComponents *actualURLComponents = [NSURLComponents componentsWithURL:URL
                                                     resolvingAgainstBaseURL:NO];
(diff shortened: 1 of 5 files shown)


FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -245,7 +245,7 @@ + (instancetype)providerWithAuth:(FIRAuth *)auth {
     @param error The error that occurred if any.
     @return The reCAPTCHA token if successful.
  */
-- (NSString *)reCAPTCHATokenForURL:(NSURL *)URL error:(NSError **)error {
+- (nullable NSString *)reCAPTCHATokenForURL:(NSURL *)URL error:(NSError **_Nonnull)error {
```
maksymmalyhin, line 248: Same here, it will be safer to check for NULL because there might be cases when compiler and analyzer may not catch nullability issue.

maksymmalyhin (approved):
LGTM with a comment.

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
paulb777: The error pointer for the one call is allocated at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m#L667

maksymmalyhin: LGTM, thank you for checking. I sill think it is safer to assume nullable pointer to make it less probable to accidentally introduce a crash in the future.

paulb777: Done in #8214

PR #8213, opened 2021-06-07
Fix Xcode 13 package manager warnings.

ryanwilson:
These excludes no longer exist in the repo. We should also figure out a
test to make sure these don't get re-introduced.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -428,8 +428,6 @@ let package = Package(
         "CHANGELOG.md",
         "LICENSE",
         "README.md",
-        "Data/",
-        "Protos/",
         "ProtoSupport/",
         "UnitTests/",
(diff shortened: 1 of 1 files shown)


FirebaseCore/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
```
paulb777, line 1: Firebase 8.2.0

paulb777 (approved):

paulb777:
To check for reintroduction, we should start checking for warnings on SPM tests.

ryanwilson:
Merging now since CI already passed and I just did a small CHANGELOG update.

PR #8215, opened 2021-06-08
Remove Swift flag from user agent

paulb777:
Xcode 13 always returns true and we don't need more data to know that Swift is the future.

#no-changelog


maksymmalyhin (approved):
👍 

Please update [Swift tests](https://github.com/firebase/firebase-ios-sdk/blob/b01af578aab7ec06defe4bd76934c56d44d0b8f2/FirebaseCore/Tests/SwiftUnit/FirebaseAppTests.swift#L34) as well.

PR #8231, opened 2021-06-10
Wrap Protobuf for Query Filters in Messages

schmidt-sebastian:
This fixes a memory leak when we fail Query validation. 

The old flow was:
- Create nanopb struct
- Validate Query
- If valid, wrap in Message

This PR creates the Message first and then validates.

### DIFF
--- a/Firestore/core/src/api/query_core.cc
+++ b/Firestore/core/src/api/query_core.cc
@@ -240,6 +240,8 @@ Query Query::Filter(const FieldPath& field_path,
           Describe(op));
     } else if (op == Operator::In || op == Operator::NotIn) {
       ValidateDisjunctiveFilterElements(*value, op);
+      // TODO(mutabledocuments): See if we can remove this copy and modify the
+      // input values directly.
       google_firestore_v1_Value references{};
       references.which_value_type = google_firestore_v1_Value_array_value_tag;
       nanopb::SetRepeatedField(
--- a/Firestore/core/src/api/query_core.cc
+++ b/Firestore/core/src/api/query_core.cc
@@ -37,6 +37,7 @@
 #include "Firestore/core/src/util/exception.h"
 #include "absl/algorithm/container.h"
 #include "absl/strings/match.h"
+#include "absl/types/span.h"
 
 namespace firebase {
 namespace firestore {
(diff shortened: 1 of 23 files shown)


Firestore/core/src/core/field_filter.cc
```diff
@@ -81,30 +82,32 @@ const char* CanonicalName(Filter::Operator op) {
...
 
-FieldFilter FieldFilter::Create(const FieldPath& path,
-                                Operator op,
-                                google_firestore_v1_Value value_rhs) {
-  model::SortFields(value_rhs);
+FieldFilter FieldFilter::Create(
+    const FieldPath& path,
+    Operator op,
+    SharedMessage<google_firestore_v1_Value> value_rhs) {
+  google_firestore_v1_Value& value = value_rhs.operator*();
```
var-const, line 89: You can just do `*value_rhs`, no need to call `operator*` explicitly.

Firestore/core/src/api/query_core.cc
```diff
@@ -238,22 +239,28 @@ Query Query::Filter(FieldPath field_path,
...
           Describe(op));
     } else if (op == Operator::In || op == Operator::NotIn) {
-      ValidateDisjunctiveFilterElements(value, op);
-      google_firestore_v1_ArrayValue& array = value.array_value;
-      for (pb_size_t i = 0; i < array.values_count; ++i) {
-        array.values[i] =
-            ParseExpectedReferenceValue(array.values[i], type_describer);
-      }
+      ValidateDisjunctiveFilterElements(*value, op);
+      google_firestore_v1_Value references{};
```
var-const, line 243: Question: is creating a separate value and copying elements into it necessary? Can this be done in place?

Firestore/core/src/nanopb/message.h
```diff
@@ -188,8 +188,8 @@ class Message {
 };
 
 /**
- * A wrapper of const Message objects that facilitates shared ownership of
- * immutable Protobuf data.
+ * A wrapper of Message objects that facilitates shared ownership of Protobuf
```
var-const, line 191: It's a little unfortunate that this class is no longer immutable. How about adding a TODO to create a `template <typename const T>` specialization?

var-const (changes requested):

var-const:
Hmm, the majority of the PR seems to deal with wrapping values inside `Filter`s in `SharedMessage` -- should this be mentioned in the PR description?

Firestore/core/src/api/query_core.cc
schmidt-sebastian: I think nothing would break if we did not create a new array here, but we would be modifying the input array and would have to manually free values from the input array if we change the value type in ParseExpectedReferenceValue. 

I added a TODO since ParseExpectedReferenceValue will return a Message object in a follow up PR. I have to see if I can address your comment easily. In general, this is not code that gets invoked very often.

Firestore/core/src/core/field_filter.cc
schmidt-sebastian: Done

Firestore/core/src/nanopb/message.h
schmidt-sebastian: Done

var-const (approved):

PR #8234, opened 2021-06-10
Speed up SPM CI by reducing full Firebase rebuilds

paulb777:
#no-changelog

### DIFF
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -48,10 +48,13 @@ jobs:
       matrix:
         target: [tvOS, macOS, catalyst]
         # Full set of Firebase-Package tests only run on iOS because of Analytics.
-        test: [objc-import-test, swift-test, version-test]
     steps:
     - uses: actions/checkout@v2
     - name: Initialize xcodebuild
       run: scripts/setup_spm_tests.sh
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):
Looks great! What kind of difference do we see with this?

.github/workflows/spm.yml
```diff
@@ -48,10 +48,13 @@ jobs:
       matrix:
         target: [tvOS, macOS, catalyst]
         # Full set of Firebase-Package tests only run on iOS because of Analytics.
```
ryanwilson, line 50: Nit: comment should go above now? or not relevant anymore?

ryanwilson (approved):

.github/workflows/spm.yml
paulb777: Still relevant

paulb777:
We go from three jobs per platform - see https://github.com/firebase/firebase-ios-sdk/runs/2800987939?check_suite_focus=true to one job per platform - see https://github.com/firebase/firebase-ios-sdk/actions/runs/926785832

tvos goes from 27 cpu minutes to 9
macos 23 to 14
catalyst 26 to 9

PR #8236, opened 2021-06-11
ObjectValue should take ownership of values

schmidt-sebastian:
This changes Set() and SetAll() to own the memory of the objects passes to it. This is important since the mutations already copy the data before passing (to make sure that the internal data is not modified).

### DIFF
--- a/Firestore/core/src/model/object_value.h
+++ b/Firestore/core/src/model/object_value.h
@@ -84,14 +84,18 @@ class ObjectValue {
   /**
    * Sets the field to the provided value.
    *
+   * Takes ownership of value.
+   *
    * @param path The field path to set. The path must not be empty.
    * @param value The value to set.
    */
-  void Set(const FieldPath& path, const google_firestore_v1_Value& value);
+  void Set(const FieldPath& path, google_firestore_v1_Value value);
 
   /**
    * Sets the provided fields to the provided values. Fields set to `nullopt`
    * are deleted.
+   *
+   * Takes ownership of value.
    *
    * @param data A map of fields to values (or nullopt for deletes)
    */
--- a/Firestore/core/src/model/object_value.h
+++ b/Firestore/core/src/model/object_value.h
@@ -95,7 +95,7 @@ class ObjectValue {
    * Sets the provided fields to the provided values. Fields set to `nullopt`
    * are deleted.
    *
-   * Takes ownership of value.
+   * Takes ownership of data.
    *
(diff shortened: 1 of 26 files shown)


schmidt-sebastian:
This might change again with https://github.com/firebase/firebase-ios-sdk/pull/8244/files

Firestore/core/src/model/patch_mutation.cc
```diff
@@ -113,7 +113,10 @@ PatchMutation::Rep::GetPatch() const {
   std::map<FieldPath, absl::optional<google_firestore_v1_Value>> result;
   for (const FieldPath& path : mask_) {
     if (!path.empty()) {
-      result[path] = value_.Get(path);
+      auto value = value_.Get(path);
+      result[path] =
```
var-const, line 117: Small nit: would using `if/else` be a little cleaner? (the type can be omitted in that case)
```cpp
if (value) {
  result[path] = DeepClone(*value);
} else {
  result[path] = absl::nullopt;
}
```

Firestore/core/src/model/object_value.h
```diff
@@ -84,15 +84,19 @@ class ObjectValue {
   /**
    * Sets the field to the provided value.
    *
+   * Takes ownership of value.
+   *
    * @param path The field path to set. The path must not be empty.
    * @param value The value to set.
    */
-  void Set(const FieldPath& path, const google_firestore_v1_Value& value);
+  void Set(const FieldPath& path, google_firestore_v1_Value value);
```
var-const, line 92: What do you think about creating a move-only wrapper over protos and using it in function signatures like this?
```cpp
void Set(const FieldPath& path, Unique<google_firestore_v1_Value> value);
// ...
foo.Set(path, std::move(value));
```

The only advantage would be as a sort of documentation -- the caller would have to explicitly move the value, and the fact that ownership is being transferred would be clear in the signature.

var-const (approved):

Firestore/core/src/model/object_value.h
schmidt-sebastian: https://github.com/firebase/firebase-ios-sdk/pull/8244 will replace this argument with `Message`, which should signal the ownership.

Firestore/core/src/model/patch_mutation.cc
schmidt-sebastian: Looks much better. Thanks.

PR #8242, opened 2021-06-11
Bundle takes SharedMessage

schmidt-sebastian:


Firestore/core/src/remote/serializer.cc
```diff
@@ -1115,21 +1115,21 @@ google_firestore_v1_Cursor Serializer::EncodeBound(const Bound& bound) const {
...
+      absl::Span<google_firestore_v1_Value>(bound.position()->values,
+                                            bound.position()->values_count),
       [](const google_firestore_v1_Value& value) { return DeepClone(value); });
   return result;
 }
 
 Bound Serializer::DecodeBound(google_firestore_v1_Cursor& cursor) const {
-  google_firestore_v1_ArrayValue index_components;
-  SetRepeatedField(&index_components.values, &index_components.values_count,
+  SharedMessage<google_firestore_v1_ArrayValue> index_components{{}};
```
var-const, line 1125: This initialization looks a little weird (though it does the right thing, AFAICT). Consider adding a named "constructor" for this case, something like:
```cpp
auto index_components = SharedMessage<google_firestore_v1_ArrayValue>::CreateDefault();
```

The name can definitely be changed (maybe it should be `Empty` or something).

Applies to the other files as well.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -1011,18 +1012,20 @@ TEST_F(BundleSerializerTest, DecodesStartAtCursor) {
   core::Query original =
       testutil::Query("colls")
           .AddingOrderBy(OrderBy("f1", "asc"))
-          .StartingAt(core::Bound::FromValue({Array("f1", 1000)},
-                                             /* is_before= */ true));
+          .StartingAt(core::Bound::FromValue(
+              SharedMessage<google_firestore_v1_ArrayValue>{Array("f1", 1000)},
```
var-const, line 1016: For cases like this, the compiler would have been able to figure out the template type if it were a function, not a class (in C++11; in C++17 class template arguments can be deduced as well). This is basically the reason for functions like `std::make_pair`. Consider adding a free function like `MakeSharedMessage` -- then this could likely be:
```cpp
MakeSharedMessage(Array("f1", 1000))
```
(applies throughout)

var-const (changes requested):
@schmidt-sebastian The PR name seems to only refer to a part of the actual code changes.

Firestore/core/src/remote/serializer.cc
schmidt-sebastian: I used the new MakeSharedMessage function. I like the explicit nature a bit better as a `SharedMessage<google_firestore_v1_ArrayValue>::Empty()` factory might suggest that it returns the same singleton instance.

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
schmidt-sebastian: Done. I was wondering why I needed to put the class parameter everywhere. Thanks for the cleanup suggestion.

Firestore/core/src/nanopb/message.h
```diff
@@ -243,6 +243,11 @@ class SharedMessage {
   std::shared_ptr<Message<T>> message_;
 };
 
+template <typename T>
+SharedMessage<T> MakeSharedMessage(const T& proto) {
+  return SharedMessage<T>(std::move(proto));
```
var-const, line 248: Nit: moving a const reference is extraneous.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -210,7 +211,7 @@ Filter InvalidFilter() {
   // The exact value doesn't matter. Note that there's no way to create the base
   // class `Filter`, so it has to be one of the derived classes.
   return FieldFilter::Create({}, {},
-                             SharedMessage<google_firestore_v1_Value>({}));
+                             MakeSharedMessage<google_firestore_v1_Value>({}));
```
var-const, line 214: Nit: perhaps it would be slightly more readable as:
```cpp
MakeSharedMessage(google_firestore_v1_Value{}));
```

var-const (approved):

Firestore/core/src/nanopb/message.h
schmidt-sebastian: Removed.

Firestore/core/src/bundle/bundle_serializer.cc
schmidt-sebastian: Agreed and fixed.

PR #8248, opened 2021-06-12
Explicitly check that document and collection IDs given to the public API aren't empty

### LINKED ISSUE #8218: Non-nil empty strings cause Firestore to crash instead of throwing an error gracefully
### Environment

  * Xcode version: 12.5
  * Firebase SDK version: 8.0.0
  * Firebase Component: Firestore

### The Problem

Currently, attempting a Firestore document load (of any nature) with an empty ID causes a runtime exception which results in a crash.  By "empty ID" I am referring to a string whose character count is zero, but where the object is non-nil. Literally, an empty string.

While Firestore handles a number of errors and returns the failure results gracefully in many cases, this is not one of those. It would be nice if the SDK could catch this and return an error

#### Steps to reproduce:

 1. Start in a Swift environment
 2. Create a Firestore document or collection request
 3. In the document ID parameter: `.document(ID)`, supply a non-nil but empty string such as `""`.
 4. Firestore throws a fatal error and causes the application to crash.
 5. Crash logs are difficult to parse, as the throw occurs deep inside the Firestore core framework.

#### Relevant Code

It appears, from crash logs that this occurs deep in the Firestore framework (`core/src/api/collection_reference.cc`):

```cc
return DocumentReference(std::move(path), firestore());
```

There seems to be an attempt to handle the issue and fail gracefully in `Source/API/FIRCollectionReference.mm`):

```objc
- (FIRDocumentReference *)documentWithPath:(NSString *)documentPath {
  if (!documentPath) {
    ThrowInvalidArgument("Document path cannot be nil.");
  }
  DocumentReference child = self.reference.Document(util::MakeString(documentPath));
  return [[FIRDocumentReference alloc] initWithReference:std::move(child)];
}
```

Unfortunately, `if(!documentPath)` does not handle cases where strings may be non-nil, yet still empty (something I think may have been made possible in a recent Swift release 🥴?).

var-const:
Currently, the error would look like `Document references must have an even number of segments, but has 0`.

Fixes #8218.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+- [changed] Passing in an empty document ID, collection group ID, or collection
+  path will now result in a more readable error (#8218).
+
 # v7.9.0
 - [feature] Added support for Firestore Bundles via
   `FIRFirestore.loadBundle`, `FIRFirestore.loadBundleStream` and
--- a/Firestore/Example/Tests/Integration/API/FIRValidationTests.mm
+++ b/Firestore/Example/Tests/Integration/API/FIRValidationTests.mm
@@ -110,10 +110,17 @@ namespace testutil = firebase::firestore::testutil;
   FSTAssertThrows([baseDocRef collectionWithPath:nil], nilError);
 }
 
+- (void)testEmptyCollectionPathsFail {
+  FIRDocumentReference *baseDocRef = [self.db documentWithPath:@"foo/bar"];
+  NSString *emptyError = @"Collection path cannot be empty.";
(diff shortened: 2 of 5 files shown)


var-const (commented):
@schmidt-sebastian Sebastian, do you think this needs to be mentioned in the changelog?

schmidt-sebastian (approved):
It should be mentioned in the changelog in case it is the only change in the release.

nidegen:
This is actually an Issue for me. I was using this to refer either to top level domain collections or child collections from the same `documentRef` variable, i.e. `database = firestore.document(isDevel ? "devel/001" : "")`.

This is breaking API for me, it would be nice if it was undone:)

PR #7900, opened 2021-04-14
Analytics 7.11.0

moemcd:
Analytics 7.11.0 podspec.json files, SPM, and Firebase podspec updates.

PR #8256, opened 2021-06-14
Fix versioning bug in release tooling

### LINKED ISSUE #8179: Firebase/AnalyticsWithoutAdIdSupport subspec still points to `'FirebaseAnalytics/WithoutAdIdSupport', '~> 8.0.0'`, not 8.1.0

  * Xcode version: 12.5
  * Firebase SDK version: 8.1.0
  * Firebase Component: Core (Auth, Core, Database, Firestore, Messaging, Storage, etc)

#### Steps to reproduce:

1. Depend on `pod 'Firebase/AnalyticsWithoutAdIdSupport', '~> 8.1.0'` in your Podfile
2. Run `pod update`
3. Inspect Podfile.lock or run `pod outdated`, and notice that `FirebaseAnalytics/WithoutAdIdSupport (8.0.0)` was installed, even though v8.1.0 of `FirebaseAnalytics` is available.

It appears that in the most recent version bump PR (#8152), the `AnalyticsWithoutAdIdSupport` subspec in `Firebase.podspec` was missed, and it still points to v8.0.0 instead of 8.1.0:

https://github.com/firebase/firebase-ios-sdk/blob/acf106a33d50e893f85a8563a19557b3df6110d1/Firebase.podspec#L72

I'm guessing that this is probably because it's a new subspec and the corresponding version bump tooling needs to be updated?

ncooke3:
When the release tooling [iterates through all of the defined `Pods`](https://github.com/firebase/firebase-ios-sdk/blob/5aaf38c022718421c389cfafa4fe2776227966d2/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift#L74), it uses the `range(of:)` String API to find the first occurrence of the `pod` and then bumps the versioning. Because the API only finds the first occurrence, each pod will only get versioned bumped once. Because the `FirebaseAnalytics` substring now occurs in two places (after PR #7900 added `'FirebaseAnalytics/WithoutAdIDSupport'`):  

https://github.com/firebase/firebase-ios-sdk/blob/acf106a33d50e893f85a8563a19557b3df6110d1/Firebase.podspec#L37

and 

https://github.com/firebase/firebase-ios-sdk/blob/acf106a33d50e893f85a8563a19557b3df6110d1/Firebase.podspec#L72

it was only bumped once (the first occurrence). 

A new String API has been added to collect all of the ranges of the occurrences of a given substring.

Fixes #8179

### DIFF
--- a/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
+++ b/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
@@ -97,7 +97,7 @@ struct InitializeRelease {
     }
   }
 
-  /// Update the existing version to the given version by writing to a given string usign the provided range.
+  /// Update the existing version to the given version by writing to a given string using the provided range.
   /// - Parameters:
   ///   - contents: A reference to a String containing a version that will be updated.
   ///   - range: The range containing a version substring that will be updated.
--- a/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
+++ b/ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
@@ -75,41 +75,18 @@ struct InitializeRelease {
       let pod = firebasePod.name
       let version = firebasePod.isBeta ? firebaseVersion + "-beta" : firebaseVersion
       if pod == "Firebase" {
-        // TODO: This then block is redundant with the updatePodspecs function above and is left
-        // until we decide to go with Swift or sed.
-        // Replace version in string like s.version = '6.9.0'
+        // TODO: This block is redundant with `updatePodspecs`. Decide to go with Swift or sed.
         guard let range = contents.range(of: "s.version") else {
           fatalError("Could not find version of Firebase pod in podspec at \(podspecFile)")
         }
-        var versionStartIndex = contents.index(range.upperBound, offsetBy: 1)
-        while contents[versionStartIndex] != "'" {
-          versionStartIndex = contents.index(versionStartIndex, offsetBy: 1)
-        }
-        var versionEndIndex = contents.index(versionStartIndex, offsetBy: 1)
-        while contents[versionEndIndex] != "'" {
-          versionEndIndex = contents.index(versionEndIndex, offsetBy: 1)
-        }
-        contents.removeSubrange(versionStartIndex ... versionEndIndex)
-        contents.insert(contentsOf: "'" + version + "'", at: versionStartIndex)
-      } else {
-        // Replace version in string like ss.dependency 'FirebaseCore', '6.3.0'
-        // Obtain all the ranges of `pod`'s occurences.
-        guard let ranges = contents.ranges(of: pod) else {
-          // This pod is not a top-level Firebase pod dependency.
-          continue
-        }
+        // Replace version in string like s.version = '6.9.0'
+        updateVersion(&contents, in: range, to: version)
 
-        for range in ranges {
-          var versionStartIndex = contents.index(range.upperBound, offsetBy: 2)
-          while !contents[versionStartIndex].isWholeNumber {
-            versionStartIndex = contents.index(versionStartIndex, offsetBy: 1)
-          }
-          var versionEndIndex = contents.index(versionStartIndex, offsetBy: 1)
-          while contents[versionEndIndex] != "'" {
-            versionEndIndex = contents.index(versionEndIndex, offsetBy: 1)
-          }
-          contents.removeSubrange(versionStartIndex ... versionEndIndex)
-          contents.insert(contentsOf: version + "'", at: versionStartIndex)
+      } else {
+        // Iterate through all the ranges of `pod`'s occurences.
+        for range in contents.ranges(of: pod) {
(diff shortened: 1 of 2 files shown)


ryanwilson (approved):
Changes LGTM - were you able to validate this works locally and doesn't have any unintended consequences?

paulb777 (approved):
LGTM assuming that a manual test has confirmed only the one expected change.

No need to wait for the zip CI.

ReleaseTooling/Sources/Utils/String+Utils.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import Foundation
+
+/// Utilities to simplify String operations.
+public extension String {
+  /// Finds and returns the ranges of all occurrences of a given string within a given range of the String, subject to given options,
+  /// using the specified locale, if any.
+  /// - Returns: An an optional array of ranges where each range corresponds to an occurence of the substring in the given string.
+  func ranges<T: StringProtocol>(of substring: T, options: CompareOptions = .literal,
+                                 locale: Locale? = nil) -> [Range<Index>]? {
```
maksymmalyhin, line 25: nit: What's the difference between `[]` and `nil` returned? If no difference, than I would suggest to use non-optional array. I think it will simplify the client code as well (no need for `guard`).

ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
```diff
@@ -93,20 +93,24 @@ struct InitializeRelease {
...
-        var versionStartIndex = contents.index(range.upperBound, offsetBy: 2)
-        while !contents[versionStartIndex].isWholeNumber {
-          versionStartIndex = contents.index(versionStartIndex, offsetBy: 1)
-        }
-        var versionEndIndex = contents.index(versionStartIndex, offsetBy: 1)
-        while contents[versionEndIndex] != "'" {
-          versionEndIndex = contents.index(versionEndIndex, offsetBy: 1)
+
+        for range in ranges {
+          var versionStartIndex = contents.index(range.upperBound, offsetBy: 2)
```
maksymmalyhin, line 103: optional: I know, it was here before your change, but it would be useful to add either comments explaining the intent or refactor it to a method(s) to make it self-documentary or both. I don't feel too strong about it but I think it will help future us.

maksymmalyhin (approved):
LGTM, a couple optional comments.

maksymmalyhin:
> Changes LGTM - were you able to validate this works locally and doesn't have any unintended consequences?

I feel we have enough logic here to start adding tests to it. Test for this piece of logic could be giving different podspecs as inputs and validate expected outputs. Fortunately SPM provides pretty good support for tests.

ReleaseTooling/Sources/Utils/String+Utils.swift
ncooke3: Agreed

ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
ncooke3: SGTM. To simplify this method even more, I'm curious if it was ever considered to bump the versions by calling
`contents.replacingOccurrences(of: oldVersion, with: newVersion)`

maksymmalyhin: I guess we may not always now old version for each pod that's why we try to parse the version from the file first. I think @paulb777 should have more context on it as the original author.

paulb777: Yep, when it started, there weren't consistent oldVersions. Also, a generic string replace, in theory, could change the string in a name or comment incorrectly. 

However, it could be fine to simplify now since `oldVersion` should be able to be defined and the context check might be not worth the extra complexity.

ncooke3: @paulb777 Makes sense. I was in the middle of refactoring some logic into a method when you replied so I figured I'd at least push those changes. WDYT of the current state? cc: @maksymmalyhin

maksymmalyhin (approved):
LGTM, thank you for the cleanup!

ncooke3 (commented):
@ryanwilson @paulb777 @maksymmalyhin thanks for the feedback. I addressed some comments and manually verified that running the releaser bumps accordingly (and fixes the bug).

Following the conversation from above, I'm happy to simplify the implementation with `replaceOccurrences`, add old/new version launch arguments, and tests. Is this something I should work on now or add to the backlog after more conversation?

The PR in it's current state should be good for next release

ReleaseTooling/Sources/FirebaseReleaser/InitializeRelease.swift
paulb777: LGTM

maksymmalyhin: IMO the concerns about unintentional changes look still relevant. We may mitigate it by using e.g. regex, but I'm not sure if it will end up more readable eventually. The current state LGTM.

PR #8258, opened 2021-06-14
Cache Functions instances by app

### LINKED ISSUE #7783: Callable function returns 'Unauthenticated' error when using the functions emulator

  * Xcode version: 12.4
  * Firebase SDK version: FirebaseFunctions (7.8.0)
  * Firebase Component: Functions

When using the emulator for functions and calling a callable function from the app, the request immediately returns an error: Unauthenticated, there are no logs from the emulator when the error is returned so i am guessing the request doesn't even reach the function.

#### Steps to reproduce:

1. Start emulators
2. Start ios app with emulators configured
3. call the callable function

Actual result: fails with error: Unauthenticated.

Expected result: The callable function is executed successfully.

#### Relevant Code:
(Swift)
```
Auth.auth().useEmulator(withHost:"localhost", port:9099)
Functions.functions().useEmulator(withHost: "localhost", port: 5001)

Functions.functions().httpsCallable("functionName").call(functionData) { (result, error) in }
```

The code above fails with the following details:
error.code: `__C.FIRFunctionsErrorCode`,
error.localizedDescription: `Unauthenticated`

morganchen12:
Fixes #7783.

### DIFF
--- a/Functions/Example/Tests/FIRFunctionsTests.m
+++ b/Functions/Example/Tests/FIRFunctionsTests.m
@@ -104,6 +104,11 @@
   functions1 = [FIRFunctions functionsForApp:app2 customDomain:@"test_domain"];
 
   XCTAssertEqualObjects(functions1, functions2);
+
+  functions1 = [FIRFunctions functionsForCustomDomain:@"test_domain"];
+  functions2 = [FIRFunctions functionsForApp:[FIRApp defaultApp]];
+
+  XCTAssertEqualObjects(functions1, functions2);
 }
 
 - (void)testURLWithName {
--- a/Functions/Example/Tests/FIRFunctionsTests.m
+++ b/Functions/Example/Tests/FIRFunctionsTests.m
@@ -90,24 +90,27 @@
   [FIRApp configureWithOptions:options];
 
   FIRFunctions *functions1 = [FIRFunctions functions];
-  FIRFunctions *functions2 = [FIRFunctions functionsForRegion:@"us-central"];
+  FIRFunctions *functions2 = [FIRFunctions functionsForApp:[FIRApp defaultApp]];
 
   XCTAssertEqualObjects(functions1, functions2);
 
   [FIRApp configureWithName:@"test" options:options];
   FIRApp *app2 = [FIRApp appNamed:@"test"];
 
-  functions2 = [FIRFunctions functionsForApp:app2 region:@"us-central"];
+  functions2 = [FIRFunctions functionsForApp:app2 region:@"us-central2"];
 
   XCTAssertNotEqualObjects(functions1, functions2);
 
-  functions1 = [FIRFunctions functionsForApp:app2 customDomain:@"test_domain"];
+  functions1 = [FIRFunctions functionsForApp:app2 region:@"us-central2"];
 
   XCTAssertEqualObjects(functions1, functions2);
 
   functions1 = [FIRFunctions functionsForCustomDomain:@"test_domain"];
-  functions2 = [FIRFunctions functionsForApp:[FIRApp defaultApp]];
+  functions2 = [FIRFunctions functionsForRegion:@"us-central1"];
 
+  XCTAssertNotEqualObjects(functions1, functions2);
+
+  functions2 = [FIRFunctions functionsForApp:[FIRApp defaultApp] customDomain:@"test_domain"];
   XCTAssertEqualObjects(functions1, functions2);
 }
 
--- a/Functions/Example/Tests/FIRFunctionsTests.m
+++ b/Functions/Example/Tests/FIRFunctionsTests.m
@@ -20,6 +20,8 @@
 #import "SharedTestUtilities/AppCheckFake/FIRAppCheckFake.h"
 #import "SharedTestUtilities/AppCheckFake/FIRAppCheckTokenResultFake.h"
 
+#import <FirebaseCore/FirebaseCore.h>
+
 #if SWIFT_PACKAGE
 @import GTMSessionFetcherCore;
 #else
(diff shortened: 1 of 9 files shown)


Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -95,30 +95,57 @@ + (void)load {
   return @[ internalProvider ];
 }
 
+// A table mapping FIRApp instances to FIRStorage instances, used to cache FIRStorage instances
```
paulb777, line 98: s/FIRStorage/FIRFunctions/

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
```diff
@@ -58,15 +58,17 @@ NS_SWIFT_NAME(Functions)
     NS_SWIFT_NAME(functions(customDomain:));
 
 /**
- * Creates a Cloud Functions client with the given app and region.
+ * Creates a Cloud Functions client with the given app and region. If an instance of Storage
```
paulb777, line 61: Storage ...

Functions/FirebaseFunctions/Public/FirebaseFunctions/FIRFunctions.h
```diff
@@ -58,15 +58,17 @@ NS_SWIFT_NAME(Functions)
...
+ * associated with the given app already exists, the region parameter will be ignored.
  * @param app The app for the Firebase project.
  * @param region The region for the http trigger, such as "us-central1".
  */
 + (instancetype)functionsForApp:(FIRApp *)app
                          region:(NSString *)region NS_SWIFT_NAME(functions(app:region:));
 
 /**
- * Creates a Cloud Functions client with the given app and region.
+ * Creates a Cloud Functions client with the given app and region. If an instance of Storage
```
paulb777, line 70: Storage

Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -95,30 +95,57 @@ + (void)load {
...
+// on a per-app basis. Does not retain either its keys or its values, since FIRApps may be deleted
+// without notifying their respective FIRStorage instances, and FIRStorage instances may be short-
+// lived.
++ (NSMapTable<FIRApp *, FIRFunctions *> *)instanceCache {
```
paulb777, line 102: @ryanwilson Please confirm this is ok from an API stability perspective

paulb777 (commented):
Is the incremental coverage checker correct with its four reports? cc: @granluo

granluo:
> Is the incremental coverage checker correct with its four reports? cc: @granluo

For the latest reports, I double checked with the xcresult artifact in this [run](https://github.com/firebase/firebase-ios-sdk/actions/runs/937564070). The reports should be correct.

Functions/FirebaseFunctions/FIRFunctions.m
morganchen12: This may result in surprising behavior since consumers don't tend to retain the main [`FIRFunctions functions`] instance. Thoughts?

Functions/Example/Tests/FIRFunctionsTests.m
```diff
@@ -81,6 +83,29 @@ - (void)tearDown {
...
 }
 
+- (void)testFunctionsInstanceIsStablePerApp {
+  FIROptions *options =
+      [[FIROptions alloc] initWithGoogleAppID:@"0:0000000000000:ios:0000000000000000"
+                                  GCMSenderID:@"00000000000000000-00000000000-000000000"];
+  [FIRApp configureWithOptions:options];
+
+  FIRFunctions *functions1 = [FIRFunctions functions];
+  FIRFunctions *functions2 = [FIRFunctions functionsForRegion:@"us-central"];
```
ryanwilson, line 93: Can you please comment that this is the default region and that's why we expect them to be the same?

Functions/Example/Tests/FIRFunctionsTests.m
```diff
@@ -81,6 +83,29 @@ - (void)tearDown {
...
+  [FIRApp configureWithName:@"test" options:options];
+  FIRApp *app2 = [FIRApp appNamed:@"test"];
+
+  functions2 = [FIRFunctions functionsForApp:app2 region:@"us-central"];
+
+  XCTAssertNotEqualObjects(functions1, functions2);
+
+  functions1 = [FIRFunctions functionsForApp:app2 customDomain:@"test_domain"];
+
+  XCTAssertEqualObjects(functions1, functions2);
```
ryanwilson, line 106: Providing a custom domain should be a different instance of Functions, no?

Functions/FirebaseFunctions/FIRFunctions.m
```diff
@@ -95,30 +95,57 @@ + (void)load {
   return @[ internalProvider ];
 }
 
+// A table mapping FIRApp instances to FIRFunctions instances, used to cache FIRFunctions instances
+// on a per-app basis. Does not retain either its keys or its values, since FIRApps may be deleted
+// without notifying their respective FIRFunctions instances, and FIRFunctions instances may be
+// short- lived.
++ (NSMapTable<FIRApp *, FIRFunctions *> *)instanceCache {
```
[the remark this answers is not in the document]
morganchen12: Yep, that's what I was mentioning in my reply to Paul above. I'm leaning toward instead retaining the FIRFunctions instance strongly and the FIRApp weakly, so if the FIRApp is deleted the FIRFunctions instance will be released but otherwise it'll be retained.

Functions/Example/Tests/FIRFunctionsTests.m
morganchen12: These are the same because the current implementation returns an existing FIRFunctions if one was already configured for the given app (in this case implicitly the default app), ignoring the region parameter. If we allow this and `functionsForCustomDomain:` to return new instances every time, they'll also have the issue of emulator settings not persisting across instances; otherwise we need to have a more complex cache key.

Functions/Example/Tests/FIRFunctionsTests.m
morganchen12: See above comment for reasoning.

morganchen12:
Talked to Ryan offline and we agreed it'd be best to move Functions to broader FIRComponent adoption, with these specific points:

- Functions should attempt to minimize creation of new instances as much as possible
- Functions should return a new instance per combination of app, region, and custom domain

It's a bit late for code freeze, but we can punt to M99 if necessary.

paulb777 (approved):
LGTM on green

Functions/FirebaseFunctions/FIRFunctionsComponent.m
```diff
@@ -0,0 +1,128 @@
...
+- (void)appWillBeDeleted:(FIRApp *)app {
+  NSString *appName = app.name;
+  if (appName == nil) {
+    return;
+  }
+
+  os_unfair_lock_lock(&_instancesLock);
+  [self.instances removeObjectForKey:appName];
+  os_unfair_lock_unlock(&_instancesLock);
+}
```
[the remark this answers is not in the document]
morganchen12: I'm making the executive decision to not test these lines.

PR #8267, opened 2021-06-17
Update versions for Release 8.2.0

ryanwilson:


paulb777:
The Firebase podspec is missing from SpecsStaging. After that LGTM on green.

paulb777 (approved):

ncooke3 (commented):
LGTM as well. Looks like all versions were correctly bumped 👍

ryanwilson:
Added the Firebase podspec to SpecsStaging, will kick off failures. Thanks!

Firebase.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
```
ncooke3, line 1: Curious if we know why the `Firebase.podspec` wasn't pushed to SpecsStaging? I just did a log-only run of firebase-releaser and it logs that `Firebase.podspec` is pod repo push'd.

ryanwilson: Good question and good catch - this is because when you run it it goes through one by one and will stop when one fails. One did fail and I was handling the rest manually. I knew it wasn't published yet but was trying to do multiple things at once (it was pushing while I opened the PR). Shouldn't happen when the tools are working as expected and all pods pass.

PR #8279, opened 2021-06-20
Update FIS IID error message

paulb777:
Update confusing error message now that IID has been removed.

Context at https://stackoverflow.com/questions/68049548/firebaseinstallations-will-not-work-correctly-with-current-version-of-firebase-i

#no-changelog

### DIFF
--- a/FirebaseInstallations/Source/Library/FIRInstallations.m
+++ b/FirebaseInstallations/Source/Library/FIRInstallations.m
@@ -256,9 +256,12 @@ static const NSUInteger kExpectedAPIKeyLength = 39;
   return;
 #else
   if (![self isIIDVersionCompatible]) {
-    [NSException raise:kFirebaseInstallationsErrorDomain
-                format:@"FirebaseInstallations will not work correctly with current version of "
-                       @"Firebase Instance ID. Please update your Firebase Instance ID version."];
+    [NSException
+         raise:kFirebaseInstallationsErrorDomain
+        format:@"Firebase Instance ID is not compatible with Firebase 8.x+. Please remove the "
+               @"dependency from the app. See the documentation at "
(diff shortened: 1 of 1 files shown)


ncooke3 (approved):

paulb777:
@maksymmalyhin Should FIS still be checking the IID version?

FirebaseInstallations/Source/Library/FIRInstallations.m
```diff
@@ -256,9 +256,12 @@ + (void)assertCompatibleIIDVersion {
...
   if (![self isIIDVersionCompatible]) {
-    [NSException raise:kFirebaseInstallationsErrorDomain
-                format:@"FirebaseInstallations will not work correctly with current version of "
-                       @"Firebase Instance ID. Please update your Firebase Instance ID version."];
+    [NSException
+         raise:kFirebaseInstallationsErrorDomain
+        format:@"Firebase Instance ID is not compatible with Firebase 8.x+. Please remove the "
+               @"dependency from the app. See the documentation at "
+               @"https://firebase.google.com/docs/cloud-messaging/ios/"
+               @"client#fetching-the-current-registration-token."];
```
paulb777, line 264: @chliangGoogle Is this the right link?

maksymmalyhin:
> @maksymmalyhin Should FIS still be checking the IID version?

As long as we have apps with old Firebase version that may potentially be updated we need to keep the migration logic and a safeguard check to warn that IID must be deleted. I think we still have pretty long tail of apps using old Firebase SDKs that may benefit from the check during the upgrade process.

maksymmalyhin (approved):
It's much more clear now, thank you Paul!

chliangGoogle (approved):

Issue #7911, opened 2021-04-16
DispatchQueue seems not truly effective in Cloud Firestore.


lorenzofiamingo:
### Environment

  * Xcode version: 12
  * Firebase SDK version: 7.10.0
  * Installation method: SPM
  * Firebase Component: Firestore

### Problem

I need to make a call to Firestore and wait for the query result. Since Firestore make its async calls on the main thread and I need my code to be on the main thread, I have to change the default `dispatchQueue` used by Firestore callbacks, in order to execute them while I'm waiting. When I run the project this is the output:

```
2021-04-16 19:06:12.080367+0200 FirestoreIncrementalStore[1794:547410] 7.10.0 - [Firebase/Firestore][I-FST000001] Could not reach Cloud Firestore backend. Backend didn't respond within 10 seconds.
 This typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.
Async Optional(<FIRQuerySnapshot: 0x12eb044e0>)
Sync Optional(<FIRQuerySnapshot: 0x12eb044e0>)
```

So the code works, since I received the correct snapshot, but I had to wait 10 secs for the timeout of the internet connection (untrue, since the query ran well). My only explanation is that exist a piece of async code inside Firestore APIs that is not performed using the `dispatchQueue` of `firestore.settings` (maybe the callback that stops the timer).

#### Steps to reproduce:

Run the following code in a sample project

```swift
let firestore: Firestore = .firestore()
let dispatchGroup = DispatchGroup()
var syncSnapshot: QuerySnapshot? = nil
dispatchGroup.enter()
let settings = FirestoreSettings()
settings.dispatchQueue = DispatchQueue(label: "com.test.my-thread")
firestore.settings = settings
firestore.collection(path).getDocuments { snapshot, error in
    guard let snapshot = snapshot else { return }
    syncSnapshot = snapshot
    print("Async", syncSnapshot)
    dispatchGroup.leave()
}
dispatchGroup.wait()
print("Sync", syncSnapshot)
```
#### PS:

`firestore.settings.dispatchQueue = DispatchQueue(label: "com.test.my-thread")` doesn't work in order to set the queue.

rizafran:
Hi @lorenzofiamingo, we'll try to reproduce this, and let you know with the updates or findings. It'll also be better if you can share a sample project that I can run locally. Thanks.

dconeybe:
[...]

PR #8288, opened 2021-06-22
Experimental auth custom callback queue feature

morganchen12:
A proposal implementation for #7911.


paulb777:
@morganchen12 Do we want to move forward with this?

morganchen12:
It does solve a real problem, but it got lost in the auth/firestore cross team comms. I'll try to pick this up again in Q1.

PR #8296, opened 2021-06-23
Punt pause task to the end of the queue

### LINKED ISSUE #8294: Latest GTMSessionFetcher's new deprecation breaks Storage CI
See https://github.com/firebase/firebase-ios-sdk/pull/8292/checks?check_run_id=2898960029 and 

Screen Shot 2021-06-23 at 1 47 16 PM

From GTMsessionFetcherService.h:
// DEPRECATED: Callers should use XCTestExpectation instead.
//
// Spin the run loop and discard events (or, if not on the main thread, just sleep the thread)
// until all running and delayed fetchers have completed.
//
// This is only for use in testing or in tools without a user interface.
//
// Synchronous fetches should never be done by shipping apps; they are
// sufficient reason for rejection from the app store.

Given that comment, it seems we might need a more significant refactor than most warnings require.

cc: @thomasvl

morganchen12:
Fixes #8294, probably.

It's unclear to me exactly what the original implementation did, though it's likely it [slept](https://github.com/google/gtm-session-fetcher/blob/main/Source/GTMSessionFetcher.m#L2055) the thread instead of allowing other closures to execute first.

### DIFF
--- a/FirebaseStorage/Sources/FIRStorageDownloadTask.m
+++ b/FirebaseStorage/Sources/FIRStorageDownloadTask.m
@@ -169,11 +169,11 @@
     strongSelf.state = FIRStorageTaskStatePausing;
     [strongSelf.fetcher stopFetching];
     // Give the resume callback a chance to run (if scheduled)
-    [strongSelf dispatchAsync:^{
+    dispatch_after(0.001 * NSEC_PER_SEC, strongSelf.dispatchQueue, ^{
       strongSelf.state = FIRStorageTaskStatePaused;
       FIRStorageTaskSnapshot *snapshot = weakSelf.snapshot;
       [strongSelf fireHandlersForStatus:FIRStorageTaskStatusPause snapshot:snapshot];
-    }];
+    });
   }];
 }
 
--- a/FirebaseStorage/Sources/FIRStorageDownloadTask.m
+++ b/FirebaseStorage/Sources/FIRStorageDownloadTask.m
@@ -165,13 +165,15 @@
 - (void)pause {
   __weak FIRStorageDownloadTask *weakSelf = self;
   [self dispatchAsync:^() {
-    weakSelf.state = FIRStorageTaskStatePausing;
-    [weakSelf.fetcher stopFetching];
+    __strong FIRStorageDownloadTask *strongSelf = weakSelf;
+    strongSelf.state = FIRStorageTaskStatePausing;
+    [strongSelf.fetcher stopFetching];
     // Give the resume callback a chance to run (if scheduled)
-    [weakSelf.fetcher waitForCompletionWithTimeout:0.001];
-    weakSelf.state = FIRStorageTaskStatePaused;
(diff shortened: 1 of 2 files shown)


paulb777:
Restarted CI that failed because of the GoogleUtilities 7.4.2 to 7.4.3 transition

paulb777 (approved):
LGTM on green

paulb777:
Thanks for addressing Morgan.

Please investigate the integration test failure - https://github.com/firebase/firebase-ios-sdk/pull/8296/checks?check_run_id=2900576792

FirebaseStorage/Sources/FIRStorageDownloadTask.m
```diff
@@ -165,13 +165,15 @@ - (void)cancelWithError:(NSError *)error {
...
-    [weakSelf.fetcher stopFetching];
+    __strong FIRStorageDownloadTask *strongSelf = weakSelf;
+    strongSelf.state = FIRStorageTaskStatePausing;
+    [strongSelf.fetcher stopFetching];
     // Give the resume callback a chance to run (if scheduled)
-    [weakSelf.fetcher waitForCompletionWithTimeout:0.001];
-    weakSelf.state = FIRStorageTaskStatePaused;
-    FIRStorageTaskSnapshot *snapshot = weakSelf.snapshot;
-    [weakSelf fireHandlersForStatus:FIRStorageTaskStatusPause snapshot:snapshot];
+    dispatch_after(0.001 * NSEC_PER_SEC, strongSelf.dispatchQueue, ^{
```
schmidt-sebastian, line 172: Would a simple call to dispatch a new block on the queue be enough here? The dispatch queue is configured to run tasks in order.

FirebaseStorage/Sources/FIRStorageDownloadTask.m
```diff
@@ -165,13 +165,15 @@ - (void)cancelWithError:(NSError *)error {
...
+      strongSelf.state = FIRStorageTaskStatePaused;
+      FIRStorageTaskSnapshot *snapshot = weakSelf.snapshot;
```
paulb777, line 174: Should this be strongSelf?

morganchen12:
I think the fix doesn't work because it's unable to invoke dispatch_group_wait as the original method does [here](https://github.com/google/gtm-session-fetcher/blob/main/Source/GTMSessionFetcher.m#L2034), which is causing the callbacks to not be executed.

FirebaseStorage/Sources/FIRStorageDownloadTask.m
morganchen12: Fixed this so it's not as silly. It turns out the `pause` method needs to wait for GTMSessionFetcher to finish its last update, or the internal state of the download task will be incorrect.

FirebaseStorage/Sources/FIRStorageDownloadTask.m
morganchen12: Fixed.

PR #8309, opened 2021-06-26
Add script to check for absl reference from Firestore/core/api

wu-hui:
Also add exception to check_imports.swift to exclude cmake build directories.

### DIFF
--- a/scripts/check_firestore_core_api_absl.py
+++ b/scripts/check_firestore_core_api_absl.py
@@ -42,8 +42,9 @@ def diff_with_absl(revision, patterns):
 
 
 def main():
+  patterns = ['Firestore/core/src/api/*.h']
   parser = argparse.ArgumentParser(
-      description='Check Absl usaging in firestore/core/api')
+      description='Check Absl usage in %s' % patterns)
   parser.add_argument(
       '--dry-run',
       '-n',
@@ -63,7 +64,6 @@ def main():
     _logger.setLevel(logging.DEBUG)
 
   revision = 'origin/master' if not args.rev else args.rev
-  patterns = ['Firestore/core/src/api/*.h']
   _logger.debug('Checking %s absl usage against %s' %
                 (patterns, revision))
   diff = diff_with_absl(revision, patterns)
--- a/scripts/check_firestore_core_api_absl.py
+++ b/scripts/check_firestore_core_api_absl.py
@@ -34,16 +34,16 @@ def diff_with_absl(revision, patterns):
   """Finds diffs containing 'absl' since a revision from specified path
   pattern.
   """
-
   # git command to print all diffs that has 'absl' in it.
-  command = ['git', 'diff', '-z', '-G', 'absl', revision, '--']
+  command = ['git', 'diff', '-G', 'absl', revision, '--']
   command.extend(patterns)
-  command_trace.log(command)
+  _logger.debug(command)
   return six.ensure_text(subprocess.check_output(command))
 
 
 def main():
-  parser = argparse.ArgumentParser(description='Lint source files.')
+  parser = argparse.ArgumentParser(
+      description='Check Absl usaging in firestore/core/api')
   parser.add_argument(
       '--dry-run',
       '-n',
@@ -60,29 +60,25 @@ def main():
   dry_run = False
   if args.dry_run:
     dry_run = True
-    command_trace.enable_tracing()
+    _logger.setLevel(logging.DEBUG)
 
-  revision = 'origin/master' if (not args.rev) else args.rev
-  _logger.debug('Checking Firestore/core/src/api absl usage against %s' %
-                revision)
-  diff = diff_with_absl(revision, ['Firestore/core/src/api/*.h'])
+  revision = 'origin/master' if not args.rev else args.rev
+  patterns = ['Firestore/core/src/api/*.h']
+  _logger.debug('Checking %s absl usage against %s' %
+                (patterns, revision))
+  diff = diff_with_absl(revision, patterns)
 
   # Check for changes adding new absl references only.
-  found = False
-  for line in diff.splitlines():
-    # Additions start with '+'
-    if line.startswith('+') and line.find('absl::') >= 0:
-      # Found a change introducing a reference to absl
+  lines = [line for line in diff.splitlines()
+           if line.startswith('+') and 'absl::' in line]
+  if lines:
       _logger.error(
-          'Found a change introducing reference to absl under'
-          'Firestore/core/api:'
-      )
-      _logger.error('  %s' % line)
-      _logger.error('')
-      found = True
-
-  if found and (not dry_run):
-    exit(-1)
+          'Found a change introducing reference to absl under %s'
+          % patterns)
+      for line in lines:
+          _logger.error(' %s' % line)
+      if not dry_run:
+          exit(-1)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+  _logger.debug('Checking Firestore/core/src/api absl usage against %s' %
+                revision)
+  diff = diff_with_absl(revision, ['Firestore/core/src/api/*.h'])
+
+  # Check for changes adding new absl references only.
+  found = False
+  for line in diff.splitlines():
+    # Additions start with '+'
+    if line.startswith('+') and line.find('absl::') >= 0:
+      # Found a change introducing a reference to absl
```
var-const, line 75: Nit: this comment repeats the message from the next line, I'd omit it.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+  command.extend(patterns)
+  command_trace.log(command)
+  return six.ensure_text(subprocess.check_output(command))
+
+
+def main():
+  parser = argparse.ArgumentParser(description='Lint source files.')
+  parser.add_argument(
+      '--dry-run',
+      '-n',
```
var-const, line 49: Question: what is the use for dry run? From what I can see, the main difference is that it always returns a successful code upon exit.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+      nargs='?',
+      help='A single revision that specifies a point in time '
+      'from which to look for changes. Defaults to '
+      'origin/master.')
+  args = command_trace.parse_args(parser)
+
+  dry_run = False
+  if args.dry_run:
+    dry_run = True
+    command_trace.enable_tracing()
```
var-const, line 63: Question: what does this command do?

var-const (changes requested):
Overall looking good, just some nitpicks.

scripts/check_firestore_core_api_absl.py
wu-hui: It just sets the log level to debug.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+
+  revision = 'origin/master' if (not args.rev) else args.rev
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+  _logger.debug('Checking Firestore/core/src/api absl usage against %s' %
+                revision)
+  diff = diff_with_absl(revision, ['Firestore/core/src/api/*.h'])
+
+  # Check for changes adding new absl references only.
+  found = False
+  for line in diff.splitlines():
+    # Additions start with '+'
+    if line.startswith('+') and line.find('absl::') >= 0:
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+      # Found a change introducing a reference to absl
+      _logger.error(
+          'Found a change introducing reference to absl under'
+          'Firestore/core/api:'
+      )
+      _logger.error('  %s' % line)
+      _logger.error('')
+      found = True
+
+  if found and (not dry_run):
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+  for line in diff.splitlines():
+    # Additions start with '+'
+    if line.startswith('+') and line.find('absl::') >= 0:
+      # Found a change introducing a reference to absl
+      _logger.error(
+          'Found a change introducing reference to absl under'
+          'Firestore/core/api:'
+      )
+      _logger.error('  %s' % line)
+      _logger.error('')
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+      'origin/master.')
+  args = command_trace.parse_args(parser)
+
+  dry_run = False
+  if args.dry_run:
+    dry_run = True
+    command_trace.enable_tracing()
+
+  revision = 'origin/master' if (not args.rev) else args.rev
+  _logger.debug('Checking Firestore/core/src/api absl usage against %s' %
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -0,0 +1,89 @@
...
+                revision)
+  diff = diff_with_absl(revision, ['Firestore/core/src/api/*.h'])
+
+  # Check for changes adding new absl references only.
+  found = False
+  for line in diff.splitlines():
+    # Additions start with '+'
+    if line.startswith('+') and line.find('absl::') >= 0:
+      # Found a change introducing a reference to absl
+      _logger.error(
```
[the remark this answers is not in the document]
wu-hui: Done.

scripts/check_firestore_core_api_absl.py
```diff
@@ -34,16 +34,16 @@ def diff_with_absl(revision, patterns):
...
   command.extend(patterns)
-  command_trace.log(command)
+  _logger.debug(command)
   return six.ensure_text(subprocess.check_output(command))
 
 
 def main():
-  parser = argparse.ArgumentParser(description='Lint source files.')
+  parser = argparse.ArgumentParser(
+      description='Check Absl usaging in firestore/core/api')
```
var-const, line 46: Nits:
- `s/usaging/usage/`;
- please use the `patterns` variable instead of hardcoding `api/` here.

var-const (commented):
Question: is this possible to make it so that if this script fails, the failure is visible but the PR can still be merged? It's ok to add Abseil to the private section of a class in `api/`, but this script would still fail because it doesn't distinguish between `public` and `private` (and trying to do that would be more trouble than it's worth, IMO).

var-const:
Mostly LGTM, but need to resolve the question about whether this check blocks/should block merging.

scripts/check_firestore_core_api_absl.py
wu-hui: Done.

wu-hui:
> Question: is this possible to make it so that if this script fails, the failure is visible but the PR can still be merged? It's ok to add Abseil to the private section of a class in `api/`, but this script would still fail because it doesn't distinguish between `public` and `private` (and trying to do that would be more trouble than it's worth, IMO).

We could change the call with `--dry-run` in `check.sh`. But we can wait until such a case did happen: when someone tried to add absl in private. I doubt this would happen before we remove this check.

var-const (approved):

var-const:
@wu-hui Can you please do a test PR that adds Abseil to the `api` folder after you merge this? Just to make sure that the check fails and to take a look at the script output.

var-const:
@wu-hui Andy, is this ready to merge? Please let me know if you need anything from me.

Issue #8310, opened 2021-06-27
'Firebase/AppCheck' subspec doesn't work for macOS / tvOS


jakekrog:
### Step 1: Describe your environment

  * Xcode version: 12.5.1
  * Firebase SDK version: 8.2.0
  * Installation method: CocoaPods
  * Firebase Component: AppCheck

### Step 2: Describe the problem

`pod 'Firebase/AppCheck` does not work for macOS or tvOS targets, but `pod 'FirebaseAppCheck'` does.

#### Steps to reproduce:

Link `Firebase/AppCheck` to a macOS / tvOS target that uses `AppCheckDebugProviderFactory` and build.

PR #8311, opened 2021-06-27
Update AppCheck subspec dependency to support macOS / tvOS

jakekrog:
Fixes https://github.com/firebase/firebase-ios-sdk/issues/8310


paulb777 (approved):
Thanks for the fix!

PR #8265, opened 2021-06-16
Re-enable HTTPSCallableTests

morganchen12:
Fixes #8261.

PR #8316, opened 2021-06-29
Update base commit of code coverage

granluo:
Thanks Yifan for pointing out the issue on the test coverage workflow
The code coverage is comparing a merge commit to the common commit of a PR and the target branch. This will involve changes between this common commit and the latest commit of the target branch. This update will switch the common commit of the PR to the latest commit on the target branch.

Update indent of the workflow.

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -39,7 +39,7 @@ jobs:
           pr_branch: ${{ github.event.pull_request.head.ref }}
         run: |
           if [ ! -z "${{ env.METRICS_SERVICE_SECRET }}"  ]; then
-          ./scripts/code_coverage_report/get_updated_files.sh
+            ./scripts/code_coverage_report/get_updated_files.sh
           fi
 
   pod-lib-lint-abtesting:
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Good find. Thanks!

yifanyang (approved):

yifanyang:
**TLDR:** this is to avoid problems, for example, in #8267 where the report of coverage diff contained changes for `FirebaseFunctions` and `FirebasePerformance`, which is unexpected since the PR didn't contain any code change related to them.

The diff for `FirebaseFunctions` and `FirebasePerformance` in #8267 actually comes from #8194 and #8265 respectively. The test for PR #8267 ran on the commit [`d3b4bd`](https://github.com/firebase/firebase-ios-sdk/commit/d3b4bdcce90d9a682401eb9361c3dec112b8e6b8) (as shown in GitHub Actions [log](https://github.com/firebase/firebase-ios-sdk/runs/2844571388#step:2:557)), which merged 48df806 into 3ab732b. If we compared [`d3b4bd`](https://github.com/firebase/firebase-ios-sdk/commit/d3b4bdcce90d9a682401eb9361c3dec112b8e6b8) with 3ab732b, the resulted diff would be reflecting precisely the code change contained in the PR. However, as in the report we compared against 76d0144, an ancestor of 3ab732b, the diff included code change between 76d0144 and 3ab732b (outside of #8267), as listed in [commit history](https://github.com/firebase/firebase-ios-sdk/commits/master).

48df806: the last commit pushed by the author to the PR #8267
3ab732b: the last commit on merge target branch, in our case `master`, picked by GitHub Actions at the time of CI run

@paulb777 @ryanwilson 
One more problem in #8267 is that, the diff report also contains some number for `FirebaseDatabase`, `FirebaseDynamicLinks` and `FirebaseFirestore`. I suspect that these diff actually comes from flaky test execution, meaning even though the same test passed for both runs, but the code path executed were different, which resulted in different coverage number calculated. Do you know how Gran and I can deal with it?

paulb777:
@yifanyang Thanks for investigating. Some of the flakes are due to networking in the tests that is not deterministic run to run. We would need to update the tests to mock the networking so they become consistent.

PR #8317, opened 2021-06-29
Update storage workflow

granluo:
Remove the duplicated flag.


paulb777:
Where is the duplicate?

granluo:
> Where is the duplicate?
The matrix in the workflow has the `--skip-test`
https://github.com/firebase/firebase-ios-sdk/blob/ebdca2e2b43b2b6cc7b19d288e4369f191474e54/.github/workflows/storage.yml#L175

paulb777:
I see. Thanks.

paulb777 (approved):

PR #8328, opened 2021-06-30
update crashlytics qs test with new dir

morganchen12:


morganchen12:
@paulb777 do you know what's causing the unit test failures off the top of your head? I can dig deeper tomorrow otherwise.

paulb777:
It looks like `ruby ../scripts/info_script.rb "${SAMPLE}" "${LEGACY:-}"` is not adding GoogleService-Info.plist to the project, but I'm not sure why.

paulb777:
I'm not following the paths between running info_script in the qs repo versus this repo. In the quickstart repo, there is a `cd` to the legacy directory before running - but then the script adjusts the path here - https://github.com/firebase/quickstart-ios/blame/master/scripts/info_script.rb#L23?

morganchen12:
I fixed the path issue, but now the swift sample fails to compile because it looks like the pods aren't properly installed.

However, I'm not able to find any errors in the setup.

paulb777:
One problem that may or may not be related is the Legacy xcode project has many multiple targets matching warnings because it still has SPM configured.

Perhaps, restore the old Xcode project to the Legacy directory?

morganchen12:
Fixed with https://github.com/firebase/quickstart-ios/pull/1203. Still no clue why the top-level Firebase module refused to import.

paulb777:
Weird - it might be worth diving deeper at some point to figure it out - but I'm good to merge here for now.

paulb777 (approved):

PR #8352, opened 2021-07-08
Core API build tests

ncooke3:
#no-changelog

### DIFF
--- a/FirebaseCore/Tests/SwiftUnit/CoreAPITests.swift
+++ b/FirebaseCore/Tests/SwiftUnit/CoreAPITests.swift
@@ -92,13 +92,12 @@ final class CoreAPITests {
       break
     case FirebaseLoggerLevel.debug:
       break
-    case FirebaseLoggerLevel.min:
-      break
-    case FirebaseLoggerLevel.max:
-      break
     default:
       break
     }
+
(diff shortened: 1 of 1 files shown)


FirebaseCore/Tests/SwiftUnit/CoreAPITests.swift
```diff
@@ -0,0 +1,163 @@
...
+      break
+    case FirebaseLoggerLevel.info:
+      break
+    case FirebaseLoggerLevel.debug:
+      break
+    default:
+      break
+    }
+
+    _ = FirebaseLoggerLevel.min
```
paulb777, line 99: Is it Swift version dependent whether these can be in the switch?

paulb777 (approved):

FirebaseCore/Tests/SwiftUnit/CoreAPITests.swift
ncooke3: I don't believe so. Including them in the switch led to a [warning which failed CI](https://github.com/firebase/firebase-ios-sdk/runs/3015446535?check_suite_focus=true#step:4:32). The warning was that the `min` and `max` case would never be executed. I think it's just because `min` and `max` point to enum cases that the switch would always catch first.

PR #8354, opened 2021-07-08
Fix API availability in AppCheck API build tests

ncooke3:
I realized I labeled the API availability incorrectly for these tests. AppCheck is iOS only at the moment.

#no-changelog


paulb777:
The podspec shows osx and tvos support: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAppCheck.podspec#L25

PR #8365, opened 2021-07-09
Bump Promises dependency

### LINKED ISSUE #8334: Upgrade Promise dependency before Xcode 13 is released

  * Xcode version: 13 beta 2
  * Firebase SDK version: 8.2.0
  * Firebase Component: Core

The current version of Xcode 13 (beta 2) have a change to the swift compiler which makes `await` into a keyword. It seems this change is not guarded be a feature flag, so anything compiled with Xcode 13 (even if not targeting new versions) will not be able to use `await` as a function name (See [0296-async](https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#source-compatibility) ).

This conflicts with [google/promise](https://github.com/google/promises) which Firebase depends on (through `GoogleUtilities/Environment`).

`Google/Promises` have been updated with a fix that renames `await`, but required a major release (v2.0.0).

Since Firebase is still depending on the old version, any project that use both Firebase and Promise will not be able to compile with Xcode 13 since there's no way to allow two versions of Promise in one project.

#### Steps to reproduce:

1. Create a project that Use both Firebase and Promise and compile it with Xcode 13
2. Trying to use the old Promise will result in compile errors
3. Updating to use Promise 2 is not possible because of firebase dependencies:
```
        FirebaseCore (~> 8.0) was resolved to 8.2.0, which depends on
          GoogleUtilities/Environment (~> 7.4) was resolved to 7.4.3, which depends on
            PromisesObjC (~> 1.2)
```

ncooke3:
Plan is to:
- merge this PR
- wait for clean nightlies
- then publish GDT & GUL minor releases

Fixes #8334

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -133,7 +133,7 @@ let package = Package(
     .package(
       name: "Promises",
       url: "https://github.com/google/promises.git",
-      "1.2.8" ...< "2.0.0"
+      "1.2.8" ... "2.0.0"
     ),
     .package(
       name: "SwiftProtobuf",
--- a/Package.swift
+++ b/Package.swift
@@ -153,8 +153,7 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      .revision("8982d5f4df54fd15599287fdb4f77a8be9604221")
-//      "7.4.0" ..< "8.0.0"
+      "7.5.0" ..< "8.0.0"
     ),
     .package(
       name: "GTMSessionFetcher",
--- a/Package.swift
+++ b/Package.swift
@@ -153,7 +153,8 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      "7.4.0" ..< "8.0.0"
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -153,7 +153,7 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      "7.4.0" ..< "8.0.0"
+      "7.5.0" ..< "8.0.0"
     ),
```
ncooke3, line 157: GoogleUtilities should bump to its new minor right?

paulb777:
Test the podspec changes first, to avoid the need to prematurely tag the GoogleUtilities and GTM repos.

The podspecs should be updated.

Bumping the GoogleUtilities minimum dependency is fine, but not required since the firebase-ios-sdk SPM should still work with 7.4.x ....

ncooke3:
@paulb777 So with the specs bumped and CI green,  GDT and GULs are good to publish, correct?

paulb777:
Let's wait for a clean nightly run before publishing GDT and GoogleUtilities.

This is good to merge after adding a CHANGELOG update for 8.4.0

paulb777 (approved):

PR #8366, opened 2021-07-09
Add tests for Update methods with empty dictionaries.

ehsannas:
The Update methods (document update, batch update, transaction update) are allowed to take an empty dictionary -- in which case they do not apply any updates to the document.


ehsannas:
How does adding tests lower the coverage percentage? 😄

PR #8369, opened 2021-07-09
Fix nightly failures related to Swift API tests

ncooke3:


.github/workflows/functions.yml
```diff
@@ -113,7 +113,8 @@ jobs:
         target: [ios, tvos, macos]
         flags: [
           '--use-static-frameworks',
-          '--use-libraries'
+          '--use-libraries',
```
paulb777, line 116: Should be `'--use-libraries --skip-tests'`

PR #8371, opened 2021-07-09
Create nanopb options file.

jeremyjiang-dev:
Note: I did not remove Android/Web related fields in the option file. The reason is that these field will be created in the anyway based on `perf_metric.proto`. If I do not define these fields in the option file, all iOS field will be `FT_POINTER` but all Android/web field will be `FT_callback`, causing inconsistency.


FirebasePerformance.podspec
```diff
@@ -28,7 +28,7 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
 
   base_dir = "FirebasePerformance/"
   s.source_files = [
-    base_dir + 'Sources/**/*.[mh]',
+    base_dir + 'Sources/**/*.[cmh]',
     base_dir + 'ProtoSupport/**/*.[mh]',
```
paulb777, line 32: Can the Protobuf dependency be removed below?

paulb777 (approved):
LGTM

FirebasePerformance.podspec
visumickey: We need to remove that. But that will be done after all the Protobuf dependencies are removed. For now, we will retain them until the last PR for the NanoPb conversion gets submitted.

visumickey (approved):

FirebasePerformance.podspec
jeremyjiang-dev: Good catch! I will remove this and all Protobuf files in another PR.

PR #8372, opened 2021-07-09
Fix the handling of Unicode characters in log and assertion messages

var-const:
Currently, any UTF-8 characters outside the ascii set may be encoded incorrectly.

### DIFF
--- a/Firestore/core/src/util/exception_apple.mm
+++ b/Firestore/core/src/util/exception_apple.mm
@@ -60,8 +60,8 @@ ABSL_ATTRIBUTE_NORETURN void ObjcThrowHandler(ExceptionType type,
         handleFailureInFunction:MakeNSString(func)
                            file:MakeNSString(file)
                      lineNumber:line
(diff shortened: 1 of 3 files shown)


Firestore/core/src/util/exception_apple.mm
```diff
@@ -60,8 +60,8 @@ ABSL_ATTRIBUTE_NORETURN void ObjcThrowHandler(ExceptionType type,
         handleFailureInFunction:MakeNSString(func)
                            file:MakeNSString(file)
                      lineNumber:line
-                    description:@"%@: %s", ExceptionName(type),
-                                message.c_str()];
+                    description:@"%@: %@", ExceptionName(type),
+                                MakeNSString(message)];
```
var-const, line 64: `MakeNSString` specifies encoding [as UTF-8](https://github.com/firebase/firebase-ios-sdk/blob/d50dd0a5575331be187596cc975a856b59c88615/Firestore/core/src/util/string_apple.h#L65).

dconeybe (approved):

PR #8378, opened 2021-07-12
Cherrypick 8.3.1 SPM patch into master

ncooke3:
Cherrypicking latest [8.3.1 patch commit](../commit/ed87539cda9cfb3a3de1be22354c6e74760db91b) from [`release-8.3`](../commits/release-8.3) branch to keep `master` in sync for M100.

#no-changelog


paulb777 (approved):
Technically should be a merge, not a cherry-pick.

They should be exactly the same this time so ok with it like this.

PR #8381, opened 2021-07-13
Convert ApplicatoinInfo class to nanopb object.

jeremyjiang-dev:
ApplicationInfo class conversion to nanopb object is completed.
Other classes are WIP.


FirebasePerformance/Sources/FPRProtoUtils.h
```diff
@@ -23,22 +23,35 @@
 #import "FirebasePerformance/Sources/Public/FIRTrace.h"
 
 #import "FirebasePerformance/ProtoSupport/PerfMetric.pbobjc.h"
+#import "FirebasePerformance/Sources/Protogen/nanopb/perf_metric.nanopb.h"
```
visumickey, line 26: Instead of modifying the FPRProtoUtils, can we create something like FPRNanoPbUtils and make the changes there? This way we can make sure we are not disrupting the FPRProtoUtils and we have a better way to validate if the new conversion is working as expected.

This would also enable testing separately over disrupting FPRProtoUtils. With a separate FPRNanoPbUtils, the changes to FPRClient can be done later too. We can testing everything related to NanoPb conversion in our unit tests before modifying the consumer of NanoPbUtils.

jeremyjiang-dev: Great suggestion! I created https://github.com/firebase/firebase-ios-sdk/pull/8387. PTAL

PR #8387, opened 2021-07-14
Convert Objective-C Classes to Nanopb

jeremyjiang-dev:
### How to review this PR?
Similar to how Objective-C classes are converted to protos, this PR converts Objective-C classes to Nanopb struct.
Please refer to [FPRProtoUtils.h](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/Sources/FPRProtoUtils.h), [FPRProtoUtils.m](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/Sources/FPRProtoUtils.m), and [FPRProtoUtilsTest.m](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/Tests/Unit/FPRProtoUtilsTest.m)

### DIFF
--- a/FirebasePerformance/Sources/FPRClient.m
+++ b/FirebasePerformance/Sources/FPRClient.m
@@ -209,8 +209,8 @@
     return;
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
-    FPRMSGNetworkRequestMetric *networkRequestMetric = FPRGetNetworkRequestMetric(trace);
-    if (networkRequestMetric) {
+    if (trace.hasValidResponseCode) {
+      FPRMSGNetworkRequestMetric *networkRequestMetric = FPRGetNetworkRequestMetric(trace);
       int64_t duration = networkRequestMetric.hasTimeToResponseCompletedUs
                              ? networkRequestMetric.timeToResponseCompletedUs
                              : 0;
@@ -237,7 +237,10 @@
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
     FPRMSGPerfMetric *metric = FPRGetPerfMetricMessage(self.config.appID);
-    FPRMSGGaugeMetric *gaugeMetric = FPRGetGaugeMetric(gaugeData, sessionId);
+    FPRMSGGaugeMetric *gaugeMetric = nil;
+    if ((gaugeData != nil && gaugeData.count != 0) && (sessionId != nil && sessionId.length != 0)) {
+      gaugeMetric = FPRGetGaugeMetric(gaugeData, sessionId);
+    }
     metric.gaugeMetric = gaugeMetric;
     [self processAndLogEvent:metric];
   });
--- a/FirebasePerformance/Sources/FPRClient.m
+++ b/FirebasePerformance/Sources/FPRClient.m
@@ -209,7 +209,7 @@
     return;
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
-    if (trace.hasValidResponseCode) {
+    if ([trace isValid]) {
       FPRMSGNetworkRequestMetric *networkRequestMetric = FPRGetNetworkRequestMetric(trace);
       int64_t duration = networkRequestMetric.hasTimeToResponseCompletedUs
                              ? networkRequestMetric.timeToResponseCompletedUs
--- a/FirebasePerformance/Sources/FPRNanoPbUtils.h
+++ b/FirebasePerformance/Sources/FPRNanoPbUtils.h
@@ -24,33 +24,83 @@
 
 #import "FirebasePerformance/Sources/Protogen/nanopb/perf_metric.nanopb.h"
 
+/**nanopb struct of encoded NSDictionary<NSString *, NSString *>.*/
 typedef struct {
   pb_bytes_array_t* _Nonnull key;
   pb_bytes_array_t* _Nonnull value;
 } StringToStringMap;
 
+/**nanopb struct of encoded NSDictionary<NSString *, NSNumber *>.*/
 typedef struct {
   pb_bytes_array_t* _Nonnull key;
   bool has_value;
   int64_t value;
 } StringToNumberMap;
 
+/** Callocs a pb_bytes_array and copies the given NSData bytes into the bytes array.
+ *
+ * @note Memory needs to be free manually, through pb_free or pb_release.
+ * @param data The data to copy into the new bytes array.
+ * @return pb_byte array
+ */
 extern pb_bytes_array_t* _Nullable FPREncodeData(NSData* _Nonnull data);
 
+/** Callocs a pb_bytes_array and copies the given NSString's bytes into the bytes array.
+ *
+ * @note Memory needs to be free manually, through pb_free or pb_release.
+ * @param string The string to encode as pb_bytes.
+ * @return pb_byte array
+ */
 extern pb_bytes_array_t* _Nullable FPREncodeString(NSString* _Nonnull string);
 
+/** Creates a NSData object by copying the given bytes array and returns the reference.
+ *
+ * @param pbData The pbData to dedoded as NSData
+ * @return A reference to NSData
+ */
 extern NSData* _Nullable FPRDecodeData(pb_bytes_array_t* _Nonnull pbData);
 
+/** Creates a NSString object by copying the given bytes array and returns the reference.
+ *
+ * @param pbData The pbData to dedoded as NSString
+ * @return A reference to the NSString
+ */
 extern NSString* _Nullable FPRDecodeString(pb_bytes_array_t* _Nonnull pbData);
 
+/** Creates a NSDictionary by copying the given bytes from the StringToStringMap object and returns
+ * the reference.
+ *
+ * @param map The reference to a StringToStringMap object to be decoded.
+ * @param count The number of entries in the dictionary.
+ * @return A reference to the dictionary
+ */
 extern NSDictionary<NSString*, NSString*>* _Nullable FPRDecodeStringToStringMap(
     StringToStringMap* _Nullable map, NSInteger count);
 
+/** Callocs a nanopb StringToStringMap and copies the given NSDictionary bytes into the
+ * StringToStringMap.
+ *
+ * @param dict The dict to copy into the new StringToStringMap.
+ * @return A reference to StringToStringMap
+ */
 extern StringToStringMap* _Nullable FPREncodeStringToStringMap(NSDictionary* _Nullable dict);
 
+/** Creates a NSDictionary by copying the given bytes from the StringToNumberMap object and returns
+ * the reference.
+ *
+ * @param map The reference to a StringToNumberMap object to be decoded.
+ * @param count The number of entries in the dictionary.
+ * @return A reference to the dictionary
+ */
 extern NSDictionary<NSString*, NSNumber*>* _Nullable FPRDecodeStringToNumberMap(
     StringToNumberMap* _Nullable map, NSInteger count);
 
+/** Callocs a nanopb StringToNumberMap and copies the given NSDictionary bytes into the
+ * StringToStringMap.
+ *
+ * @param dict The dict to copy into the new StringToNumberMap.
+ * @return A reference to StringToNumberMap
+ */
 extern StringToNumberMap* _Nullable FPREncodeStringToNumberMap(NSDictionary* _Nullable dict);
 
 /** Creates a new firebase_perf_v1_PerfMetric struct populated with system metadata.
--- a/FirebasePerformance/Sources/FPRNanoPbUtils.h
+++ b/FirebasePerformance/Sources/FPRNanoPbUtils.h
@@ -1,0 +1,94 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <TargetConditionals.h>
+#if __has_include("CoreTelephony/CTTelephonyNetworkInfo.h") && !TARGET_OS_MACCATALYST
+#define TARGET_HAS_MOBILE_CONNECTIVITY
+#import <CoreTelephony/CTTelephonyNetworkInfo.h>
+#endif
+
+#import "FirebasePerformance/Sources/AppActivity/FPRTraceBackgroundActivityTracker.h"
+#import "FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.h"
+#import "FirebasePerformance/Sources/Public/FIRTrace.h"
+
+#import "FirebasePerformance/Sources/Protogen/nanopb/perf_metric.nanopb.h"
+
+typedef struct {
+  pb_bytes_array_t* _Nonnull key;
+  pb_bytes_array_t* _Nonnull value;
+} StringToStringMap;
+
+typedef struct {
+  pb_bytes_array_t* _Nonnull key;
+  bool has_value;
+  int64_t value;
+} StringToNumberMap;
+
+extern pb_bytes_array_t* _Nullable FPREncodeData(NSData* _Nonnull data);
+
+extern pb_bytes_array_t* _Nullable FPREncodeString(NSString* _Nonnull string);
+
+extern NSData* _Nullable FPRDecodeData(pb_bytes_array_t* _Nonnull pbData);
+
+extern NSString* _Nullable FPRDecodeString(pb_bytes_array_t* _Nonnull pbData);
+
+extern NSDictionary<NSString*, NSString*>* _Nullable FPRDecodeStringToStringMap(
+    StringToStringMap* _Nullable map, NSInteger count);
+
+extern StringToStringMap* _Nullable FPREncodeStringToStringMap(NSDictionary* _Nullable dict);
+
+extern NSDictionary<NSString*, NSNumber*>* _Nullable FPRDecodeStringToNumberMap(
+    StringToNumberMap* _Nullable map, NSInteger count);
+
+extern StringToNumberMap* _Nullable FPREncodeStringToNumberMap(NSDictionary* _Nullable dict);
+
+/** Creates a new firebase_perf_v1_PerfMetric struct populated with system metadata.
+ *  @param appID The Google app id to put into the message
+ *  @return A firebase_perf_v1_PerfMetric struct.
+ */
+extern firebase_perf_v1_PerfMetric GetPerfMetricMessage(NSString* _Nonnull appID);
+
+/** Creates a new firebase_perf_v1_ApplicationInfo struct populated with system metadata.
+ *  @return A firebase_perf_v1_ApplicationInfo struct.
+ */
+extern firebase_perf_v1_ApplicationInfo GetApplicationInfoMessage(void);
+
+/** Converts the FIRTrace object to a firebase_perf_v1_TraceMetric struct.
+ *  @return A firebase_perf_v1_TraceMetric struct.
+ */
+extern firebase_perf_v1_TraceMetric GetTraceMetric(FIRTrace* _Nonnull trace);
+
+/** Converts the FPRNetworkTrace object to a firebase_perf_v1_NetworkRequestMetric struct.
+ *  @return A firebase_perf_v1_NetworkRequestMetric struct.
+ */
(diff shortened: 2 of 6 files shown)


FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+NSMutableDictionary<NSString *, NSString *> *FPRDecodeCustomAttributes(
+    struct _firebase_perf_v1_ApplicationInfo_CustomAttributesEntry *customAttributes,
+    NSInteger count) {
+  NSMutableDictionary<NSString *, NSString *> *dict = [NSMutableDictionary dictionary];
+  for (int i = 0; i < count; i++) {
+    NSString *key = FPRDecodeString(customAttributes[i].key);
+    NSString *value = FPRDecodeString(customAttributes[i].value);
+    dict[key] = value;
+  }
+  return dict;
```
visumickey, line 78: can you return an immutable copy here? [dict copy]. Unless required keeping the object immutable would enable a lot of predictability to the code.

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+}
+
+#pragma mark - Public methods
+
+firebase_perf_v1_PerfMetric GetPerfMetricMessage(NSString *appID) {
+  firebase_perf_v1_PerfMetric perfMetricMessage = firebase_perf_v1_PerfMetric_init_default;
```
visumickey, line 84: Do you know the difference between init_default and init_zero? I'm curious on how default handles the optional fields. We do not want to automatically set values to zero, when there are no values set.

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+    iosAppInfo.mcc_mnc = FPREncodeString(mccMnc);
+  }
+  if (iosAppInfo.network_connection_info.network_type ==
+      firebase_perf_v1_NetworkConnectionInfo_NetworkType_MOBILE) {
+    iosAppInfo.network_connection_info.mobile_subtype = FPRCellularNetworkType();
+  }
+#endif
+  appInfoMessage.ios_app_info = iosAppInfo;
+
+  NSDictionary<NSString *, NSString *> *attributes =
```
visumickey, line 113: Similar to decodeCustomAttributes, can you create a method as encodeCustomAttributes?

Or even better if we could generalize that as "encodeMap" and "decodeMap". WDYT?

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+  // TODO(visum) Enable subtraces
+  //  NSMutableArray<FPRMSGTraceMetric *> *subtraces = [[NSMutableArray alloc] init];
+  //  [trace.stages
+  //      enumerateObjectsUsingBlock:^(FIRTrace *_Nonnull stage, NSUInteger idx, BOOL *_Nonnull
+  //      stop) {
+  //        [subtraces addObject:FPRGetTraceMetric(stage)];
+  //      }];
+  //  traceMetric.subtracesArray = subtraces;
+
+  // TODO(visum) Enable custom attributes
```
visumickey, line 154: The next couple of lines can be cleaned up.

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+  //      processState = FPRMSGApplicationProcessState_Background;
+  //      break;
+  //
+  //    case FPRTraceStateBackgroundAndForeground:
+  //      processState = FPRMSGApplicationProcessState_ForegroundBackground;
+  //      break;
+  //
+  //    default:
+  //      break;
+  //  }
```
visumickey, line 309: Should these be commented out? It looks like we were able to bring in ENUMs. Can we respond back with the proper NanoPb structure for this data?

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,452 @@
...
+  return networkInfo;
+}
+#endif
+
+#pragma mark - Proto creation utilities
+
+/** Converts a dictionary of <NSString *, NSNumber *> to a GPBStringInt64Dictionary proto object.
+ *  @return Reference to a GPBStringInt64Dictionary object.
+ */
+static GPBStringInt64Dictionary *FPRGetProtoCounterForDictionary(
```
visumickey, line 331: This might not be required. For now, lets comment out its implementation.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: From my understanding, if a default value is specified in the option file, then init_default will have the default value. Otherwise, init_default is the same as init_zero. All of these values could be found in `perf_metric.nanopb.h`. In our case, `firebase_perf_v1_PerfMetric_init_default == firebase_perf_v1_PerfMetric_init_zero`

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Good idea! Fixed.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Sounds good! Create encode and decode for stringToStringMap and stringToIntMap

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Updated.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Updated.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Updated.

FirebasePerformance/Sources/FPRClient.m
```diff
@@ -237,7 +237,10 @@ - (void)logGaugeMetric:(nonnull NSArray *)gaugeData forSessionId:(nonnull NSStri
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
     FPRMSGPerfMetric *metric = FPRGetPerfMetricMessage(self.config.appID);
-    FPRMSGGaugeMetric *gaugeMetric = FPRGetGaugeMetric(gaugeData, sessionId);
+    FPRMSGGaugeMetric *gaugeMetric = nil;
+    if ((gaugeData != nil && gaugeData.count != 0) && (sessionId != nil && sessionId.length != 0)) {
+      gaugeMetric = FPRGetGaugeMetric(gaugeData, sessionId);
+    }
```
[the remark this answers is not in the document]
jeremyjiang-dev: The tests will added in a following PR for updating FPRClient to use nanopb object.

jeremyjiang-dev:
@paulb777 Hey Paul! The check keeps failing due to style errors for non-FirebasePerformance files from the master branch. Do you know what went wrong? Thanks~

paulb777:
@jeremyjiang-dev You need to merge or rebase origin/master to your branch to get the clang-format fixes

jeremyjiang-dev:
> @jeremyjiang-dev You need to merge or rebase origin/master to your branch to get the clang-format fixes

It worked. Thanks Paul!

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,501 @@
...
+
+/** Reorders the list of sessions to make sure the first session is verbose if at least one session
+ *  in the list is verbose.
+ *  @return Ordered list of sessions.
+ */
+NSArray<FPRSessionDetails *> *MakeFirstSessionVerbose(NSArray<FPRSessionDetails *> *sessions) {
+  NSMutableArray<FPRSessionDetails *> *orderedSessions =
+      [[NSMutableArray<FPRSessionDetails *> alloc] initWithArray:sessions];
+
+  __block NSInteger firstVerboseSessionIndex = -1;
```
paulb777, line 486: Why not a plain for loop to iterate through the array? The block overhead seems like expensive overhead for just checking one field of each session.

On further research, it might be just a style difference - https://stackoverflow.com/questions/4486622/when-to-use-enumerateobjectsusingblock-vs-for

paulb777 (commented):
LGTM, deferring approval to @visumickey

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Thanks for the reference! I still changed it to a plain for loop to make it more straightforward.

FirebasePerformance/Sources/FPRClient.m
```diff
@@ -209,8 +209,8 @@ - (void)logNetworkTrace:(nonnull FPRNetworkTrace *)trace {
     return;
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
-    FPRMSGNetworkRequestMetric *networkRequestMetric = FPRGetNetworkRequestMetric(trace);
-    if (networkRequestMetric) {
+    if (trace.hasValidResponseCode) {
```
visumickey, line 212: Consider creating an API `-(BOOL) isValid` inside FPRNetworkTrace that can determine if the trace is valid. Checking on the responseCode, though fine for now, is not scalable/maintainable. The FPRNetworkTrace is the right object that can determine if the trace is valid/not.

FirebasePerformance/Sources/FPRClient.m
```diff
@@ -237,7 +237,10 @@ - (void)logGaugeMetric:(nonnull NSArray *)gaugeData forSessionId:(nonnull NSStri
   }
   dispatch_group_async(self.eventsQueueGroup, self.eventsQueue, ^{
     FPRMSGPerfMetric *metric = FPRGetPerfMetricMessage(self.config.appID);
-    FPRMSGGaugeMetric *gaugeMetric = FPRGetGaugeMetric(gaugeData, sessionId);
+    FPRMSGGaugeMetric *gaugeMetric = nil;
+    if ((gaugeData != nil && gaugeData.count != 0) && (sessionId != nil && sessionId.length != 0)) {
```
visumickey, line 241: Why would we have to check this here? Also, thinking wild if we can do this check in the GaugeMetric data structure for completeness?

FirebasePerformance/Sources/FPRNanoPbUtils.h
```diff
@@ -0,0 +1,94 @@
...
+#import <CoreTelephony/CTTelephonyNetworkInfo.h>
+#endif
+
+#import "FirebasePerformance/Sources/AppActivity/FPRTraceBackgroundActivityTracker.h"
+#import "FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.h"
+#import "FirebasePerformance/Sources/Public/FIRTrace.h"
+
+#import "FirebasePerformance/Sources/Protogen/nanopb/perf_metric.nanopb.h"
+
+typedef struct {
```
visumickey, line 27: Add comments for this structure.

FirebasePerformance/Sources/FPRNanoPbUtils.h
```diff
@@ -0,0 +1,94 @@
...
+  pb_bytes_array_t* _Nonnull key;
+  pb_bytes_array_t* _Nonnull value;
+} StringToStringMap;
+
+typedef struct {
```
visumickey, line 32: Add comments for this structure.

FirebasePerformance/Sources/FPRNanoPbUtils.h
```diff
@@ -0,0 +1,94 @@
...
+  int64_t value;
+} StringToNumberMap;
+
+extern pb_bytes_array_t* _Nullable FPREncodeData(NSData* _Nonnull data);
+
+extern pb_bytes_array_t* _Nullable FPREncodeString(NSString* _Nonnull string);
+
+extern NSData* _Nullable FPRDecodeData(pb_bytes_array_t* _Nonnull pbData);
+
+extern NSString* _Nullable FPRDecodeString(pb_bytes_array_t* _Nonnull pbData);
```
visumickey, line 44: Would appreciate some comments for each of these functions for better reference lookup.

visumickey: May be even consider something like FPRNanoPbDecodeUtils?

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+  pb_bytes_array_t *pbBytesArray = calloc(1, PB_BYTES_ARRAY_T_ALLOCSIZE(data.length));
+  if (pbBytesArray != NULL) {
+    [data getBytes:pbBytesArray->bytes length:data.length];
+    pbBytesArray->size = (pb_size_t)data.length;
+  }
+  return pbBytesArray;
+}
+
+pb_bytes_array_t *FPREncodeString(NSString *string) {
+  NSData *stringBytes = [string dataUsingEncoding:NSUTF8StringEncoding];
```
visumickey, line 55: With UTF-8 encoding, do we know the limitations? For eg. if the traceName comes up in non-english characters, would UTF-8 encoding still work? If not, consider using UTF16encoding.

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+}
+
+NSData *FPRDecodeData(pb_bytes_array_t *pbData) {
+  NSData *data = [NSData dataWithBytes:&(pbData->bytes) length:pbData->size];
+  return data;
+}
+
+NSString *FPRDecodeString(pb_bytes_array_t *pbData) {
+  NSData *data = FPRDecodeData(pbData);
+  return [NSString stringWithCString:[data bytes] encoding:NSUTF8StringEncoding];
```
visumickey, line 66: Same as above. Should we consider UTF16encoding?

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+    map[index].key = FPREncodeString(key);
+    map[index].value = FPREncodeString(value);
+    index++;
+  }];
+  return map;
+}
+
+NSDictionary<NSString *, NSString *> *FPRDecodeStringToStringMap(StringToStringMap *map,
+                                                                 NSInteger count) {
+  NSMutableDictionary<NSString *, NSString *> *dict = [NSMutableDictionary dictionary];
```
visumickey, line 82: General recommendation is to use [[NSMutableDictionary alloc] init].

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+    map[index].value = [value longLongValue];
+    map[index].has_value = true;
+    index++;
+  }];
+  return map;
+}
+
+NSDictionary<NSString *, NSNumber *> *_Nullable FPRDecodeStringToNumberMap(
+    StringToNumberMap *_Nullable map, NSInteger count) {
+  NSMutableDictionary<NSString *, NSNumber *> *dict = [NSMutableDictionary dictionary];
```
visumickey, line 105: General recommendation is to use [[NSMutableDictionary alloc] init].

FirebasePerformance/Sources/FPRNanoPbUtils.m
visumickey: Sounds good. Thanks Jeremy for clarifying!

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+
+firebase_perf_v1_GaugeMetric GetGaugeMetric(NSArray *gaugeData, NSString *sessionId) {
+  firebase_perf_v1_GaugeMetric gaugeMetric = firebase_perf_v1_GaugeMetric_init_default;
+  gaugeMetric.session_id = FPREncodeString(sessionId);
+
+  __block NSInteger cpuReadingsCount = 0;
+  __block NSInteger memoryReadingsCount = 0;
+
+  struct _firebase_perf_v1_CpuMetricReading *cpuReadings =
+      calloc([gaugeData count], sizeof(firebase_perf_v1_CpuMetricReading));
```
visumickey, line 310: If we are anyways going to reallocate at the end of counting the readings, should we start with less memory request first and later expand it? Instead of [gaugeData count], can we use just 1?

FirebasePerformance/Sources/FPRNanoPbUtils.m
```diff
@@ -0,0 +1,502 @@
...
+#endif
+
+#pragma mark - Nanopb creation utilities
+
+/** Converts the network method string to a value defined in the enum
+ *  firebase_perf_v1_NetworkRequestMetric_HttpMethod.
+ *  @return Enum value of the method string. If there is no mapping value defined for the method
+ * firebase_perf_v1_NetworkRequestMetric_HttpMethod_HTTP_METHOD_UNKNOWN is returned.
+ */
+static firebase_perf_v1_NetworkRequestMetric_HttpMethod FPRHTTPMethodForString(
```
visumickey, line 392: I would recommend moving the static functions to top to make it easier to reference. The older code suffered on this. Might as well fix it here.

FirebasePerformance/Sources/FPRClient.m
jeremyjiang-dev: Great suggestion! Updated.

FirebasePerformance/Sources/FPRClient.m
jeremyjiang-dev: The check was  in https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/Sources/FPRProtoUtils.m#L198-L204. Because `GetGaugeMetric` cannot return nil, I moved the checks here. I think it makes sense to have the check in FPRClient because it was passed by FPRClient and is not a part of a GaugeMetric object.

FirebasePerformance/Sources/FPRNanoPbUtils.h
jeremyjiang-dev: Done.

FirebasePerformance/Sources/FPRNanoPbUtils.h
jeremyjiang-dev: Done.

FirebasePerformance/Sources/FPRNanoPbUtils.h
jeremyjiang-dev: Comments are added.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: FPRProtoUtils was using UTF-8, so UTF-8 should be fine.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: FPRProtoUtils was using UTF-8, so UTF-8 should be fine.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Done.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Done.

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Good point! Expanding one by one could be expensive though according https://stackoverflow.com/questions/6727136/performance-impact-of-realloc

FirebasePerformance/Sources/FPRNanoPbUtils.m
jeremyjiang-dev: Sg! done.

FirebasePerformance/Sources/FPRNanoPbUtils.m
visumickey: Sounds good.

FirebasePerformance/Sources/FPRNanoPbUtils.m
visumickey: Sounds good.

FirebasePerformance/Sources/FPRClient.m
visumickey: Sounds good, makes sense.

visumickey (approved):

Issue #8305, opened 2021-06-25
Exclude code coverage for cross product changes


granluo:
#8295
RTDB was changed in this PR so incremental code coverage is running, but this change is in FirebaseCore and we wouldn't expect full code coverage on FirebaseCore from the RTDB tests.
https://github.com/firebase/firebase-ios-sdk/pull/8295#discussion_r658008412_

granluo:
The tool will actually search files in the xcresult bundle. And FirebaseDatabase xcresult bundles contains files under `firebase-ios-sdk/FirebaseCore/Sources/`,  that why the `FirebaseCore/Sources/FIRApp.m` is caught. The xcresult bundle contains files from the following dirs:
```
firebase-ios-sdk/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m
firebase-ios-sdk/FirebaseCore/Sources/*
firebase-ios-sdk/FirebaseDatabase/Sources/*
firebase-ios-sdk/FirebaseDatabase/Tests/*
firebase-ios-sdk/SharedTestUtilities/*

 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/CocoaPods-Lint-20210625-2536-4gmwka-FirebaseDatabase/Pods/*
```
We probably just need to involve `firebase-ios-sdk/FirebaseCore/Sources/*`

granluo:
[files-for-target](https://gist.github.com/mnem/d6d6e25c3ae647b422a7ac9dc73109b9#file-man-xccov-L31) seems to be a way to just output coverage for a specific target.
For files, we might need to parse the outcome in postprocess, like filter files by dir/path.

paulb777:
Closing for staleness. Please open a new issue and refer to this one if it is still occurring in recent releases.

PR #8396, opened 2021-07-15
Database API build tests

jrcrespoh:
#no-changelog

### DIFF
--- a/FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
+++ b/FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
@@ -79,21 +79,26 @@ final class DatabaseAPITests {
     var databaseQuery: DatabaseQuery = DatabaseQuery()
 
     // Observe for data
+
+    // observe(_ eventType:with block:)
     databaseHandle = databaseQuery.observe(dataEventType) { dataSnapshot in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     }
 
+    // observe(_ eventType:andPreviousSiblingKeyWith block:)
     databaseHandle = databaseQuery.observe(dataEventType) { dataSnapshot, optionalString in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
       let /* optionalString */ _: String? = optionalString
     }
 
+    // observe(_ eventType:with block:withCancel cancelBlock:)
     databaseHandle = databaseQuery.observe(dataEventType) { dataSnapshot in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     } withCancel: { error in
       let /* error */ _: Error = error
     }
 
+    // observe(_ eventType:andPreviousSiblingKeyWith block:withCancel cancelBlock:)
     databaseHandle = databaseQuery.observe(dataEventType) { dataSnapshot, optionalString in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
       let /* optionalString */ _: String? = optionalString
@@ -102,6 +107,8 @@ final class DatabaseAPITests {
     }
 
     // Get data
+
+    // getData(completion block:)
     databaseQuery.getData { optionalError, dataSnapshot in
       let /* optionalError */ _: Error? = optionalError
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
@@ -121,10 +128,13 @@ final class DatabaseAPITests {
     #endif // swift(>=5.5)
 
     // Observe Single Event
+
+    // observeSingleEvent(of eventType:with block:)
     databaseQuery.observeSingleEvent(of: dataEventType) { dataSnapshot in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     }
 
+    // observeSingleEvent(of eventType:andPreviousSiblingKeyWith block:)
     databaseQuery.observeSingleEvent(of: dataEventType) { dataSnapshot, optionalString in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
       let /* optionalString */ _: String? = optionalString
@@ -135,8 +145,8 @@ final class DatabaseAPITests {
         // async/await is a Swift 5.5+ feature available on iOS 15+
         async {
           do {
-            let /* (dataSnapshot, optionalString) */ _: (DataSnapshot,
-                                                         String?) = try await databaseQuery
+            // observeSingleEvent(of eventType:)
+            let _: (DataSnapshot, String?) = try await databaseQuery
               .observeSingleEventAndPreviousSiblingKey(of: dataEventType)
           } catch {
             // ...
@@ -145,12 +155,14 @@ final class DatabaseAPITests {
       }
     #endif // swift(>=5.5)
 
+    // observeSingleEvent(of eventType:with block:withCancel cancelBlock:)
     databaseQuery.observeSingleEvent(of: dataEventType) { dataSnapshot in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     } withCancel: { error in
       let /* error */ _: Error = error
     }
 
+    // observeSingleEvent(of eventType:andPreviousSiblingKeyWith block:withCancel cancelBlock:)
     databaseQuery.observeSingleEvent(of: dataEventType) { dataSnapshot, optionalString in
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
       let /* optionalString */ _: String? = optionalString
@@ -187,6 +199,7 @@ final class DatabaseAPITests {
 
     // MARK: - DatabaseReference
 
+    let priorityAny: Any = "priority"
     var values = [AnyHashable: Any]()
     var transactionResult = TransactionResult()
 
@@ -196,6 +209,8 @@ final class DatabaseAPITests {
 
     // Set value
     databaseReference.setValue(value)
+
+    // setValue(_ value:withCompletionBlock block:)
     databaseReference.setValue(value) { optionalError, databaseReference in
       let /* optionalError */ _: Error? = optionalError
       let /* databaseReference */ _: DatabaseReference = databaseReference
@@ -206,6 +221,7 @@ final class DatabaseAPITests {
         // async/await is a Swift 5.5+ feature available on iOS 15+
         async {
           do {
+            // setValue(_ value:)
             databaseReference = try await databaseReference.setValue(value)
           } catch {
             // ...
@@ -215,6 +231,8 @@ final class DatabaseAPITests {
     #endif // swift(>=5.5)
 
     databaseReference.setValue(value, andPriority: priority)
+
+    // setValue(_ value:andPriority priority:withCompletionBlock block:)
     databaseReference.setValue(value, andPriority: priority) { optionalError, databaseReference in
       let /* optionalError */ _: Error? = optionalError
       let /* databaseReference */ _: DatabaseReference = databaseReference
@@ -225,6 +243,7 @@ final class DatabaseAPITests {
         // async/await is a Swift 5.5+ feature available on iOS 15+
         async {
           do {
+            // setValue(_ value:andPriority priority:)
             databaseReference = try await databaseReference.setValue(value, andPriority: priority)
           } catch {
             // ...
@@ -235,6 +254,8 @@ final class DatabaseAPITests {
 
     // Remove value
     databaseReference.removeValue()
+
+    // removeValue(completionBlock block:)
     databaseReference.removeValue { optionalError, databaseReference in
       let /* optionalError */ _: Error? = optionalError
       let /* databaseReference */ _: DatabaseReference = databaseReference
@@ -255,6 +276,8 @@ final class DatabaseAPITests {
 
     // Set priority
     databaseReference.setPriority(priority)
+
+    // setPriority(_ priority:withCompletionBlock block:)
     databaseReference.setPriority(priority) { optionalError, databaseReference in
       let /* optionalError */ _: Error? = optionalError
       let /* databaseReference */ _: DatabaseReference = databaseReference
@@ -265,6 +288,7 @@ final class DatabaseAPITests {
         // async/await is a Swift 5.5+ feature available on iOS 15+
         async {
           do {
+            // setPriority(_ priority:)
             databaseReference = try await databaseReference.setPriority(priority)
           } catch {
             // ...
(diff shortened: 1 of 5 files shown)


jrcrespoh:
WIP, other Database APIs yet to be tested. Using type casting to verify compile time types.

jrcrespoh:
I'm considering replacing literal values (e.g. `"path"`, `0`) with empty init calls (`String()`, `Int()`) to avoid the type casting (`as String`, `as Int`). Should I go ahead with this change?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,598 @@
...
+      _ = Database.database(app: app as FirebaseApp, url: "url" as String) as Database
+      _ = Database.database(app: app as FirebaseApp) as Database
+    }
+    let instance = Database.database() as Database
+    // Retrieve FirebaseApp
+    _ = instance.app as FirebaseApp?
+    // Retrieve DatabaseReference
+    _ = instance.reference() as DatabaseReference
+    _ = instance.reference(withPath: "path" as String) as DatabaseReference
+    _ = instance.reference(fromURL: "url" as String) as DatabaseReference
```
ncooke3, line 41: @jrcrespoh So is this where a warning appears asking you to do the cast? Does doing `instance.reference(withPath: "path")` not work?

I would say keep it as is because, ideally, I think the API should be able to handle the literal. Showing the cast in the build test highlights some awkwardness in the API that we can improve.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,598 @@
...
+    databaseReference.runTransactionBlock({ (mutableData: MutableData) in
+                                            TransactionResult()
+                                          },
+                                          andCompletionBlock: { (
+                                            optionalError: Error?,
+                                            bool: Bool,
+                                            optionalDataSnapshot: DataSnapshot?
+                                          ) in
+                                          // ...
+                                          }, withLocalEvents: true as Bool)
```
ncooke3, line 527: Curious, are you able to use this API using the Swift's multiple trailing closure syntax? This [doc](https://github.com/apple/swift-evolution/blob/main/proposals/0279-multiple-trailing-closures.md) shows what they look like.

It'd probably look like this:
```swift
databaseReference.runTransactionBlock { (mutableData: MutableData) in 
  TransactionResult()
}, { (optionalError: Error?, bool: Bool, optionalDataSnapshot: DataSnapshot?) in
  // ...
}, withLocalEvents: true as Bool)
```

ncooke3 (commented):
Looking really good!

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
jrcrespoh: I don't need to cast for it to compile successfully, I chose to do so to be explicit about the type since some function calls used parameters of type `Any?` so I had to provide a literal and I arbitrarily used string literals, so I wanted to highlight the arbitrariness.

Also, the explicit type casting would provide compile-time errors whenever types used are changed (say from `Any` to `Any?`) so I thought that would be useful and thus went ahead and used type casting.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,598 @@
...
+
+    // Retrieve Database Instance
+    _ = Database.database() as Database
+    _ = Database.database(url: "url" as String) as Database
+    if let app = FirebaseApp.app() {
+      _ = Database.database(app: app as FirebaseApp, url: "url" as String) as Database
+      _ = Database.database(app: app as FirebaseApp) as Database
+    }
+    let instance = Database.database() as Database
+    // Retrieve FirebaseApp
```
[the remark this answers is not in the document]
jrcrespoh: I recently noticed that the header files already have `#pragma mark` sections so I think I'll also replace mine with those already in use. I will also add the new lines as suggested 👍

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
jrcrespoh: I don't think I can because the last named parameter is not a closure but rather a `Bool`, though I'll investigate further...

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,598 @@
...
+final class DatabaseAPITests {
+  func usage() {
+    // MARK: - Database
+
+    // Retrieve Database Instance
+    _ = Database.database() as Database
+    _ = Database.database(url: "url" as String) as Database
+    if let app = FirebaseApp.app() {
+      _ = Database.database(app: app as FirebaseApp, url: "url" as String) as Database
+      _ = Database.database(app: app as FirebaseApp) as Database
```
ncooke3, line 33: @paulb777 @ryanwilson @maksymmalyhin  based on the style of the other API tests, this line would look like:
```swift
_ = Database.database(app: app)
```
But with the added type casts here, I think it adds a bit more context around the API's types and would test if the param/return type of `database()` were to change.

Only downside is that it kinda muddles the usage of each API since a developers wouldn't normally do this.

Perhaps we should adopt a hybrid approach that:
- demonstrates intended usage
- does type checking buildtime to catch API changes.

Possible approach:
```swift 
let database = Database.database(app: app)  // Usage example
_ = database as Database // Type check
```
I think this would be a middle ground that'd allow us to evaluate API's intended usage while also also being a compile time type check to catch API changes

Great work @jrcrespoh for approaching it from this angle, I'd like to know what you think as well!

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
ncooke3: Ah, good point. Ok, cool, no worries. I dont know enough about the API but this could be an example where the API could be improved to be like: `func runTransactionBlock(localEvents: Bool, transactionHandler: blah, completion: blah)`

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
ncooke3: Gotcha, I think an alternative approach that shows clearer usage while still checking explicit types could be:
```swift
let url = "url"
databaseRef = instance.reference(fromURL: url)
_ = databaseRef as DatabaseReference
```
WDYT?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
ncooke3: Another option that does both in one line is:
```swift
let database: Database = Database.database(app: app)
```
But we'd have to keep the style script from running on it because it will replace `database` with an "_" since in most cases of these tests we don't use it again.

jrcrespoh: I do agree that the separation of developer usage and compile-time type check would be even better and could be implemented as suggested. I would also wonder if there would be a way to further separate these beyond just lines, say on separate functions (e.g `usage()` vs `explicit()`) or even separate files? I imagine further separation would make it even more readable.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
jrcrespoh: LGTM. How would you propose we tackle function calls that take in a parameter of type, say, `Any?`?

jrcrespoh:
@paulb777 Should FirebaseDatabase be added to [SwiftPMTests/swift-tests/main.swift](../blob/master/SwiftPMTests/swift-test/main.swift)?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,602 @@
...
+    // MARK: - ServerValue
+
+    _ = ServerValue.timestamp() as [AnyHashable: Any]
+    _ = ServerValue.increment(0 as NSNumber) as [AnyHashable: Any]
+
+    // MARK: - TransactionResult
+
+    _ = TransactionResult.success(withValue: MutableData()) as TransactionResult
+    _ = TransactionResult.abort() as TransactionResult
+  }
```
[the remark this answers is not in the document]
paulb777: @granluo Test files should not be included in coverage reports

paulb777:
> @paulb777 Should FirebaseDatabase be added to [SwiftPMTests/swift-tests/main.swift](../blob/master/SwiftPMTests/swift-test/main.swift)?

Yes. Looks like an oversight. Good catch!

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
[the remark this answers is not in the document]
granluo: It seems to be the same as  #8305. This could be resolved by filtering files.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,602 @@
...
+    _ = databaseReference.queryStarting(afterValue: "value" as Any?) as DatabaseQuery
+    _ = databaseReference.queryStarting(
+      afterValue: "value" as Any?,
+      childKey: "key" as String?
+    ) as DatabaseQuery
+    _ = databaseReference.queryEnding(atValue: "value" as Any?) as DatabaseQuery
+    _ = databaseReference.queryEnding(
+      atValue: "value" as Any?,
+      childKey: "key" as String?
+    ) as DatabaseQuery
```
jrcrespoh, line 385: `DatabaseReference` inherits from `DatabaseQuery` the following methods:
- `open func queryEnding(beforeValue endValue: Any?) -> DatabaseQuery`
- `open func queryEnding(beforeValue endValue: Any?, childKey: String?) -> DatabaseQuery`

These (and possibly other similar methods) are not currently tested since they do not show up in the generated Swift 5 Interface for FIRDatabaseReference.h, but should they?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
paulb777: I like the clear distinction between the usage example and the type checking example.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,639 @@
...
+    _ = mutableData.childrenCount as UInt
+    _ = mutableData.children as NSEnumerator
+    _ = mutableData.key as String?
+
+    // MARK: - ServerValue
+
+    _ = ServerValue.timestamp() as [AnyHashable: Any]
+    _ = ServerValue.increment(0 as NSNumber) as [AnyHashable: Any]
+
+    // MARK: - TransactionResult
```
paulb777, line 634: I like the MARK indicating the header 👍

Package.swift
```diff
@@ -498,13 +498,23 @@ let package = Package(
...
+        // Disable Swift tests as mixed targets are not supported (Xcode 12.4).
+        "Unit/Swift",
         "Integration/",
       ],
       resources: [.process("Resources")],
       cSettings: [
         .headerSearchPath("../.."),
       ]
     ),
+    .testTarget(
```
paulb777, line 510: A corresponding scheme should be added to .swiftpm/xcode/xcshemes

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -623,6 +625,10 @@ final class DatabaseAPITests {
       break
     case .value:
       break
+    case .none:
+      break
+    case .some:
+      break
```
morganchen12, line 631: nit: you can shorten this code by putting value, none, and some in the same statement.

```swift
case .value, .none, .some:
  break
```

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
paulb777: Probably not necessary - at least in the first version

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,683 @@
...
+
+    var url = "url"
+    let path = "path"
+    let host = "host"
+    let port = 0
+    let yes = true
+
+    // Retrieve Database Instance
+    var database = Database.database()
+    _ = database as Database
```
ncooke3, line 36: Looking at this again, since we use `database` more than once, I think we should should add line 36's type info to line 35. So:
```swift
var database: Database = Database.database()
```

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
jrcrespoh: Just to confirm, `.value` is added by `DataEventType` while `.none` and `.some` aren't. Should all three still be grouped or should I just group `.none` and `.some`?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -0,0 +1,683 @@
...
+      database = Database.database(app: app, url: url)
+      database = Database.database(app: app)
+    }
+
+    // Retrieve FirebaseApp
+    let /* app */ _: FirebaseApp? = database.app
+
+    // Retrieve DatabaseReference
+    var databaseReference = database.reference()
+    _ = databaseReference as DatabaseReference
```
ncooke3, line 50: Same again here:
```swift
var databaseReference: DatabaseReference = database.reference()
```

Package.swift
paulb777: Disregard. This is no longer the process.

paulb777 (approved):
LGTM, when the last comments are resolved and CI is green.

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
morganchen12: It's up to you :)

ncooke3 (approved):
LGTM on green CI. Great work! 🎉

morganchen12 (approved):

PR #8399, opened 2021-07-16
Bump App Check version to 8.5.0-beta

maksymmalyhin:
Should fix https://github.com/firebase/firebase-ios-sdk/runs/3083903668?check_suite_focus=true

Test run: https://github.com/firebase/firebase-ios-sdk/actions/runs/1037955080

### DIFF
--- a/FirebaseAppCheck.podspec
+++ b/FirebaseAppCheck.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'FirebaseAppCheck'
-  s.version          = '8.4.0-beta'
+  s.version          = '8.5.0-beta'
   s.summary          = 'Firebase App Check SDK.'
 
   s.description      = <<-DESC


paulb777:
Isn't the prerleease script supposed to manage the versions? @granluo

maksymmalyhin:
> Isn't the prerleease script supposed to manage the versions? @granluo

It doesn't look like we bump the version for prerelease testing (see [here](https://github.com/firebase/firebase-ios-sdk/blob/41b00739990d098c48fa4fc63dbf764af434c34e/scripts/release_testing_setup.sh#L37)) which, I think, we should do.

paulb777:
I'm going to merge this as a workaround to get the prerelease nightly working. We should still get the prerelease scripting fixed to manage the version correctly.

paulb777 (approved):

granluo:
> Isn't the prerleease script supposed to manage the versions? @granluo

Yes, the prerelease will manage the versions but for `beta` version will be excluded. Should we include `beta` version?

maksymmalyhin:
> Yes, the prerelease will manage the versions but for `beta` version will be excluded. Should we include `beta` version?

@granluo  Yes, we need update update all tags including `beta` for prerelease testing.

PR #8400, opened 2021-07-16
Remove allow-warnings in pod push

granluo:


paulb777:
It's still needed where `allowWarnings: true` is in https://github.com/firebase/firebase-ios-sdk/blob/master/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift#L41

PR #8403, opened 2021-07-17
Fix tvOS SPM tests

paulb777:
#no-changelog

### DIFF
--- a/FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
+++ b/FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
@@ -14,25 +14,24 @@
 
 import FirebaseInAppMessaging
 
-@available(iOS 13, tvOS 13, *)
-@available(iOSApplicationExtension, unavailable)
+@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, *)
 public struct InAppMessagingPreviewHelpers {
   public static func cardMessage(campaignName: String = "Card message campaign",
                                  title: String = "Title for modal message",
                                  body: String? = "Body for modal message",
                                  textColor: UIColor = UIColor.label,
-                                 backgroundColor: UIColor = UIColor.systemBackground,
+                                 backgroundColor: UIColor = defaultBackgroundColor(),
                                  portraitImage: UIImage = UIImage(systemName: "rectangle")!,
                                  landscapeImage: UIImage? = UIImage(systemName: "square"),
                                  primaryButtonText: String = "Click me!",
                                  primaryButtonTextColor: UIColor = UIColor.systemBlue,
-                                 primaryButtonBackgroundColor: UIColor = UIColor
-                                   .secondarySystemBackground,
+                                 primaryButtonBackgroundColor: UIColor =
+                                   defaultPrimaryBackgroundColor(),
                                  primaryActionURL: URL? = nil,
                                  secondaryButtonText: String? = "Dismiss",
                                  secondaryButtonTextColor: UIColor? = UIColor.secondaryLabel,
-                                 secondaryButtonBackgroundColor: UIColor? = UIColor
-                                   .tertiarySystemBackground,
+                                 secondaryButtonBackgroundColor: UIColor? =
+                                   defaultSecondaryBackgroundColor(),
                                  secondaryActionURL: URL? = nil,
                                  appData: [String: String]? = nil) -> InAppMessagingCardDisplay {
     // This may crash the preview if an invalid portrait image is provided, card messages must have
@@ -73,6 +72,30 @@ public struct InAppMessagingPreviewHelpers {
       secondaryActionURL: secondaryActionURL,
       appData: appData
     )
+  }
+
+  public static func defaultBackgroundColor() -> UIColor {
+#if os(tvOS)
+    return UIColor.white
+#else
+    return UIColor.systemBackground
+#endif
+  }
+
+  public static func defaultPrimaryBackgroundColor() -> UIColor {
+#if os(tvOS)
+    return UIColor.gray
+#else
+    return UIColor.secondarySystemBackground
+#endif
+  }
+
+  public static func defaultSecondaryBackgroundColor() -> UIColor {
+#if os(tvOS)
+    return UIColor.gray
+#else
+    return UIColor.tertiarySystemBackground
+#endif
   }
 
   public static func modalMessage(campaignName: String = "Modal message campaign",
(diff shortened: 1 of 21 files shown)


paulb777:
The travis failure is related to recent FIAM changes and unrelated. I'll deal with that separately.

This PR is ready for review.

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -73,6 +74,30 @@ public struct InAppMessagingPreviewHelpers {
     )
   }
 
+  public static func defaultBackgroundColor() -> UIColor {
+    #if os(tvOS)
+      return UIColor.white
```
ryanwilson, line 79: Nit: typically tvOS has dark interfaces, so I'd prefer black over white here.

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -73,6 +74,30 @@ public struct InAppMessagingPreviewHelpers {
...
+    #else
+      return UIColor.systemBackground
+    #endif
+  }
+
+  public static func defaultPrimaryBackgroundColor() -> UIColor {
+    #if os(tvOS)
+      return UIColor.gray
```
ryanwilson, line 87: Nit: `systemGray` https://developer.apple.com/documentation/uikit/uicolor/3173143-systemgray

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -73,6 +74,30 @@ public struct InAppMessagingPreviewHelpers {
...
+    #else
+      return UIColor.secondarySystemBackground
+    #endif
+  }
+
+  public static func defaultSecondaryBackgroundColor() -> UIColor {
+    #if os(tvOS)
+      return UIColor.gray
```
ryanwilson, line 95: Nit: `systemGray2` https://developer.apple.com/documentation/uikit/uicolor/3255071-systemgray2

maksymmalyhin (approved):
LGTM

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -73,6 +74,30 @@ public struct InAppMessagingPreviewHelpers {
     )
   }
 
+  public static func defaultBackgroundColor() -> UIColor {
```
ryanwilson, line 77: Hmm, could we do these inline by any chance? Otherwise these are now part of the public API surface of the product which we don't want (this is a limitation that the default value needs to have the same visibility).

If not, maybe we change it to make it colours available on all platforms.

ryanwilson (changes requested):

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -14,23 +14,24 @@
 
 import FirebaseInAppMessaging
 
+@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, *)
 public struct InAppMessagingPreviewHelpers {
```
ncooke3, line 18: If we're gating `InAppMessagingPreviewHelpers` with this availability check (that includes `tvOS 13.0`, are the added background color APIs necessary?

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
ncooke3, line 79: I think there should be an additional check within these helpers that returns a light or dark color based on [interface style](https://developer.apple.com/documentation/uikit/uitraitcollection/1651063-userinterfacestyle).

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: Screen Shot 2021-07-19 at 9 22 25 AM

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: I couldn't figure out a good way to inline.  Any suggestions on colo(u)rs?

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: Makes sense, would you do a follow up PR to update?

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: done

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: I changed to colors for both, since I can't figure out a good way to make the API use a platform-specific initializer without expanding the public API.

More testing is needed to verify the colors work, but I'd like to merge this here to fix CI

@ryanwilson PTAL

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
```diff
@@ -14,25 +14,25 @@
 
 import FirebaseInAppMessaging
 
-@available(iOS 13, tvOS 13, *)
+@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, *)
```
ryanwilson, line 17: Nit: FIAM isn't available on macOS or macCatalyst yet as far as I'm aware. This could be marked as unavailable instead.

ryanwilson (approved):

ryanwilson (approved):

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: Done

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: Verify deferred to fix for #8416

FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift
paulb777: Added to #8416 description

PR #8351, opened 2021-07-08
Add support for testing in-app messages in SwiftUI previews

christibbs:
In core FIAM SDK: Opens up initializers for image subclasses
- See changes in `FirebaseInAppMessaging/Sources/Public/FirebaseInAppMessaging/FIRInAppMessagingRendering.h`

In core FIAM UI test project, refactor to use open message initializers
- See changes in test files
`FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/FiamDisplaySwiftExample/BannerMessageViewController.swift`
`FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/FiamDisplaySwiftExample/ImageOnlyMessageViewController.swift`
`FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/FiamDisplaySwiftExample/ModalMessageViewController.swift`
`FirebaseInAppMessaging/Tests/Integration/DefaultUITestApp/FiamDisplaySwiftExample/CardMessageViewController.swift`

In FIAM SwiftUI SDK: Adds preview helper methods to create canned in-app messages
- See changes in new file `FirebaseInAppMessaging/Swift/Source/SwiftUIPreviewHelpers.swift`

In FIAM SwiftUI integration test project: Updates project to use SwiftUI preview helpers
- See changes in new file `FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift`

PR #8404, opened 2021-07-17
Turn off failing travis FIAM tests

paulb777:
Something about #8351 caused travis infra to fail.  See https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/774672431 

Since travis is going away and these tests are running on GHA, deleting to get travis green again.

Note `pod lib lint` is still running for FIAM on travis.

#no-changelog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,23 +101,6 @@ jobs:
 
     - stage: test
       env:
-        - PROJECT=InAppMessaging PLATFORM=iOS METHOD=xcodebuild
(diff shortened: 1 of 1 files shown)


ryanwilson (approved):

ryanwilson:
Maybe it was the missing `@available` check? I'm still fine deleting this though.

paulb777:
I don't think so, since it looks like a test hang. I'll go ahead and delete.

PR #8406, opened 2021-07-19
Involve beta and regular version

granluo:


scripts/release_testing_setup.sh
```diff
@@ -48,7 +48,7 @@ if [ "$TESTINGMODE" = "release_testing" ]; then
   # Latest release branch with prefix CocoaPods, e.g. CocoaPods-7.11
   latest_branch_version=$(git branch -l -r --sort=-v:refname origin/release-* | head -n 1 | sed -n 's/^[ \t]*origin\/release/CocoaPods/p' )
   # Latest tag of release branch on the repo, e.g. Cocoapods-7.9.0
-  latest_release_branch_tag=$(git tag -l --sort=-version:refname "$latest_branch_version"*[0-9] | head -n 1)
+  latest_release_branch_tag=$(git tag -l --sort=-version:refname "CocoaPods-*" | grep ".*-beta$\|.*[0-9]$" | head -n 1)
```
maksymmalyhin, line 51: For actual releases we create two tags
- CocoaPods-<version>
- CocoaPods-<version>-beta

It doesn't look like this change alone will fix the issue.

paulb777: For prerelease testing, both the beta and non-beta versions could be changed to use the same nightly tag.

Issue #8407, opened 2021-07-19
FR: Firestore Property Wrapper to Fetch Documents


chFlorian:
### Feature proposal

* Firebase Component: FirestoreSwift

I am proposing a property wrapper that works like FetchRequest (CoreData). It should fetch a specified type of documents from a specified collection, ideally with predicates such as whereField and limit.

PR #8408, opened 2021-07-19
Draft: Add property wrapper to easily fetch Firestore documents

chFlorian:
### Discussion

  * [Issue](https://github.com/firebase/firebase-ios-sdk/issues/8407)
  * This PR resembles a draft of a possible implementation
  * Currently the `@Published` part of the wrappedValue is not working inside ViewModels. It does however work inside a View.
  * I already started adding some predicates (see FIRPredicate), but not all are implemented yet (e.g. limit)
  * Perhaps it would make sense to create a snapshotListener instead of reading the documents once

### Testing

  * I have not created tests for this feature yet

### API Changes

  * Add FIRPredicate Enum
  * Add FIRRequest Property Wrapper with `@Published` wrappedValue


chFlorian:
CI is throwing these errors, which I can't understand or reproduce. 
As far as I understand `compactMap` does not return an optional (which is the whole idea of compactMap). 
image

I also noticed that CI uses Xcode 11 ( could be the issue here?) and this warning:
image

chFlorian:
After rewriting the Property Wrapper to internally contain a StateObject, it is now usable inside Views.
image

paulb777:
Thanks for the PR! We know longer support Xcode 11 and are in the process of deprecating travis from our CI. I've approved the Git Hub Actions CI run for the PR.  I'll send a separate PR to either update the travis Xcode version or turn it off that you can merge or rebase to resolve the travis failure.

chFlorian:
Thanks @paulb777 !
As this is just a draft, the feature is still WIP, so it won't need to get merged quite yet.
Is there a place where I can find style guidelines? Right now the test failes because of some linter check, that I didn't know about.

ncooke3:
Hi @chFlorian! For styling, see the instructions for installing the styling tools [here](https://github.com/firebase/firebase-ios-sdk/blob/master/CONTRIBUTING.md#getting-started). With those installed, you'll be able to run the style script. Instructions for running that are [here](https://github.com/firebase/firebase-ios-sdk/blob/master/CONTRIBUTING.md#style-guide)

paulb777:
And this specific error is from `scripts/check_whitespace.sh` which you can run locally to confirm. We don't allow trailing whitespace.

Firestore/Swift/Source/Property Wrapper/FIRRequest.swift
```diff
@@ -0,0 +1,142 @@
...
+    @Published var items: [T] = []
+    
+    private let store = Firestore.firestore()
+    
+    init(_ collection: String, _ predicates: [FIRPredicate]) {
+        load(from: collection, withPredicates: predicates)
+    }
+    
+    private func load(from collection: String, withPredicates predicates: [FIRPredicate]) {
+        let reference = store.collection(collection)
```
schmidt-sebastian, line 37: Note that a CollectionReference is a subclass of Query. You should be able to simplify your code if you treat this as a Query.

schmidt-sebastian (commented):
Thanks for sending this. Have you thought about what a snapshot listener approach would look like? That would be a big value add for our users.

Firestore/Swift/Source/Property Wrapper/FIRRequest.swift
chFlorian: Thanks for the hint, this really does simplify it a lot!

chFlorian:
> Thanks for sending this. Have you thought about what a snapshot listener approach would look like? That would be a big value add for our users.

I think nothing much would change actually. Just replace getDocuments with addSnapshotListener. Memory-wise this also shouldn't be an issue, as the `FirebaseStore` class is handled as a StateObject, meaning that it will deallocated once the view is dismissed. 

The only thing that I am not sure about yet is error handling (e.g. bad snapshots or internal errors). Currently I don't see any easy way to handle errors that doesn't break the development experience when using this...

schmidt-sebastian:
We also couldn't come up with a good way to handle errors. We probably just need to return an empty set fo results.

As we are transitioning this to use snapshot listeners, we also need to keep in mind that we need to unregister the listener, as listeners incur network and billing overhead. This can likely be done using `deinit`.

chFlorian:
I am sorry for the amount of commits, but I'm testing this by importing my version of the sdk to an app.

chFlorian:
> We also couldn't come up with a good way to handle errors. We probably just need to return an empty set fo results.
> 
> As we are transitioning this to use snapshot listeners, we also need to keep in mind that we need to unregister the listener, as listeners incur network and billing overhead. This can likely be done using `deinit`.

I have implemented exactly this in my latest commit. The listener is working :)

schmidt-sebastian:
I think this looks pretty good now. I will follow up with some Swift experts to get a full review.

Can you add `isNotEqual`?

chFlorian:
> I think this looks pretty good now. I will follow up with some Swift experts to get a full review.
> 
> Can you add `isNotEqual`?

Thanks! I still have to look into the style guidelines/linting and I was thinking about adding other predicates such as `limit`. What do you think about that idea?

schmidt-sebastian:
Oh yes - we do need to add limit/limitToLast and cursor support (`startingAt` and friends).

peterfriese (changes requested):
I created a small sample project and played around with this, and it looks very promising!

I left a bunch of comments regarding naming and DevXP.

In addition, I'd love to see support for
- ordering (see https://firebase.google.com/docs/firestore/query-data/order-limit-data)
- limiting

W.r.t. error handling: properties can throw (starting with Swift 5.5) - see https://www.swiftbysundell.com/articles/throwing-async-properties/ and https://www.hackingwithswift.com/articles/233/whats-new-in-swift-5-5. It seems this isn't supported for property wrappers yet - when trying to add a `throw` clause, the compiler says that `Property wrappers currently cannot define an 'async' or 'throws' accessor`.

Until we're able to use this, I think it might be a good idea to not gloss over any mapping errors, but expose them to the user instead.

Firestore/Swift/Source/Property Wrapper/FIRRequest.swift
```diff
@@ -0,0 +1,97 @@
+//
+//  FIRRequest
+//
+//  Created by Florian Schweizer on 12.07.21.
+//
+
+import FirebaseFirestore
+import SwiftUI
+
+public enum FIRPredicate {
```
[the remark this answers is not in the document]
chFlorian: As far as I know adding a parameter name doesn't work for enum cases if there are multiple with the same type signatures.
e.g.
```swift
case isIn(field: String, condition: [Any])
case isNotIn(field: String, condition: [Any])
```
would be considered the same case by the compiler. Hence I will go with factory methods as they provide exactly what's needed here.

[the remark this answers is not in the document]
chFlorian: I really like the factory methods as they provide the same semantics as the well known .whereField functions.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,158 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import FirebaseFirestore
+import SwiftUI
+
+public enum FIRPredicate {
```
chFlorian, line 20: @peterfriese Should we rename this enum as well?

chFlorian:
> Hi @chFlorian! For styling, see the instructions for installing the styling tools [here](https://github.com/firebase/firebase-ios-sdk/blob/master/CONTRIBUTING.md#getting-started). With those installed, you'll be able to run the style script. Instructions for running that are [here](https://github.com/firebase/firebase-ios-sdk/blob/master/CONTRIBUTING.md#style-guide). Don't hesitate to reach out with issues!

```
(base) user@device firebase-ios-sdk % brew upgrade clang-format                              
Warning: clang-format 12.0.1 already installed
(base) user@device firebase-ios-sdk % scripts/style.sh Firestore/Swift/Source/PropertyWrapper
Please upgrade to clang-format version 12.
If it's installed via homebrew you can run:
brew upgrade clang-format
```

It looks to me like the style script doesn't work?

paulb777:
You likely need to merge or rebase origin/master.  The GitHub Actions clang-format version updated to 12.0.1 after your branch started.

mortenbekditlevsen:
I saw this cool PR and had some random thoughts:

`@Published` itself uses the property wrapper initialized variation that allows the wrapper to know about the 'enclosing self'. If that 'self' confirms to `ObservableObject` it calls the 'objectWillChange' on it. (It does something more too, because it pseudo-am synthesizes the objectWillChange in case it doesn't exist.)

Perhaps that could be used to make the wrapper work inside a view model?

A thought about errors: perhaps the array of items could be an array of Results (if it makes sense to have a possible failure per item) or a Result of an array (in case it makes sense to have only one). Then the wrappedValue could be a compactMap of the success cases but the actual results could be exposed through a projected value of anyone is interested in them - and if the projected value should not be reserved for some other purpose.

I'm not at a computer right now, so I can't experiment, but in case any of my thoughts could be of use, I'd gladly help experimenting. 😊

mortenbekditlevsen:
I would personally love the possibility of customization of the decoder and the Firestore instance (in case we get multiple instances in the future, that is).

With the current values as defaults.

chFlorian:
> `@Published` itself uses the property wrapper initializer variation that allows the wrapper to know about the 'enclosing self'. If that 'self' conforms to `ObservableObject` it emits the 'objectWillChange' on it. (It does something more too, because it pseudo-synthesizes the objectWillChange in case it doesn't exist.)

Thanks for looking into this PR! I actually already played around with the projected Value and tried to make it a PassthroughSubject, since I saw that `@Published` also has a Publisher in that place. I don't understand how to connect everything then though. So how do I forward the publishers values to the enclosing ObservableObject? Do you have a concrete idea for this?

mortenbekditlevsen:
Hi @chFlorian ,

I don't think my 'enclosing self' property wrapper suggestion works anyway - at least it's too constrained.
Since an `ObservableObject` can have any `Publisher` as it's `objectWillChange`, we can in general not emit a new value to it. In the special case where the `Publisher` is an `ObjectWillChangePublisher` (which is the one you get if you have one auto-synthesized by adding a `@Published` property to the `ObservableObject`), then we can emit values.

Here's my testing that shows the idea. The 'willUpdate' closure should be defined on the `QueryStore` and called on every new value in Firestore.

```
@available(iOS 14.0, *)
@propertyWrapper
public struct FirestoreQuery<T: Decodable>: DynamicProperty {
  @StateObject private var store: QueryStore<T>

    var willUpdate: () -> Void = {}

    public var wrappedValue: Int? {
      get { fatalError("called wrappedValue getter") }
      set { fatalError("called wrappedValue setter") }
    }

    public static subscript<EnclosingSelf: ObservableObject>(
      _enclosingInstance observed: EnclosingSelf,
      wrapped wrappedKeyPath: ReferenceWritableKeyPath<EnclosingSelf, [T]>,
      storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, FirestoreQuery>
    ) -> [T] where EnclosingSelf.ObjectWillChangePublisher: ObservableObjectPublisher {
      get {
        observed[keyPath: storageKeyPath].willUpdate = {
            observed.objectWillChange.send()
    }
        return observed[keyPath: storageKeyPath].store.items
      }
      set {
        // TODO
      }
    }


  public init(collectionPath: String, predicates: [FIRPredicate] = []) {
    _store =
      StateObject(wrappedValue: QueryStore<T>(collectionPath: collectionPath,
                                              predicates: predicates))
  }
}
```

There should of course also be a non-restricted initializer for when the enclosing self is not an `ObservableObject`.
One could also add another initializer that works when the `objectWillChange` publisher is a `Subject` with `Void` input...
Don't know if it's too much hazzle and too unpredictable how to use it.

mortenbekditlevsen:
Another consideration is that the 'enclosing self' feature is not official. It is, however, a compiler transformation, so it won't stop working at runtime for an app built with the compiler today. And we can rest assured that some form of the 'enclosing self' property wrapper will continue to exist since Combine/SwiftUI makes use of it.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
peterfriese: Yes, I suggested `QueryPredicate` in https://github.com/chFlorian/firebase-ios-sdk/pull/1/commits/41114476945ce040df34e26c6113d35b7759eef3#diff-78e468ae8bbbcd59d12162d4ba9b4399e9a4aa2c48898f52cb83bbdb7a1a6c79R20

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+    .orderBy(lhs, rhs)
+  }
+
+  public static func limit(to lhs: Int) -> FIRPredicate {
+    .limitTo(lhs)
+  }
+}
+
+@available(iOS 13.0, *)
+internal class QueryStore<T: Decodable>: ObservableObject {
```
peterfriese, line 88: Not sure `QueryStore` is a good choice, as the term _Store_ has such an overloaded meaning already in the context of Firestore.

@ryanwilson @paulb777 @schmidt-sebastian WDYT?

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+      case let .orderBy(lhs, rhs):
+        query = query.order(by: lhs, descending: rhs)
+      case let .limitTo(lhs):
+        query = query.limit(to: lhs)
+      }
+    }
+
+    listener = query.addSnapshotListener { [weak self] snapshot, error in
+      if let error = error {
+        print(error)
```
peterfriese, line 135: We still need to decide how to deal with errors.

peterfriese (changes requested):

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+  case isGreaterThanOrEqualTo(_ lhs: String, _ rhs: Any)
+
+  case orderBy(_ lhs: String, _ rhs: Bool)
+
+  case limitTo(_ lhs: Int)
+
+  /*
+   Factory methods to expose the underlying enum cases with a nicer development experience and improved semantics.
+   */
+  public static func whereField(_ lhs: String, isEqualTo rhs: Any) -> FIRPredicate {
```
peterfriese, line 42: I find `rhs` and `lhs` a bit confusing. Let's something more descriptive, like `field` and `value`/`values` (see https://github.com/chFlorian/firebase-ios-sdk/pull/1/commits/41114476945ce040df34e26c6113d35b7759eef3#diff-78e468ae8bbbcd59d12162d4ba9b4399e9a4aa2c48898f52cb83bbdb7a1a6c79R37).

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+    .isEqualTo(lhs, rhs)
+  }
+
+  public static func whereField(_ lhs: String, isIn rhs: [Any]) -> FIRPredicate {
```
peterfriese, line 46: See my comment above. This applies to all occurrences of `lhs` and `rhs` in this file.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright 2021 Google
```
peterfriese, line 2: This needs to read 

```
Copyright 2021 Google LLC
```

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+
+  private func removeListener() {
+    listener?.remove()
+  }
+}
+
+@available(iOS 14.0, *)
+@propertyWrapper
+public struct FirestoreQuery<T: Decodable>: DynamicProperty {
+  @StateObject private var store: FirestoreQueryObservable<T>
```
peterfriese, line 163: Rename `store` to `firestoreQueryObservable` or similar.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+  public static func whereField(_ field: String,
+                                isGreaterThanOrEqualTo value: Any) -> QueryPredicate {
+    .isGreaterThanOrEqualTo(field, value)
+  }
+
+  public static func order(by field: String, descending value: Bool = false) -> QueryPredicate {
+    .orderBy(field, value)
+  }
+
+  public static func limit(to field: Int) -> QueryPredicate {
```
schmidt-sebastian, line 85: We also need to add `limitToLast`.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
chFlorian: Absolutely right. Currently I set the items array to empty and print out errors.

mortenbekditlevsen: Would it be too strange to have  projectedValue with the following signature:
```
public var projectedValue: AnyPublisher<[Result<T, Error>], Never>
```
?
That is: If you're interested in getting errors, then you can get a publisher of an array of results.
That signature of course doesn't match top level errors - I don't know if it would be too hacky if a top level error was represented as an array containing a single failure result?

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+@propertyWrapper
+public struct FirestoreQuery<T: Decodable>: DynamicProperty {
+  @StateObject private var firestoreQueryObservable: FirestoreQueryObservable<T>
+
+  public var wrappedValue: [T] {
+    get {
+        firestoreQueryObservable.items
+    }
+  }
+
```
mortenbekditlevsen, line 170: In my opinion it would be beneficial to be able to override 'firestore` as well as the 'decoder' in use. Something like:

```
    public var firestore: Firestore {
        get { store.firestore }
        set { store.firestore = newValue }
    }

    public var decoder: Firestore.Decoder {
        get { store.decoder }
        set { store.decoder = newValue }
    }

```

and then in the `FirestoreQueryObservable`:
```
    fileprivate var firestore: Firestore = Firestore.firestore() {
        didSet {
            resubscribe()
        }
    }
    fileprivate var decoder: Firestore.Decoder = Firestore.Decoder() {
        didSet {
            resubscribe()
        }
    }

    private func resubscribe() {
        removeListener()
        self.items = []
        setupListener(from: collectionPath, withPredicates: predicates)
    }

```
This of course requires you to store the collectionPath and predicates on the `FirestoreQueryObservable`.
And the `removeListener` should then also nil out the listener...

What do you think?

peterfriese: Hi Morten, more configuration options add more flexibility, but also more liability. Can you elaborate why you'd want to be able to configure the Firestore instance and the decoder? This will help us decide whether to add this to this PR or add it later.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
chFlorian: Why would you use a Result type and set the error type to Never? Wouldn't it be much easier to use
`public var projectedValue: AnyPublisher<[T], FirebaseError>`
?

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
mortenbekditlevsen: Main reason is just parity between this and other versions of the same API (the API meaning: fetching data from a Firestore collection somehow).
If `Firestore.Decoder` implemented different 'strategies' like `JSONDecoder` and also `Database.Decoder`, then there would naturally be an even bigger reason.
The complexity could be alleviated a bit by only supporting overrides through the initializer - although property wrapper initializers are 'hard' to inject values into...

paulb777:
It would be great to get this PR merged.

Besides tests, what else needs to be done, what are the next steps, and are there any blockers?

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,176 @@
...
+  case isLessThan(_ field: String, _ value: Any)
+  case isGreaterThan(_ field: String, _ value: Any)
+
+  case isLessThanOrEqualTo(_ field: String, _ value: Any)
+  case isGreaterThanOrEqualTo(_ field: String, _ value: Any)
+
+  case orderBy(_ field: String, _ value: Bool)
+
+  case limitTo(_ field: Int)
+
```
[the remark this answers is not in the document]
chFlorian: What would be an advantage of this? I like how it feels, but I think this adds some unneeded complexity.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
chFlorian: > It would be great to get this PR merged.
> 
> Besides tests, what else needs to be done, what are the next steps, and are there any blockers?

@peterfriese What do you think here?

chFlorian:
> It would be great to get this PR merged.
> 
> Besides tests, what else needs to be done, what are the next steps, and are there any blockers?

@peterfriese What do you think here?

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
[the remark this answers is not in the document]
nidegen: Yeah, mostly it would just feel more correct to me.

peterfriese:
> It would be great to get this PR merged.
> 
> Besides tests, what else needs to be done, what are the next steps, and are there any blockers?

Flo and I worked on it today, and I think it is in good shape. A couple of things that are still missing:

- [ ] API docs
- [ ] tests
- [ ] sample app (I already started one [over here](https://github.com/chFlorian/firebase-ios-sdk/pull/1/files?file-filters%5B%5D=.export_metadata&file-filters%5B%5D=.json&file-filters%5B%5D=.overall_export_metadata&file-filters%5B%5D=.png&file-filters%5B%5D=.rules&file-filters%5B%5D=.swift&file-filters%5B%5D=No+extension&file-filters%5B%5D=dotfile#diff-08be69695d95b9e65d790d69ade5276e332585bb0d8d629cb9a6684b34c8beb5)
- [ ] API review (in progress)

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
paulb777: Perhaps adding the tests will help clarify an error handling strategy?

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -24,19 +24,26 @@ internal class FirestoreQueryObservable<T: Decodable>: ObservableObject {
...
   }
 
   deinit {
     removeListener()
   }
 
-  private func setupListener(from collectionPath: String,
-                             withPredicates predicates: [QueryPredicate]) {
-    var query: Query = firestore.collection(collectionPath)
+  private func setupListener() {
```
mortenbekditlevsen, line 43: The QueryPredicate and the built in construction of a query seems very powerful. Might that be extracted so it’s possible to use this with the basic firestore api? That can of course come later.

mortenbekditlevsen:
Regarding the chat about possibly vending arrays of errors as well as arrays of compactMapped items, it is at least possible to get pretty far by juggling around the generics and providing an implementation for both T: Decodable and U: Decodable, T == Result<U, Error>.
The only thing is that I can't get the 'setupListener' called when updating the configuration since I made `setupListener` generic.

Here's my experiments - and sorry for the incomplete code, I didn't have a lot of time to experiment:

```
import Combine
import Firebase
import FirebaseFirestore
import SwiftUI

@available(iOS 14.0, *)
@propertyWrapper
public struct FirestoreQuery<T>: DynamicProperty {
    @StateObject private var firestoreQueryObservable: FirestoreQueryObservable<T>

    public var wrappedValue: [T] {
        firestoreQueryObservable.items
    }

    public var projectedValue: FirestoreQueryConfiguration {
        get {
            firestoreQueryObservable.configuration
        }
        nonmutating set {
            firestoreQueryObservable.objectWillChange.send()
            firestoreQueryObservable.configuration = newValue
        }
    }

    public init(collectionPath: String, predicates: [QueryPredicate] = []) where T: Decodable {
        let configuration = FirestoreQueryConfiguration(path: collectionPath, predicates: predicates)

        _firestoreQueryObservable =
            StateObject(wrappedValue: FirestoreQueryObservable<T>(configuration: configuration))
    }

    public init<U: Decodable>(collectionPath: String, predicates: [QueryPredicate] = []) where T == Result<U, Error> {
        let configuration = FirestoreQueryConfiguration(path: collectionPath, predicates: predicates)

        _firestoreQueryObservable =
            StateObject(wrappedValue: FirestoreQueryObservable<T>(configuration: configuration))
    }

}

public struct FirestoreQueryConfiguration {
    public var path: String
    public var predicates: [QueryPredicate]
}

@available(iOS 13.0, *)
internal class FirestoreQueryObservable<T>: ObservableObject {
    @Published var items: [T] = []

    private let firestore = Firestore.firestore()
    private var listener: ListenerRegistration? = nil

    internal var configuration: FirestoreQueryConfiguration {
        didSet {
            removeListener()
            //      setupListener()
        }
    }

    init(configuration: FirestoreQueryConfiguration) where T: Decodable {
        self.configuration = configuration
        setupListener()
    }

    init<U: Decodable>(configuration: FirestoreQueryConfiguration) where T == Result<U, Error> {
        self.configuration = configuration
        setupListener()
    }


    deinit {
        removeListener()
    }

    private func setupListener() where T: Decodable {
        var query: Query = firestore.collection(configuration.path)

        for predicate in configuration.predicates {
            switch predicate {
            case let .isEqualTo(field, value):
                query = query.whereField(field, isEqualTo: value)
            case let .isIn(field, values):
                query = query.whereField(field, in: values)
            case let .isNotIn(field, values):
                query = query.whereField(field, notIn: values)
            case let .arrayContains(field, value):
                query = query.whereField(field, arrayContains: value)
            case let .arrayContainsAny(field, values):
                query = query.whereField(field, arrayContainsAny: values)
            case let .isLessThan(field, value):
                query = query.whereField(field, isLessThan: value)
            case let .isGreaterThan(field, value):
                query = query.whereField(field, isGreaterThan: value)
            case let .isLessThanOrEqualTo(field, value):
                query = query.whereField(field, isLessThanOrEqualTo: value)
            case let .isGreaterThanOrEqualTo(field, value):
                query = query.whereField(field, isGreaterThanOrEqualTo: value)
            case let .orderBy(field, value):
                query = query.order(by: field, descending: value)
            case let .limitTo(field):
                query = query.limit(to: field)
            case let .limitToLast(field):
                query = query.limit(toLast: field)
            }
        }

        listener = query.addSnapshotListener { [weak self] snapshot, error in
            if let error = error {
                print(error)
                self?.items = []
                return
            }

            guard let snapshot = snapshot else {
                print("FirestoreQuery: Registering the SnapshotListener returned a bad snapshot.")
                self?.items = []
                return
            }

            self?.items = snapshot.documents.compactMap { document in
                try? document.data(as: T.self)
            }
        }
    }

    private func setupListener<U: Decodable>() where T == Result<U, Error> {
        var query: Query = firestore.collection(configuration.path)

        for predicate in configuration.predicates {
            switch predicate {
            case let .isEqualTo(field, value):
                query = query.whereField(field, isEqualTo: value)
            case let .isIn(field, values):
                query = query.whereField(field, in: values)
            case let .isNotIn(field, values):
                query = query.whereField(field, notIn: values)
            case let .arrayContains(field, value):
                query = query.whereField(field, arrayContains: value)
            case let .arrayContainsAny(field, values):
                query = query.whereField(field, arrayContainsAny: values)
            case let .isLessThan(field, value):
                query = query.whereField(field, isLessThan: value)
            case let .isGreaterThan(field, value):
                query = query.whereField(field, isGreaterThan: value)
            case let .isLessThanOrEqualTo(field, value):
                query = query.whereField(field, isLessThanOrEqualTo: value)
            case let .isGreaterThanOrEqualTo(field, value):
                query = query.whereField(field, isGreaterThanOrEqualTo: value)
            case let .orderBy(field, value):
                query = query.order(by: field, descending: value)
            case let .limitTo(field):
                query = query.limit(to: field)
            case let .limitToLast(field):
                query = query.limit(toLast: field)
            }
        }

        listener = query.addSnapshotListener { [weak self] snapshot, error in
            if let error = error {
                print(error)
                self?.items = []
                return
            }

            guard let snapshot = snapshot else {
                print("FirestoreQuery: Registering the SnapshotListener returned a bad snapshot.")
                self?.items = []
                return
            }

            self?.items = snapshot.documents.map { document in
                Result {
                    try document.data(as: U.self)!
                }
            }
        }
    }


    private func removeListener() {
        listener?.remove()
        listener = nil
    }
}

public enum QueryPredicate {
    case isEqualTo(_ field: String, _ value: Any)

    case isIn(_ field: String, _ values: [Any])
    case isNotIn(_ field: String, _ values: [Any])

    case arrayContains(_ field: String, _ value: Any)
    case arrayContainsAny(_ field: String, _ values: [Any])

    case isLessThan(_ field: String, _ value: Any)
    case isGreaterThan(_ field: String, _ value: Any)

    case isLessThanOrEqualTo(_ field: String, _ value: Any)
    case isGreaterThanOrEqualTo(_ field: String, _ value: Any)

    case orderBy(_ field: String, _ value: Bool)

    case limitTo(_ value: Int)
    case limitToLast(_ value: Int)

    /*
     Factory methods to expose the underlying enum cases with a nicer development experience and improved semantics.
     */
    public static func whereField(_ field: String, isEqualTo value: Any) -> QueryPredicate {
        .isEqualTo(field, value)
    }

    public static func whereField(_ field: String, isIn values: [Any]) -> QueryPredicate {
        .isIn(field, values)
    }

    public static func whereField(_ field: String, isNotIn values: [Any]) -> QueryPredicate {
        .isNotIn(field, values)
    }

    public static func whereField(_ field: String, arrayContains value: Any) -> QueryPredicate {
        .arrayContains(field, value)
    }

    public static func whereField(_ field: String,
                                  arrayContainsAny values: [Any]) -> QueryPredicate {
        .arrayContainsAny(field, values)
    }

    public static func whereField(_ field: String, isLessThan value: Any) -> QueryPredicate {
        .isLessThan(field, value)
    }

    public static func whereField(_ field: String, isGreaterThan value: Any) -> QueryPredicate {
        .isGreaterThan(field, value)
    }

    public static func whereField(_ field: String,
                                  isLessThanOrEqualTo value: Any) -> QueryPredicate {
        .isLessThanOrEqualTo(field, value)
    }

    public static func whereField(_ field: String,
                                  isGreaterThanOrEqualTo value: Any) -> QueryPredicate {
        .isGreaterThanOrEqualTo(field, value)
    }

    public static func order(by field: String, descending value: Bool = false) -> QueryPredicate {
        .orderBy(field, value)
    }

    public static func limit(to value: Int) -> QueryPredicate {
        .limitTo(value)
    }

    public static func limit(toLast value: Int) -> QueryPredicate {
        .limitToLast(value)
    }
}

struct Test: Decodable {}

@available(iOS 14.0, *)
struct Foo: View {
    @FirestoreQuery(collectionPath: "bar") var bar: [Result<Test, Error>]
    @FirestoreQuery(collectionPath: "baz") var baz: [Test]

    var body: some View {
        Text("foo")
    }
}


```

mortenbekditlevsen:
The `setupListener` can of course be captured in a type erased closure, but let me know if you think this exploration is worth continuing. 😊

paulb777:
Please add instructions for setting up and running the sample to make it easier for others to review.

peterfriese:
> Please add instructions for setting up and running the sample to make it easier for others to review.

I need to rebase the sample on the current implementation anyway, I'll use this opportunity to add a launch script for the local emulator, as well as a short README.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,101 @@
...
+@available(iOS 14.0, *)
+@available(macOS 11.0, *)
+@available(tvOS 14.0, *)
+@propertyWrapper
+public struct FirestoreQuery<T: Decodable>: DynamicProperty {
+  @StateObject private var firestoreQueryObservable: FirestoreQueryObservable<T>
+
+  /// The query's configurable properties.
+  public struct Configuration {
+    /// The query's collectin path.
```
paulb777, line 65: nit: collection

Example/FirestoreSample/firestore.rules
```diff
@@ -0,0 +1,16 @@
...
+      // This rule allows anyone with your database reference to view, edit,
+      // and delete all data in your database. It is useful for getting
+      // started, but it is configured to expire after 30 days because it
+      // leaves your app open to attackers. At that time, all client
+      // requests to your database will be denied.
+      //
+      // Make sure to write security rules for your app before that time, or
+      // else all client requests to your database will be denied until you
+      // update your rules.
+      allow read, write: if request.time < timestamp.date(2021, 9, 2);
```
paulb777, line 13: will need an update next week

Example/FirestoreSample/.firebaserc
```diff
@@ -0,0 +1 @@
+{}
```
paulb777, line 1: I believe `Firestore/Swift/Tests/FirestoreSample` would be a more consistent location with the rest of the repo than `Example/FirestoreSample`.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,101 @@
...
+/// to modify the query criteria. Changing the filter predicates results in the
+/// underlying snapshot listener being unregistered and a new one registered.
+///
+///     Button("Show only Apples and Oranges") {
+///       $fruits.predicates = [.whereField("name", isIn: ["Apple", "Orange]]
+///     }
+///
+/// This property wrapper does not support updating the `wrappedValue`, i.e.
+/// you need to use Firestore's other APIs to add, delete, or modify documents.
+@available(iOS 14.0, *)
```
paulb777, line 56: Why 14 and not 13?

Firestore/Swift/Source/PropertyWrapper/QueryPredicate.swift
```diff
@@ -0,0 +1,137 @@
...
+    .limitTo(value)
+  }
+
+  public static func limit(toLast value: Int) -> QueryPredicate {
+    .limitToLast(value)
+  }
+
+  // Alternate naming
+
+  public static func `where`(field name: String, isEqualTo value: Any) -> QueryPredicate {
```
paulb777, line 101: Just learned that the backquotes enable a reserved word to be used as an identifier. 😄

Example/FirestoreSample/.gitignore
```diff
@@ -0,0 +1,66 @@
+# Logs
```
paulb777, line 1: Is this `.gitignore` providing anything needed not covered by the repo level `.gitignore`?

 We've had a policy of keeping all `.gitignore`s  at the top of the repo. It might be ok to make an exception for Samples, but I'm not sure it's needed here.

paulb777 (commented):
This is looking good! My understanding is the biggest open questions are API review and error handling. 

I'm looking forward to getting it merged.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
chFlorian: Because of the StateObject. That only became available with iOS 14.

Example/FirestoreSample/.gitignore
peterfriese: This one is generated by `firebase init`, I believe. We can remove it, and add any patterns that are relevant to the top level ignore file.

paulb777:
Closing in favor of #8614.  @chFlorian Thanks for getting this rolling!

PR #8414, opened 2021-07-20
Migrate SpecRepoBuilder tool

granluo:
Move the tool under the ReleaseTooling so that the tool is able to fetch the manifest.

### DIFF
--- a/ReleaseTooling/Sources/SpecRepoBuilder/main.swift
+++ b/ReleaseTooling/Sources/SpecRepoBuilder/main.swift
@@ -306,19 +306,19 @@ struct SpecRepoBuilder: ParsableCommand {
   }
 
   // Create `pod repo push` flags based on the manifest.
-  func createSpecPushFlags(of sdk:String, from manifest: [Pod]) -> [String]{
-      if sdk == "Firebase" {
-          return Constants.umbrellaPodFlags
+  func createSpecPushFlags(of sdk: String, from manifest: [Pod]) -> [String] {
+    if sdk == "Firebase" {
+      return Constants.umbrellaPodFlags
+    }
+    var flags = Constants.flags
+    for podConfig in manifest {
+      if sdk == podConfig.name {
+        if podConfig.allowWarnings {
+          flags.append("--allow-warnings")
+        }
       }
-      var flags = Constants.flags
-      for podConfig in manifest{
-          if sdk == podConfig.name{
-              if podConfig.allowWarnings{
-                  flags.append( "--allow-warnings")
-              }
-          }
-      }
-      return flags
+    }
+    return flags
   }
 
   mutating func run() throws {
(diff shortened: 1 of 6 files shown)


ReleaseTooling/Sources/SpecRepoBuilder/main.swift
```diff
@@ -0,0 +1,396 @@
+/*
+ * Copyright 2020 Google LLC
```
paulb777, line 2: 2021

ReleaseTooling/Sources/SpecRepoBuilder/main.swift
```diff
@@ -0,0 +1,396 @@
...
+  ]
+}
+
+// flags for 'pod push'
+extension Constants {
+  static let flags = ["--skip-tests"]
+  static let umbrellaPodFlags = Constants.flags + ["--skip-import-validation", "--use-json"]
+}
+
+// SpecFiles is a wraper of dict mapping from required pods to their path. This
```
paulb777, line 40: sp: wrapper

ReleaseTooling/Sources/ZipBuilder/main.swift
```diff
@@ -236,11 +236,8 @@ struct ZipBuilderTool: ParsableCommand {
                                            logsOutputDir: outputDir?
                                              .appendingPathComponent("build_logs"))
 
-    // Populate the platforms list if it's empty. This isn't a great spot, but the argument parser
```
paulb777, line 239: Was this fixed in 0.3 of the Argument Parser?

scripts/release_testing_setup.sh
```diff
@@ -66,4 +66,9 @@ elif [ "$TESTINGMODE" = "prerelease_testing" ]; then
   # Update source and tag, e.g.  ":tag => 'CocoaPods-' + s.version.to_s" to
   # ":tag => ${test_version}.nightly"
   sed  -i "" "s/\s*:tag.*/:tag => '${tag_version}'/" *.podspec
+  # Cocoapods is throwing warnings version not tagged to the repo. This warning
+  # will not affect test but still fail the workflow. To avoid this warning,
+  # the version will be updated to be the same as tag.
```
paulb777, line 71: +1

paulb777 (commented):
Here's an initial review. I'd like to another when CI finishes.

ReleaseTooling/Sources/ZipBuilder/main.swift
granluo: Yes this is fixed in 0.3. Also, default setup for parameters are  updated as well.
Instead of using `default: "default_value`, a default value could be assign to an option. So not just array, string/int could be assigned directly.

granluo:
> Here's an initial review. I'd like to another when CI finishes.

It turns out the warning still exists but `pod repo push` is working fine locally even the version is not in the git tag.
```
    - WARN  | source: The version should be included in the Git tag.

```
is not showing up locally. Will investigate what the version refer to.

paulb777:
Closing for staleness. Please reopen if still needed.

PR #8420, opened 2021-07-21
Pod binary size measurement

granluo:


paulb777:
The changed files look like a mix of code coverage and size changes. Is that intentional?  Would you add a description to the PR to summarize its intended scope.

granluo:
> The changed files look like a mix of code coverage and size changes. Is that intentional? Would you add a description to the PR to summarize its intended scope.

Will update the dir name then, the binary size metrics is sharing the same workflow with binary size.
The job `check` is to determine SDKs to run binary size and code coverage, so the workflow is not only for code coverage then.

paulb777:
I see. It generally looks good to me with renaming the directory. Is it possible to see example output anywhere?

granluo:
> I see. It generally looks good to me with renaming the directory. Is it possible to see example output anywhere?

Yes, In the `binary_size_metrics` job of this [run](https://github.com/firebase/firebase-ios-sdk/runs/3145844656), you can see 
```
{"results":[{"sdk":"FirebaseABTesting","type":"firebase-ios-sdk-testing","value":350528},{"sdk":"FirebaseAuth","type":"firebase-ios-sdk-testing","value":1031136}],"metric":"BinarySize","log":"testing.log"}
```
This run triggered ABT and auth on purpose to test the tool. The json data is the one will be send in an API request to the Metrics Service. The log is long and , instead of scrolling down, you can get into the raw log to search the data above.

granluo:
Created another PR #8539 due to outdated files.

PR #8422, opened 2021-07-21
Auth integration with App Multicast Delegate

maksymmalyhin:
b/193554796


FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -77,7 +77,14 @@
 #import "FirebaseAuth/Sources/SystemService/FIRAuthAppCredentialManager.h"
 #import "FirebaseAuth/Sources/SystemService/FIRAuthNotificationManager.h"
 #import "FirebaseAuth/Sources/Utilities/FIRAuthURLPresenter.h"
-#endif
+
+#if SWIFT_PACKAGE
```
paulb777, line 81: Structure public headers as documented in https://github.com/firebase/firebase-ios-sdk/blob/master/HeadersImports.md#header-file-types-and-locations---for-header-file-creators to avoid need for SPM/CocoaPods specific imports.

Package.swift
```diff
@@ -153,7 +153,7 @@ let package = Package(
     .package(
       name: "GoogleUtilities",
       url: "https://github.com/google/GoogleUtilities.git",
-      "7.4.0" ..< "8.0.0"
+      .branch("mm/app-delegate-sample")
```
paulb777, line 156: This should not merge to master.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -482,7 +489,14 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
     }
     UIApplication *application = [applicationClass sharedApplication];
 
-    [GULAppDelegateSwizzler proxyOriginalDelegateIncludingAPNSMethods];
+    // Use Multicast App Delegate when available but fallback to App Delegate Swizzler for backward compatibility.
+    id<GULMulticastAppDelegateProtocol> multicastDelegate =
```
paulb777, line 493: Why is this in iOS specific code?

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -482,7 +489,14 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey appName:(NSString *)a
...
+        [GULMulticastAppDelegate installedMulticastDelegate];
+    if (multicastDelegate == nil) {
+      FIRLogInfo(kFIRLoggerAuth, @"I-AUT000018", @"Multicast App delegate was not detected. Make sure app events forwarding is configured manually for you app. <Add link to docs>");
```
paulb777, line 496: nit: long line

FirebaseAuth/Sources/Auth/FIRAuth.m
ryanwilson: My guess is it's due to phone auth not being available on other platforms, but that was out of ease-of-implementation not any product decisions.

FirebaseAuth/Sources/Auth/FIRAuth.m
maksymmalyhin: Hm, not sure I understand how it's applicable here because there are no headers in the lib currently, but only a single Swift file. But probably it won't be relevant any more after https://github.com/google/GoogleUtilities/pull/50#discussion_r674200503 since we will have to re-implement the multicast delegate in Objective-C.

paulb777: Oh right. I made this comment before I saw the implementation was Swift.

maksymmalyhin: TODO: Validate in Firebase

paulb777:
Closing since we no longer plan to do the multicast delegate

PR #8424, opened 2021-07-22
Fix leak in reference value parsing

schmidt-sebastian:
This fixes the last leak in the field value port.

### DIFF
--- a/Firestore/core/src/api/query_core.cc
+++ b/Firestore/core/src/api/query_core.cc
@@ -255,7 +255,7 @@ Query Query::Filter(const FieldPath& field_path,
             return *ParseExpectedReferenceValue(value, type_describer)
                         .release();
           });
-      value = references;
+      value = std::move(references);
     } else {
       value = ParseExpectedReferenceValue(*value, type_describer);
     }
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
FYI - clang format failed. I will fix in the feature branch since I need to merge `master`.

Firestore/core/src/api/query_core.cc
```diff
@@ -244,17 +244,18 @@ Query Query::Filter(const FieldPath& field_path,
...
+          &references->array_value.values,
+          &references->array_value.values_count,
           absl::Span<google_firestore_v1_Value>(
               value->array_value.values, value->array_value.values_count),
           [&](const google_firestore_v1_Value& value) {
             return *ParseExpectedReferenceValue(value, type_describer)
                         .release();
           });
-      value = MakeSharedMessage(references);
+      value = references;
```
var-const, line 258: Nit: you can move `references`.

var-const (approved):

Firestore/core/src/api/query_core.cc
schmidt-sebastian: done

PR #8426, opened 2021-07-22
Use GoogleAppMeasurement main branch instead of release tag for CI

paulb777:
Use GoogleAppMeasurement main branch instead of release tag for CI.  This enables us to use CI for release tagging without tagging the GoogleAppMeasurement repository until release and avoid needing to make last minute commits to the firebase repo to update to the new tag. b/194422721

### DIFF
--- a/scripts/setup_spm_test_app_measurement.sh
+++ b/scripts/setup_spm_test_app_measurement.sh
@@ -20,4 +20,3 @@
 # until after testing.
 
 sed -i '' 's#exact("[0-9.]*#branch("main#' Package.swift
- 
--- a/scripts/setup_spm_test_app_measurement.sh
+++ b/scripts/setup_spm_test_app_measurement.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


scripts/setup_spm_test_app_measurement.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Point SPM CI to the tip of main of https://github.com/google/GoogleAppMeasurement
+# so that the release process can defer publish the GoogleAppMeasurement tag
+# until after testing.
+
+sed -i '' 's#exact("[0-9.]*#branch("main#' Package.swift
```
maksymmalyhin, line 22: An example of input and output would be really helpful for future us.

maksymmalyhin (approved):
LGTM.

The only question is if we should use `main` unconditionally or there are cases when we should keep using a specific version? If we don't know exact conditions we may want to have a way to specify either a branch or a tag or not change package manifest for a manual workflow trigger.

scripts/setup_spm_test_app_measurement.sh
paulb777: Good idea. Done!

paulb777:
I'm proposing that we use `main` unconditionally. It should always work in CI. The biggest issue will be local SPM development after the new version is updated and before it is published will need to run the script locally.

I'll add a step to go/firi to verify that the Package.swift successfully loads after publishing the GoogleAppMeasurement tag.

maksymmalyhin:
> I'm proposing that we use `main` unconditionally. It should always work in CI. The biggest issue will be local SPM development after the new version is updated and before it is published will need to run the script locally.
> 
> I'll add a step to go/firi to verify that the Package.swift successfully loads after publishing the GoogleAppMeasurement tag.

SGTM, though I would still prefer to reduce space for an error with some automatic validation. I'm thinking, maybe we can set up a pre-push hook in our local git repos to validate pushed tags, e.g. when a SPM tag push is detected in Firebase repo then SPM build is run locally before pushing. It's perhaps outside of scope of this PR, but something to consider for future.

PR #8436, opened 2021-07-26
MetricKit SDK implementation

elenadoty:


Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+  [[MXMetricManager sharedManager] addSubscriber:self];
+  _metricKitDataAvailable = [FBLPromise pendingPromise];
+
+  // If there was no crash on the last run of the app, then we aren't expecting a MetricKit
+  // diagnostic report and should resolve the promise immediately.
+  NSString *crashedMarkerFileName = [NSString stringWithUTF8String:FIRCLSCrashedMarkerFileName];
+  NSString *crashedMarkerFileFullPath =
+      [[_fileManager rootPath] stringByAppendingPathComponent:crashedMarkerFileName];
+  if (![_fileManager fileExistsAtPath:crashedMarkerFileFullPath]) {
+    [_mutex lock];
```
samedson, line 63: For all these, you can use `@synchronized (self) { ... }` blocks instead of a mutex

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+#import "Crashlytics/Crashlytics/Models/FIRCLSExecutionIdentifierModel.h"
+#import "Crashlytics/Crashlytics/Models/FIRCLSInternalReport.h"
+#import "Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlyticsReport.h"
+
+@interface FIRCLSMetricKitManager () {
+  NSLock *_mutex;
+  FBLPromise *_metricKitDataAvailable;
+  FIRCLSExistingReportManager *_existingReportManager;
+  FIRCLSFileManager *_fileManager;
+  FIRCLSManagerData *_managerData;
```
samedson, line 30: For all these variables (except mutex), we're tryna move away from this style and move towards using `@property`. Check out `FIRCLSExistingReportManager` as an example. For usage of these variables, keep using `_varName` in the `init` method, but for the rest of the instances replace it with `self.varName`

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+ */
+- (void)registerMetricKitManager {
+  [[MXMetricManager sharedManager] addSubscriber:self];
+  _metricKitDataAvailable = [FBLPromise pendingPromise];
+
+  // If there was no crash on the last run of the app, then we aren't expecting a MetricKit
+  // diagnostic report and should resolve the promise immediately.
+  NSString *crashedMarkerFileName = [NSString stringWithUTF8String:FIRCLSCrashedMarkerFileName];
+  NSString *crashedMarkerFileFullPath =
+      [[_fileManager rootPath] stringByAppendingPathComponent:crashedMarkerFileName];
```
samedson, line 61: Can you turn these 2 lines into a method on `fileManager`, and replace this instance and the FIRCrashlytics.m one with the new method you make?

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+                 payloadCount);
+
+  if ([payloads count] == 0) {
+    return;
+  }
+
+  MXDiagnosticPayload *diagnosticPayload = [payloads objectAtIndex:0];
+  if (!diagnosticPayload) {
+    return;
+  }
```
samedson, line 90: Might want to iterate here? Maybe make a new method and put the rest of this method in it, with a single payload as an arg.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+  // Get file path for the active reports directory.
+  NSString *activePath = [[_fileManager activePath] stringByAppendingString:@"/"];
+
+  // If newestReportID is nil or the file no longer exists, then this callback function was
+  // triggered for nonfatal MetricKit events.
+  NSString *newestUnsentReportID =
+      [_existingReportManager.newestUnsentReport.reportID stringByAppendingString:@"/"];
+  NSString *metricKitReportFile;
+  BOOL fatal =
+      (newestUnsentReportID != nil) && ([_fileManager fileExistsAtPath:metricKitReportFile]);
```
samedson, line 106: I don't think `newestUnsentReportID` will ever be nil because you append a slash to it. Can just check `self.existingReportManager.newestUnsentReport` is nil

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+  // TODO: Time stamp information is only available in begin and end time periods. Hopefully this is
+  // updated with iOS 15.
+  NSTimeInterval beginSecondsSince1970 = [diagnosticPayload.timeStampBegin timeIntervalSince1970];
+  NSTimeInterval endSecondsSince1970 = [diagnosticPayload.timeStampEnd timeIntervalSince1970];
+
+  // Get file path for the active reports directory.
+  NSString *activePath = [[_fileManager activePath] stringByAppendingString:@"/"];
+
+  // If newestReportID is nil or the file no longer exists, then this callback function was
+  // triggered for nonfatal MetricKit events.
```
samedson, line 101: Couldn't it be triggered for non-fatals even if there was a crash on the previous run? Eg. if they happened to keep the crash report on disk (maybe cause they didn't call `sendUnsentReports`), and then had a hang?

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,311 @@
...
+    NSString *currentReportID =
+        [_managerData.executionIDModel.executionID stringByAppendingString:@"/"];
+    metricKitReportFile = [[activePath stringByAppendingString:currentReportID]
+        stringByAppendingString:FIRCLSMetricKitNonfatalReportFile];
+  } else {
+    metricKitReportFile = [[activePath stringByAppendingString:newestUnsentReportID]
+        stringByAppendingString:FIRCLSMetricKitFatalReportFile];
+  }
+
+  // Need to copy the string passed to the log method or it causes a crash (GUI logging issue).
```
samedson, line 123: Super duper nit: `GUL` logging issue

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
samedson: Can we just use the existence of `crashDiagnostics` here? It is tricky - like would you ever get a crash diagnostic with other non-fatal ones? I wonder if we can assume crash ones go with the last report, while non-fatal ones go with the current

Crashlytics/Crashlytics/Components/FIRCLSProcess.c
```diff
@@ -524,19 +527,24 @@ void FIRCLSProcessRecordThreadNames(FIRCLSProcess *process, FIRCLSFile *file) {
...
+  FIRCLSFileWriteSectionEnd(metricKitFile);
 }
 
-void FIRCLSProcessRecordDispatchQueueNames(FIRCLSProcess *process, FIRCLSFile *file) {
+void FIRCLSProcessRecordDispatchQueueNames(FIRCLSProcess *process, FIRCLSFile *file, FIRCLSFile *metricKitFile) {
   uint32_t threadCount;
   uint32_t i;
 
   FIRCLSFileWriteSectionStart(file, "dispatch_queue_names");
+  FIRCLSFileWriteSectionStart(metricKitFile, "dispatch_queue_names");
```
samedson, line 544: So thinking more about this, I feel like we should skip the part about writing thread names and not cross streams like this. If MetricKit doesn't provide it then we should just go with what they provide. If we want to do this, wait for customer feedback or solve it later. I feel like this is unintuitive implementation-wise, and seems like issues could come up with timing or missing MetricKit reports (eg. if we require thread names to process them successfully, and there isn't a Mach Exception signal triggered).

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
elenadoty: Good callout - I like that approach, I'm just unsure whether we'd see an exception and a crash diagnostic for the same event, because in that case we wouldn't want to separate them out.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
elenadoty: So it turns out that it can be nil, the append fails if the original string is nil

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
elenadoty: Going to use a combination of checking for a crash diagnostic and looking at the last report id

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
elenadoty: My worry with that is that each type of payload has different metadata fields, so we'd end up with a lot of the same code in that new method anyway

samedson: But is it possible that this `payloads` array has more than one element? Even if each element has multiple types of payloads themselves?

samedson: We humaned - gonna iterate, but only take the first or last crash event

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -207,6 +221,24 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData
...
+ * This method returns a promise that is resolved once
+ * MetricKit diagnostic reports have been received by `metricKitManager`.
+ */
+- (FBLPromise *)waitForMetricKitData {
+  // If the platform is not iOS or the iOS version is less than 15, immediately resolve the promise
+  // since no MetricKit diagnostics will be available.
+#if CLS_METRICKIT_SUPPORTED
+  if (@available(iOS 15, *)) {
+    FBLPromise *metricKitDataAvailable = [self.metricKitManager waitForMetricKitDataAvailable];
+    return metricKitDataAvailable;
```
samedson, line 234: Nit: don't need the variable here

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -107,6 +108,9 @@ @interface FIRCLSReportManager () {
   // for all the unsent reports.
   FBLPromise<FIRCLSWrappedReportAction *> *_reportActionProvided;
 
+  // A promise that will be resolved when we've received diagnostic reports from MetricKit.
+  FBLPromise *_metricKitDiagnosticsAvailable;
```
samedson, line 112: Seems unused

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,320 @@
...
+ */
+- (void)registerMetricKitManager {
+  [[MXMetricManager sharedManager] addSubscriber:self];
+  self.metricKitDataAvailable = [FBLPromise pendingPromise];
+
+  // If there was no crash on the last run of the app, then we aren't expecting a MetricKit
+  // diagnostic report and should resolve the promise immediately.
+  if (![self.fileManager didCrashOnPreviousExecution] && ![self.fileManager didCrashPreviously]) {
+    @synchronized(self) {
+      [self.metricKitDataAvailable fulfill:nil];
```
samedson, line 63: So this is really tricky, but I'm worried about this block of code for 2 reasons:
 - If we don't detect the crash and don't write out the marker file, but MetricKit does, we won't capture the MetricKit crash, so we might miss out on events.
 - If we detect a crash by writing out the marker file, but MetricKit does NOT, then this doesn't look like it'll resolve, and we won't upload crash reports.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,320 @@
...
+/*
+ * This method receives diagnostic payloads from MetricKit whenever a fatal or nonfatal MetricKit
+ * event occurs. If a fatal event, this method will be called when the app restarts. Since we're
+ * including a MetricKit report file in the Crashlytics report to be sent to the backend, we need
+ * to make sure that we process the payloads and write the included information to file before
+ * the report is sent up. If this method is called due to a nonfatal event, it will be called
+ * immediately after the event. Since we send nonfatal events on the next run of the app, we can
+ * write out the information but won't need to resolve the promise.
+ */
+- (void)didReceiveDiagnosticPayloads:(NSArray<MXDiagnosticPayload *> *)payloads {
```
samedson, line 77: How hard would it be to write tests for this method?

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,320 @@
...
+  if ([payloads count] == 0) {
+    return;
+  }
+
+  for (MXDiagnosticPayload *diagnosticPayload in payloads) {
+    if (!diagnosticPayload) {
+      return;
+    }
+
+    // TODO: Time stamp information is only available in begin and end time periods. Hopefully this
```
samedson, line 91: Can we move everything from this line to 224 to a new method? We return early from this method in some cases, which means we don't resolve the promise and don't upload reports. We can have it return success or failure instead.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,320 @@
...
+    }
+  }
+}
+
+/*
+ * Required for MXMetricManager subscribers. Since we aren't currently collecting any MetricKit
+ * metrics, this method is left empty.
+ */
+- (void)didReceiveMetricPayloads:(NSArray<MXMetricPayload *> *)payloads {
+  FIRCLSDebugLog(@"[Crashlytics:Crash] received %d MetricKit metric payloads\n", payloads.count);
```
samedson, line 244: Should we log if we're not doing anything with it? Will we or customers use this?

Crashlytics/Crashlytics/FIRCrashlytics.m
```diff
@@ -147,14 +147,12 @@ - (instancetype)initWithApp:(FIRApp *)app
                                                      analyticsManager:_analyticsManager];
 
     // Process did crash during previous execution
-    NSString *crashedMarkerFileName = [NSString stringWithUTF8String:FIRCLSCrashedMarkerFileName];
-    NSString *crashedMarkerFileFullPath =
-        [[_fileManager rootPath] stringByAppendingPathComponent:crashedMarkerFileName];
-    _didPreviouslyCrash = [_fileManager fileExistsAtPath:crashedMarkerFileFullPath];
-
-    if (_didPreviouslyCrash) {
+    if ([_fileManager didCrashOnPreviousExecution]) {
```
samedson, line 150: We're no longer setting `self.didPreviouslyCrash`, which means the public API `didCrashDuringPreviousExecution` is gonna be wrong

Crashlytics/Crashlytics/Models/FIRCLSFileManager.h
```diff
@@ -55,6 +56,7 @@
 - (BOOL)removeItemAtPath:(NSString *)path;
 - (BOOL)removeContentsOfDirectoryAtPath:(NSString *)path;
 - (BOOL)moveItemAtPath:(NSString *)path toDirectory:(NSString *)destDir;
+- (BOOL)didCrashOnPreviousExecution;
```
samedson, line 59: What's the reason for having both `didCrashPreviously` and `didCrashOnPreviousExecution`?

samedson: Can we call this `crashMarkerFileExists` so we don't confuse it with the public API `didCrashDuringPreviousExecution`

Crashlytics/Crashlytics/Models/FIRCLSFileManager.m
```diff
@@ -124,6 +126,15 @@ - (BOOL)moveItemAtPath:(NSString *)path toDirectory:(NSString *)destDir {
   return YES;
 }
 
+- (BOOL)didCrashOnPreviousExecution {
+  NSString *crashedMarkerFileName = [NSString stringWithUTF8String:FIRCLSCrashedMarkerFileName];
+  NSString *crashedMarkerFileFullPath =
+      [[self rootPath] stringByAppendingPathComponent:crashedMarkerFileName];
+  self.didCrashPreviously =
+      self.didCrashPreviously || [self fileExistsAtPath:crashedMarkerFileFullPath];
```
samedson, line 134: If you want to memoize this you can use dispatch_once eg: https://github.com/firebase/firebase-ios-sdk/blob/0d73868de6df07272705d98c98950e8f82b8601b/Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m#L386-L389

Crashlytics/Crashlytics/Models/FIRCLSFileManager.h
elenadoty: Since we call didCrashOnPreviousExecution twice and remove the file marker the first time it's called, it won't be correct the second time.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
elenadoty: If we don't detect the crash we would still receive the diagnostic payload from MetricKit, we just wouldn't send it up this run of the app (it would be treated as a non-fatal, in terms of timing). I'm not sure I understand what you mean in the second point though - the registerMetricKitManager method will be called whenever the app opens, regardless of if something crashed.

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
samedson: Thinking we should remove the log cause it might confuse customers.

Crashlytics/Crashlytics/Models/FIRCLSFileManager.h
```diff
@@ -46,6 +46,7 @@
 @property(nonatomic, readonly) NSArray *activePathContents;
 @property(nonatomic, readonly) NSArray *preparedPathContents;
 @property(nonatomic, readonly) NSArray *processingPathContents;
+@property(nonatomic) BOOL crashFileMarkerExists;
```
samedson, line 49: You can put this in the `.m` file's interface so it's not public

Crashlytics/Crashlytics/Models/FIRCLSFileManager.m
```diff
@@ -124,6 +126,17 @@ - (BOOL)moveItemAtPath:(NSString *)path toDirectory:(NSString *)destDir {
   return YES;
 }
 
+- (BOOL)didCrashOnPreviousExecution {
+  NSString *crashedMarkerFileName = [NSString stringWithUTF8String:FIRCLSCrashedMarkerFileName];
+  NSString *crashedMarkerFileFullPath =
+      [[self rootPath] stringByAppendingPathComponent:crashedMarkerFileName];
+  static dispatch_once_t checkCrashFileMarketExistsOnceToken;
+  dispatch_once(&checkCrashFileMarketExistsOnceToken, ^{
```
samedson, line 134: I would actually wrap the rest of this method (besides the return) so we don't pay the cost of file lookups more than once

Crashlytics/Crashlytics/Controllers/FIRCLSMetricKitManager.m
```diff
@@ -0,0 +1,322 @@
...
+      continue;
+    }
+
+    BOOL processedPayload = [self processMetricKitPayload:diagnosticPayload];
+    BOOL fatal = processedPayload && ([diagnosticPayload.crashDiagnostics count] > 0);
+    if (fatal) {
+      // Since we only want to handle one crash per report, if this was a fatal event
+      // don't process any additional payloads.
+      break;
+    }
```
samedson, line 91: Let's sync on this

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -178,6 +182,13 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData
                                                            googleAppID:self.googleAppID];
 
   _notificationManager = [[FIRCLSNotificationManager alloc] init];
+#if CLS_METRICKIT_SUPPORTED
+  if (@available(iOS 15, *) && _settings.metricKitCollectionEnabled) {
```
samedson, line 186: nit: self.settings

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -207,6 +218,23 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData
...
 
+/*
+ * This method returns a promise that is resolved once
+ * MetricKit diagnostic reports have been received by `metricKitManager`.
+ */
+- (FBLPromise *)waitForMetricKitData {
+  // If the platform is not iOS or the iOS version is less than 15, immediately resolve the promise
+  // since no MetricKit diagnostics will be available.
+#if CLS_METRICKIT_SUPPORTED
+  if (@available(iOS 15, *) && _settings.metricKitCollectionEnabled) {
```
samedson, line 229: Same here

samedson: Do we have any debug logs that print when MetricKit is just enabled? Might be good to put one here just so we know when customers write in

samedson (approved):
Just a couple really small things

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
elenadoty: Would it be better to put the log where we initially initialize the metrickit manager?

elenadoty: Put it in the initial part (can move later if needed).

PR #8443, opened 2021-07-27
Prevent firelog from logging invalid data

chliangGoogle:


FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -211,8 +211,25 @@ - (void)exportDeliveryMetricsToBigQueryWithMessageInfo:(NSDictionary *)info {
...
   fm_MessagingClientEvent clientEvent = fm_MessagingClientEvent_init_default;
+  if (!info[kFIRMessagingSenderID]) {
+    FIRMessagingLoggerError(kFIRMessagingServiceExtensionInvalidProjectID,
+                            @"Delivery logging failed: Invalid project ID");
+    return;
+  }
   clientEvent.project_number = (int64_t)[info[kFIRMessagingSenderID] longLongValue];
+
+  if (!info[kFIRMessagingMessageIDKey]) {
+    FIRMessagingLoggerWarn(kFIRMessagingServiceExtensionInvalidMessageID,
```
maksymmalyhin, line 222: Seems like this code path is not covered by tests. I would suggest adding a test/tests for each condition.

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -211,8 +211,25 @@ - (void)exportDeliveryMetricsToBigQueryWithMessageInfo:(NSDictionary *)info {
   fm_MessagingClientEventExtension eventExtension = fm_MessagingClientEventExtension_init_default;
 
   fm_MessagingClientEvent clientEvent = fm_MessagingClientEvent_init_default;
+  if (!info[kFIRMessagingSenderID]) {
```
maksymmalyhin, line 214: optional: I usually prefer using a local var in such cases like:
```
NSString *senderID = info[kFIRMessagingSenderID];
if(senderID.length == 0) {
...
}
clientEvent.project_number = (int64_t)[info[kFIRMessagingSenderID] longLongValue];
```

It makes relation between validation and usage more clear.

maksymmalyhin (approved):
LGTM with a couple minor comments

maksymmalyhin (approved):

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
chliangGoogle: what if it's not a string? I worry it might still pass and crash the protobuf.

maksymmalyhin: The validation is just an example, my main point here is that declaring a local variable is more readable and change proof than accessing a value from the dict by key. I started doing it after I had changed a key in one place but forgot to change at another by accident.

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
chliangGoogle: added tests

FirebaseMessaging/Sources/FIRMessagingExtensionHelper.m
```diff
@@ -211,8 +211,26 @@ - (void)exportDeliveryMetricsToBigQueryWithMessageInfo:(NSDictionary *)info {
...
+    return;
+  }
   clientEvent.project_number = (int64_t)[info[kFIRMessagingSenderID] longLongValue];
+
+  if (!info[kFIRMessagingMessageIDKey] ||
+      ![info[kFIRMessagingMessageIDKey] isKindOfClass:NSString.class]) {
+    FIRMessagingLoggerWarn(kFIRMessagingServiceExtensionInvalidMessageID,
+                           @"Delivery logging failed: Invalid Message ID");
+    return;
+  }
```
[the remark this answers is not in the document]
chliangGoogle: Done.

Issue #8448, opened 2021-07-28
Unaligned pointer warning building MessagingUnit target


paulb777:
When building the MessagingUnit target with Xcode 13 beta 4 in Swift Package Manager:

`ld: warning: pointer not aligned at address 0x179466 (_fm_MessagingClientEventExtension_fields + 6 from /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/firebase-ios-sdk-dzxqfpowrkksnfdbrdsjzildavlw/Build/Products/Debug-iphonesimulator/FirebaseMessaging.o)`

The warning should be fixed or at least tested on device to make sure it does not cause runtime crashes on ARM hardware.

maksymmalyhin:
Seems like the [proto code](https://github.com/firebase/firebase-ios-sdk/blob/3a5bdd881672f752b50d8111f11c7b617556687b/FirebaseMessaging/Sources/Protogen/nanopb/me.nanopb.c#L18) was generated for older nanopb version. Currently we install 2.30908.0 with SPM.

maksymmalyhin:
never mind, I tried to generate with 2.30908.0 version from https://github.com/firebase/nanopb but there were no code changes.

paulb777:
Fixed via missing defines in Package.swift in #8460

PR #8440, opened 2021-07-27
Update changelog

chliangGoogle:

PR #6181, opened 2020-07-31
 Adding a new API that allows iOS users to log notification delivery signal that exports to BigQuery.

chliangGoogle:
We got approved within our team to proceed with this launch. Please review this again, the API review is approved at 
go/fm-ios-delivery

PR #8458, opened 2021-07-29
Revert "Adding a new API that allows iOS users to log notification delivery signal that exports to BigQuery."

maksymmalyhin:
Reverts firebase/firebase-ios-sdk#8443, firebase/firebase-ios-sdk#8440, firebase/firebase-ios-sdk#6181

Reverting to give more time for investigation of #8448.


chliangGoogle (approved):

maksymmalyhin:
SPM checks fails where introduced in #8456 and are not relevant to the PR. `pod spec lint` fails because the podspec uses an existing version with existing tag pointing to incompatible code. 
I'm merging the PR.

PR #8459, opened 2021-07-29
Run spm workflow when changes in SwiftPMTests/*

ncooke3:
Also:
- add AppCheck to swift import tests
- add AppCheck to objc import tests
#no-changelog


paulb777 (commented):
LGTM, presumably CI will fail because of the pending Catalyst AppCheck issue.

paulb777 (approved):
Thanks!

PR #8465, opened 2021-07-30
Add DatabaseUnitSwift scheme and add it to GHA

paulb777:
Enable SPM testing of DatabaseUnitSwift other than as part of the full Firebase-package.

#no-changelog


ncooke3 (approved):
LGTM on green and re-enabling cron.

paulb777:
@jrcrespoh This PR is what I wasn't remembering when we last talked about Swift PM and managing test target schemes.

Issue #8357, opened 2021-07-08
RTDB getData Crashes When Offline


ChristianGoodridgeLVM:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.5.1
  * Firebase SDK version: 8.2.0
  * Installation method: `CocoaPods
  * Firebase Component:  Database

### [REQUIRED] Step 2: Describe the problem

Performing a getData call while not connected to the internet produces the following error:
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[FChildrenNode node]: unrecognized selector sent to instance

#### Steps to reproduce:
Make any getData call while not connected to the internet

paulb777:
Sorry about the trouble. Please share the full backtrace of the crash and if feasible, a reproducible example.

ChristianGoodridgeLVM:
I cannot share a full stack trace into our codebase but here it is from the point of entry into the Firebase framework

https://user-images.githubusercontent.com/81970493/124939094-69c4af80-dfe3-11eb-921e-93d499debdec.png)

ChristianGoodridgeLVM:
Additional information if it helps, the requested data does in fact already exist in the cache and can be retrieved with an observeSingleCall

morganchen12:
Can you share the exception message?

ChristianGoodridgeLVM:
`	0   CoreFoundation                      0x000000018b177270 37A3D601-9398-3DC2-9D00-1B162726A4C9 + 1159792
	1   libobjc.A.dylib                     0x000000019ef0c480 objc_exception_throw + 56
	2   CoreFoundation                      0x000000018b086278 37A3D601-9398-3DC2-9D00-1B162726A4C9 + 172664
	3   CoreFoundation                      0x000000018b1798b8 37A3D601-9398-3DC2-9D00-1B162726A4C9 + 1169592
	4   CoreFoundation                      0x000000018b17b85c _CF_forwarding_prep_0 + 92
	5   FirebaseDatabase                    0x0000000103acd0f8 -[FIRDataSnapshot value] + 80
	6   [Redacted]                   0x000000010082ea98 $s10[REDACTED]9RTDBRelayC7getData4with3for7closureySS_SaySSGySDySSyXlGctFys5Error_pSg_So15FIRDataSnapshotCtcfU_ + 232
[...]

PR #8470, opened 2021-07-30
Add diagnostic for type error

morganchen12:
Something to help diagnose #8357.

### DIFF
--- a/FirebaseDatabase/Sources/Core/View/FCacheNode.m
+++ b/FirebaseDatabase/Sources/Core/View/FCacheNode.m
@@ -30,6 +30,9 @@
 - (id)initWithIndexedNode:(FIndexedNode *)indexedNode
        isFullyInitialized:(BOOL)fullyInitialized
                isFiltered:(BOOL)filtered {
+    NSAssert([indexedNode isKindOfClass:[FIndexedNode class]],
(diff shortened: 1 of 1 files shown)


FirebaseDatabase/Sources/Core/View/FCacheNode.m
```diff
@@ -30,6 +30,9 @@ @implementation FCacheNode
 - (id)initWithIndexedNode:(FIndexedNode *)indexedNode
        isFullyInitialized:(BOOL)fullyInitialized
                isFiltered:(BOOL)filtered {
+    NSAssert([indexedNode isKindOfClass:[FIndexedNode class]],
+             @"Expected node with type FIndexNode, but instead got %@",
+             indexedNode);
```
paulb777, line 35: Will it always lead to a crash if it's not an `FIndexedNode`? Won't `NSAssert` be compiled out in Release mode?

morganchen12: Loosened this check to only check for the crashing selector and made it a thrown exception instead of an assert.

paulb777:
@morganchen12 Is this PR still necessary after #8484?

morganchen12:
It is not.

Issue #8473, opened 2021-08-01
FirebaseInstallations shouldn't depend on FirebaseCore private headers


omarzl:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Any
  * Firebase SDK version: 8.5.0
  * Installation method: CocoaPods
  * Firebase Component: Installations and Core

#### Steps to reproduce:

In [FirebaseInstallations.podspec](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInstallations.podspec) you are using FirebaseCore private headers which shouldn't be done that way, this causes issues with pure iOS build systems like Bazel since we get duplicated headers errors.

These issues doesn't happen with Xcode build system since there exists Project headers additionally to Public and Private but in other build systems this concept doesn't exist and we need to do workarounds to integrate Firebase.

For good practices each pod should be independent and don't use other pod private source code, instead if FirebaseInstallations needs them then FirebaseCore should expose those headers as public.

ncooke3:
Hi @omarzl, 

Thanks for bringing this to our attention and apologies for the trouble. The team will be looking into this and we will update the thread accordingly. Thanks!

maksymmalyhin:
@omarzl Currently we use private headers in Cocoapods in Firebase to expose Firebase SDK API to other Firebase SDKs when the API doesn't make sense for external users. The intention is keep public API surface cleaner. 

Could you please provide more details on the way you use Firebase SDKs? We need a usage example to make a reasonable progress with the issue.

Additionally, I would like to point out to the approach we use in Swift Package Manager which also doesn't have a concept of private headers. There we laverage our mono-repo layout and add the repo root to the header search paths, e.g. like [here](https://github.com/firebase/firebase-ios-sdk/blob/26b561aba60422f99e90f64ba4e9e6fcb9e349d7/Package.swift#L758). Using repo-relative imports in the code enables us doing this. This approach allows us to support both Cocoapods and Swift Package Manager with the minimal overhead. I wonder if this approach may be used in Bazel?

PR #8474, opened 2021-08-02
Remove private Core headers from Installations sources

ncooke3:


morganchen12:
AFAIK every Firebase component uses the internal Core headers for FIRComponent registry. Though the original issue referenced Installations, it's potentially an issue that affects all Firebase libraries for bazel users.

https://cs.opensource.google/search?q=%22FirebaseCore%2FSources%2FPrivate%2FFirebaseCoreInternal.h%22&start=1

The least-change way to fix this is probably to just make component registry a public API, but we'll have to wait for Paul and 
Ryan's input.

paulb777:
Yep, see @maksymmalyhin's comment on #8473 and the CI failure. 

We have several examples of relying on non-public internal headers in Firebase.

Issue #7127, opened 2020-12-08
Twitter sign in button won't build 


justBearded:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: _12.2____
  * Firebase SDK version: _7.2____
  * Installation method: `CocoaPods
  * Firebase Component: __Auth___ (Auth, Core, Database, Firestore, Messaging, Storage, etc)

### [REQUIRED] Step 2: Describe the problem
MacOS via SwiftUI:
following the documentation to setup the twitter sign in button Im getting an error 
" Value of type 'OAuthProvider' has no member 'getCredentialWith' " 

#### Steps to reproduce:
build a SwiftUI app for macOS 
add pods:
pod 'Firebase/Auth', '~> 7.2-M1'
  pod 'Firebase/Firestore', '~> 7.2-M1'
  #pod 'Firebase/Analytics', '~> 7.2-M1'
  pod 'Firebase/Storage', '~> 7.2-M1'
  pod 'Firebase/Functions', '~> 7.2-M1'
  pod 'Firebase/DynamicLinks', '~> 7.2-M1'
  pod 'Firebase/Messaging', '~> 7.2-M1'
pod 'FirebaseFirestoreSwift', '7.2-beta'

init firebase in the apps initializer 

follow twitter auth documentation 

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

If you have a downloadable sample project that reproduces the bug you're reporting, you will
likely receive a faster response on your issue.

#### Relevant Code:

```
// TODO(you): code here to reproduce the problem
```
var provider = OAuthProvider(providerID: "twitter.com")
    func signIN(){
    provider.getCredentialWith(nil) { credential, error in
          if error != nil {
            // Handle error.
          }
          if credential != nil {
            Auth().signIn(with: credential) { authResult, error in
              if error != nil {
                // Handle error.
              }
              // User is signed in.
              // IdP data available in authResult.additionalUserInfo.profile.
              // Twitter OAuth access token can also be retrieved by:
              // authResult.credential.accessToken
              // Twitter OAuth ID token can be retrieved by calling:
              // authResult.credential.idToken
              // Twitter OAuth secret can be retrieved by calling:
              // authResult.credential.secret
            }
          }
        }
    }

ryanwilson:
[...]

PR #8493, opened 2021-08-04
Remove conditional compilation in app check headers

morganchen12:


FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
```diff
@@ -20,9 +20,6 @@
 
 #pragma mark - DeviceCheck
 
-// Targets where DeviceCheck framework is available to be used in preprocessor conditions.
-#define FIR_DEVICE_CHECK_SUPPORTED_TARGETS TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_TV
```
maksymmalyhin, line 24: I'm not sure if this check can be removed entirely because, e.g. on watchOS DeviceCheck is not present so imports of it's headers won't work.

morganchen12: I've replaced this with checking directly if the DeviceCheck framework can be imported.

FirebaseAppCheck/Sources/DeviceCheckProvider/FIRDeviceCheckProvider.m
```diff
@@ -16,10 +16,12 @@
 
 #import "FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h"
 
-#if FIR_DEVICE_CHECK_SUPPORTED_TARGETS
-
 #import <Foundation/Foundation.h>
 
+#if __has_include(<DeviceCheck/DeviceCheck.h>)
+#import <DeviceCheck/DeviceCheck.h>
+#endif
```
maksymmalyhin, line 23: None of the code below makes any sense if `DeviceCheck` is unavailable. I wonder why you would prefer to still keep include it?

morganchen12: This is entirely so it shows up in Xcode if a user tries to invoke any of these APIs with a helpful message saying "this is unavailable on the platform you're targeting" instead of saying "No known method for selector".

maksymmalyhin: I probably don't have enough context here. Could you please give an example of what error we are trying to avoid and what will be the replacement. I would like to understand better if the unnecessary binary size increase is a fair trade of here.

morganchen12: Here's an example where a user was trying to find a method in Auth: https://github.com/firebase/firebase-ios-sdk/issues/7127#issuecomment-740850842

I'll see if I can expose the more helpful method without increasing binary size on platforms where the DeviceCheck classes are unused.

morganchen12: I wasn't able to find a way to keep the more helpful diagnostics without increasing the code size. If this module is built statically and linked without the `-ObjC` flag, the linker would be able to remove the two unused object files on watchOS, but that seems unlikely for most app targets. For a dynamically-built module the symbols would always be included in the final app bundle (from what I understand).

I'm fine with reverting these changes except the documentation comments, let me know what you think.

morganchen12:
Latest test failure appears to be Keychain-related.

FirebaseAppCheck/Sources/DeviceCheckProvider/FIRDeviceCheckProvider.m
maksymmalyhin: I don't feel too strongly about it, but excluding the code on completely unsupported platform (watchOS) looks a bit cleaner to me. In this specific case it should not create too much confusion because the underlying Apple API also is completely unavailable.

maksymmalyhin (approved):
LGTM.

FirebaseAppCheck/Sources/DeviceCheckProvider/FIRDeviceCheckProvider.m
morganchen12: Done.

PR #8494, opened 2021-08-04
Configure test keychain for macOS tests

maksymmalyhin:
Should fix some keychain failures from https://github.com/firebase/firebase-ios-sdk/issues/8471#issuecomment-892646717.
- script to create, unlock and set as default a new keychain
- the script added as a step to FCM and App Check GHA workflows

I cannot reproduce the failure reliably in the PR (sometimes it succeeds without the fix), so I cannot be 100% sure if the approach actually fixes the problem. Let's try it and revert if it doesn't work.


maksymmalyhin:
@chliangGoogle [This failure](https://github.com/firebase/firebase-ios-sdk/pull/8494/checks?check_run_id=3246443320) looks similar to what we have with nanopb.

maksymmalyhin:
> @chliangGoogle [This failure](https://github.com/firebase/firebase-ios-sdk/pull/8494/checks?check_run_id=3246443320) looks similar to what we have with nanopb.

Sorry, I re-ran the job and it overwrote the logs.

chliangGoogle (approved):

scripts/configure_test_keychain.sh
```diff
@@ -0,0 +1,43 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+set -o xtrace
+
+TEST_KEYCHAIN_NAME="temp.keychain"
+TEST_KEYCHIN_PASSWORD="temp.keychain.password"
```
paulb777, line 21: nit: KEYCHAIN

paulb777 (approved):

maksymmalyhin:
Seems like the fix works. I'll merge it and apply the fix to the remaining workflows in following PRs.

PR #8497, opened 2021-08-05
FirebaseInAppMessagingSwift: explicitly import UIKit

maksymmalyhin:
Fixes the [failure](https://github.com/firebase/firebase-ios-sdk/runs/3250028778?check_suite_focus=true) from https://github.com/firebase/firebase-ios-sdk/issues/8471#issuecomment-893449537.

### DIFF
--- a/FirebaseInAppMessaging.podspec
+++ b/FirebaseInAppMessaging.podspec
@@ -74,6 +74,8 @@ See more product details at https://firebase.google.com/products/in-app-messagin
     'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
   }
 
+  s.framework = 'UIKit'
+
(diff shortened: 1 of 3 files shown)


maksymmalyhin:
I also noticed that UIKit framework is not declared as a dependency in [FirebaseInAppMessaging.podspec](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInAppMessaging.podspec) and [FirebaseInAppMessagingSwift](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInAppMessagingSwift.podspec). Do you know a reason for that?

paulb777 (approved):
I don't think we're consistent about adding UIKit to the package dependencies since an import is sufficient to pull it in. However, it probably would be good to add them and be explicit.

PR #8499, opened 2021-08-05
Add functions spm unit and build tests

jrcrespoh:
Tests ran successfully with watchos

`pod lib lint` fails on travis for watchos, but succeeds locally. Disabling test for now, pending GHA update to Xcode 12.5+


Package.swift
```diff
@@ -690,6 +690,24 @@ let package = Package(
...
     ),
+    .testTarget(
+      name: "FunctionsUnit",
+      dependencies: ["FirebaseFunctions",
+                     "SharedTestUtilities"],
+      path: "Functions/Example/Tests/",
+      exclude: [
+        "en.lproj",
+        "Tests-Info.plist",
+      ],
```
jrcrespoh, line 701: This was added because otherwise it would complain about not having a `defaultLocalization` for spm.

ncooke3 (commented):
Nice! I think we should add `watchos` to the each matrix in `functions.yml` [Here's one such matrix.](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/functions.yml#L27)

paulb777:
Looks like `pod lib lint` does not work for watch OS. https://travis-ci.org/github/firebase/firebase-ios-sdk/jobs/774699071

Functions/Tests/Unit/Swift/FunctionsAPITests.swift
```diff
@@ -61,9 +61,9 @@ final class FunctionsAPITests {
     }
 
     #if swift(>=5.5)
-      if #available(iOS 15.0, macOS 12.0, macCatalyst 15.0, tvOS 15.0, *) {
+      if #available(iOS 15.0, macOS 12.0, macCatalyst 15.0, tvOS 15.0, watchOS 8.0, *) {
```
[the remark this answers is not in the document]
ncooke3: cc: @granluo This is a compile time test

Package.swift
ncooke3: @paulb777 Have you seen this issue before?

ncooke3 (approved):
This LGTM 👍 . Added one comment about SwiftPM excludes as I havent seen that issue before

Package.swift
paulb777: Good catch @ncooke3! I think those are files that are no longer used and should be deleted.  I'll investigate.

jrcrespoh:
What should be done about incremental code coverage reports? I imagine these should not surface on tests?

paulb777:
@granluo is working on the incremental coverage reporting separately.

@jrcrespoh This PR should be good to resolve and merge here.

PR #8507, opened 2021-08-07
Fix failing swift-test for Perf SPM

paulb777:


jeremyjiang-dev (approved):
Thank you so much, Paul!

Issue #8495, opened 2021-08-05
Firebase performance crashes with objc_msgSend in method updateTraceWithCurrentSession of class FPRNetworkTrace


AbhishekBB:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: _12.4____
  * Firebase SDK version: _8.4.0____
  * Installation method: `CocoaPods`
  * Firebase Component: _Performance____ 

### [REQUIRED] Step 2: Describe the problem

We are unable to reproduce this, We are seeing this crash in our production app in crashlytics.

<img width="1276" alt="Screenshot 2021-08-05 at 5 11 02 PM" src="https://user-images.githubusercontent.com/15157308/128344167-485e2cb2-ef17-4bb3-b2c8-b8c0951fb307.png">

What happened? How can we make the problem occur?
We are unable to reproduce this.

If you have a downloadable sample project that reproduces the bug you're reporting, you will
likely receive a faster response on your issue.

NO

paulb777:
This looks like one of the traces reported in #8321

AbhishekBB:
Can we downgrade to any firebase version to avoid this crash?

luan-sala:
+1

I have the same crash.

![image](https://user-images.githubusercontent.com/38911319/128401069-ac01a424-590b-4276-8d17-d2b8355337b3.png)

<details>
<summary>2 other threads</summary>

![image](https://user-images.githubusercontent.com/38911319/128401178-06fbeb68-9879-4ef4-bf35-e363fe08cba9.png)

![image](https://user-images.githubusercontent.com/38911319/128401208-da14dab9-be3b-4478-8995-ac2a91b2e44c.png)

</details>

jposuna:
Thanks for reporting this! We are having some trouble reproducing the issue, but we have a theory about what could be going wrong and will submit a fix as soon as possible.

jposuna:
We just submitted a PR which should make it to the next iOS SDK release (scheduled for August 17th). Please feel free to reopen if that version still has this error.

caffieneToCode:
We too are getting these crashes, is Firebase version 8.6.0 released?
<img width="1256" alt="Screenshot 2021-08-13 at 12 01 05 PM" src="https://user-images.githubusercontent.com/11851138/129314821-a42b1a77-0224-4e09-9090-e96510535792.png">

Issue #8321, opened 2021-06-30
A crash of an any application with  Firebase/Performance.


K-Be:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.3
  * Firebase SDK version: 8.2.0
  * Installation method: CocoaPods
  * Firebase Component: Performance

### [REQUIRED] Step 2: Describe the problem
An application crashes on launch.

#### Steps to reproduce:

1. Clone a code from https://github.com/firebase/quickstart-ios/tree/master/performance and install pods
2. Turn on diagnostic flags for launch scheme PerformanceExample named "Malloc Scribble", "Zombie Objects": 

<img width="644" alt="Screenshot 2021-06-30 at 10 03 51" src="https://user-images.githubusercontent.com/128474/123895720-7bc2a480-d98a-11eb-943d-7fb07ec03f6d.png">

3. Launch an application named PerformanceExample.

Application will crash with a Zombie access message: 
```
#0	0x00007fff20176876 in object_setClass ()
#1	0x00007fff2042fbd2 in -[NSObject(NSObject) __dealloc_zombie] ()
#2	0x00007fff207f34d6 in -[NSOperation dealloc] ()
#3	0x00007fff2018f834 in objc_object::sidetable_release(bool, bool) ()
#4	0x000000010cc362a1 in __destroy_helper_block_e8_32s40s ()
#5	0x00007fff2007f42f in _Block_release ()
#6	0x000000010ca468d3 in __destroy_helper_block_e8_32s ()
#7	0x00007fff2007f42f in _Block_release ()
#8	0x000000010ca467f1 in __destroy_helper_block_e8_32s40s ()
#9	0x00007fff2007f42f in _Block_release ()
#10	0x000000010ca488ff in __destroy_helper_block_e8_32s40s48s ()
#11	0x00007fff2007f42f in _Block_release ()
#12	0x000000010ce2d9c8 in _dispatch_client_callout ()
#13	0x000000010ce34440 in _dispatch_lane_serial_drain ()
#14	0x000000010ce34f67 in _dispatch_lane_invoke ()
#15	0x000000010ce40de2 in _dispatch_workloop_worker_thread ()
#16	0x00007fff60c8f45d in _pthread_wqthread ()
#17	0x00007fff60c8e42f in start_wqthread ()
Enqueued from com.google.GDTCCTUploader (Thread 5) Queue : com.google.GDTCCTUploader (serial)
[...]

PR #8512, opened 2021-08-09
Resolve a possible race condition

visumickey:
**Key objects/DS:**
FPRSessionDetails - Object that contains a session details (sessionID)
FPRSessionManager - Manages the creation and updating of sessions.
FIRTrace/FPRNetworkTrace - Objects who care about the session change and acquires the recently modified sessionId.

**Current implementation:**
Whenever there is a change in the sessionID, session manager throws out a notification. FIRTrace and FPRNetwork trace receives the notification and talks to sessionManager to get the recent sessionId. That sessionID gets added to the FIRTrace and FPRNetworkTrace objects asynchronously.

**Potential threat:**
The notification for session change which leads to session details fetch from FIRTrace and FPRNetworkTrace would be stake as the session manager could have initiated a new session. A potential race condition where the object reference is lost before being added to FIRTrace and FPRNetworkTrace.

**Potential fix:**
As a part of the notification, ship out the sessionDetails so that object is never lost in memory and session manager can continue to change the sessionID without disrupting the subscribers of the sessionID change.

Potential fix for #8495.


FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
```diff
@@ -104,8 +105,12 @@ - (void)updateSessionId:(NSNotification *)notification {
   FPRSessionDetails *sessionInfo = [[FPRSessionDetails alloc] initWithSessionId:sessionIdString
                                                                         options:sessionOptions];
   self.sessionDetails = sessionInfo;
+  NSMutableDictionary<NSString *, FPRSessionDetails *> *userInfo =
+      [[NSMutableDictionary alloc] init];
+  [userInfo setObject:sessionInfo forKey:kFPRSessionIdNotificationKey];
   [self.sessionNotificationCenter postNotificationName:kFPRSessionIdUpdatedNotification
-                                                object:self];
+                                                object:self
+                                              userInfo:[userInfo copy]];
```
jposuna, line 113: Do we need to copy this? Would this not have a similar race condition?

jposuna (approved):

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
visumickey: This is the userInfo object copy and this is a local variable. So no real possibility of a race condition here. The FPRSessionDetails object is already help by the dictionary, so that is not going out of memory to cause a crash.

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
```diff
@@ -137,14 +140,20 @@ - (NSString *)description {
...
+    FPRSessionDetails *sessionDetails = [userInfo valueForKey:kFPRSessionIdNotificationKey];
+    if (sessionDetails) {
+      [self updateTraceWithCurrentSession:sessionDetails];
+    }
+  }
+}
+
+- (void)updateTraceWithCurrentSession:(FPRSessionDetails *)sessionDetails {
+  if (sessionDetails != nil) {
+    dispatch_sync(self.sessionIdSerialQueue, ^{
```
maksymmalyhin, line 155: Does it have to be `dispatch_sync`? Using `dispatch_sync` usually increases probability to introduce a deadlock unintentionally so should be avoided when possible.

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
```diff
@@ -191,9 +200,9 @@ - (void)start {
     [self checkpointState:FPRNetworkTraceCheckpointStateInitiated];
 
     FPRSessionManager *sessionManager = [FPRSessionManager sharedInstance];
-    [self updateTraceWithCurrentSession];
+    [self updateTraceWithCurrentSession:[sessionManager sessionDetails]];
```
maksymmalyhin, line 203: Why is it safe to access `[sessionManager sessionDetails]` here? Should we copy it before passing to the method? 

The same questions to similar changes in other traces.

FirebasePerformance/Tests/Unit/FPRSessionManagerTest.m
```diff
@@ -91,4 +91,35 @@ - (void)testSessionIdUpdationThrowsNotification {
...
+      addObserverForName:kFPRSessionIdUpdatedNotification
+                  object:instance
+                   queue:[NSOperationQueue mainQueue]
+              usingBlock:^(NSNotification *note) {
+                NSDictionary<NSString *, FPRSessionDetails *> *userInfo = note.userInfo;
+                FPRSessionDetails *sessionDetails =
+                    [userInfo valueForKey:kFPRSessionIdNotificationKey];
+                if (sessionDetails != nil) {
+                  containsSessionDetails = YES;
+                  updatedSessionDetails = sessionDetails;
```
maksymmalyhin, line 112: optional: Currently the test relies on the implementation details so the block happen to be called synchronously with posting the notification. According to the [docs](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1411723-addobserverforname) we may pass `nil` to `queue` to make sure the block is called on the poster thread. Alternatively `XCTestExpectation` can be used to make sure the test works when the block is called asynchronously. Otherwise the test may unexpectedly fail on a different platform or OS version.

maksymmalyhin (commented):
Could you please provide a bit more information on why do you think the crash happens and how the changes should fix it? Is there a way to design a test that reproduces the crash with a reasonable probability?

visumickey:
> Could you please provide a bit more information on why do you think the crash happens and how the changes should fix it? Is there a way to design a test that reproduces the crash with a reasonable probability?

Added some details notes on the description. Please let me know if that works.

paulb777:
Thanks for adding the detail. Since #8321 is reproducible, can you verify that it is fixed? And update the description to remove "potential" at least for that issue.

maksymmalyhin:
> > Could you please provide a bit more information on why do you think the crash happens and how the changes should fix it? Is there a way to design a test that reproduces the crash with a reasonable probability?
> 
> Added some details notes on the description. Please let me know if that works.

Thank you for the explanation! One thing is not 100% to me - what is meant by "the object reference is lost before being added to FIRTrace and FPRNetworkTrace"? I guess memory management issues may a result of a concurrent read and write access to the property holding FPRSessionDetails. If it is true, then I'm curious if we are sure that concurrent access does not happen e.g. [here](https://github.com/firebase/firebase-ios-sdk/pull/8512#discussion_r685540676).

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
visumickey: Good point. Makes sense. Updated it.

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
visumickey: I thought quite a bit about this and preferred to change this to sync (had it as async earlier). Agree with the possibility of the deadlock situation. Since this queue is internally handled and it not used for any purpose other than queueing the addition of sessionIds, I felt it is safe to go with a synchronous operation for this.

FirebasePerformance/Tests/Unit/FPRSessionManagerTest.m
visumickey: Sure. That makes sense. We can definitely try Expectation, but expectation also comes with a timeout value.

I would prefer to keep that change (if we decide to do that) outside this PR since that would require multiple tests to be changed as well.

visumickey:
> Thanks for adding the detail. Since #8321 is reproducible, can you verify that it is fixed? And update the description to remove "potential" at least for that issue.

I don't think this is related to issue 8321, so I removed it from the list. I will work with Maksym on figuring out the rootcause of that issue outside this PR.

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
```diff
@@ -191,9 +200,9 @@ - (void)start {
     [self checkpointState:FPRNetworkTraceCheckpointStateInitiated];
 
     FPRSessionManager *sessionManager = [FPRSessionManager sharedInstance];
-    [self updateTraceWithCurrentSession];
+    [self updateTraceWithCurrentSession:[[sessionManager sessionDetails] copy]];
```
maksymmalyhin, line 203: Is it still possible that `sessionManager.sessionDetails` is concurrently modified at the time of reading so it may trigger a similar crash?

maksymmalyhin: I guess we can declare [FPRSessionManager.sessionDetails](https://github.com/firebase/firebase-ios-sdk/blob/b19fb5f2e7326e64d23feb1db2311468f1b4e06d/FirebasePerformance/Sources/AppActivity/FPRSessionManager%2BPrivate.h#L25) property as `atomic` to prevent concurrent access.

FirebasePerformance/Sources/Timer/FIRTrace.m
```diff
@@ -139,9 +142,9 @@ - (void)start {
     self.backgroundActivityTracker = [[FPRTraceBackgroundActivityTracker alloc] init];
     FPRSessionManager *sessionManager = [FPRSessionManager sharedInstance];
     if (!self.isStage) {
-      [self updateTraceWithSessionId];
+      [self updateTraceWithSessionId:[[sessionManager sessionDetails] copy]];
```
maksymmalyhin, line 145: nit: Dot syntax should be used to access a property:
```
[sessionManager.sessionDetails copy]
```

FirebasePerformance/Sources/Instrumentation/FPRNetworkTrace.m
visumickey: That is a good suggestion. Added protection by making it an atomic property.

FirebasePerformance/Sources/Timer/FIRTrace.m
visumickey: Done.

visumickey:
> LGTM 👍

Thanks a ton for a thorough review @maksymmalyhin ! Greatly appreciate that 👍

paulb777 (approved):

PR #8513, opened 2021-08-10
Fix race condition where checkin is not reset properly when app first installed

### LINKED ISSUE #8491: Messaging: `deleteToken` no longer deletes token correctly

    file a Github issue.
    with the `firebase` tag.
    google group.

  * Xcode version: 12.5
  * Firebase SDK version: 8.5.0
  * Firebase Component: Messaging (Auth, Core, Database, Firestore, Messaging, Storage, etc)

FlutterFire: Issue: https://github.com/FirebaseExtended/flutterfire/issues/6766
RNFB: Issue https://github.com/invertase/react-native-firebase/issues/5570 P/R https://github.com/invertase/react-native-firebase/pull/5571

#### Steps to reproduce:

Call `deleteTokenWithCompletion` and inspect the token from `tokenWithCompletion` and see that it has not changed and is still valid.

From debugging, I can see that when the installations id is refreshed via ` installations.deleteWithCompletion`, the token will always refresh when calling `messaging.deleteTokenWithCompletion`

This behaviour can be replicated both in FlutterFire and RNFB. My question is what is the correct solution/method to refresh an FCM token?

I've looked through the changelog and the documentation, and it's still not 100% clear to me on how the installations id should be managed in relation to the FCM token.

Any help or guidance on this would be greatly appreciated.

#### Relevant Code:
By changing the following code, additional requests to `deleteTokenWithCompletion` will refresh the token:
From:

```
[[FIRMessaging messaging] deleteTokenWithCompletion:^(NSError *_Nullable error) {
    if (error) {
      [RNFBSharedUtils rejectPromiseWithNSError:reject error:error];
    } else {
      resolve([NSNull null]);
    }
  }];
```

To:

```
[[FIRInstallations installations] deleteWithCompletion:^(NSError * _Nullable error) {
    if (error) {
      [RNFBSharedUtils rejectPromiseWithNSError:reject error:error];
    } else {
      resolve([NSNull null]);
    }
  }];
```

chliangGoogle:
This fixes #8491.

There's a race condition when checkin sometimes not reset properly during app first install, then delete token request contains an old chekcin that causes token is not deleted properly.

### DIFF
--- a/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
+++ b/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
@@ -51,8 +51,7 @@
   self = [super init];
   if (self) {
     _tokenStore = [[FIRMessagingTokenStore alloc] init];
-    _checkinStore = [[FIRMessagingCheckinStore alloc] init];
-    _authService = [[FIRMessagingAuthService alloc] initWithCheckinStore:_checkinStore];
+    _authService = [[FIRMessagingAuthService alloc] init];
     [self resetCredentialsIfNeeded];
     [self configureTokenOperations];
     _installations = [FIRInstallations installations];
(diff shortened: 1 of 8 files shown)


FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -458,41 +457,34 @@ - (void)deleteWithHandler:(void (^)(NSError *))handler {
...
-      }
-      if (oldCheckinPreferences.deviceID.length && oldCheckinPreferences.secretToken.length) {
-        FIRMessagingLoggerDebug(kFIRMessagingMessageCodeStore006,
-                                @"App reset detected. Will delete server registrations.");
-        // We don't really need to delete old FCM tokens created via IID auth tokens since
-        // those tokens are already hashed by APNS token as the has so creating a new
-        // token should automatically delete the old-token.
-        [self didDeleteFCMScopedTokensForCheckin:oldCheckinPreferences];
-      } else {
-        FIRMessagingLoggerDebug(kFIRMessagingMessageCodeStore009,
```
maksymmalyhin, line 490: Is this message still relevant?

maksymmalyhin (commented):
Could you please provide a bit more info on how the race condition occurs  and causes the issue as well as how the changes fix it?

Is it possible to add a tests that triggers the race condition with the old code and succeeds with the changes?

maksymmalyhin (approved):
Thank you for the explanation. LGTM on CI green.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
ncooke3: FYI since the message is still relevant: The conditional block where it's logged is missing in the latest commit.

chliangGoogle: You are right, it's still relevant. I'm adding it back.

spundun:
@chliangGoogle Does this mean that the race condition is between these two threads of execution?
[pre-condition] app just got reinstalled and reloaded
[thread 1] detect reinstall -> reset checkin credentials in storage through checkinStore class
[thread 2] authService class is loaded -> reads and caches checkin credentials

chliangGoogle:
@spundun That is correct.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -458,41 +457,40 @@ - (void)deleteWithHandler:(void (^)(NSError *))handler {
...
   if (checkinPlistExists) {
     return;
   }
-
-  // Resets checkin in keychain if a fresh install.
   // Keychain can still exist even if app is uninstalled.
-  FIRMessagingCheckinPreferences *oldCheckinPreferences = [_checkinStore cachedCheckinPreferences];
-
-  if (oldCheckinPreferences) {
-    [_checkinStore removeCheckinPreferencesWithHandler:^(NSError *error) {
```
spundun, line 472: Note for reviewers: The _authService wraps _checkinService and the replacement call `[_authService resetCheckinWithHandler]` internally calls this method `[_checkinStore removeCheckinPreferencesWithHandler]` internally.

PR #8516, opened 2021-08-10
Fix Xcode 13 beta 4 build issues for Firebase Performance

jeremyjiang-dev:
This is a follow up PR for https://github.com/firebase/firebase-ios-sdk/pull/8412.

Do not submit before https://github.com/firebase/firebase-ios-sdk/pull/8412 is merged.


paulb777:
Please either merge #8412 or change the target branch of this PR to make its changes reviewable.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
```diff
@@ -38,6 +38,7 @@ typedef NS_ENUM(NSInteger, FPRApplicationState) {
 /** This class is used to track the app activity and create internal traces to capture the
  *  performance metrics.
  */
+NS_EXTENSION_UNAVAILABLE_IOS("Firebase Performance is not supported for iOS extensions.")
```
paulb777, line 41: Should be `NS_EXTENSION_UNAVAILABLE("Firebase Performance is not supported for extensions.")`

I suspect the current version will fail to build for tvOS on Xcode 13.

jeremyjiang-dev:
> Please either merge #8412 or change the target branch of this PR to make its changes reviewable.

Done. (I clicked the reviewers button too fast before realizing there are 100+ files need to be reviewed

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
jeremyjiang-dev: I tested for tvOS and the build succeeded, but I agree that `NS_EXTENSION_UNAVAILABLE` should be the right one.

paulb777: I might be confusing macOS and tvOS issues I saw before. Anyway, agreed `NS_EXTENSION_UNAVAILABLE` is better and will make adding macOS easier as well as if tvOS adds extensions

paulb777 (approved):

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
visumickey: I honestly did not like this MACRO being spread all over the code. We need to find a better elegant solution to this.

Can we huddle for a better solution here? I don't want to block this PR. But this can't scale.

visumickey (approved):

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
maksymmalyhin: I could not come up with a dramatically different alternative. As slight maintainability improvement you may declare a an alias macro like e.g. [here](https://github.com/firebase/firebase-ios-sdk/blob/ec193d015078cfdf084aa9714da6e8838ac2058b/FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h#L41) to simplify potential changes in the future (you will have to change availability only at one place instead of updating all the affected files.

PR #8529, opened 2021-08-14
First pass at Swift cancellation

paulb777:
Investigate feasibility of Swift cancellation implementation as a wrapper over existing APIs.

We're going to put this PR on hold pending more customer input and additional Swift Concurrency evolution. In the meantime, cancellation continues to be available via the original API.

Thanks to @maksymmalyhin 's last commit,  inspired by https://github.com/apple/swift-evolution/blob/main/proposals/0300-continuation.md#additional-examples, we believe we're on track to a solution. The race condition in the comments could be solved by exposing the FIRStorageUploadTask init API https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/FIRStorageReference.m#L227, call it before to get the task, and then pass the task into a modified putFile API. The two new public APIs could be prefixed with double underscores to indicate they're not for public consumption.

Beyond that, it is still an open question about how to implement cancellation for the auto-generated APIs from Objective C.  I don't see anything about overriding or modifying the generated functions at https://github.com/apple/swift-evolution/blob/main/proposals/0297-concurrency-objc.md.

### DIFF
--- a/FirebaseStorageSwift/Sources/AsyncAwait.swift
+++ b/FirebaseStorageSwift/Sources/AsyncAwait.swift
@@ -86,7 +86,7 @@ public enum StorageError: Error {
         }
       }, onCancel: { [storageTask] in
         // A possible race condition here. `onCancel` and `operation` are performed concurrently,
-        // so it's possible for `storageTask` still be `nil` but `Task.isCancelled` still be `false`, 
+        // so it's possible for `storageTask` still be `nil` but `Task.isCancelled` still be `false`,
         // so `storageTask` can be initialized and started after `onCancel` block has been
         // performed.
         storageTask?.cancel()
--- a/FirebaseStorageSwift/Sources/AsyncAwait.swift
+++ b/FirebaseStorageSwift/Sources/AsyncAwait.swift
@@ -14,6 +14,10 @@
 
 import FirebaseStorage
 
+public enum StorageError: Error {
+  case cancelled
+}
+
 #if swift(>=5.5)
   @available(iOS 15, tvOS 15, macOS 12, *)
   public extension StorageReference {
@@ -66,12 +70,33 @@ import FirebaseStorage
     func putFileAsync(from url: URL,
                       metadata: StorageMetadata? = nil) async throws -> StorageMetadata {
       typealias MetadataContinuation = CheckedContinuation<StorageMetadata, Error>
-      return try await withCheckedThrowingContinuation { (continuation: MetadataContinuation) in
-        // TODO: Use task to handle progress and cancellation.
-        _ = self.putFile(from: url, metadata: metadata) { result in
-          continuation.resume(with: result)
+
+      let metadata =
+        try await withCheckedThrowingContinuation { (continuation: MetadataContinuation) in
+
+          if Task.isCancelled {
+            print("xxxxxxxxx")
+            continuation.resume(throwing: StorageError.cancelled)
+            return
+          }
+          let storageTask = self.putFile(from: url, metadata: metadata) { result in
+            print("yyyyy")
+            continuation.resume(with: result)
+          }
+
+          storageTask.observe(StorageTaskStatus.progress) { snapshot in
+            guard let progress = snapshot.progress else {
+              fatalError("Missing progress")
+            }
+            print("\(progress.completedUnitCount) of \(progress.totalUnitCount)")
+            if Task.isCancelled {
+              print("zzzzzzzzzzzzz")
+              storageTask.cancel()
+              continuation.resume(throwing: StorageError.cancelled)
+            }
+          }
         }
-      }
+      return metadata
     }
 
     /// Asynchronously downloads the object at the current path to a specified system filepath.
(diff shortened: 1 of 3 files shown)


FirebaseStorageSwift/Sources/AsyncAwait.swift
```diff
@@ -66,12 +70,33 @@ import FirebaseStorage
...
-        // TODO: Use task to handle progress and cancellation.
-        _ = self.putFile(from: url, metadata: metadata) { result in
-          continuation.resume(with: result)
+
+      let metadata =
+        try await withCheckedThrowingContinuation { (continuation: MetadataContinuation) in
+
+          if Task.isCancelled {
+            print("xxxxxxxxx")
+            continuation.resume(throwing: StorageError.cancelled)
```
paulb777, line 79: This makes the test pass, but only works if the cancellation happens early.

FirebaseStorageSwift/Sources/AsyncAwait.swift
```diff
@@ -66,12 +70,33 @@ import FirebaseStorage
...
+          }
+
+          storageTask.observe(StorageTaskStatus.progress) { snapshot in
+            guard let progress = snapshot.progress else {
+              fatalError("Missing progress")
+            }
+            print("\(progress.completedUnitCount) of \(progress.totalUnitCount)")
+            if Task.isCancelled {
+              print("zzzzzzzzzzzzz")
+              storageTask.cancel()
```
paulb777, line 94: I can't figure out a test to get into this block.  

Similar issue as https://forums.swift.org/t/how-can-i-get-the-current-task/50677

FirebaseStorageSwift/Tests/Integration/StorageAsyncAwait.swift
```diff
@@ -155,17 +155,53 @@ import XCTest
...
+      let fileName = "hello&+@_ .txt"
+      let ref = storage.reference(withPath: "ios/public/" + fileName)
+      let fileURL = try setupFile(fileName)
+
+      let task = Task { () -> StorageMetadata in
+        let metadata = try await ref.putFileAsync(from: fileURL)
+        return metadata
+      }
+
+      let result = try await task.value
```
paulb777, line 184: Is this the best way for a client to call the `putFileAsync` API so that they can have a handle to cancel it?

FirebaseStorageSwift/Tests/Integration/StorageIntegration.swift
```diff
@@ -193,6 +193,26 @@ class StorageResultTests: StorageIntegrationCommon {
     waitForExpectations()
   }
 
+  func testSimplePutFileCancel() throws {
```
paulb777, line 196: Add a test for cancellation with the old APIs

FirebaseStorageSwift/Sources/AsyncAwait.swift
```diff
@@ -66,12 +70,33 @@ import FirebaseStorage
     func putFileAsync(from url: URL,
                       metadata: StorageMetadata? = nil) async throws -> StorageMetadata {
       typealias MetadataContinuation = CheckedContinuation<StorageMetadata, Error>
-      return try await withCheckedThrowingContinuation { (continuation: MetadataContinuation) in
-        // TODO: Use task to handle progress and cancellation.
-        _ = self.putFile(from: url, metadata: metadata) { result in
-          continuation.resume(with: result)
+
+      let metadata =
+        try await withCheckedThrowingContinuation { (continuation: MetadataContinuation) in
```
paulb777, line 75: There is `withTaskCancellationHandler` but I'm not sure how to combine that with `withCheckedThrowingContinuation` and not sure it would work with cancelling the Objective C task anyway.

https://forums.swift.org/t/how-to-cancel-a-publisher-when-using-withtaskcancellationhandler/49688/5

FirebaseStorageSwift/Sources/AsyncAwait.swift
maksymmalyhin: I guess you may wrap the storage task into a parent task (I did not actually tested it myself):
```
let taskHandle = async {
  return try await storage.putFileAsync(...) // The task is created and started.
} 

taskHandle.cancel() // The storage task is being cancelled because the parent task has been cancelled. 
```
See "Unstructured Concurrency" [here](https://docs.swift.org/swift-book/LanguageGuide/Concurrency.html) and [cancel()](https://developer.apple.com/documentation/swift/task/3851218-cancel) method docs.

Another approach could be using a [TaskGroup](https://developer.apple.com/documentation/swift/taskgroup) with has [cancelAll](https://developer.apple.com/documentation/swift/taskgroup/3814852-cancelall) method.

paulb777: Thanks, but I'm not sure I follow.  The first suggestion is how the invocation of the `putFileAsync` API is done in the test below.  Here, I'm trying to figure out the implementation.

The challenge I'm having is finding a way to cancel the Objective C task that is running from GTMSessionFetcher. And I can't figure out a way to get a handle to the Swift task to be able to check if it has been cancelled to know that the Objective C task should be cancelled.

paulb777:
Succeeded by #10161

PR #8543, opened 2021-08-17
GHA on macOS 11

paulb777:
Update GHA to use macOS 11 and Xcode 12.5

All jobs are updated from `macos-latest` to `macos-11` except the zip build jobs which still need to use Xcode 12.2

The messaging CI failure is unrelated and the same as the nightly issue being tracked in #8525

This PR should unblock enabling watchOS testing on CI, since XCTest support for watchOS was added in Xcode 12.5

#no-changelog


paulb777:
The release CI failure in the storage qs is a flake. See the previous success at https://github.com/firebase/firebase-ios-sdk/actions/runs/1140914577

ncooke3 (approved):
Thanks Paul!

PR #8550, opened 2021-08-19
Update binary size tool to connect to the Metrics Service.

granluo:


granluo:
A report will not be posted in PRs. Binary reports will run for all SDKs in postsubmits and the report will be sent to the Metrics Service.

paulb777 (approved):

.github/workflows/health_metrics.yml
```diff
@@ -44,13 +44,25 @@ jobs:
 
   binary_size_metrics:
     needs: check
-    if: always() && github.repository == 'Firebase/firebase-ios-sdk' && (github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.pull_request.merged)
+    if: always() && github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.merged
```
maksymmalyhin, line 47: Could you please add a comment explaining what the check does. If it read it correctly it limits the job to the post-pr of the PR target repo. Does it mean it will be run for forks like FirebasePrivate as well? If yes, will it be a problem?

scripts/health_metrics/create_binary_size_report.sh
```diff
@@ -19,43 +19,49 @@
 # This will generate a JSON output of changed files and their newly added
 # lines.
 
+set -ex
+
 BINARY_SIZE_SDK=()
-if $FirebaseABTesting == 'true'; then
+if [[ "${POSTSUBMIT}" == true || "${FirebaseABTesting}" == 'true' ]]; then
```
maksymmalyhin, line 25: A comment will be useful for future us.

scripts/health_metrics/create_binary_size_report.sh
```diff
@@ -19,43 +19,49 @@
 # This will generate a JSON output of changed files and their newly added
 # lines.
```
maksymmalyhin, line 21: It's a bit difficult to understand how to use the script. Could you please either:
- replace environment variable with the script input parameters and provide docs on the usage explaining the parameters
- provide docs on the environment variables that are used as input parameters

maksymmalyhin (approved):
LGTM 👍  but a couple questions and requests for docs.

PR #8555, opened 2021-08-19
Version bump Analytics specs with release tooling

ncooke3:
Up until now, the release tooling only bumped the podspecs of open source libraries.
Since the podspecs for the close source libraries (`FirebaseAnalytics` and `GoogleAppMeasurement`) are in `JSON`, the  tooling had to be slightly adjusted to bump the versions in that file type.

#no-changelog


paulb777 (commented):
LGTM. Thanks!

I'll also double check with M103.

paulb777 (approved):

PR #8557, opened 2021-08-20
Remove pod_lib_lint script workaround

paulb777:
With CocoaPods 1.10.2, it's no longer necessary to remove the Twitter URL in the `pod_lib_lint` script. 

See https://github.com/CocoaPods/CocoaPods/pull/10398


maksymmalyhin (approved):
I'm glad we can remove it 👍

PR #8558, opened 2021-08-20
Functions: don't send dummy FAC token

maksymmalyhin:
Fixes https://github.com/FirebaseExtended/flutterfire/issues/6794 for Apple platforms.

b/197236400

### DIFF
--- a/Functions/CHANGELOG.md
+++ b/Functions/CHANGELOG.md
@@ -1,7 +1,6 @@
-# v8.7.1
-- [changed] Don't set the App Check header in the case of App Check error. (#8558)
 # v8.7.0
 - [fixed] Add watchOS support (#8499).
+- [changed] Don't set the App Check header in the case of App Check error (#8558). 
 
 # v8.3.0
 - [fixed] Fixed an issue where subclassing Functions was unusable (#8265).
--- a/Functions/CHANGELOG.md
+++ b/Functions/CHANGELOG.md
@@ -1,3 +1,5 @@
+# v8.7.1
+- [changed] Don't set the App Check header in the case of App Check error. (#8558)
 # v8.7.0
 - [fixed] Add watchOS support (#8499).
 
--- a/Functions/CHANGELOG.md
+++ b/Functions/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v8.7.0
 - [fixed] Add watchOS support (#8499).
-- [changed] Don't set the App Check header in the case of App Check error (#8558). 
+- [changed] Don't set the App Check header in the case of App Check error (#8558).
 
 # v8.3.0
 - [fixed] Fixed an issue where subclassing Functions was unusable (#8265).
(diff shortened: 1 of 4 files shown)


Functions/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# v8.7.1
```
paulb777, line 1: Should either be part of 8.7.0 or 8.6.1 if you want to cherry-pick to a `release-8.6` branch created from the 8.6.0 tag to patch the current release.

Functions/Example/Tests/FIRFunctionsTests.m
```diff
@@ -189,6 +189,43 @@ - (void)testCallFunctionWhenAppCheckIsInstalled {
...
+
+  NSError *networkError = [NSError errorWithDomain:self.name code:-2 userInfo:nil];
+
+  XCTestExpectation *httpRequestExpectation =
+      [self expectationWithDescription:@"HTTPRequestExpectation"];
+  __weak __auto_type weakSelf = self;
+  _fetcherService.testBlock = ^(GTMSessionFetcher *_Nonnull fetcherToTest,
+                                GTMSessionFetcherTestResponse _Nonnull testResponse) {
+    // Fixes retain cycle warning for Xcode 11 and earlier.
+    // __unused to avoid warning in Xcode 12+.
```
paulb777, line 205: Xcode 11 is no longer supported

Functions/FirebaseFunctions/FUNContext.m
```diff
@@ -101,7 +101,10 @@ - (void)getContext:(void (^)(FUNContext *context, NSError *_Nullable error))comp
 
     [_appCheck getTokenForcingRefresh:NO
                            completion:^(id<FIRAppCheckTokenResultInterop> _Nonnull tokenResult) {
-                             appCheckToken = tokenResult.token;
+                             // Send only valid token to functions.
+                             if (tokenResult.error == nil) {
+                               appCheckToken = tokenResult.token;
```
paulb777, line 106: Why does the `getTokenForcingRefresh` return a token when there are errors?

A Swift Result type could help this API.  😛

maksymmalyhin: The current App Check semantic doesn't exactly fit Swift Result type. For most of the integrator SDKs setting a value to FAC header even in the case of the failure is a way to indicate that App Check is integrated with the app (see more details in go/app-attestation-sdk-api#heading=h.605kjfzcbxum). To simplify integration for these products it was decided to always return a token even if an error happened (it is a dummy token in the case of the error). 

For Functions the dummy token is useless and it appeared to trigger another issue because of the way Firebase Functions wrapper handle invalid tokens. So we are fixing it by:
1. Not sending the dummy tokens for Functions on the SDK side
2. Fixing handling of invalid FAC tokens on the callable functions side

Functions/Example/Tests/FIRFunctionsTests.m
maksymmalyhin: Good catch, let me clean it up!

paulb777 (approved):
Fix style and LGTM on green CI

weixifan:
Adding this for reference: See FirebaseExtended/flutterfire#6794 for more details.

PR #8560, opened 2021-08-20
Handle the case when refreshToken is null

yoyomyo:
Handle the case when refreshToken is null for Custom Auth. 
If refresh token is absent, we won't send a getAccountInfoRequest to populate FIRUser fields. Instead, a FIRUser is created from a Firebase ID token, with only the user_id field populated from the ID token.
#no-changelog

### DIFF
--- a/FirebaseAuth/Sources/User/FIRUser.m
+++ b/FirebaseAuth/Sources/User/FIRUser.m
@@ -282,7 +282,11 @@ static void callInMainThreadWithAuthDataResultAndError(
   user.auth = auth;
   user.tenantID = auth.tenantID;
   user.requestConfiguration = auth.requestConfiguration;
-  if (refreshToken.length) {
+  if (!refreshToken.length) {
+    [user updateWithIDToken:accessToken];
+    callback(user, nil);
+    return;
+  }
     [user
         internalGetTokenWithCallback:^(NSString *_Nullable accessToken, NSError *_Nullable error) {
           if (error) {
@@ -306,10 +310,6 @@ static void callInMainThreadWithAuthDataResultAndError(
                                   callback(user, nil);
                                 }];
         }];
-  } else {
-    [user updateWithIDToken:accessToken];
-    callback(user, nil);
-  }
 }
 
 - (instancetype)initWithTokenService:(FIRSecureTokenService *)tokenService {
--- a/FirebaseAuth/Sources/User/FIRUser.m
+++ b/FirebaseAuth/Sources/User/FIRUser.m
@@ -282,29 +282,34 @@ static void callInMainThreadWithAuthDataResultAndError(
   user.auth = auth;
   user.tenantID = auth.tenantID;
   user.requestConfiguration = auth.requestConfiguration;
-  [user internalGetTokenWithCallback:^(NSString *_Nullable accessToken, NSError *_Nullable error) {
-    if (error) {
-      callback(nil, error);
-      return;
+    if (refreshToken.length) {
+      [user internalGetTokenWithCallback:^(NSString *_Nullable accessToken, NSError *_Nullable error) {
+        if (error) {
+          callback(nil, error);
+          return;
+        }
+        FIRGetAccountInfoRequest *getAccountInfoRequest =
+            [[FIRGetAccountInfoRequest alloc] initWithAccessToken:accessToken
+                                             requestConfiguration:auth.requestConfiguration];
+        [FIRAuthBackend
+            getAccountInfo:getAccountInfoRequest
+                  callback:^(FIRGetAccountInfoResponse *_Nullable response, NSError *_Nullable error) {
+                    if (error) {
+                      // No need to sign out user here for errors because the user hasn't been signed in
+                      // yet.
+                      callback(nil, error);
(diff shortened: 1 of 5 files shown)


yoyomyo:
quickstart test is failing on master branch as well, starting review early while I find someone to look into the test failure

yoyomyo:
Thank you, Paul!

FirebaseAuth/Sources/User/FIRUser.m
```diff
@@ -282,28 +282,34 @@ + (void)retrieveUserWithAuth:(FIRAuth *)auth
...
-                  // yet.
-                  callback(nil, error);
-                  return;
-                }
-                user.anonymous = anonymous;
-                [user updateWithGetAccountInfoResponse:response];
-                callback(user, nil);
-              }];
-  }];
+  if (refreshToken.length) {
```
rosalyntan, line 285: nit: Maybe to avoid indenting this entire section, do something like:

```
if (!refreshToken.length) {
  [user updateWithIdToken:accessToken];
  callback(user, nil);
  return;
}

[user internalGetTokenWithCallback:^( ... ) {
  ...

}];

```

FirebaseAuth/Tests/Unit/FIRAuthTests.m
```diff
@@ -1513,6 +1521,39 @@ - (void)testSignInWithCustomTokenSuccess {
...
+ */
+- (void)testSignInWithCustomTokenWithoutRefreshTokenSuccess {
+  OCMExpect([_mockBackend verifyCustomToken:[OCMArg any] callback:[OCMArg any]])
+      .andCallBlock2(^(FIRVerifyCustomTokenRequest *_Nullable request,
+                       FIRVerifyCustomTokenResponseCallback callback) {
+        XCTAssertEqualObjects(request.APIKey, kAPIKey);
+        XCTAssertEqualObjects(request.token, kCustomToken);
+        XCTAssertTrue(request.returnSecureToken);
+        dispatch_async(FIRAuthGlobalWorkQueue(), ^() {
+          id mockVeriyCustomTokenResponse = OCMClassMock([FIRVerifyCustomTokenResponse class]);
```
rosalyntan, line 1536: nit: `mockVerifyCustomTokenResponse`

FirebaseAuth/Tests/Unit/FIRAuthTests.m
```diff
@@ -1513,6 +1521,39 @@ - (void)testSignInWithCustomTokenSuccess {
...
+          [self stubTokensWithMockResponseNoRefreshToken:mockVeriyCustomTokenResponse];
+          callback(mockVeriyCustomTokenResponse, nil);
+        });
+      });
+  // If refreshToken is not present, GetAccountInfo is not invoked.
```
rosalyntan, line 1541: Can you add a verification statement that verifies `GetAccountInfo` is not invoked?

FirebaseAuth/Tests/Unit/FIRAuthTests.m
```diff
@@ -1513,6 +1521,39 @@ - (void)testSignInWithCustomTokenSuccess {
   OCMVerifyAll(_mockBackend);
 }
 
+/** @fn testSignInWithCustomTokenWithoutRefreshTokenSuccess
+    @brief Tests the flow of a successful @c signInWithCustomToken:completion: call without a
+   refresh token.
+ */
+- (void)testSignInWithCustomTokenWithoutRefreshTokenSuccess {
```
ncooke3, line 1528: Could you add a test to assert the intended error occurs when it is _not_ successful? I believe it should be the error with the `kRefreshTokenError` message.

ncooke3 (commented):
Added a comment for testing a new code path. Otherwise, LGTM.

FirebaseAuth/Tests/Unit/FIRAuthTests.m
yoyomyo: Added `OCMReject([_mockBackend getAccountInfo:[OCMArg any] callback:[OCMArg any]]);`

FirebaseAuth/Tests/Unit/FIRAuthTests.m
yoyomyo: Hi Nick,
Thank you for suggesting to add more tests. Added `testGetIDTokenResultForcingRefreshFailureNoRefreshToken` in `FIRUsersTests.m` to test the scenario when a user tries to refresh without a refresh token. However I don't know the syntax to verify NSError with message `kRefreshTokenError` yet: I am still pretty new to iOS development. Will keep digging. If you know an existing example, please do point me to it.

yoyomyo: Added syntax to verify the correct NSError is returned from SecureTokenService. Mark this as resolved.

paulb777:
@ncooke3 Please take another look. Thanks.

rosalyntan (approved):

ncooke3 (approved):
LGTM. Thank you for the very well documented tests!  👍

paulb777 (approved):

PR #8593, opened 2021-08-30
Create a random delay before initiating a remote config fetch.

visumickey:
**Rationale:**
This is required to ensure that a remote notification does not trigger multiple config fetches from remote config at the same time.

### DIFF
--- a/FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags+Private.h
+++ b/FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags+Private.h
@@ -24,7 +24,7 @@ static NSString *const kFPRConfigPrefix = @"com.fireperf";
 static NSInteger const kFPRConfigFetchIntervalInSeconds = 12 * 60 * 60;
 
 /** Interval after which the configurations can be fetched. Specified in seconds. */
-static NSInteger const kFPRConfigAppStartDelayInSeconds = 1 * 30;
+static NSInteger const kFPRMinAppStartConfigFetchDelayInSeconds = 1 * 30;
 
 /** This extension should only be used for testing. */
 @interface FPRRemoteConfigFlags ()
@@ -37,6 +37,9 @@ static NSInteger const kFPRConfigAppStartDelayInSeconds = 1 * 30;
 
 /** @brief User defaults used for caching. */
 @property(nonatomic) NSUserDefaults *userDefaults;
+
+/** @brief Number of seconds delayed until the first config is made during app start. */
+@property(nonatomic) NSTimeInterval appStartConfigFetchDelayInSeconds;
 
 /**
  * Creates an instance of FPRRemoteConfigFlags.
--- a/FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
+++ b/FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
@@ -23,6 +23,8 @@
 
 #define ONE_DAY_SECONDS 24 * 60 * 60
 
+static NSDate *kFPRAppStartTime = nil;
+
 typedef NS_ENUM(NSInteger, FPRConfigValueType) {
   // Config value type String.
   FPRConfigValueTypeString,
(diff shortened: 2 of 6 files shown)


FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
```diff
@@ -69,6 +75,8 @@ - (instancetype)initWithRemoteConfig:(FIRRemoteConfig *)config {
     _fprRemoteConfig = config;
     _userDefaults = [FPRConfigurations sharedInstance].userDefaults;
     self.fetchInProgress = NO;
+    self.appStartConfigFetchDelayInSeconds =
+        kFPRMinAppStartConfigFetchDelayInSeconds + arc4random_uniform(30);
```
paulb777, line 79: How is it know that 30 seconds is enough of a spread to address the issue?

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags+Private.h
```diff
@@ -24,7 +24,7 @@ static NSString *const kFPRConfigPrefix = @"com.fireperf";
 static NSInteger const kFPRConfigFetchIntervalInSeconds = 12 * 60 * 60;
 
 /** Interval after which the configurations can be fetched. Specified in seconds. */
-static NSInteger const kFPRConfigAppStartDelayInSeconds = 1 * 30;
+static NSInteger const kFPRMinAppStartConfigFetchDelayInSeconds = 1 * 30;
```
jeremyjiang-dev, line 27: Nit: curious why is it "1 * 30" instead of "30"

paulb777 (commented):
Please fix CI

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
```diff
@@ -23,6 +23,8 @@
 
 #define ONE_DAY_SECONDS 24 * 60 * 60
 
+static NSDate *kFPRAppStartTime = nil;
```
jeremyjiang-dev, line 26: Nit: why do we need `k` as the prefix?

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
```diff
@@ -110,8 +118,10 @@ - (void)update {
 
   NSTimeInterval timeIntervalSinceLastFetch =
       [self.fprRemoteConfig.lastFetchTime timeIntervalSinceNow];
-  if (!self.fprRemoteConfig.lastFetchTime ||
-      ABS(timeIntervalSinceLastFetch) > kFPRConfigFetchIntervalInSeconds) {
+  NSTimeInterval timeSinceAppStart = [kFPRAppStartTime timeIntervalSinceNow];
+  if ((ABS(timeSinceAppStart) > self.appStartConfigFetchDelayInSeconds) &&
```
jeremyjiang-dev, line 122: When is `update()` gets called? Will it be possible that `update()` always gets called within 30 seconds, so that the remote config will never be fetched?

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
visumickey: We are taking cue-s from the RC team to see what works well for them.

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
visumickey: update() is called whenever we bring the app to foreground/app launch.

So, this 30 seconds is applied only for a cold start of the application. For the all subsequent launches, this delay does not kick in.

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
visumickey: Done.

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags+Private.h
visumickey: done.

FirebasePerformance/Sources/Configurations/FPRRemoteConfigFlags.m
jeremyjiang-dev: Makes sense! Thanks!

jeremyjiang-dev (approved):
LGTM on CI green

PR #8606, opened 2021-08-31
Fix GoogleAppMeasurement SPM CI

paulb777:
I added a script in #8426 for GoogleAppMeasurement SPM CI but invoked it too late. This should fix it.

Also, do the version bump for 8.7.0 for GoogleAppMeasurement


Package.swift
```diff
@@ -149,7 +149,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Please keep the version specification aligned with
       // scripts/setup_spm_test_app_measurement.sh.
-      .exact("8.3.1")
+      .exact("8.7.0")
```
ncooke3, line 152: nit: update the above comment now that `scripts/setup_spm_test_app_measurement.sh` is being deleted

ncooke3 (commented):
LGTM on green

ncooke3 (approved):

htcgh (approved):

Package.swift
paulb777: Good catch! Thanks.

AdieOlami:
I think this is breaking my project as I cannot set up on a new machine. 

Dependencies could not be resolved because no versions of 'GoogleAppMeasurement' match the requirement 8.7.0 and root depends on 'GoogleAppMeasurement' 8.7.0.

paulb777:
We don't generally support working off the master branch. We're currently testing and staging the 8.7.0 release and it's not fully functional externally. Instead, you should use a release version as is documented at https://github.com/firebase/firebase-ios-sdk/blob/master/SwiftPackageManager.md.

AdieOlami:
Thanks for the response, however, using version 8.6.1 gives error `Missing package product 'GoogleAppMeasurement'` that was the reason for using master branch which worked prior.

paulb777:
You might need to `Reset Package Caches` with 8.6.1.

AdieOlami:
I had that issue the first time using SPM. I am doing a fresh install now with that version and that error keeps coming up. That was the reason why I opted to use the master branch then.

paulb777:
Hmm, I'm not sure what is going on. Please open an issue with the Xcode version and a reproducible example and we can take a look.

AdieOlami:
Works now after reseting cache.

AdieOlami:
I also just want to say you guys are doing an amazing amazing job.

PR #8610, opened 2021-09-01
Do not erase from the map while iterating over it.

ehsannas:


var-const (approved):

var-const:
This might be related to #8607.

var-const:
Might also be related to #8612.

PR #8614, opened 2021-09-02
Property wrapper for Firestore

peterfriese:
This is a continuation of #8408, including some explorations for error handling contributed by @mortenbekditlevsen.


Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+          // them in the `documents` property of a snapshot
+          try queryDocumentSnapshot.data(as: U.self)!
+        }
+      }
+    }
+
+    setupListener()
+  }
+
+  init<U: Decodable>(configuration: FirestoreQuery<T>.Configuration) where T == Result<[U], Error> {
```
peterfriese, line 89: This property wrapper offers 3 different ways for exposing fetched data. This method here makes data or errors available like this:

```swift
  @FirestoreQuery(collectionPath: "fruits") var fruitResults: Result<[Fruit], Error>
```

Later on in the code, the user can then check if `fruitResults` contains a .`success` or `.failure`:

```swift
  var body: some View {
    if case let .success(fruits) = fruitResults {
      List(fruits) { fruit in
        Text(fruit.name)
      }
    } else if case let .failure(error as NSError) = fruitResults {
      // Handle error
      Text("Couldn't map data: \(error)")
    }
  }
```

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+        let items = try documents.compactMap { queryDocumentSnapshot in
+          try queryDocumentSnapshot.data(as: U.self)
+        }
+        self?.items = .success(items)
+      } catch {
+        self?.items = .failure(error)
+      }
+    }
+
+    setupListener()
```
peterfriese, line 113: Do we have a canonical way to implement snapshot listeners in Swift, @morganchen12 and @schmidt-sebastian? In [my article](https://peterfriese.dev/firestore-codable-the-comprehensive-guide/), I've tried to condense what we've got in the documentation, but some aspects aren't documented very clearly.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+  init<U: Decodable>(configuration: FirestoreQuery<T>.Configuration) where T == Result<[U], Error> {
+    items = .success([])
+    self.configuration = configuration
+    setupListener = createListener { [weak self] querySnapshot, error in
+      if let error = error {
+        self?.items = .failure(error)
+        return
+      }
+
+      guard let documents = querySnapshot?.documents else {
```
peterfriese, line 98: Will a snapshot listener ever return a non-nil `querySnapshot` that has a `nil` `documents` reference?

@schmidt-sebastian @morganchen12

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+        self?.items = .success([])
+        return
+      }
+
+      do {
+        let items = try documents.compactMap { queryDocumentSnapshot in
```
peterfriese, line 104: @morganchen12 I know you had this `Result`-driven mapping code in https://firebase.google.com/docs/firestore/query-data/get-data#custom_objects - any suggestions how to make use of this here? I am not unhappy with how this code looks like, but was wondering if we can improve it.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      guard let documents = querySnapshot?.documents else {
+        self?.items = []
+        return
+      }
+
+      self?.items = documents.compactMap { queryDocumentSnapshot in
+        Result {
+          // NOTE: The `try` handles the parse failure
```
peterfriese, line 76: I am not 100% sure this is correct. @schmidt-sebastian, can you comment?

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+        self?.items = []
+        return
+      }
+
+      guard let documents = querySnapshot?.documents else {
+        self?.items = []
+        return
+      }
+
+      self?.items = documents.compactMap { document in
```
peterfriese, line 52: In all three different scenarios, we use `compactMap` to filter out any `nil` results. This is mostly convenience, to relieve developers from having to `compactMap` the results themselves (which would somewhat defat the whole purpose of having a one-stop property wrapper), but do you think we might have situations where users would *want* to see `nil` values in the result set, @morganchen12 @schmidt-sebastian?

chFlorian:
@googlebot I consent.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
mortenbekditlevsen: I could more easily find a use case for preserving nil values for a wrapped value that were a Dictionary from key to decoded entity.
With the exploration in my PR you can basically add overloads for many wrapped types. This could perhaps be used to provide several combinations of values or optional values or result values - in arrays or in dictionaries keyed by the document id - and all of the above wrapped in outer results.
The possibility is there, I just can't say if it's a nice API. 😊

peterfriese: > The possibility is there, I just can't say if it's a nice API. 😊

Agree - if the community requests this, it should be easy enough to extend the API accordingly.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
morganchen12: We do not, though the method in your article feels a lot like the culmination of all of the scattered recommendations in our docs. Which aspects are you referring to in particular?

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
morganchen12: Yes, since users can explicitly set nil documents. A non-nil query snapshot with a nil document indicates a successful fetch of nil, which, while confusing if not an intentional choice by the developer, is not an error.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    setupListener()
+  }
+
+  init<U: Decodable>(configuration: FirestoreQuery<T>.Configuration) where T == Result<[U], Error> {
+    items = .success([])
+    self.configuration = configuration
+    setupListener = createListener { [weak self] querySnapshot, error in
+      if let error = error {
```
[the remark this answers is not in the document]
morganchen12: That can be found here: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Public/FirebaseFirestore/FIRFirestoreErrors.h

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
morganchen12: Honestly, using Swift's built-in `try` makes this code nearly identical to the Result code I wrote for that snippet. I'm not too worried about it.

The only advantage of the `Result` is you could potentially pass the whole enum off somewhere else to handle in a different location.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
schmidt-sebastian: Documents in a QuerySnapshot always exist.

This is not true in a DocumentSnapshot (for single document fetches). We need to keep that in mind when expanding this to single document fetches.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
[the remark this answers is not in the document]
schmidt-sebastian: (in addition to decoding issues if I am not mistaken)

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
schmidt-sebastian: I am not suer I can follow this. A user can set `nil` values, but I don't see how documents can be nil here. It could very well be an empty list.

peterfriese: @morganchen12 interesting - I did not know this. Can this be done in code only? This doesn't seem to be possible via the console.

@schmidt-sebastian are you saying that the Objective-C code never passes a `nil` value in `querySnapshot.documents`?

schmidt-sebastian: Yes.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,175 @@
...
+
+  private var setupListener: (() -> Void)!
+
+  internal var preventUpdates = false
+  internal var configuration: FirestoreQuery<T>.Configuration {
+    didSet {
+      if !preventUpdates {
+        removeListener()
+        setupListener()
+      }
```
ncooke3, line 35: WDYT about changing `preventUpdate` to `shouldUpdate`/`shouldUpdateListener` (and flipping all assigned values to account for renaming).

There is something about `!preventUpdates` that reads like a double negative and confuses me.
Proposing:
```swift
// --snip --
  internal var shouldUpdateListener = true
  internal var configuration: FirestoreQuery<T>.Configuration {
    didSet {
      if shouldUpdateListener {
        removeListener()
        setupListener()
      }
// --snip --
```
And, later in `projectError(_ error:)`:
```swift
  private func projectError(_ error: Error?) {
    shouldUpdateListener = false
    configuration.error = error
    shouldUpdateListener = true
  }
```

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -0,0 +1,138 @@
...
+///         collectionPath: "fruits",
+///         predicates: [.whereField("isFavourite", isEqualTo: true)]
+///       ) var fruits: [Fruit]
+///
+///       var body: some View {
+///         List(fruits) { fruit in
+///           Text(fruit.name)
+///         }
+///       }
+///     }
```
ncooke3, line 40: The code example in the doc comment is awesome. WDYT about including another one for the Result type implementation, or linking to your sample app?

Example/FirestoreSample/FirestoreSample/Views/FavouriteFruitsView.swift
```diff
@@ -0,0 +1,72 @@
...
+    collectionPath: "fruits",
+    predicates: [
+      .where("isFavourite", isEqualTo: true),
+    ]
+  ) fileprivate var fruitResults: Result<[Fruit], Error>
+
+  @State var showOnlyFavourites = true
+
+  var body: some View {
+    if case let .success(fruits) = fruitResults {
```
ncooke3, line 35: this is great!

At least in the newer Xcodes, you can use a switch statement with implicit returns for some elegant approaches!
```swift
var body: some View {
      switch fruitResults {
      case .success(fruits):
        FruitsList(fruits)
      case .failure(error):
        EmptyView(error)
      }
}
```

ncooke3 (commented):
LGTM after addressing a few comments, questions, and suggestions.

This goes without saying but this is **amazing** work! Congrats to everyone that has contributed on this! 👏

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,175 @@
...
+ */
+
+import SwiftUI
+import FirebaseFirestore
+
+@available(iOS 14.0, macOS 11.0, macCatalyst 14.0, tvOS 14.0, watchOS 7.0, *)
+internal class FirestoreQueryObservable<T>: ObservableObject {
+  @Published var items: T
+
+  private let firestore = Firestore.firestore()
```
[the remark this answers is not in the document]
chFlorian: I'm not sure wether this is a usecase for the property wrapper, as it is meant to be used inside a `View`, similarly to an `@FetchRequest`.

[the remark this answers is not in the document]
ncooke3: Aha I see, SGTM. I'll resolve this then 👍

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
peterfriese: The reason why I chose `preventUpdates` is very specific - we need to actively _prevent_ any updates to `configuration.error` to trigger a SwiftUI state update  cycle (as this results in an endless loop, no matter the state of the error).

You will notice that updates will only be blocked when assigning `error`.

That's why I chose `prevent`. I also considered `blockUpdates`, `stopBubblingUp` (just briefly), and others. If anyone has any better names, LMK.

Example/FirestoreSample/FirestoreSample/Views/FavouriteFruitsView.swift
peterfriese: That's a great suggestion - my goal is to strike a balance between making the code easy to understand for casual reader who are looking for an example how to use this in their own apps (i.e. potentially just wrapping their existing code in a conditional statement), and making it as elegant as possible (as one might do for a production app).

People new to SwiftUI might not be immediately familiar with the idea of extracting views into sub views, properties, or view builders.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
```diff
@@ -0,0 +1,175 @@
...
+        self?.projectError(error)
+        return
+      }
+
+      guard let documents = querySnapshot?.documents else {
+        self?.items = []
+        return
+      }
+
+      self?.projectError(nil)
```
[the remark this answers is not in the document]
peterfriese: I thing you're right - I moved this up.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
ncooke3: Aha I see the endless loop issue now, that's smart 👍  `preventUpdates` LGTM

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
peterfriese: Good idea! Added.

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
peterfriese: After a bit of deliberation, changed the name to `shouldUpdateListener`, used a `guard` , and added a comment to make this clearer.

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
```diff
@@ -39,10 +48,54 @@ public enum DecodingFailureStrategy {
...
+///               Text("There was an error")
+///                 .foregroundColor(Color(UIColor.systemBackground))
+///               Spacer()
+///             }
+///             .padding(30)
+///             .background(Color.red)
+///           }
+///         }
+///       }
+///     }`
```
ncooke3, line 98: nit: looks like there is a trailing ` character

Firestore/Swift/Source/PropertyWrapper/FirestoreQueryObservable.swift
ncooke3: Latest revisions look great! Thanks 👍

ncooke3 (approved):
LGTM. The docs are amazing @peterfriese!

Firestore/Swift/Source/PropertyWrapper/FirestoreQuery.swift
peterfriese: Eagle-eyed review! I wonder how this sneaked in...

peterfriese:
API Request has been approved!

Issue #8286, opened 2021-06-22
Possible bug in “queryOrdered(byChild:).queryEqual(toValue:).getData" function


fawzirifai:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: Version 12.5 (12E262)
  * Firebase SDK version: 8.1.1
  * Installation method: Swift Package Manager
  * Firebase Component: Database

### [REQUIRED] Step 2: Describe the problem
I’m observing how many children a person called "james" has using `child("james").child("children").observe(.value)` function, in my case he has 2 children called "child1" and "child2" so it prints `children count = 2`. Then I check if "james" has a child called "child2" using `child("james").child("children").queryOrdered(byChild: "name").queryEqual(toValue: "child2").getData` function, in my case he has a child called "child2" so it prints `found`. Then _unexpectedly_ the first `observe` function that observe and print children count triggers again and prints `children count = 1`. I think the first `observe` function should not be triggered and the children count should remain 2. If I replaced `getData` function with `observe` function to search for a child named "child2" it finds it and prints `found` and the first `observe` function that observe and print children count not triggered and the children count remains 2.

#### Steps to reproduce:
I’m observing how many children a person called "james" has using `child("james").child("children").observe(.value)` function, in my case he has 2 children called "child1" and "child2" so it prints `children count = 2`. Then I check if "james" has a child called "child2" using `child("james").child("children").queryOrdered(byChild: "name").queryEqual(toValue: "child2").getData` function, in my case he has a child called "child2" so it prints `found`. Then _unexpectedly_ the first `observe` function that observe and print children count triggers again and prints `children count = 1`. I think the first `observe` function should not be triggered and the children count should remain 2. If I replaced `getData` function with `observe` function to search for a child named "child2" it finds it and prints `found` and the first `observe` function that observe and print children count not triggered and the children count remains 2.

What happened? How can we make the problem occur?
[...]

PR #8618, opened 2021-09-03
attempt to reproduce #8286

IanWyszynski:


FirebaseDatabase/Tests/Integration/FIRDatabaseQueryTests.m
```diff
@@ -4468,4 +4468,67 @@ - (void)testGetSkipsPersistenceCacheWhenOnline {
...
 
+- (void)testGetTriggersListener {
+  FIRDatabase* db = [self databaseForURL:self.databaseURL name:@"fir-multi-shards"];
+
+  FIRDatabaseReference* ref = [db referenceWithPath:@"james"];
+  FIRDatabaseReference* children = [ref child:@"children"];
+
+  __block BOOL done = NO;
+
+  [children removeValueWithCompletionBlock:^(NSError* _Nullable error,
```
schmidt-sebastian, line 4479: You could use an `autoId` instead of `children` (via `[FTestHelpers getRandomNode]`). Then you won't have to do a delete.

FirebaseDatabase/Tests/Integration/FIRDatabaseQueryTests.m
```diff
@@ -4468,4 +4468,67 @@ - (void)testGetSkipsPersistenceCacheWhenOnline {
...
+                                             FIRDatabaseReference* _Nonnull ref) {
+    XCTAssertNil(error);
+    done = YES;
+  }];
+
+  WAIT_FOR(done);
+  done = NO;
+
+  [[children childByAutoId] setValue:@{@"name" : @1}
```
schmidt-sebastian, line 4488: One difference between this test and the issue is that the test does not write values from the same instance, but rather all values exist on the backend. We can emulate this using a reader/writer test setup.

FirebaseDatabase/Tests/Integration/FIRDatabaseQueryTests.m
```diff
@@ -4468,4 +4468,67 @@ - (void)testGetSkipsPersistenceCacheWhenOnline {
...
+                     if (done == YES) {
+                       XCTAssertTrue(snapshot.childrenCount == 2U);
+                       nextDone = YES;
+                       return;
+                     }
+                     XCTAssertTrue(snapshot.childrenCount == 2U);
+                     done = YES;
+                   }];
+
+  WAIT_FOR(done);
```
schmidt-sebastian, line 4519: The original repo also does not wait for "done". It runs both operation in parallel.

Issue #7977, opened 2021-04-27
Deadlock caused by RemoteConfig fetch


1oo7:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.3
  * Firebase SDK version: 7.11
  * Installation method: `Zip file` (select one)
  * Firebase Component: RemoteConfig (Auth, Core, Database, Firestore, Messaging, Storage, etc)

### [REQUIRED] Step 2: Describe the problem

We are experiencing a deadlock when performing a "fetch" RemoteConfig operation on a fresh install of the app.

On the main thread we have:

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x00000001b1c10404 in kevent_id ()
#1	0x000000010f6a5ee0 in _dispatch_kq_poll ()
#2	0x000000010f6a6988 in _dispatch_event_loop_wait_for_ownership ()
#3	0x000000010f694a74 in __DISPATCH_WAIT_FOR_QUEUE__ ()
#4	0x000000010f6944c0 in _dispatch_sync_f_slow ()
#5	0x0000000109736960 in -[FIRRemoteConfig configValueForKey:] ()
#6	0x0000000109805368 in __41-[FPRRemoteConfigFlags cacheConfigValues]_block_invoke ()
#7	0x00000001863e2cc8 in __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__ ()
#8	0x00000001863610bc in -[__NSFrozenDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:] ()
#9	0x00000001098052e4 in -[FPRRemoteConfigFlags cacheConfigValues] ()
#10	0x0000000109805128 in __30-[FPRRemoteConfigFlags update]_block_invoke ()
[...]

PR #8626, opened 2021-09-06
Fix RemoteConfig multithreading

ArtVolkov:
Sorry, here is story:
Some time ago I had to use FirebaseRemoteConfig 7.4.0.
So I encountered problem https://github.com/firebase/firebase-ios-sdk/pull/7975 and fixed it on my own in private fork.
Here is PR to adjust current master with parts of my fix. 
I hope it will increase thread safety, and will possibly fix https://github.com/firebase/firebase-ios-sdk/issues/7977. 
Please read on for details.

#### (I) Main idea

Replace 
`semaphore 'wait/signal'`
with
`dispatch_group 'enter/wait/leave'`

#### (II) General thoughts

_Semaphore_ is one of most generic synchronisation mechanisms, so it can be used to solve probably any synch problem.
But _semaphore_ requires to thoroughly craft special code for every small slightly non-trivial problem.
On other side _dispatch_group_ is not so generic, but it's already crafted specially for case that we have in `RCNConfigContent.m`.
It's purpose of `dispatch_group 'enter/wait/leave'` to ensure that some threads blocked until some other threads complete their work.
So I believe that this out of the box solution will be less error prone.

#### (III) Concrete caveats with current code in master in `RCNConfigContent.m`

Caveat 1.
```
[_DBManager
      loadMainWithBundleIdentifier:_bundleIdentifier
                 completionHandler:^(BOOL success, NSDictionary *fetchedConfig,
                                     NSDictionary *activeConfig, NSDictionary *defaultConfig) {
                   self->_fetchedConfig = [fetchedConfig mutableCopy];
                   self->_activeConfig = [activeConfig mutableCopy];
                   self->_defaultConfig = [defaultConfig mutableCopy];
                   dispatch_semaphore_signal(self->_configLoadFromDBSemaphore);
                 }];

  // TODO(karenzeng): Refactor personalization to be returned in loadMainWithBundleIdentifier above
  [_DBManager loadPersonalizationWithCompletionHandler:^(
                  BOOL success, NSDictionary *fetchedPersonalization,
                  NSDictionary *activePersonalization, NSDictionary *defaultConfig) {
    self->_fetchedPersonalization = [fetchedPersonalization copy];
    self->_activePersonalization = [activePersonalization copy];
  }];
```
`loadPersonalizationWithCompletionHandler` has no synch with semaphore. So later in `checkAndWaitForInitialDatabaseLoad` we have no guarantees that `loadPersonalizationWithCompletionHandler` completion finished.

 Caveat 2.
```
- (BOOL)checkAndWaitForInitialDatabaseLoad {
  /// Wait on semaphore until done. This should be a no-op for subsequent calls.
  if (!_isConfigLoadFromDBCompleted) {
    long result = dispatch_semaphore_wait(
        _configLoadFromDBSemaphore,
        dispatch_time(DISPATCH_TIME_NOW, (int64_t)(kDatabaseLoadTimeoutSecs * NSEC_PER_SEC)));
    if (result != 0) {
      FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000048",
                  @"Timed out waiting for fetched config to be loaded from DB");
      return false;
    }
    _isConfigLoadFromDBCompleted = true;
  }
  return true;
}
```
Let's imagine we have multiple threads waiting for semaphore. When `dispatch_semaphore_signal` fired from `loadMainWithBundleIdentifier` it will wake _only one_ thread. So all other thread will wait until timeout. In later calls of cause `_isConfigLoadFromDBCompleted` will save situation, but some first calls will possibly be stuck.

#### (IV) Details on this PR

1. Changes in`RCNConfigContent.m` remove caveats (1) and (2) 
2. Changes in`RCNConfigDBManager.m` address slightly different but related to work of `checkAndWaitForInitialDatabaseLoad` moments:
2.1. Change: two new `handler(NO, ...` make potential work of subsequent `return` statement more consistent.
2.2. Change: replacement `dispatch_async(dispatch_get_main_queue()...` with `dispatch_async(dispatch_get_global_queue` should increase general performance of RemoteConfig because initialisation of RemoteConfig often runs while app launching and main thread is very busy.

### Testing

  * All tests in RemoteConfig pass with changes from this PR.
  * Did not added new tests as it's very hard to test multithreading.

### API Changes

  * No APi changes.

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
@@ -868,6 +868,9 @@ static NSArray *RemoteConfigMetadataTableColumnsInOrder() {
   dispatch_async(_databaseOperationQueue, ^{
     RCNConfigDBManager *strongSelf = weakSelf;
     if (!strongSelf) {
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        handler(NO, [NSMutableDictionary new], [NSMutableDictionary new], nil);
+      });
       return;
     }
 
@@ -898,7 +901,7 @@ static NSArray *RemoteConfigMetadataTableColumnsInOrder() {
     }
 
     if (handler) {
-      dispatch_async(dispatch_get_main_queue(), ^{
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
         handler(YES, fetchedPersonalization, activePersonalization, nil);
       });
     }
@@ -972,6 +975,9 @@ static NSArray *RemoteConfigMetadataTableColumnsInOrder() {
   dispatch_async(_databaseOperationQueue, ^{
     RCNConfigDBManager *strongSelf = weakSelf;
     if (!strongSelf) {
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        handler(NO, [NSDictionary new], [NSDictionary new], [NSDictionary new]);
+      });
       return;
     }
     __block NSDictionary *fetchedConfig =
--- a/FirebaseRemoteConfig/Sources/RCNConfigContent.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigContent.m
@@ -41,10 +41,10 @@
   RCNConfigDBManager *_DBManager;
   /// Current bundle identifier;
   NSString *_bundleIdentifier;
-  /// Dispatch semaphore to block all config reads until we have read from the database. This only
+  /// Blocks all config reads until we have read from the database. This only
   /// potentially blocks on the first read. Should be a no-wait for all subsequent reads once we
   /// have data read into memory from the database.
-  dispatch_semaphore_t _configLoadFromDBSemaphore;
+  dispatch_group_t _dispatch_group;
   /// Boolean indicating if initial DB load of fetched,active and default config has succeeded.
   BOOL _isConfigLoadFromDBCompleted;
   /// Boolean indicating that the load from database has initiated at least once.
@@ -86,7 +86,8 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
       _bundleIdentifier = @"";
     }
     _DBManager = DBManager;
-    _configLoadFromDBSemaphore = dispatch_semaphore_create(0);
+    // Waits for both config and Personalization data to load.
+    _dispatch_group = dispatch_group_create();
     [self loadConfigFromMainTable];
   }
   return self;
@@ -112,6 +113,7 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
   NSAssert(!_isDatabaseLoadAlreadyInitiated, @"Database load has already been initiated");
   _isDatabaseLoadAlreadyInitiated = true;
 
+  dispatch_group_enter(_dispatch_group);
   [_DBManager
       loadMainWithBundleIdentifier:_bundleIdentifier
                  completionHandler:^(BOOL success, NSDictionary *fetchedConfig,
@@ -119,15 +121,17 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
                    self->_fetchedConfig = [fetchedConfig mutableCopy];
                    self->_activeConfig = [activeConfig mutableCopy];
                    self->_defaultConfig = [defaultConfig mutableCopy];
-                   dispatch_semaphore_signal(self->_configLoadFromDBSemaphore);
+                   dispatch_group_leave(self->_dispatch_group);
                  }];
 
   // TODO(karenzeng): Refactor personalization to be returned in loadMainWithBundleIdentifier above
+  dispatch_group_enter(_dispatch_group);
   [_DBManager loadPersonalizationWithCompletionHandler:^(
                   BOOL success, NSDictionary *fetchedPersonalization,
                   NSDictionary *activePersonalization, NSDictionary *defaultConfig) {
     self->_fetchedPersonalization = [fetchedPersonalization copy];
     self->_activePersonalization = [activePersonalization copy];
+    dispatch_group_leave(self->_dispatch_group);
   }];
 }
 
@@ -359,12 +363,11 @@ static const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
 /// configs until load is done.
 /// @return Database load completion status.
 - (BOOL)checkAndWaitForInitialDatabaseLoad {
-  /// Wait on semaphore until done. This should be a no-op for subsequent calls.
+  /// Wait until load is done. This should be a no-op for subsequent calls.
   if (!_isConfigLoadFromDBCompleted) {
-    long result = dispatch_semaphore_wait(
-        _configLoadFromDBSemaphore,
-        dispatch_time(DISPATCH_TIME_NOW, (int64_t)(kDatabaseLoadTimeoutSecs * NSEC_PER_SEC)));
-    if (result != 0) {
+    intptr_t isErrorOrTimeout =
+        dispatch_group_wait(_dispatch_group, kDatabaseLoadTimeoutSecs * NSEC_PER_SEC);
+    if (isErrorOrTimeout) {
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000048",
                   @"Timed out waiting for fetched config to be loaded from DB");
       return false;


FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
```diff
@@ -898,7 +901,7 @@ - (void)loadPersonalizationWithCompletionHandler:(RCNDBLoadCompletion)handler {
     }
 
     if (handler) {
-      dispatch_async(dispatch_get_main_queue(), ^{
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
maksymmalyhin, line 904: Note for reviewers. Validate if it affects calling a public callback. If yes, it's a breaking change and will require a major version bump.

maksymmalyhin:
@ArtVolkov Thanks a lot for your investigation and proposed fix! From the first glance the change looks good, but we will need a bit more time to do more detailed review.

maksymmalyhin:
As far as I can see the changes are not covered by unit tests. We most likely will need to provide tests before merging it to master. @ArtVolkov Will you have time to provide the tests. Seems like updating [RCNConfigContentTest](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m) will require significant changes. Please let us know if you are interested in updating the tests yourself and if you need help with it.

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
ArtVolkov: Just for additional info: 
1. There is same kind of callback in `- (void)loadMainWithBundleIdentifier` in this file.
In `- (void)loadMainWithBundleIdentifier` callback already looks like:
```
    if (handler) {
      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
      fetchedConfig = fetchedConfig ? fetchedConfig : [[NSDictionary alloc] init];
      activeConfig = activeConfig ? activeConfig : [[NSDictionary alloc] init];
      defaultConfig = defaultConfig ? defaultConfig : [[NSDictionary alloc] init];
      handler(YES, fetchedConfig, activeConfig, defaultConfig);
    });
```
2. Another thought on this is that "Personalization" feature is in alpha now (afaik).

But if this change still needs major version bump, then I can move it to separate PR, to make main part sooner to merge.

ArtVolkov:
> As far as I can see the changes are not covered by unit tests. We most likely will need to provide tests before merging it to master. @ArtVolkov Will you have time to provide the tests. Seems like updating [RCNConfigContentTest](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m) will require significant changes. Please let us know if you are interested in updating the tests yourself and if you need help with it.

Hi @maksymmalyhin! Yes, I interested to got this PR merged. I will try my best to look closer to RCNConfigContentTest during this week.

maksymmalyhin:
> Hi @maksymmalyhin! Yes, I interested to got this PR merged. I will try my best to look closer to RCNConfigContentTest during this week.

Thank you! Alternatively one of us can implement the tests on top of your changes if you don't have enough time to do it, just let us know.

ArtVolkov:
> Thank you! Alternatively one of us can implement the tests on top of your changes if you don't have enough time to do it, just let us know.

Adjusted RCNConfigContentTest with test for `-checkAndWaitForInitialDatabaseLoad`, any comments appreciated.

ArtVolkov:
Fixed stile of my changes in RCNConfigContentTest.

ArtVolkov:
@maksymmalyhin @karenyz Did I miss something to do? Please guide me, as I newbie at contributing.

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
maksymmalyhin: Thank you for pointing out to that the handler is already called on the global queue in some cases. Modifications in the file look good to me. 

cc @karenyz

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
```diff
@@ -25,7 +25,36 @@
...
+}
+- (void)loadPersonalizationWithCompletionHandler:(RCNDBLoadCompletion)handler {
+  double justOtherSmallDelay = 0.009;
+  XCTAssertTrue(justOtherSmallDelay < kDatabaseLoadTimeoutSecs);
+  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(justOtherSmallDelay * NSEC_PER_SEC)),
+                 dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                   self.isLoadPersonalizationCompleted = YES;
+                   handler(YES, nil, nil, nil);
+                 });
+}
```
maksymmalyhin, line 57: I can see that OCMock is used for in other tests for mocking, but defining the mock as a separate class doesn't look like an issue to me in this case. @karenyz WDYT?

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
```diff
@@ -262,4 +291,38 @@ - (void)testCopyFromDictionaryUpdatesActiveConfig {
                         [_configContent.activeConfig[@"dummy_namespace"][@"new_key"] dataValue]);
 }
 
+- (void)testCheckAndWaitForInitialDatabaseLoad {
```
maksymmalyhin, line 294: @ArtVolkov  Great, thank you for adding the test! Does it fail if running without the code change?

maksymmalyhin (commented):
@ArtVolkov Thank you for your contribution and sorry for the slow review. The change looks good to me but I defer the final approval to @karenyz as the maintainer of the SDK.

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
ArtVolkov: AFAIK OCMock can not mock `- init`. But I wanted to override `- (void)createOrOpenDatabase` before `- init` fired to prevent heavy code of database loading.

Also in current way:
```
@property(nonatomic, assign) BOOL isLoadMainCompleted;
@property(nonatomic, assign) BOOL isLoadPersonalizationCompleted;
```
located in separate mock class, not in unit test class itself.

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
ArtVolkov: @maksymmalyhin. Yes, two falls were here with old code: one for each caveat in PR description.

maksymmalyhin: Great, thank you for confirming.

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
maksymmalyhin: With OCMock approach you probably would create a instance of a class mock with `OCMClassMock()`  macro and configure stubs and expectations before passing it to the `RCNConfigContent` initializer. There are examples of this approach in the repo e.g. a [relatively simple one](https://github.com/firebase/firebase-ios-sdk/blob/23b6acefb9afaf7dc4f5471a14fc087971878072/FirebaseRemoteConfig/Tests/Unit/RCNConfigAnalyticsTest.m#L32)  or more advanced like [here](https://github.com/firebase/firebase-ios-sdk/blob/23b6acefb9afaf7dc4f5471a14fc087971878072/FirebaseInstallations/Source/Tests/Unit/FIRInstallationsStoreTests.m#L140) or [here](https://github.com/firebase/firebase-ios-sdk/blob/23b6acefb9afaf7dc4f5471a14fc087971878072/FirebaseAuth/Tests/Unit/FIRPhoneAuthProviderTests.m#L537).

We usually prefer using OCMock to make our tests more uniform. I can see room for better test coverage for this class and adopting OCMock now should make this work easier in the future. On the other hand we have exceptions as well where manually created mocks/fakes are used, so I personally don't think it is a blocker for the current PR.

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
```diff
@@ -262,4 +291,38 @@ - (void)testCopyFromDictionaryUpdatesActiveConfig {
...
+  RCNConfigDBManagerMock *mockDBManager = [[RCNConfigDBManagerMock alloc] init];
+  RCNConfigContent *configContent = [[RCNConfigContent alloc] initWithDBManager:mockDBManager];
+
+  // Check that no one of first three calls of `-checkAndWaitForInitialDatabaseLoad` do not produce
+  // timeout error <begin>
+  XCTestExpectation *expectation1 =
+      [self expectationWithDescription:
+                @"1st `checkAndWaitForInitialDatabaseLoad` return without timeout"];
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [configContent checkAndWaitForInitialDatabaseLoad];
```
karenyz, line 304: should we assert that `checkAndWaitForInitialDatabaseLoad` returns true here too?

karenyz (approved):
thanks for the detailed description of the problem/fix! good to know about `dispatch_group` and its usage :)

FirebaseRemoteConfig/Tests/Unit/RCNConfigContentTest.m
ArtVolkov: It would be cool to have this assert while it not strictly necessary. I will add it now.

Additional info: it seems safe to put XCTAssert in non-main queue. 
Apple example: https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations/testing_asynchronous_operations_with_expectations?language=objc
Callback in example "is executed on the delegate queue" https://developer.apple.com/documentation/foundation/nsurlsession/1410330-datataskwithurl?language=objc

ArtVolkov: @karenyz ready :)

morganchen12:
Deferring to @karenyz to merge. Thanks @ArtVolkov for the contribution!

ArtVolkov:
@maksymmalyhin @karenyz @morganchen12 many thanks for your support!

Issue #8577, opened 2021-08-26
FPRDecodeString causes heap-buffer overflow (potential security issue)


ortekka:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.5.1
  * Firebase SDK version: 8.6.0
  * Installation method: CocoaPods
  * Firebase Component: Performance

### [REQUIRED] Step 2: Describe the problem

While running an app that includes Firebase Performance with the Address Sanitizer on, a heap-buffer-overflow was reported in FPRDecodeString() (FPRNanoPbUtils.m:160).

FPRDecodeString() is calling [NSString stringWithCString:[data bytes] encoding:NSUTF8StringEncoding] without making sure that data contains a NUL terminator.  In fact, in this particular instance, data does not contains NULs, and stringWithCString keeps scanning for NULs beyond the end of the buffer.

What follows is a full copy of the console log:

```
VocalFeel(45943,0x10de71e00) malloc: nano zone abandoned due to inability to preallocate reserved vm space.
objc[45943]: Class _NSZombie_OS_xpc_uint64 is implemented in both ?? (0x60400008f990) and ?? (0x60400008dad0). One of the two will be used. Which one is undefined.
2021-08-26 09:54:07.700467+0200 VocalFeel[45943:5299360] [Firebase/Crashlytics] Version 8.6.0
2021-08-26 09:54:07.961767+0200 VocalFeel[45943:5299818] [Client] Synchronous remote object proxy returned error: Error Domain=NSCocoaErrorDomain Code=4099 "The connection to service on pid 0 named com.apple.commcenter.coretelephony.xpc was invalidated." UserInfo={NSDebugDescription=The connection to service on pid 0 named com.apple.commcenter.coretelephony.xpc was invalidated.}
[...]

PR #8631, opened 2021-09-08
Fix heap-buffer overflow from FPRDecodeString 

### LINKED ISSUE #8628: crash on FPRDecodeString when using Thread Sanitizer

    file a Github issue.
    with the `firebase` tag.
    google group.

  * Xcode version: 12.5.1
  * Firebase SDK version: 8.6.0
  * Firebase Component: Analytics, Crashlytics, Performance

#### Steps to reproduce:

For reproducing, I just check the Adress Sanitizer in Edit Scheme > Run and I run my project, it crashes systematically.

Capture d’écran 2021-09-07 à 14 05 10
Capture d’écran 2021-09-07 à 14 05 24

jeremyjiang-dev:
This is to fix #8628 and #8577 
The issue was not reproducible.
The fix is to remove the usage of FPRDecodeString  in the code.


visumickey (commented):
@jeremyjiang-dev Since FPRClient is the only place where we are using FPRDecodeString, we might be safe here. But, I think we need to get to the root cause if the code exists in our ecosystem. As an alternate, do you think we can FPRDecodeString as an API? This is not used anywhere else except for unit tests.

Heap overflow occurs when there is a byte array conversion to NSData. If the byte array does not contain a 'null' or '\0' at the end, there is no way for the system to undertsand where the character array ends.

One way to attempt to reproduce this issue would be to force a really long character array to see if FPRDecodeString cannot decode it.

jeremyjiang-dev:
> @jeremyjiang-dev Since FPRClient is the only place where we are using FPRDecodeString, we might be safe here. But, I think we need to get to the root cause if the code exists in our ecosystem. As an alternate, do you think we can FPRDecodeString as an API? This is not used anywhere else except for unit tests.
> 
> Heap overflow occurs when there is a byte array conversion to NSData. If the byte array does not contain a 'null' or '\0' at the end, there is no way for the system to undertsand where the character array ends.
> 
> One way to attempt to reproduce this issue would be to force a really long character array to see if FPRDecodeString cannot decode it.

Discussed offline. I moved the decoding methods to testUtil and note the methods may cause heap-buffer overflow.

FirebasePerformance/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Pending
 * Create a random number of delay for remote config fetch during app starts.
 * Fix log spamming when Firebase Performance is disabled. (#8423)
+* Fix heap-heap-buffer overflow when decoding strings. (#8628)
```
visumickey, line 4: I thought there were 2 bugs related to this. Can you tag both of them here?

FirebasePerformance/Tests/Unit/FPRTestUtils.m
```diff
@@ -153,4 +154,39 @@ + (GDTCOREvent *)createRandomNetworkGDTEvent:(NSString *)url {
   return gdtEvent;
 }
 
+#pragma mark - Decode nanoPb pbData
+NSData *FPRDecodeData(pb_bytes_array_t *pbData) {
```
visumickey, line 158: nit: New line here.

visumickey (approved):

leotianlizhan (approved):
LGTM

FirebasePerformance/CHANGELOG.md
jeremyjiang-dev: Done.

FirebasePerformance/Tests/Unit/FPRTestUtils.m
jeremyjiang-dev: Done

PR #8634, opened 2021-09-08
Fix implementation of `Hash` for DocumentSnapshot.

ehsannas:
The implementation of the `Hash` function for DocumentSnapshots has regressed. Without this fix two document snapshots with the same key and different content have the same hash.

The previous implementation took all parameters (key, type, version, data, state) into account for calculating the hash. See [here](https://github.com/firebase/firebase-ios-sdk/blob/04ab074265633049d5518866515e3bd7fc0de246/Firestore/core/src/model/document.cc#L88) and [here](https://github.com/firebase/firebase-ios-sdk/blob/04ab074265633049d5518866515e3bd7fc0de246/Firestore/core/src/model/maybe_document.cc#L33).

I verified that the test added in the PR fails without the code change, and passes with the code change.

#no-changelog

### DIFF
--- a/Firestore/Example/Tests/API/FIRDocumentSnapshotTests.mm
+++ b/Firestore/Example/Tests/API/FIRDocumentSnapshotTests.mm
@@ -30,6 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)testEquals {
   FIRDocumentSnapshot *base = FSTTestDocSnapshot("rooms/foo", 1, @{@"a" : @1}, NO, NO);
   FIRDocumentSnapshot *baseDup = FSTTestDocSnapshot("rooms/foo", 1, @{@"a" : @1}, NO, NO);
+  FIRDocumentSnapshot *baseChangedData = FSTTestDocSnapshot("rooms/foo", 1, @{@"b" : @1}, NO, NO);
   FIRDocumentSnapshot *nilData = FSTTestDocSnapshot("rooms/foo", 1, nil, NO, NO);
   FIRDocumentSnapshot *nilDataDup = FSTTestDocSnapshot("rooms/foo", 1, nil, NO, NO);
   FIRDocumentSnapshot *differentPath = FSTTestDocSnapshot("rooms/bar", 1, @{@"a" : @1}, NO, NO);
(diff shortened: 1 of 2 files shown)


ehsannas:
#no-changelog

schmidt-sebastian:
This was actually done by design since we do not want the hash to change if the document contents change. By only hashing "final" values we can store the document in a hash map and find it later if any mutable value changes.

I think we should keep the behavior as is but add a comment that explains this behavior.

As a note, if we do keep this change then we should in fact add a changelog entry as this is a user visible change.

ehsannas:
Hmmm... I have mixed feelings about this change now 😄 ! The current implementation _is_ inconsistent with the Android [implementation](https://github.com/firebase/firebase-android-sdk/blob/82b02af331d4674682bfb90c195da45543b808c3/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java#L534). (And Unity users will see different hash results depending on whether they run on Android or iOS).

> we can store the document in a hash map

one could use the `key()` [getter](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/model/mutable_document.h#L146) and use those results as the keys in their hash maps.


> if we do keep this change then we should in fact add a changelog entry as this is a user visible change

Sure.

Does the current documentation provide any implementation details and/or guarantees about the hashes?

var-const:
Agree with Ehsan's points, particularly that this appears to differ from e.g. the Android SDK.

schmidt-sebastian:
The Android implementation of `DocumentSnapshot.hashCode()` can use the hash of the data as the data is guaranteed to not change at this point. Not that it does not call "internalDocument.hash()" but rather "internalDocument.getData().hash()". The hash for MutableDocument (which is what this PR changes) [does not](https://github.com/firebase/firebase-android-sdk/blob/82b02af331d4674682bfb90c195da45543b808c3/firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java#L241) hash the ObjectValue since it is still mutable at this point.

Another difference is that hashing ObjectValues on Android is generally pretty cheap as it uses memoized hash codes. The iOS implementation is [very expensive](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/model/object_value.cc#L342).

We do not promise any guarantees about our hash function outside of what software design dictates. I personally don't think that just hashing on the key should cause any problems for our users.

schmidt-sebastian:
cc @dconeybe who reviewed the Android implementation. It took some convincing to get me to agree with the basic assumption that we should not hash mutable data, but I am fully on board now.

PR #8644, opened 2021-09-10
Remove potentially unnecessary secret from workflows

ncooke3:
Removing a secret from a workflow that is no longer used in the objc quickstart testing.

Comment with some context below:
> So it looks like account in FIREGSignInInfo.h doesn't get tested anymore. Related to b/140411106 which is marked obsolete. And FWIW, these tests were in the legacy ObjC Auth quickstart.

#no-changelog


ncooke3:
So it looks like account in `FIREGSignInInfo.h` doesn't get tested anymore. Related to b/140411106 which is marked obsolete. And FWIW, these tests were in the legacy ObjC Auth quickstart.

ncooke3:
Re-running CI since it's been a few days and quickstart secrets have been updated.

morganchen12 (approved):

PR #8645, opened 2021-09-10
Add back fix for g3 warnings

ncooke3:
Looks like we still need to do the weak self bit to avoid a fatal buildtime warning that occurs when copybara'ing into g3. 

#no-changelog


morganchen12 (approved):

maksymmalyhin (approved):
Thanks!

PR #8647, opened 2021-09-10
Do not consider MutableDocument's state for DocumentSnapshot comparison

ehsannas:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/Firestore/core/src/api/document_snapshot.cc
+++ b/Firestore/core/src/api/document_snapshot.cc
@@ -88,12 +88,12 @@ absl::optional<google_firestore_v1_Value> DocumentSnapshot::GetValue(
 bool operator==(const DocumentSnapshot& lhs, const DocumentSnapshot& rhs) {
   return lhs.firestore_ == rhs.firestore_ &&
          lhs.internal_key_ == rhs.internal_key_ &&
-         lhs.internal_document_.has_value() ==
-             rhs.internal_document_.has_value() &&
-         (lhs.internal_document_.has_value()
-              ? lhs.internal_document_->get().CompareIgnoringState(
-                    rhs.internal_document_->get())
-              : true) &&
+         lhs.exists() == rhs.exists() &&
+         (lhs.exists() ? lhs.internal_document_->get().key() ==
+                                 rhs.internal_document_->get().key() &&
+                             lhs.internal_document_->get().data() ==
+                                 rhs.internal_document_->get().data()
+                       : true) &&
          lhs.metadata_ == rhs.metadata_;
 }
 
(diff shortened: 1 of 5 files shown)


Firestore/core/src/model/mutable_document.cc
```diff
@@ -113,6 +113,11 @@ std::string MutableDocument::ToString() const {
   return stream.str();
 }
 
+bool MutableDocument::CompareIgnoringState(const MutableDocument& other) const {
```
schmidt-sebastian, line 116: If we keep this, this should be called "EqualsSomethingSomething" as it does not compare documents.

Firestore/core/src/api/document_snapshot.cc
```diff
@@ -88,7 +88,12 @@ absl::optional<google_firestore_v1_Value> DocumentSnapshot::GetValue(
 bool operator==(const DocumentSnapshot& lhs, const DocumentSnapshot& rhs) {
   return lhs.firestore_ == rhs.firestore_ &&
          lhs.internal_key_ == rhs.internal_key_ &&
-         lhs.internal_document_ == rhs.internal_document_ &&
+         lhs.internal_document_.has_value() ==
+             rhs.internal_document_.has_value() &&
+         (lhs.internal_document_.has_value()
+              ? lhs.internal_document_->get().CompareIgnoringState(
```
schmidt-sebastian, line 94: Nit (or personal preference): What do you think of comparing the value and the key from the document, and then using `exists()` and `snapshotMetadata` for the comparison of the remaining state (all encapsulated in this equals function).

Firestore/core/src/model/mutable_document.cc
```diff
@@ -113,6 +113,11 @@ std::string MutableDocument::ToString() const {
...
+  return key_ == other.key_ && document_type_ == other.document_type_ &&
+         version_ == other.version_ && *value_ == *other.value_;
```
schmidt-sebastian, line 118: I think we might want to ignore version as well since it is also not exposed.

schmidt-sebastian (commented):
Sorry for the slow review! Please ping me if you don't hear back within 24 hours (minus weekends - DO NOT PING ME ON WEEKENDS (*))

(*) at your own discretion.

schmidt-sebastian:
You should also add a changelog entry.

Firestore/core/src/api/document_snapshot.cc
ehsannas: Good idea. Done.

schmidt-sebastian (approved):

Firestore/core/src/api/document_snapshot.cc
```diff
@@ -88,7 +88,12 @@ absl::optional<google_firestore_v1_Value> DocumentSnapshot::GetValue(
 bool operator==(const DocumentSnapshot& lhs, const DocumentSnapshot& rhs) {
   return lhs.firestore_ == rhs.firestore_ &&
          lhs.internal_key_ == rhs.internal_key_ &&
-         lhs.internal_document_ == rhs.internal_document_ &&
+         lhs.exists() == rhs.exists() &&
+         (lhs.exists() ? lhs.internal_document_->get().key() ==
```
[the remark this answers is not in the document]
ehsannas: oops. I was checking the key twice (`lhs.internal_key_ == rhs.internal_key_` is already included).

PR #8653, opened 2021-09-13
Fall back to reCAPTCHA verification in phone auth if push notification is not received

rosalyntan:
For phone auth app verification, if the push notification is not received before the timeout, fall back to reCAPTCHA for app verification.

[Google-internal bug: b/198507858]

### DIFF
--- a/FirebaseAuth/Tests/Unit/FIRPhoneAuthProviderTests.m
+++ b/FirebaseAuth/Tests/Unit/FIRPhoneAuthProviderTests.m
@@ -324,13 +324,7 @@ static const NSTimeInterval kExpectationTimeout = 2;
         number was provided.
  */
 - (void)testVerifyEmptyPhoneNumber {
-  id mockBundle = OCMClassMock([NSBundle class]);
-  OCMStub(ClassMethod([mockBundle mainBundle])).andReturn(mockBundle);
-  OCMStub([mockBundle objectForInfoDictionaryKey:@"CFBundleURLTypes"]).andReturn(@[
-    @{@"CFBundleURLSchemes" : @[ kFakeReverseClientID ]}
-  ]);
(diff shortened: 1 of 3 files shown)


FirebaseAuth/Tests/Unit/FIRPhoneAuthProviderTests.m
```diff
@@ -1434,6 +1434,167 @@ - (void)testMissingAPNSToken {
...
+        XCTAssertFalse(callbackMatcher([components URL]));
+
+        // `completion` is at index 5
+        [invocation getArgument:&unretainedArgument atIndex:5];
+        FIRAuthURLPresentationCompletion completion = unretainedArgument;
+        dispatch_async(FIRAuthGlobalWorkQueue(), ^() {
+          completion([NSURL URLWithString:[kFakeEncodedFirebaseAppID
+                                              stringByAppendingString:
+                                                  kFakeRedirectURLStringWithReCAPTCHAToken]],
+                     nil);
```
maksymmalyhin, line 1566: Optional: It seems like this verification is repeated 4 times with little or none modifications. It may be useful to refactor it into a helper  method to reduce code duplication and improve readability. I don't feel to strongly about it and don't object to doing it in a separate PR or skipping it so far.

maksymmalyhin (approved):
LGTM 👍

yoyomyo (commented):
Thanks for sending this PR! I learned how delegates are passed. That said, there is nothing I could contribute as a reviewer 😂

FirebaseAuth/Tests/Unit/FIRPhoneAuthProviderTests.m
rosalyntan: Done, refactored duplicate code into helper methods.

PR #8654, opened 2021-09-14
Add workflow to update Firebase C++ SDK deps when iOS SDK release is published

jonsimantov:
Whenever a  release is published (except for CocoaPods-* releases), trigger the C++ SDK to do an iOS dependency update, pulling latest deps from public Cocoapods HEAD.

It requests a token that has access to firebase-cpp-sdk and uses that token to trigger the C++ update-dependencies workflow.

(Android equivalent is here: https://github.com/firebase/firebase-android-sdk/pull/2977)

### DIFF
--- a/.github/workflows/update-cpp-sdk-on-release.yml
+++ b/.github/workflows/update-cpp-sdk-on-release.yml
@@ -1,10 +1,14 @@
+# Whenever a new Firebase iOS SDK is released, this workflow triggers *another*
+# workflow on the Firebase C++ SDK, which will check for the iOS version update
+# and create a PR updating its iOS dependencies if the version number has
+# changed.
 name: update-cpp-sdk-on-release
 on:
   release:
     types: [ published ]
 
 jobs:
-  trigger-cpp-sdk-update:
+  trigger_cpp_sdk_update:
     # Only when a new release (not just the CocoaPods-* tag) is published.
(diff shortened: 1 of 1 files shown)


maksymmalyhin:
@jonsimantov Please note that we push Cocoapods tags as a part of the release staging about a week before pushing the podspecs to trunk. From the first glance the [release](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release) trigger will be more suitable for you, since we create a release on GitHub right after we pushed Cocoapods podspecs to trunk and pushed Swift Package Manager tag.

jonsimantov:
> @jonsimantov Please note that we push Cocoapods tags as a part of the release staging about a week before pushing the podspecs to trunk. From the first glance the [release](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release) trigger will be more suitable for you, since we create a release on [GitHub](https://github.com/firebase/firebase-ios-sdk/releases) right after we pushed Cocoapods podspecs to trunk and pushed Swift Package Manager tag.

Makes sense - I'll switch over to that.

jonsimantov:
Example of this working: https://github.com/firebase/firebase-ios-sdk/actions/runs/1235639116

maksymmalyhin (approved):
LGTM

Issue #4087, opened 2019-10-16
GoogleUtilities installed twice, impossible to archive app on Xcode 11.x


Tokta:
### [READ] Step 1: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ make sure the issue title starts with "FR:".
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Xcode version: 11.0 and 11.1
  * Firebase SDK version: 6.10.0 installed via Cocoapods
  * Firebase Component and version: 
Using Auth0 (1.19.0)
Using Crashlytics (3.14.0)
Using Fabric (1.10.2)
Using Firebase 6.10.0
Using FirebaseABTesting (3.1.2)
Using FirebaseAnalytics (6.1.3)
Using FirebaseAnalyticsInterop (1.4.0)
Using FirebaseCore (6.3.1)
Using FirebaseCoreDiagnostics (1.1.1)
Using FirebaseCoreDiagnosticsInterop (1.0.0)
Using FirebaseDynamicLinks (4.0.5)
Using FirebaseInAppMessaging (0.15.5)
Using FirebaseInstanceID (4.2.5)
Using FirebaseMessaging (4.1.6)
Using FirebasePerformance (3.1.5)
Using FirebaseRemoteConfig (4.4.2)
Using GTMSessionFetcher (1.2.2)
Using GoogleAppMeasurement (6.1.3)
Using GoogleDataTransport (3.0.1)
Using GoogleDataTransportCCTSupport (1.2.1)
Using GoogleToolboxForMac (2.2.1)
Using GoogleUtilities 6.3.1

#### Steps to reproduce:

Using Xcode 11.x the app compile with this warning on my target:

```
Multiple targets match implicit dependency for linker flags '-framework GoogleUtilities'. Consider adding an explicit dependency on the intended target to resolve this ambiguity. (in target 'Lepaya-Dev' from project 'Lepaya')
Target 'GoogleUtilities-00567490' (in project 'Pods')
[...]

PR #8665, opened 2021-09-16
Switch to Xcode 13 for all CI jobs.

yifanyang:
Upon submission of [#4087](https://github.com/actions/virtual-environments/pull/4087), Xcode 13 is officially supported on GitHub Actions.

[`zip.yml`](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/zip.yml) seems to be intentionally left out of the Xcode upgrade in #8543. I suppose it is still the case now.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,8 @@ source 'https://rubygems.org'
 # favorite editor. See https://github.com/firebase/firebase-ios-sdk/pull/8498
 # for examples.
 
+# TODO(yifanyang): Delete this line after all jobs are triggered and passed on GitHub Actions.
+
 # gem 'cocoapods', git: "https://github.com/CocoaPods/CocoaPods.git", ref: "9cebcde577f56aa26f27d8aa501b51fdd4d6abdb"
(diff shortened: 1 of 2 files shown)


ncooke3:
I would push a commit with a modified Gemfile to trigger all CI checks. Then, revert that commit before submitting. Paul recently added a write up in https://github.com/firebase/firebase-ios-sdk/pull/8569#issue-978516682

paulb777:
Looks like new Xcode 13 build errors for Crashlytics and StorageSwift. I'll take a look ...

paulb777:
zip.yml needs to build with the oldest supported version of Xcode since Xcode-generated binaries are not backwards compatible.

yifanyang:
> zip.yml needs to build with the oldest supported version of Xcode since Xcode-generated binaries are not backwards compatible.

Got it. I'm still in the progress of getting familiar with Apple toolchain. Thanks for the explanation!

paulb777:
Additional progress ongoing in #8676

Issue #8652, opened 2021-09-13
FCM flaky tests


maksymmalyhin:
FCM test `-[FIRMessagingTokenManagerTest testResetCredentialsWithNoCachedCheckin]` is regularly fails in nightly builds. Sometimes (but not always) the test passes after re-run. See failure examples: 
- https://github.com/firebase/firebase-ios-sdk/runs/3587412371?check_suite_focus=true#step:4:10085
- https://github.com/firebase/firebase-ios-sdk/runs/3578204793?check_suite_focus=true#step:4:677

chliangGoogle:
hmm I'm having trouble reproduce. I tried resetting the simulator and cleaning the project build and test results. And I tested SPM and nonSPM as well.
I wonder what nightly build setup is different than mine.

maksymmalyhin:
seems like it's flaky internally as well, see cl/397422878.

maksymmalyhin:
FCM tests are still flaky in nightly builds: #8663.

PR #8669, opened 2021-09-20
GHA setup for debugging flaky test

ncooke3:
Attempt to reproduce #8652

Note, I disabled as many workflows and tests as possible so that only FIRMessagingTokenManagerTest is getting tested.

cc: @maksymmalyhin @chliangGoogle 

Hopefully, this PR can be a good way to repro the flaky tests and verify a fix.

#no-changelog


scripts/build.sh
```diff
@@ -651,6 +651,9 @@ case "$product-$platform-$method" in
     RunXcodebuild \
       -scheme $product \
       "${xcb_flags[@]}" \
+      -test-iterations 10000 \
+      -retry-tests-on-failure \
+      -test-repetition-relaunch-enabled YES \
```
ncooke3, line 656: Tried to repro the flaky test with Xcode 13's new test iteration flags. In the context of this PR, `FIRMessagingTokenManagerTest` ran 10000 times and I wasn't able to see any crashes.

paulb777: Could the issue be related to running multiple incarnations of the tests in parallel?

ncooke3: Yeah, good point. I added back all of Messaging's tests and have only seen one flake so far. 4a11772 ran 10000 times of the Messaging spm tests

paulb777: Maybe try running all the platforms at the same time

paulb777: cc @chliangGoogle

ncooke3: > Maybe try running all the platforms at the same time

How would this done? @paulb777

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenManagerTest.m
```diff
@@ -70,7 +72,10 @@ - (void)setUp {
 }
 
 - (void)tearDown {
-  [_messaging.tokenManager stopAllTokenOperations];
+  [_messaging.tokenManager.tokenOperations addBarrierBlock:^{
```
maksymmalyhin, line 75: I guess, [waitUntilAllOperationsAreFinished](https://developer.apple.com/documentation/foundation/nsoperationqueue/1407971-waituntilalloperationsarefinishe) method may do the trick here.

ncooke3: I tried that one and that one waits indefinitely.

ncooke3: `stopAllTokenOperations` and `stopAllTokenOperations` both worked locally. `stopAllTokenOperations` didn't work as i expected it to on CI though.

maksymmalyhin: In the case `waitUntilAllOperationsAreFinished` creates a deadlock, you may use `addBarrierBlock` along with an expectation to postpone tearDown completion.

maksymmalyhin: If I'm not mistaken, `addBarrierBlock` returns immediately, so you will need an expectation here e.g.:
```
XCTestExpectation *exp = [...];
[_messaging.tokenManager.tokenOperations addBarrierBlock:^{
  [exp fulfill];
}];

[self waitForExpectations: @[exp]];
```

ncooke3:
The current state here is that I know why the test is flaking but am having issues finding a solid solution. Solving the problem at hand leads to another problem cropping up.

ncooke3:
Latest commit has a good solution to the problem.

Ran each test 1000 times with no failures. See GHA [messaging / spm (iOS) (pull_request)](https://github.com/firebase/firebase-ios-sdk/pull/8669/checks?check_run_id=3678724539)

ncooke3:
Will open a separate PR with just that fix and close this one.

PR #8672, opened 2021-09-20
[FDL] Refactoring and fixing url validation

eldhosembabu:
[FDL] Refactoring and fixing url validation

- Refactored the existing code to use Regex to validate FDL URL formats.

- Added checks to make sure the 'link' query param is having an http/https prefix input.
        == This is going to be breaking the existing SDK's unintended behavior to support links with non http/https content. The existing SDK was constructing the Dynamic link eventhough the backend response to resolve the link is not successfull.

- Added support for URL(s) with Query string starting with '?' alone. Previously it was supporting '/?' formats only. This fixes https://github.com/firebase/firebase-ios-sdk/issues/7087

- Added more test cases to be in sync with FDL backend validation.

- Removed invalid test inputs and corrected few test cases to make the input in proper format.

### DIFF
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,5 +1,5 @@
 # v8.8.0
-- [fixed] Firebase dynamic links with custom domain will only work if the custom domai has a trailing '/'. (#7087)
+- [fixed] Firebase dynamic links with custom domain will only work if the custom domain has a trailing '/'. (#7087)
 
 # v8.7.0
 - [added] Refactoring and adding helper class. (#8432)
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v8.7.1
+# v8.8.0
 - [fixed] Firebase dynamic links with custom domain will only work if the custom domai has a trailing '/'. (#7087)
 
 # v8.7.0
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v8.7.1
+- [fixed] Firebase dynamic links with custom domain will only work if the custom domai has a trailing '/'. (#7087)
+
 # v8.7.0
 - [added] Refactoring and adding helper class. (#8432)
 
--- a/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
+++ b/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
@@ -214,6 +214,10 @@ BOOL FIRDLIsURLForAllowedCustomDomain(NSURL *_Nullable URL) {
                          options:NSCaseInsensitiveSearch | NSAnchoredSearch]
                .location) == 0) {
         NSString *urlWithoutDomainURIPrefix = [urlStr substringFromIndex:domainURIPrefixStr.length];
+
+        //For a valid custom domain DL Suffix:
+        //1. At least one path exists OR
+        //2. Should have a link query param with an http/https link
         BOOL matchesRegularExpression =
             ([urlWithoutDomainURIPrefix
                  rangeOfString:@"((\\/[A-Za-z0-9]+)|((\\?|\\/\\?)link=https?.*))"
@@ -230,7 +234,7 @@ BOOL FIRDLIsURLForAllowedCustomDomain(NSURL *_Nullable URL) {
   return customDomainMatchFound;
 }
 
-/*We are validating following domains in proper format.
+/* We are validating following domains in proper format.
  *.page.link
  *.app.goo.gl
  *.page.link/i/
--- a/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
+++ b/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
@@ -204,30 +204,24 @@ NSString *FIRDLDeviceTimezone() {
 BOOL FIRDLIsURLForAllowedCustomDomain(NSURL *_Nullable URL) {
   BOOL customDomainMatchFound = false;
   for (NSURL *allowedCustomDomain in FIRDLCustomDomains) {
-    // All custom domain host names should match at a minimum.
+    // Atleast one custom domain host name should match at a minimum.
     if ([allowedCustomDomain.host isEqualToString:URL.host]) {
       NSString *urlStr = URL.absoluteString;
       NSString *domainURIPrefixStr = allowedCustomDomain.absoluteString;
 
       // Next, do a string compare to check if entire domainURIPrefix matches as well.
-      if (([URL.absoluteString rangeOfString:allowedCustomDomain.absoluteString
+      if (([urlStr rangeOfString:domainURIPrefixStr
                                      options:NSCaseInsensitiveSearch | NSAnchoredSearch]
                .location) == 0) {
-        // The (short) URL needs to be longer than the domainURIPrefix, it's first character after
-        // the domainURIPrefix needs to be '/' and should be followed by at-least one more
-        // character.
-        if (urlStr.length > domainURIPrefixStr.length + 1 &&
-            ([urlStr characterAtIndex:domainURIPrefixStr.length] == '/')) {
-          // Check if there are any more '/' after the first '/'trailing the
-          // domainURIPrefix. This does not apply to unique match links copied from the clipboard.
-          // The clipboard links will have '?link=' after the domainURIPrefix.
-          NSString *urlWithoutDomainURIPrefix =
-              [urlStr substringFromIndex:domainURIPrefixStr.length + 1];
-          if ([urlWithoutDomainURIPrefix rangeOfString:@"/"].location == NSNotFound ||
-              [urlWithoutDomainURIPrefix rangeOfString:@"?link="].location != NSNotFound) {
-            customDomainMatchFound = true;
-            break;
-          }
+
+        NSString *urlWithoutDomainURIPrefix =
+            [urlStr substringFromIndex:domainURIPrefixStr.length];
+        BOOL matchesRegularExpression =
+            ([urlWithoutDomainURIPrefix rangeOfString:@"((\\/[A-Za-z0-9]+)|((\\?|\\/\\?)link=https?.*))" options:NSRegularExpressionSearch].location != NSNotFound);
+
+        if(matchesRegularExpression){
+          customDomainMatchFound = true;
+          break;
         }
       }
     }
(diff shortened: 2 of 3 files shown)


FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v8.7.1
```
paulb777, line 1: Should be 8.8.0 or 8.9.0. See the Milestones tab for upcoming releases

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -204,64 +204,92 @@ BOOL FIRDLOSVersionSupported(NSString *_Nullable systemVersion, NSString *minSup
 BOOL FIRDLIsURLForAllowedCustomDomain(NSURL *_Nullable URL) {
   BOOL customDomainMatchFound = false;
   for (NSURL *allowedCustomDomain in FIRDLCustomDomains) {
-    // All custom domain host names should match at a minimum.
+    // Atleast one custom domain host name should match at a minimum.
```
paulb777, line 207: nit: At least

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
```diff
@@ -1099,9 +1099,23 @@ - (void)testResolveLinkRespectsResponseErrorStatusCode {
   [self waitForExpectationsWithTimeout:kAsyncTestTimout handler:nil];
 }
 
-- (void)testMatchesShortLinkFormat {
-  NSArray<NSString *> *urlStrings =
-      @[ @"https://test.app.goo.gl/xyz", @"https://test.app.goo.gl/xyz?link=" ];
+- (void)testPassMatchesShortLinkFormatForDDLDomains {
+  NSArray<NSString *> *urlStrings = @[
+    @"https://someapp.app.goo.gl/somepath", @"https://someapp.app.goo.gl/link",
```
paulb777, line 1104: Nice add!

paulb777 (commented):
Also, since we don't yet have Xcode 13 in CI pending #8665, please confirm that the PR builds and tests with Xcode 13.

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
eldhosembabu: Thanks

FirebaseDynamicLinks/CHANGELOG.md

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: Done

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -204,64 +204,92 @@ BOOL FIRDLOSVersionSupported(NSString *_Nullable systemVersion, NSString *minSup
...
-              [urlStr substringFromIndex:domainURIPrefixStr.length + 1];
-          if ([urlWithoutDomainURIPrefix rangeOfString:@"/"].location == NSNotFound ||
-              [urlWithoutDomainURIPrefix rangeOfString:@"?link="].location != NSNotFound) {
-            customDomainMatchFound = true;
-            break;
-          }
+        NSString *urlWithoutDomainURIPrefix = [urlStr substringFromIndex:domainURIPrefixStr.length];
+        BOOL matchesRegularExpression =
+            ([urlWithoutDomainURIPrefix
+                 rangeOfString:@"((\\/[A-Za-z0-9]+)|((\\?|\\/\\?)link=https?.*))"
```
ryanwilson, line 219: Can you add a comment about what we're looking for here? Is this any different than the above code? If so, why?

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -204,64 +204,92 @@ BOOL FIRDLOSVersionSupported(NSString *_Nullable systemVersion, NSString *minSup
...
+          customDomainMatchFound = true;
+          break;
         }
       }
     }
   }
   return customDomainMatchFound;
 }
 
+/*We are validating following domains in proper format.
```
ryanwilson, line 233: Nit: space after the `*`.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -204,64 +204,92 @@ BOOL FIRDLOSVersionSupported(NSString *_Nullable systemVersion, NSString *minSup
...
+ *.page.link
+ *.app.goo.gl
+ *.page.link/i/
+ *.app.goo.gl/i/
+ */
```
ryanwilson, line 238: Nit: I think there's an extra space at the beginning of this line

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -204,64 +204,92 @@ BOOL FIRDLOSVersionSupported(NSString *_Nullable systemVersion, NSString *minSup
...
+
+    if (!matchesRegularExpression) {
+      // Matches the *.page.link/i/ and *.app.goo.gl/i/ domains.
+      matchesRegularExpression =
+          ([urlStr rangeOfString:
+                       @"^https?:\\/\\/[a-zA-Z0-9]+((\\.app\\.goo\\.gl)|(\\.page\\.link))\\/i\\/.*$"
+                         options:NSRegularExpressionSearch]
+               .location != NSNotFound);
+    }
+  }
```
ryanwilson, line 259: Is there a reason why we're not checking for `.page.link` and `.page.link/i/` together and `.app.goo.gl` with `.app.goo.gl/i/`? Seems like the regex would be simpler.

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
```diff
@@ -491,7 +491,7 @@ - (void)testUniversalLink_NoDeepLink {
 //  https://a.firebase.com/mypath
 - (void)testDynamicLinkFromUniversalLinkURLWithCustomDomainLink {
   self.service = [[FIRDynamicLinks alloc] init];
-  NSString *durableDeepLinkString = @"https://a.firebase.com/mypath/?link=abcd";
+  NSString *durableDeepLinkString = @"https://a.firebase.com/mypath/?link=http://abcd";
```
ryanwilson, line 494: Does this mean there's a change in functionality?

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: I think its expected and we ran the style script as well. LMK if we need to make any changes here.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: The expected suffix for .page.link and .page.link/i are different.

https://$appdomain.page.link/i/$anything is valid but https://$appdomain.page.link/ has many restrictions to be a valid link. For example: 

https://$appdomain.page.link/p1 is valid but https://$appdomain.page.link/p1/p2 is invalid 

https://$appdomain.page.link?link=https://somedomain is valid but https://$appdomain.page.link?url=https://somedomain is invalid 

Since my expertise on regex is not that great, I find it difficult to mix these two checks together.

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
eldhosembabu: The Client SDK will make sure the link query parameter is having and http/https link to consider it as a valid link which was not done before. So we won't make a backend call to validate the URL. Previously, for any value in link, we were making backend calls considering it as a valid DL.  This is in sync with Server side validations.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: The changes added here are: 
1. Moved the logic to Regex.
2. Old code wont consider any number of paths after the whitelisted custom domain as valid.
3. Added check to make sure the link query param starts with http/https
4. Previous code was expecting the presence of either one path after the custom domain prefix or a link query param which is not the case for a valid custom domain.

eldhosembabu: Usually this code execution flow will happens only once after the app is started ie. when the app is opened through universal link, so I believe its better to not cache it since it would stay in the memory even though we dont need it further. LMK if you think otherwise.

maksymmalyhin: SGTM

eldhosembabu:
> #8665



> Also, since we don't yet have Xcode 13 in CI pending #8665, please confirm that the PR builds and tests with Xcode 13.

I was doing my testing in XCode 13 Beta 5 only

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
ryanwilson: > 3. Added check to make sure the link query param starts with http/https

Re: 3, would a link query param work without the http/https before and now it won't? Or would it fail in a different way (silently or not) previously and now we're failing in a more sensible way? Just want to make sure we're not breaking any existing setups.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
ryanwilson: Ahh, that makes sense. Adding a comment above the regex would be helpful for those that aren't familiar with the rules around the `/i/` if you don't mind

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
ryanwilson: So to confirm, previously the SDK would send off a request to the backend but the backend would reject it, now we're doing that rejection in the SDK?

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: No...previously, the link param without http/https will bypass client validation and will gets rejected from the backend. Now with this change, we will filter it from the client itself.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: Sure

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
eldhosembabu: Yes

ryanwilson (approved):

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: As discussed over GVC with @ryanwilson , Upon checking it is found that even though the server response for this link is un successful, the client SDK was locally creating a dynamic link object and passing to the SDK users. This is going to be breaking with this change as the old behavior is un intended.

eldhosembabu:
Checked on xCode 13 and it is working.

Issue #8684, opened 2021-09-22
Firebase Messaging crash during app loading if Messaging was a added via SPM


DenisVerkhovod:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.5
  * Firebase SDK version: 8.7.0
  * Installation method: Swift Package Manager
  * Firebase Component: Messaging

#### Steps to reproduce:

During app loading I get crash cause `createTableWithName` method. You can see stack on attached screenshot.

<img width="1438" alt="Screenshot 2021-09-21 at 18 13 39" src="https://user-images.githubusercontent.com/39265089/134311853-a6430f53-0adc-4130-b780-f05fe0c30f81.png">

paulb777:
It looks like the sqlite3 database creation is failing. Does the app or any of its other dependencies do anything unusual that could change the state of the system sqlite3 library that FirebaseMessaging is depending on?

DenisVerkhovod:
I think no, moreover everything is good with Messaging when it was installed via CocoaPod. This crash is specific to SPM

paulb777:
The 8.9.0 release will improve the logging when this crash occurs, but the crash occurs after a table create failure, so something has already gone wrong.

chliangGoogle:
@DenisVerkhovod Can you try deleting the app and install again?

PR #8699, opened 2021-09-23
Improve error logging when table creation fails

morganchen12:
SQLite's [documentation](https://www.sqlite.org/capi3ref.html#sqlite3_exec) implies that `sqlite3_exec` is not guaranteed to return a non-null error message for every failure.

Improves error reporting for #8684, though it does not fix the crash.


chliangGoogle (approved):
Sorry I just saw this.

PR #8714, opened 2021-09-27
Enable binary data upload in post submit

granluo:
Found binary size reports have more question marks than expected. 
This seemed to be caused by binary size data which were not uploaded properly, since `needs:check` kept the binary data job from being run in post submits, which uploaded binary data to Metrics Service. The lack of data in base commits eventually lead to questions marks in reports.
The `check` job is helping check changing files in a PR and so to determine jobs to be run and this is not applicable in post submit and `always()` will run the binary job regardless of `needs:check`.

### DIFF
--- a/.github/workflows/health_metrics.yml
+++ b/.github/workflows/health_metrics.yml
@@ -47,7 +47,8 @@ jobs:
   binary_size_metrics:
     needs: check
     # Prevent the job from being triggered in fork.
-    if: github.event.pull_request.head.repo.full_name == github.repository && (github.event.action != 'closed' || github.event.pull_request.merged)
+    # always() will trigger this job when `needs` are skipped in a `merge` pull_request event.
+    if: always() && github.event.pull_request.head.repo.full_name == github.repository && (github.event.action != 'closed' || github.event.pull_request.merged)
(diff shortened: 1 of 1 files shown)


.github/workflows/health_metrics.yml
```diff
@@ -47,7 +47,8 @@ jobs:
   binary_size_metrics:
     needs: check
```
maksymmalyhin, line 48: I wonder if we can just remove `needs: check` for `binary_size_metrics` job?

maksymmalyhin (approved):
LGTM, one question.

.github/workflows/health_metrics.yml
granluo: No, we need `needs: check` in presubmits. In presubmits, `check` will determine which SDK test to run by updating [these variable](https://github.com/firebase/firebase-ios-sdk/blob/3438778ce70c61f521be29d54e0ffe480d1e6fe2/.github/workflows/health_metrics.yml#L72-L82). 
In postsubmit, we do not need `check` since we will upload all SDK binary data regardless of changing SDKs in PRs.

PR #8717, opened 2021-09-28
build(release, zip): pin zip build to macOS-10.15

mikehardy:
PR #8543 specifically leaves the zip build on macOS-latest (vs macOS-11) so that the build has access to the oldest supported Xcode (12.2 as of now) to avoid bitcode-incompatibility

However, macOS-latest is currently transitioning to macOS-11 which will lead to build failures as macOS-11 does not have access to Xcode 12.2

migration plan:
https://github.com/actions/virtual-environments/issues/4060

macOS-11 without Xcode 12.2:
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode

macOS-10.15 still has Xcode 12.2:
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode

This pins the zip build to macOS-10.15 in order to maintain Xcode 12.2 access

### Testing

I rely on @paulb777's keen eyes here

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -24,7 +24,7 @@ jobs:
   package-release:
     # Don't run on private repo.
     if: (github.repository == 'Firebase/firebase-ios-sdk' && github.event_name == 'schedule') || github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
-    runs-on: macOS-latest
+    runs-on: macOS-10.15
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


mikehardy:
Saw this while clicking through the 8.8.0 release notes and the related github links and their transitives, easier to PR it then write a note, given GitHub's web UI working pretty well...cheers

ncooke3 (approved):
LGTM. Good catch, thanks @mikehardy!

PR #8722, opened 2021-09-29
Update swiftpm docs for Xcode 13

ncooke3:
Preview: https://github.com/firebase/firebase-ios-sdk/blob/befebff/SwiftPackageManager.md

#no-changelog


ryanwilson:
One thing in reviewing that I realized we should add, can you please change from `Requires Xcode 12.5` to `Requires Xcode 12.5 or above`?

ryanwilson:
Another thing, can you please add ABTesting to the watchOS product availability?

ryanwilson (approved):
Looks great, thanks! If you don't mind those two small tweaks as well that would be great.

ncooke3:
Small improvements

ryanwilson (approved):
Thanks!

PR #8764, opened 2021-10-07
Update versions for Release 8.9.0

paulb777:
#no-changelog


paulb777:
Actually I don't believe those podspec's shouldn't even be in the repo since we decided to only support Combine via Swift Package Manager. I'll send a PR to delete them tomorrow.

PR #8766, opened 2021-10-07
Don't invalidate token if we were able to receive data

schmidt-sebastian:
This PR changes our token invalidation to optimize for the general case when we hit UNAUTHORIZED. If we fetch a token, successfully talk to Watch and then receive UNAUTHORIZED we will not force refresh tokens. Instead, we rely on the AuthTokenProvider to always give us fresh tokens. Only if we UNAUTHORIZED before we receive any message to forcefully invalidate the token - as this indicates that the AuthTokenProvider did not refresh the token.

Will verify this manually and report back.

### DIFF
--- a/Firestore/core/test/unit/remote/stream_test.cc
+++ b/Firestore/core/test/unit/remote/stream_test.cc
@@ -511,7 +511,7 @@ TEST_F(StreamTest, RefreshesTokenUponExpiration) {
   StartStream();
   ForceFinish({{Type::Read, CompletionResult::Error},
                {Type::Finish, grpc::Status{grpc::UNAUTHENTICATED, ""}}});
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
cc @ehsannas The test does not work yet but we should do the same thing on Android. Basically, we invalidate both tokens in the rare case that we get an "invalid token" error right away.

schmidt-sebastian:
Verified the assumption that this is based on manually. Unfortunately, the backend does not immediately reject the Stream when invalid credentials are passed. Rather, it sends a couple messages before the stream is closed. Hence, I updated this PR to use the same timeout that we use in our OnlineStateTracking. If the stream stays active for more than 10 seconds, we consider the credentials to be valid and do not force refresh.

PR #8774, opened 2021-10-08
Fix Remote Config Integration CI

paulb777:
Similar to #7874, the RC integration tests have been silently failing ever since Firebase 7.0.0 after the GoogleServiceInfo.plist does not get found because FirebaseCore is linked dynamically as a separate bundle instead of statically into the app.

This PR updates the resources to be added to the Test Container App instead of the test bundle. 

It also fixes a setup ordering bug that caused a failure when running locally in the fakeconsole tests.

The IntegrationTest's are commented out for now, pending finding the right Firebase project.


paulb777:
Going to merge here and follow up in another PR on the disabled integration tests.

PR #6637, opened 2020-10-01
Remove deprecated RC APIs

karenyz:
Removing following methods:
- `isDeveloperModeEnabled`
- `initWithDeveloperModeEnabled:developerModeEnabled`
- `activateWithCompletionHandler:completionHandler`
- `activateFetched`
- `configValueForKey:namespace`
- `configValueForKey:namespace:source`
- `allKeysFromSource:namespace`
- `keysWithPrefix:namespace`
- `setDefaults:namespace`
- `setDefaultsFromPlistFileName:namespace`
- `defaultValueForKey:namespace`

Fixes #6546

PR #8788, opened 2021-10-11
Restore 6.x fetchAndActivate status behavior

paulb777:
This PR restores the 6.x fetchAndActivate status behavior returning `successUsingPreFetchedData`. The behavior was changed in the 7.0.0 release via #6637 and not noticed because the tests also silently broke in 7.0.0. Details about the test regression in #8774 and I'll continue with fixing the CI for the Swift API tests in a subsequent PR.

This PR fixes the test at https://github.com/firebase/firebase-ios-sdk/blob/v8.8.1/FirebaseRemoteConfig/Tests/SwiftAPI/APITests.swift#L157 from returning `successFetchedFromRemote` to returning `successUsingPreFetchedData`


paulb777:
After talking with @ryanwilson, we agreed that even this wasn't an intentional change for Firebase 7.x, it would be a breaking change to reset it back now.  I'll leave it to the RC team to decide if this is something to address for Firebase 9.  In the meantime, I'll close this PR and update the test with a comment in #8792

PR #8802, opened 2021-10-13
Disable resource management for binary pods

### LINKED ISSUE #8796: Nightly Testing Report
# Nightly Testing Report
This issue([log](https://github.com/firebase/firebase-ios-sdk/actions/runs/1337569085)) is generated at 10/13/2021 06:10 AM, fetching workflow runs triggered in the last 24 hrs.
| Workflow |10/12|
| -------- | -------- |
| [zip](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/zip.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/1336521123) |
| [database](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/database.yml) | [failure](https://github.com/firebase/firebase-ios-sdk/actions/runs/1336948587) |

paulb777:
Fix #8796 

Facebook switched to binary xcframework pods in their 12.0.0 release and that exposed a crash in code that probably doesn't need to run that was reorganizing resources in binary frameworks. xcframework support was never added and I'm not sure we even need the original code.  

Disabling for now.

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/ZipBuilder.swift
@@ -808,13 +808,15 @@ struct ZipBuilder {
     }
     var frameworks: [URL] = []
 
+    // TODO: packageAllResources is disabled for binary frameworks since it's not needed for Firebase
+    // and it does not yet support xcframeworks.
     // Package all resources into the frameworks since that's how Carthage needs it packaged.
-    do {
-      // TODO: Figure out if we need to exclude bundles here or not.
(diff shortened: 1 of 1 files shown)


ryanwilson:
> Thanks @ryanwilson. I'm not convinced there's an issue beyond some dead code. WDYT?

SGTM, wasn't sure if we wanted to fix this eventually and would get caught in a situation where we add resources somewhere and they're missing from the distribution since it silently fails. I suspect that's going to be quite rare so I'm fine leaving it commented out without an issue.

Issue #8790, opened 2021-10-12
FNextPushId 'successor' crash


mortenbekditlevsen:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13
  * Firebase SDK version: master
  * Installation method: `Swift Package Manager`
  * Firebase Component: Database

### [REQUIRED] Step 2: Describe the problem

The methods `successor` and `predecessor` erronously deal with PUSH_CHARS even though it's entirely valid to use unicode keys that are not within the PUSH_CHARS set.

Around line 98 in FNextPushId.m we have:
```
    NSInteger sourceIndex = [PUSH_CHARS rangeOfString:source].location;
    NSString *sourcePlusOne = [NSString
        stringWithFormat:@"%C", [PUSH_CHARS characterAtIndex:sourceIndex + 1]];
```
The issue is that if `source` is for instance a space, then sourceIndex will be max int and the sourcePlusOne  calculation causes a crash (since max int + 1 gives min int in objective-c, and min int is out of bounds).

Thread 1: "-[__NSCFConstantString characterAtIndex:]: Range or index out of bounds"

#### Steps to reproduce:
Easily reproducible with the one-liner below:

#### Relevant Code:

```
        let ref = Database.database().reference().queryOrderedByKey().queryStarting(afterValue: " ".padding(toLength: 786, withPad: "z", startingAt: 0))
```

#### Suggestion for a solution

I guess that 'successor' and 'predecessor' ought to deal with unicode character points instead of the PUSH_CHARs used to auto-generate keys.
[...]

PR #8817, opened 2021-10-16
Attempt at fixing issue with getting next and previous push id

mortenbekditlevsen:
This is an attempt at fixing #8790. It's not very polished yet, and the test is currently too slow (loops over all single unicode character strings).

### DIFF
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.h
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.h
@@ -20,8 +20,8 @@
 
 + (NSString *)get:(NSTimeInterval)now;
 
-+ (NSString *)successor:(NSString *)key;
++ (NSString *)from:(NSString *)fn successor:(NSString *)key;
 
-+ (NSString *)predecessor:(NSString *)key;
++ (NSString *)from:(NSString *)fn predecessor:(NSString *)key;
 
 @end
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.m
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.m
@@ -106,7 +106,8 @@ static NSInteger const MAX_KEY_LEN = 786;
         break;
 
     case 0x2E:
-        plusOne = 0x2F;
+    case 0x2F:
+        plusOne = 0x30;
         break;
 
     case 0x5B:
@@ -115,10 +116,6 @@ static NSInteger const MAX_KEY_LEN = 786;
 
     case 0x5D:
         plusOne = 0x5E;
-        break;
-
-    case 0x2F:
-        plusOne = 0x30;
         break;
 
     case 0x7F:
@@ -219,6 +216,7 @@ static NSInteger const MAX_KEY_LEN = 786;
         break;
 
     case 0x2E:
+    case 0x2F:
         minusOne = 0x2D;
         break;
 
@@ -228,10 +226,6 @@ static NSInteger const MAX_KEY_LEN = 786;
 
     case 0x5D:
         minusOne = 0x5C;
-        break;
-
-    case 0x2F:
-        minusOne = 0x2E;
         break;
 
     case 0x7F:
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.m
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.m
@@ -229,7 +229,7 @@ static unichar const HIGH_SURROGATE_PAIR_END = 0xDBFF;
         minusOne = 0x5A;
         break;
 
-    case 0x5B: // ']'
+    case 0x5D: // ']'
         minusOne = 0x5C;
         break;
 
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.m
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.m
@@ -105,25 +105,25 @@ static unichar const HIGH_SURROGATE_PAIR_END = 0xDBFF;
     BOOL removePreviousCharacter = NO;
     BOOL replaceWithLowestSurrogatePair = NO;
     switch (plusOne) {
-    case 0x23:
-    case 0x24:
+    case 0x23: // '#'
+    case 0x24: // '$'
         plusOne = 0x25;
         break;
 
-    case 0x2E:
-    case 0x2F:
+    case 0x2E: // '.'
+    case 0x2F: // '/'
         plusOne = 0x30;
         break;
 
-    case 0x5B:
+    case 0x5B: // '['
         plusOne = 0x5C;
         break;
 
-    case 0x5D:
+    case 0x5D: // ']'
         plusOne = 0x5E;
         break;
 
-    case 0x7F:
+    case 0x7F: // control character: del
         plusOne = 0x80;
         break;
 
@@ -215,25 +215,25 @@ static unichar const HIGH_SURROGATE_PAIR_END = 0xDBFF;
     BOOL replaceWithHighestSurrogatePair = NO;
     switch (minusOne) {
     // NOTE: We already handled the case of min char (0x20)
-    case 0x23:
-    case 0x24:
+    case 0x23: // '#'
+    case 0x24: // '$'
         minusOne = 0x22;
         break;
 
-    case 0x2E:
-    case 0x2F:
+    case 0x2E: // '.'
+    case 0x2F: // '/'
         minusOne = 0x2D;
         break;
 
-    case 0x5B:
+    case 0x5B: // '['
         minusOne = 0x5A;
         break;
 
-    case 0x5D:
+    case 0x5B: // ']'
         minusOne = 0x5C;
         break;
 
-    case 0x7F:
+    case 0x7F: // control character: del
         minusOne = 0x7E;
         break;
 
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.m
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.m
@@ -16,6 +16,7 @@
 
 #import "FirebaseDatabase/Sources/Utilities/FNextPushId.h"
 #import "FirebaseDatabase/Sources/Utilities/FUtilities.h"
+#import "FirebaseDatabase/Sources/Utilities/FValidation.h"
 
 static NSString *const PUSH_CHARS =
     @"-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz";
@@ -65,8 +66,8 @@ static NSInteger const MAX_KEY_LEN = 786;
     return [NSString stringWithString:id];
 }
 
-+ (NSString *)successor:(NSString *_Nonnull)key {
-    // TODO: Wouldn't it be good to perform actual key validation on the input?
++ (NSString *)from:(NSString *)fn successor:(NSString *_Nonnull)key {
+    [FValidation validateFrom:fn validKey:key];
     NSInteger keyAsInt;
     if ([FUtilities tryParseString:key asInt:&keyAsInt]) {
         if (keyAsInt == [FUtilities int32max]) {
@@ -177,9 +178,8 @@ static NSInteger const MAX_KEY_LEN = 786;
     return [next substringWithRange:NSMakeRange(0, length)];
 }
 
-// `key` is assumed to be non-empty.
-+ (NSString *)predecessor:(NSString *_Nonnull)key {
-    // TODO: Wouldn't it be good to perform actual key validation on the input?
++ (NSString *)from:(NSString *)fn predecessor:(NSString *_Nonnull)key {
+    [FValidation validateFrom:fn validKey:key];
     NSInteger keyAsInt;
     if ([FUtilities tryParseString:key asInt:&keyAsInt]) {
         if (keyAsInt == [FUtilities int32min]) {
--- a/FirebaseDatabase/Sources/Utilities/FNextPushId.m
+++ b/FirebaseDatabase/Sources/Utilities/FNextPushId.m
@@ -142,8 +142,8 @@ static unichar const HIGH_SURROGATE_PAIR_END = 0xDBFF;
             // Error, encountered low surrogate without matching high surrogate
         } else {
             unichar high = [next characterAtIndex:i - 1];
-            if (high ==
-                HIGH_SURROGATE_PAIR_END) { /* highest value for the high part of the pair */
+            if (high == HIGH_SURROGATE_PAIR_END) { /* highest value for the high
+                                                      part of the pair */
                 // Replace pair with 0xE000 (the value of plusOne)
                 removePreviousCharacter = YES;
             } else {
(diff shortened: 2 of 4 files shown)


mortenbekditlevsen:
@IanWyszynski would you do a formal review of the PR, or would it be better if someone not involved closely with the details did that?

paulb777 (commented):
Please update https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDatabase/CHANGELOG.md with a note for 8.10.0 with a summary of the fix.

mortenbekditlevsen:
Another small thought about this API:
Since it requires so much pretty specific unicode/utf16 'knowledge' to be built in to the clients, it might actually be prettier if the API was implemented server-side, so all clients don't have to re-implement this rather complex logic.

I am guessing that the API was added directly to the client since it does provide good value, and it seemed like an easy thing to add to the clients without involving the server. But as it turns out that it's not so easy, perhaps it would make sense for the server to provide the API out-of-the-box.

Just my two cents. :-)

mortenbekditlevsen:
As an addition to my comment above (that it might be better to implement the start after / start before queries server-side), I realized that the implementation in this PR might actually _still_ not be accurate.

If two keys are integer based, then the textually _longest_ key comes after the _shorter_.
This means that the next key after 12345 is not 12346, but 012345.
Worse, the next key _before_ 12345 is 00000...000012344 (prefix zero-padding to max key length).
I haven't tried it out, but I assume that the similar situation holds for negative numbers, but here we are on the negative axis, so it may be non-obvious that -01 comes _after_ -1 and 01 comes _after_ 1.

I haven't tested how -00000 sorts compared to 000, but I would assume from the algorithm that -00000 comes after 000... Hmmm.

This is a _really_ hard thing to do right.

@IanWyszynski

mortenbekditlevsen:
Hmm, I just got the thought that there is actually an undefined sort order in the Obj-C and Javascript implementations of key sorting that I have seen.

'-0' and '00' have the same length and the same numerical value, but they are not the same. What is their sort order?

The RTDB console appears to be hiding my 00 key since there was already a -0 value... This looks like a bug, but I don't know how pervasive it is...

Hmm, I guess internally 00 and -0 resolves to the same key internally... That's a bit confusing.

Playing around in the console I get the following:

I add a key "-0" with the value "minus zero"
Then I add a key "00" with the value "double zero"
The console does not update with my "00" key/value, but still shows "-0": "minus zero"
If I refresh the UI then I see the "00": "double zero" value and not the "-0" key.
If I try accessing the "-0" key directly I see the "double zero" value.

My conclusion is that "-0" and "00" resolve to the same key (as does "-00" and "000", etc.)

mortenbekditlevsen:
Now you see it, now you don't:

```
{
  "rules": {
    "00": {
    ".read": true,
    ".write": true
    },
    "-0": {
    ".read": false,
    ".write": false      
    }
  }
}
```

mortenbekditlevsen:
Final comment about this:
I think that I may have destroyed my database. I can't access it through the console anymore.

Skærmbillede 2021-11-05 kl  10 41 41

That's a bit unfortunate - luckily it is just a toy db...

mortenbekditlevsen:
I think I found a difference in the key sort between ios and js:

In js there can be any number of 0's prefixed to a 32-bit int:
```
export const INTEGER_REGEXP_ = new RegExp('^-?(0*)\\d{1,10}$');

export const tryParseInt = function (str: string): number | null {
  if (INTEGER_REGEXP_.test(str)) {
    const intVal = Number(str);
    if (intVal >= INTEGER_32_MIN && intVal <= INTEGER_32_MAX) {
      return intVal;
    }
  }
  return null;
};
```

On iOS the length of the string is checked before conversion is attempted, so any string longer than 11 characters is never considered to be an integer value.
```
static inline BOOL tryParseStringToInt(__unsafe_unretained NSString *str,
                                       NSInteger *integer) {
    // First do some cheap checks (NOTE: The below checks are significantly
    // faster than an equivalent regex :-( ).
    NSUInteger length = str.length;
    if (length > 11 || length == 0) {
        return NO;
    }
    ...
```

I haven't yet tested which of the two implementations the server appears to use.

mortenbekditlevsen:
Now I've tested, and it appears that the javascript implementation matches the server implementation in this regard.

paulb777:
Thanks @mortenbekditlevsen 

@IanWyszynski Would you follow up on the open questions and the review?

mortenbekditlevsen:
@IanWyszynski @paulb777 
Let me know if there's anything I can do to help here. 
Should I continue to add fixes on top of the existing fixes? Or would you find it more appropriate to try and implement these APIs (the query after and before, I mean) in the backend somehow?
And what about the key sorting issue I pointed out above? Should I file a separate issue for that?

IanWyszynski:
> 



> @IanWyszynski @paulb777 Let me know if there's anything I can do to help here. Should I continue to add fixes on top of the existing fixes? Or would you find it more appropriate to try and implement these APIs (the query after and before, I mean) in the backend somehow? And what about the key sorting issue I pointed out above? Should I file a separate issue for that?

Given the complexity of getting this right, I'm leaning towards moving startAfter/endBefore to the backend so we at least don't have to maintain a separate implementation in each SDK. @jsdt wdyt?

paulb777 (approved):
In email exchange with @IanWyszynski and @schmidt-sebastian, we agreed that even if a better long-term solution is a back end fix, this is a good interim solution.

paulb777:
@mortenbekditlevsen Anything outstanding to resolve before merging?  I just tried the new test and the speed seems fine.

mortenbekditlevsen:
It's still missing a fix for next/prev of an integer key prefixed by a number of zeroes.

mortenbekditlevsen:
I'll look into fixing that part.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
```diff
@@ -93,19 +100,91 @@ + (NSString *)successor:(NSString *_Nonnull)key {
...
-        stringWithFormat:@"%C", [PUSH_CHARS characterAtIndex:sourceIndex + 1]];
+    unichar character = [next characterAtIndex:i];
+    unichar plusOne = character + 1;
+    BOOL removePreviousCharacter = NO;
+    BOOL replaceWithLowestSurrogatePair = NO;
+    switch (plusOne) {
+    case 0x23: // '#'
+    case 0x24: // '$'
+        plusOne = 0x25;
+        break;
```
schmidt-sebastian, line 111: Can we do:

```suggestion
    case '#':
    case '$':
        plusOne ='.';
        break;
```

FirebaseDatabase/Sources/Utilities/FNextPushId.m
```diff
@@ -93,19 +100,91 @@ + (NSString *)successor:(NSString *_Nonnull)key {
...
+        replaceLocation--;
+        replaceLength++;
+    }
+
+    [next replaceCharactersInRange:NSMakeRange(replaceLocation, replaceLength)
                         withString:sourcePlusOne];
-    return [next substringWithRange:NSMakeRange(0, i + 1)];
+    NSInteger length = i + 1;
+    if (removePreviousCharacter) {
+        length--;
```
schmidt-sebastian, line 179: Google-style usually suggest to use prefix operators where possible: `--length`

FirebaseDatabase/Sources/Utilities/FNextPushId.m
```diff
@@ -93,19 +100,91 @@ + (NSString *)successor:(NSString *_Nonnull)key {
...
+        plusOne = 0x5E;
+        break;
 
-    [next replaceCharactersInRange:NSMakeRange(i, i + 1)
+    case 0x7F: // control character: del
+        plusOne = 0x80;
+        break;
+
+    case HIGH_SURROGATE_PAIR_START: // 0xD800
+        // We added one to 0xD7FF and entered surrogate pair zone
```
schmidt-sebastian, line 131: To me, this logic feels a bit hard to parse since you are operating on the already mutated value. Do you think it is possible to use the base value and then increment it based on these special cases?

FirebaseDatabase/Tests/Unit/FNextPushIdTest.m
```diff
@@ -18,71 +18,161 @@
...
 
-  actual = [FNextPushId predecessor:[NSString stringWithFormat:@"abc%@", MIN_PUSH_CHAR]];
+  actual = [FNextPushId from:@"test"
+                 predecessor:[NSString stringWithFormat:@"abc%@", MIN_PUSH_CHAR]];
   expected = @"abc";
   XCTAssertEqualObjects(expected, actual, @"predecessor(abc + MIN_PUSH_CHAR) == abc");
 }
 
+- (void)testPredecessorWild {
+  NSString *actual = [FNextPushId from:@"test" predecessor:@"\uE000"];
```
schmidt-sebastian, line 97: s/Wild/Unicode?

schmidt-sebastian (commented):
Thanks for this. Please let us know if you have time to address the feedback.

FirebaseDatabase/Sources/Utilities/FNextPushId.h
```diff
@@ -20,8 +20,8 @@
 
 + (NSString *)get:(NSTimeInterval)now;
 
-+ (NSString *)successor:(NSString *)key;
++ (NSString *)from:(NSString *)fn successor:(NSString *)key;
```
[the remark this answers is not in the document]
mortenbekditlevsen: Right, I was trying to follow the precedent.

I'll gladly change it, but I'm uncertain about wether you're actually asking me to change it based on your preference, or if you're stating that it's probably ok to go with the precedent?

FirebaseDatabase/Sources/Utilities/FNextPushId.m
mortenbekditlevsen: Yes, we can certainly do that. My reasoning for using numerical values is that I think that it is easier to follow that we are adding or skipping or what we are doing. In the 'ordering' sense, the actual character value doesn't really matter as I see it.

But let me know if you think I should change it.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
mortenbekditlevsen: Right, so you mean make the entire switch statement operate on the `character` variable rather than on `plusOne`? That can of course be done, but switching on `plusOne` makes the values in the various cases exactly match the 'invalid' ones - or the ones where the `plusOne` is not the next value, but must be replaced by something else.

But I agree that the logic is both hard to read and it was also not easy to write.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
mortenbekditlevsen: Will fix.

FirebaseDatabase/Tests/Unit/FNextPushIdTest.m
mortenbekditlevsen: Will fix.

mortenbekditlevsen:
Ok, I have briefly considered the fix for 'integer' base key ordering.
I think that inconsistencies between the server key sorting and the `[FUtilities compareKey:toKey:]` sorting needs to be fixed first, because otherwise it's not clear which ordering this fix should be against - and if it should use the server ordering, then we can't use `[FUtilities compareKey:toKey:]` to test it.

So here are some things to consider:
1. Fix up `[FUtilities compareKey:toKey:]` to not only consider keys of character length less than 11 to be integer keys - in order to align with server algorithm. I think I noted somewhere upthread that this appears to be an issue in the Android SDK too, so someone should warn that team about the inconsistency too.
2. Introduce special casing for the fact that the two keys "-0" and "00" are in fact the same key (the sort equally and can be used interchangeably). Similarly for "-00" and "000", etc. up to max key length.
3. Next key after any positive integer key is "0"+key - until the key length is max key length, then the integer is incremented.
4. Next key after any negative integer key is "-0" + (negated) key
5. Next key after 0 is 00 (or "-0", but I guess the former is the more obvious choice)
6. Next key after a 'negative 0' '-0' is also either '000' or '-00', but I feel like 'normalising' to a sequence of zeroes is probably better.
7. Previous key after any positive integer (without zero prefix padding) is the integer value before, prepended with the appropriate amount of zeroes to make the key of length 'max key length'.
8. Ditto for negative integers
9. For zero prefixed integers (or '-0') then one 0 can just be removed.
10. For zero, the previous integer is -00...001
11. For values like "-0000" I also suggest normalising to the equivalent key "00000" before decrementing to the previous value "0000".

Does that sound about right?
@schmidt-sebastian @paulb777 Would you comment about the suggested fix for `[FUtilities compareKey:toKey:]` and let me know if I should do anything about the similar issue in the Android SDK?

FirebaseDatabase/Sources/Utilities/FNextPushId.h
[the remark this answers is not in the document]
schmidt-sebastian: I will leave this up to you. There is something to be said about consistency within a repo, so I am okay with either. I just wanted to point this out in case you had similar thoughts.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
schmidt-sebastian: I didn't think about it like that - let's leave it as is. The comments do help with the overall understanding.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
schmidt-sebastian: I will play with this today and let you know if I can come up with something clean. Thanks for your patience.

FirebaseDatabase/Sources/Utilities/FNextPushId.h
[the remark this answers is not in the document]
mortenbekditlevsen: I must admit that I didn't really consider this, I just added the extra arguments rather automatically. I'll just leave it as is for now.

FirebaseDatabase/Sources/Utilities/FNextPushId.m
schmidt-sebastian: Okay, so after two hours of playing with this I have now managed to understand why you wrote it this way. It is much easier to deal with the "forbidden" codepoints once translated. My understanding of Unicode is a bit murky, so I am not 100% convinced that the logic below works in all cases - but I think we can merge it, especially since it will be short lived. 

It would be good to add a test for the successor method similar to `testPredessorUnicode()`. In the meantime, I will try to see if I can improve my understanding.

mortenbekditlevsen: I'm no Unicode expert at all, but luckily all we need to know is about utf-16 encodings of Unicode code points, and that seems very well defined. My source of information on that actually just comes from the Wikipedia entry for utf-16.

schmidt-sebastian:
cc @jsdt and @IanWyszynski for comments on the ordering in `compareKey`.

schmidt-sebastian (approved):

IanWyszynski (approved):
There is an outstanding conversation here, but I think we can go ahead and merge this since it fixes the immediate issue. I plan to re-implement startAfter/endBefore on the server side to avoid introducing this complexity in all the different SDKs, but I haven't been able to prioritize that work yet.

paulb777:
Thanks @IanWyszynski  I resolved to unblock the merge and merged.

mortenbekditlevsen:
Should I create an issue to track the discrepancy with key sorting on iOS and Android vs js and server?

Issue #8731, opened 2021-10-01
FR: Support watchOS platform in xcframework, Carthage distribution


tiepvuvan:
### Issue
Downloading frameworks from Firebase distribution (https://firebase.google.com/download/ios) is not included watchOS platform support.
watchOS support is included in SPM and Cocoapods. 

### Proposal
Add support watchOS in the xcframework.

paulb777:
Thanks for the feature request. Our [ZipBuilder tool](https://github.com/firebase/firebase-ios-sdk/tree/master/ReleaseTooling) needs to be updated to generate a watchOS and watchOS simulator slice.

asendra:
Is there any work being done on this front? I tried to generate them myself, but so far no luck. Currently we are using it through CocoaPods, which is working. But we are trying to migrate everything to SPM / XCFrameworks

paulb777:
@asendra Your question is unclear. This issue is about binary distribution, and your question seems to be about SPM which supports watchOS for all the Firebase products that support it. Please open another issue to describe what isn't working.

asendra:
I was talking about binary distribution, should not have mentioned SPM, sorry.

PR #8821, opened 2021-10-18
Added support to ZipBuilder for adding watchOS and watchOS Simulator slices

simonmaddox:
This was first mentioned in #8731. @paulb777 said that ZipBuilder needed to be updated to add watchOS and watchOS Simulator slices. I believe that this Pull Request implements these changes.

Unfortunately I've been unable to test - running `swift run zip-builder` on my machine, even with a fresh clone, causes a crash somewhere deep in Ruby. I've searched for every mention of `tvOS` and re-implemented for `watchOS`. If this doesn't 100% work, then perhaps it will get you most of the way there.

I implemented this because my app requires iOS 10 support, which means I cannot use SPM distribution. However, watchOS support appears to only be distributed in the SPM version of the frameworks.

### DIFF
--- a/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
+++ b/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
@@ -23,8 +23,8 @@ import Foundation
 public let shared = Manifest(
   version: "8.9.0",
   pods: [
-    Pod("FirebaseCoreDiagnostics", zip: true),
-    Pod("FirebaseCore", zip: true),
+    Pod("FirebaseCoreDiagnostics", platforms: ["ios", "macos", "tvos", "watchos"], zip: true),
+    Pod("FirebaseCore", platforms: ["ios", "macos", "tvos", "watchos"], zip: true),
     Pod("FirebaseInstallations", zip: true),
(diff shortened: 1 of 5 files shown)


paulb777:
The CI is failing after two and a half hours - likely after building all the other platforms and failing when it starts to try watchos. Likely there are problems related to trying to build products not supported for watchOS. https://github.com/firebase/firebase-ios-sdk/blob/master/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift#L26 probably needs updates to turn watchos off in several places. 

Either way, a good next step would be to locally edit https://github.com/firebase/firebase-ios-sdk/blob/master/ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift and try to get a watchos build working for a single pod.

simonmaddox:
I've already followed that pattern for Analytics, but adding "Firebase" appears to require "Analytics". Not sure where to go now.

paulb777:
I looked a bit more. We don't currently support the Firebase pod for watchOS.   A workaround to the `ZipBuilder/ZipBuilder.swift:344: Fatal error:` fatal error could be to save the Firebase pod version from one of the other platforms as a class variable, since there is nothing currently platform specific in it.

paulb777:
Closing in favor of #12624

PR #8824, opened 2021-10-18
[Heartbeat Logging] Implement `HeartbeatInfo` data structure

ncooke3:
The `HeartbeatInfo` data model is basically a ring buffer of heartbeats. Its goal is to be optimized for limit on write with constant* time `append`s. 

Since there may be multiple `type`s of heartbeats: i.e. daily, weekly, monthly, a cache is used to store the last heartbeat for each `type`. This prevents the need to iterate through the buffer's contents to find when the last, e.g. `daily`, heartbeat was logged. I'm happy to remove the cacheing optimization if it's not worth it though.

I'd appreciate feedback on how I expressed how to decide _when_ we want to log a new heartbeat. 

This PR is next step from #8822, then I'll continue on to building test suite and refining.

#no-changelog

### DIFF
--- a/HeartbeatLogging/Sources/Heartbeat.swift
+++ b/HeartbeatLogging/Sources/Heartbeat.swift
@@ -15,7 +15,7 @@
 import Foundation
 
 /// An enumeration of time periods.
-internal enum TimePeriod: Int, CaseIterable, Codable {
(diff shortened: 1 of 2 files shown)


ncooke3:
Thanks @maksymmalyhin for awesome review!

PR #8880, opened 2021-10-28
Fix URL handling bug when path is a substring of host

### LINKED ISSUE #8874: Does 'sanitizing' url in `FUtilities parseUrl:` accidentally corrupt host?
Regarding FirebaseDatabase.

In the referenced line:

https://github.com/firebase/firebase-ios-sdk/blob/aab84c83209a7cc5a810913956e796a3e7f788a8/FirebaseDatabase/Sources/Utilities/FUtilities.m#L158

it appears as if the host may be corrupted in case the path is a substring of the host.

For instance:
```
- (void)testUrlParsedWithPathPartOfHost {
  FParsedUrl *parsedUrl = [FUtilities parseUrl:@"https://sample.firebaseio.com/a"];
  XCTAssertEqualObjects(parsedUrl.repoInfo.host, @"sample.firebaseio.com");
  XCTAssertEqualObjects(parsedUrl.repoInfo.namespace, @"sample");
  XCTAssertTrue(parsedUrl.repoInfo.secure);
  XCTAssertEqualObjects(parsedUrl.path, [FPath pathWithString:@"a"]);
}
```

Since 'a' is part of 'sample', it will replace it with 'smple'.
This could be an issue in the API: `FIRDatabaseReference referenceFromURL:` that parses a user-supplied URL.

paulb777:
Fix #8874 

Change logic to only replace the last match in the URL string instead of all of them

### DIFF
--- a/FirebaseDatabase/Sources/Utilities/FUtilities.m
+++ b/FirebaseDatabase/Sources/Utilities/FUtilities.m
@@ -154,9 +154,12 @@ void firebaseJobsTroll(void) {
 
     // Sanitize the database URL by removing the path component, which may
     // contain invalid URL characters.
-    NSRange lastMatch = [url rangeOfString:originalPathString options:NSBackwardsSearch];
-    NSString *sanitizedUrlWithoutPath = (lastMatch.location != NSNotFound) ?
-        [url stringByReplacingCharactersInRange:lastMatch withString: @""] : url;
+    NSRange lastMatch = [url rangeOfString:originalPathString
+                                   options:NSBackwardsSearch];
+    NSString *sanitizedUrlWithoutPath =
+        (lastMatch.location != NSNotFound)
+            ? [url stringByReplacingCharactersInRange:lastMatch withString:@""]
+            : url;
 
     NSURLComponents *urlComponents =
         [NSURLComponents componentsWithString:sanitizedUrlWithoutPath];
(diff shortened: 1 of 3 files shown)


FirebaseDatabase/Sources/Utilities/FUtilities.m
```diff
@@ -154,9 +154,13 @@ + (FParsedUrl *)parseUrl:(NSString *)url {
 
     // Sanitize the database URL by removing the path component, which may
     // contain invalid URL characters.
+    NSRange lastMatch = [url rangeOfString:originalPathString
+                                   options:NSBackwardsSearch];
     NSString *sanitizedUrlWithoutPath =
-        [url stringByReplacingOccurrencesOfString:originalPathString
-                                       withString:@""];
+        (lastMatch.location != NSNotFound)
+            ? [url stringByReplacingCharactersInRange:lastMatch withString:@""]
```
schmidt-sebastian, line 161: Isn't this just a substring now? I think it might be easier to parse if we use `substring(with: NSRange)`.

schmidt-sebastian (approved):

FirebaseDatabase/Sources/Utilities/FUtilities.m
paulb777: Good catch. Thanks!

PR #8887, opened 2021-10-29
Wait for Cocoapods update to be complete before updating C++ SDK.

jonsimantov:
This can prevent a race condition when the update is triggered before the Cocoapods main branch is updated.

### DIFF
--- a/.github/workflows/update-cpp-sdk-on-release.yml
+++ b/.github/workflows/update-cpp-sdk-on-release.yml
@@ -39,19 +39,22 @@ jobs:
         run: |
           # If the new version is simply vX.Y.Z, wait up to an hour for its podspec to be present.
           if [[ "$GITHUB_REF" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Checking Cocoapods repo for associated Firebase ${GITHUB_REF:1} podspec."
             podspec_url="https://github.com/CocoaPods/Specs/blob/master/Specs/0/3/5/Firebase/${GITHUB_REF:1}/Firebase.podspec.json"
             for retry in {1..12} error; do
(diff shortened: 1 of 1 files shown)


.github/workflows/update-cpp-sdk-on-release.yml
```diff
@@ -35,6 +35,25 @@ jobs:
           private_key: ${{ secrets.CPP_WORKFLOW_TRIGGER_APP_PRIVATE_KEY }}
           repository: firebase/firebase-cpp-sdk
 
+      - name: Ensure Cocoapods repo has been updated
+        run: |
+          # If the new version is simply vX.Y.Z, wait up to an hour for its podspec to be present.
+          if [[ "$GITHUB_REF" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            podspec_url="https://github.com/CocoaPods/Specs/blob/master/Specs/0/3/5/Firebase/${GITHUB_REF:1}/Firebase.podspec.json"
```
paulb777, line 42: FYI it may still take additional time to get from the GitHub repo to the CDN

jonsimantov: How long?

jonsimantov: (Should I have this check the repo downloaded via pod repo update?)

jonsimantov: Oh, actually, all of the podspecs we use to build C++ have `source 'https://github.com/CocoaPods/Specs.git'` in them, does that mean they will bypass the CDN?

paulb777: Yep, they'll be slower, but bypass the CDN in that case. So this flow could be fine

paulb777 (approved):

PR #8897, opened 2021-11-01
Fix RC db flags

### LINKED ISSUE #8807: Remote Config doesn't work if the device is locked (due to File Protection settings)

  * Xcode version: 12.4
  * Firebase SDK version: 8.7.0

If the app is started in the background (eg a background fetch or silent notification) then the remote config database is unable to be opened and all config settings return their default value. This is due to the file protection settings in the open database call here: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/RCNConfigDBManager.m#L217

#### Steps to reproduce:

Run an app that uses remote config on a real device with a passcode set, lock the device after the app has been installed and is just starting up.

Observe errors like the following in the logs:
```
[Firebase/RemoteConfig][I-RCN000015] Error message: authorization denied. Error code: 23.
[Firebase/RemoteConfig][I-RCN000016] Failed with SQL: SELECT bundle_identifier, namespace, key, value FROM main WHERE bundle_identifier = ?.
```

I have fixed this issue by changing the flag file protection flag to `SQLITE_OPEN_FILEPROTECTION_COMPLETEUNTILFIRSTUSERAUTHENTICATION` but I'm unsure of the reasons this is explicitly set to `SQLITE_OPEN_FILEPROTECTION_COMPLETE` currently

karenyz:
Fixes #8807

### DIFF
--- a/FirebaseRemoteConfig/CHANGELOG.md
+++ b/FirebaseRemoteConfig/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Fixed cached config not loading if device is locked. (#8807)
+
 # v8.0.0
 - [fixed] Fixed throttling issue when fetch fails due to no network. (#6628)
 - [fixed] Fixed issue where sometimes the local config returned is empty. (#7424)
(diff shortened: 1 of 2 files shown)


paulb777 (commented):
Thanks!

Please add a release note for 8.10.0

How did you verify/test?

karenyz:
Similar to what was mentioned in the github issue, tested on device and saw the auth error if the phone was locked during RC DB loading:

Screen Shot 2021-11-02 at 4 15 24 PM

No longer saw the error after updating the flag

paulb777 (approved):
LGTM

Issue #8895, opened 2021-11-01
Fatal Crash on ARKit with 8.9.0 (not with 8.8.0)


nidegen:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.1
  * Firebase SDK version: 8.9
  * Installation method: `Swift Package Manager`
  * Firebase Component: not sure (Auth, Core, Database, Firestore, Messaging, Storage, etc)

#### Steps to reproduce:

Launch ARKit while using: Auth, Firestore Listeners

#### Relevant Code:

Unfortunately, I am not sure where the bug gets triggered. It seems to originate in some App Check code.

What I am doing is launch a view controller for an ARKit Session. If i go back to 8.8 it works without issues.

Also what I have noticed is that `tokenDidChangeNotificationName`'s notification observer gets called very often before the crash happens. I'd say a few dozen times per second.

![Screenshot 2021-11-01 at 15 30 56](https://user-images.githubusercontent.com/8779423/139694576-3ce1ae3e-b9c2-4714-8057-07dbe64461d2.png)
![Screenshot 2021-11-01 at 15 30 43](https://user-images.githubusercontent.com/8779423/139694581-87478890-a279-483c-bc96-858d0380dc3c.png)
`

paulb777:
@nidegen Thanks for the report. The issue is likely related to the Firestore AppCheck integration that was introduced in 8.9.0

nidegen:
That's what I was also thinking, but was unable to trace the exact origin. The often called AppCheck Notification Observer of `tokenDidChangeNotificationName` did always have `if (![[contents->app name] isEqual:notified_app_name]) { return }` return in FirebaseAppCheckCredentialsProvider.

nidegen:
(Line 52 of firebase_app_check_credentials_provider_apple.mm)

maksymmalyhin:
@nidegen Do you install App Check SDK as well?

I guess the issue may be due to `app_check == nil` [here](https://github.com/firebase/firebase-ios-sdk/blob/71fa12f4133d74eb6c972ffdc133e3ffdb4b8b69/Firestore/core/src/credentials/firebase_app_check_credentials_provider_apple.mm#L37)  so `nil` is passed as a notification name to the [method](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1411723-addobserverforname) which means the class subscribes to basically all possible notifications.

cc @schmidt-sebastian

nidegen:
[...]

PR #8898, opened 2021-11-01
Check for AppCheck's existance before subscribing

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-ios-sdk/issues/8895

### DIFF
--- a/Firestore/core/src/credentials/firebase_app_check_credentials_provider_apple.mm
+++ b/Firestore/core/src/credentials/firebase_app_check_credentials_provider_apple.mm
@@ -31,8 +31,12 @@ namespace credentials {
 FirebaseAppCheckCredentialsProvider::FirebaseAppCheckCredentialsProvider(
     FIRApp* app, id<FIRAppCheckInterop> app_check) {
   contents_ = std::make_shared<Contents>(app, app_check);
-  std::weak_ptr<Contents> weak_contents = contents_;
 
+  if (app_check == nil) {
+    return;
+  }
+
+  std::weak_ptr<Contents> weak_contents = contents_;
   app_check_listener_handle_ = [[NSNotificationCenter defaultCenter]
       addObserverForName:[app_check tokenDidChangeNotificationName]
                   object:nil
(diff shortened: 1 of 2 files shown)


Firestore/core/src/credentials/firebase_app_check_credentials_provider_apple.mm
```diff
@@ -31,8 +31,12 @@
 FirebaseAppCheckCredentialsProvider::FirebaseAppCheckCredentialsProvider(
     FIRApp* app, id<FIRAppCheckInterop> app_check) {
   contents_ = std::make_shared<Contents>(app, app_check);
-  std::weak_ptr<Contents> weak_contents = contents_;
 
+  if (app_check == nil) {
```
maksymmalyhin, line 35: I wonder if it would be better to avoid creating an instance of `FirebaseAppCheckCredentialsProvider` at all if `app_check == nil`. Can `app_check == nil` lead to any unintended behaviour of `FirebaseAppCheckCredentialsProvider` instance?

maksymmalyhin (approved):
LGTM. I would encourage adding tests for this case to avoid regressions in the future.

Firestore/core/src/credentials/firebase_app_check_credentials_provider_apple.mm
schmidt-sebastian: The setup here matches our existing code for FirebaseAuth. The only difference is that the notification key for FirebaseAuth is a constant and always available. I would rather leave this as is.

I am also working on a test now, but my build setup is currently broken so it might take a while.

maksymmalyhin: SGTM 👍

schmidt-sebastian:
Merging as is to unblock the release. I will work on tests in a follow up PR (still unable to build locally).

paulb777:
@schmidt-sebastian It looks like there are several CI failures. Please revert or fix.

schmidt-sebastian:
I have started the CI run for the fix (https://github.com/firebase/firebase-ios-sdk/pull/8900). Is it possible to disable auto-merge until the CI is green?

paulb777:
I don't know a way to make that possible. I asked at https://github.com/github/feedback/discussions/7114

maksymmalyhin:
Currently it is possible if we make green checks a requirement for manual merge as well. We decided against it so far as it may block some workflows we use and also is a bit tedious to configure as it requires manually specifying each required check and we have a lot of them.

PR #8899, opened 2021-11-01
When downloading external projects, provide GitHub token if available.

jonsimantov:
When downloading external projects, if the GITHUB_TOKEN environment variable is set, pass it to the external download cmake function via HTTP_HEADER.

 This will prevent GitHub throttling the HTTP requests. Unauthenticated requests are limited by GitHub to 60 per IP address per hour, whereas authenticated requests are limited to 5,000 per user per hour.

(GitHub Actions runners automatically have the GITHUB_TOKEN environment variable set.)

Similar change in the Firebase C++ SDK: https://github.com/firebase/firebase-cpp-sdk/pull/717


paulb777:
Thanks. Please send cmake reviews to the Firestore team

dconeybe (approved):

PR #8900, opened 2021-11-01
Remove assert that AppCheckListener is always registered

schmidt-sebastian:
This fixes a regression in https://github.com/firebase/firebase-ios-sdk/pull/8898

### DIFF
--- a/Firestore/core/src/remote/grpc_root_certificates_generated.h
+++ b/Firestore/core/src/remote/grpc_root_certificates_generated.h
@@ -3,7 +3,7 @@
 #ifndef FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 #define FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 
-#include <cstdlib>
+#include <stddef.h>
 
 namespace firebase {
 namespace firestore {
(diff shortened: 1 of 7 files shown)


paulb777 (approved):
LGTM to merge once at least one of the previously failing tests succeeds.

Firestore/core/src/remote/grpc_root_certificates_generated.h
```diff
@@ -3,7 +3,7 @@
 #ifndef FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 #define FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 
-#include <stddef.h>
+#include <cstdlib>
```
paulb777, line 6: Is this intended?

schmidt-sebastian: Reverted

schmidt-sebastian:
FYI There still seems to be a test failure related to the previous PR. Will look in 30 minutes.

maksymmalyhin:
@schmidt-sebastian @paulb777 I would be more comfortable to go for release only once we have comprehensive code coverage for the cases when App Check or/and Auth are not installed in an app.

maksymmalyhin (approved):
LGTM on green checks. Tests for missing Auth and/or App Check are preferable but are not a blocker if you prefer manual testing before a release.

PR #8914, opened 2021-11-02
Add test that sends an invalid userInfo to AppCheck

schmidt-sebastian:


Firestore/core/test/unit/credentials/firebase_app_check_credentials_provider_test.mm
```diff
@@ -220,6 +220,45 @@ - (nonnull NSString*)tokenDidChangeNotificationName {
...
+      [token_promise](const std::string& result) {
+        if (result != "") {
+          token_promise->set_value(result);
+        }
+      });
+
+  // Sending this notifcation would cause a crash if it was processed in the
+  // AppCheck notification handlder since AppCheck expects the userInfo object
+  // to an NSDictionary.
+  id userInfo = @"this_should_be_a_dictionary";
```
maksymmalyhin, line 242: Hm, [userInfo](https://developer.apple.com/documentation/foundation/nsnotification/1409222-userinfo?language=objc) is expected to be a dictionary. If it is not then there is a bug in somewhere else on the notification poster side.

maksymmalyhin: I guess if we expect `userInfo` to be not a dictionary then we should handle it gracefully in the code. From the other side I'm not sure if we should code defensively against such types of issues. I don't feel strongly about it but this specific test seems to be a bit redundant to me.

maksymmalyhin (approved):
LGTM in general.

Firestore/core/test/unit/credentials/firebase_app_check_credentials_provider_test.mm
schmidt-sebastian: It's the only way I could reproduce the issue, so while it is a bit redundant it is IMHO worth keeping as it tests the entire problematic code flow, even if the fix is of much narrower scope.

PR #8923, opened 2021-11-03
CMakeLists.txt: add -DPB_NO_PACKED_STRUCTS=1 to the nanopb compiler flags

dconeybe:
This fixes a flurry of "ld: warning: pointer not aligned at address" warnings that reportedly fail the build of the C++ SDK.

Here are some customer reports of these warnings affecting builds:
- https://github.com/firebase/firebase-cpp-sdk/issues/712
- https://github.com/firebase/firebase-unity-sdk/issues/83

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,7 +242,7 @@ add_external_subdirectory(nanopb)
 
 target_compile_definitions(
   protobuf-nanopb-static PUBLIC
-  -DPB_FIELD_32BIT -DPB_ENABLE_MALLOC
+  -DPB_FIELD_32BIT -DPB_ENABLE_MALLOC -DPB_NO_PACKED_STRUCTS=1
 )
 
 # Enable #include <nanopb/pb.h>


CMakeLists.txt
```diff
@@ -242,7 +242,7 @@ add_external_subdirectory(nanopb)
 
 target_compile_definitions(
   protobuf-nanopb-static PUBLIC
-  -DPB_FIELD_32BIT -DPB_ENABLE_MALLOC
+  -DPB_FIELD_32BIT -DPB_ENABLE_MALLOC -DPB_NO_PACKED_STRUCTS=1
```
schmidt-sebastian, line 245: Do you know if this setting is enabled by default on older Mac OS versions? There might be some memory usage implications if this change enables it for the first time.

dconeybe: I do not know if this setting is enabled by default on older Mac OS versions.

My guess is that the unaligned pointers were always there but for whatever reason the build toolchain in macOS Monterey fails instead of warning about this.

It is true that disabling packed structs will likely cause the size of the protos in memory to increase; however, we already disable packed structs everywhere else and it hasn't been an issue that I am aware of. I can only guess that this location here was overlooked by accident.

The C++ SDK builds with `-DPB_NO_PACKED_STRUCTS=1` so we've probably just gotten incredibly lucky that the compiler flag mismatch hasn't caused any noticeable crashes or other undefined behavior.

paulb777: See https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseFirestore.podspec#L118 and https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L612

dconeybe: @paulb777 Exactly. This is the ONLY place where `PB_NO_PACKED_STRUCTS=1` is missing. It _really_ looks like an oversight.

schmidt-sebastian (approved):

PR #8952, opened 2021-11-09
Refactor library and add WIP controller tests

ncooke3:
Refactored library to improve testability and clarity.

Key Changes:
- All of the heartbeat logging logic (whether or not to log) is now in `HeartbeatController`
- Moved `HeartbeatInfo`'s visibility to internal and decoupled it's internal dependencies.
    -  Motivation: `HeartbeatInfo` represents stored heartbeats. Its structure is optimized for storage requirements (limit on write) and differs significantly from the structure the heartbeats will be in when they are sent.
- Added a new public model called `HeartbeatsPayload` 
   - Allows type safe encoding into header string.
- Added instance management system to `HeartbeatStorage` (tests for `HeartbeatStorage` are a WIP)
  -  Motivation: Multiple instances of `HeartbeatStorage` with the same identifier should share the same instance.

### DIFF
--- a/HeartbeatLogging/Sources/Heartbeat.swift
+++ b/HeartbeatLogging/Sources/Heartbeat.swift
@@ -18,7 +18,7 @@ import Foundation
 enum TimePeriod: Int, CaseIterable, Codable {
   /// The raw value is the number of calendar days within each time period.
   // TODO: Enable disabled types in future iterations.
-  case daily = 1 // , case weekly = 7, monthly = 28
+  case daily = 1 // , weekly = 7, monthly = 28
 
   /// The number of seconds in a given time period.
   var timeInterval: TimeInterval { Double(rawValue) * 86400 /* seconds in day */ }
@@ -33,9 +33,11 @@ struct Heartbeat: Codable, Equatable {
   private static let version: Int = 0
 
   /// An anonymous piece of information (i.e. user agent) to associate the heartbeat with.
-  let info: String
+  let agent: String
+
   /// The date when the heartbeat was recorded (standardized to be the start of a calendar day).
   let date: Date
+
   /// The heartbeat's model version.
   let version: Int
 
@@ -45,20 +47,20 @@ struct Heartbeat: Codable, Equatable {
   /// various time periods. Because a single heartbeat can help calculate moving averages for multiple
   /// time periods, this property serves to capture all the time periods that the heartbeat can represent in
   /// a moving average.
-  var timePeriods: [TimePeriod] = []
+  let timePeriods: [TimePeriod]
 
   /// Intializes a `Heartbeat` with given `info` and, optionally, a `date` and `version`.
   /// - Parameters:
-  ///   - info: An anonymous piece of information to associate the heartbeat with.
+  ///   - agent: An anonymous piece of information to associate the heartbeat with.
   ///   - date: The date when the heartbeat was recorded. Defaults to the current date.
   ///   - version: The heartbeat's model version. Defaults to the current model version.
-  init(info: String,
+  init(agent: String,
        date: Date = .init(),
-       version: Int = Self.version) {
-    self.info = info
-    // A heartbeat's date is a calendar day standardized at the start of day.
-    // TODO: Verify backend is using same calendar.
-    self.date = Calendar(identifier: .gregorian).startOfDay(for: date)
+       timePeriods: [TimePeriod],
+       version: Int = version) {
+    self.agent = agent
+    self.date = date
+    self.timePeriods = timePeriods
     self.version = version
   }
 }
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -20,10 +20,10 @@ public final class HeartbeatController {
   private let storage: HeartbeatStorageProtocol
   // TODO: Document.
   private let limit: Int = 30 // TODO: Decide on default value.
-  // TODO: Document.
+  /// Current date provider. It is used instead of `Date.init` for testability.
   private let dateProvider: () -> Date
   // TODO: Verify that this standardization aligns with backend.
-  // TODO: Document.
+  /// Used for standardizing dates for calendar-day comparision.
   static let dateStandardizer = Calendar(identifier: .gregorian).startOfDay(for:)
 
   /// Public initializer.
@@ -50,11 +50,12 @@ public final class HeartbeatController {
   ///
   /// - Note: This API is thread-safe.
   ///
-  /// - Parameter info: A `String` identifier to associate a new heartbeat with.
-  public func log(_ info: String) {
-    let (agent, date, capacity) = (info, dateProvider(), limit)
+  /// - Parameter agent: A `String` identifier to associate a new heartbeat with.
+  public func log(_ agent: String) {
+    let date = dateProvider()
+    let capacity = limit
 
-    storage.async { heartbeatInfo in
+    storage.readAndWriteAsync { heartbeatInfo in
       var heartbeatInfo = heartbeatInfo ?? HeartbeatInfo(capacity: capacity)
 
       let timePeriods = heartbeatInfo.cache.filter { timePeriod, lastDate in
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -75,7 +75,7 @@ public final class HeartbeatController {
   ///
   /// - Note: This API is thread-safe.
   ///
-  /// - Returns: The flushed heartbeats in the form of `HeartbeatInfo`.
+  /// - Returns: The flushed heartbeats in the form of `HeartbeatsPayload`.
   @discardableResult
   public func flush() -> HeartbeatsPayload {
     let capacity = limit
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -80,14 +80,16 @@ public final class HeartbeatController {
   public func flush() -> HeartbeatsPayload {
     let capacity = limit
 
-    let heartbeatInfo = try? storage.getAndReset { heartbeatInfo in
-      if let heartbeatInfo = heartbeatInfo {
-        // The new value that's stored will use the old's cache.
-        return HeartbeatInfo(capacity: capacity, cache: heartbeatInfo.cache)
-      } else {
+    let resetTransform: (HeartbeatInfo?) -> HeartbeatInfo? = { heartbeatInfo in
+      guard let oldHeartbeatInfo = heartbeatInfo else {
         return nil // Storage was empty.
       }
+
+      // The new value that's stored will use the old's cache.
+      return HeartbeatInfo(capacity: capacity, cache: oldHeartbeatInfo.cache)
     }
+
+    let heartbeatInfo = try? storage.getAndReset(using: resetTransform)
 
     return HeartbeatsPayload.makePayload(heartbeatInfo: heartbeatInfo)
   }
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -89,6 +89,9 @@ public final class HeartbeatController {
       return HeartbeatInfo(capacity: capacity, cache: oldHeartbeatInfo.cache)
     }
 
+    // Synchronously gets and returns the stored heartbeats and resets storage
+    // using the given transform. If the operation threw an error, assume no
+    // heartbeats were retrieved/reset.
     let heartbeatInfo = try? storage.getAndReset(using: resetTransform)
 
     return HeartbeatsPayload.makePayload(heartbeatInfo: heartbeatInfo)
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -22,6 +22,7 @@ public final class HeartbeatController {
   private let limit: Int = 30 // TODO: Decide on default value.
   // TODO: Document.
   private let dateProvider: () -> Date
+  // TODO: Verify that this standardization aligns with backend.
   // TODO: Document.
   static let dateStandardizer = Calendar(identifier: .gregorian).startOfDay(for:)
 
@@ -30,7 +31,7 @@ public final class HeartbeatController {
   /// - Parameter id: The `id` to associate this logger's internal storage with.
   public convenience init(id: String) {
     // TODO: Sanitize id.
-    let storage = HeartbeatStorage.storage(with: id)
+    let storage = HeartbeatStorage.getInstance(id: id)
     self.init(storage: storage)
   }
 
--- a/HeartbeatLogging/Sources/HeartbeatController.swift
+++ b/HeartbeatLogging/Sources/HeartbeatController.swift
@@ -18,21 +18,31 @@ import Foundation
 public final class HeartbeatController {
   /// The thread-safe storage object to log and flush heartbeats from.
   private let storage: HeartbeatStorageProtocol
+  // TODO: Document.
+  private let limit: Int = 30 // TODO: Decide on default value.
+  // TODO: Document.
+  private let dateProvider: () -> Date
+  // TODO: Document.
+  static let dateStandardizer = Calendar(identifier: .gregorian).startOfDay(for:)
 
   /// Public initializer.
   ///
   /// - Parameter id: The `id` to associate this logger's internal storage with.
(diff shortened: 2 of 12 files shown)


ncooke3:
CI is failing because the HeartbeatStorage tests are currently a WIP.

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -14,57 +14,84 @@
...
-/// A type that can store and remove hearbeats.
 protocol HeartbeatStorageProtocol {
-  func offer(_ heartbeat: Heartbeat)
-  func flush() -> HeartbeatInfo?
+  typealias HeartbeatInfoTransform = (HeartbeatInfo?) -> HeartbeatInfo?
+
+  func async(_ transform: @escaping HeartbeatInfoTransform)
+  // TODO: Evaluate if async variant of below API is needed.
+  func getAndReset(using transform: HeartbeatInfoTransform?) throws -> HeartbeatInfo?
 }
```
ncooke3, line 23: Note to readers: In an effort to shift the "whether or not to log a heartbeat" logic to `HeartbeatController`, I needed to make simpler API for this class.

HeartbeatLogging/Sources/Weak.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+// TODO: Document.
```
ncooke3, line 17: Note to readers: See `HeartbeatStorage`'s Instance Management for how this weak ref box was used.

HeartbeatLogging/Sources/HeartbeatInfo.swift
```diff
@@ -14,151 +14,108 @@
...
   }
-}
 
-private extension Date {
-  /// Calculates the time interval since a given `date`.
-  func timeIntervalSince(_ date: Date) -> TimeInterval {
-    timeIntervalSinceReferenceDate - date.timeIntervalSinceReferenceDate
+  func makeIterator() -> IndexingIterator<[Element]> {
+    circularQueue
+      .compactMap { $0 } // Remove `nil` elements.
```
paulb777, line 114: Why are there nil elements?

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -14,57 +14,86 @@
 
 import Foundation
 
-/// A type that can store and remove hearbeats.
 protocol HeartbeatStorageProtocol {
-  func offer(_ heartbeat: Heartbeat)
-  func flush() -> HeartbeatInfo?
+  typealias HeartbeatInfoTransform = (HeartbeatInfo?) -> HeartbeatInfo?
+
+  func async(_ transform: @escaping HeartbeatInfoTransform)
```
paulb777, line 20: Could be confusing to use a Swift keyword as a function name.  Is there something more descriptive?

paulb777 (approved):
A few questions, but nothing blocking. Good progress!

HeartbeatLogging/Sources/HeartbeatInfo.swift
ncooke3: The `circularQueue` is an array with a fixed size (Buffer is initialized with fixed capacity at initialization). Empty/placeholder array elements are `nil`. 

```swift
let fixedCapacityArray: Int? = Array(repeating: nil, count: 3)  // [nil, nil, nil]
```

HeartbeatLogging/Sources/Heartbeat.swift
```diff
@@ -45,20 +47,20 @@ struct Heartbeat: Codable, Equatable {
...
   /// time periods, this property serves to capture all the time periods that the heartbeat can represent in
   /// a moving average.
-  var timePeriods: [TimePeriod] = []
+  let timePeriods: [TimePeriod]
 
   /// Intializes a `Heartbeat` with given `info` and, optionally, a `date` and `version`.
   /// - Parameters:
-  ///   - info: An anonymous piece of information to associate the heartbeat with.
+  ///   - agent: An anonymous piece of information to associate the heartbeat with.
   ///   - date: The date when the heartbeat was recorded. Defaults to the current date.
```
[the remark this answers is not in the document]
ncooke3: Yes, dates are based on device. Foundation's `Date` type represents [A specific point in time, independent of any calendar or time zone.](https://developer.apple.com/documentation/foundation/date#:~:text=a%20specific%20point%20in%20time%2C%20independent%20of%20any%20calendar%20or%20time%20zone.). Given that, I believe traveling across time zones won't matter as dates are compared independent of calendar/timezone. 

I think it's worth adding a unit test that tests for this behavior so I'll do that!

There is an edge case worth calling out though since dates are based on device. If device time is manually changed, the date returned from `Date()` will be distorted based on the actual time IRL. The current implementation is susceptible to this behavior as well though. It's another layer of complexity though for sure though.

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -14,57 +14,86 @@
...
     self.queue = queue ?? DispatchQueue(label: "com.heartbeat.storage.\(id)")
   }
 
-  func offer(_ heartbeat: Heartbeat) {
+  // MARK: - Instance Management
+
+  // TODO: Add tests for instance management.
+
+  /// <#Description#>
+  static var cachedInstances: [String: Weak<HeartbeatStorage>] = [:]
```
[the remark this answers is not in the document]
ncooke3: We need a weak box because collections strongly retain values so cached `HeartbeatStorage` will never be deallocated otherwise.

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -41,17 +52,48 @@ public final class HeartbeatController {
   ///
   /// - Parameter info: A `String` identifier to associate a new heartbeat with.
   public func log(_ info: String) {
-    let newHeartbeat = Heartbeat(info: info)
-    storage.offer(newHeartbeat)
+    let (agent, date, capacity) = (info, dateProvider(), limit)
```
maksymmalyhin, line 55: Optional. Is it same as just assigning values of the variables on separate lines like `let agent = info`? If so, using 3 separate lines for it looks more readable to me.

HeartbeatLogging/Sources/HeartbeatStorage.swift
maksymmalyhin: I like the more verbose `readAndWriteAsync` version so far as it saves the reader time since more information can be get from code without looking at API docs.

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -70,11 +56,23 @@ extension FileStorage: PersistentStorage {
...
   }
+
+  private func createDirectories(in url: URL) throws {
+    do {
+      try fileManager.createDirectory(
+        at: url,
+        withIntermediateDirectories: true,
+        attributes: nil
+      )
+    } catch CocoaError.fileWriteFileExists {
```
maksymmalyhin, line 67: I like Swift error handling so much more than Objective-C one!

@ncooke3 @paulb777 I think improving Firebase error system to optimize for Swift handling can be a next significant API improvement.

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -14,57 +14,86 @@
...
-  func offer(_ heartbeat: Heartbeat) {
+  // MARK: - Instance Management
+
+  /// <#Description#>
+  private(set) static var cachedInstances: [String: Weak<HeartbeatStorage>] = [:]
+
+  /// <#Description#>
+  /// - Parameter id: <#id description#>
+  /// - Returns: <#description#>
+  static func getInstance(id: String) -> HeartbeatStorage {
```
maksymmalyhin, line 54: FYI. We had a discussion with @VinayGuthal and @vkryachko yesterday and they mostly convinced me that a case when we have two Firebase app instances with the same app ID is very rare, so we can ignore it. If we don't account for this case then we can have an instance of storage Firebase app instance which may simplify the implementation (it was the case for Android). In this case storage would need to be keyed by the app ID and app name. No immediate implications for Apple SDK so far, but if implementation becomes too complex, feel free to apply this simplification as well.

HeartbeatLogging/Sources/Weak.swift
```diff
@@ -0,0 +1,22 @@
...
+
+/// <#Description#>
+struct Weak<Object: AnyObject> {
```
maksymmalyhin, line 20: The name looks too generic to me. I would prefer more verbose like `WeakContainer` or something along these lines.

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
```diff
@@ -15,45 +15,269 @@
...
-      return flushed
+    func getAndReset(using transform: HeartbeatInfoTransform?) throws -> HeartbeatInfo? {
+      let oldHeartbeatInfo = heartbeatInfo
+      heartbeatInfo = transform?(heartbeatInfo)
+      return oldHeartbeatInfo
     }
   }
+
+  /// Simulates the device system time.
+  class SystemClock {
```
maksymmalyhin, line 233: This name associates with a real system clock implementation for me. I would suggest updating the name with more reference to its intended use, e.g. `TestClock`. Also it seems it could be just a private extension on `Date` type. WDYT?

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
```diff
@@ -15,45 +15,269 @@
...
+    )
+
+    assertHeartbeatControllerFlushesEmptyPayload(controller)
+
+    // When
+    // - Clock time 2021-11-01 @ 00:00:00 (EST)
+    controller.log("dummy_agent")
+
+    // - Advance to 2021-11-01 @ 23:59:59 (EST)
+    do { clock.advance(by: 60 * 60 * 24 - 1) }
```
maksymmalyhin, line 114: I'm curious what is the motivation to wrap this statement into `do {}`?

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
```diff
@@ -15,45 +15,269 @@
...
+
+    var formattedDate: String {
+      HeartbeatsPayload.dateFormatter.string(from: date)
+    }
+  }
+
+  // TODO: - Revisit below assertion implementation.
+  // This can be simplified further by making HeartbeatsPayload conform to Equatable...
+  func assertEqualPayloadStrings(_ encoded: String, _ literal: String) {
+    let encodedData = Data(base64Encoded: encoded)!
```
maksymmalyhin, line 252: Will using `try XCTUnwrap()` here and below make a potential failure easier to investigate in the future comparing to force unwrap?

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
```diff
@@ -17,178 +17,315 @@ import XCTest
...
+    XCTAssertNotNil(heartbeatStorage2)
+    XCTAssertEqual(
+      HeartbeatStorage.cachedInstances,
+      ["sparky": Weak(object: heartbeatStorage2)]
+    )
+    // When
+    heartbeatStorage2 = nil
+    // - Then
+    XCTAssertNil(heartbeatStorage2)
+    XCTAssertEqual(HeartbeatStorage.cachedInstances, [:])
```
maksymmalyhin, line 116: Note that implementation details are used here which force the test to change when the implementation changes even without changing the public API. Asserting implementation details should be avoided when possible. Doesn't look like a big concern in this case, but it seems to be avoidable. Not a blocker for the PR.

maksymmalyhin (approved):
LGTM, a couple of non-blocking comments.

HeartbeatLogging/Sources/Weak.swift
ncooke3: `WeakContainer` is good. Would you prefer `WeakReference` or `WeakContainer`?

maksymmalyhin: Both LGTM, but I like WeakContainer slightly more.

ncooke3: SGTM, thanks

HeartbeatLogging/Sources/HeartbeatStorage.swift
ncooke3: Thanks @maksymmalyhin. I think I'll go with:
```swift
func readAndWriteAsync(using transform: @escaping HeartbeatInfoTransform)
```

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
ncooke3: I was experimenting with trying to differentiate public API usage from test logic to get us in the desired state.  I was trying to emphasize that the instructions in the `do {}` block are not public API related operations but test related code to get us in the desired state.

I also didnt know you could use `do {}` blocks like that. It makes for a nicer syntax then doing `_ = { ... }()`!

BUT, from looking at this again, I think renaming `clock`  to `testClock` makes it clear enough without silly `do {}` stuff.

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
```diff
@@ -15,45 +15,248 @@
...
+    )
+
+    assertHeartbeatControllerFlushesEmptyPayload(controller)
+
+    // When
+    // - Clock time 2021-11-01 @ 00:00:00 (EST)
+    controller.log("dummy_agent")
+
+    // - Advance to 2021-11-01 @ 23:59:59 (EST)
+    testDate.addTimeInterval(60 * 60 * 24 - 1)
```
ncooke3, line 114: Thanks to your comments @maksymmalyhin, I took a closer look and realized I was able to remove the `SystemTime` class entirely because the `HeartbeatController` initializer captures the `testDate` in an `@escaping` closure so we can treat the date with reference semantics (`testDate.addTimeInterval(60 * 60 * 24 - 1)`) and the controller's date provider will reflect those changes when we use it in public API implementation.

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
ncooke3: Reworked in e9d0361

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
ncooke3: Reworked in e9d0361. This led to some nice cleanup, see comment below.

HeartbeatLogging/Sources/HeartbeatController.swift
ncooke3: Reworked in e9d0361

HeartbeatLogging/Sources/Weak.swift
ncooke3: Reworked in e9d0361

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: Hmm, thanks for pointing this out. I'd like to understand your point more. So in this case, are the exposed implementation details the fact that it is a `Dictionary<String:WeakContainer>`? 

So then `HeartbeatStorage.cachedInstances` should be private? 
Without checking backing dictionary, I'm not sure how to confirm that an instance is removed from `cachedInstances`... would checking that a new instance is returned be enough?

```swift
var heartbeatStorage: HeartbeatStorage? = HeartbeatStorage.getInstance(id: "sparky")
heartbeatStorage = nil

// Instead of inspecting dictionary contents, try to get Instance again and assert that it is not the same instance as before (implying that the old one was discarded and new one was created?).
let newHeartbeatStorage= HeartbeatStorage.getInstance(id: "sparky")
XCTAssert(newHeartbeatStorage !== heartbeatStorage) 
```

maksymmalyhin: Yes, this looks sufficient to validate the API contract.

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
ncooke3: Reworked in 48a668b

PR #8957, opened 2021-11-11
Port Functions integration tests to Swift

paulb777:
- Translate https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseFunctions/Tests/Integration/FIRIntegrationTests.m to Swift
- Add SPM targets for the integration tests
- Add FIRFunctionsFake.m to workaround the `init` function not bridging to Swift because of its parameters using forward declarations and to use the existing Objective C fake implementations

Future PRs will:
- Add a CocoaPods test_spec for the Swift integration tests
- Add async/await variations of the tests
- Review API coverage
- Prepare for Codable PR integration - See #8854

### DIFF
--- a/FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
+++ b/FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
@@ -19,24 +19,29 @@ import FirebaseFunctionsTestingSupport
 import XCTest
 
 class IntegrationTests: XCTestCase {
-  let functions = FunctionsFake.init(projectID: "functions-integration-test", region: "us-central1", customDomain: nil, withToken: nil)
+  let functions = FunctionsFake(
+    projectID: "functions-integration-test",
+    region: "us-central1",
+    customDomain: nil,
+    withToken: nil
+  )
   let projectID = "functions-swift-integration-test"
-  
+
   override func setUp() {
     super.setUp()
     functions.useLocalhost()
   }
-  
+
   func testData() {
     let expectation = self.expectation(description: #function)
     let data = [
-      "bool" : true,
-      "int" : 2 as Int32,
-      "long" : 9876543210,
-      "string" : "four",
-      "array" : [5 as Int32, 6 as Int32],
-      "null" : nil
-    ] as [String : Any?]
+      "bool": true,
+      "int": 2 as Int32,
+      "long": 9_876_543_210,
+      "string": "four",
+      "array": [5 as Int32, 6 as Int32],
+      "null": nil,
+    ] as [String: Any?]
     let function = functions.httpsCallable("dataTest")
     XCTAssertNotNil(function)
     function.call(data) { result, error in
@@ -56,7 +61,7 @@ class IntegrationTests: XCTestCase {
     }
     waitForExpectations(timeout: 1)
   }
-  
+
   func testScalar() {
     let expectation = self.expectation(description: #function)
     let function = functions.httpsCallable("scalarTest")
@@ -73,12 +78,17 @@ class IntegrationTests: XCTestCase {
     }
     waitForExpectations(timeout: 1)
   }
-  
+
   func testToken() {
     // Recreate _functions with a token.
-    let functions = FunctionsFake.init(projectID: "functions-integration-test", region: "us-central1", customDomain: nil, withToken: "token")
+    let functions = FunctionsFake(
+      projectID: "functions-integration-test",
+      region: "us-central1",
+      customDomain: nil,
+      withToken: "token"
+    )
     functions.useLocalhost()
-    
+
     let expectation = self.expectation(description: #function)
     let function = functions.httpsCallable("FCMTokenTest")
     XCTAssertNotNil(function)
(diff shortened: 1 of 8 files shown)


FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
```diff
@@ -0,0 +1,258 @@
...
+  )
+  let projectID = "functions-swift-integration-test"
+
+  override func setUp() {
+    super.setUp()
+    functions.useLocalhost()
+  }
+
+  func testData() {
+    let expectation = self.expectation(description: #function)
```
maksymmalyhin, line 36: nit: `self` is not required here, can be just `expectation(description: #function)`.

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
```diff
@@ -0,0 +1,258 @@
...
+        let data = try XCTUnwrap(result?.data as? [String: Any])
+        let message = try XCTUnwrap(data["message"] as? String)
+        let long = try XCTUnwrap(data["long"] as? Int64)
+        let code = try XCTUnwrap(data["code"] as? Int32)
+        XCTAssertEqual(message, "stub response")
+        XCTAssertEqual(long, 420)
+        XCTAssertEqual(code, 42)
+        expectation.fulfill()
+      } catch {
+        XCTAssert(false, "Failed to unwrap the function result")
```
maksymmalyhin, line 59: I'm curious if it is worth printing the error here or `XCTUnwrap` provides enough information in place?

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
```diff
@@ -0,0 +1,258 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+import FirebaseFunctions
+import FirebaseFunctionsTestingSupport
+import XCTest
+
+class IntegrationTests: XCTestCase {
```
maksymmalyhin, line 21: Could you please add a comment somewhere about the fact that test function definitions are in `FirebaseFunctions/Backend/index.js` and mentions preretirements for the test like starting emulator, etc. It will safe some time for future us.

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  func testMissingResult() {
+    let expectation = self.expectation(description: #function)
+    let function = functions.httpsCallable("missingResultTest")
+    XCTAssertNotNil(function)
+    function.call(nil) { result, error in
+      do {
+        XCTAssertNotNil(error)
+        let error = try XCTUnwrap(error! as NSError)
```
maksymmalyhin, line 149: I wonder what `!` is needed for in this line?

maksymmalyhin (approved):
👍  LGTM, a couple non-blocking nits.

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
paulb777: Good idea! Done.

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
paulb777: It can only help to print. Done!

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
paulb777: The paren's were in the wrong place. Fixed. Good catch!

paulb777 (commented):
Thanks for the thorough review and good suggestions!

FirebaseFunctions/Tests/SwiftIntegration/IntegrationTests.swift
```diff
@@ -0,0 +1,264 @@
...
+    function.call(nil) { result, error in
+      do {
+        XCTAssertNotNil(error)
+        let error = try XCTUnwrap(error) as NSError
+        XCTAssertEqual(FunctionsErrorCode.internal.rawValue, error.code)
+        XCTAssertEqual("Response is missing data field.", error.localizedDescription)
+        expectation.fulfill()
+      } catch {
+        XCTAssert(false, "Failed to unwrap the function result: \(error)")
+      }
```
ncooke3, line 161: LGTM. You can avoid the `do {} catch {}` blocks by adding `throws` to the test method signature. If the `XCTUnwrap` fails, it'll highlight the line with the failing unwrap.

maksymmalyhin: I don't think it will work if `XCTUnwrap` is used inside of the closure.

paulb777: Right @maksymmalyhin That's what I found:

Screen Shot 2021-11-11 at 8 40 56 AM

ncooke3: Oh, never thought about that. Makes sense, thanks for pointing that out y'all!

PR #8969, opened 2021-11-15
Hackweek Firebase Performance -> Crashlytics Interop

samedson:
Hackweek


Interop/Crashlytics/Public/FIRCrashlyticsInterop.h
```diff
@@ -0,0 +1,63 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Adds logging that is sent with your crash data. The logging does not appear  in the
+ * system.log and is only visible in the Crashlytics dashboard.
+ *
+ * @param msg Message to log
+ */
+- (void)log:(NSString *)msg;
```
pydichandra, line 29: Is this same as breadcrumb in dashboard?
If not, do we need to define a method to log breadcrumbs as well here?

FirebasePerformance.podspec
```diff
@@ -32,6 +32,7 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
     'FirebaseCore/Sources/Private/*.h',
     'FirebaseInstallations/Source/Library/Private/*.h',
     'FirebaseRemoteConfig/Sources/Private/*.h',
+    'Interop/Crashlytics/Public/*.h',
```
paulb777, line 35: Fine for hackweek, but the Interop directory is deprecated and new ones should be created like FirebaseAppCheck/Sources/Interop

samedson: Ooh cool thanks for that tip

Interop/Crashlytics/Public/FIRCrashlyticsInterop.h
samedson: Yep that'll do breadcrumbs

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
```diff
@@ -92,6 +94,23 @@ - (void)updateSessionId:(NSNotification *)notification {
...
+  NSError *error;
+  NSData *crashlyticsSessionJsonBreadcrumb =
+      [NSJSONSerialization dataWithJSONObject:crashlyticsSessionBreadcrumb options:0 error:&error];
+  if (!crashlyticsSessionJsonBreadcrumb) {
+    NSLog(@"Got an error: %@", error);
+  } else {
+    NSString *jsonString = [[NSString alloc] initWithData:crashlyticsSessionJsonBreadcrumb
+                                                 encoding:NSUTF8StringEncoding];
+    // Getting the FirePerf shared instance here. Is there a better way to do that internally?
+    [[FIRPerformance sharedInstance].crashlytics log:jsonString];
```
samedson, line 112: Only weirdish part, but might be ok

PR #8976, opened 2021-11-16
Attempt to make Slice/ByteBuffer work

schmidt-sebastian:


Firestore/core/CMakeLists.txt
```diff
@@ -206,7 +206,8 @@ firebase_ios_glob(
   src/objc/*.h
   src/remote/*.cc
   src/remote/*.h
-  src/remote/grpc_adapt/grpc_adaption.h
+  src/remote/grpc_adapt/*.h
```
wu-hui, line 209: CMake should not include these. They should only be there to support grpc-swift.

Firestore/core/src/remote/grpc_adapt/grpc_swift_byte_buffer.h
```diff
@@ -0,0 +1,56 @@
...
+  ByteBuffer();
+  ByteBuffer(const Slice* slices, size_t nslices);
+
+  /// Buffer size in bytes.
+  size_t Length() const;
+  /// Dump (read) the buffer contents into \a slices.
+  Status Dump(std::vector<Slice>* slices) const;
+
+ private:
+  ByteBufferShim* shim_;
```
wu-hui, line 48: Headers cannot reference to Swift shims..because they will be included in .cc files.

You can have a ByteBufferImpl class defined in .mm file, and subclass ByteBuffer.

Firestore/core/src/remote/grpc_adapt/grpc_swift_slice.h
```diff
@@ -0,0 +1,56 @@
...
+  /// Byte size.
+  size_t size() const;
+
+  /// Raw pointer to the beginning (first element) of the slice.
+  const uint8_t* begin() const;
+
+ private:
+  friend class ByteBuffer;
+
+  GRPCSliceShim* shim_;
```
wu-hui, line 48: Similar comment about referencing Swift Shims here.

wu-hui (commented):
Left some comments..

Firestore/core/src/remote/grpc_adapt/grpc_swift_byte_buffer.h
schmidt-sebastian: Can I also not use forward declaration?

Firestore/core/CMakeLists.txt
schmidt-sebastian: Removed

Firestore/core/src/remote/grpc_adapt/grpc_swift_byte_buffer.h
wu-hui: FWIW, you can probably follow `grpc_swift_channel.h/mm`. There is a `ClientContext` that does similar things.

PR #8981, opened 2021-11-18
Update Functions public API docs for FCM Token instead of IID

paulb777:
Similar changes are needed for the Codable APIs introduced in #8854

#no-changelog

### DIFF
--- a/FirebaseFunctions/Sources/Public/FirebaseFunctions/FIRHTTPSCallable.h
+++ b/FirebaseFunctions/Sources/Public/FirebaseFunctions/FIRHTTPSCallable.h
@@ -47,11 +47,11 @@ NS_SWIFT_NAME(HTTPSCallable)
  * Executes this Callable HTTPS trigger asynchronously without any parameters.
  *
  * The request to the Cloud Functions backend made by this method automatically includes a
- * Firebase Installations token to identify the app instance. If a user is logged in with Firebase
+ * FCM token to identify the app instance. If a user is logged in with Firebase
  * Auth, an auth ID token for the user is also automatically included.
  *
- * Firebase Installations sends data to the Firebase backend periodically to collect information
- * regarding the app instance. To stop this, see `Installations.delete()`. It
+ * Firebase Cloud Messaging sends data to the Firebase backend periodically to collect information
+ * regarding the app instance. To stop this, see `Messaging.deleteData()`. It
  * resumes with a new Installations ID the next time you call this method.
  *
  * @param completion The block to call when the HTTPS request has completed.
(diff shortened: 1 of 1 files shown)


FirebaseFunctions/Sources/Public/FirebaseFunctions/FIRHTTPSCallable.h
```diff
@@ -47,12 +47,12 @@ NS_SWIFT_NAME(HTTPSCallable)
...
  * The request to the Cloud Functions backend made by this method automatically includes a
- * Firebase Instance ID token to identify the app instance. If a user is logged in with Firebase
+ * Firebase Installations token to identify the app instance. If a user is logged in with Firebase
  * Auth, an auth ID token for the user is also automatically included.
  *
- * Firebase Instance ID sends data to the Firebase backend periodically to collect information
- * regarding the app instance. To stop this, see `[FIRInstanceID deleteIDWithHandler:]`. It
- * resumes with a new Instance ID the next time you call this method.
+ * Firebase Installations sends data to the Firebase backend periodically to collect information
+ * regarding the app instance. To stop this, see `Installations.delete()`. It
```
maksymmalyhin, line 54: Hm, actually Functions use FCM token now. FCM token will also be refreshed when `Installations.delete()` is called but maybe we should be more precise here. @chliangGoogle WDYT?

maksymmalyhin (approved):
LGTM, it looks more precise and useful than before, but it is still slightly inaccurate. I guess  @chliangGoogle can help with better wording.

FirebaseFunctions/Sources/Public/FirebaseFunctions/FIRHTTPSCallable.h
```diff
@@ -47,12 +47,12 @@ NS_SWIFT_NAME(HTTPSCallable)
  * Executes this Callable HTTPS trigger asynchronously without any parameters.
  *
  * The request to the Cloud Functions backend made by this method automatically includes a
- * Firebase Instance ID token to identify the app instance. If a user is logged in with Firebase
+ * Firebase Installations token to identify the app instance. If a user is logged in with Firebase
```
chliangGoogle, line 50: I think it's called FCM Token now: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/Public/FirebaseMessaging/FIRMessaging.h#L261

And it's not part of Firebase installations

FirebaseFunctions/Sources/Public/FirebaseFunctions/FIRHTTPSCallable.h
chliangGoogle: Yeah and if function uses token, then they also need to call Messaging.deleteData() to remove FCM token info.
https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/Public/FirebaseMessaging/FIRMessaging.h#L397

chliangGoogle (changes requested):

chliangGoogle (approved):

PR #8986, opened 2021-11-18
Stop posting reports in PRs not from master

### LINKED ISSUE #8985: Size and coverage reports broken for PRs targeting non-master branches
See #8983 

It's probably fine to disable the reports if master is not the PR merge target.

granluo:
This is to fix #8985 
Reports will not be posted to PRs rooted from master in presubmirs, but data will still be uploaded in postsubmits.


paulb777 (approved):
Thanks!

PR #8987, opened 2021-11-19
Copy test app into new target and migrate ApplicationDelegate to Swift

rosalyntan:
First PR to make a copy of the Firebase Auth test app in Swift. Development will be done in the `auth-swift-testapp` branch, and merged into `master` once entirely complete.

The majority of this PR is just copying the existing Objective-C test app into a new target. The only actual code changes are:
- Updates to `Podfile` and `project.pbxproj` to configure the new target
- Adding a bridging header
- Replacing `ApplicationDelegate.h`, `ApplicationDelegate.m`, and `main.m` with `AppDelegate.swift`
- Update `FacebookAuthProvider.m` and `GoogleAuthProvider.m` to reference `AppDelegate.swift`


paulb777:
Thanks for putting this together @rosalyntan!

Instead of duplicating all of the files and adding a new directory, how about managing the two apps via the two different Xcode targets and only adding the new files? 

Screen Shot 2021-11-19 at 11 15 13 AM

Also, please add a build test for the new Swift app to CI - see https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/auth.yml#L66 and https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh#L297

rosalyntan:
> Instead of duplicating all of the files and adding a new directory, how about managing the two apps via the two different Xcode targets and only adding the new files?

Done. Since `FacebookAuthProvider.m` and `GoogleAuthProvider.m` have changes in their imports to reference the new Swift `AppDelegate`, I kept the duplication of those files. Let me know if there's a better way to do that.

> Also, please add a build test for the new Swift app to CI - see https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/auth.yml#L66 and https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh#L297

Added `RunXcodebuild` entries for `AuthSample` and `SwiftSample`, although I'm not sure how to verify that they're correctly running in CI (I saw references to `SwiftSample` in the `auth/integration-tests` logs, so I think it's configured correctly?).

scripts/build.sh
```diff
@@ -296,6 +296,18 @@ case "$product-$platform-$method" in
 
   Auth-*-xcodebuild)
     if check_secrets; then
+      RunXcodebuild \
```
paulb777, line 299: Redundant since the integration tests below depend on AuthSample - but fine for completeness.

FirebaseAuth/Tests/Sample/SwiftSample/GoogleAuthProvider.m
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright 2019 Google
```
paulb777, line 2: Could avoid duplicating these two files with ifdef's and a build setting in the Xcode project, but this could be fine, especially if they're going to evolve separately.

paulb777 (approved):
I see the SwiftSample build success at https://github.com/firebase/firebase-ios-sdk/runs/4270425361?check_suite_focus=true#step:6:1485

scripts/build.sh
rosalyntan: Ack, thanks! I will leave this for completeness.

FirebaseAuth/Tests/Sample/SwiftSample/GoogleAuthProvider.m
rosalyntan: Good to know, thanks! I'll leave it as is for now, as I expect there to be more changes.

PR #8994, opened 2021-11-22
[Heartbeat Logging] - Add underlying storage tests

ncooke3:
This PR adds tests for the wrapper classes used to read and write from file system and user defaults.

#no-changelog

### DIFF
--- a/HeartbeatLogging/Sources/Storage.swift
+++ b/HeartbeatLogging/Sources/Storage.swift
@@ -46,11 +46,12 @@ final class FileStorage: Storage {
 
   func write(_ value: Data?) throws {
     do {
+      try createDirectories(in: url.deletingLastPathComponent())
       if let value = value {
-        try createDirectories(in: url.deletingLastPathComponent())
         try value.write(to: url, options: .atomic)
       } else {
-        try? fileManager.removeItem(at: url)
+        let emptyData = Data()
+        try emptyData.write(to: url, options: .atomic)
       }
     } catch {
       throw StorageError.writeError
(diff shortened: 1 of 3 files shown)


HeartbeatLogging/Tests/Unit/StorageTests.swift
```diff
@@ -0,0 +1,186 @@
...
+    let storedData = try fileStorage.read()
+    XCTAssertEqual(storedData, modifiedData)
+  }
+
+  func testWriteNil_WhenFileDoesNotExist_DoesNothing() throws {
+    // Given
+    let fileStorage = FileStorage(url: makeTemporaryFileURL())
+    XCTAssertThrowsError(try fileStorage.read())
+    // When
+    XCTAssertNoThrow(try fileStorage.write(nil))
```
paulb777, line 69: Why does write(nil) remove versus write an empty file?

I would expect `remove` to be a different API and `write` nil to be an empty operation.

paulb777 (approved):

HeartbeatLogging/Tests/Unit/StorageTests.swift
ncooke3: Ok, thanks for the input. I'll rework this to make `write(nil)` empty the file rather than remove it.

ncooke3: I reworked things in 26f5246 so that write(nil) is an operation that write empty data to the storage.

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -46,11 +46,12 @@ final class FileStorage: Storage {
```
paulb777, line 46: Should Storage have a `remove` function that still calls `removeItem`?

paulb777 (approved):

HeartbeatLogging/Sources/Storage.swift
ncooke3: There is no need for it now. The main API requirements are to:
-  read 
-  write data
-  empty file(`write(nil)`)

PR #8995, opened 2021-11-22
Add platform flag to environment description.

ryanwilson:
This is to call out specific platform issues. Open to suggestions but this seemed like the smallest possible change to get the extra information.


paulb777 (approved):
Thanks!

ncooke3 (approved):

PR #8999, opened 2021-11-23
[WIP][HeartbeatLogging] `FirebaseCore` integration

ncooke3:
Drafting the HeartbeatLogging integration into `FirebaseCore`.

#no-changelog


paulb777:
I'm getting convinced that there's no way to make requiring a Swift library a non-breaking change for the CocoaPods `use-libraries` use case:

```
 -> FirebaseStorage (8.10.0)
    - ERROR | [iOS] unknown: Encountered an unknown error (The following Swift pods cannot yet be integrated as static libraries:
```

ncooke3:
The heartbeat Swift library (branch: heartbeats-main) has changed a fair amount since opening this PR. Will close in favor of a new PR.

PR #9005, opened 2021-11-24
Fixing the short url logic

eldhosembabu:
Added back the regex to check for more than one path in short url.

### DIFF
--- a/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
+++ b/FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
@@ -295,7 +295,10 @@ BOOL FIRDLMatchesShortLinkFormat(NSURL *URL) {
   // Path cannot be prefixed with /link/dismiss
   BOOL isDismiss = [[URL.path lowercaseString] hasPrefix:@"/link/dismiss"];
 
-  return hasPathOrCustomDomain && !isDismiss && canParse;
+  BOOL matchesRegularExpression =
+        ([URL.path rangeOfString:@"/[^/]+" options:NSRegularExpressionSearch].location != NSNotFound);
+
+  return hasPathOrCustomDomain && !isDismiss && canParse && matchesRegularExpression;
 }
 
 NSString *FIRDLMatchTypeStringFromServerString(NSString *_Nullable serverMatchTypeString) {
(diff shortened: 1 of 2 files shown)


FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
```diff
@@ -295,7 +295,10 @@ BOOL FIRDLMatchesShortLinkFormat(NSURL *URL) {
   // Path cannot be prefixed with /link/dismiss
   BOOL isDismiss = [[URL.path lowercaseString] hasPrefix:@"/link/dismiss"];
 
-  return hasPathOrCustomDomain && !isDismiss && canParse;
+  BOOL matchesRegularExpression =
+      ([URL.path rangeOfString:@"/[^/]+" options:NSRegularExpressionSearch].location != NSNotFound);
```
ncooke3, line 299: I saw `/[^/]+` pattern was used previously int he SDK. Do you think it is worth adding a comment for what this pattern means?

ncooke3 (approved):
Thanks for addressing this!

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
```diff
@@ -1198,7 +1198,6 @@ - (void)testFailMatchesShortLinkFormatForCustomDomains {
 - (void)testPassMatchesShortLinkFormatForCustomDomains {
   NSArray<NSString *> *urlStrings = @[
     @"https://google.com/xyz", @"https://google.com/xyz/?link=https://somedomain",
-    @"https://google.com?link=https://somedomain", @"https://google.com/?link=https://somedomain",
```
paulb777, line 1201: From the description I thought we were missing a type of link.  Why is a test being deleted versus adding an additional pattern?

eldhosembabu: The link that we are testing here is a valid FDL link, but not a short link. I added this one when i did re writing the logic but its actually a long link. When I added back the missed regex validation, this test case is caught. So removing it now.

FirebaseDynamicLinks/Sources/Utilities/FDLUtilities.m
eldhosembabu: Based on my understanding, this regex is checking for a "/" and should not be having anymore succeeding "/"'s. I'll add it as a comment there.

eldhosembabu: Done

paulb777 (approved):

ncooke3 (approved):

PR #9007, opened 2021-11-24
Swift Error Enum for Storage

paulb777:
Initial thoughts about a Swift Error enum for Firebase Storage.

Thoughts about this approach and recommendations for next steps appreciated.

This implementation takes advantage that most of the APIs that return error codes also have a Result API so that the Result implementation can map the Objective C error to a Swift error.

When implementing the tests, I realized that the auto-generated async APIs hide the Result version so the strategy does not work for both supporting async/await and Swift error codes.

Swift error codes may need to wait until we can do a breaking release and fully wrap and hide all of the Objective C APIs in favor of a purely Swift API.

### DIFF
--- a/FirebaseStorageSwift/Sources/StorageError.swift
+++ b/FirebaseStorageSwift/Sources/StorageError.swift
@@ -16,34 +16,78 @@ import FirebaseStorage
 
 public enum StorageError: Error {
   case unknown
-  case objectNotFound
-  case bucketNotFound
-  case projectNotFound
-  case quotaExceeded
+  case objectNotFound(String)
+  case bucketNotFound(String)
+  case projectNotFound(String)
+  case quotaExceeded(String)
   case unauthenticated
-  case unauthorized
+  case unauthorized(String, String)
   case retryLimitExceeded
   case nonMatchingChecksum
-  case downloadSizeExceeded
+  case downloadSizeExceeded(Int64, Int64)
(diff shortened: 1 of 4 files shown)


paulb777:
@ryanwilson Let me know what you think before I write up an API proposal

FirebaseStorageSwift/Sources/StorageError.swift
```diff
@@ -0,0 +1,49 @@
...
+  case projectNotFound
+  case quotaExceeded
+  case unauthenticated
+  case unauthorized
+  case retryLimitExceeded
+  case nonMatchingChecksum
+  case downloadSizeExceeded
+  case cancelled
+  case invalidArgument
+  case internalError
```
ryanwilson, line 30: Are there any of these that contain extra metadata in the `userInfo` dictionary? I could see potentially the `invalidArgument` or something having the argument passed back, or the project / bucket  info, download size max, etc.

paulb777: Yep. Good call. I'm working through an update.

paulb777:
Since it looks like adding Swift error codes will require a breaking change release. I'm going to merge this to a storage-v9 branch and investigate merging FirebaseStorage and FirebaseStorageSwift to a single public library before continuing with the error code implementation.

PR #9012, opened 2021-11-25
[Heartbeat Logging] - Add `RingBuffer` & `WeakContainer` unit tests

ncooke3:
- Add unit tests for generic `RingBuffer` collection and `WeakContainer` box.
Other changes:
- Move `RingBuffer` to its own file
- Make `RingBuffer` indexing logic depend on `count` rather than `capacity`

#no-changelog


HeartbeatLogging/Sources/RingBuffer.swift
```diff
@@ -0,0 +1,61 @@
...
+  mutating func push(_ element: Element) -> Element? {
+    guard circularQueue.count > 0 else {
+      // Do not push if `circularQueue` is a fixed empty array.
+      return nil
+    }
+
+    defer {
+      // Increment index, wrapping around to the start if needed.
+      tailIndex += 1
+      tailIndex %= circularQueue.count
```
paulb777, line 44: What if `count` is 0?

paulb777 (approved):

HeartbeatLogging/Sources/RingBuffer.swift
ncooke3: There is a guard statement at the top of the `push` API to ensure that count is greater than 0 by the time this block is reached.

HeartbeatLogging/Tests/Unit/RingBufferTests.swift
```diff
@@ -0,0 +1,95 @@
...
+  func testPushStressTest() throws {
+    // Given
+    var ringBuffer = RingBuffer<Int>(capacity: 10)
+    // When
+    for index in 1 ... 100 {
+      ringBuffer.push(index)
+    }
+    // Then
+    XCTAssertTrue(
+      ringBuffer.elementsEqual([91, 92, 93, 94, 95, 96, 97, 98, 99, 100]),
```
maksymmalyhin, line 91: Optional: it may be `ringBuffer.elementsEqual(Array(91...100))`.

maksymmalyhin (approved):
👍 LGTM

HeartbeatLogging/Sources/RingBuffer.swift
paulb777: Just for my education, since I see there is a test for this - but I thought that defer blocks execute even after a guard? - see https://www.hackingwithswift.com/new-syntax-swift-2-defer

ncooke3: Ah I see, good point. I think the difference here is that the `defer` block is defined after the `guard` statement. In the article above, the `defer` is defined before the `guard` statements.

ncooke3: So when `count` is 0. The control flow exits in the guard before the defer block can ever be scheduled. 

To test it out, I ran the `testPush_WhenCapacityIsZero_DoesNothing` with a breakpoint in the defer block and the breakpoint was never triggered.

PR #9015, opened 2021-11-25
[Heartbeat Logging] - Integration tests, documentation, refactoring

ncooke3:
- Add integration tests
- Add documentation 
- Some minor refactoring for better organization
#no-changelog


HeartbeatLogging/Tests/Unit/StorageFactoryTests.swift
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 Google LLC
```
ncooke3, line 1: This PR is quite large so I'm deferring this to the next one.

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -48,29 +61,50 @@ final class HeartbeatStorage: HeartbeatStorageProtocol {
...
+
   deinit {
     // Removes the instance if it was cached.
     Self.cachedInstances.removeValue(forKey: id)
   }
 
   // MARK: - HeartbeatStorageProtocol
 
+  /// Asynchronously reads from and writes to storage using the given transform block.
+  /// - Parameter transform: A block to transform `HeartbeatInfo?` to `HeartbeatInfo?`.
```
paulb777, line 107: transform `HeartbeatInfo?` to `HeartbeatInfo?` ?

paulb777 (approved):
Generally LGTM at a high level. If you'd like to go deeper, I'm find to do it pairing in a GVC.

HeartbeatLogging/Sources/HeartbeatStorage.swift
ncooke3: Take a look at this [piece of documentation](https://github.com/firebase/firebase-ios-sdk/blob/60a39cb6cea4daceada1a50566f1226b4e7bf2d1/HeartbeatLogging/Sources/HeartbeatStorage.swift#L17-L23) which I added in this PR to explain transformation. 

PTAL and LMK what you think. Im open to suggestions

ncooke3: If you agree with the idea behind the `HeartbeatInfo?` → `HeartbeatInfo?` semantics, then maybe this doc could instead read something like:
```suggestion
  /// - Parameter transform: A block that transforms the current stored value to a new value.
```

paulb777: Maybe I'm missing too much context, but when I see the word transform, I expect to see an explanation of the before state and the after state. Instead I see two exactly matching strings and nothing that describes what was transformed or why it was transformed. And the documentation link talks about managing nil, but not about the transformation that I can tell.

ncooke3: I see. I'll schedule some then to discuss because I'd like the intent to be clear.

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -73,16 +96,27 @@ final class FileStorage: Storage {
...
+  /// The key mapping to the object's associated resource in `defaults`.
   private let key: String
 
+  /// Designated initializer.
+  /// - Parameters:
+  ///   - defaults: The defaults container. Defaults to the `standard` instance if `nil`.
+  ///   - key: The key mapping to the value stored in the defaults container.
   init(defaults: UserDefaults? = nil,
        key: String) {
     self.defaults = defaults ?? .standard
```
ryanwilson, line 112: I'm thinking maybe we should never use `.standard` by default since it just opens up room for errors?

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -91,9 +125,14 @@ final class UserDefaultsStorage: Storage {
...
+  ///
+  /// When the given `data` is `nil`, this object's associated defaults is emptied.
+  ///
+  /// - Parameter data: The `Data?` to write to this object's associated defaults.
+  func write(_ data: Data?) throws {
+    if let data = data {
+      defaults.set(data, forKey: key)
     } else {
       let emptyData = Data()
       defaults.set(emptyData, forKey: key)
```
ryanwilson, line 138: Hmm, I think you can set `nil` on user defaults now and it will wipe the key instead of having empty data written. Is there a reason why you chose to write data instead of clear it?

HeartbeatLogging/Sources/StorageFactory.swift
```diff
@@ -16,37 +16,45 @@ import Foundation
...
 /// A factory type for `Storage`.
 protocol StorageFactory {
-  static func makeStorage(id: String) -> Self
+  static func makeStorage(id: String) -> Storage
 }
 
 // MARK: - FileStorage + StorageFactory
 
+/// The name of the file system directory where heartbeat data is stored.
+let kHeartbeatFileStorageDirectoryPath = "google-heartbeat-storage"
```
ryanwilson, line 25: From the Google Swift style guide https://google.github.io/swift/#global-constants we shouldn't be using a `k` prefix. I'm a fan of having a private `Constants` struct that has constants for this file so it's obvious at the call site where the variable is coming from. WDYT?

HeartbeatLogging/Sources/StorageFactory.swift
```diff
@@ -16,37 +16,45 @@ import Foundation
...
 
-// MARK: - FileManager + Extension
+// MARK: - UserDefaultsStorage + StorageFactory
 
-private extension FileManager {
-  var applicationSupportDirectory: URL {
-    // TODO: The below bang! should be safe but re-evaluate.
-    urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
+/// The name of the user defaults suite where heartbeat data is stored.
+let kHeartbeatUserDefaultsSuiteName = "com.google.heartbeat.storage"
```
ryanwilson, line 51: Nit: same message about the constant. It could be grouped with the above one

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,297 @@
...
   override func tearDownWithError() throws {
-    // Put teardown code here. This method is called after the
-    // invocation of each test method in the class.
+    try removeUnderlyingHeartbeatStorageContainers()
+  }
+
+  /// This test may flake if it is executed during the transtion from one day to the next.
+  func testLogAndFlush() throws {
+    // Given
+    let heartbeatController = HeartbeatController(id: #function)
```
ryanwilson, line 33: This uses the shared storage, right? It may run into issues with other tests. It probably makes sense to use the initializer that passes in the custom storage that won't conflict with any other functions.

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,297 @@
...
+    try removeUnderlyingHeartbeatStorageContainers()
   }
 
   override func tearDownWithError() throws {
-    // Put teardown code here. This method is called after the
-    // invocation of each test method in the class.
+    try removeUnderlyingHeartbeatStorageContainers()
+  }
+
+  /// This test may flake if it is executed during the transtion from one day to the next.
```
ryanwilson, line 30: Can we use the `dateProvider` to inject the exact same date?

HeartbeatLogging/Sources/AnyEncoder.swift
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// A type that can encode an `Encodable` object into `Data`.
+protocol AnyEncoder {
+  func encode<T>(_ value: T) throws -> Data where T: Encodable
+}
```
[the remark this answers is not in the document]
ncooke3: So the only reason I did this was to inject a decoder (or encoder for `AnyEncoder`) into a `HeartbeatStorage` instance could easily be faked. https://github.com/firebase/firebase-ios-sdk/blob/6f1c4eb1187a9056ac9cc07a4f98102cab1467c2/HeartbeatLogging/Sources/HeartbeatStorage.swift#L51-L54

Here's a test doing so: https://github.com/firebase/firebase-ios-sdk/blob/6f1c4eb1187a9056ac9cc07a4f98102cab1467c2/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift#L133-L166

We spoke briefly offline about it, this may be a lot of extra code just to improve testability for a code path that we should be able to verify most of the correctness at compile time:
https://github.com/firebase/firebase-ios-sdk/blob/6f1c4eb1187a9056ac9cc07a4f98102cab1467c2/HeartbeatLogging/Sources/HeartbeatStorage.swift#L139-L143

I'm ok with removing the `Any*` types and cleaning up the tests that tested this behavior.

WDYT?

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -14,92 +14,104 @@
...
+    self.init(id: id, dateProvider: Date.init)
+  }
+
+  /// Convenience initializer. Mirrors the semantics of the public intializer with the added benefit of
+  /// injecting a custom date provider for improved testability.
+  /// - Parameters:
+  ///   - id: The id to associate this controller's heartbeat storage with.
+  ///   - dateProvider: A date provider.
+  convenience init(id: String,
+                   dateProvider: @escaping () -> Date) {
```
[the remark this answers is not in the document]
ncooke3: > I wonder where this should be used compared to the one below

So for integration tests, my understanding was I should use the public API. I felt I couldnt test important stuff though (like logging heartbeats over a period of time) by using the public initializer alone because it does not allowing injecting a `dateProvider`. 

So to compromise, I created this internal initializer that the **public init calls directly into** and **allows for injecting a `dateProvider`**. 

So, to answer your question, this initializer should be used in integration tests.

> since from a testability standpoint this is accessing a singleton (using shared storage) and could conflict with other tests.

That's correct. To avoid this, I use pass `#function` as the `id` in test methods to ensure they are getting an instance that is specific to the specific test case. 

https://github.com/firebase/firebase-ios-sdk/blob/6f1c4eb1187a9056ac9cc07a4f98102cab1467c2/HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift#L68-L71


LMK what you think and I'm very open to feedback on this approach. So far, it did come in handy for writing a test like this:
https://github.com/firebase/firebase-ios-sdk/blob/6f1c4eb1187a9056ac9cc07a4f98102cab1467c2/HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift#L131-L206

HeartbeatLogging/Sources/AnyEncoder.swift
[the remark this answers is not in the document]
ryanwilson: Thanks for sharing! For decoding, maybe we could avoid it by having it return `Data` that doesn't match the expected format? Even something like `"BAD_DATA".data(using: .utf8)`

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -14,92 +14,104 @@
...
     self.dateProvider = { Self.dateStandardizer(dateProvider()) }
   }
 
-  /// Asynchronously log a new heartbeat, if needed.
+  /// Asynchronously logs a new heartbeat, if needed.
   ///
   /// - Note: This API is thread-safe.
   ///
-  /// - Parameter agent: A `String` identifier to associate a new heartbeat with.
+  /// - Parameter agent: The string agent to associate the logged heartbeat with.
```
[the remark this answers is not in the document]
ncooke3: Yes, I can expand it. TBH, I wasn't sure how specific I should get so this is good feedback!

> is it the identifiers like `fire-analytics`

In the context of how this API will be wrapped by FirebaseCore, the Firebase User Agent will be injected each time.

Here's what the integration will look like:

```objc
// Firebase Core provide this API to product SDKs

// FIRHeartbeatLogger.m
- (void)log {
  NSString *firebaseUserAgent = FirebaseUserAgent();
  [_heartbeatController logWithAgent:firebaseUserAgent];
}

// Clients that want to log Firebase heartbeats use  Firebase Core's API like:
[heartbeatLogger log];
```

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -14,92 +14,104 @@
...
   public func log(_ agent: String) {
     let date = dateProvider()
-    let capacity = limit
+    let capacity = heartbeatsStorageCapacity
 
     storage.readAndWriteAsync { heartbeatInfo in
       var heartbeatInfo = heartbeatInfo ?? HeartbeatInfo(capacity: capacity)
 
+      // Filter for the time periods where the last heartbeat to be logged for
+      // that time period was logged more than one time period (i.e. day) ago.
```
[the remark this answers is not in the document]
ncooke3: The `cache` probably needs more descriptive name. I like that comment structure. I will work on it.

HeartbeatLogging/Sources/HeartbeatInfo.swift
```diff
@@ -14,40 +14,40 @@
...
-/// A type that can be represented as a `HeartbeatsPayload`.
+/// A type that can be converted to a `HeartbeatsPayload`.
 protocol HeartbeatsPayloadConvertible {
   func makeHeartbeatsPayload() -> HeartbeatsPayload
 }
 
-/// <#Description#>
+/// A codable collection of heartbeats that has a fixed capacity and optimizations for storing heartbeats of
+/// multiple time periods.
 struct HeartbeatInfo: Codable, HeartbeatsPayloadConvertible {
```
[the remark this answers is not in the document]
ncooke3: Absolutely! Some ideas:
- HeartbeatsBundle
- HeartbeatsBuffer
- Heartbeats

HeartbeatLogging/Sources/Storage.swift
ncooke3: Good idea. I'm not sure how to maneuver around the fact that `UserDefaults`'s[`init?(suiteName suitename: String?)`](https://developer.apple.com/documentation/foundation/userdefaults/1409957-init) returns an optional instance. 

I guess it will be up to clients of this class to unwrap the defaults they pass in.

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
ncooke3: Yes, it uses shared storage, but because shared storage is keyed by the `id` used to create the controller instance. We are safe from conflict here because we pass in a unique `id` specific to this test case: `#function`

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
ncooke3: We could. The reason I didn't here was to make sure I had at least one integration test that used the public initializer (and defaulted to using `Date.init` as the date provider). 

I realized there was a small potential of a flake in doing so this these tests would run on CI and added the comment to explain why.

WDYT? I'm cool with injecting a `dateProvider` to keep things controlled but am curious about your thoughts on  my thinking above.

HeartbeatLogging/Sources/Storage.swift
ncooke3: > Is there a reason why you chose to write data instead of clear it?

The `Storage` protocol is implemented by `UserDefaultsStorage` and `FileStorage`. I was not sure how similar each conforming type should behave. From the conversations in https://github.com/firebase/firebase-ios-sdk/pull/8994, I made `FileStorage`'s `write(nil)` empty the file rather than remove it. When implementing `write(nil)` for `UserDefaultsStorage`, I was not sure if it should follow the similar behavior (empty the value vs removing the key). I chose writing `emptyData` to keep the behavior the same across both types.

I'm curious to know what you think though. I'm unsure if there is a best practice here.

HeartbeatLogging/Sources/HeartbeatsPayload.swift
```diff
@@ -14,57 +14,101 @@
...
     let encoder = JSONEncoder()
     encoder.dateEncodingStrategy = .formatted(Self.dateFormatter)
 
     if let data = try? encoded(using: encoder) {
+      // TODO: Compress the payload before base64 encoding.
+      // TODO: Confirm in e2e test if base64 URL safe string is needed.
       return data.base64EncodedString()
     } else {
       return "" // Return empty string if encoding failed.
     }
```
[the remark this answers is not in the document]
ncooke3: Reworked in 56ffaae

HeartbeatLogging/Sources/Storage.swift
ncooke3: Reworked in 29cf736

HeartbeatLogging/Sources/StorageFactory.swift
ncooke3: Reworked in 29cf736

HeartbeatLogging/Sources/StorageFactory.swift
ncooke3: Reworked in 29cf736

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -58,12 +76,17 @@ final class FileStorage: Storage {
     }
   }
 
+  /// Creates all directories in the given file system URL.
+  ///
+  /// If the directory for the given URL already exists,  the error is ignored because the directory
```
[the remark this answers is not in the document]
ncooke3: Reworked in 29cf736

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -14,28 +14,41 @@
 
 import Foundation
 
+/// A type that reads from and writes to an underlying storage container.
 protocol Storage {
   func read() throws -> Data
-  func write(_ value: Data?) throws
+  func write(_ data: Data?) throws
```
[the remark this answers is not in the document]
ncooke3: Reworked in 29cf736

HeartbeatLogging/Sources/HeartbeatInfo.swift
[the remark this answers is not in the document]
paulb777: +1 for `HeartbeatsBundle` or `HeartbeatBundle`

HeartbeatLogging/Sources/HeartbeatStorage.swift
```diff
@@ -14,31 +14,44 @@
...
 import Foundation
 
+/// A typealias for a function that transforms a `HeartbeatInfo?` into `HeartbeatInfo?`. This
+/// typealias is used in `HeartbeatStorageProtocol` to model the transformation of the contents
+/// of heartbeat storage.
+///
+/// The parameter is marked as optional as the heartbeat info provided to the closure may be `nil` if
+/// an error occurred or storage is empty. The return type is marked as optional because the storage can
+/// be emptied by returning `nil` from the transformation closure.
 typealias HeartbeatInfoTransform = (HeartbeatInfo?) -> HeartbeatInfo?
```
paulb777, line 24: Consider eliminating the typealias as we discussed.

HeartbeatLogging/Sources/HeartbeatController.swift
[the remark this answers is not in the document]
ncooke3: How does ⬇️  sound?
> The cache contains the last added date for each time period. We only need to cache the date since it's the only piece of information used to determine whether or not to append a new heartbeat. 

The cache can be used by clients (as done here) to decide what time periods need a new heartbeat. 

WDYT is most helpful? 
- renaming `cache` to `lastAddedHeartbeatDateByTimePeriod` (I like the conciseness of `cache`, esp. in the `HeartbeatInfo` implementation.)
- Adding additional documentation here
- Adding additional documentation to the `cache` property declaration in `HeartbeatInfo`

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -14,92 +14,104 @@
...
-  /// Synchronously flushes heartbeats from storage.
+  /// Synchronously flushes heartbeats from storage into a heartbeats payload.
   ///
   /// - Note: This API is thread-safe.
   ///
   /// - Returns: The flushed heartbeats in the form of `HeartbeatsPayload`.
   @discardableResult
   public func flush() -> HeartbeatsPayload {
-    let capacity = limit
+    let capacity = heartbeatsStorageCapacity
```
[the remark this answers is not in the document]
ncooke3: Reworked in 4b1ef3e

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -14,92 +14,104 @@
...
+  /// Asynchronously logs a new heartbeat, if needed.
   ///
   /// - Note: This API is thread-safe.
   ///
-  /// - Parameter agent: A `String` identifier to associate a new heartbeat with.
+  /// - Parameter agent: The string agent to associate the logged heartbeat with.
   public func log(_ agent: String) {
     let date = dateProvider()
-    let capacity = limit
+    let capacity = heartbeatsStorageCapacity
```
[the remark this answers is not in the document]
ncooke3: Reworked in 4b1ef3e

HeartbeatLogging/Sources/HeartbeatStorage.swift
ncooke3: Reworked in 861028b. Im going to followup with some other improvements that should further simplify things

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,307 @@
...
+    let heartbeatsDirectoryURL = FileManager.default
+      .applicationSupportDirectory
+      .appendingPathComponent(
+        Constants.heartbeatsDirectoryPath, isDirectory: true
+      )
+    do {
+      try FileManager.default.removeItem(at: heartbeatsDirectoryURL)
+    } catch CocoaError.fileNoSuchFile {
+      // Do nothing.
+    }
```
ryanwilson, line 317: Nice _catch_!

HeartbeatLogging/Tests/Unit/HeartbeatControllerTests.swift
```diff
@@ -169,63 +164,176 @@ class HeartbeatControllerTests: XCTestCase {
...
+
+    let heartbeatController = HeartbeatController(
+      storage: HeartbeatStorageFake(),
+      dateProvider: { testDate }
+    )
+
+    // When
+    heartbeatController.log("dummy_agent")
+
+    // Device travels from NYC to Tokyo.
```
ryanwilson, line 245: 😍

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
```diff
@@ -87,201 +87,218 @@ class HeartbeatStorageTests: XCTestCase {
...
       expectation.fulfill()
-      return heartbeatInfo
+      throw StorageError.writeError
     }
 
+    // Then
+    XCTAssertThrowsError(
+      try heartbeatStorage.getAndSet { heartbeatInfo in
+        heartbeatInfo
+      }
```
ryanwilson, line 251: Nit / in case you wanted to make this shorter, you can use `try heartbeatStorage.getAndSet { $0 }` to do the same thing (using implicitly named arguments). Totally fair if you wanted to be explicit here, but just in case you weren't aware!

ryanwilson (commented):
Sorry for the delay, wanted to take a close look at the whole thing. Thanks for the amazingly thorough tests!

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,307 @@
...
+      // to achieve a payload with multiple user agent groupings.
+      if day < 5 {
+        heartbeatController.log("dummy_agent_1")
+      } else if day < 13 {
+        heartbeatController.log("dummy_agent_2")
+      } else {
+        heartbeatController.log("dummy_agent_3")
+      }
+
+      testdate.addTimeInterval(60 * 60 * 24) // Advance the test date by 1 day.
```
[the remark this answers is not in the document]
ncooke3: No, it shouldn't because the whatever is passed in as the `dateProvider` is "standardized" with a configured `Calendar`.  

https://github.com/firebase/firebase-ios-sdk/blob/10410ec6cc5763fe55dfe941f366008b7abb3952/HeartbeatLogging/Sources/HeartbeatController.swift#L52-L60

The standardizer is statically configured at the top of class: 

https://github.com/firebase/firebase-ios-sdk/blob/10410ec6cc5763fe55dfe941f366008b7abb3952/HeartbeatLogging/Sources/HeartbeatController.swift#L26-L33

So this ensures a standardized date is ultimately recorded that isn't biased by factors like daylight savings.

FWIW, when I first wrote this test, it did fail due to DLS because Nov 7 (i think?) was daylight savings time so the payload string one day off and i couldn't figure out why for the life of me. It was because the `HeartbeatController` date was not standardized to ignore factors like DSL so adding `60 * 60 * 24` resulted in the same day on the day of DLS.

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,307 @@
...
+    try removeUnderlyingHeartbeatStorageContainers()
   }
 
   override func tearDownWithError() throws {
-    // Put teardown code here. This method is called after the
-    // invocation of each test method in the class.
+    try removeUnderlyingHeartbeatStorageContainers()
+  }
+
+  /// This test may flake if it is executed during the transition from one day to the next.
```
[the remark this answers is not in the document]
ncooke3: Reworked in fd199870e

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -13,16 +13,307 @@
 // limitations under the License.
 
 import XCTest
-import HeartbeatLogging
+@testable import HeartbeatLogging
+
+private enum Constants {
+  static let heartbeatsDirectoryPath = "google-heartbeat-storage"
+  static let heartbeatsSuiteName = "com.google.heartbeat.storage"
```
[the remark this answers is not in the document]
ncooke3: > at least one test fails when these values get out of sync with the other private enum, right?

Correct, this will catch hardcoded storage location regressions. 

Reworked in 66719b75d

HeartbeatLogging/Sources/Storage.swift
```diff
@@ -14,28 +14,48 @@
...
   private let url: URL
+  /// The file manager used to perform file system operations.
   private let fileManager: FileManager
 
+  /// Designated initializer.
+  /// - Parameters:
+  ///   - url: A file system URL for the underlying file resource.
+  ///   - fileManager: A file manager. Defaults to `default` manager.
   init(url: URL,
        fileManager: FileManager = .default) {
```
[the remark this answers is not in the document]
ncooke3: Reworked in 1aec84937

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
ncooke3: Reworked in fd19987

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: Reworked in 0c827b651

HeartbeatLogging/Sources/HeartbeatInfo.swift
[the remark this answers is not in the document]
ncooke3: Great point on the extra `s`. I'll rename `HeartbeatsBundle`. Thanks team.

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: Reworked in 0c827b6

HeartbeatLogging/Sources/HeartbeatController.swift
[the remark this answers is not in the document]
ncooke3: Great, I'll keep it as is. I have a test that does the exact thing you describe to verify this behavior. Injecting the date prevents risk of flake here so we can test the instance management logic.

https://github.com/firebase/firebase-ios-sdk/blob/0c827b651262bb0fafc9ba8c14190659cd7bb9a2/HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift#L73-L96

HeartbeatLogging/Sources/HeartbeatInfo.swift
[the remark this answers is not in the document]
ncooke3: Reworked in bc0699f81 and 7291791e3

HeartbeatLogging/Sources/HeartbeatController.swift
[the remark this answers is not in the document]
ncooke3: I added the comment to the `cache` property declaration. I think the hover doc provides a nice glimpse into what it is/does:

Screen Shot 2021-12-08 at 5 26 42 PM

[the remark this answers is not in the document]
ncooke3: Thanks for the naming suggestions! Heh, I went back and forth between a the concise `cache` vs a more self-documenting name. PTAL at commit abbaa4f18 and lmk what you think.

HeartbeatLogging/Sources/Storage.swift
ncooke3: FWIW: In terms of how the library uses `Storage`, it does not matter whether `write(nil)` removes the object vs. writing empty data.

paulb777: My mental model for UsersDefaults is a Dictionary in which a key with no value does not makes sense.

My mapping for writing nil to a file system is empty versus a non-existent file which would be created with a remove operation.

I don't feel too strongly though and fine either way, especially since it doesn't seem to make a practical difference for this use case.

ncooke3: Thanks team, it was very helpful to hear how y'all interpreted this. I've reworked in c0b7dd969

To summarize the behavior now:

    writing `nil` to:
      - user defaults → removes default
      - file system → empties file

ncooke3:
Hi @ryanwilson, thanks for the second pass! Could you please take one more look at the new changes to make sure you're cool with how everything was addressed?

ryanwilson (approved):
Thanks for working through all the comments, I love how thorough the testing and comments are!

PR #9023, opened 2021-11-27
Update for GoogleUtilitiesComponents 1.1.0 release

### LINKED ISSUE #9021: Need to publish new podspec for GoogleUtilitiesComponents
The latest podspec published to Cocoapods contains an IOS platform version of 8:
https://cocoapods.org/pods/GoogleUtilitiesComponents
https://github.com/CocoaPods/Specs/blob/master/Specs/e/a/f/OpenCV2/4.3.0/OpenCV2.podspec.json

However, this repo source code has contained version 10 since last year:
https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleUtilitiesComponents.podspec
This update needs to be published. Version 8 is just so old, and breaks other tooling that has higher minimum versions, e.g., cordova.

paulb777:
Fix #9021 

Bump version to 1.1.0 to update published minimum iOS version from 8.0 to 10.0.

Restore s.static_framework = true, so that the library is an exact match to 1.0.0 except for minimum iOS version.

This should technically be a major version update, but doing a minor version since the only CocoaPods dependent, MLKitCommon, which has always required iOS 10.0 and the minor update will get the fix with publishing only this pod and only a `pod update` from users.

Checks:
`git diff UtilitiesComponents-1.0.0 -- GoogleUtilitiesComponents`
` git diff UtilitiesComponents-1.0.0 -- GoogleUtilitiesComponents.podspec`
`~/.cocoapods/repos/cocoapods (master) $ git grep GoogleUtilitiesComponents`

### DIFF
--- a/GoogleUtilitiesComponents.podspec
+++ b/GoogleUtilitiesComponents.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'GoogleUtilitiesComponents'
-  s.version          = '1.0.0'
+  s.version          = '1.1.0'
   s.summary          = 'Google Utilities Component Container for Apple platforms.'
 
   s.description      = <<-DESC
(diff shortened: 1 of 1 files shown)


paulb777:
cc: @bcdj

ryanwilson (approved):

PR #9028, opened 2021-11-29
Swift Modernization Dashboard

paulb777:
Reviewers: See https://github.com/firebase/firebase-ios-sdk/blob/pb-markdown/SwiftDashboard.md

#no-changelog


SwiftDashboard.md
```diff
@@ -0,0 +1,56 @@
...
+| **Swift Library**     | ❌   |   ✔   | ❌     |❌     | ❌     | ❌     | ❌      |  ✔     |  ✔    | 1      |  ✔     | ❌    | ❌     | ✔      | ❌     | ❌    | ✔     |
+| **API Tests**         | ❌   |  ❌    |  ✔    |❌     | ❌     | ✔       | ❌     | 3      | 2     |  ✔     | 2      | ❌     | ❌     | 2      | ❌    |  ✔     | ✔    |
+| **async/await**       | ❌   |  ❌    |  ✔    |❌     | ❌     |  ✔      | ❌     | 3     | ❌     | ❌     | ❌     | ❌    | ❌     | ❌    | ❌     |  ✔    | ✔    |
+| **Swift Errors**      |  ❌  |  ❌    | ❌    |❌     | 4      | ❌     | ❌     | ❌     | ❌    | ❌     | ❌     | ❌    | ❌     | ✔      | ❌     | ❌   | 5   |
+| **Codable**           | n/a  | n/a     | n/a   |n/a    | n/a     | n/a    |n/a     |  ✔     |  ✔     | 1      | n/a     | n/a   | ❌     | n/a    | n/a    | n/a   |n/a   |
+| **SwiftUI Lifecycle** | n/a  |  ❌    | n/a    |❌     | ❌     | n/a    |n/a     | n/a    | n/a    | n/a     | n/a    | n/a   | ❌     | n/a    | n/a    | n/a   |n/a  |
+| **Property Wrappers** |  ❌  |  ❌    | ❌    |❌     | ❌     | ❌     | ❌     | ❌     | ✔     | ❌     | ❌     | ❌    | ❌     | ❌    | ❌     | ❌   |❌     |
```
ryanwilson, line 15: Thanks for putting this together! I'm torn about having a specific row for Property Wrappers since it's an implementation detail and may not (likely doesn't?) make sense for all products. For example, Functions may not make sense to have any property wrappers but feels wrong to have an X or even N/A there.

I sort of think of it in the same way as a row for enums with associated types, we'd love to add them where the API makes sense but shouldn't add them just because we can.

SwiftDashboard.md
```diff
@@ -0,0 +1,56 @@
+# Firebase Swift Modernization Dashboard
+
+This dashboard summarizes the status of Firebase's [2022 Swift Modernization Project](Roadmap.md).
+Please upvote or create a [feature request](https://github.com/firebase/firebase-ios-sdk/issues)
+to help prioritize any particular cell(s).
+
+|                       | AB  | An     | ApC    | ApD    | Aut    | Cor    | Crs    | DB     | Fst    | Fn     | IAM    | Ins    | Msg    | MLM    | Prf    | RC     |    Str |
```
ryanwilson, line 7: Two other things I can think of that may make sense to track:

1.  SwiftUI improvements/support/enhancements
  - Specifically where there is some element of UI, or dealing with UIViewControllers. This could also encompass some Combine work (publishers can fit in nicely with SwiftUI)

2. Swift Generics
  - This sort of contradicts my other comment... but if we have APIs that are using weakly typed information we should try to use proper generics. Not sure if it's worth having a row for this though.

SwiftDashboard.md
```diff
@@ -0,0 +1,56 @@
...
+| **API Tests**         | ❌   |  ❌    |  ✔    |❌     | ❌     | ✔       | ❌     | 3      | 2     |  ✔     | 2      | ❌     | ❌     | 2      | ❌    |  ✔     | ✔    |
+| **async/await**       | ❌   |  ❌    |  ✔    |❌     | ❌     |  ✔      | ❌     | 3     | ❌     | ❌     | ❌     | ❌    | ❌     | ❌    | ❌     |  ✔    | ✔    |
+| **Swift Errors**      |  ❌  |  ❌    | ❌    |❌     | 4      | ❌     | ❌     | ❌     | ❌    | ❌     | ❌     | ❌    | ❌     | ✔      | ❌     | ❌   | 5   |
+| **Codable**           | n/a  | n/a     | n/a   |n/a    | n/a     | n/a    |n/a     |  ✔     |  ✔     | 1      | n/a     | n/a   | ❌     | n/a    | n/a    | n/a   |n/a   |
+| **SwiftUI Lifecycle** | n/a  |  ❌    | n/a    |❌     | ❌     | n/a    |n/a     | n/a    | n/a    | n/a     | n/a    | n/a   | ❌     | n/a    | n/a    | n/a   |n/a  |
+| **Property Wrappers** |  ❌  |  ❌    | ❌    |❌     | ❌     | ❌     | ❌     | ❌     | ✔     | ❌     | ❌     | ❌    | ❌     | ❌    | ❌     | ❌   |❌     |
+
+### Other Projects
+- Automatic Swift API generation and coverage validation
```
ryanwilson, line 18: Can we clarify that this is more surfacing the Objective-C translated Swift APIs in PRs and validating coverage of them? I read this more as generating Swift APIs automatically which isn't what we're doing (unless I missed something)

SwiftDashboard.md
paulb777: I think that property wrappers might be similar to Codable. We should think about opportunities to add them and then go to either n/a or ❌  /✔

SwiftDashboard.md
paulb777: Done. PTAL

SwiftDashboard.md
paulb777: For 1, can you give an example or two.  Is there a more up to date reference than https://docs.google.com/document/d/1UrC4eqwSLG00nSH6V5ttwTkcQt5ULQKkKuiZbtsAGGA/edit?resourcekey=0-sEstblfahgcurSGOzrteXw#heading=h.hdi9c76jzlbq?

For 2, do you have any examples in mind?  If it's just one or two, it might be better in the Other Projects section.

SwiftDashboard.md
```diff
@@ -0,0 +1,57 @@
+# Firebase Swift Modernization Dashboard
+
+This dashboard summarizes the status of Firebase's [2022 Swift Modernization Project](Roadmap.md).
+Please upvote or create a [feature request](https://github.com/firebase/firebase-ios-sdk/issues)
+to help prioritize any particular cell(s).
+
+|                       | AB  | An     | ApC    | ApD    | Aut    | Cor    | Crs    | DB     | Fst    | Fn     | IAM    | Ins    | Msg    | MLM    | Prf    | RC     |    Str |
+|   :---                | :--- | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: |
+| **Swift Library**     | ❌   |   ✔   | ❌     |❌     | ❌     | ❌     | ❌      |  ✔     |  ✔    | 1      |  ✔     | ❌    | ❌     | ✔      | ❌     | ❌    | ✔     |
+| **API Tests**         | ❌   |  ❌    |  ✔    |❌     | ❌     | ✔       | ❌     | 3      | 2     |  ✔     | 2      | ❌     | ❌     | 2      | ❌    |  ✔     | ✔    |
```
ncooke3, line 10: Installations can have a check for API Tests: Installations API build tests (#8353)

ncooke3 (commented):
Thanks Paul- this is a great way to track everything!

SwiftDashboard.md
paulb777: Good catch. Thanks! Fixed.

SwiftDashboard.md
ryanwilson: Sounds reasonable to me, thanks.

SwiftDashboard.md
ryanwilson: 1: At a quick glance AppDistro, Auth, FIAM, Perf, and Analytics still have APIs that include UIViewControllers (or implementations that depend on them) which should have an associated API or implementation that work with SwiftUI.  This will overlap with Property Wrappers and likely the SwiftUI lifecycle bits, but an audit and improvements could likely be made. The existing FIAM and Analytics View modifier APIs would fit into this category.

2. Other projects sounds like a good place for this, good call.

mortenbekditlevsen:
I see async/await, but would AsyncSequence or AsyncStream support also make sense for some of the frameworks?

paulb777:
@mortenbekditlevsen Thanks. We should consider AsyncSequence and AsyncStream support, but so far, my feeling is that those will generally wait for a phase 2.  A general goal with this dashboard is to help track an initial full Swift review of Firebase along with addressing low-hanging fruit. I would expect it to identify several follow up tasks including AsyncSequence and AsyncStream. I'll make this clear in the introduction.

mortenbekditlevsen:
I can add these as pitches or feature requests or even try assisting with the implementation, but off of the top of my mind, here are a few things that could be implemented:

Firestore:
A `PublishedFirestoreQuery` property wrapper that is similar to `FirestoreQuery` in all respects except that it is a `DynamicProperty` class that publishes will change notifications to the default implementation `objectWillChange` of `ObservableObject`.

RTDB:
Two property wrappers similar to `FirestoreQuery` and `PublishedFirestoreQuery`

For both Firestore and RTDB:
Explore if the Swift type system can help with query creation for Firestore and RTDB by avoiding invalid query composition.

I did a small spike, and something is definitely possible. The generic types used to make this happen do however leak out of the implementation. This might not matter, but it may - depending on taste:

https://gist.github.com/mortenbekditlevsen/9cd48ca0b0d2b1d4ceda9332b6645a3a

paulb777:
@mortenbekditlevsen Good ideas! Feature requests are probably the best way to track those. The dashboard is primarily for us to track the list of items we want to consider for all Firebase products. As we get into specific property wrappers, we can track them individually just with the particular product.

peterfriese:
> @mortenbekditlevsen Good ideas! Feature requests are probably the best way to track those. The dashboard is primarily for us to track the list of items we want to consider for all Firebase products. As we get into specific property wrappers, we can track them individually just with the particular product.

On that note - I was just wondering if we should set up some tags to makes it easier to track FRs and match them to the specific columns / rows of the dashboard. We could even link out from the dashboard to the respective tags' page, like so: https://github.com/firebase/firebase-ios-sdk/issues?q=is%3Aissue+is%3Aopen+label%3A%22type%3A+feature+request%22+label%3A%22Swift+API%22 (and now that I see it, maybe the tags we already have in use are already good enough... WDYT?)

paulb777:
Yep the Swift API tag should be used for associated FR's and PR's. I thought about links directly from the cells but concluded that would make the markdown hard to read and maintain - and chose to use the low tech number notes instead.  :)

SwiftDashboard.md
```diff
@@ -4,6 +4,9 @@ This dashboard summarizes the status of Firebase's [2022 Swift Modernization Pro
 Please upvote or create a [feature request](https://github.com/firebase/firebase-ios-sdk/issues)
 to help prioritize any particular cell(s).
 
+This dashboard is intended to track an initial full Swift review of Firebase along with addressing low-hanging fruit. I would expect it to identify additional follow up
```
ryanwilson, line 7: Nit: `s/I would/We/`

ryanwilson (approved):
LGTM with one small nit!

PR #9037, opened 2021-12-01
Investigate binary size changes in 8.10.0

granluo:
This pr is to trigger the `health-metrics-presubmit `workflow to get the size from a certain release.


paulb777:
Closing now that the test is over

PR #9044, opened 2021-12-02
Install Python dependencies before running trigger_workflow script.

jonsimantov:
Updated version of the script requires the "requests" module, which needs to be explicitly installed.


ncooke3 (approved):
👍

PR #9045, opened 2021-12-02
Merge FirebaseStorageSwift and FirebaseStorage

paulb777:
We originally thought that we could move to combined Swift/ObjC libraries with a single combined CocoaPod and a dependent SwiftPM library.  However, that strategy is problematic since it makes usability of public imports different between the two package managers, along with testing and usability complications.

This PR prototypes using the same library structure for both. It also solves the problem of providing Swift error codes for any API that needs them by using `REFINED_FOR_SWIFT` for any applicable API and wrapping the error handling in a Swift specific API. See the `list` API implementation.

I believe we have two unavoidable breakages:

- When we start to depend on Swift libraries, we will require all CocoaPods users to use frameworks (static or dynamic). This is probably fine since most users are probably already there - but will require a breaking release.
- I don't know of a good way to continue to support the Firebase pod and Firebase.h to support APIs only available through Swift.  This will be a big breaking change and we will need to come up with a good migration strategy.

Open Questions:

- Should we use `@_exported import FirebaseStorageObjC` to pass through Objective C APIs or should we explicitly wrap all of them in Swift after marking them as `REFINED_FOR_SWIFT`?
- Can we do a Swift version of a Firebase module with `can_import` to continue to support `import Firebase`?

Next Steps for Storage:

-  Update all APIs that can return an error similarly to `list`


FirebaseStorageSwift/Sources/Result.swift
```diff
@@ -91,9 +91,9 @@ public extension StorageReference {
   func list(maxResults: Int64,
             pageToken: String,
             completion: @escaping (Result<StorageListResult, Error>) -> Void) {
-    list(maxResults: maxResults,
-         pageToken: pageToken,
-         completion: getResultCallback(completion: completion))
+    __list(withMaxResults: maxResults,
```
paulb777, line 94: list implementation for Swift clients that will find the error handling in Result.swift

FirebaseStorage/Sources/Public/FirebaseStorage/FIRStorageReference.h
```diff
@@ -251,9 +251,9 @@ NS_SWIFT_NAME(putData(_:metadata:));
  * prefixes under the current StorageReference.
  */
 - (void)listWithMaxResults:(int64_t)maxResults
-                completion:
-                    (void (^)(FIRStorageListResult *result, NSError *_Nullable error))completion
-    NS_SWIFT_NAME(list(maxResults:completion:));
+                completion:(void (^)(FIRStorageListResult *result,
+                                     NSError *_Nullable error))completion NS_REFINED_FOR_SWIFT;
```
paulb777, line 255: Hide API for Swift

FirebaseStorageSwift/Sources/Result.swift
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import FirebaseStorage
+@_exported import FirebaseStorageObjC
```
paulb777, line 15: Pass Objective C APIS through to Swift.

.github/workflows/storage.yml
```diff
@@ -108,56 +101,28 @@ jobs:
...
       run: scripts/setup_bundler.sh
     - name: Build and test
       run: |
-       scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb ${{ matrix.spec }} --platforms=${{ matrix.target }}
+       scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseStorage.podspec --test-specs=unit --platforms=${{ matrix.target }}
 
   storage-cron-only:
     # Don't run on private repo.
-    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
+#    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
```
ncooke3, line 114: should this still be commented out?

FirebaseCombineSwift/Tests/Unit/Storage/StorageReferenceTests.swift
```diff
@@ -54,16 +54,9 @@ class StorageReferenceTests: XCTestCase {
...
           putFileExpectation.fulfill()
-
-          XCTAssertEqual(error.domain, StorageErrorDomain)
-          XCTAssertEqual(error.code, StorageErrorCode.unknown.rawValue)
-
-          let expectedDescription =
-            "File at URL: \(dummyFileURL.absoluteString) is not reachable. Ensure file URL is not" +
-            " a directory, symbolic link, or invalid url."
-          XCTAssertEqual(error.localizedDescription, expectedDescription)
+          XCTAssertEqual("unknown", "\(String(describing: error))")
```
ncooke3, line 59: nit: i think this can be simplified to
```suggestion
          XCTAssertEqual("unknown", String(describing: error))
```

FirebaseStorage/Sources/Public/FirebaseStorage/FIRStorageReference.h
```diff
@@ -251,9 +251,9 @@ NS_SWIFT_NAME(putData(_:metadata:));
...
+//  NS_SWIFT_NAME(list(maxResults:completion:));
```
ncooke3, line 256: nit: should this be removed now?

FirebaseStorage/Sources/Public/FirebaseStorage/FIRStorageReference.h
```diff
@@ -274,9 +274,9 @@ NS_SWIFT_NAME(putData(_:metadata:));
  */
 - (void)listWithMaxResults:(int64_t)maxResults
                  pageToken:(NSString *)pageToken
-                completion:
-                    (void (^)(FIRStorageListResult *result, NSError *_Nullable error))completion
-    NS_SWIFT_NAME(list(maxResults:pageToken:completion:));
+                completion:(void (^)(FIRStorageListResult *result,
+                                     NSError *_Nullable error))completion NS_REFINED_FOR_SWIFT;
+// NS_SWIFT_NAME(list(maxResults:pageToken:completion:));
```
ncooke3, line 279: nit: same here. should this be removed now?

ncooke3 (approved):
LGTM – thanks for the thorough writeup! Left mostly nits below. Feel free to defer to future PRs.

paulb777:
Going to merge this here to storage-v9 since this now demonstrates an always Swift-on top library approach along with providing the structure to pass through and/or override any Objective C API.

I'll address open questions and next steps in subsequent PRs.

Feel free to continue to add comments and I'll address those as well.

.github/workflows/storage.yml
paulb777: Yes - for this branch while we work through the Storage changes for v9.  We'll need a complete review when we merge back to master.

PR #9046, opened 2021-12-03
Add app ID header to Firebase Auth

rosalyntan:
Add the `X-Firebase-GMPID` header to requests made from Firebase Auth.


paulb777 (approved):
LGTM. Presumably needs to wait for the backend to merge?

rosalyntan:
> LGTM. Presumably needs to wait for the backend to merge?

The auth backend is ready to handle the `X-Firebase-GMPID` header (but not the heartbeat header, which I've removed from this PR).

PR #9050, opened 2021-12-03
Fix flaky messaging test

ncooke3:
The flaky test requires internet access. I think it's easiest to increase the expectation timeout time by a few seconds but would like to verify this works.

**Update** I was not able to repro the flaky behavior seen in the nightlies. I still think it may be worth it to extend expectation wait time and see if this test flakes again in nightlies.

#no-changelog

### DIFF
--- a/FirebaseMessaging/Tests/UnitTests/FIRMessagingExtensionHelperTest.m
+++ b/FirebaseMessaging/Tests/UnitTests/FIRMessagingExtensionHelperTest.m
@@ -83,7 +83,8 @@ static NSString *const kValidImageURL =
     [_mockExtensionHelper populateNotificationContent:content withContentHandler:handler];
     OCMVerify([_mockExtensionHelper loadAttachmentForURL:[OCMArg any]
                                        completionHandler:[OCMArg any]]);
-    [self waitForExpectationsWithTimeout:1.0 handler:nil];
+    // Wait longer to accomodate increased network latency when running on CI.
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
   }
 }
 #endif  // COCOAPODS


scripts/test_catalyst.sh
```diff
@@ -57,6 +57,8 @@ args=(
   "CODE_SIGN_IDENTITY=-" "CODE_SIGNING_REQUIRED=NO" "CODE_SIGNING_ALLOWED=NO"
   # GHA is still running 10.15.
   "MACOSX_DEPLOYMENT_TARGET=10.15"
+  # Run tests multiple times.
+  -test-iterations 1000
```
ncooke3, line 61: When testing on Catalyst, add flags here

scripts/test_catalyst.sh
```diff
@@ -57,6 +57,8 @@ args=(
   "CODE_SIGN_IDENTITY=-" "CODE_SIGNING_REQUIRED=NO" "CODE_SIGNING_ALLOWED=NO"
   # GHA is still running 10.15.
   "MACOSX_DEPLOYMENT_TARGET=10.15"
+  # Run tests multiple times.
+  -test-iterations 10000
```
ncooke3, line 61: It didn't flake when running 10k times yet still flakes on the nightlies on a sort of regular basis. I propose setting the test in question to have a longer timeout.

paulb777: Could it be related to running the tests for multiple platforms at the same time? Is anything in the tests related to changing and resetting external state?

ncooke3: > Could it be related to running the tests for multiple platforms at the same time?

I don't think so, because i isolated the GHA so _only_ the `messaging / catalyst` was running.

> Is anything in the tests related to changing and resetting external state?

Maybe. Previously i was only running the flaky test but i will reenable all tests and try to repro one more time. If it still doesn't flake there, then I suggest extended expectation wait time and seeing if flake reoccurs in nightly

ncooke3: **Update** it did not flake when running all the `messaging` tests. I suggest extending expectation wait time and investigating further if it flakes in nightly again.

chliangGoogle (approved):

paulb777 (approved):

Issue #8052, opened 2021-05-07
Use buildcache to speed up CI


paulb777:
See https://github.com/marketplace/actions/buildcache-action and https://twitter.com/fastsquatch/status/1390465013983367176

There is potentially a 40-50% speedup opportunity

paulb777:
Next step is to verify that the merged #9055 makes a difference.

paulb777:
A minimal random survey of about 10 before and after shows some nice speedups - like 9 minutes to 6 minutes. 

There is likely another project for a deeper investigation to look at more data and understand why some jobs look to be unchanged.  

But I'll close this one here. cc: @granluo 

Thanks @a-25

a-25:
Hello, @paulb777. Glad to hear.
Could you post examples of unchanged jobs? I used buildcache not in all of them, only in those who needed building.

paulb777:
@a-25 See the spm job in https://github.com/firebase/firebase-ios-sdk/runs/4531139622?check_suite_focus=true and https://github.com/firebase/firebase-ios-sdk/runs/4604330680?check_suite_focus=true.  The Swift build went from 5 minutes to 1 minute, but the iOS Unit Tests - which is also primarily a build  - stayed about the same.

a-25:
I see, thanks. Possibly I could suggest adding statistics (`...buildcache -s`) after each build job instead of an aggregate one. It could help speed debugging.

PR #9055, opened 2021-12-06
Use GitHub buildcache #8052

a-25:
Used github buildcache plugin, #8052


paulb777:
@a-25 Thanks for the tackling this potential CI improvement.  A good next step would be to demonstrate the performance improvement of using the buildcache plugin. That might be easier to do in your forked repo and then provide links that show the advantage of the cache.

a-25:
@paulb777: Hello. I made some measurements, they are below. Used only core.yml file for instance.

We are interested in "iOS Unit Tests" and "Post Run mikehardy/buildcache-action@v1" sections.

- Without buildcache - https://github.com/a-25/firebase-ios-sdk/runs/4440188088?check_suite_focus=true
- After buildcache, first time (cache warm up) - https://github.com/a-25/firebase-ios-sdk/runs/4440226702?check_suite_focus=true
- After buildcache, second time (no files changed) - https://github.com/a-25/firebase-ios-sdk/runs/4440252091?check_suite_focus=true

As I see, the cache at least doesn't slow the process down. And building with cache gives about 40% increase.
I think the more accurate statistics should be collected later, after using. But as I see most files will remain unchanged (because one PR is usually devoted to only one project while others stay the same).
Also I tried to use separate caches for different OS to reduce cache rebuilding.

paulb777:
Thanks @a-25 . The 40% improvement is great!

For security concerns, we're not comfortable depending on a third-party action specified with a tag. Would you update to a commit hash?  See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions and https://michaelheap.com/improve-your-github-actions-security/

paulb777:
@a-25 Let me know if you want to continue this. Otherwise I can pick it up in the next week or two.

a-25:
@paulb777: Hello. Yes, I'll going to finish this one soon. Was busy lately.

a-25:
@paulb777: updated the hash, waiting for the checks to prove everything is working.

paulb777:
@a-25 Thanks. Sorry for the delay in unblocking the CI. It looks like something about the change is causing the jobs to be cancelled before even starting.

a-25:
@paulb777: hello. 
This looks strange, I tested this configuration in my repository, and it's working fine: https://github.com/a-25/firebase-ios-sdk/runs/4520127644?check_suite_focus=true
Reverted buildcache-action to previous stable version - let's see if it helps.

a-25:
@paulb777: I think yes.

I compared two jobs:
Without cache - https://github.com/firebase/firebase-ios-sdk/runs/4516567370?check_suite_focus=true
With cache - https://github.com/firebase/firebase-ios-sdk/runs/4521400838?check_suite_focus=true
Slowing is about 15% as I see. And the last job had almost no cache hits.

Maybe you could replay the last job to compare the results for the second time? Or you can merge this PR and see if it helps on the longer prospective. If not - it will be easy to revert it, I guess, because there are just additions, not code changes.

paulb777:
@a-25 Good point about merging and giving it a try for a few days and reverting if we don't see a real improvement.  Let's get one more approval and then we can merge.

.github/workflows/abtesting.yml
```diff
@@ -35,6 +35,9 @@ jobs:
     runs-on: macos-11
     steps:
     - uses: actions/checkout@v2
+    - uses: mikehardy/buildcache-action@50738c6c77de7f34e66b870e4f8ede333b69d077
```
paulb777, line 38: This is 1.2.1. See https://github.com/mikehardy/buildcache-action/tags

paulb777 (approved):

.github/workflows/abtesting.yml
a-25: Yes, this is on purpose - https://github.com/firebase/firebase-ios-sdk/pull/9055#issuecomment-993510558

ryanwilson (approved):

paulb777:
Thanks for the contribution @a-25! We'll merge this now and keep and eye on our CI times.

paulb777:
Hey @a-25 I'm working on Google Open Source Peer Bonus nominations and would like to recognize your contributions on this PR and others. I couldn't find a name or email address for you, so if you're interested, please email them to paulbeusterien at google.com and I'll make the nomination.

PR #9057, opened 2021-12-06
Add callAsFunction and inferred function types

peterfriese:
Using inferred request and response types for HTTP callable functions still is something we need to discuss, but using callAsFunction shouldn't be controversial.

I suggest adding both to Morten's PR, as they are part of the API review.


paulb777:
@ryanwilson PTAL from an API perspective. See the last two commits.

paulb777:
Cherry-picked into #9091 and closing in favor of that.

PR #9065, opened 2021-12-08
Implement CustomTokenProviderDelegate logic

rosalyntan:
Add a `FIRCustomTokenProviderDelegate` to `FIRAuth`, for obtaining new custom token to exchange for a Firebase ID token if a refresh token is not available.

Note: All development is being done on the `passthrough` branch, to be merged into `master` once the feature is complete.

API Proposal: go/firebase-auth-passthrough-sdk-api

#no-changelog


paulb777 (commented):
I added some questions in the API review. The PR looks like a good implementation of the API review. As I pointed out there, I'd like to better understand the impact on other Firebase products that depend on Auth and explore if we can be more Swift-first with new APIs.

paulb777 (approved):
As @ryanwilson has already approved these API changes, I'm fine to follow up on the Swift API improvements in subsequent PRs.

PR #9067, opened 2021-12-08
Experiment with Firebase module options

paulb777:
wip


paulb777:
Closing for now. It looks like we'll need a more slash and burn approach of completely removing the Firebase library for this closer to the breaking change release. See #9074

PR #9071, opened 2021-12-09
Fix build errors in `FirebaseCombineSwift`

ncooke3:
- Fix `combine` regression from #9069
- Run `combine` on dependency source changes
#no-changelog

### DIFF
--- a/.github/workflows/combine.yml
+++ b/.github/workflows/combine.yml
@@ -30,6 +30,14 @@ on:
     # Rebuild on Ruby infrastructure changes.
     - 'Gemfile'
 
+    # Dependencies
+    - 'FirebaseCore/**'
+    - 'FirebaseAuth/**'
+    - 'FirebaseFunctions/**'
+    - 'Firestore/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/combine.yml
```diff
@@ -30,6 +30,14 @@ on:
     # Rebuild on Ruby infrastructure changes.
     - 'Gemfile'
 
+    # Dependencies
+    - 'FirebaseCore/**'
+    - 'FirebaseAuth/**'
+    - 'FirebaseFunctions/**'
+    - 'Firestore/**'
+    - 'FirebaseStorage/**'
+    - 'FirebaseStorageSwift/**'
```
ncooke3, line 39: WDYT about adding the deps of `FirebaseCombineSwift` to catch these breaks on PRs in the future?

paulb777: SGTM, maybe sufficient to do 'FirebaseCore/**/*.h', etc though

paulb777 (approved):

PR #9077, opened 2021-12-10
Specstesting Copy files to `.cocoapods` dir

granluo:
Updated podspecs will be copied to a local `.cocoapods` dir, so when podspec testing is run and source to the specstesting spec repo, the latest dependencies will be from the head of a PR.

### DIFF
--- a/ReleaseTooling/Sources/PodspecsTester/InitializeSource.swift
+++ b/ReleaseTooling/Sources/PodspecsTester/InitializeSource.swift
@@ -98,7 +98,7 @@ struct InitializeSpecTesting {
   }
 
   // Copy updated specs to the `${HOME}/.cocoapods/` dir.
-  private static func copyPodspecs(from specsDir: URL, manifest: FirebaseManifest.Manifest, exclude) {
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Will you add the output of `find .` on the  new local repo in `.cocoapods` after running the tool as a PR comment?

granluo:
> Will you add the output of `find .` on the new local repo in `.cocoapods` after running the tool as a PR comment?

Added in to the tool. Will determine if we need to add a flag to display in a later PR for GHA workflow setup.

paulb777:
Oh sorry. I meant that I wanted to see the output of `find .` in a PR comment to check my understanding of the changes. I'm not sure it should be in the tool itself, but maybe?

granluo:
> Oh sorry. I meant that I wanted to see the output of `find .` in a PR comment to check my understanding of the changes. I'm not sure it should be in the tool itself, but maybe?
I will create another PR for running the tool. `find .` could be added to the script in a workflow. We can check that out in the next PR.

PR #9080, opened 2021-12-11
Control async/await testing from source instead of podspec

paulb777:
The minimum versions in the podspec cause `pod install` and build warnings.  Also using `@available` is consistent with the rest of Firebase.

### DIFF
--- a/FirebaseRemoteConfig/Tests/SwiftAPI/AsyncAwaitTests.swift
+++ b/FirebaseRemoteConfig/Tests/SwiftAPI/AsyncAwaitTests.swift
@@ -28,131 +28,134 @@ private enum Constants {
   static let darthSidious = "Darth Sidious"
 }
 
-class AsyncAwaitTests: APITestBase {
-  var console: RemoteConfigConsole!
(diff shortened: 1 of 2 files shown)


FirebaseRemoteConfig/Tests/SwiftAPI/AsyncAwaitTests.swift
```diff
@@ -28,131 +28,134 @@ private enum Constants {
...
-  override func setUp() {
-    super.setUp()
-    if APITests.useFakeConfig {
-      fakeConsole.config = [Constants.key1: Constants.value1]
-    } else {
-      console = RemoteConfigConsole()
-      console.updateRemoteConfigValue(Constants.obiwan, forKey: Constants.jedi)
+#if compiler(>=5.5) && canImport(_Concurrency)
+  @available(iOS 15, tvOS 15, macOS 12, watchOS 8, *)
+  class AsyncAwaitTests: APITestBase {
```
paulb777, line 33: Only indentation changes follow

ncooke3 (approved):

PR #9081, opened 2021-12-12
Make `CoreUnit` scheme buildable

ncooke3:
When developing `FirebaseCore` with SPM, I commonly try to build `FirebaseCore` in Xcode and get: 
Screen Shot 2021-12-11 at 7 01 31 PM

So I tweaked the scheme to allow for running (building) the library.

#no-changelog


paulb777 (approved):
Good spot. We might want to do this for all the spm test schemes

LGTM on green

PR #9085, opened 2021-12-12
Make all SPM test schemes support `run`,  `test`, and `analyze` configs

ncooke3:
**All** SPM test schemes now support run, test, and analyze configs.

Follow-up work to #9081. See comment https://github.com/firebase/firebase-ios-sdk/pull/9081#pullrequestreview-829573573 for motivation.

This makes for a better devx while developing Firebase while using SPM, imo.

You can now build a test scheme without running the test suite.

All test schemes now support below configuration:
Screen Shot 2021-12-12 at 6 18 14 PM

#no-changelog


ryanwilson (approved):
👏  nice work, this will be super handy!

paulb777 (approved):

PR #9089, opened 2021-12-13
Update remote `generate_access_token` script

ncooke3:
#no-changelog

### DIFF
--- a/scripts/generate_access_token.sh
+++ b/scripts/generate_access_token.sh
@@ -66,7 +66,7 @@ if [[ $GHA_SECRET == "local_dev" ]]; then
     cp $SERVICE_ACCOUNT ~/.credentials/
 else
     echo "Decrypting ${SERVICE_ACCOUNT_FILE}.gpg"
-    scripts/decrypt_GHA_SECRET.sh $SERVICE_ACCOUNT ~/.credentials/$SERVICE_ACCOUNT_FILE "$plist_secret"
+    scripts/decrypt_gha_secret.sh $SERVICE_ACCOUNT ~/.credentials/$SERVICE_ACCOUNT_FILE "$GHA_SECRET"
 fi
 
 echo "GOOGLE_APPLICATION_CREDENTIALS=${HOME}/.credentials/${SERVICE_ACCOUNT_FILE}" >> $GITHUB_ENV
--- a/scripts/generate_access_token.sh
+++ b/scripts/generate_access_token.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+    
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


scripts/generate_access_token.sh
```diff
@@ -66,7 +66,7 @@ if [[ $GHA_SECRET == "local_dev" ]]; then
     cp $SERVICE_ACCOUNT ~/.credentials/
 else
     echo "Decrypting ${SERVICE_ACCOUNT_FILE}.gpg"
-    scripts/decrypt_GHA_SECRET.sh $SERVICE_ACCOUNT ~/.credentials/$SERVICE_ACCOUNT_FILE "$plist_secret"
```
ncooke3, line 69: This shouldn't fix the nightly test errors seen in #9086 but are improvements I made while trying to reproduce nightly failures.

scripts/generate_access_token.sh
```diff
@@ -66,7 +66,7 @@ if [[ $GHA_SECRET == "local_dev" ]]; then
...
+    scripts/decrypt_gha_secret.sh $SERVICE_ACCOUNT ~/.credentials/$SERVICE_ACCOUNT_FILE "$GHA_SECRET"
```
ncooke3, line 69: This shouldn't fix the nightly test errors seen in #9086 but are improvements I made while trying to reproduce nightly failures.

paulb777: Assuming that `"$plist_secret"` and `"$GHA_SECRET"` are equivalent?

paulb777 (approved):

scripts/generate_access_token.sh
ncooke3: Yes. `plist_secret` is the env var that GHA knows about. We pass it to the script like so:

https://github.com/firebase/firebase-ios-sdk/blob/0e855f4ef6dffe30e2e9375b6034d5d76e5345f4/.github/workflows/remoteconfig.yml#L35

After some head scratching while running locally, I realized that I was using the expected GHA env var in the diff'd code rather than the local param.

ncooke3:
Hmm i reverted the script and tripped CI and I'm seeing weird script failure. The script ran successfully in the nightlies: https://github.com/firebase/firebase-ios-sdk/runs/4517060855?check_suite_focus=true so I'm not sure what the issue could be. Maybe GHA related. Ill try again later.

paulb777:
Looks like a failure building NIOCore?  Hopefully a flake. I restarted the job.

ncooke3:
Ah yeah, you're right @paulb777. Looks like SwiftNIO had a [release](https://github.com/apple/swift-nio/releases/tag/2.36.0) about 5 hours ago so that may explain why nightlies ran well but this now fails.

ncooke3:
SwiftNIO built successfully on Xcode 13.2 and RC tests passed. 

@paulb777  Does it make sense for a SwiftNIO minor version release to break building on Xcode 13.0?

paulb777:
It's possible. Would have to try to see.

ncooke3:
The SwiftNIO-related failure was that Xcode 13.0 doesn't have the macOS 12 SDK and so the `macOS 12` availability checks were failing. I'll kick off CI after #9098 (implicitly bumps GHA to use Xcode 13.1 (default) over hard-coded Xcode 13.0) merges and we should be green and ready to merge.

PR #8983, opened 2021-11-18
Add CI for Functions Codable

paulb777:
Pending #8854

PR #9091, opened 2021-12-13
Functions Codable after sharing implementation with RTDB

paulb777:
Umbrella PR that rolls up the following

#8853 from @mortenbekditlevsen to share Codable implementation
#8854 from  @mortenbekditlevsen to add Firebase Functions Codable implementation
Additional Functions Codable APIs cherry-picked from @peterfriese's #9057 
CI implementation cherry-picked from @paulb777's #8983

Reviewers, you may want to review commit by commit

The FirebaseFunctionsSwift CocoaPods implementation, additional CI and associated documentation will come in follow up PR(s)

I plan to merge this PR without squashing to keep the contribution history in master.


paulb777:
@mortenbekditlevsen @peterfriese Please confirm I merged everything correctly.

@ryanwilson This should now be a one-stop PR for review of the Codable changes for Functions and RTDB

cc: @schmidt-sebastian

ryanwilson:
Thanks a ton for putting this together, @paulb777!

ryanwilson:
I struggle with the naming of `StructuredEncoder` - it's very generic which accurately matches the implementation, but I worry that it could be confused with a non-Firebase type and used for other purposes. 

What about something like `FirebaseDataEncoder/Decoder`? I recognize it's code in `third_party` but it's a fork that we want to use exclusively for Firebase Codable implementations and that helps set the expectations.

paulb777:
Updated Callable API to use `requestAs` and `responseAs` instead of `requestType` and `requestAs`

Enabled second way of using API after an internal conversation. See the new test 

```
    func testScalarAsyncAlternateSignature() async throws {
      let function: Callable<Int16, Int> = functions.httpsCallable("scalarTest")
      let result = try await function.call(17)
      XCTAssertEqual(result, 76)
    }
```

versus

```
func testScalarAsync() async throws {
      let function = functions.httpsCallable(
        "scalarTest",
        requestType: Int16.self,
        responseType: Int.self
        requestAs: Int16.self,
        responseAs: Int.self
      )

      let result = try await function.call(17)
      XCTAssertEqual(result, 76)
    }
```

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,199 @@
...
+   *
+   * - Parameter name: The name of the Callable HTTPS trigger
+   * - Parameter requestType: The type of the `Encodable` entity to use for requests to this `Callable`
+   * - Parameter responseType: The type of the `Decodable` entity to use for responses from this `Callable`
+   */
+  func httpsCallable<Request: Encodable,
+    Response: Decodable>(_ name: String,
+                         requestAs requestType: Request.Type = Request.self,
+                         responseAs responseType: Response.Type = Response.self,
+                         encoder: StructureEncoder = StructureEncoder(),
```
mortenbekditlevsen, line 34: Just a quick question: With default parameters for `requestAs` and `responseAs` (and neat that you can actually default them in that fashion - I did not know that), do you need the overload on line 52 too? It appears to be doing the same thing as the default parameters...

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,199 @@
...
+   * resumes with a new Instance ID the next time you call this method.
+   *
+   * - Parameter data: Parameters to pass to the trigger.
+   * - Parameter completion: The block to call when the HTTPS request has completed.
+   *
+   * - Throws: An error if any value throws an error during encoding.
+   */
+  public func call(_ data: Request,
+                   completion: @escaping (Result<Response, Error>)
+                     -> Void) throws {
```
mortenbekditlevsen, line 110: In the review thread of #8854 we discussed if it could make sense to 'squash' the two types of error handling into one:

https://github.com/firebase/firebase-ios-sdk/pull/8854#issuecomment-974697150

and @schmidt-sebastian answered that he thought it would be a good solution in this case.

Should I make a PR for that, or can you change the method to be non-throwing and rather have a `do {} catch {} around the entire function body that calls the completion with the error in the `catch` clause?

mortenbekditlevsen (approved):

mortenbekditlevsen:
I am glad to see that this got through your internal API review process.

Regarding the choice to supply both explicit typing of the callable and inferred typing:
Should this design choice form precedence for other similar API? As in: should `DocumentSnapshot` then also have an inferred version of 
```
                                 decoder: Firestore.Decoder? = nil) throws -> T?
```
without the `as type: T.Type`? And the same in other similar situations?

And if not: Why might that be?

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Good catch! Removed.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done in bdacb29c6.

paulb777 (commented):
@mortenbekditlevsen Thanks for the review.  Comments addressed in bdacb29c6

paulb777:
@mortenbekditlevsen The design choice of supplying both explicit typing of the callable and inferred typing sounds like a good standard to me. @ryanwilson What do you think?

BTW, we're not completely through the internal API review yet - but hopefully getting close. 😄

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,184 @@
...
+  /// Creates a directly callable function.
+  ///
+  /// This allows users to call a HTTPS Callable Funciton like a normal Swift function:
+  ///
+  ///     let greeter = functions.httpsCallable("greeter",
+  ///                                           requestType: GreetingRequest.self,
+  ///                                           responseType: GreetingResponse.self)
+  ///     try greeter(data) { result in
+  ///       print(result.greeting)
+  ///     }
```
ryanwilson, line 123: Can we use DocC formatting here? https://developer.apple.com/documentation/xcode/formatting-your-documentation-content#Add-Code-Listings

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * The data passed into the trigger must be of the generic `Request` type:
+     *
+     * The request to the Cloud Functions backend made by this method automatically includes a
+     * Firebase Instance ID token to identify the app instance. If a user is logged in with Firebase
+     * Auth, an auth ID token for the user is also automatically included.
+     *
+     * Firebase Instance ID sends data to the Firebase backend periodically to collect information
+     * regarding the app instance. To stop this, see `[FIRInstanceID deleteIDWithHandler:]`. It
+     * resumes with a new Instance ID the next time you call this method.
```
ryanwilson, line 146: Same here re: out of date IID docs

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * - Parameter data: The `Request` representing the data to pass to the trigger.
+     *
+     * - Throws: An error if any value throws an error during encoding.
+     * - Throws: An error if any value throws an error during decoding.
```
ryanwilson, line 151: Nit: should this just be `during encoding or decoding`?

FirebaseDatabaseSwift/Tests/Codable/ServerValueCodingTests.swift
```diff
@@ -87,3 +87,114 @@ extension CurrencyAmount: ExpressibleByFloatLiteral {
     self.value = Decimal(value)
   }
 }
+
```
ryanwilson, line 90: I think I'm missing something here - none of these functions or classes are used anywhere above, are they?

ryanwilson (commented):
I still need to look at the `IntegrationTests.swift` but looking good so far!

FirebaseDatabaseSwift/Tests/Codable/ServerValueCodingTests.swift
paulb777: All except the bottom function are used by the `assertThat` implementation. These functions are uses more extensively in the Firestore tests and we may want to unify them.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,199 @@
...
+      callable.timeoutInterval
+    }
+    set {
+      callable.timeoutInterval = newValue
+    }
+  }
+
+  enum CallableError: Error {
+    case internalError
+  }
```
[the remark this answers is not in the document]
paulb777: Done.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,199 @@
...
+   *
+   * The data passed into the trigger must be of the generic `Request` type:
+   *
+   * The request to the Cloud Functions backend made by this method automatically includes a
+   * Firebase Instance ID token to identify the app instance. If a user is logged in with Firebase
+   * Auth, an auth ID token for the user is also automatically included.
+   *
+   * Firebase Instance ID sends data to the Firebase backend periodically to collect information
+   * regarding the app instance. To stop this, see `[FIRInstanceID deleteIDWithHandler:]`. It
+   * resumes with a new Instance ID the next time you call this method.
```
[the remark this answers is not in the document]
paulb777: Done. I already caught this and merged #8981 for the ObjC version.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done. PTAL since I'm not sure how to test.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Fixed here and above.

FirebaseSharedSwift/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
paulb777: Done.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,171 @@
...
+      let encoded = try encoder.encode(data)
+      let result = try await callable.call(encoded)
+      return try decoder.decode(Response.self, from: result.data)
+    }
+
+    /// Creates a directly callable function.
+    ///
+    /// This allows users to call a HTTPS Callable Function like a normal Swift function:
+    /// ```swift
+    ///     let greeter = functions.httpsCallable("greeter",
```
peterfriese, line 157: Now that we support both type-specific parameters and inferred types, shouldn't we add another sample to show how to use this as well?

I.e. add the following:

    /// You can also call a HTTPS Callable function using the following syntax:
    /// ```swift
    ///     let greeter: Callable<GreetingRequest, GreetingResponse> = functions.httpsCallable("greeter")
    ///     let result = try await greeter(data)
    ///     print(result.greeting)
    /// ```

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,171 @@
...
+                       StructureDecoder()) async throws -> Response {
+      let encoded = try encoder.encode(data)
+      let result = try await callable.call(encoded)
+      return try decoder.decode(Response.self, from: result.data)
+    }
+
+    /// Creates a directly callable function.
+    ///
+    /// This allows users to call a HTTPS Callable Function like a normal Swift function:
+    /// ```swift
```
peterfriese, line 156: You specified the language here, but in line 107 you didn't.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,171 @@
...
+    } catch {
+      completion(.failure(error))
+    }
+  }
+
+  /// Creates a directly callable function.
+  ///
+  /// This allows users to call a HTTPS Callable Funciton like a normal Swift function:
+  ///
+  ///     let greeter = functions.httpsCallable("greeter",
```
peterfriese, line 107: Code block not fenced, and language not specified.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,171 @@
...
+        }
+      }
+    } catch {
+      completion(.failure(error))
+    }
+  }
+
+  /// Creates a directly callable function.
+  ///
+  /// This allows users to call a HTTPS Callable Funciton like a normal Swift function:
```
peterfriese, line 105: Typo: Funciton

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -0,0 +1,171 @@
...
+  ///
+  ///     let greeter = functions.httpsCallable("greeter",
+  ///                                           requestType: GreetingRequest.self,
+  ///                                           responseType: GreetingResponse.self)
+  ///     try greeter(data) { result in
+  ///       print(result.greeting)
+  ///     }
```
peterfriese, line 112: Add sample code using inferred types:

  /// You can also call a HTTPS Callable function using the following syntax:
  /// ```swift
  ///     let greeter: Callable<GreetingRequest, GreetingResponse> = functions.httpsCallable("greeter")
  ///     try greeter(data) { result in
  ///       print(result.greeting)
  ///     }
  /// ```

peterfriese (changes requested):
Found a couple of typos. Also, should we add code samples for calling using inferred types?

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: Done.

paulb777 (commented):
Thanks @peterfriese !

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -78,116 +50,129 @@ public struct Callable<Request: Encodable, Response: Decodable> {
...
+  ///
+  /// - Parameter data: Parameters to pass to the trigger.
+  /// - Parameter completion: The block to call when the HTTPS request has completed.
+  ///
+  /// - Throws: An error if any value throws an error during encoding.
   public func call(_ data: Request,
                    completion: @escaping (Result<Response, Error>)
-                     -> Void) throws {
-    let encoded = try encoder.encode(data)
+                     -> Void) {
```
ryanwilson, line 80: This isn't marked as throws anymore - that makes sense given the implementation, right? If so, we need to update the doc

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -78,116 +50,129 @@ public struct Callable<Request: Encodable, Response: Decodable> {
...
   ///                                           responseType: GreetingResponse.self)
   ///     try greeter(data) { result in
   ///       print(result.greeting)
   ///     }
-  ///
+  /// ```
+  /// You can also call a HTTPS Callable function using the following syntax:
+  /// ```swift
+  ///     let greeter: Callable<GreetingRequest, GreetingResponse> = functions.httpsCallable("greeter")
+  ///     try greeter(data) { result in
```
ryanwilson, line 117: Nit: this isn't marked as throw anymore so `try` should be removed.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -78,116 +50,129 @@ public struct Callable<Request: Encodable, Response: Decodable> {
...
     ///                                           requestType: GreetingRequest.self,
     ///                                           responseType: GreetingResponse.self)
     ///     let result = try await greeter(data)
     ///     print(result.greeting)
-    ///
+    /// ```
+    /// You can also call a HTTPS Callable function using the following syntax:
+    /// ```swift
+    ///     let greeter: Callable<GreetingRequest, GreetingResponse> = functions.httpsCallable("greeter")
+    ///     try await greeter(data)
```
ryanwilson, line 173: `let result = ` is missing

paulb777:
As discussed above, I renamed `StructuredEncoder` to `FirebaseDataEncoder`. 

All comments should be addressed and looking for approvals from @ryanwilson and @peterfriese and optionally @ncooke3, @schmidt-sebastian, and anyone else interested.

paulb777:
I squashed the commits from 15 to 5.

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+
+import Foundation
+
+import FirebaseFunctions
+import FirebaseFunctionsSwift
+import FirebaseFunctionsTestingSupport
+import XCTest
+
+/// This file was intitialized as a direct port of the Objective C
+/// FirebaseFunctions/Tests/Integration/FIRIntegrationTests.m
```
ryanwilson, line 23: Nit: please update this line to the Swift file

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+  var bool: Bool
+  var int: Int32
+  var long: Int64
+  var string: String
+  var array: [Int32]
+  // NOTE: Auto-synthesized Encodable conformance uses 'encodeIfPresent' to
+  // encode Optional values. To encode Optional.none as null you either need
+  // to write a manual encodable conformance or use a helper like the
+  // propertyWrapper here:
+  @NullEncodable var null: Bool?
```
ryanwilson, line 38: I think this is from another set of tests.. but is this necessary for customers to use the `@NullEncodable` wrapper? i.e. what happens if we don't use it, does Codable fail? I'm guessing it means the key doesn't get written in the backend, but maybe it's fine for Functions that the key doesn't exist? If this is required, maybe we need to publicize this and make it part of the public API. It may be helpful to add the context as to how this relates to the Functions backend as well in the comment.

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+    withToken: nil
+  )
+  let projectID = "functions-swift-integration-test"
+
+  override func setUp() {
+    super.setUp()
+    functions.useLocalhost()
+  }
+
+  func testData() throws {
```
ryanwilson, line 78: This doesn't throw anymore, does it? Looks like the only `try` is caught.

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+      let expected = DataTestResponse(
+        message: "stub response",
+        long: 420,
+        code: 42
+      )
+      XCTAssertEqual(response, expected)
+    }
+  #endif
+
+  func testScalar() throws {
```
ryanwilson, line 134: Same here - this doesn't throw anymore

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+    @available(iOS 15, tvOS 15, macOS 12, watchOS 8, *)
+    func testScalarAsyncAlternateSignature() async throws {
+      let function: Callable<Int16, Int> = functions.httpsCallable("scalarTest")
+      let result = try await function.call(17)
+      XCTAssertEqual(result, 76)
+    }
+
+  #endif
+
+  func testToken() throws {
```
ryanwilson, line 175: Also doesn't throw

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+        requestAs: [String: Int].self,
+        responseAs: [String: Int].self
+      )
+
+      let data = try await function.call([:])
+      XCTAssertEqual(data, [:])
+    }
+  #endif
+
+  func testFCMToken() throws {
```
ryanwilson, line 227: Also doesn't throw

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+  #endif
+
+  // No parameters to call should be the same as passing nil.
+  // If no parameters are required, then the non-typed API
+  // is more appropriate since it specifically avoids defining
+  // type.
+  //  func testParameterless() {
+  //  }
+  //
+  //
```
ryanwilson, line 300: Is this still relevant?

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+  func testMissingResult() throws {
+    let expectation = expectation(description: #function)
+    let function = functions.httpsCallable(
+      "missingResultTest",
+      requestAs: Int?.self,
+      responseAs: Int?.self
+    )
+    function.call(nil) { result in
+      do {
+        _ = try result.get()
```
ryanwilson, line 310: Should this also have a `XCTFail("Failed to throw error for missing result")`?

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,652 @@
...
+  func testUnhandledError() throws {
+    let expectation = expectation(description: #function)
+    let function = functions.httpsCallable(
+      "unhandledErrorTest",
+      requestAs: [Int].self,
+      responseAs: Int.self
+    )
+    function.call([]) { result in
+      do {
+        _ = try result.get()
```
ryanwilson, line 349: Same here re: `XCTFail("Failed to throw error for missing result")`

ryanwilson (commented):
Just a couple of nits then the tests LGTM, thanks!

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
paulb777: It's a good question. The argument makes sense to me.  I wrote the test and updated the APIs' first parameter to be `_ data: Request? = nil`. @ryanwilson Let me know if you agree and I'll update the API proposal.

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
paulb777: Yep. Good catch! Added here and below.

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
paulb777: Yep. Good catch! Added here and below.

paulb777 (commented):
Thanks @ryanwilson !

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
mortenbekditlevsen: My thoughts is that it was likely added since one of the tests contain examples of each primitive encodable type. 

I'd say that if someone needs to represent nil to their callable function, then they would likely know how to do this by either adding a wrapper like this themselves or implement the `Codable` conformance by hand instead of using the synthesized version.

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -73,7 +73,7 @@ public struct Callable<Request: Encodable, Response: Decodable> {
   ///
   /// - Parameter data: Parameters to pass to the trigger.
   /// - Parameter completion: The block to call when the HTTPS request has completed.
-  public func call(_ data: Request,
+  public func call(_ data: Request? = nil,
```
ryanwilson, line 76: Hmm, how common do you think it is for people to have functions with no parameters? I worry that we're over optimizing here and the call site is less clear that no arguments are being passed (in the situation where you want that)

peterfriese (approved):
LGTM - excited to see this land!

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
mortenbekditlevsen: I second that. If you have a truly optional parameter you just use `Param?.self` as your generic parameter. 
The only thing I can see missing is if you either never have any parameters or never have return values. In those situations it would be nice to use `Void.self` as generic params, but unfortunately `Void` does not conform to `Codable` afair.
But I also think that it's probably overkill to model an overload with void request param or void response param and one with both. Instead we could start a discussion on the swift evolution forums to argue that `Void` ought to be `Codable`.

paulb777: Reverted parameterless

FirebaseFunctionsSwift/Tests/IntegrationTests.swift
ryanwilson: SGTM, thanks! Discussed offline and I agree it shouldn't be part of the API surface. Can always revisit in the future if we get feedback that it's needed.

ryanwilson (approved):

PR #9093, opened 2021-12-13
Add Swift Doc Scrub to dashboard

paulb777:
See https://firebase.google.com/docs/reference/swift/firebaseremoteconfig/api/reference/Classes/RemoteConfig?authuser=0 for an example where almost every API references an Objective C type or call even though the APIs themselves are Swift


SwiftDashboard.md
```diff
@@ -45,6 +46,7 @@ for Storage.
 Property Wrappers and likely the SwiftUI lifecycle bits, but an audit and improvements could likely be made. The existing FIAM and Analytics View modifier
 APIs would fit into this category.
 * **Property Wrappers**: Property wrappers are used to improve the API.
+* **Swift Doc Scrub**: Review and update to change Objective C types and call examples to Swift.
```
ryanwilson, line 49: Nice! Do you think it's worth adding something about ensuring future support for DocC in our formatting?

ryanwilson (approved):

SwiftDashboard.md
paulb777: Good idea! Added a sentence.

ryanwilson (approved):

PR #9094, opened 2021-12-13
Remove unused STS auth code flow.

rosalyntan:
Removing the unused Secure Token Service authorization code flow, as it is never invoked anywhere and is just dead code. In addition, the authorization code flow is deprecated on the backend.

#no-changelog non-user-visible refactoring


paulb777 (approved):
Nice cleanup!

PR #9100, opened 2021-12-14
Add multi-tenancy support to generic OAuth providers.

### LINKED ISSUE #7990: OAuthProvider with tenant support

  * Xcode version:  12.4
  * Firebase SDK version: 7.10.0
  * Installation method: `CocoaPods ` 
  * Firebase Component: Auth

#### Steps to reproduce:

there is a problem when we try to signin with a specific tenant and specific provider using OAuthProvider(see code below for a better understand)

in the file `FIROAuthProvider.m` line `311` where the params are added in the URL, the param `tid` is missed, and in some cases this param is needed to the correct functionality of the authentication, don't know if it's missed in the other providers, but in this case of the oauth provider is needed with a multi tenant implementation.

something to add is that Android SDK create the url in a right way when we use the tenant id in the same way.

#### Relevant Code:

let's use this code to perform the signin...
```
let auth = Auth.auth()
auth.tenantID = "tenant_id"
let provider = OAuthProvider(providerID: "provider_id", auth: auth)

auth.signIn(with: provider, uiDelegate: nil) { authResult, error in
// handle here the result
}
```

rosalyntan:
Fixes #7990

### DIFF
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
-- [changed] Added a `X-Firebase-GMPID` header to network requests.
+- [changed] Added a `X-Firebase-GMPID` header to network requests. (#9046)
+- [fixed] Added multi-tenancy support to generic OAuth providers. (#9100)
 
 # 8.9.0
 - [changed] Improved error logging. (#8704)
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 - [changed] Added a `X-Firebase-GMPID` header to network requests. (#9046)
-- [fixed] Added multi-tenancy support to generic OAuth providers. (#9100)
+- [fixed] Added multi-tenancy support to generic OAuth providers. (#7990)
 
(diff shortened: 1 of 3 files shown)


paulb777 (commented):
Thanks @rosalyntan! Please add a Changelog entry.

renkelvin (approved):
Thanks @rosalyntan!

rosalyntan:
> Thanks @rosalyntan! Please add a Changelog entry.

Done!

FirebaseAuth/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
-- [changed] Added a `X-Firebase-GMPID` header to network requests.
+- [changed] Added a `X-Firebase-GMPID` header to network requests. (#9046)
+- [fixed] Added multi-tenancy support to generic OAuth providers. (#9100)
```
paulb777, line 3: Recommended practice is to specify the issue number when there is both a bug and PR - #7990 in this case.

paulb777 (approved):

FirebaseAuth/CHANGELOG.md
rosalyntan: Ah, gotcha! Done.

PR #9101, opened 2021-12-15
Require explicit typing for DocumentSnapshot decoding. DocumentReference decoding.

mortenbekditlevsen:
A while back, the Codable API for RTDB was accepted.
That API differs from the existing Codable implementation for Firestore in one important aspect:

Decoding a `DocumentSnapshot` in Firestore always returns an `Optional` result even though you may have prior knowledge (you are iterating over documents in a collection, which means they must exist) or you have prior explicit expectations (for instance: I consider it to be an error if I cannot find a `Configuration` entity in this specific Firestore document).

The similar RTDB API allows the user to be more explicit about their expectations. If they know or have a hard expectation about the presence of an entity, they can use:
```
let configuration = try snapshot.data(as: Configuration.self) 
```
while if they only know that a specific reference only _might_ contain an entity, they can be explicit about that expectation as well:
```
let optionalConfiguration = try snapshot.data(as: Configuration?.self)
```

I suggest to allow the users of Firestore the same amount of flexibility and explicit control, decreasing the amount of required 'unwrapping' when dealing with Codable entities in Firestore.

I also added a `getDocument(as: ...)` API in order to allow for simpler usage of the API.

To take a real world example, here's the guide from @peterfriese to simplify Firestore usage using Codable:
[https://github.com/peterfriese/Swift-Firestore-Guide/blob/main/FirestoreCodableSamples/Screens/MappingSimpleTypesScreen.swift](https://github.com/peterfriese/Swift-Firestore-Guide/blob/main/FirestoreCodableSamples/Screens/MappingSimpleTypesScreen.swift)
```swift
  private func fetchBook(documentId: String) {
    let docRef = db.collection("books").document(documentId)
    
    docRef.getDocument { document, error in
      if let error = error as NSError? {
        self.errorMessage = "Error getting document: \(error.localizedDescription)"
      }
      else {
        let result = Result { try document?.data(as: Book.self) }
        switch result {
        case .success(let book):
          if let book = book {
            // A Book value was successfully initialized from the DocumentSnapshot.
            self.book = book
            self.errorMessage = nil
          }
          else {
            // A nil value was successfully initialized from the DocumentSnapshot,
            // or the DocumentSnapshot was nil.
            self.errorMessage = "Document doesn't exist."
          }
        case .failure(let error):
          // A Book value could not be initialized from the DocumentSnapshot.
          switch error {
          case DecodingError.typeMismatch(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
          case DecodingError.valueNotFound(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
          case DecodingError.keyNotFound(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
          case DecodingError.dataCorrupted(let key):
            self.errorMessage = "\(error.localizedDescription): \(key)"
          default:
            self.errorMessage = "Error decoding document: \(error.localizedDescription)"
          }
        }
      }
    }
  }
```

with this PR this can be simplified a bit further to:

```swift
private func fetchBook(documentId: String) {
    let docRef = db.collection("books").document(documentId)
    
    docRef.getDocument(as: Book.self) { result in
      do {
        // A Book value was successfully initialized from the DocumentSnapshot.
        self.book = try result.get()
        self.errorMessage = nil
      } catch {
        switch error {
        case DecodingError.typeMismatch(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
        case DecodingError.valueNotFound(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
        case DecodingError.keyNotFound(_, let context):
            self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
        case DecodingError.dataCorrupted(let key):
            self.errorMessage = "\(error.localizedDescription): \(key)"
        default:
          self.errorMessage = "Error decoding document: \(error.localizedDescription)"
        }
      }
    }
  }
```

If you wish a dedicated `FirestoreDecodingError` for the situation where the snapshot can't be decoded and the reason is that the data doesn't exist, then it is fairly easy to accommodate that too by changing the decoding of the snapshot a bit:

```swift
                                 decoder: Firestore.Decoder? = nil) throws -> T {
    let d = decoder ?? Firestore.Decoder()
    let data = data(with: serverTimestampBehavior)
    do {
      return try d.decode(T.self, from: data ?? NSNull(), in: reference)
    } catch {
      if case DecodingError.valueNotFound = error, data == nil {
        // If the data is nil, and we're expecting a value, then map the error
        // into our own domain.
        throw FirestoreDecodingError.missingValue
      } else {
        throw error
      }
    }
  }
```

Let me know what you think: @ryanwilson, @paulb777, @schmidt-sebastian


Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,53 @@
...
+  ///
+  /// - Parameters:
+  ///   - value: An instance of `Encodable` to be encoded to a document.
+  ///   - encoder: An encoder instance to use to run the encoding.
+  ///   - completion: A block to execute once the document has been successfully
+  ///                 written to the server. This block will not be called while
+  ///                 the client is offline, though local changes will be visible
+  ///                 immediately.
+
+  public func getDocument<T: Decodable>(as type: T.Type,
```
schmidt-sebastian, line 35: @ryanwilson - Should we consider making this our first async/await API?

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,53 @@
...
+  /// See `Firestore.Encoder` for more details about the encoding process.
+  ///
+  /// - Parameters:
+  ///   - value: An instance of `Encodable` to be encoded to a document.
+  ///   - encoder: An encoder instance to use to run the encoding.
+  ///   - completion: A block to execute once the document has been successfully
+  ///                 written to the server. This block will not be called while
+  ///                 the client is offline, though local changes will be visible
+  ///                 immediately.
+
```
peterfriese, line 34: Nit: blank line

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import FirebaseFirestore
+
+extension DocumentReference {
+  /// Encodes an instance of `Encodable` and overwrites the encoded data
```
peterfriese, line 21: This comment doesn't match what the code does. Seems to have been copied from `setData`.

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -35,11 +35,9 @@ extension DocumentSnapshot {
...
   public func data<T: Decodable>(as type: T.Type,
                                  with serverTimestampBehavior: ServerTimestampBehavior = .none,
-                                 decoder: Firestore.Decoder? = nil) throws -> T? {
+                                 decoder: Firestore.Decoder? = nil) throws -> T {
     let d = decoder ?? Firestore.Decoder()
-    if let data = data(with: serverTimestampBehavior) {
-      return try d.decode(T.self, from: data, in: reference)
-    }
-    return nil
+    let data: Any = data(with: serverTimestampBehavior) ?? NSNull()
```
peterfriese, line 40: The comment says "By default, server-provided timestamps that have not yet been set to their final value will be returned as `NSNull`. Pass `serverTimestampBehavior` configure this behavior.", but I don't think this is what's happening here.

1. We do pass `serverTeimstampBehaviour` - that much is true
2. If `data(with:)` returns `nil`, we convert this to `NSNull()` - it's not entirely clear to me why.  The following two alternatives compile just the same:
    - `let data: Any = data(with: serverTimestampBehavior) ?? .none`
    -  `let data: Any = data(with: serverTimestampBehavior)`

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,53 @@
...
+  /// to the document referred by this `DocumentReference`. If no document exists,
+  /// it is created. If a document already exists, it is overwritten.
+  ///
+  /// See `Firestore.Encoder` for more details about the encoding process.
+  ///
+  /// - Parameters:
+  ///   - value: An instance of `Encodable` to be encoded to a document.
+  ///   - encoder: An encoder instance to use to run the encoding.
+  ///   - completion: A block to execute once the document has been successfully
```
peterfriese, line 30: "completion block" is a bit Objective-C-esque. I might rather use "closure" instead. But again - this doc comment seems to be copied from `setData` - let's update the comment over there.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,53 @@
...
+  ///                 immediately.
+
+  public func getDocument<T: Decodable>(as type: T.Type,
+                                        with serverTimestampBehavior: ServerTimestampBehavior =
+                                          .none,
+                                        decoder: Firestore.Decoder? = nil,
+                                        completion: @escaping (Result<T, Error>) -> Void) {
+    getDocument { snapshot, error in
+      guard let snapshot = snapshot else {
+        completion(.failure(error ?? FirestoreDecodingError.internal))
```
peterfriese, line 42: @schmidt-sebastian if we decide to introduce a new error constant to indicate internal errors, we should make sure to update other similar locations as well.

peterfriese (changes requested):
1. Docs needs to be cleaned up, 
2. I had a question around NSNull()
3. The tests should be updated
4. Adding a sample would be nice

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
mortenbekditlevsen: I'll have a look at the comments. The reason for using NSNull() is that the decoder can decode it into the .none case of an Optional. So if the type being decoded is optional, then it will decode just fine, if it is not, it will fail decoding.
Similar trick is used in the current Snapshot decoding for RTDB.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: Another solution could be to fatalError - in case the underlying api guarantees that either error or value is present.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: I can add that too. Only thing is that I don't know how to correctly annotate iOS 15 support on Xcode 13 and iOS 13+ support on Xcode 13.2. Does anyone have an example for this?

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: You're correct - I am in the process of updating.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: I'll fix this in `setData` and in the updated documentation.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: I guess something similar to a `fatalError()` or other means of trapping must be present in the auto-synthesised async interfaces from Objective-C methods with completion handlers with an optional value and an optional error. 
So on that account it would seem 'safe' just to `fatalError()` here too. What do you think?

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
mortenbekditlevsen: Now I've looked a bit more: The initial mention of `NSNull` is from the existing implementation and must be referring to what happens internally when getting data with serverside timestamp values in them.

That comment does not relate to the usage of passing `NSNull` to the unboxing in case there is no data. The latter is what I explained above.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: I now added this too - but currently only for iOS 15 and friends.

mortenbekditlevsen:
And here's the async use case version of the code:
```
  private func fetchBook(documentId: String) async {
    let docRef = db.collection("books").document(documentId)
    do {
      self.book = try await docRef.getDocument(as: Book.self)
      self.errorMessage = nil
    } catch {
      switch error {
      case DecodingError.typeMismatch(_, let context):
        self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
      case DecodingError.valueNotFound(_, let context):
        self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
      case DecodingError.keyNotFound(_, let context):
        self.errorMessage = "\(error.localizedDescription): \(context.debugDescription)"
      case DecodingError.dataCorrupted(let key):
        self.errorMessage = "\(error.localizedDescription): \(key)"
      default:
        self.errorMessage = "Error decoding document: \(error.localizedDescription)"
      }
    }
  }
```

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
paulb777: That's fine for now. We haven't found a good incantation that will also work for Xcode 13.0 because of its missing macOS support. We may end up requiring 13.1 or 13.2 for Swift libraries - and we can update all Swift Concurrency code globally, including this, when we make that decision.

paulb777 (commented):
Looks like a nice usability improvement.  Thanks @mortenbekditlevsen! What is the impact on migrating code using the existing library?

If others agree, let's try to run this through API review in January.

mortenbekditlevsen:
If you have code that decodes a snapshot into an entity, the result will no longer be an `Optional`, so an attempt at unwrapping will fail. 

There are two fixes: Inserting a question mark on the decidable type will give you the old behavior. 

The other fix - for instance in case the snapshot comes from a collection, meaning that you know the data exists, is to then skip the extra unwrapping. 

I should probably sketch out both scenarios in an example to provide a bit of guidance on the change.

mortenbekditlevsen:
I believe that all requested changes have been addressed. Would you mind a second review, @peterfriese ?

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,85 @@
...
+      let result = Result {
+        try snapshot.data(as: T.self,
+                          with: serverTimestampBehavior,
+                          decoder: decoder)
+      }
+      completion(result)
+    }
+  }
+
+  // TODO: How do you annotate that using Xcode 13.2 you can actually compile to earlier os versions
```
peterfriese, line 59: If I understood @paulb777 correctly (in [this comment](https://github.com/firebase/firebase-ios-sdk/pull/9101/files#r770006320), we're OK with targeting this at iOS 15 and friends for now, and update later once we have a better handle on the concurrency story for iOS 13.x). Let's remove this TODO.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,85 @@
...
+  #if compiler(>=5.5) && canImport(_Concurrency)
+    /// Fetches and decodes the document referenced by this `DocumentReference`.
+    ///
+    /// This allows users to retrieve a Firestore document and have it decoded to an instance of
+    /// caller-specified type.
+    /// ```swift
+    ///     let book = try await ref.getDocument(as: Book.self)
```
peterfriese, line 66: In [line 27](https://github.com/firebase/firebase-ios-sdk/pull/9101/files#diff-89265322325090eca3813c8141dd0a5d41d63ae0d0f96228b9303c98d9e0130cR27https://github.com/firebase/firebase-ios-sdk/pull/9101/files#diff-89265322325090eca3813c8141dd0a5d41d63ae0d0f96228b9303c98d9e0130cR27), you wrapped the code in a do/try/catch block. I think it'd be good to keep the examples symmetric - either add the do//try/catch  here as well, or remove it there.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,85 @@
...
+  ///       } catch {
+  ///         // Handle error
+  ///       }
+  ///     }
+  /// ```
+  /// - Parameters:
+  ///   - as: A `Decodable` type to convert the document fields to.
+  ///   - serverTimestampBehavior: Configures how server timestamps that have
+  ///     not yet been set to their final value are returned from the snapshot.
+  ///   - decoder: The decoder to use to convert the document. `nil` to use
```
peterfriese, line 38: The sentence is missing an end: 

> nil to use ... (what?)

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// This allows users to retrieve a Firestore document and have it decoded to an instance of
+    /// caller-specified type.
+    /// ```swift
+    ///     let book = try await ref.getDocument(as: Book.self)
+    /// ```
+    /// - Parameters:
+    ///   - as: A `Decodable` type to convert the document fields to.
+    ///   - serverTimestampBehavior: Configures how server timestamps that have
+    ///     not yet been set to their final value are returned from the snapshot.
+    ///   - decoder: The decoder to use to convert the document. `nil` to use
```
peterfriese, line 72: Same as above - sentence is missing an end.

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -35,11 +35,9 @@ extension DocumentSnapshot {
   ///     default decoder.
   public func data<T: Decodable>(as type: T.Type,
                                  with serverTimestampBehavior: ServerTimestampBehavior = .none,
-                                 decoder: Firestore.Decoder? = nil) throws -> T? {
+                                 decoder: Firestore.Decoder? = nil) throws -> T {
```
peterfriese, line 38: This is a breaking change. I compiled my FirestoreCodableSamples app against this PR, and every single `snapshot..data(as: )` now breaks:


CleanShot 2022-01-13 at 20 29 44@2x

peterfriese (changes requested):
I ❤️ the improved usability of this change. However, making `T` non-optional is a breaking change for all Firestore users. Can we try implementing this in a non-breaking way?

Other than that, a couple of small issues in the docs.

mortenbekditlevsen:
Hi Peter,
Thank you for the review! I'll get to fixing the comments ASAP. :-)

Regarding the breaking change:
A way to do it without breaking would be to supply a new API and keep the old with a deprecation warning.
A downside is that we have to come up with a new name for this API. :-)

Second best for users already using the API would be to provide a good guide with an update path.
This is of course only feasible if the extension is still considered to be in beta.
Regarding a guide:
Adding a question mark to your types - i.e. changing `.data(as: T.self)` -> `.data(as: T?.self)` would give the behavior of today, but many call sites could be examined to see whether the return type should actually be optional.

What do you think? New API and deprecation of existing - or rely on beta testers to be open to such a change?

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
mortenbekditlevsen: Right - should I try going for keeping the old api with a deprecation warning - and then see if I can come up with a new name for the replacement?
Any suggestions are very welcome. :-)

paulb777:
@ryanwilson Any suggestions about the best way to deprecate the current and naming of the new?

ryanwilson:
> What do you think? New API and deprecation of existing - or rely on beta testers to be open to such a change?

> @ryanwilson Any suggestions about the best way to deprecate the current and naming of the new?

Great questions, and thanks as always @mortenbekditlevsen for working through this with us.

Since this is beta and we're already at the ideal naming, I'm in favour of making the breaking change and add instructions / an explanation in the CHANGELOG (or a link to instructions). 

cc @schmidt-sebastian do you have a preference one way or the other?

schmidt-sebastian:
We can ship this as a breaking change. There is no reason to have `@beta` APIs if we don't take advantage of them. Let's just make sure that we always move in the right direction.

paulb777:
I updated the SwiftFormat we're using in #9239, so please rebase and update. Sorry about the inconvenience.

mortenbekditlevsen:
I rebased and reacted to the 'changes requested' by requesting a re-review.
Let me know what else I can do to move this forward.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
ryanwilson: Please correct me if I'm wrong but this introduces a fork in the functionality with the async/await version, right? The async/await version has the same issue where it expects the API to honour the contract or else it'll have unexpected behaviour. 

Firestore guarantees that only one of `snapshot` or `error` will be instantiated and we should expect that - I think it's fine to force unwrap the `error` here to avoid introducing a new public API for something that should never happen. If this ever crashes, I'm happy to take the blame 😛 

.... I love that this isn't something we have to think about when consuming Swift APIs 😄

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -35,11 +35,9 @@ public extension DocumentSnapshot {
   ///     default decoder.
   func data<T: Decodable>(as type: T.Type,
                           with serverTimestampBehavior: ServerTimestampBehavior = .none,
-                          decoder: Firestore.Decoder? = nil) throws -> T? {
+                          decoder: Firestore.Decoder? = nil) throws -> T {
```
ryanwilson, line 38: While we're making the breaking change anyways... think it's worthwhile to make the `decoder` parameter match RTDB and make it non-optional and have the default value be `Firestore.Decoder()`? cc @schmidt-sebastian @peterfriese

mortenbekditlevsen: I like that change. While technically breaking, it will only affect the calling code if it explicitly called it with a `nil` argument.

schmidt-sebastian: SGTM

mortenbekditlevsen: Implemented with latest commit.

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
ryanwilson: Updated note after chatting with @paulb777: we can test the async/await behaviour by hardcoding `nil/nil` and see what happens and try to have a reasonable behaviour between the two.

peterfriese:
Thanks for the changes, Morten!

Here are the next steps:

- [ ] Put together an internal API proposal (in progress, internal link: go/firebase-codable-explicit-typing)
- [ ] Once this is approved, merge the PR
- [ ] Update the code in my Codable sample project
- [ ] Reach out the community and make them aware of this breaking change
- [ ] Potentially update a bunch of StackOverflow answers I (and others) wrote

I will update this comment as I make progress.

Firestore/Swift/Source/Codable/DocumentReference+WriteEncodable.swift
```diff
@@ -27,7 +27,7 @@ public extension DocumentReference {
   /// - Parameters:
   ///   - value: An instance of `Encodable` to be encoded to a document.
   ///   - encoder: An encoder instance to use to run the encoding.
-  ///   - completion: A block to execute once the document has been successfully
+  ///   - completion: A closure to execute once the document has been successfully
   ///                 written to the server. This block will not be called while
```
peterfriese, line 31: Change "block" to "closure".

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -19,7 +19,7 @@ import FirebaseFirestore
 
 public extension DocumentSnapshot {
   /// Retrieves all fields in a document and converts them to an instance of
-  /// caller-specified type. Returns `nil` if the document does not exist.
+  /// caller-specified type.
   ///
   /// By default, server-provided timestamps that have not yet been set to their
   /// final value will be returned as `NSNull`. Pass `serverTimestampBehavior`
```
peterfriese, line 25: "to" missing at the end of the line:

Pass `serverTimestampBehavior` to

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
```diff
@@ -31,15 +31,12 @@ public extension DocumentSnapshot {
   ///   - type: The type to convert the document fields to.
   ///   - serverTimestampBehavior: Configures how server timestamps that have
   ///     not yet been set to their final value are returned from the snapshot.
-  ///   - decoder: The decoder to use to convert the document. `nil` to use
+  ///   - decoder: The decoder to use to convert the document. Defaults to use
   ///     default decoder.
```
peterfriese, line 35: Add "the":

the default decoder.

Firestore/Swift/Source/Codable/DocumentSnapshot+ReadDecodable.swift
peterfriese: I just looked into this particular spot again today. There are two issues I'd like to see resolved:

1) The mention of `NSNull` is a bit confusing. Can we phrase the doc comment differently? Maybe we can just say "The default behaviour is to leave timestamps untouched" or something like this. @schmidt-sebastian can you chime in here to provide guidance?

2) I *think* this code changes the default behaviour in case a document doesn't exist. Previously, when calling `getDocument` for a non-existing document, you would get `nil` as a result. Now, you get a `DecodingError.valueNotFound` instead ("The data couldn’t be read because it is missing." / "Cannot get keyed decoding container -- found null value instead."). To me, these are two different kinds of errors, and this might be misleading for developers. The error message indicates that something with the decoding process, when in fact the document was missing. This can be alleviated by passing in an optional type (`Book?.self`), and having to be more explicit about this might force developers to think more about their expectations when mapping documents, but I'd like to discuss this concern so we can properly document it.

peterfriese (approved):

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
peterfriese: Thanks for adding an inline comment documenting the behaviour of the code, @mortenbekditlevsen.

From my perspective, this LGTM.
@schmidt-sebastian, what's your perspective?

mortenbekditlevsen:
@peterfriese I updated the code, but in blind as I haven't been able to figure out how to run the codable integration tests.
I ran `./scripts/setup_spm_tests.sh` as per the documentation and I got a lot of new test schemes, but I can't find any that suggest they are running integration tests.
If I somehow missed something in my commit, I'd of course love to know how to build and run the tests - and please excuse my ignorance if I am missing something completely obvious. :-)

peterfriese (approved):

paulb777 (approved):
Great work! Thanks @mortenbekditlevsen and @peterfriese

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
```diff
@@ -0,0 +1,97 @@
...
+  /// }
+  /// ```
+  /// - Parameters:
+  ///   - as: A `Decodable` type to convert the document fields to.
+  ///   - serverTimestampBehavior: Configures how server timestamps that have
+  ///     not yet been set to their final value are returned from the snapshot.
+  ///   - decoder: The decoder to use to convert the document. Defaults to use
+  ///     default decoder.
+  ///   - completion: The closure to call when the document snapshot has been fetched and decoded.
+  func getDocument<T: Decodable>(as type: T.Type,
```
schmidt-sebastian, line 41: Ideally (but optional):
- Rewrap the comments for the `param` section
- Describe the behavior for nil/offline

schmidt-sebastian (approved):

Firestore/Swift/Source/Codable/DocumentReference+ReadDecodable.swift
mortenbekditlevsen: I'm sorry if I already asked this previously, but I'm afraid I don't remember the answer: do you use a tool to wrap comments?

mortenbekditlevsen: I found `fold` and tried rewrapping - I hope I did it as you envisioned it.
I added offline behavior description from Objective-C method.

I'm not sure what you mean by 'nil' in "Describe the behavior for nil/offline" - could you elaborate?

schmidt-sebastian: I am actually not sure how the tooling works - ./check/style.sh might wrap comments with lines that are too long, but I don't remember out of the top of my head. I certainly do a lot of manual rewrapping. 

As for describing the method behavior, what about:

```
  /// This method attempts to provide up-to-date data when possible by waiting
  /// for data from the server, but it may return cached data or fail if you are
  /// offline and the server cannot be reached. If `T` denotes an optional type, the 
  /// method returns a successful status with a value of `nil` for non-existing 
  /// documents.
```

mortenbekditlevsen: Perfect - thanks for the suggestion. Latest commit includes this extended comment. :-)

schmidt-sebastian (approved):

Issue #6876, opened 2020-10-30
Auth not Shared Correctly with App Groups


ccjernigan:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.1
  * macOS version: 10.15.7
  * Firebase SDK version: 7.0.0
  * Installation method: Swift Package
  * Firebase Component: Auth

### [REQUIRED] Step 2: Describe the problem
Firebase Auth is not properly shared between multiple bundles within the same app group.

I have a macOS app where the main app and a second process (in the same app bundle) share the same app group.  After the main app signs in, the second process launches via `SMLoginItemSetEnabled` as part of the setup process.  When the second process touches Firebase Auth, a system password prompt appears to grant keychain access (for a keychain item named for my API key).

#### Steps to reproduce:
1. Create an app that uses app groups and has a sub-bundle that runs in a second process
2. Sign into Firebase with the main app (call `Auth.auth().useUserAccessGroup(AppGroupId)` first)
3. Launch the second process, which should call `Auth.auth().useUserAccessGroup(AppGroupId)` and listens for auth changes `Auth.auth().addStateDidChangeListener({ (_, user) in
            ...
        })` 

Results:
Actual: Second process triggers a system keychain dialog asking for a password.
Expected: No keychain dialog should appear, since these are using a shared app group

I realize that macOS is "community supported" although this seems to be a bug that should be resolved to better support app extensions and other multiprocess use cases.

malcolmdeck:
Hey there! I've filed b/172260275 to track this issue internally.

ZComwiz:
@malcolmdeck could this be related to why on the Mac I am getting a token mismatch but on iOS with the same bundleID— it works fine when using Auth() for phone login?

OGmetamonkey:
Did anything come of this? When you say "shared app group", does that include an extension within the main app? If so, I'm having the same issue.[]( https://stackoverflow.com/questions/68199352/ios-extension-authentication-in-firestore)

OGmetamonkey:
> Hey there! I've filed b/172260275 to track this issue internally.

Any way you can tell me how I track this bug #?

kamilpowalowski:
@OGmetamonkey I have the same issue in the macOS extension. Did you manage to solve it?

rosalyntan:
Should be fixed in #9102.

PR #5603, opened 2020-05-13
Mac OS FIS: use Firebase App ID as a suffix to Keychain service name Mac OS

maksymmalyhin:
Addresses an issue detected during investigation of #5540.

- Use Firebase App ID from `GoogleServices-Info.plist` as a suffix for Keychain Service name on Mac OS to make it unique per Mac OS app (see [comments](https://github.com/firebase/firebase-ios-sdk/pull/5603/files#diff-837cf256e9fc57b3d2bdacf30e6771c3R464) for details).

NOTE: This change will lead to reset of Firebase Installation ID on Mac OS. Migration is not provided because for it we will have to access the old storage which will trigger the user alert described in #5540. 
The Identifiers on iOS/TVOS/WatchOS won't be affected.

PR #9102, opened 2021-12-15
[Auth] Fixed macOS extension keychain access by adding recommended kS…

### LINKED ISSUE #8950: SwiftUI/macOS Auth: "[App] wants to use your confidential information stored in 'com.firebase.FIRInstallations.installations' in your keychain."

Xcode version: Version 13.1 (13A1030d)
Firebase SDK version: 8.9.1
Installation method: Swift Package Manager
Firebase Component: Auth

#### Steps to reproduce:

When build on MacOS, user sees permissions popup saying, "[App] wants to use your confidential information stored in 'com.firebase.FIRInstallations.installations' in your keychain." 

#### Relevant Code:

```
// TODO(you): code here to reproduce the problem
```

kamilpowalowski:
Fixes #8950, #6876

I'm working on the app that used Firebase Auth for authentication. I have an iOS app with iOS Share Extension and macOS app with macOS Share Extension.
By adding Keychain Access Group, I can use authenticated accounts between the main iOS app and iOS Share Extension. But the same code doesn't work on macOS.
When the app is installed on macOS or when I'm trying to access it from macOS Share Extension, I always get a prompt to keychain access. I started to dig, and I found the keychain parameter https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain
from its documentation, we can find

> use the data protection key when adding, searching for, or deleting an item to which the kSecAttrAccessible or **kSecAttrAccessGroup** attributes apply.

> It's highly recommended that you set the value of this key to true for all keychain operations. This key helps to improve the portability of your code across platforms.

It can be added to all platforms:

> The data protection key affects operations only in macOS. Other platforms automatically behave as if the key is set to true, and ignore the key in the query dictionary. You can safely use the key on all platforms.

`kSecUseDataProtectionKeychain` key is also mentioned in https://developer.apple.com/documentation/security/ksecattraccessgroup documentation

> This attribute applies to macOS keychain items only if you also set a value of true for the kSecUseDataProtectionKeychain key, the kSecAttrSynchronizable key, or both.

But I have to limit it a bit with `@available` parameter.

I took screenshot from keychain without that key (only main app listed)
Screen Shot 2021-12-15 at 13 15 43

and a screenshot from keychain after key added (KeychainGroup access instead is listed)
Screen Shot 2021-12-15 at 13 15 36

### DIFF
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 - [changed] Added a `X-Firebase-GMPID` header to network requests. (#9046)
 - [fixed] Added multi-tenancy support to generic OAuth providers. (#7990)
+- [fixed] macOS Extension access to Shared Keychain by adding kSecUseDataProtectionKeychain recommended key (#6876)
 
 # 8.9.0
 - [changed] Improved error logging. (#8704)
--- a/FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
+++ b/FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
@@ -81,6 +81,9 @@ static NSString *kStoredUserCoderKey = @"firebase_auth_stored_user_coder_key";
   query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
   query[(__bridge id)kSecAttrService] = projectIdentifier;
   query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
+  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
+    query[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
+  }
   if (shareAuthStateAcrossDevices) {
     query[(__bridge id)kSecAttrSynchronizable] = (__bridge id)kCFBooleanTrue;
   }
@@ -125,6 +128,9 @@ static NSString *kStoredUserCoderKey = @"firebase_auth_stored_user_coder_key";
   query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
   query[(__bridge id)kSecAttrService] = projectIdentifier;
   query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
+  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
+    query[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
+  }
   if (shareAuthStateAcrossDevices) {
     query[(__bridge id)kSecAttrSynchronizable] = (__bridge id)kCFBooleanTrue;
   }
@@ -168,6 +174,9 @@ static NSString *kStoredUserCoderKey = @"firebase_auth_stored_user_coder_key";
   query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
   query[(__bridge id)kSecAttrService] = projectIdentifier;
   query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
+  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
+    query[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
+  }
 
   return [self.keychainServices removeItemWithQuery:query error:outError];
 }
--- a/FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
+++ b/FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
@@ -75,19 +75,9 @@ static NSString *kStoredUserCoderKey = @"firebase_auth_stored_user_coder_key";
              shareAuthStateAcrossDevices:(BOOL)shareAuthStateAcrossDevices
                        projectIdentifier:(NSString *)projectIdentifier
                                    error:(NSError *_Nullable *_Nullable)outError {
-  NSMutableDictionary *query = [[NSMutableDictionary alloc] init];
-  query[(__bridge id)kSecClass] = (__bridge id)kSecClassGenericPassword;
-
-  query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
-  query[(__bridge id)kSecAttrService] = projectIdentifier;
-  query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
-  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
-    query[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
-  }
-  if (shareAuthStateAcrossDevices) {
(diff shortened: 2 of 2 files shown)


ncooke3 (commented):
Thanks a lot for the investigation and contribution @kamilpowalowski! I've approved the CI workflows. 

Have you completed the [Google CLA](https://cla.developers.google.com/clas)?

kamilpowalowski:
@ncooke3 No problem :) I need that for my own app so I was motivated.

Yes, I completed Google CLA yesterday.

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
```diff
@@ -81,6 +81,9 @@ - (FIRUser *)getStoredUserForAccessGroup:(NSString *)accessGroup
   query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
   query[(__bridge id)kSecAttrService] = projectIdentifier;
   query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
+  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
+    query[(__bridge id)kSecUseDataProtectionKeychain] = (__bridge id)kCFBooleanTrue;
+  }
```
ncooke3, line 86: So IIUC: when the `kSecUseDataProtectionKeychain` is added to the query, the prompt on macOS is prevented. When it's not (i.e. we are on running on pre-macOS 10.15), the prompt will appear?

FirebaseAuth/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 - [changed] Added a `X-Firebase-GMPID` header to network requests.
+- [fixed] macOS Extension access to Shared Keychain by adding kSecUseDataProtectionKeychain recommended key (#8950, #6876)
```
ncooke3, line 3: Based on #8950's title, the keychain prompt was mentioning `com.firebase.FIRInstallations.installations`.

Could you confirm the bundle ID(s) mentioned in the keychain prompts you are seeing (without this fix)?

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
kamilpowalowski: @ncooke3 yes. If macOS pre 10.15 don't behave differently, then on that system, after reinstall/update or when the account is checked from an extension there will be a keychain access prompt.

FirebaseAuth/CHANGELOG.md
kamilpowalowski: #8950 is a bit misleading because right now a keychain entry name is an API KEY, not a `com.firebase.FIRInstallations.installations`. Give me a few minutes, I'll find a commit that changed this.
So for me, a prompt was asking to access to `AlzaSy...` which is my app Firebase API KEY.

kamilpowalowski: I found it https://github.com/firebase/firebase-ios-sdk/pull/5603. It was a critical PR. Otherwise, every macOS application with Firebase Auth tries to use the same keychain entry to store app auth information.

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
kamilpowalowski: For <10.15 a working sulution may be to set `shareAuthStateAcrossDevices` to true. because [AccessGroup documentation](https://developer.apple.com/documentation/security/ksecattraccessgroup) says 

> This attribute applies to macOS keychain items only if you also set a value of true for the kSecUseDataProtectionKeychain key, the kSecAttrSynchronizable key, or both.

FirebaseAuth/CHANGELOG.md
ncooke3: I see- thanks! I was wondering because I think this PR's solution may be applied to other parts of Firebase that use the keychain. Again, great investigation!

kamilpowalowski: You are right. It should be applied everywhere where `kSecAttrAccessGroup` is used, but `FIRAuthStoredUserManager` is the only place I found.

ncooke3, line 3: Hi, could you remove the #8950 part from the changelog entry, please? Reason being is that this patch fixes Auth's keychain access but I believe we still need to investigate the root cause for the `com.firebase.FIRInstallations.installations`-related prompts, which may be a separate fix.

```suggestion
- [fixed] macOS Extension access to Shared Keychain by adding kSecUseDataProtectionKeychain recommended key (#6876)
```

ncooke3 (approved):
LGTM. I added one request below. Thanks again @kamilpowalowski!

FirebaseAuth/CHANGELOG.md
kamilpowalowski: Done, thank you.

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
```diff
@@ -81,6 +81,9 @@ - (FIRUser *)getStoredUserForAccessGroup:(NSString *)accessGroup
   query[(__bridge id)kSecAttrAccessGroup] = accessGroup;
   query[(__bridge id)kSecAttrService] = projectIdentifier;
   query[(__bridge id)kSecAttrAccount] = kSharedKeychainAccountValue;
+  if (@available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)) {
```
rosalyntan, line 84: From the [docs](https://developer.apple.com/documentation/security/ksecusedataprotectionkeychain): "The data protection key affects operations only in macOS. Other platforms automatically behave as if the key is set to true, and ignore the key in the query dictionary. You can safely use the key on all platforms."

Do you mind adding a comment with this information wherever `kSecUseDataProtectionKeychain` is used?

rosalyntan (approved):
Thank you for the contribution! One minor comment, and there's a new merge conflict in the changelog that needs to be resolved before merging this, but otherwise LGTM!

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
kamilpowalowski: @rosalyntan Will this be a problem if I did that (and rebase to resolve conflicts) beginning next week? Until that time, I only have access to the iPad, which is not a convenient tool for code editing and git operations.

paulb777: Thanks @kamilpowalowski It's fine to do next week. Our next release from master will be in January.

ncooke3 (approved):

FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m
ncooke3: Thanks for the additional cleanup @kamilpowalowski!

rosalyntan: Thank you, @kamilpowalowski! LGTM.

nidegen:
@kamilpowalowski do you have an idea by any chance if this would also fix a similar bug on tvOS?

kamilpowalowski:
@nidegen, I don’t think so. It fixes a behavior related strictly to macOS. On macOS, you have access to the whole keychain. On other systems, you shouldn't have that problems. But (if possible) you should try to set `shareAuthStateAcrossDevices` to true. It may solve your problem.

PR #9104, opened 2021-12-16
Fireperf: dropping app start events when they are prewarmed.

visumickey:


FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -66,12 +68,24 @@ @implementation FPRAppActivityTracker
...
   gAppStartCPUGaugeData = fprCollectCPUMetric();
   gAppStartMemoryGaugeData = fprCollectMemoryMetric();
   [[NSNotificationCenter defaultCenter] addObserver:self
                                            selector:@selector(windowDidBecomeVisible:)
                                                name:UIWindowDidBecomeVisibleNotification
                                              object:nil];
+  
+  [[NSNotificationCenter defaultCenter] addObserver:self
+                                           selector:@selector(applicationDidFinishLaunching:)
+                                               name:UIApplicationDidFinishLaunchingNotification
```
jeremyjiang-dev, line 87: Why are you comparing the double dispatch time with `UIApplicationDidFinishLaunchingNotification`? The Hub app is using the `ApplicationWillFinishLaunchingWithUserActivity`

jeremyjiang-dev (approved):
Thanks for fixing the bug and adding tests!

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
```diff
@@ -50,6 +51,9 @@ NS_EXTENSION_UNAVAILABLE("Firebase Performance is not supported for extensions."
 /** Current running state of the application. */
 @property(nonatomic, readonly) FPRApplicationState applicationState;
 
+/** Firebase Remote Configuration object */
```
visumickey, line 54: This is Firebase Performance configuration object. Remote Config configuration object is abstracted within.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
```diff
@@ -50,6 +51,9 @@ NS_EXTENSION_UNAVAILABLE("Firebase Performance is not supported for extensions."
...
+@property(nonatomic) FPRConfigurations *configurations;
```
visumickey, line 55: Should this be in the header? I don't think anyone else needs to use the app activity tracker's dependency on FPRConfigurations

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -66,12 +73,26 @@ @implementation FPRAppActivityTracker
 + (void)load {
   // This is an approximation of the app start time.
   appStartTime = [NSDate date];
+
+  if (![[[NSBundle mainBundle] objectForInfoDictionaryKey:@"fireperf_disable_dd"] boolValue]) {
```
visumickey, line 77: Are we just checking for an entry in the PLIST or are we checking for the value of the key in the plist?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
visumickey: WillFinishLaunching is something that cannot be piggybacked by an SDK as we do not have any notifications in that aspect. So, we are using the closest, which is applicationDidFinishLaunchingWithOptions.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
++ (BOOL)isPrewarmAvailable {
+  if (![[GULAppEnvironmentUtil applePlatform] isEqualToString:@"ios"]) {
+    return NO;
+  }
+  NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
+  if ([systemVersion length] == 0) {
+    return NO;
+  }
+  NSArray *versionComponents = [systemVersion componentsSeparatedByString:@"."];
+  NSInteger majorVersion = [versionComponents[0] integerValue];
```
visumickey, line 162: Please add a check to make sure the array length is greater than 0.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
   return self.backgroundSessionTrace;
 }
 
++ (BOOL)isPrewarmAvailable {
```
visumickey, line 153: Can we add some comments about what this method does?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+  NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
+  if ([systemVersion length] == 0) {
+    return NO;
+  }
+  NSArray *versionComponents = [systemVersion componentsSeparatedByString:@"."];
+  NSInteger majorVersion = [versionComponents[0] integerValue];
+  return majorVersion >= 15;
+}
+
+- (BOOL)isActivePrewarmEnabled {
```
visumickey, line 166: Please add comments

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+  return ([self.configurations prewarmDetectionMode] == 1 ||
```
visumickey, line 167: Can we replace these constants with something more understandable. For example: you can create a macro/enums for these values and we can use that check here. A check for the value of 1 or 3 is not so trivial to understand.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+          [self.configurations prewarmDetectionMode] == 3);
+}
+
+- (BOOL)isApplicationPreWarmed {
+  if (![FPRAppActivityTracker isPrewarmAvailable]) {
+    return NO;
+  }
+
+  // Force drop all events according to RC flag
+  if ([self.configurations prewarmDetectionMode] == 0) {
```
visumickey, line 182: Same comment as above.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+    return YES;
+  }
+
+  NSDictionary<NSString *, NSString *> *environment = [NSProcessInfo processInfo].environment;
+  if ([self isActivePrewarmEnabled] && environment[@"ActivePrewarm"] != nil &&
```
visumickey, line 187: Checking the environment variable seem to be happening more than once. Is there a way to abstract this?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+      [environment[@"ActivePrewarm"] isEqualToString:@"1"]) {
+    [self.activeTrace incrementMetric:kFPRPrewarmedAppStartIdentifiedByActivePrewarm byInt:1];
+    return YES;
+  }
+
+  if ([self isDoubleDispatchEnabled] &&
+      [doubleDispatchTime compare:applicationFinishLaunchTime] == NSOrderedAscending) {
+    [self.activeTrace incrementMetric:kFPRPrewarmedAppStartIdentifiedByDoubleDispatch byInt:1];
```
visumickey, line 195: Can you verify if self.activeTrace is available at this time? Most likely the answer is yes, but we can make sure we are not missing anything here.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,55 @@ - (FIRTrace *)activeTrace {
...
+  }
+
+  NSDictionary<NSString *, NSString *> *environment = [NSProcessInfo processInfo].environment;
+  if ([self isActivePrewarmEnabled] && environment[@"ActivePrewarm"] != nil &&
+      [environment[@"ActivePrewarm"] isEqualToString:@"1"]) {
+    [self.activeTrace incrementMetric:kFPRPrewarmedAppStartIdentifiedByActivePrewarm byInt:1];
+    return YES;
+  }
+
+  if ([self isDoubleDispatchEnabled] &&
```
visumickey, line 193: Should we also add a check got [self isActivePrewarmEnabled] here?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -173,10 +251,13 @@ - (void)appDidBecomeActiveNotification:(NSNotification *)notification {
       // happens a lot later.
       // Dropping the app start trace in such situations where the launch time is taking more than
       // 60 minutes. This is an approximation, but a more agreeable timelimit for app start.
-      if (currentTimeSinceEpoch - startTimeSinceEpoch < gAppStartMaxValidDuration) {
+      if ((currentTimeSinceEpoch - startTimeSinceEpoch < gAppStartMaxValidDuration) &&
+          ![self isApplicationPreWarmed]) {
         [self.appStartTrace stop];
       } else {
         [self.appStartTrace cancel];
+        [self.activeTrace incrementMetric:kFPRPrewarmedAppStartDroppedWithDuration
```
visumickey, line 259: I don't think it is a good idea to mark the foreground/background session trace with app start duration. We should find another mechanism to record this if required.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Current implementation checks for value. The value must be a `boolean` `1`. Checking for just entry is an alternative, what would be best for this case?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: `isActivePrewarmEnabled` is already checked above

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Yes, activeTrace is set earlier in the same method that calls `isApplicationPreWarmed`

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Will move this discussion to another PR.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,66 @@ - (FIRTrace *)activeTrace {
...
+ */
+- (BOOL)isApplicationPreWarmed {
+  if (![FPRAppActivityTracker isPrewarmAvailable]) {
+    return NO;
+  }
+
+  // Force drop all events according to RC flag
+  if ([self.configurations prewarmDetectionMode] == DropAllEvents) {
+    return YES;
+  }
```
jeremyjiang-dev, line 197: This method should only answer `isApplicationPreWarmed` instead of answering if events should be dropped or not. Can we move this call the caller of `isApplicationPreWarmed`?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
```diff
@@ -35,6 +35,15 @@ typedef NS_ENUM(NSInteger, FPRApplicationState) {
   FPRApplicationStateBackground,
 };
 
+/** Different modes of prewarm-detection */
+typedef NS_ENUM(NSInteger, PrewarmDetectionMode) {
+  DropAllEvents = 0,
+  OnlyActivePrewarm = 1,
+  OnlyDoubleDispatch = 2,
+  EitherActivePrewarmOrDoubleDispatch = 3,
+  KeepAllEvents = 4
```
jeremyjiang-dev, line 44: Under what scenario, we will use `KeepAllEvents`? Does it mean that we will include all prewarmed events?

jeremyjiang-dev (commented):
Could you update the title and click `ready for review` when this PR is ready?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Moved to a new method `isAppStartEnabled`, called right before `isApplicationPreWarmed`

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
leotianlizhan: Yes, it means all app start are included, essentially reverting back to pre-iOS15 SDK. There's not much scenario I suppose where this would be used, probably only if Apple completely reverts prewarming feature. This is there just for completion so RC flag covers all possibilities.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Discussed and decided that we won't be capturing duration of dropped `_as`

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
visumickey: Typically Enums have prefixes for better understanding:

For example: For the above case, my recommendation of Enums would be:

```
typedef NS_ENUM(NSInteger, PrewarmDetectionMode) {
  PrewarmDetectionModeNone = 0,
  PrewarmDetectionModeActivePrewarm = 1,
  PrewarmDetectionModeDoubleDispatch = 2,
  PrewarmDetectionModeActivePrewarmOrDoubleDispatch = 3,
  PrewarmDetectionModeAll = 4
}
```

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -24,7 +24,11 @@
 #import "FirebasePerformance/Sources/Timer/FIRTrace+Internal.h"
 #import "FirebasePerformance/Sources/Timer/FIRTrace+Private.h"
 
+#import <GoogleUtilities/GULAppEnvironmentUtil.h>
+
 static NSDate *appStartTime = nil;
+static NSDate *doubleDispatchTime = nil;
+static NSDate *applicationFinishLaunchTime = nil;
```
visumickey, line 31: can we change this to applicationDidFinishLaunchTime? (To avoid the ambiguity of applicationDidFinishLaunchTime and applicationWillFinishLaunchTime)

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
visumickey: Since we are checking for a bool value, I would say lets check for the booValue to return to "Yes".

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -83,6 +104,13 @@ + (void)windowDidBecomeVisible:(NSNotification *)notification {
                                                 object:nil];
 }
 
++ (void)applicationDidFinishLaunching:(NSNotification *)notification {
+  applicationFinishLaunchTime = [NSDate date];
+  [[NSNotificationCenter defaultCenter] removeObserver:self
+                                                  name:UIApplicationDidFinishLaunchingNotification
+                                                object:nil];
```
visumickey, line 111: Just make sure this is formatted correctly.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,68 @@ - (FIRTrace *)activeTrace {
...
 }
 
+/**
+ * Checks if prewarming is available for the platform on current device.
+ * It is available when running iOS 15 and above.
+ *
+ * @return true if the platform could prewarm apps on the current device
+ */
++ (BOOL)isPrewarmAvailable {
+  if (![[GULAppEnvironmentUtil applePlatform] isEqualToString:@"ios"]) {
```
visumickey, line 160: Curious if they have an ENUM to check for iOS. With a a string equality check, the check would fail if they change the case of the value. If there are not other options available, lets make a comparison that ignores case sensitivity.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,68 @@ - (FIRTrace *)activeTrace {
...
+    return NO;
+  }
+  NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
+  if ([systemVersion length] == 0) {
+    return NO;
+  }
+  return [systemVersion compare:@"15" options:NSNumericSearch] != NSOrderedAscending;
```
visumickey, line 167: This check concerns me. Are there any other better way to do this? Doing a numeric search of "15" is super concerning. Eg: if we have a version that is 14.15.1, this check would return "YES" and it should not.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,68 @@ - (FIRTrace *)activeTrace {
...
+/**
+ Checks if the current app start is a prewarmed app start
+ */
+- (BOOL)isApplicationPreWarmed {
+  if (![FPRAppActivityTracker isPrewarmAvailable]) {
+    return NO;
+  }
+
+  NSDictionary<NSString *, NSString *> *environment = [NSProcessInfo processInfo].environment;
+  BOOL isActivePrewarmVariableSetToYes = [environment[@"ActivePrewarm"] boolValue];
```
visumickey, line 202: `isActivePrewarmEnvVariableSet` instead of `isActivePrewarmVariableSetToYes`.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
visumickey: All the previous checks have more than one, so it is not clear if there is an explicit check for isActivePrewarmEnabled. May be I'm missing something here.

FirebasePerformance/Sources/Configurations/FPRConfigurations.h
```diff
@@ -62,6 +62,18 @@ FOUNDATION_EXTERN FPRConfigName kFPRConfigInstrumentationEnabled;
...
+ * Controls how the SDK detects and drops prewarmed app start traces:
+ * 0 = drop all app start traces;
+ * 1 = drop only if ActivePrewarm returns true;
+ * 2 = drop only if double dispatch returns true;
+ * 3 = drop if either ActivePrewarm or double dispatch returns true;
+ * 4 = do not drop any app start traces.
+ *
+ * @return an integer representing the detection mode of prewarm app start traces.
+ */
+- (uint32_t)prewarmDetectionMode;
```
visumickey, line 75: Can we make this function return the ENUM as well? Typically the enum should be defined in the FPRConfigurations.h and we use those definitions in the whole of SDK. The ENUM definition should not be in the implementation file of FPRAppActivityTracker.m

FirebasePerformance/Tests/Unit/FPRAppActivityTrackerTest.m
```diff
@@ -193,4 +205,88 @@ - (void)testApplicationStateManagement {
   XCTAssertEqual(appTracker.applicationState, FPRApplicationStateBackground);
 }
 
+- (void)testIsApplicationPrewarmedReturnsYesBecauseOfDoubleDispatch {
```
visumickey, line 208: Can you add comments on the tests as to what they are doing?

FirebasePerformance/Tests/Unit/FPRAppActivityTrackerTest.m
```diff
@@ -193,4 +205,88 @@ - (void)testApplicationStateManagement {
...
+  id mockAppTracker = OCMPartialMock([FPRAppActivityTracker sharedInstance]);
+  OCMStub([mockAppTracker isPrewarmAvailable]).andReturn(YES);
+  OCMStub([mockAppTracker isDoubleDispatchEnabled]).andReturn(YES);
+  OCMStub([mockAppTracker isActivePrewarmEnabled]).andReturn(NO);
+
+  [FPRAppActivityTracker load];
```
visumickey, line 214: This should be called automatically. isn't it?

FirebasePerformance/Tests/Unit/FPRAppActivityTrackerTest.m
```diff
@@ -193,4 +205,88 @@ - (void)testApplicationStateManagement {
...
+  [[NSNotificationCenter defaultCenter]
+      postNotificationName:UIApplicationDidFinishLaunchingNotification
+                    object:[UIApplication sharedApplication]];
+
+  XCTAssertTrue([mockAppTracker isApplicationPreWarmed]);
+}
+
+- (void)testIsApplicationPrewarmedReturnsYesBecauseOfActivePrewarm {
```
visumickey, line 222: Just a thought on the testing function names - Consider shortening the names to a simple format.

go/unit-test-practices#naming

Typically the format should be "function/featture_scenario_expectedOutcome".

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Will change to case insensitive compare

FirebasePerformance/Sources/Configurations/FPRConfigurations.h
leotianlizhan: Will move it to `FPRConfigurations.h`

FirebasePerformance/Tests/Unit/FPRAppActivityTrackerTest.m
leotianlizhan: this test actually fails if `+load()` isn't manually called. Somehow `appDidFinishLaunching` callback happens before `+load()` in the unit test.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.h
leotianlizhan: Done

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -66,12 +73,26 @@ @implementation FPRAppActivityTracker
 + (void)load {
   // This is an approximation of the app start time.
   appStartTime = [NSDate date];
+
+  if ([[[NSBundle mainBundle] objectForInfoDictionaryKey:@"fireperf_disable_dd"] boolValue] == NO) {
```
visumickey, line 77: minor nit: Add comments to qualify what are we doing below and why? This is a critical piece of code for someone who wants take a look at the code.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,70 @@ - (FIRTrace *)activeTrace {
...
+- (BOOL)isAppStartEnabled {
+  return [self.configurations prewarmDetectionMode] != PrewarmDetectionModeNone;
+}
+
+/**
+ RC flag for enabling prewarm-detection using ActivePrewarm environment variable
+ */
+- (BOOL)isActivePrewarmEnabled {
+  PrewarmDetectionMode mode = [self.configurations prewarmDetectionMode];
+  return (mode == PrewarmDetectionModeActivePrewarm ||
```
visumickey, line 182: Is ActivePrewarmEnabled when the mode is PrewarmDetectionModeAll?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -121,6 +150,70 @@ - (FIRTrace *)activeTrace {
...
+          mode == PrewarmDetectionModeActivePrewarmOrDoubleDispatch);
+}
+
+/**
+ RC flag for enabling prewarm-detection using double dispatch method
+ */
+- (BOOL)isDoubleDispatchEnabled {
+  PrewarmDetectionMode mode = [self.configurations prewarmDetectionMode];
```
visumickey, line 190: Is double disapatch enabled when the mode is PrewarmDetectionModeAll?

FirebasePerformance/Sources/Configurations/FPRConfigurations.h
```diff
@@ -16,6 +16,15 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** Different modes of prewarm-detection */
+typedef NS_ENUM(NSInteger, PrewarmDetectionMode) {
```
visumickey, line 20: nit: Can you add comments to the individual ENUMs here? I felt a bit lost when reviewing the code. It would have helped to have contextual reference on the value for these ENUMs

FirebasePerformance/Sources/Configurations/FPRConfigurations.m
```diff
@@ -300,6 +300,20 @@ - (int)logSource {
   return logSource;
 }
 
+- (PrewarmDetectionMode)prewarmDetectionMode {
+  // 0 means use neither and drop all _as events
+  // 1 means use only ActivePrewarm
+  // 2 means use only double dispatch
+  // 3 means use both ActivePrewarm and double dispatch
+  // 4 means keep all _as events
```
visumickey, line 308: Is this a repeat of the comments in the headers. Feel free to remove it.

jeremyjiang-dev (approved):
Great work!

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: It depends on how we define "None" and "All". in the old naming, 0 = "DropAllEvents", 4 = "KeepAllEvents". If "None" means no events are allowed and "All" means all events are allowed, then it works. If "None" means prewarm-detection is not going to identify any prewarm and "All" means prewarm-detection is going to identify all events as prewarm, then it we need to swap the ENUM names or swap 0 and 4.

visumickey: Perfect. Makes sense. Can we make it clear in the code on this definition and align the code accordingly? This was not evident when I looked at the code. May be better comments in the enum definition or the comments above this method can help.

leotianlizhan: Added comments to ENUM definitions. Changed "None" and "All" to "KeepNone" and "KeepAll" to be more clear.

PR #9111, opened 2021-12-17
[Heartbeat Logging] Add API for flushing today's heartbeat 

ncooke3:
**Context**
Both v1 and v2 clients will use the new heartbeat library. V1 client payloads only support the "heartbeat code" payload (see [FIRHeartbeatInfo.h](https://github.com/firebase/firebase-ios-sdk/blob/526f8ca35f9dcf3f1b0aeb94c995f058691b2c26/FirebaseCore/Sources/Private/FIRHeartbeatInfo.h#L22)). 

To allow **v1** clients to upload with the new new heartbeat library, a new API must be added to allow flushing only today's heartbeat into a payload for sending. 

---
Remaining work (planned for next PR):
- refactor HeartbeatStorage class for better thread-safety
- update HeartbeatStorage tests

#no-changelog


HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -108,12 +105,41 @@ public final class HeartbeatController {
...
+      return heartbeatsBundle
+    }
+
+    // `todaysHeartbeat`'s value is updated in the `readWriteBlock`.
+    guard let todaysHeartbeat = todaysHeartbeat else {
+      return HeartbeatsPayload.emptyPayload
+    }
+
+    return todaysHeartbeat.makeHeartbeatsPayload()
+  }
```
ncooke3, line 144: I'm not too keen on this approach but I think it's ok for now. Specifically, I don't like how the `todaysHeartbeat` variable is set within the `readAndWriteSync` block. I believe it's safe in this case because the block is executed synchronously, but I'd prefer if the storage API supported this behavior in a safer way.

I'm going to continue with testing and hope to refine after thinking about it more.

HeartbeatLogging/Tests/Unit/HeartbeatsBundleTests.swift
```diff
@@ -0,0 +1,322 @@
+// Copyright 2021 Google LLC
```
ncooke3, line 1: This file is a rename of HeartbeatLogging/Tests/Unit/HeartbeatInfoTests.swift with added tests for new remove API.

HeartbeatLogging/Sources/HeartbeatController.swift
ncooke3: I have an idea for how to refactor HeartbeatStorage class. To keep the PR smaller, I will follow up with a HeartbeatStorage refactor PR after this one.

HeartbeatLogging/Tests/Unit/HeartbeatsBundleTests.swift
ncooke3: I used `git mv` in 5d8f6db to keep its history but it didn't seem to work.

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -108,12 +105,41 @@ public final class HeartbeatController {
...
   }
+
+  /// Synchronously flushes the heartbeat for today.
+  ///
+  /// If no heartbeat was logged today, the returned payload is empty.
+  ///
+  /// - Note: This API is thread-safe.
+  /// - Returns: A heartbeats payload for the flushed heartbeat.
+  @discardableResult
+  public func flushHeartbeatFromToday() -> HeartbeatsPayload {
```
ncooke3, line 124: reviewers: The existing v1 ObjC (`+ FIRHeartbeatInfo heartbeatCodeForTag`) API's implementation will instead call into this API. I will post ObjC wrapper in next PR.

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -108,12 +105,41 @@ public final class HeartbeatController {
...
+    let todaysDate = dateProvider()
+    var todaysHeartbeat: Heartbeat?
+
+    storage.readAndWriteSync { heartbeatsBundle in
+      guard var heartbeatsBundle = heartbeatsBundle else {
+        return nil // Storage was empty.
+      }
+
+      todaysHeartbeat = heartbeatsBundle.removeHeartbeat(from: todaysDate)
```
paulb777, line 133: How does `todaysHeartbeat` get used since there is a return statement following?

HeartbeatLogging/Sources/HeartbeatsBundle.swift
```diff
@@ -72,6 +72,30 @@ struct HeartbeatsBundle: Codable, HeartbeatsPayloadConvertible {
...
+  /// Removes the heartbeat associated with the given date.
+  /// - Parameter date: The date of the heartbeat needing removal.
+  /// - Returns: The heartbeat that was removed or `nil` if there was no heartbeat to remove.
+  @discardableResult
+  mutating func removeHeartbeat(from date: Date) -> Heartbeat? {
+    var removedHeartbeat: Heartbeat?
+
+    var poppedHeartbeats: [Heartbeat] = []
+
+    while let poppedHeartbeat = buffer.pop() {
```
paulb777, line 84: Why a stack data structure that requires popping and repushing to extract versus a list in which extracting would be simpler?

paulb777 (approved):

HeartbeatLogging/Sources/HeartbeatController.swift
ncooke3: `todaysHeartbeat`  is defined in the outer scope (relative to the block in which it is assigned). 

The block captures  `todaysHeartbeat` and assigns it. The return after the assignment applies to the block's scope . The block is executed synchronously so by line 138, The read and write operation is done and `todaysHeartbeat` has been assigned.

HeartbeatLogging/Sources/HeartbeatsBundle.swift
ncooke3: Good question. 2 factors informed this decision. 

- 1. The underlying heartbeat data structure is a ring buffer* and it internally manages a fixed capacity array and index (`tailIndex`) where it pushes/pops from. The buffer does conform to Swift's `Sequence` so we could filter the heartbeat out but then we are left with an array that is has one less capacity than the buffer it came from and we arent sure where the index index should be. So a Ringbuffer initializer would need to be added that takes in a list, a capacity, and a tail index and we'd have to rebuild the data structure through that. 

- 2. Because of reason one, I figured it'd be easier to add an API to the RingBuffer type. I added `pop` API because it is another API you would expect a ring buffer implementation to have. With `push` and `pop` you can remove elements from the buffer without risking the insertion index being compromised.

FWIW, 99% of the time, the last heartbeat in the buffer will be the heartbeat from today so this logic will take the happiest path (meaning first heartbeat popped is the heartbeat from today). The reason it is written this way is to handle the 1% edge case where the device's system time has been modified by end user and the expected ordering breaks down. Because of that edge case, we cant make assumption that last heartbeat inserted is heartbeat from today. 

It's very important that the heartbeat we ultimately remove is from today since the date will be assigned by the backend (this is all for v1 clients that send FIRHeartbeatInfo codes.)

> *Ring buffer was chosen since we want to store a fixed capacity of heartbeats where the oldest heartbeat is overwritten when the capacity is reached. I thought it worked well for this use case since we append based on date. Plus it has constant time operations.

HeartbeatLogging/Sources/HeartbeatController.swift
```diff
@@ -108,12 +105,41 @@ public final class HeartbeatController {
...
+
+      return heartbeatsBundle
+    }
+
+    // Note that `todaysHeartbeat` is updated in the above read/write block.
+    guard let todaysHeartbeat = todaysHeartbeat else {
```
paulb777, line 139: I think an if - else might be clearer than assigning a var todaysHeartbeat to a let todaysHeartbeat.

HeartbeatLogging/Sources/HeartbeatController.swift
paulb777: I see. Thanks!

paulb777 (approved):

HeartbeatLogging/Sources/HeartbeatController.swift
ncooke3: Reworked in ead3a8c

PR #9117, opened 2021-12-20
[Heartbeat Logging] Add ObjC interop API and Firebase Core integration

ncooke3:
**Context**
The heartbeat logging library is written in Swift. Integrating it into Firebase Core means adding:
- adding objc interop types (see `_ObjC_HeartbeatController.swift` and `_ObjC_HeartbeatsPayload.swift`)
- adding Firebase interface to call into heartbeat logging library, injecting Firebase specific info when doing so

**TODOs**
- [x] Add unit tests for Firebase Core's new `FIRHeartbeatLogger` class
- [x] Organize structure for CocoaPods can run the **Swift** library's tests too

#no-changelog


HeartbeatLogging/Sources/StorageFactory.swift
```diff
@@ -14,7 +14,7 @@
 
 import Foundation
 
-private enum Constants {
+enum Constants {
```
ncooke3, line 17: `private` access removed since these constants are now used in HeartbeatLogging/Sources/HeartbeatLoggingTestUtils.swift

HeartbeatLogging/Sources/HeartbeatLoggingTestUtils.swift
```diff
@@ -0,0 +1,76 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import XCTest
+
+@objc(FIRHeartbeatLoggingTestUtils)
+@objcMembers
+public class HeartbeatLoggingTestUtils: NSObject {
```
ncooke3, line 20: When integrating with FirebaseCore, I realized during tests that it'd be helpful to have the `assertEqualPayloadStrings` custom assertion so i created this file to add it to the `HeartbeatLogging` module.

HeartbeatLogging/Sources/_ObjC_HeartbeatController.swift
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+import Foundation
+
+/// An object that provides API to log and flush heartbeats from a synchronized storage container.
+@objc(FIRHeartbeatController)
+@objcMembers
+public class _ObjC_HeartbeatController: NSObject {
+  /// The underlying Swift object.
+  private let heartbeatController: HeartbeatController
```
ncooke3, line 22: The heartbeat logging module has two native swift types that need to be bridged over for use outside module (i.e. FirebaseCore).

I prefixed the swift type with _ObjC_ to "hide" it from Swift clients.

HeartbeatLogging/Sources/_ObjC_HeartbeatsPayload.swift
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+import Foundation
+
+/// A model object representing a payload of heartbeat data intended for sending in network requests.
+@objc(FIRHeartbeatsPayload)
+@objcMembers
+public class _ObjC_HeartbeatsPayload: NSObject, HTTPHeaderRepresentable {
+  /// The underlying Swift structure.
+  private let heartbeatsPayload: HeartbeatsPayload
```
ncooke3, line 22: Same as above comment.

ncooke3 (commented):
Some additional context

Package.swift
```diff
@@ -226,22 +227,39 @@ let package = Package(
...
+    ),
     .testTarget(
       name: "HeartbeatLoggingTests",
-      dependencies: ["HeartbeatLogging"],
+      dependencies: [
+        "HeartbeatLogging",
+        "HeartbeatLoggingTestUtils",
+      ],
       path: "HeartbeatLogging/Tests/"
     ),
```
ncooke3, line 261: This organization seems to be working well for SPM.

ncooke3: I've been experimenting with how to make a similar setup work for CocoaPods without making FirebaseCore depend on a new, public HeartbeatLogging.podspec. 

I'm going to try following approach: 
- Create a HeartbeatLoggingTestingUtils.podspec
- Depend on new test util podspec's Swift bridging header in the FirebaseCore objc unit testing spec

ncooke3: WIP with above approach at 4e9c9a9. I'm running into two compiler errors that I added comments for.

paulb777: That might work. We may need to wrap the test spec in an environment variable like https://github.com/google/GoogleDataTransport/blob/main/GoogleDataTransport.podspec#L64 so that the test is only seen by `pod lib lint` but not by `pod spec lint` or in production.

HeartbeatLoggingTestUtils/Sources/HeartbeatLoggingTestUtils.swift
```diff
@@ -13,10 +13,12 @@
 // limitations under the License.
 
 import Foundation
-import XCTest
+import XCTest // COMPILE ERROR: No such module 'XCTest'
```
ncooke3, line 16: I'm not sure what why this is happening. I looked at how popular swift testing cocoapods do it and they specify `XCTest` as a framework.

https://github.com/firebase/firebase-ios-sdk/blob/4e9c9a907ddaf41d3590041b5d2c6b688389de04/HeartbeatLoggingTestingUtils.podspec#L37

https://github.com/firebase/firebase-ios-sdk/blob/4e9c9a907ddaf41d3590041b5d2c6b688389de04/HeartbeatLoggingTestingUtils.podspec#L48-L51

paulb777: And https://github.com/erikdoe/ocmock/blob/master/OCMock.podspec.

ncooke3:
Nice to see green CI 😅 . I think there are still open questions about the organization in the FirebaseCore.podspec.

FirebaseCore.podspec
```diff
@@ -68,11 +73,35 @@ Firebase Core includes FIRApp and FIROptions which provide central configuration
...
+    s.test_spec 'heartbeat-unit' do |unit_tests|
+      unit_tests.scheme = { :code_coverage => true }
+      unit_tests.platforms = {
+        :ios => ios_deployment_target,
+        :osx => osx_deployment_target,
+        :tvos => tvos_deployment_target
+      }
+      unit_tests.source_files = [
+        'FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m',
+        # TODO: Is testing HeartbeatLogging on SPM enough?
```
paulb777, line 98: That would be fine - in which case this test_spec should be deleted. Do you understand why it doesn't work?

FirebaseStorage.podspec
```diff
@@ -27,6 +27,8 @@ Firebase Storage provides robust, secure file uploads and downloads from Firebas
   s.tvos.deployment_target = tvos_deployment_target
   s.watchos.deployment_target = watchos_deployment_target
 
+  s.swift_version = '5.3'
```
paulb777, line 30: I'm not sure this is needed since there's no Swift in the pod. Or did this fix a warning with tests?

HeartbeatLogging/Sources/_ObjC_HeartbeatController.swift
paulb777: Why not use `NS_REFINED_FOR_SWIFT`? https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/improving_objective-c_api_declarations_for_swift

HeartbeatLoggingTestUtils.podspec
```diff
@@ -0,0 +1,46 @@
...
+  s.ios.deployment_target     = ios_deployment_target
+  s.osx.deployment_target     = osx_deployment_target
+  s.tvos.deployment_target    = tvos_deployment_target
+  s.watchos.deployment_target = watchos_deployment_target
+
+  s.source_files = [
+    'HeartbeatLoggingTestUtils/Sources/**/*.swift',
+  ]
+
+  s.dependency 'FirebaseCore', '~> 8.0'
```
paulb777, line 43: 8.11

paulb777 (commented):
Just a few nits/questions.

Looking good!

FirebaseStorage.podspec
ncooke3: Yes, it was to fix a warning that popped up during CI for this PR. See here: https://github.com/firebase/firebase-ios-sdk/runs/4610188684?check_suite_focus=true

The storage workflows get triggered very easily, my guess was the warning was caused by Storage depending on Core which now has swift?

HeartbeatLoggingTestUtils.podspec
ncooke3: Done

FirebaseCore.podspec
ncooke3: Why what doesn't work– the mixed swift and objc unit tests? I'm not sure. I added the swift tests and then saw I had to change imports so I figured i'd hold off until we decided what we wanted

HeartbeatLogging/Sources/_ObjC_HeartbeatController.swift
ncooke3: Because `NS_REFINED_FOR_SWIFT` is for ObjC → Swift. 

For Swift → ObjC, types need to be public and conform to `NSObject`.

ncooke3: Some extra context.

I tried making `HeartbeatController` class conform to `NSObject` to make it objc-compatible. But, the `flush` API returns a `HeartbeatsPayload` which is a Swift struct and does not bridge. So I had to create a bridging type for `HeartbeatsPayload`.

We could eliminate the need for these custom bridging types by turning `HeartbeatController` and `HeartbeatsPayload` into `NSObject`s but I felt `HeartbeatsPayload` should stay a value type. OTOH, `HeartbeatsPayload` is immutable as it is so making it a reference type wouldn't affect its usage.

WDYT?

paulb777: But wouldn't `NS_REFINED_FOR_SWIFT` also hide the Objective C API from Swift so it wouldn't need to be prefixed with the underscore?

FirebaseStorage.podspec
paulb777: In that case, we might need to add it to all the podspecs that depend on FirebaseCore ...

HeartbeatLogging/Sources/_ObjC_HeartbeatController.swift
ncooke3: Where are you suggesting using `NS_REFINED_FOR_SWIFT`?

paulb777: Thanks. I see my confusion now. It make sense.

FirebaseStorage.podspec
ncooke3: Good point. I guess that a podspec depending on a Swift pod means the swift_version must also be set to avoid the warning. 

Once this PR is sorted out. I could run touch the Gemfile to run CI for everything and see what happens.

FirebaseCore.podspec
ncooke3: So if this test_spec is deleted, I can wrap the "unit" test_spec in the env var check. Is OK for FirebaseCore's "unit" test spec to be gated by the env var?

paulb777: Yes. It's fine to use env vars in test specs.

ncooke3: So if i set the env var locally and then use the `pod gen` tool to generate a workspace, the gated scheme does not work. Is this expected?

paulb777: It looks like it's set to intentionally fail:
Screen Shot 2021-12-23 at 3 13 05 PM

ncooke3: Hmm I cant even run that scheme.

ncooke3: Screen Shot 2021-12-23 at 6 21 13 PM

ncooke3:
I'm expecting the latest commit to be 🟢

paulb777 (approved):

FirebaseCore/Sources/Private/FIRHeartbeatLogger.h
```diff
@@ -0,0 +1,58 @@
...
+// TODO: Remove in future PR when `FIRHeartbeatInfo` symbol is removed.
+#import "FirebaseCore/Sources/Private/FIRHeartbeatInfo.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class FIRHeartbeatsPayload;
+
+/// <#Description#>
+/// @param heartbeatsPayload <#heartbeatsPayload description#>
+NSString *FIRHeaderValueFromHeartbeatsPayload(FIRHeartbeatsPayload *heartbeatsPayload);
```
ncooke3, line 26: Finish doc.

FirebaseStorage.podspec
ncooke3: TODO

HeartbeatLogging/Sources/StorageFactory.swift
ncooke3: Make private again since they have been copied over to new HeartbeatLoggingTestUtils target

HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
```diff
@@ -14,23 +14,24 @@
 
 import XCTest
 @testable import HeartbeatLogging
+import HeartbeatLoggingTestUtils
 
 /// See `Sources/StorageFactory.swift` for details regarding where the module stores client data.
 private enum Constants {
```
ncooke3, line 20: Remove in favor of HeartbeatLoggingTestUtils.Constants

ncooke3: Also try just importing symbol again...

HeartbeatLogging/Tests/Unit/StorageTests.swift
```diff
@@ -100,14 +100,7 @@ class UserDefaultsStorageTests: XCTestCase {
...
   override func setUpWithError() throws {
-    // Clean up suite in case prior test case did not complete.
-    UserDefaults.standard.removePersistentDomain(forName: suiteName)
-
-    defaults = try XCTUnwrap(UserDefaults(suiteName: suiteName))
-  }
-
-  override func tearDownWithError() throws {
-    defaults.removePersistentDomain(forName: suiteName)
+    defaults = try XCTUnwrap(UserDefaultsFake(suiteName: suiteName))
```
ncooke3, line 103: Using the real user defaults generates a plist file in the test suite that needs to be manually removed before committing changes. Deleting the file in the class tear down did not work so I added a fake user defaults to inject.

PR #9118, opened 2021-12-21
[Heartbeat Logging] Add compression for heartbeats payloads

ncooke3:
**Context**
Compressing heartbeats payloads reduces size and should help reduce bandwidth when sending in requests

- [x] Do a manual backend test to ensure the backend can decode data from this implementation

**Experiments**
| payload type | not gzipped  (bytes) | gzipped (bytes) |
| --- | --- | --- |
| small | 75 | 85 |
| large | 490 | 171 |

#no-changelog

### DIFF
--- a/FirebaseCore.podspec
+++ b/FirebaseCore.podspec
@@ -53,6 +53,7 @@ Firebase Core includes FIRApp and FIROptions which provide central configuration
   # Remember to also update version in `cmake/external/GoogleUtilities.cmake`
   s.dependency 'GoogleUtilities/Environment', '~> 7.6'
   s.dependency 'GoogleUtilities/Logger', '~> 7.6'
+  s.dependency 'GoogleUtilities/NSData+zlib', '~> 7.6'
   s.dependency 'FirebaseCoreDiagnostics', '~> 8.0'
 
   s.pod_target_xcconfig = {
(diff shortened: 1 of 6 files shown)


HeartbeatLogging/Sources/HeartbeatsPayload.swift
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Foundation
+import GoogleUtilities_NSData
```
ncooke3, line 16: Weird. When I command click on the import in Xcode, it brings me to the swift interface. Yet, it still fails to build with `no such module GoogleUtilities_NSData`

FirebaseCore.podspec
```diff
@@ -53,6 +53,7 @@ Firebase Core includes FIRApp and FIROptions which provide central configuration
   # Remember to also update version in `cmake/external/GoogleUtilities.cmake`
   s.dependency 'GoogleUtilities/Environment', '~> 7.6'
   s.dependency 'GoogleUtilities/Logger', '~> 7.6'
+  s.dependency 'GoogleUtilities/NSData+zlib', '~> 7.6'
```
ncooke3, line 56: .

ncooke3: Is there some other config I'm missing? @paulb777

ncooke3: Maybe related: https://stackoverflow.com/questions/67894280/how-to-import-a-cocoapods-subspec-module-instead-of-the-entire-pod-in-swift

paulb777 (approved):

PR #9123, opened 2021-12-21
Fix RTDB crash from race

### LINKED ISSUE #9096: Intermittent crash in FUtilities.intForString

  * Xcode version: 13.0
  * Firebase SDK version:8.9.1
  * Firebase Component: Database
  * Target platform(s): `iOS`

- It appeared when we started using the SDK as dependency of one of our SPM package, before, we were using cocoapods and it was fine.
- FirebaseDatabase entities get mapped when app is launched and it intermittently crashes in `FUtilities.intForString`

#### Steps to reproduce:
- Happens when we launch the app since we starting mapping entities immediately
- Attached (crashlog.txt) is the crash report

#### Relevant Code:

```
// TODO(you): code here to reproduce the problem
+ (NSNumber *)intForString:(NSString *)string {
    static NSCharacterSet *notDigits = nil;
    if (!notDigits) {
        notDigits = [[NSCharacterSet decimalDigitCharacterSet] invertedSet];
    }
    if ([string rangeOfCharacterFromSet:notDigits].length == 0) {
        NSInteger num;
        NSScanner *scanner = [NSScanner scannerWithString:string];
        if ([scanner scanInteger:&num]) {
            return [NSNumber numberWithInteger:num];
        }
    }
    return nil;
}

```
Screen Shot 2021-12-14 at 11 11 55 AM
[crashlog.txt](https://github.com/firebase/firebase-ios-sdk/files/7708426/crashlog.txt)

paulb777:
Fix #9096

### DIFF
--- a/FirebaseDatabase/Sources/Utilities/FUtilities.m
+++ b/FirebaseDatabase/Sources/Utilities/FUtilities.m
@@ -295,10 +295,11 @@ void firebaseJobsTroll(void) {
 }
 
 + (NSNumber *)intForString:(NSString *)string {
-    static NSCharacterSet *notDigits = nil;
-    if (!notDigits) {
+    static dispatch_once_t once;
+    static NSCharacterSet *notDigits;
+    dispatch_once(&once, ^{
         notDigits = [[NSCharacterSet decimalDigitCharacterSet] invertedSet];
-    }
+    });
     if ([string rangeOfCharacterFromSet:notDigits].length == 0) {
         NSInteger num;
         NSScanner *scanner = [NSScanner scannerWithString:string];
--- a/FirebaseDatabase/Sources/Utilities/FUtilities.m
+++ b/FirebaseDatabase/Sources/Utilities/FUtilities.m
@@ -298,7 +298,7 @@ void firebaseJobsTroll(void) {
     static dispatch_once_t once;
     static NSCharacterSet *notDigits;
     dispatch_once(&once, ^{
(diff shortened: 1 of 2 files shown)


FirebaseDatabase/Sources/Utilities/FUtilities.m
```diff
@@ -295,10 +295,11 @@ + (NSError *)errorForStatus:(NSString *)status andReason:(NSString *)reason {
...
 + (NSNumber *)intForString:(NSString *)string {
-    static NSCharacterSet *notDigits = nil;
-    if (!notDigits) {
-        notDigits = [[NSCharacterSet decimalDigitCharacterSet] invertedSet];
-    }
+    static dispatch_once_t once;
+    static NSCharacterSet *notDigits;
+    dispatch_once(&once, ^{
+      notDigits = [[NSCharacterSet decimalDigitCharacterSet] invertedSet];
+    });
```
ncooke3, line 302: Curious, what is the reason we want to assign the `notDigits` just once? Is it expensive to `[[NSCharacterSet decimalDigitCharacterSet] invertedSet]`?

paulb777: I was focused on correcting the existing implementation - but presumably it is converting a operation processing the Unicode character set to a cached value which could add up if this function is called a lot. cc: @schmidt-sebastian

ncooke3: Gotcha, interesting, makes sense. Nice fix!

schmidt-sebastian (approved):

Issue #9070, opened 2021-12-09
Duplicate messages can occur when two campaigns are triggered by different events in In-App Messaging


kouki-dan:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.1
  * Firebase SDK version: 8.10.0
  * Installation method: `Swift Package Manager` 
  * Firebase Component: In-App Messaging 
  * Target platform(s): `iOS`

### [REQUIRED] Step 2: Describe the problem
Duplicate messages can occur when two campaigns are triggered by different events. (The first message must have an image to reproduce it.)

Setup like this in the console.
|First Event|Second Event|
|-|-|
|<img width="934" alt="スクリーンショット 2021-12-09 23 07 04" src="https://user-images.githubusercontent.com/1401711/145411325-d14fc25c-8a57-4af1-b62c-ad2ac3ac01b6.png">|<img width="913" alt="スクリーンショット 2021-12-09 23 07 35" src="https://user-images.githubusercontent.com/1401711/145411364-a5ccd7eb-ec5a-4177-9ee1-43d54caa477f.png">|

And trigger events consecutively like this code.

```swift
struct ContentView: View {
    var body: some View {
        Button("Trigger In-App") {
            InAppMessaging.inAppMessaging().triggerEvent("first_event")
            InAppMessaging.inAppMessaging().triggerEvent("second_event")
        }
            .padding()
    }
}
```

It shows multiple dialogs in a row.

https://user-images.githubusercontent.com/1401711/145411619-850d6c24-845f-4989-9217-24a5c06e1dee.mov

It seems [FIRIAMDisplayExecutor](https://github.com/firebase/firebase-ios-sdk/blob/c7f80c14c2ff207173f1a3ba294df12c019c266a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m) checks message is displayed or not in this code.
https://github.com/firebase/firebase-ios-sdk/blob/c7f80c14c2ff207173f1a3ba294df12c019c266a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m#L422-L428

However, `isMsgBeingDisplayed` becomes true only after the image has finished loading.
https://github.com/firebase/firebase-ios-sdk/blob/c7f80c14c2ff207173f1a3ba294df12c019c266a/FirebaseInAppMessaging/Sources/Flows/FIRIAMDisplayExecutor.m#L665

[...]

PR #9138, opened 2021-12-23
Fix : Duplicate messages can occur when two campaigns are triggered by different events

eldhosembabu:
Fix: Duplicate messages can occur when two campaigns are triggered by different events in In-App Messaging #9070

The flags to prevent multiple calling of inapp message was set after image downloading and so if two events are fired simultaneously, the second message display will be called prior to flag set. So adjusted the flag.

### DIFF
--- a/FirebaseInAppMessaging/CHANGELOG.md
+++ b/FirebaseInAppMessaging/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2021-12 -- v8.12.0
+# 2022-01 -- v8.12.0
 - [fixed] Duplicate messages can occur when two campaigns are triggered by different events in In-App Messaging (#9070).
 
 # 2021-12 -- v8.11.0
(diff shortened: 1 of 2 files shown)


FirebaseInAppMessaging/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2021-12 -- v8.12.0
+- [fixed] Duplicate messages can occur when two campaigns are triggered by different events in In-App Messaging (#9070).
```
paulb777, line 2: Move to 8.11.0 section. 8.11.0 won't release until next month.

paulb777 (approved):
Thanks!

FirebaseInAppMessaging/CHANGELOG.md
eldhosembabu: Updated the year and month to current. Please lmk if it looks good.

paulb777: It should be 8.11.0 if it merges before the codefreeze. See go/firetracker

eldhosembabu: Moved under 8.11.0. Please review.

paulb777 (approved):
Thanks

PR #9140, opened 2021-12-28
Don't fail on empty bundles

schmidt-sebastian:
Nodejs Firestore creates a bundle with missing "seconds" and "nanos" properties when no documents or queries are added to the bundle. These should be valid to load, so this PR adds this ability.


Firestore/Example/Firestore.xcodeproj/project.pbxproj
```diff
@@ -2371,13 +2370,6 @@
 			path = core;
 			sourceTree = "<group>";
 		};
-		AB38D9312023962A000A432D /* auth */ = {
```
schmidt-sebastian, line 2374: This change is just to remove the "Auth" folder which is now "credentials"

wu-hui (approved):
Thanks for the fix!

wu-hui:
> Thanks for the fix!

Have you verify this for JS/Android?

schmidt-sebastian:
Yeah, this works as is on JS/Android.

PR #9145, opened 2021-12-29
Update Functions Swift public API docs

paulb777:
#no-changelog

### DIFF
--- a/FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
+++ b/FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
@@ -23,11 +23,11 @@ public extension Functions {
   /// - Parameter requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
   /// - Parameter responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
   /// - Parameter encoder: The encoder instance to use to run the encoding.
-  /// - Parameter decoder: The decoder instance to use to run the encoding.
+  /// - Parameter decoder: The decoder instance to use to run the decoding.
   func httpsCallable<Request: Encodable,
     Response: Decodable>(_ name: String,
-                         requestAs requestType: Request.Type = Request.self,
-                         responseAs responseType: Response.Type = Response.self,
+                         requestAs: Request.Type = Request.self,
+                         responseAs: Response.Type = Response.self,
                          encoder: FirebaseDataEncoder = FirebaseDataEncoder(),
                          decoder: FirebaseDataDecoder = FirebaseDataDecoder())
     -> Callable<Request, Response> {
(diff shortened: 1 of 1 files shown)


FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -139,6 +141,8 @@ public struct Callable<Request: Encodable, Response: Decodable> {
     /// resumes with a new FCM Token the next time you call this method.
     ///
     /// - Parameter data: The `Request` representing the data to pass to the trigger.
+    /// - Parameter encoder: The encoder instance to use to run the encoding.
+    /// - Parameter decoder: The decoder instance to use to run the encoding.
```
ncooke3, line 145: s/encoding/decoding

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -20,8 +20,10 @@ public extension Functions {
   /// Creates a reference to the Callable HTTPS trigger with the given name, the type of an `Encodable`
   /// request and the type of a `Decodable` response.
   /// - Parameter name: The name of the Callable HTTPS trigger
-  /// - Parameter requestType: The type of the `Encodable` entity to use for requests to this `Callable`
-  /// - Parameter responseType: The type of the `Decodable` entity to use for responses from this `Callable`
+  /// - Parameter requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  /// - Parameter responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
+  /// - Parameter encoder: The encoder instance to use to run the encoding.
+  /// - Parameter decoder: The decoder instance to use to run the encoding.
```
ncooke3, line 26: s/encoding/decoding

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
```diff
@@ -20,8 +20,10 @@ public extension Functions {
   /// Creates a reference to the Callable HTTPS trigger with the given name, the type of an `Encodable`
   /// request and the type of a `Decodable` response.
   /// - Parameter name: The name of the Callable HTTPS trigger
-  /// - Parameter requestType: The type of the `Encodable` entity to use for requests to this `Callable`
-  /// - Parameter responseType: The type of the `Decodable` entity to use for responses from this `Callable`
+  /// - Parameter requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  /// - Parameter responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
```
ncooke3, line 24: Based on my understanding of [Swift argument labels vs parameter names](https://docs.swift.org/swift-book/LanguageGuide/Functions.html#:~:text=implicit%20return%20value.-,Function%20Argument%20Labels%20and%20Parameter%20Names,-Each%20function%20parameter), I think the before diff is the correct form because `requestType` is the parameter name and `requestAs` is the argument label. The `- Parameter` doc should therefore declare the parameter name. 

Xcode offers an opinion if you select the API signature and use the `option + cmd + / ` "add-docs" shortcut.

I'm not firm on this, but I remember wondering about it once and this ^ was the rationale that informed my decision then.

ncooke3 (approved):

FirebaseFunctionsSwift/Sources/Codable/Callable+Codable.swift
paulb777: I didn't know that one. TIL! Unfortunately it fails for this signature, but it works for me for others.  It turns out there is no need for parameter names for this one, so I removed them.

paulb777 (commented):
Good catches. thanks!

Issue #8949, opened 2021-11-09
FIRAnalytics setConsent request hangs when `FIRConsentTypeAdStorage` is `FIRConsentStatusDenied`


russellwheatley:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.1
  * Firebase SDK version: 8.9.0
  * Installation method: `CocoaPods` (select one)
  * Firebase Component: Analytics 

### [REQUIRED] Step 2: Describe the problem

Try to set `FIRConsentTypeAdStorage` to `FIRConsentStatusDenied` in the `setConsent` API will cause the request to hang.

#### Steps to reproduce:

Simply run the below noted `Relevant Code` and you ought to see the request hang. 
#### Relevant Code:

```obj-c
NSMutableDictionary<FIRConsentType, FIRConsentStatus> *parameters = [[NSMutableDictionary alloc] init];
parameters[FIRConsentTypeAdStorage] = FIRConsentStatusDenied;
[FIRAnalytics setConsent:parameters];
```

rizafran:
Hi @russellwheatley, it looks like I was able to reproduce the issue. To make sure that we're on the same page, is the request hanging only upon opening the app and is not reproducible if using `FIRConsentStatusGranted`?

russellwheatley:
Hi @rizafran, The request is fine if you use `FIRConsentStatusGranted`. The request never completes if you use `FIRConsentStatusDenied `. I'm afraid I don't have it open to test the specific timing of when the request fails.

rizafran:
Thanks for confirming, @russellwheatley. It looks like we're having the same behavior. With this, I'll notify an engineer and see what we can do here.

tsunghung:
Thanks for the bug report. We found a deadlock in the Analytics SDK during initialization. We will fix it in the next release.
In the meantime, to disable Ad storage by default, you can set the value of `GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_STORAGE` to `NO` in your app's Info.plist file. For example, viewed in the source XML:
```
<key>GOOGLE_ANALYTICS_DEFAULT_ALLOW_AD_STORAGE</key>
<false/>
```
You can still use the `setConsent` API to change the consent status afterward.

paulb777:
This fix is now targeted for the release after next - 8.12.0

eddie8balltoliver:
[...]

PR #9147, opened 2021-12-29
Fireperf EAP: custom `cold_start` and `prewarmed_start` trace for prewarming identification protytpe

leotianlizhan:
b/211905766


FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -143,26 +147,38 @@ - (FIRTrace *)activeTrace {
...
+  if (![self prewarmAvailable]) {
     return NO;
   }
 
-  NSDictionary<NSString *, NSString *> *environment = [NSProcessInfo processInfo].environment;
-  if (environment[@"ActivePrewarm"] != nil &&
-      [environment[@"ActivePrewarm"] isEqualToString:@"1"]) {
+  if ([self isActivePrewarm]) {
     return YES;
   }
```
jeremyjiang-dev, line 157: Nit: `return [self isActivePrewarm] || [self isDoubleDispatchPrewarm]`

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -221,6 +248,29 @@ - (void)appDidBecomeActiveNotification:(NSNotification *)notification {
...
         [self.appStartTrace cancel];
       }
+
+      if ([self prewarmAvailable] &&
+          (currentTimeSinceEpoch - startTimeSinceEpoch < gAppStartMaxValidDuration)) {
+        BOOL activePrewarm = [self isActivePrewarm];
+        BOOL doubleDispatchPrewarm = [self isDoubleDispatchPrewarm];
+        if (!activePrewarm && !doubleDispatchPrewarm) {
+          [self.coldStartTrace stop];
+          [self.prewarmedStartTrace cancel];
```
jeremyjiang-dev, line 258: Why do you need `return`? 
Nit: Can you connect `if (!activePrewarm && !doubleDispatchPrewarm)` and `if (activePrewarm && doubleDispatchPrewarm)` with `else if` for better readability?

jeremyjiang-dev (commented):
The changes LGTM! Can you add some tests?

jeremyjiang-dev (approved):

leotianlizhan:
> We also need to update [the SDK version](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance.podspec#L3) to something like `ios-15-eap` in this branch.

does the SDK version have to follow a number pattern like `8.12.0` or does it work with any string like `ios-15-eap`?

jeremyjiang-dev:
> > We also need to update [the SDK version](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance.podspec#L3) to something like `ios-15-eap` in this branch.
> 
> does the SDK version have to follow a number pattern like `8.12.0` or does it work with any string like `ios-15-eap`?

Let's add Paul to review the PR so that we are following the best practice.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -143,24 +146,49 @@ - (FIRTrace *)activeTrace {
...
+    [trace setValue:@"active_prewarm" forAttribute:@"prewarm_detection"];
+    [trace setValue:@"yes" forAttribute:@"is_prewarmed"];
+  } else {
+    [trace setValue:@"double_dispatch" forAttribute:@"prewarm_detection"];
+    [trace setValue:@"yes" forAttribute:@"is_prewarmed"];
   }
+}
+
+- (BOOL)prewarmAvailable {
+  NSArray *versionComponents =
```
paulb777, line 178: May also check that the platform is iOS. I'm not aware of the prewarm issue is also on tvOS. Also consider using https://github.com/google/GoogleUtilities/blob/main/GoogleUtilities/Environment/third_party/GULAppEnvironmentUtil.m#L245 to get the OS version

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -143,24 +146,49 @@ - (FIRTrace *)activeTrace {
...
+      [[UIDevice currentDevice].systemVersion componentsSeparatedByString:@"."];
+  NSInteger majorVersion = [versionComponents[0] integerValue];
+  return majorVersion >= 15;
+}
+
+- (BOOL)isActivePrewarm {
+  NSDictionary<NSString *, NSString *> *environment = [NSProcessInfo processInfo].environment;
+  return (environment[@"ActivePrewarm"] != nil &&
```
paulb777, line 186: May want to use a plist value to check for enabling the EAP code to make it easy to include in a release and avoid separate release management for the EAP.  For example, see https://github.com/firebase/firebase-ios-sdk/issues/8949#issuecomment-975970983

leotianlizhan: Do you mean we make a normal release, then ask EAP customers to set a plist value to enable this prototype?

paulb777: Yes. It's easier from a documentation/release management/installation/usability perspective.

paulb777: Discussed offline and agreed this is appropriate for a limited EAP

leotianlizhan:
Since there has been no signups for EAP, this branch will be closed.

PR #9149, opened 2021-12-29
[Won't merge] Testing auth's keychain implementation on GHA

ncooke3:
#no-changelog


ncooke3:
See failures at https://github.com/firebase/firebase-ios-sdk/actions/runs/1635758403

PR #9158, opened 2022-01-04
`schedule` integration tests to run on prewarm feature branch

leotianlizhan:
b/211905617

## Reason for PR to master
According to https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#schedule, 

> Scheduled workflows run on the latest commit on the default or base branch. 

Thus a PR to master branch is required to have the E2E pipeline automatically test the prototype for us.

## Interval
Currently set to every 2 hours. Alternatively could keep the 4 hours interval and use the old `performance-integration-tests.yml` file instead of making a new one.

### DIFF
--- a/.github/performance-integration-tests-prewarm.yml
+++ b/.github/performance-integration-tests-prewarm.yml
@@ -35,4 +35,4 @@ jobs:
         scripts/decrypt_gha_secret.sh scripts/gha-encrypted/Performance/GoogleService-Info_e2e_prod.plist.gpg \
             FirebasePerformance/Tests/FIRPerfE2E/FIRPerfE2EProd/GoogleService-Info.plist "$plist_secret"
     - name: BuildAndTest # can be replaced with pod lib lint with CocoaPods 1.10
-      run: scripts/third_party/travis/retry.sh scripts/build.sh Performance iOS-device integration
+      run: scripts/third_party/travis/retry.sh scripts/build.sh Performance all integration
--- a/.github/performance-integration-tests-prewarm.yml
+++ b/.github/performance-integration-tests-prewarm.yml
@@ -1,0 +1,38 @@
+# Merge the yml file to master branch for the cron job schedule to be effective.
+# Reference: https://github.community/t/on-schedule-per-branch/17525
+name: performance-integration-tests-prewarm
+
+on:
+  # See cron syntax references:
+  #   - https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule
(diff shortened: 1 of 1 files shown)


jeremyjiang-dev:
The changes LGTM. Have you tested the workflow? I think you have run the workflow here[https://github.com/firebase/firebase-ios-sdk/actions/workflows/performance-integration-tests-prewarm.yml] Just want to confirm that you have received the events.

Nit: could you also comment in to bug to remove this workflow after experiments?

leotianlizhan:
> The changes LGTM. Have you tested the workflow? I think you have run the workflow [here](https://github.com/firebase/firebase-ios-sdk/actions/workflows/performance-integration-tests-prewarm.yml). Just want to confirm that you have received the events.

It works when the arguments are `Performance all integration`, however I notice that that only uses the simulator. When I switch to `Performance iOS-device integration`, it fails because of the [following](https://github.com/firebase/firebase-ios-sdk/runs/4707962860?check_suite_focus=true#step:6:72):

> xcodebuild: error: option '-sdk' may only be provided once

Looking at `scripts/build.sh`, it looks like no matter what the simulator arguments are added [here](https://github.com/firebase/firebase-ios-sdk/blob/6a3e8398965b342d50917f6b562493ebdf5f27b3/scripts/build.sh#L660). Then the [next line](https://github.com/firebase/firebase-ios-sdk/blob/6a3e8398965b342d50917f6b562493ebdf5f27b3/scripts/build.sh#L661) adds whatever the 2nd argument is (e.g. `iOS-device`, `all`). This looks like anything but `all` will fail, and it always just use the simulator. Is that intentional?

leotianlizhan:
I've experimented with changing `scripts/build.sh` to use only the given argument platform, but now I get this error when I specify `iOS-device`:

> ❌  error: Signing for "FIRPerfE2EAutopush" requires a development team. Select a development team in the Signing & Capabilities editor. (in target 'FIRPerfE2EAutopush' from project 'FIRPerfE2E')

Is this related to Google's iOS dev device approval requirements?

leotianlizhan:
Given that we cannot run on real device due to security issues mentioned at go/ftl-github-ios-setup, what is our next step of action? Do we just use simulator for now and see if prewarming ever happens, or setup a pipeline that pulls code into g3 then use kokoro?

jeremyjiang-dev:
> 
I think we have to skip the E2E test setup in this case, which will involve a huge effort. Can you take the lead on collecting the EAP apps?

PR #9162, opened 2022-01-05
Remote Config Swift APIs

fumito-ito:
refs: https://github.com/firebase/firebase-ios-sdk/discussions/9108

# Why ?

I got following message on twitter for my library, [SwiftyRemoteConfig](https://github.com/fumito-ito/SwiftyRemoteConfig).

> Looks promising. We'll be happy to work with you if you want to turn this into PR(s) for a FirebaseRemoteConfigSwift library at

https://twitter.com/paulbeusterien/status/1448001604536307712?s=20

This feature is being considered on https://github.com/firebase/firebase-ios-sdk/discussions/9108.

# What ?

This PR add Swifty API interfaces for RemoteConfig (e.g. RawRepresentable and Custom type support !). All features are on [CHANGELOG.md](https://github.com/firebase/firebase-ios-sdk/pull/9162/files#diff-5987a9347751405d13e565dba489d0f005de9b8c2cf4d5a215420524956c3a4f)

# Todo 

- [x] copy and rename files
- [x] add interface for Swift Package Manager and Cocoapods
- [x] fix some warnings caused by older OS version.
- [ ] fix some CI settings
- [ ] use project Codable infrastructure a.k.a [FirebaseSharedSwift](https://github.com/firebase/firebase-ios-sdk/tree/master/FirebaseSharedSwift)
- [ ] format code according to project style
- [ ] discuss and fix some interface, naming and so on.

### DIFF
--- a/FirebaseRemoteConfigSwift.podspec
+++ b/FirebaseRemoteConfigSwift.podspec
@@ -1,0 +1,43 @@
+#
+# Be sure to run `pod lib lint FirebaseRemoteConfigSwift.podspec' to ensure this is a
+# valid spec before submitting.
+#
+
+Pod::Spec.new do |s|
+  s.name                    = 'FirebaseRemoteConfigSwift'
+  s.version                 = '8.10.0-beta'
+  s.summary                 = 'Swift Extensions for Google Cloud RemoteConfig'
+
+  s.description      = <<-DESC
+  Firebase Remote Config is a cloud service that lets you change the
+  appearance and behavior of your app without requiring users to download an
+  app update.
+                       DESC
+
+  s.homepage                = 'https://developers.google.com/'
+  s.license                 = { :type => 'Apache', :file => 'LICENSE' }
+  s.authors                 = 'Google, Inc.'
+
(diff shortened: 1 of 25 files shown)


paulb777:
@fumito-ito Thanks so much for creating SwiftyRemoteConfig and sharing it with the Firebase developer community.

Regarding merging this PR, we still need to do some more analysis, but here are some initial thoughts: It looks like SwiftyRemoteConfig strives to address the requirement of defining a great API for Remote Config for users starting without any previous Firebase code or experience.

That does not exactly meet our requirements. We have a large existing developer base and we need to support the developers migrating incrementally on whatever schedule works for them. We want to avoid breaking existing code as much as possible. And we want to support existing code remaining unchanged while new code is developed with a more Swifty API.

Therefore, the best way to make SwiftyRemoteConfig mergeable would be to explore how it can complement the additive approach taken in #9084 to provide additional features.

fumito-ito:
@paulb777 thanks your feedback. #9084 is great work. I have few reviews on it.

Issue #8096, opened 2021-05-17
Sign in alert shown inconsistently when triggered in viewDidAppear 


lfkellogg:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 12.4
  * Firebase SDK version: 8.0.0
  * Installation method: `CocoaPods`
  * Firebase Component: App Distribution
  * iOS version: 13+

#### Steps to reproduce:

1. Create a new Project, choose the "App" template, then set Interface to `Storyboard`
2. Follow the docs to set up the SDK in the "Basic" configuration, calling `checkForUpdate` from `viewDidAppear`
3. Run the app on a device running iOS 13+
4. [Optional] If the "Enable new build alerts" alert is shown on launch, kill the app from the app switcher and re-launch it

Expected behavior: on first launch and on subsequent launches before signing in, the SDK should show the "Enable new build alerts" alert.

Actual behavior: It is inconsistent. Often the alert will show up on first launch, but not on subsequent launches. Sometimes not even on first launch.

#### Relevant Code:

Debugging the SDK locally, the issue appears to be with this line: https://github.com/firebase/firebase-ios-sdk/blob/f920f2feaad50aa7d6a196e956574b21fcab927a/FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m#L208

Sometimes, the scene has not become active by the time `viewDidAppear` is called. In that case we do not find a foreground scene here, and the alert cannot be shown. This is illustrated by a workaround where `checkForUpdate` is triggered from the SceneDelegate's `sceneDidBecomeActive` instead.

tejasd:
This has been fixed for single scene apps in Firebase SDK version: 10.1.0

PR #9169, opened 2022-01-06
Fixed inconsistent appdistribution prompt (#8096)

a-25:
Fix for issue (#8096), prompt is shown consistently.


ryanwilson:
Thanks for the contribution @a-25. A quick question, did you look into the [`UISceneDidActivateNotification`](https://developer.apple.com/documentation/uikit/uiscenedidactivatenotification?language=objc) at all to see if it could be used instead? Seems like it could be useful and would let us avoid just looping without knowing the state.

a-25:
@ryanwilson Hello. No, I didn't. Will look at it closely and return with changes or conclusions.

a-25:
Hello, @ryanwilson. Implemented using UISceneDidActivateNotification. Seems to me that it is a cleaner way.

ryanwilson:
Apologies for the delay in review, will be getting to it soon! Thanks for the update as well.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+- (void)onFirstSceneActivated API_AVAILABLE(ios(13.0)) {
+  UIWindowScene *foregroundedScene = nil;
+  BOOL areAllScenesUnattached = YES;
+  [self findScene](&foregroundedScene, &areAllScenesUnattached);
+  if (areAllScenesUnattached) {
+    return;
+  }
```
ryanwilson, line 55: Is there a situation where all scenes are unattached but we get a UIWindowScene back? I'm curious as to why this flag is needed as it could vastly simplify things if we just checked if the foreground scene is there.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
```diff
@@ -0,0 +1,82 @@
...
+                                                  name:UISceneDidActivateNotification
+                                                object:nil];
+  if (self.pendingCompletion != nil) {
+    self.pendingCompletion(foregroundedScene);
+    self.pendingCompletion = nil;
+  }
+}
+
+- (void (^)(UIWindowScene **, BOOL *))findScene API_AVAILABLE(ios(13.0)) {
+  return ^(UIWindowScene **foregroundedScene, BOOL *areAllScenesUnattached) {
```
ryanwilson, line 67: This format makes the call site quite difficult to read. If we don't need the unattached flag, this could be simplified as a function returning the `UIWindowScene *`. if it's still needed, we should still return that and pass in the BOOL as a `BOOL *` as an argument.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
```diff
@@ -0,0 +1,82 @@
...
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(onFirstSceneActivated)
+                                                 name:UISceneDidActivateNotification
+                                               object:nil];
+  } else {
+    completionBlock(nil);
+  }
+}
+
+- (void)onFirstSceneActivated API_AVAILABLE(ios(13.0)) {
```
ryanwilson, line 49: The notification that triggers this method also passes in the `UIScene` that is activated as the notification's object - would we be able to use that instead somehow? Maybe checking if it's a `UIWindowScene *` and if it's the foregrounded one? We can access it by adding an `NSNotification` parameter to this method.

ryanwilson (commented):
A couple comments to start!

FirebaseAppDistribution/Sources/FIRSceneHelper.m
a-25: Perhaps it was too cautious, removed that flag.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
a-25: You're right, fixed.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
a-25: Removed the boolean parameter, it is not needed.

a-25:
> A couple comments to start!

Hello, @ryanwilson. Thanks for your comments, worked them out.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
```diff
@@ -0,0 +1,75 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "FIRSceneHelper.h"
+#import <UIKit/UIKit.h>
+
+API_AVAILABLE(ios(13.0))
+typedef void (^FindSceneCompletionBlock)(UIWindowScene *scene);
```
ryanwilson, line 20: Can we add a `FIR` prefix to the `FindSceneCompletionBlock` name to avoid potential symbol collisions?

FirebaseAppDistribution/Sources/FIRSceneHelper.m
```diff
@@ -0,0 +1,75 @@
...
+
+- (void)findForegroundedSceneWithCompletionBlock:(void (^)(UIWindowScene *scene))completionBlock
+    API_AVAILABLE(ios(13.0)) {
+  if (@available(iOS 13.0, *)) {
+    UIWindowScene *foregroundedScene = [self findScene];
+    if (foregroundedScene != nil) {
+      completionBlock(foregroundedScene);
+      return;
+    }
+    self.pendingCompletion = completionBlock;
```
ryanwilson, line 37: The one concern I have for this is it now introduces a race condition, if two call this method before the `onFirstSceneActivated:` method gets called, the first one's completion handler will never be called. Open to suggestions but I'll think about this in terms of the best approach to move forward.

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -198,28 +205,38 @@ - (void)logRegistrationCompletion:(NSError *)error authType:(NSString *)authType
...
-    } else {
-      FIRFADInfoLog(@"No foreground scene found.");
-      self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-    }
-  } else {
-    self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+  self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+  [self windowFinishInitialization];
+  if (completion != nil) {
+    completion();
```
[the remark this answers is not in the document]
a-25: Fixed

FirebaseAppDistribution/Sources/FIRSceneHelper.m
a-25: Yes, agree.

FirebaseAppDistribution/Sources/FIRSceneHelper.m
a-25: Turned callbacks into an array to prevent from race conditions.

a-25:
Hi, @ryanwilson. Finished with comments fixing. Perhaps, I'd better resolve all conversations so that you received a notification?

tejasd:
Hi @a-25 - thanks for the PR!

There's a simpler fix (focused on single scenes) - does this work in your case? Or does your use case involve multiple scenes?

https://github.com/firebase/firebase-ios-sdk/pull/10288

a-25:
Hi, @tejasd. 
My case was for multiple scenes, as I remember. At least, I cannot guarantee that the original bug appears only in one-scene apps.

paulb777:

a-25:
@ryanwilson, @tejasd: what do you think, should it be merged?

paulb777:
Looks like no immediate plans to move forward on this. I'll close and we can reopen if plans change.

leelee855:
No don't close it

Issue #8783, opened 2021-10-11
FirebaseApp.app() logs an error when used as documented


paulb777:
FirebaseApp.app() is [documented](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Sources/Public/FirebaseCore/FIRApp.h#L78) to "Returns the default app, or nil if the default app does not exist".

However, when used to check if the default app exists and it doesn't, it logs an [error](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Sources/FIRApp.m#L251).

It's confusing to log an error when an API is used as documented. Perhaps, there should be separate APIs for getting the default app versus checking its existence?

ryanwilson:
We do have a `isDefaultAppConfigured` private API: https://github.com/firebase/firebase-ios-sdk/blob/63c1673255f6ffb444f2ec69a772b9734354e806/FirebaseCore/Sources/Private/FIRAppInternal.h#L103

but have generally recommended against people conditionally calling configure. I'm curious to hear the context as to how this came up and how we can address it!

paulb777:
It comes up in tests like at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Tests/SwiftAPI/APITestBase.swift#L28. The private API is not an option since the code is Swift.

paulb777:
On further discussion with the team, we don't want to extend the public API to solve this - because we always want Firebase configuration to be explicit. 

It would be good to solve the issue for tests only though.  There should be a good solution after #8973.  I'll add this issue to the Firebase 9 milestone.

paulb777:
This is unblocked by Firebase 9, but not necessarily required for it.

ryanwilson:
For the original issue of it being an error when used as described - do you think it should be dropped to a warning instead? 

Also what else do we need to do to close this issue?

paulb777:
This looks fixed now that `isDefaultAppConfigured` is available to Swift clients via FirebaseCoreExtension.

PR #9170, opened 2022-01-06
Added autocreate feature in default Firebase app (#8783)

a-25:
Added autocreate option in default app method (#8783). Removed a warning message when using defaultApp without autocreation.


paulb777:
@a-25 Thanks for the PR. Unfortunately, it's not mergable since we'd like to solve the test issue without impacting the public API.  See the latest comment in #8783 for more context.

PR #9173, opened 2022-01-06
Bump GoogleUtilities to 7.7.

asafkorem:
Bump GoogleUtilities to 7.7.

This version allows to disable the GULObjectSwizzler's generated classes dispose calls when `GULGeneratedClassDisposeDisabled` environment variable is set (see https://github.com/google/GoogleUtilities/pull/66).


paulb777 (approved):
Thanks! This will be part of the upcoming 8.11.0. In the meantime, `pod update` should update to GoogleUtilities 7.7

PR #9176, opened 2022-01-07
Reuse the content view from other test app in the new SwiftUI test app

chliangGoogle:
Use the same SwiftUI ContentView from other test app under Messaging to test functionalities of Messaging APIs.

I do discover the Swizzling is not working properly with the new SwiftUI, developers must implement the delegate method for the swizzling to happen.

#no-changelog


FirebaseMessaging/Apps/SwiftUISample/SwiftUISample/SwiftUISampleApp.swift
```diff
@@ -35,17 +42,61 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
       }
     }
     application.registerForRemoteNotifications()
+
+    // Observe token refresh - two ways
```
paulb777, line 46: Is there a third way with async/await?

paulb777 (approved):
Not necessary for this PR, but would be nice to have a build test in CI for the new Sample app.

FirebaseMessaging/Apps/SwiftUISample/SwiftUISample/SwiftUISampleApp.swift
chliangGoogle: Not for token refresh listener. It is for the async getToken call.

PR #9177, opened 2022-01-07
Enable checking diff on fork.

granluo:
`common_commit` was the common ancestor of a target branch and base branch. This cannot be applied to fork. This variable block the workflow of spec testing on fork. This variable is not used in any workflows, and will be removed to unblock the workflow.
```
git diff --name-only remotes/origin/${GITHUB_BASE_REF} ${GITHUB_SHA} 
```
should be sufficient for spec testing on fork, where `remotes/origin/${GITHUB_BASE_REF}` is generally the `master` branch while `GITHUB_SHA` is a merge commit from a fork repo.

### DIFF
--- a/scripts/health_metrics/get_updated_files.sh
+++ b/scripts/health_metrics/get_updated_files.sh
@@ -25,10 +25,7 @@ done
 
 dir=$(pwd)
 
-# Get most rescent ancestor commit.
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Thanks!

PR #9178, opened 2022-01-07
Add remaining features of ordered code.

wu-hui:
#no-changelog

### DIFF
--- a/Firestore/Protos/nanopb/google/type/latlng.nanopb.h
+++ b/Firestore/Protos/nanopb/google/type/latlng.nanopb.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Firestore/core/src/util/ordered_code.cc
+++ b/Firestore/core/src/util/ordered_code.cc
@@ -16,6 +16,9 @@
 
 #include "Firestore/core/src/util/ordered_code.h"
 
+#include <stddef.h>
+
+#include <cstdint>
 #include <limits>
 
 #include "Firestore/core/src/util/bits.h"
--- a/Firestore/core/src/util/ordered_code.cc
+++ b/Firestore/core/src/util/ordered_code.cc
@@ -16,8 +16,7 @@
 
 #include "Firestore/core/src/util/ordered_code.h"
 
-#include <stddef.h>
-
+#include <cstddef>
 #include <cstdint>
 #include <limits>
 
--- a/Firestore/core/src/util/ordered_code.cc
+++ b/Firestore/core/src/util/ordered_code.cc
@@ -16,11 +16,15 @@
 
 #include "Firestore/core/src/util/ordered_code.h"
 
+#include <limits>
+
 #include "Firestore/core/src/util/bits.h"
 #include "Firestore/core/src/util/hard_assert.h"
+#include "absl/base/casts.h"
 #include "absl/base/internal/endian.h"
 #include "absl/base/internal/unaligned_access.h"
 #include "absl/base/port.h"
+#include "absl/strings/internal/resize_uninitialized.h"
 
 #if !defined(ABSL_IS_LITTLE_ENDIAN) && !defined(ABSL_IS_BIG_ENDIAN)
 #error \
@@ -28,8 +32,10 @@
        "ABSL_IS_LITTLE_ENDIAN must be defined"
 #endif
 
+#define UNALIGNED_LOAD16 ABSL_INTERNAL_UNALIGNED_LOAD16
 #define UNALIGNED_LOAD32 ABSL_INTERNAL_UNALIGNED_LOAD32
 #define UNALIGNED_LOAD64 ABSL_INTERNAL_UNALIGNED_LOAD64
+#define UNALIGNED_STORE16 ABSL_INTERNAL_UNALIGNED_STORE16
 #define UNALIGNED_STORE32 ABSL_INTERNAL_UNALIGNED_STORE32
 #define UNALIGNED_STORE64 ABSL_INTERNAL_UNALIGNED_STORE64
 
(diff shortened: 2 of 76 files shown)


Firestore/core/src/util/ordered_code.h
```diff
@@ -122,6 +151,21 @@ class OrderedCode {
   OrderedCode& operator=(const OrderedCode&) = delete;
```
cherylEnkidu, line 151: Could you please share with me why we don't need to delete move ctor and move assignment operator here?

Firestore/Protos/nanopb/google/type/latlng.nanopb.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
```
cherylEnkidu, line 2: Why only the years under Protos changed, but not the others?

Firestore/core/src/util/ordered_code.h
```diff
@@ -61,21 +61,29 @@ class OrderedCode {
...
   static void WriteInfinity(std::string* dest);
+  static void WriteInfinityDecreasing(std::string* dest);
 
-  /**
-   * Special string append that can only be used at the tail end of
-   * an encoded string -- blindly appends "str" to "*dest".
-   */
+  // Special string append that can only be used at the tail end of
+  // an encoded string -- blindly appends "str" to "*dest".
   static void WriteTrailingString(std::string* dest, absl::string_view str);
```
cherylEnkidu, line 87: Is there a reason to skip `WriteTrailingCord(std::string* dest, const absl::Cord& cord);` ?

Firestore/core/src/util/ordered_code.h
```diff
@@ -122,6 +151,21 @@ class OrderedCode {
   OrderedCode& operator=(const OrderedCode&) = delete;
 };
 
+// -----------------------------
+// Implementation details follow
```
cherylEnkidu, line 155: Why we put implementation here instead of .cc ?

Firestore/core/src/util/ordered_code.cc
```diff
@@ -16,20 +16,26 @@
 
 #include "Firestore/core/src/util/ordered_code.h"
 
+#include <limits>
+
```
cherylEnkidu, line 20: Is there a reason to skip `#include <cstdint>` Although it might be included in some other libraries but according to what I found online: `that is not something you can rely on. Standard library implementors may one day decide that they will no longer do it like that.`

Firestore/core/src/util/ordered_code.h
wu-hui: My little research seems to suggest they should be deleted explicitly as well, at least there would be no harm. Thanks.

Firestore/Protos/nanopb/google/type/latlng.nanopb.h
cherylEnkidu: For script generating nanopb files, do we need to run it manually every time when changing the codebase?

paulb777: The copyright date should be the first year a file is published externally. It's ok to update after, but not necessary.

cherylEnkidu (changes requested):

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -53,47 +60,96 @@ bool OCReadIncreasing<std::string>(absl::string_view* src,
...
+  return OrderedCode::ReadDoubleDecreasing(src, result);
+}
 
 template <typename T>
-std::string OCWrite(T val, Direction direction) {
-  EXPECT_EQ(INCREASING, direction);  // DECREASING never implemented.
+std::string OCWrite(const T& val, Direction direction) {
   std::string result;
-  OCWriteIncreasing<T>(&result, val);
+  if (Direction::INCREASING == direction) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Please fix.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -410,14 +706,52 @@ TEST(OrderedCodeString, Infinity) {
...
+  OrderedCode::WriteString(&encoding, value);
+  EXPECT_THAT(encoding, testing::ElementsAre(0, 1));
+
+  encoding.clear();
+  OrderedCode::WriteStringDecreasing(&encoding, value);
+  EXPECT_THAT(encoding, testing::ElementsAre(0xff, 0xfe));
 }
 
-TEST(OrderedCodeString, EncodeDecode) {
+TEST(OrderedCodeStringString, EncodeDecode) {
```
schmidt-sebastian, line 752: Why the name change?

schmidt-sebastian (approved):
Please address @cherylEnkidu's comments before merging.

Firestore/core/src/util/ordered_code.cc
wu-hui: It's because the internal version used c style, i changed it back. Seems to be more sensesible.

Firestore/core/src/util/ordered_code.cc
```diff
@@ -16,20 +16,26 @@
...
 #include "Firestore/core/src/util/bits.h"
 #include "Firestore/core/src/util/hard_assert.h"
+#include "absl/base/casts.h"
 #include "absl/base/internal/endian.h"
 #include "absl/base/internal/unaligned_access.h"
 #include "absl/base/port.h"
+#include "absl/strings/internal/resize_uninitialized.h"
 
 #if !defined(ABSL_IS_LITTLE_ENDIAN) && !defined(ABSL_IS_BIG_ENDIAN)
```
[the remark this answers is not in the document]
wu-hui: The internal version uses `#include "util/endian/endian.h"` which is not available to us..which is why it uses the absl equivalent here i guess.

Firestore/core/src/util/ordered_code.cc
```diff
@@ -270,87 +407,102 @@ void OrderedCode::WriteNumIncreasing(std::string* dest, uint64_t val) {
...
-  return ReadStringInternal(src, result);
+  return ReadStringInternal<false>(src, result);
+}
+
+bool OrderedCode::ReadStringDecreasing(absl::string_view* src,
+                                       std::string* result) {
+  return ReadStringInternal<true>(src, result);
 }
 
 bool OrderedCode::ReadNumIncreasing(absl::string_view* src, uint64_t* result) {
```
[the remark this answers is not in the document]
wu-hui: This optimization uses internal code `~BigEndian::Load<uint64_t>` which is not available here. Again, I *guess* this was why we did not copy the optimization.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -16,10 +16,11 @@
 
 #include "Firestore/core/src/util/ordered_code.h"
 
-#include <iostream>
 #include <limits>
 
 #include "Firestore/core/src/util/secure_random.h"
+#include "absl/base/casts.h"
+#include "gmock/gmock.h"
```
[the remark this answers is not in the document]
wu-hui: This seems to be an optimization for PowerPC instructions. The internal code tries its best to support as many platforms as possible. Our SDK OTOH, is unlikely to run on PowerPCs.

Firestore/core/test/unit/util/ordered_code_test.cc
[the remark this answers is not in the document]
wu-hui: Done.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -135,7 +199,7 @@ static void TestWriteAppends(Direction d, T first, U second) {
 
 template <typename T>
 static void TestNumbers(T multiplier) {
-  for (int j = 0; j < 1; ++j) {
+  for (int j = 0; j < 2; ++j) {
```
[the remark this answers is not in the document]
wu-hui: Yes, I think that is correct.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -344,39 +633,46 @@ TEST(OrderedCodeInvalidEncodingsTest, NonCanonical) {
...
 
     EXPECT_NE(OCWrite<uint64_t>(0, INCREASING), non_minimal);
-
-#if defined(NDEBUG)
-    EXPECT_ANY_THROW(TestRead<uint64_t>(INCREASING, non_minimal));
-#else   // defined(NDEBUG)
-    absl::string_view s(non_minimal);
-    EXPECT_ANY_THROW(OrderedCode::ReadNumIncreasing(&s, NULL));
-#endif  // defined(NDEBUG)
+    {
```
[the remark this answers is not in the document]
wu-hui: Because the SDK build's behavior is similar to the DEBUG_MODE from google3, that it reports error and die immediately if this happens.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -398,7 +694,7 @@ TEST(OrderedCodeString, Infinity) {
   EXPECT_EQ(1u, s.size());
```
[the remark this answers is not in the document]
wu-hui: I do not think this is necessary?

Firestore/core/test/unit/util/ordered_code_test.cc
wu-hui: Fixed.

Firestore/core/test/unit/util/ordered_code_test.cc
```diff
@@ -176,48 +264,83 @@ static bool CompareStrings(const std::string& a,
...
 }
 
 // Helper routine for testing TEST_SkipToNextSpecialByte
-static size_t FindSpecial(const std::string& x) {
+static int64_t FindSpecial(const std::string& x) {
   const char* p = x.data();
   const char* limit = p + x.size();
   const char* result = OrderedCode::TEST_SkipToNextSpecialByte(p, limit);
-  return static_cast<size_t>(result - p);
+  return result - p;
```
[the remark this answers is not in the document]
wu-hui: Because we are not THAT performance sensitive, and I am too lazy to include something only exist for C++..

wu-hui (commented):
Thanks..if you have more questions about my replies, you can also ping me directly.

cherylEnkidu (approved):
LGTM

PR #9181, opened 2022-01-07
Update endpoint of the Metrics Service

granluo:
Update endpoint of the Metrics Service


yifanyang (approved):
Does this change also cover binary size calls?

granluo:
> Does this change also cover binary size calls?

Yes, This is used in [binary size measurement ](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/health_metrics/generate_code_coverage_report/Sources/BinarySizeReportGenerator/main.swift#L81)as well.

PR #9182, opened 2022-01-07
Updates to the SDK for future Flutter support

elenadoty:
#no-changelog

### DIFF
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
@@ -31,7 +31,8 @@ extern NSString* const FIRCLSUserNameKey;
 extern NSString* const FIRCLSUserEmailKey;
 extern NSString* const FIRCLSDevelopmentPlatformNameKey;
 extern NSString* const FIRCLSDevelopmentPlatformVersionKey;
-extern NSString* const FIRCLSFlutterOnDemandExceptionKey;
+extern NSString* const FIRCLSFlutterOnDemandRecordedExceptionsKey;
+extern NSString* const FIRCLSFlutterOnDemandDroppedExceptionsKey;
 #endif
 
 extern const uint32_t FIRCLSUserLoggingMaxKVEntries;
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
@@ -31,8 +31,8 @@ extern NSString* const FIRCLSUserNameKey;
 extern NSString* const FIRCLSUserEmailKey;
 extern NSString* const FIRCLSDevelopmentPlatformNameKey;
 extern NSString* const FIRCLSDevelopmentPlatformVersionKey;
-extern NSString* const FIRCLSFlutterOnDemandRecordedExceptionsKey;
-extern NSString* const FIRCLSFlutterOnDemandDroppedExceptionsKey;
+extern NSString* const FIRCLSOnDemandRecordedExceptionsKey;
+extern NSString* const FIRCLSOnDemandDroppedExceptionsKey;
 #endif
 
 extern const uint32_t FIRCLSUserLoggingMaxKVEntries;
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
@@ -31,6 +31,7 @@ extern NSString* const FIRCLSUserNameKey;
 extern NSString* const FIRCLSUserEmailKey;
 extern NSString* const FIRCLSDevelopmentPlatformNameKey;
 extern NSString* const FIRCLSDevelopmentPlatformVersionKey;
+extern NSString* const FIRCLSFlutterOnDemandExceptionKey;
 #endif
 
 extern const uint32_t FIRCLSUserLoggingMaxKVEntries;
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
@@ -32,7 +32,10 @@ NSString *const FIRCLSUserIdentifierKey = @"com.crashlytics.user-id";
 NSString *const FIRCLSDevelopmentPlatformNameKey = @"com.crashlytics.development-platform-name";
 NSString *const FIRCLSDevelopmentPlatformVersionKey =
     @"com.crashlytics.development-platform-version";
-NSString *const FIRCLSFlutterOnDemandExceptionKey = @"com.crashlytics.on-demand-flutter-exception";
+NSString *const FIRCLSFlutterOnDemandRecordedExceptionsKey =
+    @"com.crashlytics.on-demand-recorded-flutter-exceptions";
+NSString *const FIRCLSFlutterOnDemandDroppedExceptionsKey =
+    @"com.crashlytics.on-demand-dropped-flutter-exceptions";
 
 // Empty string object synchronized on to prevent a race condition when accessing AB file path
 NSString *const FIRCLSSynchronizedPathKey = @"";
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
@@ -32,6 +32,7 @@ NSString *const FIRCLSUserIdentifierKey = @"com.crashlytics.user-id";
 NSString *const FIRCLSDevelopmentPlatformNameKey = @"com.crashlytics.development-platform-name";
 NSString *const FIRCLSDevelopmentPlatformVersionKey =
     @"com.crashlytics.development-platform-version";
+NSString *const FIRCLSFlutterOnDemandExceptionKey = @"com.crashlytics.on-demand-flutter-exception";
 
 // Empty string object synchronized on to prevent a race condition when accessing AB file path
 NSString *const FIRCLSSynchronizedPathKey = @"";
--- a/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
+++ b/Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
@@ -32,10 +32,10 @@ NSString *const FIRCLSUserIdentifierKey = @"com.crashlytics.user-id";
 NSString *const FIRCLSDevelopmentPlatformNameKey = @"com.crashlytics.development-platform-name";
 NSString *const FIRCLSDevelopmentPlatformVersionKey =
     @"com.crashlytics.development-platform-version";
-NSString *const FIRCLSFlutterOnDemandRecordedExceptionsKey =
-    @"com.crashlytics.on-demand-recorded-flutter-exceptions";
-NSString *const FIRCLSFlutterOnDemandDroppedExceptionsKey =
-    @"com.crashlytics.on-demand-dropped-flutter-exceptions";
+NSString *const FIRCLSOnDemandRecordedExceptionsKey =
+    @"com.crashlytics.on-demand-recorded-exceptions";
+NSString *const FIRCLSOnDemandDroppedExceptionsKey =
+    @"com.crashlytics.on-demand-dropped-exceptions";
 
 // Empty string object synchronized on to prevent a race condition when accessing AB file path
 NSString *const FIRCLSSynchronizedPathKey = @"";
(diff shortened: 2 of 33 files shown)


Crashlytics/Crashlytics/Components/FIRCLSUserLogging.m
```diff
@@ -32,6 +32,10 @@
 NSString *const FIRCLSDevelopmentPlatformNameKey = @"com.crashlytics.development-platform-name";
 NSString *const FIRCLSDevelopmentPlatformVersionKey =
     @"com.crashlytics.development-platform-version";
+NSString *const FIRCLSFlutterOnDemandRecordedExceptionsKey =
+    @"com.crashlytics.on-demand-recorded-flutter-exceptions";
```
mrober, line 36: Can we rename this to just be on-demand, not flutter specific? I hope we use this for other platforms in the future.

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,193 @@
...
+      [self calculateUploadDelay:YES];
+      return NO;  // Didn't record or submit then exception because no quota was available.
+    }
+  }
+}
+
+// Helper methods
+
+- (double)calculateUploadDelay:(BOOL)shouldUpdate {
+  // TODO: should return 0 if queue is empty?
```
mrober, line 127: If the queue is empty it should still calculate the delay just like a "NotFull" queue.

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,193 @@
...
+  double calculatedStepDuration = [self calculateStepDuration:shouldUpdate];
+  double power = pow(self.baseExponent, calculatedStepDuration);
+  NSNumber *calculatedUploadDelay =
+      [NSNumber numberWithDouble:(SEC_PER_MINUTE / self.uploadRate) * power];
+  NSComparisonResult result =
+      [[NSNumber numberWithDouble:MAX_DELAY_SEC] compare:calculatedUploadDelay];
+  return (result == NSOrderedAscending) ? MAX_DELAY_SEC : [calculatedUploadDelay doubleValue];
+}
+
+- (double)calculateStepDuration:(BOOL)shouldUpdate {
```
mrober, line 138: I cannot tell if you already handled this case, but be careful how you initialize self.currentStep and self.lastUpdated so that if nothing happens for a long time after the app initialized, then the first on-demand fatal comes through, that the queueNotFullDuration still gets calculated correctly.

mrober: Nit: I think "step duration" is a bit ambiguous here. The method calculateStepDuration returns an exponent to use in a formula that gives a duration of time, and the member variable self.stepDuration is an actual duration of time. Maybe name this method just "calculateStep" ?

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
elenadoty: It defaults to a six second delay though - I think we want to have it upload immediately if there haven't been any events uploaded before

Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
```diff
@@ -31,6 +31,8 @@ extern NSString* const FIRCLSUserNameKey;
 extern NSString* const FIRCLSUserEmailKey;
 extern NSString* const FIRCLSDevelopmentPlatformNameKey;
 extern NSString* const FIRCLSDevelopmentPlatformVersionKey;
+extern NSString* const FIRCLSOnDemandRecordedExceptionsKey;
+extern NSString* const FIRCLSOnDemandDroppedExceptionsKey;
```
samedson, line 35: Are we committed to using "Exceptions" as the name. Should we use "Events"?

Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
```diff
@@ -60,7 +62,10 @@ - (instancetype)initWithManagerData:(FIRCLSManagerData *)managerData {
   _installIDModel = managerData.installIDModel;
   _fileManager = managerData.fileManager;
   _analytics = managerData.analytics;
-
+  _onDemandModel = [[FIRCLSOnDemandModel alloc]
+      initWithOnDemandUploadRate:[managerData.settings onDemandUploadRate]
+                            base:[managerData.settings onDemandBackoffBase]
+                    stepDuration:[managerData.settings onDemandBackoffStepDuration]];
```
samedson, line 68: Nit: I think you can use `.onDemandUploadRate`, etc if you make these `@property(readonly)` and implement the getter

samedson: Kindof a nit since we're not doing this for other settings

Crashlytics/Crashlytics/Components/FIRCLSUserLogging.h
elenadoty: Since there's already a cl submitted with this to change the Procella schema, let's leave it as exception

Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
elenadoty: done

Crashlytics/Crashlytics/Models/FIRCLSInternalReport.m
```diff
@@ -135,7 +136,7 @@ + (NSArray *)crashFileNames {
 #if CLS_MACH_EXCEPTION_SUPPORTED
       FIRCLSReportMachExceptionFile,
 #endif
-      FIRCLSReportSignalFile, FIRCLSMetricKitFatalReportFile
+      FIRCLSReportSignalFile, FIRCLSMetricKitFatalReportFile, FIRCLSCustomCrashIndicatorFile
```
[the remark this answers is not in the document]
elenadoty: done

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,202 @@
...
+// Helper methods
+
+- (double)calculateUploadDelay {
+  double calculatedStepDuration = [self calculateStepDuration];
+  double power = pow(self.base, calculatedStepDuration);
+  NSNumber *calculatedUploadDelay =
+      [NSNumber numberWithDouble:(SEC_PER_MINUTE / self.uploadRate) * power];
+  NSComparisonResult result =
+      [[NSNumber numberWithDouble:MAX_DELAY_SEC] compare:calculatedUploadDelay];
+  [self incrementQueuedOperationCount:-1];
```
samedson, line 132: This seems hidden in this method based on its name - I would expect `calculateUploadDelay` to be pure.  Can we call this in the dispatch_async block?

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+          self->_recordedOnDemandExceptionCount -= 1;
+          self->_droppedOnDemandExceptionCount += 1;
+        }
+      }
+      double uploadDelay = [self calculateUploadDelay];
+      if (dataCollectionEnabled) {
+        FIRCLSDebugLog(@"Submitted an on-demand exception, starting delay %.20f", uploadDelay);
+      } else {
+        FIRCLSDebugLog(@"Stored an on-demand exception, starting delay %.20f", uploadDelay);
+      }
```
samedson, line 129: Can we move these logs into the if statements above so we don't duplicate the if statement? If I'm reading correctly `calculateUploadDelay` doesn't depend on the output of the above if statement

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+  return [self incrementQueuedOperationCount:0];
+}
+
+- (void)setQueuedOperationsCount:(int)queuedOperations {
+  [self incrementQueuedOperationCount:queuedOperations];
+}
+
+- (BOOL)isQueueFull {
+  return [self incrementQueuedOperationCount:0] >= self.uploadRate;
+  ;
```
samedson, line 206: Extra

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+- (int)getOrIncrementOnDemandEventCountForCurrentRun:(BOOL)increment {
+  @synchronized(self) {
+    if (increment) {
+      _recordedOnDemandExceptionCount += 1;
+    }
+    return self.recordedOnDemandExceptionCount;
+  }
+}
+
+- (int)getOrIncrementDroppedOnDemandEventCountForCurrentRun:(BOOL)increment {
```
samedson, line 180: I think we can call these 2 methods `increment...` and drop the `getOr...` - think it's clear enough what it returns

samedson: Wait nvm

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+        [existingReportManager handleOnDemandReportUpload:activeReportPath
+                                      dataCollectionToken:dataCollectionToken
+                                                 asUrgent:YES];
+      } else {
+        [self.storedActiveReportPaths insertObject:activeReportPath atIndex:0];
+        if ([self.storedActiveReportPaths count] > FIRCLSMaxUnsentReports) {
+          [self.fileManager removeItemAtPath:[self.storedActiveReportPaths lastObject] error:nil];
+          [self.storedActiveReportPaths removeLastObject];
+          self->_recordedOnDemandExceptionCount -= 1;
+          self->_droppedOnDemandExceptionCount += 1;
```
samedson, line 121: I think we can do `self.recordedOnDemandExceptionCount ...` for these

elenadoty: they're read-only so we can't

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+    self.lastUpdated = currentTime;
+  }
+
+  return calculatedDuration;
+}
+
+- (int)getOrIncrementOnDemandEventCountForCurrentRun:(BOOL)increment {
+  @synchronized(self) {
+    if (increment) {
+      _recordedOnDemandExceptionCount += 1;
```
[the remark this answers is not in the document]
elenadoty: likewise, readonly properties - could make them readwrite if you think that's better

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
```diff
@@ -0,0 +1,209 @@
...
+- (int)incrementQueuedOperationCount:(int)increment {
+  @synchronized(self) {
+    _queuedOperationsCount += increment;
+    return self.queuedOperationsCount;
+  }
+}
+
+- (int)getQueuedOperationsCount {
+  return [self incrementQueuedOperationCount:0];
+}
```
samedson, line 198: So in general I think you can ditch the `increment...` methods altogether because they're used inconsistently.

For this, just implement the getter as synchronized, and this is a time when you do want to use the underscore version, so

```
- (int)queuedOperationsCount {
  @synchronized(self) {
    return _queuedOperationsCount;
  }
}
```

Crashlytics/UnitTests/FIRCLSOnDemandModelTests.m
```diff
@@ -0,0 +1,206 @@
...
+                                             withDataCollectionEnabled:NO
+                                            usingExistingReportManager:self.existingReportManager];
+
+    XCTAssertTrue(success);
+  }
+
+  // Once we've finished processing, there should be only FIRCLSMaxUnsentReports recorded with the
+  // rest considered dropped. The recorded events should be stored in storedActiveReportPaths which
+  // is kept in sync with the contents of the active path.
+  sleep(60);  // wait for the queue to empty
```
samedson, line 148: We unfortunately can't sleep in tests - I believe there's a way around this though

Crashlytics/Crashlytics/Models/FIRCLSOnDemandModel.m
samedson: Or use the increment / decrement methods?

Crashlytics/UnitTests/FIRCLSOnDemandModelTests.m
elenadoty: Can't in what way - shouldn't or actually can't?

samedson (approved):

Issue #9074, opened 2021-12-09
Deprecate Firebase pod, Firebase module, and Firebase.h


paulb777:
Firebase has historically enabled all public APIs for all installed products to be available via importing the Firebase module or Firebase.h.  This is implemented via preprocessor commands in Firebase.h along with some fake module infrastructure.

As Firebase becomes more Swift-centric, it's not clear we can continue supporting that functionality since Swift does not allow similar preprocessor logic and its modules would require all dependencies to be built - defeating the purpose of only installing/building what is needed.

As a result, starting the next major release, Firebase users will need to import a Firebase module for each Firebase product used.

Let us know if you have any concerns about this proposal or suggestions for alternatives.

emmanuel-montoya:
Could you add some screenshots or snippets for future references?

paulb777:
@emmanuel-montoya Yep, we'll do a full docs/samples/snippets scrub if we implement this.

a-25:
Hello, @paulb777. Are you planning to implement this feature?
I replaced internal links to several Firebase/<framework>s so let me know if it is good to go on.

paulb777:
Thanks @a-25! We do plan to implement, but not until our next major release, which isn't yet scheduled.  It will likely be sometime between the spring and fall of this year.  Once we get closer, we'll evaluate and merge these changes.

paulb777:
Perhaps, we should keep the Firebase pod even without the module to ease transition and beta version specification?  cc: @ryanwilson

paulb777:
Based on experimentation and https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/importing_swift_into_objective-c, I'm not able to find a way to check for existence from Swift Package Manager. `#import "FirebaseFunctions-Swift.h"` is only visible from the current module. Otherwise a module import is needed and there doesn't seem to be a corresponding directive in Objective C to the Swift `canImport`

paulb777:
Swift Forum question at https://forums.swift.org/t/canimport-for-objective-c-clients-of-swift-packages/55294

a-25:
> #9333 demonstrates a way to use the Firebase pod for CocoaPods with its Firebase.h. If we can find a similar solution for Swift PM, we may be able to keep the Firebase pod. cc: @a-25

@paulb777, do you see any advantage of keeping Firebase pod? I think it could be technically removed.

paulb777:
[...]

PR #9186, opened 2022-01-10
fix: `Firebase` podspec missing ML Model Downloader for macOS

Salakar:
### Discussion

[FirebaseMLModelDownloader.podspec](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMLModelDownloader.podspec) supports osx as a target however the subspec in `Firebase.podspec` does not, this leads to build errors when adding the following as a dependency in podspecs:

```ruby
s.dependency 'Firebase/MLModelDownloader'
```

A work around for now is do just depend on an exact version of the ML pod directly e.g.:

```ruby
s.dependency 'FirebaseMLModelDownloader', '~> 8.10.0-beta'
```

### Testing

I've tested ML Model Downloader pod does work on macOS, via a Flutter macOS app that downloads/deletes models:

image

 - it builds without issues and downloading our test model works fine

### API Changes

  * None

### DIFF
--- a/Firebase.podspec
+++ b/Firebase.podspec
@@ -171,7 +171,7 @@ Simplify your app development, grow your user base, and monetize more effectivel
 
   s.subspec 'MLModelDownloader' do |ss|
     ss.dependency 'Firebase/CoreOnly'
-    ss.ios.dependency 'FirebaseMLModelDownloader', '~> 8.11.0-beta'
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Thanks! 

FYI, in case you missed it, we will likely deprecate the Firebase podspec and remove it in our next release. We recommend to migrate away from using it. See #9074

PR #8388, opened 2021-07-14
Make App Check core availability for all supported platforms

maksymmalyhin:
b/193554541

- App Check SDK availability now matches to Firebase Core availability
- App Attest and DeviceCheck providers availability matches the underlying platform API availability (see [DCDevice](https://developer.apple.com/documentation/devicecheck/dcdevice) and [DCAppAttestService](https://developer.apple.com/documentation/devicecheck/dcappattestservice))

PR #9187, opened 2022-01-10
fix: issue where deployment targets for subspec `AppCheck` in `Firebase` podspec does not match `FirebaseAppCheck` podspec

Salakar:
### Discussion

The deployment targets defined in https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAppCheck.podspec#L20-L23 are not reflected in the [`AppCheck` subspec](https://github.com/firebase/firebase-ios-sdk/blob/master/Firebase.podspec#L102-L104) in the main `Firebase` podspec. This PR updates them to match.

### Testing

  * N/A

### API Changes

  * N/A

### DIFF
--- a/Firebase.podspec
+++ b/Firebase.podspec
@@ -99,9 +99,10 @@ Simplify your app development, grow your user base, and monetize more effectivel
   s.subspec 'AppCheck' do |ss|
     ss.dependency 'Firebase/CoreOnly'
     ss.dependency 'FirebaseAppCheck', '~> 8.11.0-beta'
-    ss.ios.deployment_target = '11.0'
-    ss.osx.deployment_target = '10.15'
-    ss.tvos.deployment_target = '11.0'
+    ss.ios.deployment_target = '9.0'
+    ss.osx.deployment_target = '10.12'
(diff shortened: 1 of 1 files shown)


paulb777 (changes requested):
@Salakar Thanks for pointing this out. The fix is actually the obvious to update FirebaseAppCheck.podspec to match since AppCheck requires iOS 11 etc.  Let us know if you want to continue. Otherwise, we'll make the correction before the next release.

paulb777:
Hmm, on further investigation, I'm not sure the correct setup here. https://firebase.google.com/docs/ios/learn-more#firebase_library_support_by_platform shows three different lines for AppCheck.

paulb777 (approved):
Based on the updates in #8388, this change should have been done then as well, so it looks good. Thanks!  and thanks to @ryanwilson for tracking down the provenance.

Semty:
@paulb777 hey! it seems that the desired effect is seen only with using Cocoapods, but with SPM I get this after adding AppCheck for watchOS target:
```DeviceCheck is not available when building for watchOS.```

PR #9190, opened 2022-01-10
Add pods in spec testing

granluo:


scripts/health_metrics/code_coverage_file_list.json
```diff
@@ -40,6 +40,39 @@
       "\\.github/workflows/auth\\.yml"
     ]
   },
+  {
+    "sdk": "combine",
```
paulb777, line 44: The combine podspecs are only for testing and not published, so I'm not sure they should be here?

scripts/health_metrics/code_coverage_file_list.json
```diff
@@ -118,6 +167,14 @@
       "\\.github/workflows/messaging\\.yml"
     ]
   },
+  {
+    "sdk": "mlmodeldownloader",
```
paulb777, line 171: Does code coverage work for Swift pods? Do you have an example report?

granluo: Code coverage is not working for newly added pods in this PR. Code coverage is more about hard coding here.
https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/health-metrics-presubmit.yml#L20-L35
Will also update the file name to file_patterns.json to avoid confusion.

paulb777 (approved):

PR #9202, opened 2022-01-12
Fix 32bit Firebase Distro build regression

paulb777:
#9142 broke the Firebase zip distribution build because the Firebase build changed the minimum iOS version to 14 just for building.

This PR fixes that regression while still fixing the issue #9142 resolved. It also makes the 32 bit build management a bit more robust.


paulb777:
The release zip from https://github.com/firebase/firebase-ios-sdk/actions/runs/1689192258 compares reasonably to a past zip and shows the restored 32bit slices when compared to last night's nightly. I'll post it as the 8.11.0 RC0 zip

Issue #8973, opened 2021-11-16
Migrate FirebaseCore Private to a new FirebaseCoreInternal library


paulb777:
To enable Swift libraries to call internal methods like `registerInternalLibrary`

paulb777:
Alternatively, we may want to move the public APIs to a Swift implemented library that depends on a FirebaseCoreObj, in which the current private headers are now public - but the FirebaseCoreObj is only supported for internal use.

ryanwilson:
Pretty sure this is done in the `functions-swift-main` branch already.

paulb777:
This is merged in the v9 branch. It will likely be renamed FirebaseCoreExtensionInternal

PR #9214, opened 2022-01-17
FirebaseCoreSwift additive API

ryanwilson:
This is an additive, internal only Swift SDK that lets other products
access internal FirebaseCore APIs in a reflective but easy to consume
manner.

TODO:
~- [ ] Evaluate the best place  to call `registerLibrary:` from our extensions*~ Not needed
- [ ] Add component container and migrate MLModelDownloader to use it
- [ ] Add CocoaPods library if needed
- [X] Discuss if it's a reasonable implementation for #8973

~* none of our extensions have a base class or anything that gets called without user interaction (unlike `+initialize`). There's not a great spot to call `registerLibrary:` as far as I can see.~ Discussed and we don't need it right now.


ryanwilson:
> none of our extensions have a base class or anything that gets called without user interaction (unlike +initialize). There's not a great spot to call registerLibrary: as far as I can see.

One (not ideal) thought is to define a class visible to Obj-C and then use `NSClassFromString` inside the standard Core and check if the class exists:

```
@objc(FIRCoreSwift)
private class FirebaseCoreSwift: NSObject {}
```

This is similar to our existing but not-continued CoreDiagnostics code: https://github.com/firebase/firebase-ios-sdk/blob/bdac2b87f6d6e604deae914a970a75136b3a62b5/Firebase/CoreDiagnostics/FIRCDLibrary/FIRCoreDiagnostics.m#L378-L381. Definitely fragile and requires Core to know about all the Swift extensions, but it should work.

Package.swift
```diff
@@ -233,6 +233,16 @@ let package = Package(
         .headerSearchPath("../../.."),
       ]
     ),
+    .target(
+      name: "FirebaseCoreSwift",
```
paulb777, line 237: How about `FirebaseCoreInternalSwift`?  To be clear that these APIs are only for internal Firebase usage?

FirebaseCoreSwift/Sources/FirebaseAppInternal.swift
```diff
@@ -0,0 +1,97 @@
...
+  static var isDefaultAppConfigured: Bool {
+    // First two arguments are the class (FirebaseApp) and the selector name.
+    typealias VoidBoolFunc = @convention(c) (AnyClass, Selector) -> Bool
+    let sel = NSSelectorFromString("isDefaultAppConfigured")
+    let isDefaultAppConfigured = generatePrivateClassFunc(sel, from: self, type: VoidBoolFunc.self)
+    return isDefaultAppConfigured(self, sel)
+  }
+
+  /// Registers a given third-party library with the given version number to be reported for
+  /// analytics. Redirects to the Obj-C method of the same name.
```
paulb777, line 40: `analytics` is overloaded. Is there a better term?

FirebaseCore/Sources/Private/FIRAppInternal.h
```diff
@@ -89,6 +89,9 @@ extern NSString *const FIRAuthStateDidChangeInternalNotificationUIDKey;
 
 /**
  * A flag indicating if this is the default app (has the default app name).
+ *
+ * @note This property is referenced by selector name in `FirebaseAppInternal.swift`. If this
```
paulb777, line 93: Why `@note` here but `NOTE` above?

paulb777 (commented):
We may not need `registerLibrary` from extensions if we're soon going to merge the extensions with the main libraries.

Perhaps we can defer the podspec until we need these APIs from the libraries versus the tests?

I think this can close #8973.

FirebaseCoreSwift/Sources/FirebaseAppInternal.swift
ryanwilson: I'll come up with something - I copied straight from the old header.

FirebaseCore/Sources/Private/FIRAppInternal.h
ryanwilson: The `NOTE` above are not documentation comments, just regular comments to call out to the reader. This one is a documentation comment so it'll show up in Xcode docs under the "Note" section since it's part of the header.

Screen Shot 2022-01-18 at 12 59 04 PM

FirebaseCoreSwift/Sources/FirebaseAppInternal.swift
ryanwilson: Actually, removed the function since we don't need it right now.

Package.swift
```diff
@@ -233,6 +233,16 @@ let package = Package(
         .headerSearchPath("../../.."),
       ]
     ),
+    .target(
```
paulb777, line 236: Do we want to separate the CoreInternal APIs only for tests versus APIs needed by libraries into separate libraries?

In other words should this be FirebaseCoreInternalTestsSwift or part of FirebaseTestingSupport/Core?

Package.swift
```diff
@@ -148,7 +148,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Note that CI changes the version to the head of main for CI.
       // See scripts/setup_spm_tests.sh.
-      .exact("8.9.1")
+      .branch("main")
```
paulb777, line 151: Undo before merge.

paulb777 (approved):
OK to merge if you want to checkpoint here.

ryanwilson:
Thanks for the review! These changes aren't needed right now and it'll be pretty easy to bring them in after the fact so I'm going to close this as a proof of concept and can re-open or copy-paste as needed from here.

PR #9215, opened 2022-01-18
Remove never released Firebase Segmentation

paulb777:
#no-changelog


ryanwilson (approved):
LGTM on green, thanks!

paulb777:
The release CI failure is because it finds the podspecs from the repo instead of the PR. Since it is rare to delete a podspec, I'll go ahead and merge and the ci failure should get resolved. cc: @granluo

granluo:
> The release CI failure is because it finds the podspecs from the repo instead of the PR. Since it is rare to delete a podspec, I'll go ahead and merge and the ci failure should get resolved. cc: @granluo

The release and prerelease workflows are supposed to run from the latest releasing tag and the head of the main branch. The [pull_request paths](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/release.yml#L6-L7) are for testing workflow, not for podspecs, to ensure the update of those files will not block the workflow before a PR was merged.

PR #9226, opened 2022-01-19
Fix SPM AppCheck build for watchOS

### LINKED ISSUE #9191: AppCheck support for watchOS
### Feature proposal

* Firebase Component: AppCheck

We've implemented support for AppCheck in our app ecosystem (iOS, Android, Web) and thinking about enforcing App Check verification. 

The main obstacle is that we have watchOS as well and it seems that AppCheck doesn't support it (well, of course, DeviceCheck and AppAttest just not available on watchOS). 

What is the expected behaviour in this situation? Should we just abandon the idea of using AppCheck in a foreseeable future or there is some workaround we could do?

paulb777:
Fix #9191

### DIFF
--- a/.github/workflows/app_check.yml
+++ b/.github/workflows/app_check.yml
@@ -80,25 +80,9 @@ jobs:
     # Don't run on private repo unless it is a PR.
     if: (github.repository == 'Firebase/firebase-ios-sdk' && github.event_name == 'schedule') || github.event_name == 'pull_request'
     runs-on: macos-11
-    steps:
-    - uses: actions/checkout@v2
-    - uses: mikehardy/buildcache-action@50738c6c77de7f34e66b870e4f8ede333b69d077
-      with:
-        cache_key: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


.github/workflows/app_check.yml
```diff
@@ -80,25 +80,9 @@ jobs:
...
-      run: scripts/third_party/travis/retry.sh ./scripts/build.sh AppCheckUnitSwift iOS spm
-
-  spm-cron:
-    # Don't run on private repo.
-    if: github.event_name == 'schedule' && github.repository == 'Firebase/firebase-ios-sdk'
-    runs-on: macos-11
     strategy:
       matrix:
-        target: [tvOS, macOS, catalyst]
+        target: [iOS, tvOS, macOS, catalyst, watchOS]
```
paulb777, line 85: Adding watchOS and run all platforms for app check SPM on presubmit

ncooke3 (approved):
👍 LGTM on green

PR #9233, opened 2022-01-20
Re-establish streams when App Check token expires.

ehsannas:


Firestore/core/src/core/firestore_client.cc
```diff
@@ -145,10 +145,19 @@ std::shared_ptr<FirestoreClient> FirestoreClient::Create(
   };
 
   shared_client->app_check_credentials_provider_->SetCredentialChangeListener(
-      [](std::string) {
-        // Register an empty credentials change listener to activate token
-        // refresh.
+      [weak_client](const std::string&) {
+        auto shared_client = weak_client.lock();
+        if (!shared_client) return;
+        if (shared_client->credentials_initialized_) {
```
schmidt-sebastian, line 151: Can you add a comment here why it is okay to swallow the notification if we have not yet received an auth credential? I think it is okay since we will always fetch a token during client startup, so even if we ignore a token once here we would receive it at a later point.

schmidt-sebastian (commented):
Please add changelog here as well.

Firestore/core/src/core/firestore_client.cc
ehsannas: That's right. Done.

Firestore/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
```
schmidt-sebastian, line 1:
```suggestion
# Unreleased
```

schmidt-sebastian (approved):

PR #9239, opened 2022-01-21
Update to SwiftFormat 0.49.2

paulb777:
#no-changelog


ryanwilson (approved):
Overall LGTM, one thing we now need to be careful of is adding methods in a public extension could unintentionally publish a new API if we don't explicitly mark it as `internal` or `private`. Detection of API additions and changes will be very helpful.

Issue #9212, opened 2022-01-16
FR: Crashlytics key-value storage and user id should support nil values


ghugues:
### Feature proposal

* Firebase Component: Crashlytics

Crashlytics [allows to store a user id and arbitrary key/values](https://github.com/firebase/firebase-ios-sdk/blob/4377c7c574d2de2dde71fe3c2dfc8c2841a6f5d7/Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h#L77-L107) that will be sent along the next crash report. Currently this API is marked as non-nullable for the user id and custom values. It would be interesting to support nullable values as a way to remove a previously stored value. This is especially useful for the `setUserID:` method where it may be necessary to set this value to nil on logout.

The underlying API (`FIRCLSUserLoggingRecordUserKeyValue` and `FIRCLSUserLoggingRecordInternalKeyValue`) already supports this feature, so this is just a matter of adding the `nullable` annotation to the Objective-C header. In swift changing from the non nullable value to a nullable value should not be a breaking change.

elenadoty:
Hi @ghugues , 
Thanks for the feature request - what you describe is the intended behavior, we'll update it as soon as possible!

paulb777:
Fixed by #9289 and planned for 8.13.0

PR #9241, opened 2022-01-22
Added methods for removing Crashlytics user info (#9212)

a-25:


elenadoty:
Hi @a-25,
Thanks for the pr. Have you tried passing null as a value for a key you'd like to clear?

a-25:
Hi. @elenadoty. You mean that it is unnecessary to introduce new methods but better to add nullability to existing ones?

elenadoty:
Hi @a-25 ,
It turns out there was a long-standing bug in the SDK preventing this from working as intended. I've put up a PR that should fix the problem.

a-25:
Hi, @elenadoty. Got it.
Is there anything I could do in this PR?

elenadoty:
Hi @a-25 ,
There shouldn't be any need for more changes, but let me know if the solution doesn't work for you.

paulb777:
Thanks @a-25 for getting the fix started and thanks @elenadoty for landing it!

Issue #8878, opened 2021-10-28
tvOS Cloud-Synced Auth


nidegen:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.1
  * Firebase SDK version: 8.9.0
  * Installation method: `Swift Package Manager`
  * Firebase Component: Auth

#### Steps to reproduce:

Following [this](https://github.com/firebase/firebase-ios-sdk/pull/6924#issuecomment-772604212), I do share Auth over iCloud Keychain sync between iOS and macOS apps successfully.

However, if I try this with tvOS, it does not work. Does tvOS maybe not support Keychain Sharing?

Officially, tvOS supports it but there are some other users with [issues](https://developer.apple.com/forums/thread/18480).

I wonder if there could be an issue with Firebase's keychain access implementation.

#### Relevant Code:

```

    var tempUser: Firebase.User?
    do {
      try tempUser = Auth.auth().getStoredUser(forAccessGroup: accessGroupId)
    } catch let error as NSError {
      print("Error getting stored user: %@", error)
    }
    return tempUser // always nil
```

nidegen:
@ncooke3 any news here? Is this even supposed to be working?

ncooke3:
Hi @nidegen, apologies for the delay. No updates as of now. I'll look into it this week and report back.

nidegen:
No worries, thank you!

weixifan:
I filed an internal b/212482016 to track this.

nidegen:
nice! @weixifan so you can confirm this is not only an issue on my side, right?

nidegen:
Hi guys, how is it looking? Can you confirm this is a real bug?

ncooke3:
Hi @nidegen, sincere apologies for the delay. I'm working on reproducing the issue today. I have some questions:
1. Do you have an entitlements file associated with your tvOS target? Could you confirm that its contents match/correspond to that of your other targets?
2. Is an error being thrown? If so, could you share the underlying error message?
Thanks!

nidegen:
No worries! Unfortunately I am pretty sure everything is correct. I have it running for iOS and macOS. The thing is however that I am not sure if it is at all possible on tvOS. Some forum posts indicate that it could be unsupported: https://developer.apple.com/forums/thread/690714

So, I am still not sure if I should expect an error on my side or if I should give up:)

nidegen:
[...]

PR #9244, opened 2022-01-22
[Auth] Added note to `getStoredUser(forAccessGroup:)` documentation about tvOS keychain sharing issues

ncooke3:
While it's not a fix for \#8878, it should provide some clarity to developers who run into this issue in the future. 

Also, a radar has been filed for the issue in \#8878– details in internal b/212482016.


FirebaseAuth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
```
paulb777, line 1: Will be 8.12.0

paulb777 (approved):

FirebaseAuth/CHANGELOG.md
ncooke3: Done.

ryanwilson (approved):
LGTM - should we also log an error if that API is called on tvOS? May be more visible for folks.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2261,6 +2261,15 @@ - (nullable FIRUser *)getStoredUserForAccessGroup:(NSString *_Nullable)accessGro
 #endif  // TARGET_OS_WATCH
     user = [unarchiver decodeObjectOfClass:[FIRUser class] forKey:userKey];
   } else {
+#if TARGET_OS_TV
+    if (self.shareAuthStateAcrossDevices) {
+      FIRLogError(kFIRLoggerAuth, @"I-AUT000001",
```
paulb777, line 2266: Is this better than making the API unavailable for tvOS? Why not give customers a compile time error instead of a runtime error? cc: @ryanwilson

ncooke3: I realized I should verify if tvOS keychain sharing works on between tvOS apps on the same device. It works so this Auth API should work on tvOS when `shareAuthStateAcrossDevices` is false. Because of this, I don't think we should mark the API unavailable because we are blocking a valid use case on tvOS (sharing a user between tvOS apps) .

paulb777: Thanks for the clarification. Please make that clear in the note and log message.

ncooke3: Reworked in latest commits.

ncooke3:
@paulb777 @ryanwilson thanks for reviews. This PR is good for a final pass.

ncooke3:
CI failures are unrelated. (Facebook integration tests are failing.)

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2261,6 +2261,17 @@ - (nullable FIRUser *)getStoredUserForAccessGroup:(NSString *_Nullable)accessGro
 #endif  // TARGET_OS_WATCH
     user = [unarchiver decodeObjectOfClass:[FIRUser class] forKey:userKey];
   } else {
+#if TARGET_OS_TV
+    if (self.shareAuthStateAcrossDevices) {
+      FIRLogError(
+          kFIRLoggerAuth, @"I-AUT000001",
+          @"Due to tvOS system limitations, getting a stored user for a given access "
+          @"group is not supported on tvOS when sharing auth state across devices "
+          @"(`shareAuthStateAcrossDevices`) is enabled (#8878). This case will return `nil`. ");
```
paulb777, line 2270: run-on sentence? Perhaps a period at the end of 2269?  same below.

ncooke3: Thanks. I was trying to be precise but it's more verbose than anything. I reworked here and below in 2a944f7.

paulb777: Hmm. I still think there needs to be punctuation after "across devices"

ncooke3: Ah sorry, I simplified further in f7977ca

paulb777 (approved):

PR #9253, opened 2022-01-24
Partial fix for expired ID token issue

rosalyntan:
Partial fix for #6521, where the callback for `getIDToken` appears to sometimes be invoked hours after the method is called if the app goes into the background.

Not sure if this is the best fix, but this should at least alleviate some cases of this happening. For the remainder, developers can catch the token expiration error from the backend and request a new token.

### DIFF
--- a/FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
+++ b/FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
@@ -19,6 +19,7 @@
 #import "FirebaseAuth/Sources/Public/FirebaseAuth/FIRAuth.h"
 
 #import "FirebaseAuth/Sources/Auth/FIRAuthSerialTaskQueue.h"
+#import "FirebaseAuth/Sources/Auth/FIRAuthTokenResult_Internal.h"
 #import "FirebaseAuth/Sources/Auth/FIRAuth_Internal.h"
 #import "FirebaseAuth/Sources/Backend/FIRAuthBackend.h"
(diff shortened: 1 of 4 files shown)


FirebaseAuth/Sources/SystemService/FIRSecureTokenService.m
```diff
@@ -170,6 +172,20 @@ - (void)requestAccessToken:(FIRFetchAccessTokenCallback)callback {
...
            if (newAccessToken.length && ![newAccessToken isEqualToString:self->_accessToken]) {
+             FIRAuthTokenResult *tokenResult =
+                 [FIRAuthTokenResult tokenResultWithToken:newAccessToken];
+             // There is an edge case where the request for a new access token may be made right
+             // before the app goes inactive, resulting in the callback being invoked much later
+             // with an expired access token. This does not fully solve the issue, as if the
+             // callback is invoked less than an hour after the request is made, a token is not
+             // re-requested here but the approximateExpirationDate will still be off since that is
+             // computed at the time the token is received.
+             if (retryIfExpired && [tokenResult.expirationDate timeIntervalSinceNow] <= 0) {
```
sam-gc, line 183: Would it be worth re-fetching the token if the token will expire within 5 minutes as well? (Instead of only after it's expired)

rosalyntan: Good call, done.

paulb777 (approved):
LGTM, please add a release note

PR #9134, opened 2021-12-23
pass appData for In-App Messaging's test message

fumito-ito:
refs: #9126 

`FIRIAMFetchResponseParser` is better not to ignore appData for test message. This PR fixes #9126 problems particaly but not completely.

PR #9263, opened 2022-01-26
updating change log

eldhosembabu:
updating change log wrt https://github.com/firebase/firebase-ios-sdk/pull/9134


eldhosembabu:
We could hold this merge until the backend fix for the issue is done. wdyt?

paulb777 (approved):
I think it's fine to merge now since the 8.12.0 SDK will support the feature. If we think the back end change will take more than a few weeks, I would suggest adding something like - "This SDK fix will work once the backend is also updated."

eldhosembabu:
filed an internal bug to track the backend issue : b/216496087

eldhosembabu:
> I think it's fine to merge now since the 8.12.0 SDK will support the feature. If we think the back end change will take more than a few weeks, I would suggest adding something like - "This SDK fix will work once the backend is also updated."

Done. added the above mentioned comment. It seems like backend team needs prioritize this fix and probably would take couple of weeks.

PR #9265, opened 2022-01-27
[Draft] integrate multicast with Auth and Messaging

chliangGoogle:
First draft that mimic original implementation: https://github.com/firebase/firebase-ios-sdk/pull/8422/files.

This will be a feature branch that is not submitted after approval, will check out sub branch from here to keep working on SceneDelegate for auth and notification center for messaging.


FirebaseMessaging/Apps/SwiftUISample/SwiftUISample/SwiftUISampleApp.swift
```diff
@@ -92,11 +89,17 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
 @main
 struct SwiftUISampleApp: App {
   // Add the adapter to access notifications APIs in AppDelegate
-  @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+  @UIApplicationDelegateAdaptor(GULMulticastAppDelegate.self) var delegate
+  var appDelegate: UIApplicationDelegate
+
+  init() {
+    appDelegate = AppDelegate()
+    delegate.addInterceptor(with: appDelegate)
```
maksymmalyhin, line 97: At this point the results of `respondsToSelector` for `delegate` may have been already cached, so the methods defined in `AppDelegate` and not defined in `GULMulticastAppDelegate` would never be called. If I remember correctly the [default interceptor](https://github.com/google/GoogleUtilities/pull/50/files#diff-778491135fed4fb3384377f5073a5b61a616ee66a1816fa8d2688c37d9dc2aceR46)  and the [recommendation to create a subclass of GULMulticastAppDelegate to set it](https://github.com/peterfriese/FirebaseFrictionLogs/pull/2/files#diff-6191247189e17a8886c59848ce5471fa7761c58a786995c8612417e9916fbb62R12) were introduced. 

You can check it by debugging `GULMulticastAppDelegate.respondsToSelector` method and decide if you would like to replicate my original solution or try a different one.

chliangGoogle: Discussed offline and this is resolved. Thanks maksym!

paulb777:
Please resolve CI issues

chliangGoogle:
This PR will stay failing until we submit the googleutilities change

paulb777:
Closing since this is no longer planned.

PR #9272, opened 2022-01-29
Fix integration test target name

paulb777:
#no-changelog


paulb777:
After 23b931b, CI looks good - just failing because of the three purposeful warnings

b676af8 adds unit tests including the API coverage test which raises open questions starting with:
Screen Shot 2022-01-29 at 9 21 31 AM

paulb777:
Fixed by 08b9cd2. Three APIs were missing the `public` attribute.

paulb777:
@ryanwilson PTAL, comment and merge at your convenience

FirebaseFunctionsSwift.podspec
```diff
@@ -35,20 +35,16 @@ Swift SDK Extensions for Cloud Functions for Firebase.
     'FirebaseFunctionsSwift/Sources/**/*.swift',
   ]
 
-  s.dependency 'FirebaseFunctions', '~> 8.11'
-  s.dependency 'FirebaseSharedSwift', '~> 8.11'
-
-  # Using environment variable because of the dependency on the unpublished
-  # FirebaseFunctionsTestingSupport.
```
ryanwilson, line 42: Side note: we should see if we can now get rid of this in other products. No action needed for this PR.

FirebaseFunctionsSwift/Sources/Functions.swift
```diff
@@ -126,8 +140,10 @@ private enum Constants {
     Response: Decodable>(_ name: String,
                          requestAs: Request.Type = Request.self,
                          responseAs: Response.Type = Response.self,
-                         encoder: FirebaseDataEncoder = FirebaseDataEncoder(),
-                         decoder: FirebaseDataDecoder = FirebaseDataDecoder())
+                         encoder: FirebaseDataEncoder = FirebaseDataEncoder(
+                         ),
```
ryanwilson, line 144: Yikes, this formatting is a big rough. Curious if there's a way we can fix this later in swift lint

FirebaseFunctionsSwift/Sources/FunctionsContext.swift
```diff
@@ -38,21 +45,21 @@ internal class FunctionsContextProvider: NSObject {
...
-  internal func getContext(_ completion:  @escaping ((FunctionsContext, Error?) -> Void)) {
+  internal func getContext(_ completion: @escaping ((FunctionsContext, Error?) -> Void)) {
     let dispatchGroup = DispatchGroup()
 
-    var authToken: String? = nil
-    var appCheckToken: String? = nil
-    var error: Error? = nil
+    var authToken: String?
+    var appCheckToken: String?
+    var error: Error?
```
ryanwilson, line 53: Is this a swiftlint preference? I kind of like having the `nil` explicitly declared there, although maybe that's my ObjC days talking. WDYT?

FirebaseFunctionsSwift/Sources/FunctionsError.swift
```diff
@@ -140,7 +146,7 @@ extension FunctionsErrorCode {
     switch name {
```
ryanwilson, line 146: This and the next function could go away if we switched the raw value of the enum, but that's a breaking source change unfortunately (changing the value of the enums, and the underlying type). Likely wouldn't affect anyone but shouldn't do it now.

FirebaseFunctionsSwift/Tests/Unit/FunctionsAPITests.swift
```diff
@@ -0,0 +1,142 @@
+// Copyright 2021 Google LLC
```
ryanwilson, line 1: Nit: 2022

ryanwilson (approved):
Just a few comments!

FirebaseFunctionsSwift.podspec
paulb777: Agreed. We can also get rid of all the Objective C soon :)

FirebaseFunctionsSwift/Sources/Functions.swift
```diff
@@ -91,27 +105,27 @@ private enum Constants {
...
+                                               appCheck: appCheck)
     self.fetcherService = fetcherService
   }
 
   /**
    * Creates a reference to the Callable HTTPS trigger with the given name.
    * @param name The name of the Callable HTTPS trigger.
    */
-  private func httpsCallable(_ name: String) -> HTTPSCallable {
+  public func httpsCallable(_ name: String) -> HTTPSCallable {
```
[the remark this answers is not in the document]
paulb777: Should be passing here.

FirebaseFunctionsSwift/Sources/Functions.swift
paulb777: Maybe there's an option or we should file a bug - but prefer to defer this now - until a later polish phase.

FirebaseFunctionsSwift/Sources/FunctionsContext.swift
paulb777: Yes this is from SwiftLint.  As a Swift newbie, I'm generally fine to go with SwiftLint recommendations and learn than the nil is implied - but open to having a different opinion for our code.

FirebaseFunctionsSwift/Tests/Unit/FunctionsAPITests.swift
paulb777: This is a copy of the existing `FirebaseFunctions/Tests/SwiftUnit/FunctionsAPITests.swift`

PR #9283, opened 2022-02-01
Fireperf: prewarm counters

leotianlizhan:
Adding 2 hidden metrics to `_fs`, regardless of RC flag `fpr_prewarm_detection`:
* `_fsapc`: ActivePrewarm counter. if ActivePrewarm identifies an app start as prewarmed, `_fsapc` of 1 will be added, otherwise a `_fsapc` of 0 will b added.
* `_fsddc`: Double dispatch counter. If double dispatch identifies an app as prewarmed, `_fsddc` of 1 will be added; if double dispatch identifies as not-prewarmed, 0 will be added; if double dispatch is disabled by developers through plist, then no custom metric will be added.

### DIFF
--- a/FirebasePerformance/CHANGELOG.md
+++ b/FirebasePerformance/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+* [Fixed] Potentially drop pre-warmed app start traces on iOS 15 and above (#9026). App start measurements are made only for cold app starts (without pre-warming).
+
 # Version 8.10.0
 * Fix a crash related to FPRSessionDetails. (#8691)
 * Fix heap-buffer overflow when encoding sessions. (#8849)
--- a/FirebasePerformance/CHANGELOG.md
+++ b/FirebasePerformance/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-* [Fixed] Potentially drop pre-warmed app start traces on iOS 15 and above (#9026). App start measurements are made only for cold app starts (without pre-warming).
+* [fixed] Potentially drop pre-warmed app start traces on iOS 15 and above (#9026). App start measurements are made only for cold app starts (without pre-warming).
 
 # Version 8.10.0
 * Fix a crash related to FPRSessionDetails. (#8691)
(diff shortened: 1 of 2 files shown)


FirebasePerformance/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+* [Fixed] Potentially drop pre-warmed app start traces on iOS 15 and above (#9026). App start measurements are made only for cold app starts (without pre-warming).
```
jeremyjiang-dev, line 2: Nit: Could you change it to lower case `fix`? Not sure if this will have any impact. Just to make sure we align with all Firebase CHANGELOG.md.

leotianlizhan: Done

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -202,18 +204,25 @@ - (BOOL)isApplicationPreWarmed {
...
+    isPrewarmed = isPrewarmed || [self isActivePrewarmEnabled];
+    [self.activeTrace incrementMetric:kFPRAppCounterNameActivePrewarm byInt:1];
+  } else {
+    [self.activeTrace incrementMetric:kFPRAppCounterNameActivePrewarm byInt:0];
   }
 
-  if ([self isDoubleDispatchEnabled] &&
-      [doubleDispatchTime compare:applicationDidFinishLaunchTime] == NSOrderedAscending) {
-    return YES;
+  if ([doubleDispatchTime compare:applicationDidFinishLaunchTime] == NSOrderedAscending) {
```
jeremyjiang-dev, line 218: If devs disable double dispatch, doubleDispatchTime is always nil. Can you confirm`[doubleDispatchTime compare:applicationDidFinishLaunchTime] == NSOrderedAscending` always return false?

jeremyjiang-dev:
GitHub does not allow me to add comments for unmodified code.

I have a question about these [code](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m#L77-L85) related to plist. 

By default, there is no such plist flag, then what will be the value for `[[[NSBundle mainBundle] objectForInfoDictionaryKey:@"fireperf_disable_dd"] boolValue]`? Is it default to `NO`?

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: In that case this expression will return nil, so this if-block is skipped. Then line 221 will be checked, if doubleDispatchTime is null that block will also be skipped, so in the end no custom metric will be added to the trace if dev disables double dispatch

leotianlizhan:
> By default, there is no such plist flag, then what will be the value for `[[[NSBundle mainBundle] objectForInfoDictionaryKey:@"fireperf_disable_dd"] boolValue]`? Is it default to `NO`?

Yes. `[nil boolValue]` will evaluate to `nil`, and `nil` evaluated to a `BOOL` is `NO`.

jeremyjiang-dev (approved):

visumickey (approved):

PR #9288, opened 2022-02-02
Update functions quickstart ci.

granluo:


paulb777 (approved):
Thanks! I just ran into this.

LGTM when the quickstart passes CI

PR #9289, opened 2022-02-02
Fixes a bug in custom key and user ID logging

elenadoty:
Fixes a bug where setting a custom key's value to nil or passing nil as a user ID did not clear the value as expected.

Resolves #9241, #9212

### DIFF
--- a/Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h
+++ b/Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h
@@ -82,7 +82,7 @@ NS_SWIFT_NAME(Crashlytics)
  * @param value The value to be associated with the key
  * @param key A unique key
  */
-- (void)setCustomValue:(id)value forKey:(NSString *)key;
+- (void)setCustomValue:(nullable id)value forKey:(NSString *)key;
 
 /**
  * Sets custom keys and values to be associated with subsequent fatal and non-fatal reports.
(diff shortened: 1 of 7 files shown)


Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h
```diff
@@ -82,7 +82,7 @@ NS_SWIFT_NAME(Crashlytics)
  * @param value The value to be associated with the key
  * @param key A unique key
  */
-- (void)setCustomValue:(id)value forKey:(NSString *)key;
+- (void)setCustomValue:(nullable id)value forKey:(NSString *)key;
```
paulb777, line 85: This is potentially a breaking change for Swift users, so may need to wait until Firebase 9. cc: @ryanwilson

paulb777 (commented):
Is it feasible to add tests? At least for Objective C.  Also, maybe not necessary now, but this is a good example of what Swift API coverage tests would be useful for.

Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h
ryanwilson: Thanks for tagging! As far as I'm aware going from `nonnull` -> `nullable` isn't a breaking change as the source can stay the same, but the other way around can cause a compilation failure since the optional isn't unwrapped. 

Since this fixes a bug, I think it's reasonable to land before then. Happy to take the blame if it does break though :)

paulb777 (approved):

Crashlytics/Crashlytics/Public/FirebaseCrashlytics/FIRCrashlytics.h
elenadoty: In my local testing it didn't seem to be an issue, but I'll keep an eye out!

samedson (approved):
💯  ship it

PR #9292, opened 2022-02-04
Functions Interop + CoreInternal

ryanwilson:
- Created the CoreInternal library (needs API audit, has more APIs than needed)
- Introduced component system to Functions
- UNTESTED: register functions Swift library in FIRApp instead of at `+load`.

TODO: 
- [ ] Test the Core registration of Functions
- [ ] Create an interop SwiftPM package to get the Auth, Messaging, and AppCheck headers
- [ ] Migrate to use the above interop library

#no-changelog

### DIFF
--- a/FirebaseCoreInternal/Sources/dummy.m
+++ b/FirebaseCoreInternal/Sources/dummy.m
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,6 +12,4 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Foundation
-
-@objc protocol FunctionsProvider {}
+// Placeholder file, since SwiftPM requires a source file to create a module.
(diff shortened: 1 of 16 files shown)


Package.swift
```diff
@@ -237,6 +237,16 @@ let package = Package(
...
       ]
     ),
+
+    // Interal headers only for consuming from Swift.
+    .target(
+      name: "FirebaseCoreInternal",
+      path: "FirebaseCoreInternal/Sources",
+      publicHeadersPath: "Public",
+      cSettings: [
+//        .headerSearchPath("../../"),
```
paulb777, line 247: Delete cSettings. Only needed for libraries that need to find .h files.

FirebaseCoreInternal/Tests/FIRObjCAPITest.m
```diff
@@ -0,0 +1,37 @@
+//
```
paulb777, line 1: copyright

FirebaseCoreInternal/Sources/dummy.m
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2020 Google LLC
```
paulb777, line 1: 2022

paulb777 (commented):
Why are the files copied instead of moved?  Shouldn't it be sufficient to have them only in CoreInternal?

paulb777 (commented):
Incorporated into #9298 by moving the CoreInternal files instead of copying them

paulb777:
Closing now that #9298 has merged

PR #9297, opened 2022-02-04
Update Package.swift for CI.

granluo:


.github/workflows/functions.yml
```diff
@@ -45,6 +45,8 @@ jobs:
     # Don't run on private repo unless it is a PR.
     if: (github.repository == 'Firebase/firebase-ios-sdk' && github.event_name == 'schedule') || github.event_name == 'pull_request'
     runs-on: macos-11
+    env:
+      SWIFTCI_USE_LOCAL_DEPS: 1
```
paulb777, line 49: Let's namespace this with perhaps `FIREBASECI_` and its more use latest than local_deps, so how about `FIREBASECI_USE_LATEST_GOOGLEAPPMEASUREMENT`

granluo: The [sed](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/setup_spm_tests.sh#L24) could be removed after this is submitted. We can do it in another PR.

paulb777 (commented):
I suspect that spm.yml may also need the same `env` change

paulb777 (approved):
OK to merge this and follow up with the sed script and spm.yml separately

granluo:
> OK to merge this and follow up with the sed script and spm.yml separately

The change will not fail the current workflows with `sed`, so we can have separate PRs for that.

PR #9298, opened 2022-02-04
Set up Core Internal for SPM

paulb777:
Alternative to #9292 that moves the Core Internal files instead of copying them

#no-changelog


FirebaseCore/Internal/FirebaseCoreInternal.h
```diff
@@ -17,12 +17,13 @@
...
+#import "FIRAppInternal.h"
+#import "FIRComponent.h"
+#import "FIRComponentContainer.h"
+#import "FIRComponentType.h"
+#import "FIRCoreDiagnosticsConnector.h"
+#import "FIRDependency.h"
+#import "FIRHeartbeatInfo.h"
+#import "FIRLibrary.h"
+#import "FIRLogger.h"
+#import "FIROptionsInternal.h"
```
ryanwilson, line 29: Were the old repo-relative paths out of date then?

ryanwilson (approved):
Two small questions but LGTM, thanks!

FirebaseCore/Internal/FirebaseCoreInternal.h
paulb777: It's now a public header for module construction - so cannot use paths.

Package.swift
```diff
@@ -237,6 +237,15 @@ let package = Package(
         .headerSearchPath("../../.."),
       ]
     ),
+    // Interal headers only for consuming from Swift.
+    .target(
+      name: "FirebaseCoreInternal",
+      path: "FirebaseCore/Internal",
+      publicHeadersPath: ".",
+      cSettings: [
+        .headerSearchPath("../../"),
```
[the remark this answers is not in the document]
paulb777: It's needed here to find the Objective C main class references from the sources

PR #9301, opened 2022-02-07
firestore::mutation::Overlay class added, with unit tests

dconeybe:
Port https://github.com/firebase/firebase-android-sdk/blob/275b3eb9940f64c25ae614f75a4fd5f04eb09625/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Overlay.java

#no-changelog

### DIFF
--- a/Firestore/core/src/model/mutation/overlay.cc
+++ b/Firestore/core/src/model/mutation/overlay.cc
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 Google
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Firestore/core/src/model/mutation/overlay.h"
+#include "Firestore/core/src/util/hashing.h"
+#include "Firestore/core/src/util/to_string.h"
+
+#include "absl/strings/str_cat.h"
+
+namespace firebase {
+namespace firestore {
+namespace model {
+namespace mutation {
+
+bool operator==(const Overlay& lhs, const Overlay& rhs) {
+  return lhs.largest_batch_id_ == rhs.largest_batch_id_ &&
+         lhs.mutation_ == rhs.mutation_;
+}
+
+std::ostream& operator<<(std::ostream& os, const Overlay& result) {
+  return os << result.ToString();
+}
+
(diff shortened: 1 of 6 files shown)


Firestore/core/src/model/mutation/overlay.h
```diff
@@ -0,0 +1,90 @@
...
+
+  bool is_valid() const {
+    return mutation_.is_valid();
+  }
+
+  int largest_batch_id() const {
+    return largest_batch_id_;
+  }
+
+  Mutation mutation() const {
```
wu-hui, line 65: return by `const &`?

Firestore/core/src/model/mutation/overlay.h
```diff
@@ -0,0 +1,90 @@
...
+    return mutation_;
+  }
+
+  DocumentKey key() const {
```
wu-hui, line 69: return by `const &`

Firestore/core/src/model/mutation/overlay.h
```diff
@@ -0,0 +1,90 @@
...
+ */
+class Overlay {
+ public:
+  Overlay() = default;
+
+  Overlay(int largest_batch_id, const Mutation& mutation)
+      : largest_batch_id_(largest_batch_id), mutation_(mutation) {
+  }
+
+  bool is_valid() const {
```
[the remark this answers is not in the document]
wu-hui: Oh I see, this is for the tests. Can we do the friend class dance then?

FWIW, I am also ok with not testing `is_valid` at all. The constructors here are simply enough that the chance of them not working as expected is low..

Firestore/core/test/unit/model/mutation/overlay_test.cc
```diff
@@ -0,0 +1,257 @@
...
+#include "Firestore/core/test/unit/testutil/equals_tester.h"
+#include "Firestore/core/test/unit/testutil/testutil.h"
+#include "absl/strings/string_view.h"
+#include "gtest/gtest.h"
+
+namespace firebase {
+namespace firestore {
+namespace model {
+namespace mutation {
+namespace {
```
wu-hui, line 34: I have some general comments, no action needed however:

* There are a lot cool ways I did not know to write tests!
* A big chunk of the tests are actually testing C++ compiler doing the right thing IMO. I am not asking you to remove, it's just, just...I'd not have added these tests myself, and I think that is fine..

Firestore/core/src/model/mutation/overlay.cc
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iostream>
+
+#include "Firestore/core/src/model/mutation/overlay.h"
```
wu-hui, line 19: This should be the first include.

dconeybe:
> For Android, all mutation related classes are under `model/mutation` directory, do you plan to move the rest as part of you work, in a future PR?

Yes, I do plan to move those other files too.

PR #9311, opened 2022-02-08
Port serializer tests and fix exposed issues

paulb777:
- Ports Functions serializer tests to Swift 
- Implements UInt64 encoding
- Miscellaneous other fixes and cleanups

#no-changelog

### DIFF
--- a/.github/workflows/functions.yml
+++ b/.github/workflows/functions.yml
@@ -3,7 +3,7 @@ name: functions
 on:
   pull_request:
     paths:
-    - 'FirebaseFunctions/**'
+    - 'FirebaseFunctions**'
     - 'FirebaseSharedSwift**'
     - '.github/workflows/functions.yml'
     - 'Interop/Auth/Public/*.h'
--- a/FirebaseFunctionsSwift/Sources/FunctionsSerializer.swift
+++ b/FirebaseFunctionsSwift/Sources/FunctionsSerializer.swift
@@ -21,10 +21,10 @@ private enum Constants {
 }
 
 enum SerializerError: Error {
-  // TODO: Add paramters class name and value
+  // TODO: Add parameters class name and value
   case unsupportedType // (className: String, value: AnyObject)
   case unknownNumberType(charValue: String, number: NSNumber)
-  case unimplemented // TODO(wilsonryan): REMOVE
+  case invalidValueForType(value: String, requestedType: String)
 }
 
 class FUNSerializer: NSObject {
@@ -36,7 +36,7 @@ class FUNSerializer: NSObject {
     dateFormatter.timeZone = TimeZone(identifier: "UTC")
   }
 
-  // MARK: - Public APIs
+  // MARK: - Internal APIs
 
   internal func encode(_ object: Any) throws -> AnyObject {
     if object is NSNull {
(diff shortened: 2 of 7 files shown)


.github/workflows/functions.yml
```diff
@@ -3,7 +3,7 @@ name: functions
 on:
   pull_request:
     paths:
-    - 'FirebaseFunctions/**'
+    - 'FirebaseFunctions**'
```
ryanwilson, line 6: This will now trigger with podspec updates - I thought we didn't want that for some reason but I can't remember why (and I can't think of why we wouldn't want it triggering...). Just wanted to call it out to see if you remembered / knew.

FirebaseFunctionsSwift/Sources/FunctionsSerializer.swift
```diff
@@ -21,10 +21,10 @@ private enum Constants {
 }
 
 enum SerializerError: Error {
-  // TODO: Add paramters class name and value
```
ryanwilson, line 24: Can't wait for my laptop refresh with a reliable E key :P

FirebaseFunctionsSwift/Sources/FunctionsSerializer.swift
```diff
@@ -172,47 +166,24 @@ class FUNSerializer: NSObject {
...
-       if (error != NULL) {
-         *error = FUNInvalidNumberError(value, wrapped);
-       }
-       return nil;
-     }
-     return @(n);
-     */
+      let str = (value as NSString).utf8String
+      var endPtr: UnsafeMutablePointer<CChar>?
+      let returnValue = UInt64(strtoul(str, &endPtr, 10))
```
ryanwilson, line 181: Ah, thanks for porting this. I saw `stroul` but was looking for the `unsigned long long` variant, but now I realize that we're only expecting an unsigned long. Thanks!

FirebaseFunctionsSwift/Tests/Unit/SerializerTests.swift
```diff
@@ -63,186 +63,199 @@ class SerializerTests: XCTestCase {
...
+    let serializer = FUNSerializer()
+    XCTAssertEqual("hello", try serializer.encode("hello") as? String)
+  }
+
+  func testDecodeString() throws {
+    let serializer = FUNSerializer()
+    XCTAssertEqual("good-bye", try serializer.decode("good-bye") as? String)
+  }
+
+  // TODO: Should we add support for Array as well as NSArray?
```
ryanwilson, line 187: I think it implicitly works right now as it bridges to the ObjC implementation, but good question on whether or not it will do that now. Ideally yes if we can, but let's see what it looks like from the API surface (if you need to cast as NSArray or it implicitly does so)

ryanwilson (approved):

.github/workflows/functions.yml
paulb777: Yes, we should run tests if podspec's change. However, the new `pod spec` tests require consistent versions - so I'm changing to keep the 9.0.0 only pods consistent with the rest for now.

FirebaseFunctionsSwift/Tests/Unit/SerializerTests.swift
paulb777: Yeah, I think the API surface might be fine - since we didn't notice this with the integration tests.

.github/workflows/functions.yml
paulb777: Actually we can't since the `pod spec lint` is only for master. I disabled and created #9313 to track.

PR #9326, opened 2022-02-09
Update SwiftPM version for 8.12.1

ryanwilson:
This was missed during the process. Plan is to merge this then tag this commit as v8.12.1

#no-changelog

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@
 
 import PackageDescription
 
-let firebaseVersion = "8.12.0"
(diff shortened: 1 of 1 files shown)


paulb777:
It should also be tagged `8.12.1`

ryanwilson:
> It should also be tagged `8.12.1`

without the `v` as well?

ryanwilson:
Thanks, was confusing those instructions in 8 for the AppMeasurement instructions 👍

PR #9327, opened 2022-02-09
Merging the 8.12.1 release into master

ryanwilson:
Also updating the Firebase zip builder manifest which wasn't done.

#no-changelog


ryanwilson:
Looks like some of the quickstarts failed with networking error downloading carthage binaries - I think it's fine and I'll merge as is and keep an eye on it. Happy to revert if it causes issues for others though.

PR #9332, opened 2022-02-10
Combine tests ported to depend on Swift Functions

paulb777:
Port the Functions combine tests to depend on the FirebaseFunctionsSwift instead of FirebaseFunctions

cc: @peterfriese

### DIFF
--- a/FirebaseFunctions/Tests/CombineUnit/HTTPSCallableTests.swift
+++ b/FirebaseFunctions/Tests/CombineUnit/HTTPSCallableTests.swift
@@ -16,28 +16,43 @@ import Foundation
 
 import Combine
 import FirebaseCore
+@testable import FirebaseFunctionsSwift
 import FirebaseFunctionsCombineSwift
-@testable import FirebaseFunctionsTestingSupport
 import XCTest
 
-// hardcoded in FIRHTTPSCallable.m
-private let kFunctionsTimeout: TimeInterval = 70.0
+// hardcoded in HTTPSCallable.swift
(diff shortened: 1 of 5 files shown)


FirebaseFunctions/Tests/CombineUnit/HTTPSCallableTests.swift
```diff
@@ -16,56 +16,70 @@ import Foundation
...
 
-// hardcoded in FIRHTTPSCallable.m
-private let kFunctionsTimeout: TimeInterval = 70.0
+// hardcoded in HTTPSCallable.swift
+private let timeoutInterval: TimeInterval = 70.0
 private let expectationTimeout: TimeInterval = 2
 
-class MockFunctions: Functions {
-  var mockCallFunction: () throws -> HTTPSCallableResult?
+@objc class MockFunctions: Functions {
```
ryanwilson, line 31: I'm curious if this is needed anymore after doing the right imports - I'm fine to leave it but don't want to have it here unnecessarily!

ryanwilson (approved):

FirebaseFunctions/Tests/CombineUnit/HTTPSCallableTests.swift
paulb777: Good catch! Removed.

paulb777:
I'm going to merge here and sort out the combine CocoaPods issues after the next PR since they have to do with a disconnect between the FirebaseFunctions and FirebaseFunctionsSwift module names between the two package managers.

PR #9333, opened 2022-02-11
Remove Functions ObjC

paulb777:
Removes the Functions Objective C implementation.

All FirebaseFunctions references, schemes, modules etc come from the Swift implementation now.

#no-changelog


FirebaseFunctions.podspec
```diff
@@ -1,72 +1,57 @@
...
-    unit_tests.source_files = [
-      'FirebaseFunctions/Tests/Unit/*.[mh]',
-      'FirebaseFunctions/Tests/SwiftUnit/**/*',
-      'SharedTestUtilities/FIRAuthInteropFake*',
-      'SharedTestUtilities/FIRMessagingInteropFake*',
-      'SharedTestUtilities/AppCheckFake/*.[mh]',
-  ]
-  end
+  # Maybe TODO: The unit tests depend on setting up a podspec for SharedTestUtilities
+  # OTOH, we are running them with SwiftPM.
```
ryanwilson, line 47: I think considering there's at least an integration test running on CocoaPods, I'd be comfortable only having unit tests on SwiftPM. Any build configuration issues would show up in the integration tests. WDYT?

FirebaseFunctions/Tests/ObjCIntegration/FIRFunctions+Internal.h
```diff
@@ -16,7 +16,7 @@
 
 // This file started as a copy of FirebaseFunctions/Sources/FIRFunctions+Internal.h with only the
```
ryanwilson, line 17: Should this comment now be updated that the file doesn't exist?

Package.swift
```diff
@@ -717,35 +701,35 @@ let package = Package(
         "FirebaseSharedSwift",
         .product(name: "GTMSessionFetcherCore", package: "GTMSessionFetcher"),
       ],
-      path: "FirebaseFunctionsSwift/Sources"
+      path: "FirebaseFunctions/Sources"
     ),
     .testTarget(
       name: "FirebaseFunctionsSwiftUnit",
```
ryanwilson, line 707: Nit: should this and the ones below change from `FirebaseFunctionsSwift` to just `FirebaseFunctions`, or take the old naming of `Functions`? May be easier to read in the generated Xcode schemes.

ryanwilson (commented):
A couple small comments but overall LGTM, the SwiftPM generated schemes may need to be modified as well

FirebaseFunctions.podspec
paulb777: Agreed. Removed comment.

FirebaseFunctions/Tests/ObjCIntegration/FIRFunctions+Internal.h
paulb777: Yep. Removed, since this will now be the only variation.

Package.swift
paulb777: Updated here and the schemes.

paulb777:
Zip Test failures will be dealt with a subsequent PR. The quickstart failures are mostly due to the global symbols change.
There is also a failure from the missing ` static func functions()` API for ObjC.  I'll finish ObjC coverage tests in another PR.

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -39,7 +39,7 @@ internal enum FunctionsConstants {
 /**
  * `Functions` is the client for Cloud Functions for a Firebase project.
  */
-@objc(FIRFunctions) public class Functions: NSObject {
+@objc(FIRFunctions) open class Functions: NSObject {
```
ryanwilson, line 42: We should do the same for all of the classes, right?

ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
```diff
@@ -25,7 +25,11 @@ public let shared = Manifest(
   pods: [
     Pod("FirebaseSharedSwift"),
     Pod("FirebaseCoreDiagnostics", zip: true),
+    Pod("FirebaseCoreInternal", zip: true),
```
ryanwilson, line 28: These are only headers, right? Do they need to be in the zip?

ryanwilson (approved):

FirebaseFunctions/Sources/Functions.swift
paulb777: Done for all public classes now.

ReleaseTooling/Sources/FirebaseManifest/FirebaseManifest.swift
paulb777: Good point. We need the dependencies, but not the binaries. The change in CocoaPodUtils.swift is probably closer.

PR #9342, opened 2022-02-14
Complete Swift API coverage tests for Messaging.

ryanwilson:
This does have two warnings, but keeping them enables us to get an
actual break when the enum changes.

[skip ci] temporarily


ryanwilson:
np! Test failure is the CocoaPods podspec version being different from the test, it passes locally so I'll merge.

PR #9345, opened 2022-02-14
firestore::local::LevelDbDocumentOverlayCache added.

dconeybe:
Create `LevelDbDocumentOverlayCache`, the analogue of [`SqliteDocumentOverlayCache.java`](https://github.com/firebase/firebase-android-sdk/blob/ee6e6564f1689ccffcb6e193e54a46847cec81a1/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java) from the Android SDK. The unit tests for `LevelDbDocumentOverlayCache` re-use the ones for`MemoryDocumentOverlayCache` (added in #9330), since those tests were written in a manner that enables them to be parameterized.

Future PRs will be needed to improve the performance of this new class. The performance of `GetOverlay()` is O(log(n)), which is acceptable; however, the performance of `RemoveOverlaysForBatchId()`, `GetOverlays(const ResourcePath&, int)`, and `GetOverlays(const std::string&, int, std::size_t)` are θ(n), where _n_ is the number of overlays in the database for the user, which is unacceptable. Using manual indexing, their performances can be improved to O(m(log(m)) for `RemoveOverlaysForBatchId()` and O(m) for the two `GetOverlays()` methods, where _m_ is the number of overlays that are affected/returned by the methods.

#no-changelog


Firestore/core/src/local/leveldb_key.h
```diff
@@ -874,6 +879,51 @@ class LevelDbIndexEntryKey {
...
+  }
+
+  /** The path to the document, as encoded in the key. */
+  const model::DocumentKey& document_key() const {
+    return document_key_;
+  }
+
+ private:
+  std::string user_id_;
+  model::DocumentKey document_key_;
```
wu-hui, line 924: Just want to point this out before I review further, I think if you add `largest_batch_id` as the third segment to the key, you can avoid the new proto and remove the serialization code as well.

This PR will be much smaller this way.

dconeybe: Hmmm okay I'll give that a try.

dconeybe: Done. PTAL.

dconeybe: Actually... I found a bug. I'm working on a fix.

dconeybe: Fixed by cf04e6740bd371738cd62bed71498838a259e3db

wu-hui (approved):
Thanks for deleting the proto!

I left some comments, around a method you are about to rewrite..for feel free to ignore.

wu-hui:
LGTM

PR #9361, opened 2022-02-17
[Heartbeat Logging] Style branch and add two TODOs

ncooke3:
- Run style script on the branch
- Add two TODOs

**Context**
The [`platform-logging-main`](https://github.com/firebase/firebase-ios-sdk/tree/platform-logging-main) is new and was created off of master and then the 12 (approved) heartbeat-related commits from [`heartbeats-main`](https://github.com/firebase/firebase-ios-sdk/tree/heartbeats-main) were cherry-picked into it. 

The end goal was to have a feature branch with the heartbeat work that was cleanly rebased onto master (rather than merged like in `heartbeats-main`). See clean comparison now: https://github.com/firebase/firebase-ios-sdk/compare/platform-logging-main

### DIFF
--- a/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
+++ b/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
@@ -296,7 +296,7 @@ private class StorageFake: Storage {
     if let onWrite = onWrite {
       return try onWrite(data)
     } else {
-      self.fakeFile = data
+      fakeFile = data
     }
   }
 }
--- a/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
+++ b/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
@@ -285,8 +285,8 @@ private class StorageFake: Storage {
   func read() throws -> Data {
     if let onRead = onRead {
       return try onRead()
-    } else if let fakeFile = fakeFile {
-      return fakeFile
+    } else if let data = fakeFile {
+      return data
     } else {
       throw StorageError.readError
     }
--- a/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
+++ b/HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
@@ -278,15 +278,15 @@ class HeartbeatStorageTests: XCTestCase {
 }
 
 private class StorageFake: Storage {
-  var data: Data?
+  var fakeFile: Data?
   var onRead: (() throws -> Data)?
   var onWrite: ((Data?) throws -> Void)?
 
   func read() throws -> Data {
     if let onRead = onRead {
       return try onRead()
-    } else if let data = data {
-      return data
+    } else if let fakeFile = fakeFile {
+      return fakeFile
     } else {
       throw StorageError.readError
     }
@@ -296,7 +296,7 @@ private class StorageFake: Storage {
     if let onWrite = onWrite {
       return try onWrite(data)
     } else {
-      self.data = data
+      self.fakeFile = data
     }
   }
 }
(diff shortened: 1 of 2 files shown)


HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
```diff
@@ -285,8 +285,10 @@ private class StorageFake: Storage {
   func read() throws -> Data {
     if let onRead = onRead {
       return try onRead()
+    } else if let data = data {
+      return data
```
paulb777, line 289: Separate from this PR: 

- how can `data` ever not be nil?
- Why doesn't `onRead` return data?
- Why is data at class scope?

paulb777 (approved):

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: > how can data ever not be nil?
Data can be given a non-nil value by using the StorageFake's `write()` API.

Consider:
```swift
let storageFake = StorageFake() // `data` == `nil`
let myData: Data = // ...
storageFake.write(myData) // `data` == myData
```

Data may not be nil when non-nil data is  nilwrite()

> Why doesn't onRead return data?
`onRead` _does_ return data. It is an optional, 
throwing closure that returns data: `(() throws -> Data)?`

> Why is data at class scope?
Hmm it is not, `data` is an instance var.

paulb777: It's not a class var, but it's an instance var at class scope versus locally scoped to a `func`.

Why does `write` need to store data in `self.data` in addition to passing it to `onWrite`?
Why does `read` ignore the return value from `onRead` and instead return what was previously written by `write`?

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
```diff
@@ -285,8 +285,10 @@ private class StorageFake: Storage {
...
     } else {
-      return try data ?? { throw StorageError.readError }()
```
ncooke3, line 289: This was reworked to avoid an error where running `swiftformat` on the file broke the build.

I filed https://github.com/nicklockwood/SwiftFormat/issues/1142

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: Done in [515633d](https://github.com/firebase/firebase-ios-sdk/pull/9361/commits/515633dbab7dadaa76962f64d66d803fee1cc61f)

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
```diff
@@ -278,23 +278,25 @@ class HeartbeatStorageTests: XCTestCase {
...
 private class StorageFake: Storage {
-  var data: Data?
+  var fakeFile: Data?
   var onRead: (() throws -> Data)?
   var onWrite: ((Data?) throws -> Void)?
 
   func read() throws -> Data {
     if let onRead = onRead {
       return try onRead()
+    } else if let fakeFile = fakeFile {
```
paulb777, line 288: nit: the lhs could still be data here

paulb777 (approved):

HeartbeatLogging/Tests/Unit/HeartbeatStorageTests.swift
ncooke3: Done in [5a48301](https://github.com/firebase/firebase-ios-sdk/pull/9361/commits/5a48301019ef3e64f992a2fea16ceb4a56682233)

PR #9362, opened 2022-02-18
Fireperf: fix ActivePrewarm

leotianlizhan:
b/220342323
# Bug
Reading `ActivePrewarm` in `appDidBecomeActiveNotification` always results in `null`. Presumably Apple deletes the `ActivePrewarm` env variable some time after `didFinishLaunching`, which is the latest callback I have confirmed where reading `ActivePrewarm` works. 

# Fix
Move reading `ActivePrewarm` to `+load()`.

Additionally switch to using `@avaiable(iOS 13, *)` for version and platform checks. This aligns Hub, and is similar to Firebase App Monitoring (cl/424623478).

# Todo
* Perhaps remove `didFinishLaunching` callback and the code for double dispatch as we learned it did not work well.

### DIFF
--- a/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
+++ b/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
@@ -165,7 +165,8 @@ NSString *const kFPRAppCounterNameDoubleDispatch = @"_fsddc";
  */
 - (BOOL)isPrewarmAvailable {
   BOOL canPrewarm = NO;
-  // Guarding for double dispatch which does not work below iOS 13
+  // Guarding for double dispatch which does not work below iOS 13, and 0.1% of app start also show
+  // signs of prewarming on iOS 14 go/paste/5533761933410304
   if (@available(iOS 13, *)) {
     canPrewarm = YES;
   }
--- a/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
+++ b/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
@@ -158,13 +158,13 @@ NSString *const kFPRAppCounterNameDoubleDispatch = @"_fsddc";
 }
 
 /**
- * Checks if prewarming is available for the platform on current device.
- * It is available when running iOS 15 and above.
+ * Checks if prewarming is available for the platform and OS version on the current device.
  *
  * @return true if the platform could prewarm apps on the current device
  */
 - (BOOL)isPrewarmAvailable {
   BOOL canPrewarm = NO;
+  // Guarding for double dispatch which is only accurate in iOS 13+
   if (@available(iOS 13, *)) {
     canPrewarm = YES;
   }
--- a/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
+++ b/FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
@@ -166,19 +166,13 @@ NSString *const kFPRAppCounterNameDoubleDispatch = @"_fsddc";
  * @return true if the platform could prewarm apps on the current device
  */
 - (BOOL)isPrewarmAvailable {
-  NSString *platform = [[GULAppEnvironmentUtil applePlatform] lowercaseString];
-  BOOL iOS = [platform isEqualToString:@"ios"];
-  BOOL catalyst = [platform isEqualToString:@"maccatalyst"];
-  BOOL tvos = [platform isEqualToString:@"tvos"];
-  if (![GULAppEnvironmentUtil isSimulator] && (iOS || catalyst || tvos)) {
-    NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
-    if ([systemVersion length] > 0) {
-      return [systemVersion compare:@"15" options:NSNumericSearch] != NSOrderedAscending;
-    } else {
-      [self.activeTrace incrementMetric:kFPRAppCounterNameActivePrewarm byInt:2];
-    }
+  NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
+  if ([systemVersion length] > 0) {
+    return [systemVersion compare:@"15" options:NSNumericSearch] != NSOrderedAscending;
+  } else {
+    [self.activeTrace incrementMetric:kFPRAppCounterNameActivePrewarm byInt:2];
+    return NO;
   }
-  return NO;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -161,7 +166,9 @@ - (FIRTrace *)activeTrace {
  * @return true if the platform could prewarm apps on the current device
  */
 + (BOOL)isPrewarmAvailable {
-  if (![[[GULAppEnvironmentUtil applePlatform] lowercaseString] isEqualToString:@"ios"]) {
+  NSString *platform = [[GULAppEnvironmentUtil applePlatform] lowercaseString];
+  if (![platform isEqualToString:@"ios"] && ![platform isEqualToString:@"maccatalyst"] &&
+      ![platform isEqualToString:@"tvos"]) {
```
jeremyjiang-dev, line 171: nit: Can you add a comment? The code is a bit hard to read. Can you do something like:
```
if (iOS && catalyst && tvos) {
  if (version > 15) {
   return YES;
  }
}
return No;
```

visumickey:
> Thanks for creating this PR! My concern is that "around 25% of iOS 15 SDK 8.12.0 app starts without any counters", I ran a query and did not find that there are ~25% non-iOS devices. I wonder if we are missing anything.

Good point Jeremy. But I don't think that this PR should be blocked because of that. That exploration is due to figure out the gap in numbers. Few reasons why we could not have found counters could be the following:

1. _fs is not guaranteed to reach on time (sometimes not even sent).
2. _fs could have been rate limited which _as would not have been missed.
3. _fs are not guaranteed to be complete. A crash could lose the _fs trace.

But still 25% is still high worth exploring.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -161,7 +166,9 @@ - (FIRTrace *)activeTrace {
  * @return true if the platform could prewarm apps on the current device
  */
 + (BOOL)isPrewarmAvailable {
-  if (![[[GULAppEnvironmentUtil applePlatform] lowercaseString] isEqualToString:@"ios"]) {
+  NSString *platform = [[GULAppEnvironmentUtil applePlatform] lowercaseString];
+  if (![platform isEqualToString:@"ios"] && ![platform isEqualToString:@"maccatalyst"] &&
```
visumickey, line 170: I assume there should be macros to do this check much easily. Can you see if there are any platform MACROS?

visumickey (approved):

visumickey (approved):

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
```diff
@@ -160,15 +163,12 @@ - (FIRTrace *)activeTrace {
...
-+ (BOOL)isPrewarmAvailable {
-  if (![[[GULAppEnvironmentUtil applePlatform] lowercaseString] isEqualToString:@"ios"]) {
-    return NO;
-  }
-  NSString *systemVersion = [GULAppEnvironmentUtil systemVersion];
-  if ([systemVersion length] == 0) {
-    return NO;
+- (BOOL)isPrewarmAvailable {
+  BOOL canPrewarm = NO;
+  if (@available(iOS 13, *)) {
```
visumickey, line 168: Is this iOS 15? Curious why we have this for iOS 13 and above?

visumickey: Is this because Prewarm was available in early versions? If so, can you add that to comments as to why we are looking for iOS13 and above here?

visumickey: Also - Have you been able to verify if the check fails for pre iOS13 versions?

leotianlizhan: > Also - Have you been able to verify if the check fails for pre iOS13 versions?

Yes, verified using simulators. 

> Curious why we have this for iOS 13 and above?

Two reasons:
1. Hub uses 13, not sure why, I have sent an email to ask
2. Since we are currently too restrictive in enabling prewarm-detection, I am trying to cover as much ground as possible by letting the prewarm-detection also run on iOS 13 and 14, in case Apple did something like % rollouts on iOS 13 and 14. 

This seemed like a good choice because it is working for hub, and 13 is not too old that env variable checking could break.

leotianlizhan: Michael said their iOS 13 guard was for double-dispatch which had issues on iOS 12. He has the opinion that we don't need to guard `ActivePrewarm` in the first place because if it's iOS < 15 then the environment variable will just never be there.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: decided to go with the native `@available(iOS ..., *)` function.

jeremyjiang-dev:
The changes LGTM! Shall we remove double dispatch related code since its detection is not accurate? @visumickey

visumickey:
> The changes LGTM! Shall we remove double dispatch related code since its detection is not accurate? @visumickey

I would leave it as is since it does not hurt. On the other hand, it might be helpful to enable Double dispatch for scenarios not caught by the environment variable. I would prefer to remove that once we have confirmed that the environment variable based approach is giving us all the positive signals.

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
visumickey: Sounds good. Can you make sure to reflect these learnings as comments above this check? This would be helpful for future references.

visumickey (approved):

jeremyjiang-dev:
Could you also update changelog?

jeremyjiang-dev:
> 

Sounds good!

leotianlizhan:
> Could you also update changelog?

I am hesitant to update changelog because we would have the exact same notes as before.

visumickey:
> > Could you also update changelog?
> 
> I am hesitant to update changelog because we would have the exact same notes as before.

A possible change log entry would be "Make the identification of pre-warming more predictable by moving the identification to the earliest phase of app start."

jeremyjiang-dev (approved):

FirebasePerformance/Sources/AppActivity/FPRAppActivityTracker.m
leotianlizhan: Added comment. Also, in Michael's doc that he shared with us, it says they have seen 0.1% app starts on iOS 14 showing signs of pre-warming, so possibly Apple did make percentage roll out on iOS 14 to test it. Perhaps that could be a source of why some of our _as doesn't have counters.

visumickey: Thanks for the updates Leo. One nit suggestion would be to rephrase the comment as this "Check for the OS to be 13.x and above since prewarming feature was enabled from that OS version". The intent is for the developer to understand the rationale of why that version check and not to go into deep details.

visumickey (approved):

PR #9363, opened 2022-02-18
[DO NOT MERGE] Register Combine component with new Swift component system

peterfriese:
This PR adds a Swift component for Firebase Combine, so that Combine will be included in the user agent.

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -836,7 +836,10 @@ static FIRApp *sDefaultApp;
   SEL componentsToRegisterSEL = @selector(componentsToRegister);
   // Dictionary of class names that conform to `FIRLibrary` and their user agents. These should only
   // be SDKs that are written in Swift but still visible to ObjC.
-  NSDictionary<NSString *, NSString *> *swiftLibs = @{@"FIRFunctionsComponent" : @"fire-fun"};
+  NSDictionary<NSString *, NSString *> *swiftLibs = @{
+    @"FIRFunctionsComponent" : @"fire-fun",
+    @"FIRCombineComponent" : @"fire-comb"
+  };
   for (NSString *className in swiftLibs.allKeys) {
     Class klass = NSClassFromString(className);
     if (klass && [klass respondsToSelector:componentsToRegisterSEL]) {
--- a/FirebaseCombineSwift/Sources/Core/CombineComponent.swift
+++ b/FirebaseCombineSwift/Sources/Core/CombineComponent.swift
@@ -1,0 +1,46 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import FirebaseCore
+import FirebaseCoreInternal
+
+@objc(FIRCombineProvider)
+protocol CombineProvider {
+}
+
+@objc(FIRCombineComponent) class CombineComponent: NSObject, Library, CombineProvider {
+  // MARK: - Private Variables
+
+  /// The app associated with all functions instances in this container.
+  private let app: FirebaseApp
+
+  // MARK: - Initializers
+
+  required init(app: FirebaseApp) {
+    self.app = app
+  }
+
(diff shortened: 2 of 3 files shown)


FirebaseCore/Sources/FIRApp.m
```diff
@@ -836,7 +836,8 @@ + (void)registerSwiftComponents {
   SEL componentsToRegisterSEL = @selector(componentsToRegister);
   // Dictionary of class names that conform to `FIRLibrary` and their user agents. These should only
   // be SDKs that are written in Swift but still visible to ObjC.
-  NSDictionary<NSString *, NSString *> *swiftLibs = @{@"FIRFunctionsComponent" : @"fire-fun"};
+  NSDictionary<NSString *, NSString *> *swiftLibs =
+      @{@"FIRFunctionsComponent" : @"fire-fun", @"FIRCombineComponent" : @"fire-comb"};
```
ryanwilson, line 840: Really for the Swift extensions we don't need the components behaviour, only the user agent. I think to reduce the amount of code the Swift SDKs have to write we can do another loop here with non-components code and use the `[FIRApp registerLibrary:withVersion:]` call instead. That way the Swift code only needs to be:

```
@objc(FIRCombineSDK) private class __CombineSDK: NSObject {}
```

Or you may need to add an init, but it can be empty or a fatalerror. The version can be pulled from the wider Firebase version.

peterfriese: I added another loop to iterate the non-component libraries and check if the Combine SDK "component" is present.

FirebaseCore/Sources/FIRApp.m
```diff
@@ -836,11 +836,20 @@ + (void)registerSwiftComponents {
...
+  for (NSString *className in swiftComponents.allKeys) {
     Class klass = NSClassFromString(className);
     if (klass && [klass respondsToSelector:componentsToRegisterSEL]) {
-      [FIRApp registerInternalLibrary:klass withName:swiftLibs[className]];
+      [FIRApp registerInternalLibrary:klass withName:swiftComponents[className]];
+    }
+  }
+
+  // Swift libraries that don't need component behaviour
+  NSDictionary<NSString *, NSString *> *swifLibraries = @{@"FIRCombineSDK" : @"fire-comb"};
```
paulb777, line 848: nit: typo should be `swiftLibraries`

paulb777 (approved):

paulb777:
@peterfriese I rebased this branch onto v9. Make sure you update from origin before continuing work

FirebaseCombineSwift/Sources/Core/CombineSDK.swift
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import FirebaseCore
+import FirebaseCoreInternal
+
+@objc(FIRCombineSDK) private class __CombineSDK: NSObject {}
```
paulb777, line 19: This file should not be created. Instead a comparable line to the bottom line should be added for each product's combine implementation.

FirebaseCore/Sources/FIRApp.m
```diff
@@ -836,11 +836,20 @@ + (void)registerSwiftComponents {
...
+  for (NSString *className in swiftComponents.allKeys) {
     Class klass = NSClassFromString(className);
     if (klass && [klass respondsToSelector:componentsToRegisterSEL]) {
-      [FIRApp registerInternalLibrary:klass withName:swiftLibs[className]];
+      [FIRApp registerInternalLibrary:klass withName:swiftComponents[className]];
+    }
+  }
+
+  // Swift libraries that don't need component behaviour
+  NSDictionary<NSString *, NSString *> *swiftLibraries = @{@"FIRCombineSDK" : @"fire-comb"};
```
paulb777, line 848: How about "fire-co-auth", etc

FirebaseCombineSwift/Sources/Core/CombineSDK.swift
peterfriese: Ok, makes sense.

FirebaseCore/Sources/FIRApp.m
paulb777: Agreed except I think we had the goal to keep these strings less than 9 characters. @ncooke3 Would you check with the metrics team?

paulb777: We discussed a bit and realized that the "fire-" doesn't provide any value, especially because "fire-auth" is already there. How about "comb-auth", etc?

peterfriese: SG!

(Reminds me of the 8.3 filename requirements the DOS filesystem imposed ;-) )

peterfriese:
Closing this in favour of #9562

PR #9381, opened 2022-02-23
Fix bugs in LevelDbDocumentOverlayCache where it failed to check for equality of the DocumentKey

dconeybe:
There were several latent bugs in `LevelDbDocumentOverlayCache` where it first searched the LevelDb database for an entry with a given key prefix, but then failed to ensure that the document key encoded in the LevelDb key was exactly equal to the requested key.

This resulted in at least two bugs:

1. `GetOverlay()` would incorrectly return the overlay for some document in a subcollection if the requested document did not have an overlay.
2. `SaveOverlays()` would delete the overlay for some document in a subcollection if the document with the given DocumentKey did not have an overlay.

#no-changelog

### DIFF
--- a/Firestore/core/test/unit/local/document_overlay_cache_test.cc
+++ b/Firestore/core/test/unit/local/document_overlay_cache_test.cc
@@ -152,6 +152,25 @@ TEST_P(DocumentOverlayCacheTest, CanReadSavedOverlays) {
   });
 }
 
+TEST_P(DocumentOverlayCacheTest, GetOverlayExactlyMatchesTheGivenDocumentKey) {
+  this->persistence_->Run("Test", [&] {
+    this->SaveOverlaysWithSetMutations(1, {"coll/doc1/sub/doc2"});
+
+    EXPECT_FALSE(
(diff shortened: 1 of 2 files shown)


Firestore/core/test/unit/local/document_overlay_cache_test.cc
```diff
@@ -152,6 +152,27 @@ TEST_P(DocumentOverlayCacheTest, CanReadSavedOverlays) {
   });
 }
 
+// Ensure that the fix in https://github.com/firebase/firebase-ios-sdk/pull/9381
```
wu-hui, line 155: I do not think this comment helps here. It is more helpful to explain what this test does, if you think the name is not obvious.

wu-hui (approved):

Firestore/core/test/unit/local/document_overlay_cache_test.cc
dconeybe: Ok. I'll just remove the comment (here and below).

PR #9386, opened 2022-02-23
Improve performance of LevelDbDocumentOverlayCache::RemoveOverlaysForBatchId()

dconeybe:
The `LevelDbDocumentOverlayCache` class was added in In #9345, but with recognized performance issues. This PR improves the performance of `RemoveOverlaysForBatchId()` from θ(N), where `N` is the total size of the table, to θ(m(log(m)), where `m` is the number of rows removed. This performance was achieved by manually maintaining an "index", similar to how sqlite would do, so that the overlays with a given largest_batch_id could be directly looked up.

This improvement can be seen in the implementation of `RemoveOverlaysForBatchId()`. It used to call `ForEachOverlay()`, which iterated over the entire table. Now it calls `ForEachKeyWithLargestBatchId()`, which uses the index to only iterate over the table rows that actually match.

The index works by adding entries into a new table: `document_overlays_largest_batch_id_index`. The keys of this new table naturally order by largest_batch_id and contain all of the information needed to reconstruct the key in the main table. A new index entry is added every time that a new overlay is added to the database, and deleted when one is deleted.

A private helper method for unit testing, `GetOverlayCount()`, was added to the `DocumentOverlayCache` class and `GetLargestBatchIdIndexEntryCount()` was added to `LevelDbDocumentOverlayCache`. These methods are used in white box testing to make sure that indexes are behaving properly. I had run into several bugs during development that these helper methods helped to debug and fix.

#no-changelog


Firestore/core/src/local/leveldb_document_overlay_cache.cc
```diff
@@ -220,6 +266,43 @@ void LevelDbDocumentOverlayCache::ForEachOverlay(
...
+          user_id_, decoded_key.largest_batch_id(), encoded_key));
+}
+
+void LevelDbDocumentOverlayCache::AddLargestBatchIdIndexEntry(
+    absl::string_view encoded_key,
+    const LevelDbDocumentOverlayKey& decoded_key) {
+  db_->current_transaction()->Put(
+      LevelDbDocumentOverlayLargestBatchIdIndexKey::Key(
+          user_id_, decoded_key.largest_batch_id(), encoded_key),
+      "");
```
wu-hui, line 283: I realize there are precedences that use empty values in "index" tables. But those are because the in those tables, the location is encoded as, for example a document key: https://osscs.corp.google.com/firebase-sdk/firebase-ios-sdk/+/master:Firestore/core/src/local/leveldb_key.h;l=635?q=leveldb_k&ss=firebase-sdk%2Ffirebase-ios-sdk

So the lookup process will be get the document key, then recalculate the primary key to read the actual document.

Here you could do the same, encode document key instead of a leveldb key, then do a recalculation; or slightly better performance-wise (but consumes more space): save the leveldb key as a value in the table, then no recalculation is required (it not being part of the key also helps with read/write performance).

Either way, the introduction of `LevelDbKey = 23,` in `leveldb_key` should not be needed.

wu-hui (changes requested):
Thanks for the improvement!

I am not sure if we need a new type of leveldb key though, see below.

Firestore/core/src/local/leveldb_document_overlay_cache.cc
dconeybe: Done. I chose to use an empty value for the index entries and reconstruct the LevelDb key for the actual document at runtime. Since there will ultimately be 3 indexes (after my next PR) the savings in disk space could be more meaningful. It also removes a possibility of data getting out of sync somehow.

I also ended up making a bunch of other "improvements" to the code, things that were obvious now that I haven't looked at it for a few days. So the code churn since your last review is quite large relative to the code churn required just to implement the indexing improvement you suggested.

Here is a summary of the extra "improvements" I made:
- Improved the consistency of some variable and argument naming: a `DocumentKey` is now always `document_key` and any variable named `key` is referring to a LevelDb key. Previously, "key", "encoded_key", and "decoded_key" were used without much consistency, detracting from readability.
- Stopped passing around the "encoded" LevelDb keys; only the _decoded_ LevelDb keys are now passed around, simplifying method signatures.
- Removed the private overload of `GetOverlay()`, since it was never used.
- Removed a few one-line private methods that were only used in one place, and just inlined them (e.g. `DeleteLargestBatchIdIndexEntry()`).
- Added a constructor to `LevelDbDocumentOverlayKey` that initializes its instance variables to the given values; this enabled avoiding a few encode/decode round trips.
- Added an `Encode()` method to `LevelDbDocumentOverlayKey` to factor out boilerplate logic from other classes.
- Added methods to `LevelDbDocumentOverlayLargestBatchIdIndexKey` to convert to and from `LevelDbDocumentOverlayKey`, also to factor out boilerplate.
- Added unit test coverage for the new methods.

Firestore/core/test/unit/local/leveldb_document_overlay_cache_test.cc
```diff
@@ -39,13 +91,76 @@ TEST(LevelDbDocumentOverlayCacheTest, TypeTraits) {
                 "is_move_assignable");
 }
 
-std::unique_ptr<Persistence> PersistenceFactory() {
-  return LevelDbPersistenceForTesting();
-}
+// Verify that the index entries for `LevelDbDocumentOverlayCache` are created
+// and destroyed as they should be. When I originally wrote the indexing code
```
wu-hui, line 95: Ultranit: people in the future won't know who "i" was, so just state the purpose or even delete?

wu-hui (approved):
Nice PR!

Firestore/core/test/unit/local/leveldb_document_overlay_cache_test.cc
dconeybe: Done. I just removed the comment.

Issue #9341, opened 2022-02-14
Breaking change: Remove unused typedefs from Swift API surface.


ryanwilson:
Parse all public headers to remove the unused Swift typedefs that are surfaced from Objective-C.  These were handy in Objective-C due to the complex block syntax, but unnecessary in Swift as it makes it harder to understand what the underlying arguments are during autocomplete.

Example:

https://github.com/firebase/firebase-ios-sdk/blob/4a7a99e30162593774865badc6f22d41193e71d0/FirebaseMessaging/Sources/Public/FirebaseMessaging/FIRMessaging.h#L32-L34

should become

```
typedef void (^FIRMessagingFCMTokenFetchCompletion)(NSString *_Nullable FCMToken,
                                                    NSError *_Nullable error)
    NS_SWIFT_UNAVAILABLE;
```

This will prevent unused types from polluting the namespace.

ryanwilson:
Fixed in #9387

PR #9387, opened 2022-02-23
Mark ObjC typedefs as unavailable in Swift.

ryanwilson:
ATTN: BREAKING CHANGE FOR SWIFT USERS

This is part of the larger Swift API improvements effort. These reduce
visibility in autocomplete, and we've mostly migrated away from them
already. This hides the unused type from the API surface to avoid
cluttering the list of types available.


ryanwilson:
Fixes #9341. Need to actually fix the CombineSwift unit tests.

paulb777:
The API review is approved. This can be rebased to v9 and merged.

paulb777 (commented):
Need a discussion about Storage and Remote Config if we go with Swift-libraries for them.

Also, the conflict needs to be resolved.

paulb777 (approved):
We cam sort out the Storage conflicts when they come up.

ryanwilson:
Pretty sure the Auth lib lint is a flake, will merge and address if it's failing in other PRs.

PR #9394, opened 2022-02-25
[#2933] Removed `[FIRLibrary configureWithApp:]`

### LINKED ISSUE #2933: Remove `[FIRLibrary configureWithApp:]` calls.
This structure made sense before when a notification was fired, but doesn't make sense in the current world. If SDKs need to know the instant an app is configured, they should use [`FIRInstantiationTimingAlwaysEager` or `FIRInstantiationTimingEagerInDefaultApp`](https://github.com/firebase/firebase-ios-sdk/blob/9ec1da028e92cdd3e32c07ba88e8ca6fa30b420f/Firebase/Core/Private/FIRComponent.h#L39) to be configured. This will prevent any singleton use in the `configureWithApp:` call and clean up testing as well. Note: all libraries (internal and external) will need to remove this call in order for it to be removed from `FIRLibrary`. 

The definition is found here:
https://github.com/firebase/firebase-ios-sdk/blob/9ec1da028e92cdd3e32c07ba88e8ca6fa30b420f/Firebase/Core/Private/FIRLibrary.h#L38

yakovmanshin:
[Closes #2933]

* Removed all calls to `[FIRLibrary configureWithApp:]`
* Removed the `configureWithApp:` method from the `FIRLibrary` protocol

⚠️ Breaking Changes: Based on the [issue](https://github.com/firebase/firebase-ios-sdk/issues/2933), this PR should land in Firebase v9.

### DIFF
--- a/FirebaseCore/Internal/FIRLibrary.h
+++ b/FirebaseCore/Internal/FIRLibrary.h
@@ -32,11 +32,6 @@ NS_SWIFT_NAME(Library)
 /// FirebaseApp and participate in dependency resolution and injection.
 + (NSArray<FIRComponent *> *)componentsToRegister;
 
-@optional
-/// Implement this method if the library needs notifications for lifecycle events. This method is
-/// called when the developer calls `FirebaseApp.configure()`.
-+ (void)configureWithApp:(FIRApp *)app;
-
 @end
 
 NS_ASSUME_NONNULL_END
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -103,12 +103,6 @@ NSString *const kFirebaseCoreDefaultsSuiteName = @"com.firebase.core";
  */
 static NSString *const kPlistURL = @"https://console.firebase.google.com/";
 
-/**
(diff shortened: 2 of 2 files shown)


paulb777:
Still need to confirm with external clients. Also there is a usage in FIAM at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInAppMessaging/Sources/FIRInAppMessaging.m#L71

paulb777:
Still need to confirm with external clients. Also there is a usage in FIAM at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseInAppMessaging/Sources/FIRInAppMessaging.m#L71

ryanwilson:
Hey @yakovmanshin - thanks for the PR. After some internal discussions, we're going to leave this as is for now as to not disrupt downstream dependencies. It's not worth the breakage at the moment and we don't know who else would be impacted, plus it's not performance intensive or a ton of code so we're okay with it sitting there for another little while (potentially indefinitely). I'll close this for now but will re-open if we decide to move forward with removing it. Thanks again and sorry about that!

PR #9401, opened 2022-03-01
Improve performance of LevelDbDocumentOverlayCache::GetOverlays(ResourcePath, ...)

dconeybe:
The `LevelDbDocumentOverlayCache` class was added in In #9345, but with recognized performance issues. This PR improves the performance of `GetOverlays(ResourcePath, ...)` from θ(N), where `N` is the total size of the table, to θ(m(log(m)), where `m` is the number of overlays returned. 

This index implementation is directly based on the implementation for the `RemoveOverlaysForBatchId()` method, as implemented in #9386. See that PR for a deeper explanation of the implementation details.

This PR introduces the class `LevelDbDocumentOverlayIndexKey`, which is used as the base class for all index keys in the `document_overlays` table. As of this PR, it has two subclasses: `LevelDbDocumentOverlayLargestBatchIdIndexKey` (modified in this PR to subclass it) and `LevelDbDocumentOverlayCollectionIndexKey` (added in this PR).

#no-changelog

### DIFF
--- a/Firestore/core/src/local/leveldb_document_overlay_cache.cc
+++ b/Firestore/core/src/local/leveldb_document_overlay_cache.cc
@@ -60,7 +60,7 @@ absl::optional<Overlay> LevelDbDocumentOverlayCache::GetOverlay(
   auto it = db_->current_transaction()->NewIterator();
   it->Seek(key_prefix);
 
-  if (!(it->Valid() && absl::StartsWith(it->key(), key_prefix))) {
+  if (!it->Valid() || !absl::StartsWith(it->key(), key_prefix)) {
     return absl::nullopt;
   }
 
@@ -202,7 +202,7 @@ void LevelDbDocumentOverlayCache::DeleteOverlay(
   auto it = db_->current_transaction()->NewIterator();
   it->Seek(key_prefix);
 
-  if (!(it->Valid() && absl::StartsWith(it->key(), key_prefix))) {
+  if (!it->Valid() || !absl::StartsWith(it->key(), key_prefix)) {
     return;
   }
 
@@ -278,7 +278,7 @@ absl::optional<Overlay> LevelDbDocumentOverlayCache::GetOverlay(
   auto it = db_->current_transaction()->NewIterator();
   const std::string encoded_key = key.Encode();
   it->Seek(encoded_key);
-  if (!(it->Valid() && it->key() == encoded_key)) {
+  if (!it->Valid() || it->key() != encoded_key) {
     return absl::nullopt;
   }
   return ParseOverlay(key, it->value());
--- a/Firestore/core/src/local/leveldb_document_overlay_cache.cc
+++ b/Firestore/core/src/local/leveldb_document_overlay_cache.cc
@@ -88,28 +88,13 @@ void LevelDbDocumentOverlayCache::RemoveOverlaysForBatchId(int batch_id) {
 DocumentOverlayCache::OverlayByDocumentKeyMap
 LevelDbDocumentOverlayCache::GetOverlays(const ResourcePath& collection,
                                          int since_batch_id) const {
-  // TODO(dconeybe) Implement an index so that this query can be performed
-  // without requiring a full table scan.
-
   OverlayByDocumentKeyMap result;
-
-  const size_t immediate_children_path_length{collection.size() + 1};
-
-  ForEachOverlay([&](LevelDbDocumentOverlayKey&& key,
-                     absl::string_view encoded_mutation) {
-    if (!collection.IsPrefixOf(key.document_key().path())) {
-      return;
-    }
-    // Documents from sub-collections
-    if (key.document_key().path().size() != immediate_children_path_length) {
-      return;
-    }
-
-    if (key.largest_batch_id() > since_batch_id) {
-      result[key.document_key()] = ParseOverlay(key, encoded_mutation);
-    }
-  });
-
+  ForEachKeyInCollection(
+      collection, since_batch_id, [&](LevelDbDocumentOverlayKey&& key) {
+        absl::optional<Overlay> overlay = GetOverlay(key);
+        HARD_ASSERT(overlay.has_value());
+        result[std::move(key).document_key()] = std::move(overlay).value();
+      });
   return result;
 }
 
@@ -167,6 +152,11 @@ int LevelDbDocumentOverlayCache::GetLargestBatchIdIndexEntryCount() const {
       LevelDbDocumentOverlayLargestBatchIdIndexKey::KeyPrefix(user_id_));
 }
 
+int LevelDbDocumentOverlayCache::GetCollectionIndexEntryCount() const {
+  return CountEntriesWithKeyPrefix(
+      LevelDbDocumentOverlayCollectionIndexKey::KeyPrefix(user_id_));
+}
+
 int LevelDbDocumentOverlayCache::CountEntriesWithKeyPrefix(
     const std::string& key_prefix) const {
   int count = 0;
@@ -202,6 +192,7 @@ void LevelDbDocumentOverlayCache::SaveOverlay(int largest_batch_id,
   auto* transaction = db_->current_transaction();
   transaction->Put(key.Encode(), serializer_->EncodeMutation(mutation));
   transaction->Put(LevelDbDocumentOverlayLargestBatchIdIndexKey::Key(key), "");
+  transaction->Put(LevelDbDocumentOverlayCollectionIndexKey::Key(key), "");
 }
 
 void LevelDbDocumentOverlayCache::DeleteOverlay(
@@ -227,6 +218,7 @@ void LevelDbDocumentOverlayCache::DeleteOverlay(
   auto* transaction = db_->current_transaction();
   transaction->Delete(key.Encode());
   transaction->Delete(LevelDbDocumentOverlayLargestBatchIdIndexKey::Key(key));
+  transaction->Delete(LevelDbDocumentOverlayCollectionIndexKey::Key(key));
 }
 
 void LevelDbDocumentOverlayCache::ForEachOverlay(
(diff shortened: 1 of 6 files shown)


Firestore/core/src/local/leveldb_document_overlay_cache.cc
```diff
@@ -258,6 +250,40 @@ void LevelDbDocumentOverlayCache::ForEachKeyWithLargestBatchId(
...
+  const std::string index_key_prefix =
+      LevelDbDocumentOverlayCollectionIndexKey::KeyPrefix(user_id_, collection);
+
+  auto it = db_->current_transaction()->NewIterator();
+  for (it->Seek(index_start_key);
+       it->Valid() && absl::StartsWith(it->key(), index_key_prefix);
+       it->Next()) {
+    LevelDbDocumentOverlayCollectionIndexKey key;
+    HARD_ASSERT(key.Decode(it->key()));
+    if (key.collection() != collection) {
```
schmidt-sebastian, line 269: Will this work with nested subcollections that share the same prefix? We probably need to "continue" here like in https://osscs.corp.google.com/firebase-sdk/firebase-ios-sdk/+/master:Firestore/core/src/local/leveldb_mutation_queue.cc;l=323

We already validate that this loop does not go "out of bounds" with the check in line 265.

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -1346,10 +1348,75 @@ bool LevelDbDocumentOverlayLargestBatchIdIndexKey::Decode(
...
+  auto user_id = reader.ReadUserId();
+  auto largest_batch_id = reader.ReadBatchId();
+  auto document_key = reader.ReadDocumentKey();
+  reader.ReadTerminator();
+  Reset(std::move(user_id), std::move(largest_batch_id),
+        std::move(document_key));
+  return reader.ok();
+}
+
+std::string LevelDbDocumentOverlayCollectionIndexKey::KeyPrefix() {
```
schmidt-sebastian, line 1360: This is not used outside of tests, but we could keep it for consistency.

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -1346,10 +1348,75 @@ bool LevelDbDocumentOverlayLargestBatchIdIndexKey::Decode(
...
+    absl::string_view user_id,
+    const model::ResourcePath& collection,
+    model::BatchId largest_batch_id,
+    const model::DocumentKey& document_key) {
+  Writer writer;
+  writer.WriteTableName(kDocumentOverlaysCollectionIndexTable);
+  writer.WriteUserId(user_id);
+  writer.WriteResourcePath(collection);
+  writer.WriteBatchId(largest_batch_id);
+  writer.WriteResourcePath(document_key.path());
```
schmidt-sebastian, line 1405: Can we just write the document ID instead of the full path?

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -1346,10 +1348,75 @@ bool LevelDbDocumentOverlayLargestBatchIdIndexKey::Decode(
...
+
+bool LevelDbDocumentOverlayCollectionIndexKey::Decode(absl::string_view key) {
+  Reader reader{key};
+  reader.ReadTableNameMatching(kDocumentOverlaysCollectionIndexTable);
+  auto user_id = reader.ReadUserId();
+  collection_ = reader.ReadResourcePath();
+  auto largest_batch_id = reader.ReadBatchId();
+  auto document_key = reader.ReadDocumentKey();
   reader.ReadTerminator();
+  Reset(std::move(user_id), std::move(largest_batch_id),
```
schmidt-sebastian, line 1418: No need to move largest_batch_id.

Firestore/core/src/local/leveldb_key.h
```diff
@@ -127,6 +127,13 @@ namespace local {
 //   - user_id: string
 //   - largest_batch_id: model::BatchId
 //   - document_key: ResourcePath
+//
+// document_overlays_collection_index:
+//   - table_name: "document_overlays_collection_index"
+//   - user_id: string
+//   - collection: ResourcePath
+//   - largest_batch_id: model::BatchId
+//   - document_key: ResourcePath
```
schmidt-sebastian, line 136: As stated above, maybe just use the document_id.

Firestore/core/src/local/leveldb_key.h
```diff
@@ -1026,46 +1094,73 @@ class LevelDbDocumentOverlayLargestBatchIdIndexKey {
...
+  /** The collection_, as encoded in the key. */
+  const model::ResourcePath& collection() const {
+    return collection_;
   }
 
  private:
-  std::string user_id_;
-  model::BatchId largest_batch_id_ = -1;
-  model::DocumentKey document_key_;
+  model::ResourcePath collection_;
```
schmidt-sebastian, line 1163: FWIW, this should be part of the document_key_ in the superclass.

Firestore/core/src/local/leveldb_document_overlay_cache.cc
dconeybe: I believe it will work in that case because in the encoding of the key the collection is followed by the largest_batch_id. What would be a test case that I could write? I'm not sure I 100% understand the scenario you're describing.

I also want to avoid, at all costs, using `continue` instead of `break` because if using `continue` then the performance of this method degrades to a full table scan in the worst case.

Firestore/core/src/local/leveldb_key.cc
dconeybe: ... and even the usages of `KeyPrefix()` in tests was only to verify its behavior. I've removed these unnecessary `KeyPrefix()` methods.

Firestore/core/src/local/leveldb_key.cc
dconeybe: Done. I had moved it just for consistency since moving it is harmless. But, it can also be confusing when something is being moved unnecessarily. I've changed it to pass by value.

dconeybe: Done. I had used `std::move()` on `largest_batch_id` just for consistency with the other method arguments, since moving it is harmless. But, it can also be confusing when something is being moved unnecessarily. I've changed it to pass by value.

Firestore/core/src/local/leveldb_key.cc
dconeybe: Done. The code now writes the document ID and constructs the document key from it and the collection path.

Firestore/core/src/local/leveldb_key.h
dconeybe: Done. This comment was updated to replace `document_key: ResourcePath` with `document_id: string`.

Firestore/core/src/local/leveldb_key.h
dconeybe: Done. I've removed the `collection_` member variable.

Firestore/core/src/local/leveldb_document_overlay_cache.cc
```diff
@@ -258,6 +250,40 @@ void LevelDbDocumentOverlayCache::ForEachKeyWithLargestBatchId(
...
+    callback(std::move(key).ToLevelDbDocumentOverlayKey());
+  }
+}
+
+absl::optional<Overlay> LevelDbDocumentOverlayCache::GetOverlay(
+    const LevelDbDocumentOverlayKey& key) const {
+  auto it = db_->current_transaction()->NewIterator();
+  const std::string encoded_key = key.Encode();
+  it->Seek(encoded_key);
+  if (!(it->Valid() && it->key() == encoded_key)) {
```
schmidt-sebastian, line 281: I think this would be easier to read if this was `!it->valid() || it->key != encoded_key`. I just spent a couple seconds wondering if the statement was wrong. You can keep as is if you prefer as this is might just be my brains slowness this morning.

Firestore/core/src/local/leveldb_document_overlay_cache.cc
schmidt-sebastian: After discussing this offline I think it should work. The collection path is written in segments, and the segments are terminated by a character that cannot be part of a segment. So the key prefix you are using will always end in a terminator for the last path segment, then the component label for the batch ID and then the batch ID. The only cause of conflict would be if the the first byte after the terminator were the batch ID label, but that is not possible since it would have to be the component label for a path component, which will never match the batch ID label.

That being said, I still don't think you need this if check. Your loop should terminate based on `absl::StartsWith(it->key(), index_key_prefix)`.

Firestore/core/src/local/leveldb_key.cc
schmidt-sebastian: Looks like this change was not pushed.

schmidt-sebastian (approved):

Firestore/core/src/local/leveldb_document_overlay_cache.cc
dconeybe: The `if` check is still required because `index_key_prefix` does _not_ include the largest_batch_id and therefore it is also a prefix of the keys for overlays subcollections.

If I comment out that "if" check then the `GetAllOverlaysForCollection` test fails at line 255 because the call to `GetOverlays()` returns the overlays for the requested collection AND its subcollections (it should NOT return overlays for the subcollections, only the collection itself).

Firestore/core/src/local/leveldb_document_overlay_cache.cc
dconeybe: Done. If you find it confusing, others probably do too. I also made a similar change to two other "if" statements in this file.

Firestore/core/src/local/leveldb_key.cc
dconeybe: Oops. Fixed it.

Firestore/core/src/local/leveldb_document_overlay_cache.cc
schmidt-sebastian: Okay, let's keep it as is. The alternative of adding an additional marker or including the marker of the next component in the key prefix is not very appealing. Thanks for thinking this through with me.

schmidt-sebastian (approved):

PR #9411, opened 2022-03-03
Partial update to Swift dashboard

ryanwilson:
Reviewed some async/await APIs, removed ABTesting since it's not a public API.

Will improve text formatting when auditing the rest.

#no-changelog


ryanwilson:
@paulb777 also curious about what you think moving to a `?` for things we haven't investigated yet, and keep the `X` for something we know doesn't work yet or is missing.

paulb777:
Good idea to remove the ABT column. I like the idea of ? for uninvestigated/unknown

ryanwilson:
Also moved from `n/a` to `√` for async/await APIs for products that don't have async methods for simplicity, but happy to revert. I did an audit of all the libs quickly and didn't realize that was the convention and didn't want to go back to see if there were actually async/await APIs or just no async APIs.

PR #9419, opened 2022-03-04
[Heartbeat Logging] Refinements to how `HeartbeatsPayload`s are converted into header strings

ncooke3:
This PR

### DIFF
--- a/HeartbeatLogging/Sources/HeartbeatsPayload.swift
+++ b/HeartbeatLogging/Sources/HeartbeatsPayload.swift
@@ -84,21 +84,20 @@ extension HeartbeatsPayload: HTTPHeaderRepresentable {
   /// Returns a processed payload string intended for use in a HTTP header.
   /// - Returns: A string value from the heartbeats payload.
   public func headerValue() -> String {
-    // TODO(ncooke3): Return a empty payload instead of empty string.
-    if userAgentPayloads.isEmpty {
-      return ""
-    }
-
     let encoder = JSONEncoder()
     encoder.dateEncodingStrategy = .formatted(Self.dateFormatter)
 
-    // TODO(ncooke3): Fall back to base64URL-only if gzipping fails.
+    guard let data = try? encoder.encode(self) else {
+      // Because `HeartbeatsPayload` conforms to Codal
+      return Self.emptyPayload.headerValue()
+    }
+
     do {
-      let data = try encoder.encode(self)
       let gzippedData = try data.zipped()
       return gzippedData.base64URLEncodedString()
     } catch {
-      return "" // Return empty string if processing failed.
+      // If gzipping fails, fall back to encoding with base64URL.
+      return data.base64URLEncodedString()
     }
   }
 }
--- a/HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
+++ b/HeartbeatLogging/Tests/Integration/HeartbeatLoggingIntegrationTests.swift
@@ -133,7 +133,8 @@ class HeartbeatLoggingIntegrationTests: XCTestCase {
 
     // Then
     let nonEmptyPayloads = payloads.filter { payload in
-      payload.headerValue().count > 0
+      // Filter out non-empty payloads.
+      !payload.userAgentPayloads.isEmpty
     }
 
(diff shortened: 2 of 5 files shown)


HeartbeatLogging/Sources/HeartbeatsPayload.swift
```diff
@@ -84,21 +84,20 @@ extension HeartbeatsPayload: HTTPHeaderRepresentable {
   /// Returns a processed payload string intended for use in a HTTP header.
   /// - Returns: A string value from the heartbeats payload.
   public func headerValue() -> String {
-    // TODO(ncooke3): Return a empty payload instead of empty string.
-    if userAgentPayloads.isEmpty {
-      return ""
-    }
-
```
ncooke3, line 91: 1. Previously, when the `headerValue` is requested from an empty payload, it would return an empty string. This empty string was not process correctly by backend. So instead, we treat an empty payload like any other payload and compress and zip it.

HeartbeatLogging/Sources/HeartbeatsPayload.swift
```diff
@@ -84,21 +84,20 @@ extension HeartbeatsPayload: HTTPHeaderRepresentable {
...
+    }
+
     do {
-      let data = try encoder.encode(self)
       let gzippedData = try data.zipped()
       return gzippedData.base64URLEncodedString()
     } catch {
-      return "" // Return empty string if processing failed.
+      // If gzipping fails, fall back to encoding with base64URL.
+      return data.base64URLEncodedString()
```
ncooke3, line 100: 2. Previously, if gzipping fails, we would return an empty string. In practice, I'm not sure when gzip would fail, but if it did, we could probably just fall back to encoding the data without zipping to prevent losing the heartbeat data. From the payload analysis doc I wrote, it would be unlikely that an unzipped payload would be too large.

HeartbeatLogging/Sources/HeartbeatsPayload.swift
paulb777: When is there an empty payload? And why does it need to be sent?

ncooke3: > When is there an empty payload?

Let's say the app is running and heartbeats are flushed to be added to a product request. Then, heartbeats are again flushed for another product request. This second flushing will result in an *empty payload*. Heartbeats are removed upon retrieval to avoid being retrieved (and potentially sent!) twice.

> And why does it need to be sent?

I guess technically it doesn't.

My rationale for this is that it simplifies things for the client. I believe this is what other platforms did as well.

Current usage by the client would be like:
```swift
let heartbeatsPayload = heartbeatController.flush()
let heartbeatsPayloadHeaderString = heartbeatsPayload.headerValue()
// add `heartbeatsPayloadHeaderString` to request
```

If flushing the heartbeat controller returned `nil` when there were no heartbeats to send, things would instead look like:
```swift
let heartbeatsPayload = heartbeatController.flush()
if let heartbeatsPayload = heartbeatsPayload {
  let heartbeatsPayloadHeaderString = heartbeatsPayload.headerValue()
  // add `heartbeatsPayloadHeaderString` to request
}
```
Keeping in mind the above code would be in objc for our product SDKs.

paulb777: Oh, I think the point I was missing is that the client is going to send something to the backend either way. And we want at least a placeholder for the heartbeat data.

HeartbeatLogging/Sources/HeartbeatsPayload.swift
```diff
@@ -84,21 +84,20 @@ extension HeartbeatsPayload: HTTPHeaderRepresentable {
...
```
paulb777, line 100: Will the backend log when it receives an ungzipped packet so we know if this is happening or not?

paulb777 (approved):

HeartbeatLogging/Sources/HeartbeatsPayload.swift
ncooke3: Looking at the source, it does.

ncooke3:
CI should be fixed now. Waiting for green CI.

paulb777 (approved):

chliangGoogle (approved):

PR #9433, opened 2022-03-09
local_serializer.cc: fix lossy implicit conversion compiler warning

### LINKED ISSUE #9430: Firestore 8.13.0 - Warnings in Firestore/core/src/local/local_serializer.cc

  * Xcode version: 13.2.1
  * Firebase SDK version: 8.13.0
  * Firebase Component: `Firestore`
  * Target platform(s): `iOS`

After updating pods to new version (8.13.0) in my iOS Project, i have received following warnings as seen in attached screenshot of following file.
File path: Firestore/core/src/local/local_serializer.cc

Firestore:core:src:local:local_serializer

dconeybe:
Fix the following warning in `local_serializer.cc`:

```
Implicit conversion loses integer precision: 'std::vector<firebase::firestore::model::Segment>::size_type' (aka 'unsigned long') to 'pb_size_t' (aka 'unsigned int')
```

The fix is to explicitly cast to suppress the compiler warning. There should NEVER be a case where the size of the vector is greater than 2**32, so, in practice, this will never be a lossy conversion.

Fixes #9430

#no-changelog

### DIFF
--- a/Firestore/core/src/local/local_serializer.cc
+++ b/Firestore/core/src/local/local_serializer.cc
@@ -499,9 +499,12 @@ LocalSerializer::EncodeFieldIndexSegments(
   result->query_scope =
       google_firestore_admin_v1_Index_QueryScope_COLLECTION_GROUP;
 
-  result->fields_count = segments.size();
+  // Explicitly cast the result of segments.size() to suppress compiler warnings
+  // about implicit conversion resulting in potential loss of precision.
+  const auto segments_size = static_cast<pb_size_t>(segments.size());
+  result->fields_count = segments_size;
   result->fields =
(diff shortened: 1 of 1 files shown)


paulb777 (commented):
Thanks for fixing. Please add a changelog describing that this fixes the warnings introduced in 8.13.0.

dconeybe:
> Thanks for fixing. Please add a changelog describing that this fixes the warnings introduced in 8.13.0.

Done.

ncooke3 (approved):
LGTM. No warnings in the [pod linting CI](https://github.com/firebase/firebase-ios-sdk/runs/5474962855?check_suite_focus=true#step:4:298) 👍

PR #9373, opened 2022-02-21
[Do not merge] Tripping all CI for `platform-logging-main`

ncooke3:
After creating `platform-logging-main` and fixing some things (style), I wanted to get a clean run of all CI.

PR #9436, opened 2022-03-09
[Heartbeat Logging] Update FirebaseCore dependents for heartbeat logging changes

ncooke3:
I tripped CI in https://github.com/firebase/firebase-ios-sdk/pull/9373 to see what work needed to be done to get this ready for merging into `v9`.  

Here is some context:
1. Add a `swift_version` to all pods that depend on `FirebaseCore`. The FirebaseCore's podspec was changed to include the swift heartbeat code. Because of the presence of Swift in the spec source, CocoaPods will consider FirebaseCore to be a Swift pod. Swift pods should specify a `swift_version` to avoid a linting warning. [Example failure](https://github.com/firebase/firebase-ios-sdk/runs/5279145928?check_suite_focus=true#step:4:369) from #9373. 

   I've set the `swift_version` to `5.3`.
3. Firestore CI will build all of FirebaseCore via CMake. Currently, the Swift heartbeat code has not been configured to build via CMake.  We deemed this to be OK since customers won't use CMake to build Firestore and other Firebase SDKs (i.e. Games SDK) that use Firestore will build via CocoaPods on Apple platforms (see [Games SDK's CMake config for Firestore](https://github.com/firebase/firebase-unity-sdk/blob/main/firestore/CMakeLists.txt#L105-L117), and [C++ SDK's CMake config for Firestore](https://github.com/firebase/firebase-cpp-sdk/blob/main/firestore/CMakeLists.txt#L301-L318)).

#no-changelog

### DIFF
--- a/Firebase.podspec
+++ b/Firebase.podspec
@@ -28,6 +28,8 @@ Simplify your app development, grow your user base, and monetize more effectivel
 
   s.cocoapods_version = '>= 1.4.0'
 
+  s.swift_version = '5.3'
+
   s.default_subspec = 'Core'
 
   s.subspec 'Core' do |ss|
--- a/FirebaseABTesting.podspec
+++ b/FirebaseABTesting.podspec
@@ -35,6 +35,8 @@ Firebase Cloud Messaging and Firebase Remote Config in your app.
   s.cocoapods_version = '>= 1.4.0'
   s.prefix_header_file = false
 
+  s.swift_version = '5.3'
+
   base_dir = "FirebaseABTesting/Sources/"
   s.source_files = [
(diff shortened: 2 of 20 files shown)


paulb777:
The changes seem fine to me, but I'm not following why changing more files helps with a merge or rebase?

ncooke3:
@paulb777 right, sorry for the confusion. I just knew that I'd get these CI issues when trying to merge the branch into v9 so I thought I'd open this PR and add context.

ncooke3:
This should be good for review now. CI failures look expected. cc: @paulb777

paulb777 (approved):

PR #9445, opened 2022-03-14
Include Swift extensions in zip and Carthage

paulb777:


paulb777:
Closing in favor of a rebase onto the v9 branch in #9471 since Xcode 12 has issues resolved with Xcode 13

PR #9452, opened 2022-03-14
Record the size of the request bytes for POST/PUT URL Requests

visumickey:
Currently we record only the request bytes only for upload requests. But there could be request bytes even for non-upload requests (POST/PUT). This change would start recording the request payload bytes for POST and PUT requests for dataTask requests.


leotianlizhan (approved):
LGTM

paulb777 (approved):

paulb777 (commented):
Please fix formatting

jeremyjiang-dev (approved):
Thanks for making this change!

PR #9463, opened 2022-03-16
[v9] Remove GHA workflows that test with `--use-libraries`

ncooke3:
#no-changelog

### DIFF
--- a/FirebaseMessaging/Tests/UnitTestsSwift/FIRMessagingAPITest.swift
+++ b/FirebaseMessaging/Tests/UnitTestsSwift/FIRMessagingAPITest.swift
@@ -48,7 +48,7 @@ func apis() {
   case .sandbox: ()
   case .unknown: ()
   // The following case serves to silence the warning that this
-  // enum "may have additional unknown value". In the event that the existing
+  // enum "may have additional unknown values". In the event that the existing
   // cases change, this switch statement will generate
   // a "Switch must be exhaustive" warning.
   @unknown default: ()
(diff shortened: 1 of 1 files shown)


.github/workflows/messaging.yml
```diff
@@ -148,16 +148,15 @@ jobs:
...
+          '--use-modular-headers'
         ]
     needs: pod-lib-lint
     steps:
     - uses: actions/checkout@v2
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: PodLibLint Messaging Cron
-      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }}
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseMessaging.podspec --platforms=${{ matrix.target }} --allow-warnings
```
paulb777, line 159: Why is ` --allow-warnings` added? Are the warnings fixable?

paulb777 (commented):
Thanks for working through this one!  Let's resolve the warning question and then good to go.

.github/workflows/messaging.yml
ncooke3: Yep, so the messaging tests have an [intentional warning](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Tests/UnitTestsSwift/FIRMessagingAPITest.swift#L44-L52) that [fails a linting](https://github.com/firebase/firebase-ios-sdk/runs/5591176589?check_suite_focus=true).

paulb777: Hmmm, Maybe we should run two pod lib lints so we can still catch warnings in the library?

One with `--skip-tests` and one with `--allow-warnings`

paulb777 (approved):

.github/workflows/messaging.yml
ncooke3: Discussed offline. Added `@unkown default` to silence warning. Resolved in latest work.

PR #9467, opened 2022-03-16
[#9407]Enable Firestore data bundle to support empty array

cherylEnkidu:
#no-changelog
[Closes [#9407](https://github.com/firebase/firebase-ios-sdk/issues/9407)]

Unicode character wasn't the reason why decode failed. When user's bundle text file contains empty array, it doesn't fulfill  the `RequireArray` condition in `BundleSerializer`. This PR changes the requirements to Optional which can better support empty array decoding.


Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -614,11 +624,19 @@ FilterList BundleSerializer::DecodeCompositeFilter(JsonReader& reader,
...
+  auto filters = reader.OptionalArray("filters", filter, default_filters);
+
+  const json default_objects;
   FilterList result;
   for (const auto& f : filters) {
-    result = result.push_back(
-        DecodeFieldFilter(reader, reader.RequiredObject("fieldFilter", f)));
+    const json& field_filter =
+        reader.OptionalObject("fieldFilter", f, default_objects);
+    const json& unary_filter =
```
wu-hui, line 635: I think we can avoid handling unary if fieldFilter presents already.

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -614,11 +624,21 @@ FilterList BundleSerializer::DecodeCompositeFilter(JsonReader& reader,
     return {};
   }
 
-  auto filters = reader.RequiredArray("filters", filter);
+  std::vector<json> default_filters;
+  auto filters = reader.OptionalArray("filters", filter, default_filters);
```
wu-hui, line 628: Should it be `const auto&`?

Firestore/core/test/unit/bundle/bundle_serializer_test.cc
```diff
@@ -858,6 +850,13 @@ TEST_F(BundleSerializerTest, DecodesCompositeFilter) {
   VerifyNamedQueryRoundtrip(original);
 }
 
+TEST_F(BundleSerializerTest, DecodesCompositeNullFilter) {
+  core::Query original = testutil::Query("colls")
+                             .AddingFilter(Filter("f1", "==", nullptr))
+                             .AddingFilter(Filter("f2", "==", nullptr));
```
wu-hui, line 856: Can you add one more case here where `f1 == nullptr` and `f2 == some_value`?

wu-hui (approved):
Approved, but please address the comments.

PR #9478, opened 2022-03-18
Explicit async/await implementations for previous implicit ones

paulb777:
#no-changelog

### DIFF
--- a/FirebaseStorageSwift/Tests/Integration/StorageAsyncAwait.swift
+++ b/FirebaseStorageSwift/Tests/Integration/StorageAsyncAwait.swift
@@ -17,8 +17,8 @@ import FirebaseCore
 import FirebaseStorage
 import XCTest
 
-#if swift(>=5.9)
-  @available(iOS 15, *)
+#if swift(>=5.5)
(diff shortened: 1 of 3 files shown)


FirebaseStorageSwift/Tests/Integration/StorageAsyncAwait.swift
```diff
@@ -17,8 +17,8 @@ import FirebaseCore
 import FirebaseStorage
 import XCTest
 
-#if swift(>=5.9)
-  @available(iOS 15, *)
+#if swift(>=5.5)
```
ncooke3, line 20: Does this need `&& canImport(_Concurrency)` too?

ncooke3 (approved):
LGTM.

FirebaseStorageSwift/Tests/Integration/StorageAsyncAwait.swift
paulb777: Thanks! Technically, It might only be needed for Xcode 13.0 which we won't support in v9 - but I added it for consistency.

ncooke3 (approved):

PR #9480, opened 2022-03-19
Do we need umbrella headers?

paulb777:
Experiment with removing explicit umbrella headers, now that we're requiring modules.

#no-changelog


paulb777:
It looks like the answer is yes for Objective C. There are plenty of examples in the quickstarts.

Issue #9072, opened 2021-12-09
Better Cross Swift Package CI


paulb777:
Update CI involving GoogleAppMeasurement to depend on an environment variable like `ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil` from https://github.com/apple/swift-docc/blob/5e0729b7b63f5ef39dc5b5af05e18431c86f8427/Package.swift#L107-L132 instead of a [sed command](https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/setup_spm_tests.sh#L24).

paulb777:
This particular fix has been done - although we may want a better solution. Closing for solving the environment piece. cc: @ncooke3

PR #9482, opened 2022-03-20
[v9] Reorganize `FirebaseCore*` modules

ncooke3:
```mermaid
graph BT
    G[/Heartbeat Logging Source\] --- A
    F[/Future parts of Core written in Swift\] -.- A
    A{{FirebaseCoreInternal}} --> B(FirebaseCore)
    B --> C{{FirebaseCoreExtension}}
    B --> D((Clients))
    C --> E(Firebase Product SDKs)
    B --> E
    E --> D
```
**Key**
- Arrows indicate dependencies 
- Lines indicate the code is _contained_ by the module above it
- Trapezoid nodes represent the bottom of the dep chain within Firebase
- Hexagon nodes are for modules designed for internal Firebase use
- Rounded rectangle nodes are for modules with API that's publicly exposed to customers

#no-changelog


ncooke3:
The `heartbeat_logging` workflows are failing because they are out of date. 

Not super sure what's happening in https://github.com/firebase/firebase-ios-sdk/runs/5619184906?check_suite_focus=true...

Crashlytics/Crashlytics/DataCollection/FIRCLSDataCollectionArbiter.m
```diff
@@ -20,7 +20,7 @@
 #import "FBLPromises.h"
 #endif
 
-#import "FirebaseCore/Internal/FirebaseCoreInternal.h"
+#import "FirebaseCore/Extension/FirebaseCoreInternal.h"
```
paulb777, line 23: Now that module imports are ok, we might want to delete FirebaseCoreInternal.h in favor of `@import FirebaseCoreExtension;`  - although I'm not sure we have a g3 strategy for that and may need to keep FirebaseCoreInternal.h at least on the g3 side?

scripts/spm_test_schemes/FirebaseCoreInternalTests.xcscheme
```diff
@@ -1,26 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1300"
+   LastUpgradeVersion = "1320"
    version = "1.3">
    <BuildAction
       parallelizeBuildables = "YES"
       buildImplicitDependencies = "YES">
-      <BuildActionEntries>
```
paulb777, line 8: Keep it buildable

scripts/pod_lib_lint.rb
```diff
@@ -68,6 +68,7 @@ def main(args)
   podspec_file = pod_args[0]
   # Figure out which dependencies are local
   deps = find_local_deps(podspec_file, ignore_local_podspecs.to_set)
+  puts 'podspec_file: ' + podspec_file + '\n podspec_file_deps: ' + deps.join(', ')
```
paulb777, line 71: Don't commit these here

ncooke3: Reverted in [b704b09](https://github.com/firebase/firebase-ios-sdk/pull/9482/commits/b704b098b6c556806b8d6366265fba774105c885)

scripts/spm_test_schemes/FirebaseCoreInternalTests.xcscheme
ncooke3: Good catch! Done in [bb32a8b](https://github.com/firebase/firebase-ios-sdk/pull/9482/commits/bb32a8b550fcf3810295691b82941392d3278e40)

.github/workflows/core_internal.yml
```diff
@@ -0,0 +1,79 @@
...
+        flags: [
+          '--use-static-frameworks'
+        ]
+    needs: pod-lib-lint
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Bundler
+      run: scripts/setup_bundler.sh
+    - name: PodLibLint CoreInternal Cron
+      run: scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb FirebaseCoreInternal.podspec --platforms=${{ matrix.target }} ${{ matrix.flags }}
```
paulb777, line 79: Add newline

Crashlytics/Crashlytics/DataCollection/FIRCLSDataCollectionArbiter.m
paulb777: I'm ok to defer this decision.

Package.swift
```diff
@@ -149,7 +149,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Note that CI changes the version to the head of main for CI.
       // See scripts/setup_spm_tests.sh.
-      .exact("8.12.0")
+      .branch("main")
```
paulb777, line 152: Shouldn't be checked in. We really need to do https://github.com/firebase/firebase-ios-sdk/issues/9072

.github/workflows/core_internal.yml
```diff
@@ -0,0 +1,79 @@
+name: core_internal
```
paulb777, line 1: Don't we need a `core_extension.yml` also?

paulb777 (commented):
ReleaseTooling changes are also missing, but you may want to wait for #9471 to merge

Package.swift
ncooke3: Thanks. Done in 0826e4b42c1d929a4b926180287d428ecb8f7a58.

.github/workflows/core_internal.yml
ncooke3: Done in 0826e4b42c1d929a4b926180287d428ecb8f7a58.

.github/workflows/core_internal.yml
ncooke3: Yes, added in 0826e4b42c1d929a4b926180287d428ecb8f7a58.

Currently, the workflow is set to lint the `FirebaseCoreExtension.podspec` (which currently only contains source headers). IIUC, I can't just build the source for SPM (no scheme generates for the `FirebaseCoreExtension` target). I could think of three options:
1. No op– don't run `spm` workflows for FirebaseCoreExtension.
2. Add a `FirebaseCoreExtensionTesting` scheme that has no tests but builds `FirebaseCoreExtension`.
3. Add a `FirebaseCoreExtensionTesting` scheme and rearrange the `FirebaseCore/Extension` directory and add API build tests for both Swift and ObjC? The package modifications would look something like:
image

WDYT?

paulb777: Since it's only headers, `pod lib lint` testing should be sufficient.

ncooke3: Done in [e9483fa](https://github.com/firebase/firebase-ios-sdk/pull/9482/commits/e9483fa2565916206cf61e0763ac52f9354e5db1)

paulb777 (approved):

Issue #9258, opened 2022-01-25
FIRESTORE INTERNAL ASSERTION FAILED: Trying to compare documents on fields that don't exist. (expected value1.has_value() && value2.has_value())


sspogra:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version:13.1
  * Firebase SDK version:8.10.0
  * Installation method: `CocoaPods `
  * Firebase Component: Auth, Core, Database, Firestore, Messaging, Storage
  * Target platform(s): `iOS `

### [REQUIRED] Step 2: Describe the problem

Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x9904c __exceptionPreprocess
1  libobjc.A.dylib                0x15f54 objc_exception_throw
2  Foundation                     0x130834 -[NSMutableDictionary(NSMutableDictionary) initWithContentsOfFile:]
3  FirebaseFirestore              0x44868 (Missing)
4  FirebaseFirestore              0x44348 (Missing)
5  FirebaseFirestore              0x914b4 (Missing)
6  FirebaseFirestore              0x91554 (Missing)
7  FirebaseFirestore              0xd88a4 (Missing)
8  FirebaseFirestore              0xdf468 (Missing)
9  FirebaseFirestore              0x3b84c (Missing)
10 FirebaseFirestore              0x40650 (Missing)
11 FirebaseFirestore              0x40470 (Missing)
12 FirebaseFirestore              0x38920 (Missing)
13 FirebaseFirestore              0x38528 (Missing)
14 FirebaseFirestore              0x3836c (Missing)
15 FirebaseFirestore              0x121fec (Missing)
16 FirebaseFirestore              0x10bd58 (Missing)
17 FirebaseFirestore              0x10b860 (Missing)
18 FirebaseFirestore              0x68b18 (Missing)
19 FirebaseFirestore              0x5b40 (Missing)
20 FirebaseFirestore              0x1169c0 (Missing)
21 libdispatch.dylib              0x4660 _dispatch_client_callout
22 libdispatch.dylib              0xbde4 _dispatch_lane_serial_drain
23 libdispatch.dylib              0xc958 _dispatch_lane_invoke
24 libdispatch.dylib              0x171a8 _dispatch_workloop_worker_thread
25 libsystem_pthread.dylib        0x10f4 _pthread_wqthread
26 libsystem_pthread.dylib        0xe94 start_wqthread

Fatal Exception: NSInternalInconsistencyException
[...]

PR #9489, opened 2022-03-21
Include ToString() in Firestore's Query::Comparator() assertion failure message

dconeybe:
Improve the assertion failure message in `Query::Comparator()` to also include the result of calling `ToString()`. This may help find the root cause of https://github.com/firebase/firebase-ios-sdk/issues/9258.


cherylEnkidu (approved):
LGTM

PR #9491, opened 2022-03-21
Disable `specs_testing` on `v9`.

ncooke3:
I'm thinking this should speed up CI since these are failing on `v9` anyway.

Disabled and added `v9` TODO to re-enable when merging into `master`.

#no-changelog


ncooke3:
WDYT of this @paulb777? 

If you think this PR is a good idea, I could disable the `health-metrics-presubmit` workflows as well?

paulb777 (approved):

PR #9501, opened 2022-03-22
[v9, Infra] Add workflow cancellation for optimizing CI on `v9` branch

ncooke3:
Cherry-picked #9482 from `master` and added commit to add cancellation to new workflows on v9.

#no-changelog


paulb777:
I'd prefer to create a v9b that is v9 rebased on master and then deprecate v9.

Issue #9426, opened 2022-03-08
Firestore db.runTransaction() async causes nil unwrapping fatal error if no data is returned


ionothanus:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.2.1
  * Firebase SDK version: 8.10.0
  * Installation method: Swift Package Manager
  * Firebase Component: Firestore
  * Target platform(s): iOS 15.2

#### Steps to reproduce:
1. Start a transaction using the following async declaration of `db.runTransaction()`:
```
func runTransaction(_ updateBlock: @escaping (Transaction, NSErrorPointer) -> Any?) async throws -> Any
```
2. Successfully perform the transaction, without encountering or returning an error (i.e., leave `error?.pointee` unset).
3. `return nil` at the end of the transaction update block, as demonstrated in the basic [transaction example in the documentation](https://firebase.google.com/docs/firestore/manage-data/transactions#transactions).

Expected result: the transaction is completed with a nil return value.
Actual result: the error `Thread 1: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value` is encountered in `functional` line 1885. 

Partial stack trace:  

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x0000000185918d7c in _swift_runtime_on_report ()
#1	0x0000000185998e54 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x00000001855ce1a8 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x00000001855cdf48 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x00000001855cd83c in _assertionFailure(_:_:file:line:flags:) ()
#5	0x00000001855ce254 in _diagnoseUnexpectedNilOptional(_filenameStart:_filenameLength:_filenameIsASCII:_line:_isImplicitUnwrap:) ()
#6	0x0000000104866e3c in @objc completion handler block implementation for @escaping @callee_unowned @convention(block) (@unowned Swift.AnyObject?, @unowned NSError?) -> () with result type Any ()
[...]

PR #9502, opened 2022-03-22
Fix nil async/await import error

morganchen12:
Hopefully fixes https://github.com/firebase/firebase-ios-sdk/issues/9426.

### DIFF
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 - [changed] Add more details to the assertion failure in Query::Comparator() to
   help with future debugging (#9258).
+- [fixed] Fixed an issue where returning `nil` from the update closure when
+  running a transaction caused a crash in Swift (#9426).
 
 # v8.14.0
 - [fixed] Fixed compiler warnings in `local_serializer.cc` about "implicit
--- a/Firestore/CHANGELOG.md
+++ b/Firestore/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Unreleased
 - [changed] Add more details to the assertion failure in Query::Comparator() to
   help with future debugging (#9258).
-- [fixed] Fixed an issue where returning `nil` from the update closure when
-  running a transaction caused a crash in Swift (#9426).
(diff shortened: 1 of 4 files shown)


Firestore/Swift/Source/AsyncAwait/Firestore+AsyncAwait.swift
```diff
@@ -53,5 +53,50 @@ import Foundation
...
+
+    /// Executes the given updateBlock and then attempts to commit the changes applied within an atomic
+    /// transaction.
+    ///
+    /// The maximum number of writes allowed in a single transaction is 500, but note that each usage of
+    /// `FieldValue.serverTimestamp()`, `FieldValue.arrayUnion()`, `FieldValue.arrayRemove()`, or
+    /// `FieldValue.increment()` inside a transaction counts as an additional write.
+    ///
+    /// In the updateBlock, a set of reads and writes can be performed atomically using the
+    /// `FIRTransaction` object passed to the block. After the updateBlock is run, Firestore will attempt
```
schmidt-sebastian, line 65: Drop the FIR prefix.

Firestore/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 - [changed] Add more details to the assertion failure in Query::Comparator() to
   help with future debugging (#9258).
+- [fixed] Fixed an issue where returning `nil` from the update closure when
```
schmidt-sebastian, line 4: We should probably elude to the fact that this might be a breaking change that can be addressed by importing FirebaseFirestoreSwift.

schmidt-sebastian (approved):
We should probably wait to merge until 4/6 since this is technically a breaking change, even if it just affects a beta API surface. I would not ask for this if the breaking change release was much further out.

Firestore/Swift/Source/AsyncAwait/Firestore+AsyncAwait.swift
morganchen12: Done, also did this in `FIRFirestore.h` for consistency.

Firestore/CHANGELOG.md
morganchen12: Done.

PR #9506, opened 2022-03-23
[Heartbeat Logging] Add heartbeat logger to `FIRApp` and manage `FIRHeartbeatLogger` instances

ncooke3:
I will open up PRs of the product SDK integration once this is merged.

#no-changelog

### DIFF
--- a/FirebaseCore/Extension/FIRAppInternal.h
+++ b/FirebaseCore/Extension/FIRAppInternal.h
@@ -18,6 +18,7 @@
 
 @class FIRComponentContainer;
 @protocol FIRLibrary;
+@class FIRHeartbeatLogger;
 
 /**
  * The internal interface to `FirebaseApp`. This is meant for first-party integrators, who need to
@@ -96,6 +97,11 @@ extern NSString *const FIRAuthStateDidChangeInternalNotificationUIDKey;
  * The container of interop SDKs for this app.
  */
 @property(nonatomic) FIRComponentContainer *container;
+
+/*
+ * The heartbeat logger associated with this app.
+ */
+@property(readonly) FIRHeartbeatLogger *heartbeatLogger;
 
 /**
  * Checks if the default app is configured without trying to configure it.
--- a/FirebaseCore/Sources/FIRHeartbeatLogger.m
+++ b/FirebaseCore/Sources/FIRHeartbeatLogger.m
@@ -65,8 +65,7 @@ NSString *_Nullable FIRHeaderValueFromHeartbeatsPayload(FIRHeartbeatsPayload *he
   return payload;
 }
 
-// TODO: Rename to `heartbeatCodeForToday` in future PR's.
-- (FIRHeartbeatInfoCode)heartbeatCode {
+- (FIRHeartbeatInfoCode)heartbeatCodeForToday {
   FIRHeartbeatsPayload *todaysHeartbeatPayload = [_heartbeatController flushHeartbeatFromToday];
 
   if ([todaysHeartbeatPayload isEmpty]) {
--- a/FirebaseCore/Sources/Public/FirebaseCore/FIROptions.h
+++ b/FirebaseCore/Sources/Public/FirebaseCore/FIROptions.h
@@ -60,6 +60,7 @@ NS_SWIFT_NAME(FirebaseOptions)
  */
 @property(nonatomic, copy) NSString *GCMSenderID NS_SWIFT_NAME(gcmSenderID);
 
+// TODO(v9): Can this be non-null.
 /**
  * The Project ID from the Firebase console, for example @"abc-xyz-123".
  */
--- a/FirebaseCore/Sources/Public/FirebaseCore/FIROptions.h
+++ b/FirebaseCore/Sources/Public/FirebaseCore/FIROptions.h
@@ -60,7 +60,6 @@ NS_SWIFT_NAME(FirebaseOptions)
  */
 @property(nonatomic, copy) NSString *GCMSenderID NS_SWIFT_NAME(gcmSenderID);
 
-// TODO(v9): Can this be non-null.
 /**
  * The Project ID from the Firebase console, for example @"abc-xyz-123".
  */
--- a/FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m
+++ b/FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m
@@ -96,7 +96,7 @@
   FIRHeartbeatLogger *heartbeatLogger = self.heartbeatLogger;
   // When
   [heartbeatLogger log];
-  FIRHeartbeatInfoCode heartbeatInfoCode = [heartbeatLogger heartbeatCode];
+  FIRHeartbeatInfoCode heartbeatInfoCode = [heartbeatLogger heartbeatCodeForToday];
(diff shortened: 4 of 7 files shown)


FirebaseCore/Sources/FIRHeartbeatLogger.m
```diff
@@ -14,11 +14,30 @@
...
+/// `FIRHeartbeatLogger` instances in a dictionary.
+@interface FIRWeakContainer<Object> : NSObject
+@property(readonly, weak) Object object;
+@end
+
+@implementation FIRWeakContainer
+
++ (instancetype)containerWithObject:(id)obj {
+  FIRWeakContainer *container = [[FIRWeakContainer alloc] init];
+  container->_object = obj;
```
chliangGoogle, line 34: this is an interesting way to bypass readonly property and set the ivar value?

FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m
```diff
@@ -54,6 +54,110 @@ - (void)tearDown {
...
+  XCTAssert(strongHeartbeatLogger == weakHeartbeatLogger,
+            "Instances should reference the same object.");
+  // When
+  strongHeartbeatLogger = nil;
+  // Then
+  XCTAssertNil(strongHeartbeatLogger);
+  XCTAssertNil(weakHeartbeatLogger);
+}
+
+- (void)testCachedInstancesAreRemovedUponDeinitAndCanBeRetainedStrongly {
```
chliangGoogle, line 114: RemovedUponDeinit? is that a typo?

chliangGoogle (approved):

FirebaseCore/Sources/FIRHeartbeatLogger.m
ncooke3: Yep, no setter is generated because of `readonly` but you can bypass. this within the implementation by changing the underlying ivar.

FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m
ncooke3: It's not a typo, but I should make it clearer to readers.

Cached instances are stored at the _class_ level of `FIRHeartbeatLogger` in a dictionary of `NSString` appIDs mapping to weakly boxed `FIRHeartbeatLogger`. When a `FIRHeartbeatLogger`'s retain count goes to 0, I want it to be removed from the `cachedInstances`. IIUC, this should also prevent the (rare but possible?) growing of `FIRHeartbeatLogger` instances stored if the end user kept configuring different Firebase apps.

In the `dealloc` of `FIRHeartbeatLogger`, it removes itself from the `cachedInstances`:
https://github.com/firebase/firebase-ios-sdk/blob/722ec3e3b400df6695f16ac3ba3c48eb950970f2/FirebaseCore/Sources/FIRHeartbeatLogger.m#L103-L106

In this test method, I want to verify that cached instance are removed (from the cachedInstances static storage) when they are dealloc'd (`heartbeatLogger1` and `heartbeatLogger2`) but can be strongly retained (`heartbeatLogger3`).

I'll change to:
```suggestion
- (void)testCachedInstancesAreRemovedUponDeallocButCanBeRetainedStrongly {
```

Lmk if this sounds reasonable!

FirebaseCore/Sources/Public/FirebaseCore/FIROptions.h
```diff
@@ -60,6 +60,7 @@ NS_SWIFT_NAME(FirebaseOptions)
  */
 @property(nonatomic, copy) NSString *GCMSenderID NS_SWIFT_NAME(gcmSenderID);
 
+// TODO(v9): Can this be non-null.
```
ncooke3, line 63: Is there a historical reason as to why `projectID` is `nullable`? While re-working this PR, I saw some usage of it (in AppCheck sdk for example) that pretty much relies on the `projectID` being there for the SDK to function. I was wondering if this property could be made non-null in v9?



@paulb777 
@chliangGoogle 
@ryanwilson

FirebaseCore/Tests/Unit/FIRHeartbeatLoggerTest.m
ncooke3: Reworked in 666f6e1b3a9406904178f3902cb233f444c698e8

FirebaseCore/Sources/FIRHeartbeatLogger.m
```diff
@@ -65,8 +65,7 @@ - (FIRHeartbeatsPayload *)flushHeartbeatsIntoPayload {
   return payload;
 }
 
-// TODO: Rename to `heartbeatCodeForToday` in future PR's.
-- (FIRHeartbeatInfoCode)heartbeatCode {
+- (FIRHeartbeatInfoCode)heartbeatCodeForToday {
```
paulb777, line 68: I'm sure this has come up before, but I'd like to better understand `Today`.

- What if we decide we want more frequent heartbeats in the future? Or is the 24 hour window an ongoing constraint?
- Is today specific to the device, and how does it relate to timezone differences?

FirebaseCore/Extension/FIRAppInternal.h
```diff
@@ -97,6 +98,11 @@ extern NSString *const FIRAuthStateDidChangeInternalNotificationUIDKey;
  */
 @property(nonatomic) FIRComponentContainer *container;
 
+/*
+ * The heartbeat logger associated with this app.
+ */
+@property(readonly) FIRHeartbeatLogger *heartbeatLogger;
```
paulb777, line 104: So much cleaner! Thanks.

paulb777 (approved):

FirebaseCore/Sources/FIRHeartbeatLogger.m
ncooke3: https://github.com/firebase/firebase-ios-sdk/blob/4e7a6441a54076423b215d1fbb2f0e4fd338b20b/FirebaseCore/Extension/FIRHeartbeatLogger.h#L49-L55

This API is a direct replacement of `+[FIRHeartbeatInfo heartbeatCodeForTag:]` that should be used by product SDKs whose backend can only parse single a single day's heartbeat code (`FIRHeartbeatInfoCode`).

---

> What if we decide we want more frequent heartbeats in the future?

This is a one line change in the heartbeats logging source that would be reflected in the return type of this API:
https://github.com/firebase/firebase-ios-sdk/blob/4e7a6441a54076423b215d1fbb2f0e4fd338b20b/FirebaseCore/Extension/FIRHeartbeatLogger.h#L41-L47

The `heartbeatCodeForToday` is never designed to send more than one heartbeat. As more backends migrate to parse payloads with more heartbeats, then this `heartbeatCodeForToday` would be removed entirely.
---

> Or is the 24 hour window an ongoing constraint?

It is an ongoing constraint for product SDKs that havenot upgraded their backends to parse requests with multiple heartbeats.

> Is today specific to the device, and how does it relate to timezone differences?

---

_Today_ is specific to device and is standardized to use the GMT time zone with no time offset.

ncooke3:
I'm going to merge this to open PRs for the product integrations. Feel free to leave additional feedback here and I will address in following PRs

Issue #6978, opened 2020-11-17
handleUniversalLink always returns false when app launch after installation


svedm:
### Step 1: Describe your environment

  * Xcode version: 12.1
  * Firebase SDK version: 7.0-spm-beta
  * Installation method: Swift Package Manager
  * Firebase Component: DynamicLinks

### Step 2: Describe the problem

`handleUniversalLink` always returns `false` when app launch after installation

#### Steps to reproduce:

1. Create dynamic link in firebase console. Link should open appstore if application not installed
2. Remove application from device
3. Open dynamic link in browser, check `Save my place in the app. A link will be copied to continue to this page.`
4. Install app from Xcode in debug mode
5. `func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:])` called with url from firebase

Expected behavior: 
 -  `DynamicLinks.dynamicLinks().handleUniversalLink(url)` returns true

Real behavior:
 -  `DynamicLinks.dynamicLinks().handleUniversalLink(url)` always returns false

When Im opening installed app from this link all works as expected.

Diagnostic output:
```
---- Firebase Dynamic Links diagnostic output start ----
Firebase Dynamic Links framework version 7.0.0
System information: OS iOS, OS version 13.4, model iPhone
Current date 2020-11-17 12:06:01 +0000
Device locale ru-CA (raw ru_CA), timezone America/Edmonton
	Specified custom URL scheme is my.bundle.id and Info.plist contains such scheme in CFBundleURLTypes key.
	AppID Prefix: Z8M*****M8, Team ID:  Z8M*****M8, AppId Prefix equal to Team ID: YES
performDiagnostic completed successfully! No errors found.
---- Firebase Dynamic Links diagnostic output end ----
```

christibbs:
Internal bug number 174061690.

sandeep576:
same here, always getting false `DynamicLinks.dynamicLinks().handleUniversalLink(url)`

nuborian:
Anything new in this? :(

FitzAfful:
Any update with this?

LeonXiaoZ:
For the open after first install case, we should call 'DynamicLinks.dynamicLinks().dynamicLink(fromCustomSchemeURL: url)', right?

@available(iOS 9.0, *)
func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any]) -> Bool {
[...]

PR #9516, opened 2022-03-24
Fdl unit tests

eldhosembabu:
FDL Custom domain validation logic change.

- We were doing regex search because of which the regex might sometime matches based on the contents of the url present in the link query parameter. Now this is changed to start to end validation with. "^" and "$".
- We were not honoring the order of link query param in the url before as we were looking the link param to be the first param in the url. Now this got changed to accept link param anywhere in the url.

Issue : https://github.com/firebase/firebase-ios-sdk/issues/6978


FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v8.10.0
```
paulb777, line 1: Next release is 8.15.0. See https://github.com/firebase/firebase-ios-sdk/milestones

paulb777 (commented):
Looks like GitHub CI triggering is currently broken.

Have you confirmed the test breaks locally without the fix and passes with it?

FirebaseDynamicLinks/CHANGELOG.md

paulb777 (approved):

eldhosembabu:
Modified logic to handle @"https://google.com?some=qry&somelink=https%3A%2F%2Fsomedomain" pointed out by 3p dev in the related issue.

paulb777 (approved):
Much more readable than the regex and great to add the new tests. Thanks!

PR #9519, opened 2022-03-25
Add indexing support to Target

wu-hui:


Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +507,109 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+      value->which_value_type = value_tag;
+      value->timestamp_value.seconds = std::numeric_limits<int64_t>::min();
+      value->timestamp_value.nanos = 0;
+      return value;
+    }
+
+    case google_firestore_v1_Value_string_value_tag: {
+      Message<google_firestore_v1_Value> value;
+      value->which_value_type = value_tag;
+      value->string_value = nanopb::MakeArray<pb_bytes_array_s>(1);
```
schmidt-sebastian, line 538: You should be able to set this to nullptr (and it looks like you might be leaking 1 byte here since the size is set to zero, but the allocated size is 0)

Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +507,109 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+      value->string_value->size = 0;
+      return value;
+    }
+
+    case google_firestore_v1_Value_bytes_value_tag: {
+      Message<google_firestore_v1_Value> value;
+      value->which_value_type = value_tag;
+      value->bytes_value = nanopb::MakeArray<pb_bytes_array_t>(1);
```
schmidt-sebastian, line 546: Same, this can be a nullptr

Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +507,109 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+      value->geo_point_value.latitude = -90.0;
+      value->geo_point_value.longitude = -180.0;
+      return value;
+    }
+
+    case google_firestore_v1_Value_array_value_tag: {
+      Message<google_firestore_v1_Value> value;
+      value->which_value_type = value_tag;
+      value->array_value.values =
+          nanopb::MakeArray<google_firestore_v1_Value>(0);
```
schmidt-sebastian, line 566: Probably can be a nullptr.

Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +507,109 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+      value->array_value.values_count = 0;
+      return value;
+    }
+
+    case google_firestore_v1_Value_map_value_tag: {
+      Message<google_firestore_v1_Value> value;
+      value->which_value_type = value_tag;
+      value->map_value.fields =
```
schmidt-sebastian, line 574: Probably can be a nullptr.

Firestore/core/src/model/value_util.cc
```diff
@@ -522,7 +654,32 @@ Message<google_firestore_v1_Value> MaxValue() {
 }
 
 bool IsMaxValue(const google_firestore_v1_Value& value) {
-  return value == *MaxValue();
+  bool is_max =
+      (value.which_value_type == google_firestore_v1_Value_map_value_tag);
+  if (!is_max) {
```
schmidt-sebastian, line 659: Probably just a personal preference, but I think this would be easier to read if you directly used the conditions in the if statement rather than `is_max`.

Firestore/core/src/model/value_util.h
```diff
@@ -67,6 +68,14 @@ void SortFields(google_firestore_v1_ArrayValue& value);
 util::ComparisonResult Compare(const google_firestore_v1_Value& left,
                                const google_firestore_v1_Value& right);
 
+util::ComparisonResult MaxValue(
+    const absl::optional<google_firestore_v1_Value>& left,
+    const absl::optional<google_firestore_v1_Value>& right);
+
+const absl::optional<google_firestore_v1_Value>* MinValue(
```
schmidt-sebastian, line 75: These names clash with the already existing `MinValue`. Can you update the names to reflect that there are doing a comparison?

Firestore/core/src/remote/serializer.cc
```diff
@@ -1115,9 +1117,10 @@ OrderBy Serializer::DecodeOrderBy(
   return OrderBy(std::move(field_path), direction);
 }
 
-google_firestore_v1_Cursor Serializer::EncodeBound(const Bound& bound) const {
+google_firestore_v1_Cursor Serializer::EncodeBound(const Bound& bound,
```
schmidt-sebastian, line 1120: I am not a huge fan of `as_start` as it is not 100% self explanatory. Can we pass in  `position` and the desired value of `before` instead? FWIW I do think keeping this method is better than inlining it like we did on Android.

Firestore/core/src/remote/serializer.cc
```diff
@@ -1128,15 +1131,17 @@ google_firestore_v1_Cursor Serializer::EncodeBound(const Bound& bound) const {
   return result;
 }
 
-Bound Serializer::DecodeBound(google_firestore_v1_Cursor& cursor) const {
+Bound Serializer::DecodeBound(google_firestore_v1_Cursor& cursor,
+                              bool as_start) const {
```
schmidt-sebastian, line 1135: Same - maybe we should pass in the cursor values and `inclusive`

Firestore/core/test/unit/core/target_test.cc
```diff
@@ -0,0 +1,385 @@
...
+  }
+}
+
+TEST(TargetTest, EmptyQueryBound) {
+  Target target = testutil::Query("c").ToTarget();
+  FieldIndex index = MakeFieldIndex("c");
+
+  auto lower_bound = target.GetLowerBound(index);
+  ASSERT_TRUE(lower_bound.has_value());
+  VerifyBound(lower_bound.value(), true, {});
```
schmidt-sebastian, line 85: Can VerifyBound take an optional, so that we can move the `HasValue` check?

Firestore/core/test/unit/core/target_test.cc
```diff
@@ -0,0 +1,385 @@
...
+using testutil::Filter;
+using testutil::MakeFieldIndex;
+using testutil::Map;
+using testutil::OrderBy;
+using testutil::Query;
+using testutil::Ref;
+using testutil::Value;
+using testutil::Vector;
+
+void VerifyBound(Bound bound,
```
schmidt-sebastian, line 66: Pass Bound by reference?

Firestore/core/test/unit/core/target_test.cc
```diff
@@ -0,0 +1,385 @@
...
+                 bool inclusive,
+                 std::vector<google_firestore_v1_Value> values) {
```
schmidt-sebastian, line 68: Pass by reference?

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -112,6 +114,25 @@ class ValueUtilTest : public ::testing::Test {
     }
   }
 
+  void VerifyRelaxedAscending(Message<google_firestore_v1_ArrayValue>& left,
```
schmidt-sebastian, line 117: Can you add a method comment here and above?

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -112,6 +114,25 @@ class ValueUtilTest : public ::testing::Test {
...
+                              Message<google_firestore_v1_ArrayValue>& right) {
+    for (pb_size_t i = 0; i < left->values_count; ++i) {
+      for (pb_size_t j = 0; j < right->values_count; ++j) {
+        EXPECT_NE(ComparisonResult::Descending,
```
schmidt-sebastian, line 121: FWIW, I think this is pretty clever but it took me a while to understand how this was meant to work.

schmidt-sebastian (commented):
Thanks! I think a bulk of my "requested changes" are memory management related and I am not sure if all my suggestions are feasible.

Firestore/core/src/remote/serializer.cc
wu-hui: Done..more or less.

Firestore/core/test/unit/core/target_test.cc
wu-hui: Done.

Firestore/core/test/unit/core/target_test.cc
wu-hui: Done.

Firestore/core/test/unit/core/target_test.cc
wu-hui: Done.

Firestore/core/test/unit/model/value_util_test.cc
wu-hui: Done.

Firestore/core/test/unit/model/value_util_test.cc
wu-hui: Done.

Firestore/core/src/bundle/bundle_serializer.h
```diff
@@ -121,9 +121,19 @@ class BundleSerializer {
...
-                          const nlohmann::json& query,
-                          const char* bound_name) const;
+
+  core::Bound DecodeStartAtBound(JsonReader& reader,
+                                 const nlohmann::json& query,
+                                 const char* bound_name) const;
+  core::Bound DecodeEndAtBound(JsonReader& reader,
+                               const nlohmann::json& query,
+                               const char* bound_name) const;
+  // Decodes a `bound` JSON and returns a pair whose first element is the value
```
schmidt-sebastian, line 131: Missing empty line.

Firestore/core/src/core/target.cc
```diff
@@ -33,6 +35,38 @@ namespace firestore {
...
+std::vector<google_firestore_v1_Value> ValuesFrom(
+    std::unordered_map<std::string,
+                       nanopb::Message<google_firestore_v1_Value>>&&
+        field_value_map) {
+  std::vector<google_firestore_v1_Value> result;
+  for (auto& entry_pair : field_value_map) {
+    result.push_back(*entry_pair.second);
+  }
+
+  return result;
```
schmidt-sebastian, line 66: The spacing used in line 46-53, 61-66 and 82-88 is pretty inconsistent. Feel free to leave as is, but I thought I should point it out.

Firestore/core/src/core/target.cc
```diff
@@ -41,6 +77,275 @@ bool Target::IsDocumentQuery() const {
...
+    for (const auto& field_filter :
+         GetFieldFiltersForPath(segment.field_path())) {
+      switch (field_filter.op()) {
+        case FieldFilter::Operator::Equal:
+        case FieldFilter::Operator::In:
+          // Encode equality prefix, which is encoded in the index value before
+          // the inequality (e.g. `a == 'a' && b != 'b'` is encoded to `value !=
+          // 'ab'`).
+          field_value_map[segment.field_path().CanonicalString()] =
+              model::DeepClone(field_filter.value());
```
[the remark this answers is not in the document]
schmidt-sebastian: This is still showing the clone (here and right below)

Firestore/core/src/core/target.cc
```diff
@@ -41,6 +75,254 @@ bool Target::IsDocumentQuery() const {
...
+  for (const auto& field_filter :
+       GetFieldFiltersForPath(segment.field_path())) {
+    absl::optional<google_firestore_v1_Value> filter_value;
+    bool filter_inclusive = true;
+
+    switch (field_filter.op()) {
+      case FieldFilter::Operator::LessThan:
+      case FieldFilter::Operator::LessThanOrEqual:
+        filter_value =
+            *model::GetLowerBound(field_filter.value().which_value_type);
```
schmidt-sebastian, line 207: This could now theoretically cause a segfault if  GetLowerBound cleans up the memory location before the return type is used. My suggested fix would be to return just a nanopb Value in GetLowerBound and GetUpperBound and make sure that is does not contain any dynamically allocated memory (which we should be able to ensure).

Firestore/core/src/core/target.cc
```diff
@@ -41,6 +75,254 @@ bool Target::IsDocumentQuery() const {
...
+      case FieldFilter::Operator::GreaterThanOrEqual:
+        filter_value = field_filter.value();
+        break;
+      case FieldFilter::Operator::GreaterThan:
+        filter_value = field_filter.value();
+        filter_inclusive = false;
+        break;
+      case FieldFilter::Operator::NotEqual:
+      case FieldFilter::Operator::NotIn:
+        filter_value = *model::MinValue();
```
schmidt-sebastian, line 220: Same problem regarding the use of freed memory exists here.

Firestore/core/src/core/target.cc
```diff
@@ -41,6 +75,254 @@ bool Target::IsDocumentQuery() const {
...
+  for (const auto& field_filter :
+       GetFieldFiltersForPath(segment.field_path())) {
+    absl::optional<google_firestore_v1_Value> filter_value;
+    bool filter_inclusive = true;
+
+    switch (field_filter.op()) {
+      case FieldFilter::Operator::GreaterThanOrEqual:
+      case FieldFilter::Operator::GreaterThan:
+        filter_value =
+            *model::GetUpperBound(field_filter.value().which_value_type);
```
schmidt-sebastian, line 273: Same comment as above.

Firestore/core/src/core/target.h
```diff
@@ -38,6 +41,17 @@ class BundleSerializer;
 namespace core {
 
 using CollectionGroupId = std::shared_ptr<const std::string>;
+using IndexedValues = absl::optional<std::vector<google_firestore_v1_Value>>;
+
+struct IndexedBoundValue {
```
schmidt-sebastian, line 46: Please add comments to these types. I also think IndexedBoundValue can be private (and should maybe just be `IndexBoundValue`)

Firestore/core/src/core/target.h
```diff
@@ -38,6 +41,17 @@ class BundleSerializer;
...
+  bool is_inclusive;
```
schmidt-sebastian, line 47: Do we use the `is_` prefix with booleans in other places?

Firestore/core/src/core/target.h
```diff
@@ -125,6 +172,28 @@ class Target {
...
+   * The result is a pair with an absl::option<Message> representing the value
+   * and a bool to indicate if the result is inclusive
+   */
+  IndexedBoundValue GetAscendingBound(const model::Segment& segment,
+                                      const absl::optional<Bound>& bound);
+  /**
+   * Returns the value for a descending bound of `segment`, using `bound` to
+   * narrow down the result.
+   *
+   * The result is a pair with an absl::option<Message> representing the value
```
schmidt-sebastian, line 191: typo: optional (also returns a vector of Values now)

Here and above

Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +491,105 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+
+    case google_firestore_v1_Value_bytes_value_tag: {
+      Message<google_firestore_v1_Value> value;
+      value->which_value_type = value_tag;
+      value->bytes_value = nullptr;
+      return value;
+    }
+
+    case google_firestore_v1_Value_reference_value_tag:
+      return RefValue(DatabaseId("", ""), DocumentKey::Empty());
```
schmidt-sebastian, line 534: This access could cause a segfault since we are allocating a string. I think we can return a reference_value with a nullptr value, but would need to make sure the SDK treats this properly.

Firestore/core/src/model/value_util.cc
```diff
@@ -486,6 +491,105 @@ std::string CanonicalId(const google_firestore_v1_ArrayValue& value) {
...
+    case google_firestore_v1_Value_bytes_value_tag:
+      return GetLowerBound(google_firestore_v1_Value_reference_value_tag);
+    case google_firestore_v1_Value_reference_value_tag:
+      return GetLowerBound(google_firestore_v1_Value_geo_point_value_tag);
+    case google_firestore_v1_Value_geo_point_value_tag:
+      return GetLowerBound(google_firestore_v1_Value_array_value_tag);
+    case google_firestore_v1_Value_array_value_tag:
+      return GetLowerBound(google_firestore_v1_Value_map_value_tag);
+    case google_firestore_v1_Value_map_value_tag:
+      return MaxValue();
```
schmidt-sebastian, line 587: FYI We can return the underlying proto here as the memory allocation is static and we don't free it.

Firestore/core/src/remote/serializer.cc
```diff
@@ -716,11 +716,14 @@ google_firestore_v1_Target_QueryTarget Serializer::EncodeQueryTarget(
   }
```
schmidt-sebastian, line 716: Reviewed changes till here.

Firestore/core/src/remote/serializer.cc
```diff
@@ -716,11 +716,14 @@ google_firestore_v1_Target_QueryTarget Serializer::EncodeQueryTarget(
...
 
   if (target.start_at()) {
-    result.structured_query.start_at = EncodeBound(*target.start_at());
+    result.structured_query.start_at =
+        EncodeCursor(target.start_at()->position(),
+                     /*before*/ target.start_at()->inclusive());
```
schmidt-sebastian, line 721: This should be `/* before= */` so that the linter will warn us if we do change the variable name. Here and below.

Firestore/core/test/unit/core/target_test.cc
```diff
@@ -0,0 +1,361 @@
...
+
+TEST(TargetTest, FilterWithOrderByQueryBound) {
+  Target target = Query("c")
+                      .AddingFilter(Filter("foo", ">", "bar"))
+                      .AddingOrderBy(OrderBy("foo"))
+                      .ToTarget();
+  FieldIndex index = MakeFieldIndex("c", "foo", Segment::Kind::kAscending);
+
+  auto lower_bound = target.GetLowerBound(index);
+  EXPECT_TRUE(lower_bound.has_value());
```
schmidt-sebastian, line 203: Not needed anymore.

Firestore/core/test/unit/core/target_test.cc
```diff
@@ -0,0 +1,361 @@
...
+  VerifyBound(upper_bound.value(), true, {});
+}
+
+TEST(TargetTest, EqualsQueryBound) {
+  Target target =
+      Query("c").AddingFilter(Filter("foo", "==", "bar")).ToTarget();
+  FieldIndex index = MakeFieldIndex("c", "foo", Segment::Kind::kAscending);
+
+  auto lower_bound = target.GetLowerBound(index);
+  VerifyBound(lower_bound.value(), true, {*Value("bar")});
```
schmidt-sebastian, line 97: This should just be `lower_bound`. Right now, this is converted back to an absl::optional implicitly. Applies everywhere.

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -247,6 +274,129 @@ TEST_F(ValueUtilTest, Ordering) {
...
+  Add(comparison_groups, Map("foo", 1));
+  Add(comparison_groups, Map("foo", 2));
+  Add(comparison_groups, Map("foo", "0"));
+  Add(comparison_groups,
+      GetUpperBound(google_firestore_v1_Value_map_value_tag));
+
+  for (size_t i = 0; i < comparison_groups.size(); ++i) {
+    for (size_t j = i; j < comparison_groups.size(); ++j) {
+      VerifyStrictAscending(comparison_groups[i], comparison_groups[j],
+                            /* is_same_value */ i == j);
```
schmidt-sebastian, line 367: Can we just pass in the expected comparison result (like we did before)? I think it would be cleaner.

Firestore/core/src/core/target.cc
schmidt-sebastian: We can do that - but now we are back in the same situation where we always have to make a copy of the field value when we compute the bound, even if the bound is a value in the Target. Is that correct?

schmidt-sebastian:
Assigning this back to you for now as we should figure out what memory model we should use for in the Bounds calculation. Let me know if it makes sense to discuss "in person".

Firestore/core/src/model/value_util.cc
```diff
@@ -509,51 +509,56 @@ Message<google_firestore_v1_Value> GetLowerBound(pb_size_t value_tag) {
...
+      value.bytes_value = nullptr;
       return value;
     }
 
-    case google_firestore_v1_Value_reference_value_tag:
-      return RefValue(DatabaseId("", ""), DocumentKey::Empty());
+    case google_firestore_v1_Value_reference_value_tag: {
+      google_firestore_v1_Value result;
+      result.which_value_type = google_firestore_v1_Value_reference_value_tag;
+      result.reference_value =
```
schmidt-sebastian, line 536: This is now leaking memory. Please use a static variable.

Firestore/core/src/model/value_util.cc
```diff
@@ -600,35 +605,38 @@ bool Contains(google_firestore_v1_ArrayValue haystack,
...
-  max_value->map_value.fields =
+google_firestore_v1_Value MaxValue() {
+  google_firestore_v1_Value max_value;
+  max_value.which_value_type = google_firestore_v1_Value_map_value_tag;
+  max_value.map_value.fields_count = 1;
+  max_value.map_value.fields =
       nanopb::MakeArray<google_firestore_v1_MapValue_FieldsEntry>(1);
-  max_value->map_value.fields[0].key = nanopb::MakeBytesArray("__type__");
-  max_value->map_value.fields[0].value.which_value_type =
+  max_value.map_value.fields[0].key = nanopb::MakeBytesArray("__type__");
```
schmidt-sebastian, line 636: Hm, this is also leaking memory now. This should also use a static instance.

Firestore/core/src/model/value_util.cc
```diff
@@ -672,7 +680,7 @@ Message<google_firestore_v1_Value> RefValue(
     const model::DocumentKey& document_key) {
   Message<google_firestore_v1_Value> result;
   result->which_value_type = google_firestore_v1_Value_reference_value_tag;
-  result->string_value = nanopb::MakeBytesArray(util::StringFormat(
+  result->reference_value = nanopb::MakeBytesArray(util::StringFormat(
```
schmidt-sebastian, line 683: Loving it!

Firestore/core/src/model/value_util.h
```diff
@@ -121,15 +121,15 @@ bool IsMinValue(const google_firestore_v1_Value& value);
  * Firestore comparison logic always return true for `MaxValue() > v`, for any
  * v users can create, regardless `v`'s type and value.
  */
-nanopb::Message<google_firestore_v1_Value> MaxValue();
+google_firestore_v1_Value MaxValue();
```
schmidt-sebastian, line 124: You should update the comments in this file to clarify who owns the memory.

wu-hui: Done.

Firestore/core/src/model/value_util.cc
wu-hui: Fixed.

Firestore/core/src/model/value_util.cc
wu-hui: Fixed.

Firestore/core/src/model/value_util.cc
```diff
@@ -496,39 +621,68 @@ bool Contains(google_firestore_v1_ArrayValue haystack,
...
-  return value == *MaxValue();
+  if (value.which_value_type != google_firestore_v1_Value_map_value_tag) {
+    return false;
+  }
+
+  if (value.map_value.fields_count != 1) {
+    return false;
+  }
+
+  if (nanopb::MakeStringView(value.map_value.fields->key) !=
```
schmidt-sebastian, line 668: Instead of comparing the string value, you could also compare the pointer address first and only compare the string value if the pointer address is different. Here and below.

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -351,7 +351,7 @@ - (ParsedUpdateData)parsedUpdateData:(id)input {
     auto parsedEntry = [self parseData:entry context:context.ChildContext(idx)];
     if (!parsedEntry) {
       // Just include nulls in the array for fields being replaced with a sentinel.
-      parsedEntry = NullValue();
+      parsedEntry = nanopb::MakeMessage(NullValue());
```
schmidt-sebastian, line 354: You need to clone the proto to take ownership (maybe we should add a nanopb::CloneMessage or something)

Firestore/Source/API/FSTUserDataReader.mm
```diff
@@ -432,7 +432,7 @@ - (void)parseSentinelFieldValue:(FIRFieldValue *)fieldValue context:(ParseContex
 - (Message<google_firestore_v1_Value>)parseScalarValue:(nullable id)input
                                                context:(ParseContext &&)context {
   if (!input || [input isMemberOfClass:[NSNull class]]) {
-    return NullValue();
+    return nanopb::MakeMessage(NullValue());
```
schmidt-sebastian, line 435: Same

Firestore/core/src/bundle/bundle_serializer.cc
```diff
@@ -227,13 +232,17 @@ Filter DecodeUnaryFilter(JsonReader& reader, const json& filter) {
   }
 
   if (op == "IS_NAN") {
-    return FieldFilter::Create(path, Operator::Equal, NaNValue());
+    return FieldFilter::Create(path, Operator::Equal,
+                               nanopb::MakeSharedMessage(NaNValue()));
```
schmidt-sebastian, line 236: Same - Need to clone so that this code does not break in the future.

Firestore/core/test/unit/testutil/testutil.cc
```diff
@@ -311,7 +311,7 @@ core::FieldFilter Filter(absl::string_view key,
 core::FieldFilter Filter(absl::string_view key,
                          absl::string_view op,
                          std::nullptr_t) {
-  return Filter(key, op, NullValue());
+  return Filter(key, op, nanopb::MakeMessage(NullValue()));
```
schmidt-sebastian, line 314: Clone message

Firestore/core/test/unit/testutil/testutil.cc
```diff
@@ -108,7 +108,7 @@ ByteString Bytes(std::initializer_list<uint8_t> octets) {
 }
 
 Message<google_firestore_v1_Value> Value(std::nullptr_t) {
-  return NullValue();
+  return nanopb::MakeMessage(NullValue());
```
schmidt-sebastian, line 111: Clone message

Firestore/core/test/unit/model/value_util_test.cc
```diff
@@ -247,6 +272,131 @@ TEST_F(ValueUtilTest, Ordering) {
   Add(comparison_groups, true);
 
   // numbers
+  Add(comparison_groups, nanopb::MakeMessage(GetLowerBound(
```
schmidt-sebastian, line 275: Technically all callsites to MakeMessage need to make a clone too.

Firestore/core/src/model/value_util.h
```diff
@@ -80,10 +80,20 @@ bool Equals(const google_firestore_v1_ArrayValue& left,
  */
 std::string CanonicalId(const google_firestore_v1_Value& value);
 
-/** Returns the lowest value for the given value type (inclusive). */
+/**
+ * Returns the lowest value for the given value type (inclusive).
+ *
+ * The returned value might point to global variables, and the caller is not
+ * supposed to own or release it, without calling `DeepClone`.
```
schmidt-sebastian, line 87: Maybe: The returned value might point to heap allocated memory that is owned by this function. To take ownership of this memory, invoke `DeepClone`.

schmidt-sebastian (approved):

Firestore/core/src/model/value_util.cc
```diff
@@ -36,6 +36,23 @@
...
 namespace firestore {
 namespace model {
+namespace {
+
+/** The smallest reference value. */
+pb_bytes_array_s* kMinimumReferenceValue =
+    nanopb::MakeBytesArray("projects//databases//documents/");
+
+/** The field type of a maximum proto value. */
+std::string kRawMaxValueFieldKey = "__type__";
```
schmidt-sebastian, line 46: FYI The linter wants this to be a char pointer.

Firestore/core/src/model/value_util.h
wu-hui: Done.

Firestore/core/src/model/value_util.cc
wu-hui: Done.

PR #9521, opened 2022-03-25
Port Storage unit tests to Swift for corresponding Swift ports

paulb777:
- Port unit tests to Swift for the corresponding library ports to Swift - especially the components tests and top-level FIRStorage API
- Fix flaky Functions SPM tests by adding missing dependencies and removing unnecessary `NSObject` from Interop protocols. Without this change, `@testable` was not working to see the expose the internal methods of `FunctionsContextProvider` to the Functions unit tests.

#no-changelog


FirebaseFunctions/Sources/Internal/FunctionsContext.swift
```diff
@@ -35,7 +35,9 @@ internal class FunctionsContextProvider: NSObject {
   private var messaging: MessagingInterop?
   private var appCheck: AppCheckInterop?
 
-  init(auth: AuthInterop?, messaging: MessagingInterop?, appCheck: AppCheckInterop?) {
+  // The 'public' here is weird. It is sometimes needed for `@testable import FirebaseFunctions` to
+  // successfully access this 'init', but it depends upon the build configuration. Xcode 13.2.1.
+  public init(auth: AuthInterop?, messaging: MessagingInterop?, appCheck: AppCheckInterop?) {
```
paulb777, line 40: While putting `public` here fixed it locally for me, it doesn't on CI. It seems like there is a race in the Swift build system and the processing of `testable`.

Any ideas?

FirebaseStorageSwift/Tests/Unit/StorageComponentTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+  // MARK: Interoperability Tests
+
+  /// Tests that the right number of components are being provided for the container.
+  func testComponentsBeingRegistered() throws {
+    let components = StorageComponent.componentsToRegister()
+    XCTAssert(components.count == 1)
+  }
+
+  /// Tests that the right number of components are being provided for the container.
```
ryanwilson, line 45: Nit: comment is wrong

FirebaseStorageSwift/Tests/Unit/StorageComponentTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+class StorageComponentTests: XCTestCase {
+  static var app: FirebaseApp?
+
+  override class func setUp() {
+    let options = FirebaseOptions(googleAppID: "0:0000000000000:ios:0000000000000000",
+                                  gcmSenderID: "00000000000000000-00000000000-000000000")
+    options.projectID = "myProjectID"
+    FirebaseApp.configure(name: "test", options: options)
+    app = FirebaseApp(instanceWithName: "test", options: options)
```
ryanwilson, line 34: Hmm we both call `configure` then `init` using the same name, this shouldn't be the case. I don't think we need to call `configure()`.

FirebaseStorageSwift/Tests/Unit/StorageComponentTests.swift
```diff
@@ -0,0 +1,92 @@
...
+@testable import FirebaseStorage
+
+import FirebaseAppCheckInterop
+import FirebaseAuthInterop
+import SharedTestUtilities
+
+import XCTest
+
+class StorageComponentTests: XCTestCase {
+  static var app: FirebaseApp?
```
ryanwilson, line 27: Since we're configuring this in `setUp`, maybe we force unwrap here and avoid all the force unwraps in the tests?

FirebaseStorageSwift/Tests/Unit/StorageComponentTests.swift
```diff
@@ -0,0 +1,92 @@
...
+    let storage2 = provider.storage(for: "randomBucket")
+    XCTAssertNotNil(storage1)
+
+    // Ensure they're the same instance.
+    XCTAssert(storage1 === storage2)
+
+    let storage3 = provider.storage(for: "differentBucket")
+    XCTAssertNotNil(storage3)
+
+    XCTAssertNotEqual(storage1, storage3)
```
ryanwilson, line 90: Hmm I'm wondering why we can use `NotEqual` here but can't use `Equal` above... does this compare instances under the hood? Or should we use `!==`?

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  func testStorageDefaultApp() throws {
+    let app = try getApp(bucket: "bucket")
+    let storage = Storage.storage(app: app)
+    XCTAssertEqual(storage.app.name, app.name)
+  }
+
+  func testStorageNoBucketInConfig() throws {
+    let app = try getApp(bucket: "nil") // Special case code in getApp for "nil".
```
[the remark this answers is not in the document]
paulb777: Reworked to put special case nil code in test that needs it.

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -0,0 +1,192 @@
...
+    let storage = Storage.storage(app: app)
+    _ = storage.reference(forURL: "gs://benwu-test1.storage.firebase.com/child")
+    _ = storage.reference(forURL: "gs://benwu-test2.storage.firebase.com/child")
+  }
+
+  func testBucketEnforced() throws {
+    let app = try getApp(bucket: "bucket")
+    let storage = Storage.storage(app: app, url: "gs://benwu-test1.storage.firebase.com")
+    _ = storage.reference(forURL: "gs://benwu-test1.storage.firebase.com/child")
+    XCTAssertThrowsError(try ExceptionCatcher.catchException {
```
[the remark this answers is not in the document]
paulb777: Done locally. A future task could move these two a SharedTestUtilsSwift library

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -0,0 +1,192 @@
...
+  func testhash() throws {
+    let app = try getApp(bucket: "bucket")
+    let storage = Storage.storage(app: app)
+    let copy = try XCTUnwrap(storage.copy() as? Storage)
+    XCTAssertEqual(storage.app.name, copy.app.name)
+  }
+
+  // MARK: Private Helpers
+
+  private static var appDictionary: [String: FirebaseApp] = [:]
```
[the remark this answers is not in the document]
paulb777: done

paulb777 (commented):
great review. Thanks!

paulb777:
In addition to addressing the review, I cherry-picked the clang-format update fix from master to help getting to green here.

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -23,8 +23,16 @@ import SharedTestUtilities
...
 import XCTest
 
-@MainActor
 class StorageTests: XCTestCase {
+  static var app: FirebaseApp?
+  override class func setUp() {
+    let options = FirebaseOptions(googleAppID: "0:0000000000000:ios:0000000000000000",
+                                  gcmSenderID: "00000000000000000-00000000000-000000000")
+    options.projectID = "myProjectID"
+    FirebaseApp.configure(name: "test", options: options)
```
ryanwilson, line 32: Same issue as before in the component tests - this calls `configure` but doesn't need to.

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -92,10 +100,14 @@ class StorageTests: XCTestCase {
...
 
   func testStorageNoBucketInConfig() throws {
-    let app = try getApp(bucket: "nil") // Special case code in getApp for "nil".
-    XCTAssertThrowsError(try ExceptionCatcher.catchException {
+    XCTAssertThrowsObjCException {
+      let options = FirebaseOptions(googleAppID: "0:0000000000000:ios:0000000000000000",
+                                    gcmSenderID: "00000000000000000-00000000000-000000000")
+      options.projectID = "myProjectID"
+      let name = "StorageTestsNil"
+      let app = FirebaseApp(instanceWithName: name, options: options)
```
ryanwilson, line 108: Nit: maybe move this outside of the `AssertThrowsObjCException` block to make it clear which part we expect to throw?

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -171,6 +183,7 @@ class StorageTests: XCTestCase {
 
   // MARK: Private Helpers
 
+  // Cache the app's associated with each Storage bucket
```
ryanwilson, line 186: Nit: no apostrophe needed

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
```diff
@@ -181,12 +194,22 @@ class StorageTests: XCTestCase {
...
+    XCTAssertThrowsError(try ExceptionCatcher.catchException {
+      closure()
+    })
+  }
+
+  private func XCTAssertNoThrowObjCException(_ closure: @escaping () -> Void) {
+    XCTAssertNoThrow(try ExceptionCatcher.catchException {
+      closure()
+    })
+  }
```
ryanwilson, line 214: Love how these turned out :D 

Small nit picky question: can we just pass in the closure directly like `ExceptionCatcher.catchException(block: closure)` or will that not work?

FirebaseStorageSwift/Tests/Unit/StorageComponentTests.swift
ryanwilson: Does `static var app: FirebaseApp!`? or `static var app: FirebaseApp! = nil` work, or does it require a `let`?

ryanwilson (approved):
Just a few nits left

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
paulb777: Without it, there's this crash:
Screen Shot 2022-03-30 at 7 01 20 AM

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
paulb777: Done

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
paulb777: Done

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
paulb777: Do you mean 
```
    XCTAssertThrowsError(try ExceptionCatcher.catchException( {
      closure()
    }))
```
Don't the extra parens make it less readable?

ryanwilson: Just checked and we can use this:
```
XCTAssertThrowsError(try ExceptionCatcher.catchException(closure))
```

The idea is that the argument expected in `catchException` is a closure that takes no arguments and has no return type, which is exactly what the `closure` variable we have is. So instead of using the trailing closure syntax to create a new closure that only calls the function `closure()` that we passed in, we can pass it directly as an argument (like we would a selector in ObjC)

FirebaseStorageSwift/Tests/Unit/StorageTests.swift
ryanwilson: Discussed offline, line below needs to be adjusted.

ryanwilson (approved):

PR #9527, opened 2022-03-28
[Infra] Bump clang-format to v14

ncooke3:
Related to clang-format not managing brew versioning properly. See https://github.com/llvm/llvm-project/issues/49340

#no-changelog

### DIFF
--- a/scripts/style.sh
+++ b/scripts/style.sh
@@ -42,7 +42,7 @@ version="${version/ (*)/}"
 version="${version/.*/}"
 
 case "$version" in
-  13)
+  14)
     ;;
   google3-trunk)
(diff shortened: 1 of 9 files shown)


scripts/style.sh
```diff
@@ -42,7 +42,7 @@ version="${version/ (*)/}"
 version="${version/.*/}"
 
 case "$version" in
-  13)
+  14)
```
paulb777, line 45: change line 54

ncooke3: good catch, thanks

ncooke3:
Merging. Failed check was related to build cache

PR #9529, opened 2022-03-28
Fix hard_assert in field_filter.cc

cherylEnkidu:
#no-changelog


ehsannas (approved):
As of now, all filters are FieldFilters, but, _after_ you add the CompositeFilter class, it would be interesting to add a unit test in `field_filter_test.cc` that ensures this assertion does get triggered if a composite filter object is passed to the copy constructor of the FieldFilter class. googletest has [mechanisms](https://google.github.io/googletest/advanced.html#death-tests) for such testing.

No action required now.

PR #9532, opened 2022-03-29
Upgrade grpc to 1.44 for SPM and Cocoapods

wu-hui:


Package.swift
```diff
@@ -178,12 +178,12 @@ let package = Package(
...
       url: "https://github.com/firebase/abseil-cpp-SwiftPM.git",
-      "0.20200225.4" ..< "0.20200226.0"
+      "0.20220203.1" ..< "0.20220204.0"
     ),
     .package(
       name: "gRPC",
-      url: "https://github.com/firebase/grpc-SwiftPM.git",
-      "1.28.4" ..< "1.29.0"
+      url: "https://github.com/grpc/grpc-ios.git",
+      "1.44.0-grpc" ..< "1.44.1-grpc"
```
paulb777, line 186: Should be upper bound of 1.45.0-grpc to be consistent with CocoaPods to allow patch version updates for both

.github/workflows/firestore.yml
```diff
@@ -202,7 +202,8 @@ jobs:
       run: |
         scripts/third_party/travis/retry.sh scripts/pod_lib_lint.rb ${{ matrix.podspec }} \
             --platforms=ios \
-            --allow-warnings
+            --allow-warnings \
+            --no-analyze
```
paulb777, line 206: Please open an issue to reenable analyze

wu-hui: Done.

Package.swift
wu-hui: Done.

dennycd:
@wu-hui  @paulb777  I realize after properly patched firestore side's usage in https://github.com/firebase/firebase-ios-sdk/pull/9539 (by matching to StrSplit's template function that does not use std::move), the error still persist, since the analysis warning also comes from building the Cocoapod abseil framework itself (see https://github.com/firebase/abseil-cpp-SwiftPM/pull/11).   https://github.com/firebase/abseil-cpp-SwiftPM/pull/11 will help resolve SPM side's building warning, but since Cocoapod is not forked from abseil-cpp-SwiftPM, we will need to upstream either a proper fix or lint-surpress  to abseil.   So I think for short term, we may need to temporarily turn off the analysis switch for this one. I will open a issue and follow up from abseil side. lmk

wu-hui:
> @wu-hui @paulb777 I realize after properly patched firestore side's usage in #9539 (by matching to StrSplit's template function that does not use std::move), the error still persist, since the analysis warning also comes from building the Cocoapod abseil framework itself (see [firebase/abseil-cpp-SwiftPM#11](https://github.com/firebase/abseil-cpp-SwiftPM/pull/11)). [firebase/abseil-cpp-SwiftPM#11](https://github.com/firebase/abseil-cpp-SwiftPM/pull/11) will help resolve SPM side's building warning, but since Cocoapod is not forked from abseil-cpp-SwiftPM, we will need to upstream either a proper fix or lint-surpress to abseil. So I think for short term, we may need to temporarily turn off the analysis switch for this one. I will open a issue and follow up from abseil side. lmk

Thank you for your investigation! PR updated now.

Package.swift
paulb777: I don't see it at https://github.com/firebase/firebase-ios-sdk/issues. Users should be able to find the issue if they run the analyzer on their apps that include Firestore.

.github/workflows/firestore.yml
```diff
@@ -199,10 +199,12 @@ jobs:
       run: ./scripts/setup_bundler.sh
 
     - name: Pod lib lint
+      # TODO(b/227461966): Remove --no-analyze when absl is fixed.
```
paulb777, line 202: Should mention a public GitHub issue instead or additionally

paulb777 (approved):

PR #9539, opened 2022-03-30
Fix abseil forward ref warning for StrSplit

dennycd:
Patching StrSplit usage for [bugprone-move-forwarding-reference](https://clang.llvm.org/extra/clang-tidy/checks/bugprone-move-forwarding-reference.html) warning as seem from https://github.com/firebase/firebase-ios-sdk/pull/9486.  StrSplit requires a forwarding reference (rvalue ref inferred from a template param) to be passed into its first parameter, and we are currently passing a lvalue ref instead.


paulb777:
I believe this has been resolved. Closing the PR. Please reopen if still needed.

PR #9540, opened 2022-03-30
Enable async/await APIs for iOS 13, etc.

### LINKED ISSUE #9483: Update async/await code to be available for iOS 13, etc.
Now that it is supported for iOS 13 instead of 15 starting with Xcode 13.2.1

ryanwilson:
This bumps the required compiler version to 5.5.2 which was included in
Xcode 13.2, where backporting was first introduced.

This breaks less developers and enables developers who aren't using
these APIs to still compile using Xcode 13.0 and 13.1.

As far as I can tell, the only breakage this includes now is developers
who are using Xcode 13.0 or 13.1, targeting iOS 15 and using one of
these 7 APIs. It will result in a compiler error and they'll have to
either use the completion handler version or upgrade to Xcode 13.2+ to
fix the compilation error.

Fixes #9483.

### DIFF
--- a/FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
+++ b/FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
@@ -112,7 +112,7 @@ final class DatabaseAPITests {
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     }
 
-    #if compiler(>=5.5) && canImport(_Concurrency)
+    #if compiler(>=5.5.2) && canImport(_Concurrency)
       if #available(iOS 15.0, macOS 12.0, macCatalyst 15.0, tvOS 15.0, *) {
         // async/await is a Swift 5.5+ feature available on iOS 15+
         Task {
@@ -123,7 +123,7 @@ final class DatabaseAPITests {
           }
         }
       }
-    #endif // compiler(>=5.5) && canImport(_Concurrency)
+    #endif // compiler(>=5.5.2) && canImport(_Concurrency)
 
     // Observe Single Event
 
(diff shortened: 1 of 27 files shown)


FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -112,7 +112,7 @@ final class DatabaseAPITests {
       let /* dataSnapshot */ _: DataSnapshot = dataSnapshot
     }
 
-    #if compiler(>=5.5) && canImport(_Concurrency)
+    #if compiler(>=5.5.2) && canImport(_Concurrency)
       if #available(iOS 15.0, macOS 12.0, macCatalyst 15.0, tvOS 15.0, *) {
```
paulb777, line 116: We're inconsistent with watchOS versions.  The Package.swift requires 7.0. Most, but not all of the podspecs are 6. Perhaps, we don't need to specify it anywhere?  Or we should be consistent with 6 or 7?

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
```diff
@@ -138,7 +138,7 @@ final class DatabaseAPITests {
       let /* optionalString */ _: String? = optionalString
     }
 
-    #if compiler(>=5.5) && canImport(_Concurrency)
+    #if compiler(>=5.5.2) && canImport(_Concurrency)
       if #available(iOS 15.0, macOS 12.0, macCatalyst 15.0, tvOS 15.0, *) {
```
paulb777, line 142: 13.0 etc here and below

FirebaseDatabase/Tests/Unit/Swift/DatabaseAPITests.swift
ryanwilson: I'm fine to be consistent with 7. By not mentioning it, it is included for all version (`*`) which is incorrect but I'm guessing we just didn't see it since we didn't run the tests on watchOS where this didn't work. I'll update them.

ryanwilson: Done in the latest commit.

paulb777 (approved):

ryanwilson:
Messaging tests pass locally for me, seems like a networking error that I'm going to merge but can keep an eye out for continuous failures. 

Firestore changes are fixed in the branch but no need to rebase. Merging now.

PR #9543, opened 2022-03-30
[Heartbeat Logging] Migrate Installations from using `FIRHeartbeatInfo` to `FIRHeartbeatLogger`

ncooke3:
#no-changelog

### DIFF
--- a/.github/workflows/installations.yml
+++ b/.github/workflows/installations.yml
@@ -22,12 +22,11 @@ jobs:
     runs-on: macos-11
     env:
       plist_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+      # TODO(v9): Remove `POD_LIB_LINT_ONLY` after publishing `HeartbeatLoggingTestUtils`.
+      POD_LIB_LINT_ONLY: 1
     strategy:
       matrix:
         target: [ios, tvos, macos]
-    # TODO(v9): Remove `env` after publishing `HeartbeatLoggingTestUtils`.
-    env:
-      POD_LIB_LINT_ONLY: 1
     steps:
     - uses: actions/checkout@v2
     - name: Setup Bundler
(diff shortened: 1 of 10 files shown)


.github/workflows/installations.yml
```diff
@@ -22,6 +22,8 @@ jobs:
     runs-on: macos-11
     env:
       plist_secret: ${{ secrets.GHASecretsGPGPassphrase1 }}
+      # TODO(v9): Remove `POD_LIB_LINT_ONLY` after publishing `HeartbeatLoggingTestUtils`.
```
paulb777, line 25: I'm not sure we want to publish `HeartbeatLoggingTestUtils`. The test specs don't need to be runnable from `pod spec lint`.

paulb777 (approved):
What is the integration test plan?

.github/workflows/installations.yml
ncooke3: Ahh I see, thanks. So `HeartbeatLoggingTestUtils` would only be linted on CI when we're pulling in local sources. 

I added this TODO in a bunch of places so I can go back and revert those TODOs. cc: @ryanwilson

paulb777: Right. SGTM

ryanwilson: Sorry for not catching that, thanks!

ncooke3:
> What is the integration test plan?

@paulb777, thanks for calling that out. This was actually quite tricky, but I added verification that the heartbeat integration is working as expected in [a40e06a](https://github.com/firebase/firebase-ios-sdk/pull/9543/commits/a40e06ad540774f21751935aa0891c9c50156caa).

.github/workflows/installations.yml
ncooke3: Cleaned these up in https://github.com/firebase/firebase-ios-sdk/pull/9554

paulb777 (approved):
Thanks!

PR #9544, opened 2022-03-31
Update Firestore's public headers to use Swift class names.

ehsannas:
#no-changelog

### DIFF
--- a/Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
+++ b/Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
@@ -59,15 +59,15 @@ NS_SWIFT_NAME(Timestamp)
 /** Creates a new timestamp with the current date / time. */
 + (instancetype)timestamp;
 
-/** Returns a new NSDate corresponding to this timestamp. This may lose precision. */
+/** Returns a new `Date` corresponding to this timestamp. This may lose precision. */
 - (NSDate *)dateValue;
 
 /**
  * Returns the result of comparing the receiver with another timestamp.
  * @param other the other timestamp to compare.
- * @return NSOrderedAscending if `other` is chronologically following self,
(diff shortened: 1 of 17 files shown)


ehsannas:
#no-changelog

ehsannas:
@ryanwilson  Not sure if it makes sense to also use `Dictionary` and `String` instead of `NSDictionary` and `NSString` in the documentation?

Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
```diff
@@ -59,15 +59,15 @@ NS_SWIFT_NAME(Timestamp)
...
 
 /**
  * Returns the result of comparing the receiver with another timestamp.
  * @param other the other timestamp to compare.
- * @return NSOrderedAscending if `other` is chronologically following self,
- *     NSOrderedDescending if `other` is chronologically preceding self,
- *     NSOrderedSame otherwise.
+ * @return `NSOrderedAscending` if `other` is chronologically following self,
+ *     `NSOrderedDescending` if `other` is chronologically preceding self,
+ *     `NSOrderedSame` otherwise.
```
ryanwilson, line 70: Looks like the Swift names are different: https://developer.apple.com/documentation/foundation/comparisonresult

ryanwilson:
> @ryanwilson Not sure if it makes sense to also use `Dictionary` and `String` instead of `NSDictionary` and `NSString` in the documentation?

Yep, `Dictionary`, `String`, and `Array` all make sense as well in the docs. Thanks a ton for doing this!

Firestore/Source/Public/FirebaseFirestore/FIRTimestamp.h
ehsannas: oops. Done.

ryanwilson (approved):
Awesome, thanks!

PR #9552, opened 2022-03-31
Make AppDistro releaseNotes nullable.

ryanwilson:
These were unintentionally marked as non-null, but should be nullable.

Fixes #8602

PR #9553, opened 2022-04-01
Fix perf test regression from recent RC change

paulb777:
Fix a regression test failure introduced by the new exception introduced in #9552 

It preserves the previous (probably unintended behavior) of passing nil to `[[FPRRemoteConfigFlags alloc] initWithRemoteConfig:`

Now an unconfigured initWithRemoteConfig: is throwing instead of returning nil.


ryanwilson (approved):
Thanks for catching this!

PR #9555, opened 2022-04-01
[Heartbeat Logging] Migrate Firestore from using `FIRHeartbeatInfo` to `FIRHeartbeatLogger`

ncooke3:
Firestore doesn't have any tests right now to verify its integration with `FIRHeartbeatInfo`. 

At this time, the Firestore backend will not be migrated to accept the v2 heartbeats payload. Because of this, it will use the  direct replacement of `+[FIRHeartbeatInfo heartbeatCodeforTag:]`, which is `-[FIRHeartbeatLogger heartbeatCodeforToday]`. The string "Tag" is no longer needed because the concept of an ["SDK-only heartbeat"](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Sources/Private/FIRHeartbeatInfo.h#L24) is going away.  Additional context: https://github.com/firebase/firebase-ios-sdk/pull/9506#discussion_r836560426

I manually ran the tests and–for the tests that did touch the affected code paths in this PR– `FIRHeartbeatInfoNone` is always returned.

#no-changelog

### DIFF
--- a/Firestore/core/src/remote/firebase_metadata_provider_apple.mm
+++ b/Firestore/core/src/remote/firebase_metadata_provider_apple.mm
@@ -53,9 +53,9 @@ FirebaseMetadataProviderApple::FirebaseMetadataProviderApple(FIRApp* app)
 void FirebaseMetadataProviderApple::UpdateMetadata(
     grpc::ClientContext& context) {
   FIRHeartbeatInfoCode heartbeat = GetHeartbeat(app_);
-  // TODO: If support for notifying a heartbeat logger when a request
-  // fails is implemented, we will need to change the below code to place
-  // the heartbeat data back into heartbeat storage.
+  // TODO(ncooke3): If support for notifying a heartbeat logger when a
+  // request fails is implemented, we will need to change the below
+  // code to place the heartbeat data back into heartbeat storage.
   if (heartbeat != FIRHeartbeatInfoCodeNone) {
     context.AddMetadata(kXFirebaseClientLogTypeHeader,
                         std::to_string(heartbeat));


Firestore/core/src/remote/firebase_metadata_provider_apple.mm
```diff
@@ -51,7 +52,7 @@ FIRHeartbeatInfoCode GetHeartbeat() {
 
 void FirebaseMetadataProviderApple::UpdateMetadata(
     grpc::ClientContext& context) {
-  FIRHeartbeatInfoCode heartbeat = GetHeartbeat();
+  FIRHeartbeatInfoCode heartbeat = GetHeartbeat(app_);
```
paulb777, line 55: Is the comment below affected?

paulb777 (approved):

Firestore/core/src/remote/firebase_metadata_provider_apple.mm
ncooke3: Good question. I'm not sure I understand the context behind the comment because the current code looks to be optimized to _not_ include `FIRHeartbeatInfoCodeNone` in request when the code is `FIRHeartbeatInfoCodeNone`. 

But I dont understand why this matters:
```
 // This should only be changed once it's possible to notify the heartbeat that the 
 // previous attempt to send it has failed. 
```

I'm leaning towards removing the comment. 

@schmidt-sebastian, do you have any context here?

https://github.com/firebase/firebase-ios-sdk/blob/0386388084f1851321ba29bf7d6b3c505bb5f489/Firestore/core/src/remote/firebase_metadata_provider_apple.mm#L55-L62

schmidt-sebastian: I suspect what the comment is after is that we do want to re-send the heartbeat if the previous attempt failed. It seems like we now assume that any heartbeat attempt succeeded. Once we know that the attempt failed we can send it again, but I am not sure if this code exists yet.

ncooke3: I see, this retry behavior is not yet implemented and there are no plans to as of right now. 

I think I will remove the comment because if retry behavior is added, all integrators will be updated accordingly.

schmidt-sebastian: The comment is saying "If we implement this behavior we need to fix this line of code" so it is a good reminder that there is code here that needs to be changed. It would however helped if this wasn't a TODO for an ex-teammember anymore.

ncooke3: SGTM, Ill adjust accordingly.

ncooke3: Reworded in [721f74d](https://github.com/firebase/firebase-ios-sdk/pull/9555/commits/721f74d5386b1bde21f988270375a4a8e331a576)

PR #9557, opened 2022-04-01
Update Storage Quickstart testing dir.

granluo:


paulb777:
See https://github.com/firebase/quickstart-ios/pull/1367

scripts/setup_quickstart.sh
```diff
@@ -75,6 +75,6 @@ if check_secrets || [[ ${SAMPLE} == "installations" ]]; then
   fi
 
   # Add GoogleService-Info.plist to Xcode project
-  ruby ../scripts/info_script.rb "${SAMPLE}" "${LEGACY:-}"
+  # ruby ../scripts/info_script.rb "${SAMPLE}" "${LEGACY:-}"
```
paulb777, line 78: This makes sense to me - but why is it only a problem for Storage?

Looks like all of lines 72-79 can be now deleted.

paulb777 (approved):

scripts/setup_quickstart.sh
granluo: The diff I can find between storage and other apps is the GoogleService-Info.plist for storage is under  `Copy Bundle Resources` of `Build Phases` while others, e.g. crashlytics and functions, are in `Compile Sources`. I tired to move the plist to `Compile Sources` but the [issue](https://github.com/firebase/firebase-ios-sdk/runs/5801793775?check_suite_focus=true) still happens,  though with fewer errors comparing to putting info plist to `Copy Bundle Resources`, to storage after `info_script.rb`.

granluo: I can create another branch to investigate see what the `project.pbxproj` looks like to investigate why it happens to storage, but by any means, since the quickstart has the plist added to targets by default, we might not need to link it again.

PR #9561, opened 2022-04-03
SDK support for logical termination in Firestore BatchGetDocuments

cherylEnkidu:
#no-changelog

For BatchGetDocuments, since the SDK knows the number of document paths in the requests, as soon as the SDK receives all the responses needed, it will fire user callback to save user's waiting time.

### DIFF
--- a/Firestore/core/src/remote/datastore.cc
+++ b/Firestore/core/src/remote/datastore.cc
@@ -233,8 +233,8 @@ void Datastore::LookupDocumentsWithCredentials(
   active_calls_.push_back(std::move(call_owning));
 
   // TODO(c++14): lambda captures using move.
-  auto docs_callback =
-      [this, user_callback](const std::vector<grpc::ByteBuffer> result) {
+  auto messages_callback =
+      [this, user_callback](const std::vector<grpc::ByteBuffer>& result) {
         user_callback(datastore_serializer_.MergeLookupResponses(result));
       };
 
@@ -251,7 +251,7 @@ void Datastore::LookupDocumentsWithCredentials(
     RemoveGrpcCall(call);
   };
 
-  call->Start(keys.size(), docs_callback, close_callback);
+  call->Start(keys.size(), messages_callback, close_callback);
 }
 
 void Datastore::ResumeRpcWithCredentials(const OnCredentials& on_credentials) {
--- a/Firestore/core/src/remote/datastore.cc
+++ b/Firestore/core/src/remote/datastore.cc
@@ -233,7 +233,7 @@ void Datastore::LookupDocumentsWithCredentials(
   active_calls_.push_back(std::move(call_owning));
 
   // TODO(c++14): lambda captures using move.
-  auto messages_callback =
+  auto responses_callback =
       [this, user_callback](const std::vector<grpc::ByteBuffer>& result) {
         user_callback(datastore_serializer_.MergeLookupResponses(result));
       };
@@ -251,7 +251,7 @@ void Datastore::LookupDocumentsWithCredentials(
     RemoveGrpcCall(call);
   };
 
-  call->Start(keys.size(), messages_callback, close_callback);
+  call->Start(keys.size(), responses_callback, close_callback);
 }
 
 void Datastore::ResumeRpcWithCredentials(const OnCredentials& on_credentials) {
--- a/Firestore/core/src/remote/datastore.cc
+++ b/Firestore/core/src/remote/datastore.cc
@@ -204,17 +204,17 @@ void Datastore::CommitMutationsWithCredentials(
 }
 
 void Datastore::LookupDocuments(const std::vector<DocumentKey>& keys,
-                                LookupCallback&& callback) {
+                                LookupCallback&& user_callback) {
   ResumeRpcWithCredentials(
       // TODO(c++14): move into lambda.
-      [this, keys, callback](const StatusOr<AuthToken>& auth_token,
-                             const std::string& app_check_token) mutable {
+      [this, keys, user_callback](const StatusOr<AuthToken>& auth_token,
+                                  const std::string& app_check_token) mutable {
         if (!auth_token.ok()) {
-          callback(auth_token.status());
+          user_callback(auth_token.status());
           return;
         }
         LookupDocumentsWithCredentials(auth_token.ValueOrDie(), app_check_token,
-                                       keys, std::move(callback));
+                                       keys, std::move(user_callback));
       });
 }
 
@@ -222,7 +222,7 @@ void Datastore::LookupDocumentsWithCredentials(
     const credentials::AuthToken& auth_token,
     const std::string& app_check_token,
     const std::vector<DocumentKey>& keys,
-    LookupCallback&& callback) {
+    LookupCallback&& user_callback) {
   grpc::ByteBuffer message =
       MakeByteBuffer(datastore_serializer_.EncodeLookupRequest(keys));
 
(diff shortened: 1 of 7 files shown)


Firestore/core/test/unit/remote/grpc_connection_test.cc
```diff
@@ -162,22 +165,31 @@ TEST_F(GrpcConnectionTest, ConnectivityChangeWithSeveralActiveCalls) {
   int changes_count = 0;
 
   std::unique_ptr<GrpcStreamingReader> foo = tester.CreateStreamingReader();
-  foo->Start([&](const StatusOr<std::vector<grpc::ByteBuffer>>&) {
-    foo.reset();
-    ++changes_count;
```
cherylEnkidu, line 167: Even though all the unit and integration tests pass, I got an use after memory free [sanitizers (asan)](https://github.com/firebase/firebase-ios-sdk/runs/5807500775?check_suite_focus=true#logs) error at `++changes_count;` It was fixed by moving `foo.reset();` to the last line of callback, which has similar behaviour as `close_callback` in `datastore.cc`. But what I don’t understand is why this test wasn't failed in the previous implementation :( @schmidt-sebastian @ehsannas

schmidt-sebastian: Just a wild theory - maybe it is because you are not passing the responses by const value and they get cleaned up when we free the memory owned by `foo`?

schmidt-sebastian: Just a wild theory - maybe it is because you are not passing the responses by const value and they get cleaned up when we free the memory owned by `foo`?

Firestore/core/src/remote/grpc_streaming_reader.h
```diff
@@ -60,7 +62,9 @@ class GrpcStreamingReader : public GrpcCall, public GrpcStreamObserver {
    * results of the call. If the call fails, the `callback` will be invoked with
    * a non-ok status.
    */
-  void Start(Callback&& callback);
+  void Start(size_t key_count,
```
schmidt-sebastian, line 65: Can you update the comment? We should also not use the term `key` but rather something like`expected_response_count`.

schmidt-sebastian, line 65: Can you update the comment? We should also not use the term `key` but rather something like`expected_response_count`.

schmidt-sebastian (commented):
Thank you! This is great, I just have some cleanup suggestions.

Firestore/core/src/remote/datastore.cc
```diff
@@ -232,28 +232,26 @@ void Datastore::LookupDocumentsWithCredentials(
...
-
+  // TODO(c++14): lambda captures using move.
+  auto docs_callback =
+      [this, user_callback](const std::vector<grpc::ByteBuffer> result) {
+        user_callback(datastore_serializer_.MergeLookupResponses(result));
+      };
+
+  auto close_callback = [this, user_callback, call](const util::Status& status,
+                                                    bool callback_fired) {
+    // Trigger user_callback with an error status
```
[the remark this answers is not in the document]
cherylEnkidu: If we look up 0 documents, it will still call `docs_callback_({})` from grpc_streaming_reader.cc line 88. So this block will only be triggered with an error status.

Firestore/core/test/unit/remote/grpc_connection_test.cc
cherylEnkidu: If you refer to the response in `messagesCallback`, I made it pass by const reference in the latest commit and there is still a memory use after free after I change back two statement position (the reason why the PR is failed). On the other hand I don’t think it relates to `messagesCallback`, which I add it for the new implementation. And for `closeCallback`, both new and old implementations are passed by const reference.

ehsannas: yeah I'd imagine after calling `foo.reset()`, the parameters captured in the lambda (`changes_count`) should not be accessed. So I'm amused why the ASAN bot was not reporting this test in the past 🤷

Firestore/core/test/unit/remote/grpc_connection_test.cc
```diff
@@ -125,8 +126,9 @@ TEST_F(GrpcConnectionTest, GrpcStreamingCallsNoticeChangeInConnectivity) {
   std::unique_ptr<GrpcStreamingReader> streaming_call =
       tester.CreateStreamingReader();
   streaming_call->Start(
-      [&](const StatusOr<std::vector<grpc::ByteBuffer>>& result) {
-        if (IsConnectivityChange(result.status())) {
+      static_cast<size_t>(0), [&](const std::vector<ResponsesT>) {},
```
ehsannas, line 129: There are a lot of `static_cast<size_t>(...)` instances in the PR (for both `Start` and `StartReader`). Are you sure these are necessary? I thought the compiler should be able to implicitly cast a literal number argument to `size_t`.

cherylEnkidu: I think the compiler will do the implicitly cast, but I want to make the code less dependents on compiler behaviour :( Is it really a bad idea to increase code size a little bit?

ehsannas: These casts do not increase code size (they get optimized away by the compiler). It's more about readability. I think the code becomes much easier to read without the `static_cast`s. In this case I think all compilers should behave the same.

Firestore/core/src/remote/datastore.cc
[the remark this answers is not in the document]
schmidt-sebastian: Got it. I missed that subtlety.

Firestore/core/src/remote/grpc_streaming_reader.h
schmidt-sebastian: This is not yet fixed

Firestore/core/test/unit/remote/grpc_connection_test.cc
schmidt-sebastian: Did you flip the order again? I think the oder you proposed was actually better.

Firestore/core/test/unit/remote/grpc_streaming_reader_test.cc
```diff
@@ -73,15 +74,14 @@ class GrpcStreamingReaderTest : public testing::Test {
     tester.KeepPollingGrpcQueue();
   }
 
-  void StartReader() {
+  void StartReader(size_t key_count) {
```
schmidt-sebastian, line 77: This should also not use the term `key`.

Firestore/core/test/unit/remote/grpc_connection_test.cc
cherylEnkidu: Just flip it to show you what the "memory use after free" error looks like. Will revert the commit.

Firestore/core/src/remote/grpc_streaming_reader.h
```diff
@@ -99,8 +103,11 @@ class GrpcStreamingReader : public GrpcCall, public GrpcStreamObserver {
   std::unique_ptr<GrpcStream> stream_;
   grpc::ByteBuffer request_;
 
-  Callback callback_;
-  ResponsesT responses_;
+  size_t expected_response_count_;
+  bool callback_fired_ = false;
+  MessagesCallback docs_callback_;
```
schmidt-sebastian, line 108: Nit: Can we make this (and all references to message) "ResponsesCallback" (and rename all references to message). We are mixing the terms messages and responses freely, when I think they are all the same. The instance variable also still needs to renaming (from docs_callback_).

Firestore/core/src/remote/grpc_streaming_reader.h
```diff
@@ -60,7 +62,9 @@ class GrpcStreamingReader : public GrpcCall, public GrpcStreamObserver {
    * results of the call. If the call fails, the `callback` will be invoked with
    * a non-ok status.
    */
-  void Start(Callback&& callback);
+  void Start(size_t expected_response_count,
+             MessagesCallback&& messageCallback,
+             CloseCallback&& closeCallback);
```
schmidt-sebastian, line 67: Both of these should be camel_case.

Firestore/core/src/remote/grpc_streaming_reader.cc
```diff
@@ -45,8 +45,12 @@ GrpcStreamingReader::GrpcStreamingReader(
       request_{request} {
 }
 
-void GrpcStreamingReader::Start(Callback&& callback) {
-  callback_ = std::move(callback);
+void GrpcStreamingReader::Start(size_t expected_response_count,
+                                MessagesCallback&& docs_callback,
```
schmidt-sebastian, line 49: s/docs/responses (or messages)

schmidt-sebastian (approved):
Thanks! Please address the remaining feedback before merging.

Firestore/core/src/remote/grpc_streaming_reader.h
cherylEnkidu: My apologies, I should change it together with the function type

PR #9562, opened 2022-04-04
Register Combine modules with Swift component system

peterfriese:
This PR adds a Swift component for Firebase Combine, so that Combine will be included in the user agent.

This is a replacement for PR #9363 and contains all code already discussed there, and addresses the most recent comments/

#no-changelog

### DIFF
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -17,6 +17,8 @@
   import Combine
   import FirebaseAuth
 
+  @objc(FIRCombineAuthComponent) private class __CombineAuthComponent: NSObject {}
+
   @available(swift 5.0)
   @available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)
   public extension Auth {
--- a/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
+++ b/FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
@@ -17,7 +17,8 @@
   import Combine
   import FirebaseAuth
 
-  @objc(FIRCombineAuthComponent) private class __CombineAuthComponent: NSObject {}
+  // Make this class discoverable from Objective-C. Don't instantiate directly.
+  @objc(FIRCombineAuthLibrary) private class __CombineAuthLibrary: NSObject {}
 
   @available(swift 5.0)
   @available(iOS 13.0, macOS 10.15, macCatalyst 13.0, tvOS 13.0, watchOS 6.0, *)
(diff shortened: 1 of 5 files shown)


FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
```diff
@@ -17,6 +17,8 @@
   import Combine
   import FirebaseAuth
 
+  @objc(FIRCombineAuthComponent) private class __CombineAuthComponent: NSObject {}
```
ryanwilson, line 20: Can you add a comment above just mentioning this is exclusively used to detect the class from ObjC and shouldn't be used directly?

FirebaseCombineSwift/Sources/Storage/Storage+Combine.swift
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#if canImport(Combine) && swift(>=5.0)
+
+  import Foundation
+  @objc(FIRCombineStorageComponent) private class __CombineStorageComponent: NSObject {}
```
ryanwilson, line 18: And here

FirebaseCore/Sources/FIRApp.m
```diff
@@ -838,11 +838,25 @@ + (void)registerSwiftComponents {
...
+      [FIRApp registerInternalLibrary:klass withName:swiftComponents[className]];
+    }
+  }
+
+  // Swift libraries that don't need component behaviour
+  NSDictionary<NSString *, NSString *> *swiftLibraries = @{
+    @"FIRCombineAuthComponent" : @"comb-auth",
+    @"FIRCombineFirestoreComponent" : @"comb-firestore",
+    @"FIRCombineFunctionsComponent" : @"comb-functions",
+    @"FIRCombineStorageComponent" : @"comb-storage",
```
ryanwilson, line 854: Given we're using the `component` name for components being registered above, and `libraries` for non-components that we just want to keep track of, what do you think about changing the `Component` suffix with `Library`?

ryanwilson (commented):
Just a couple nits then LGTM, thanks!

FirebaseCombineSwift/Sources/Auth/Auth+Combine.swift
peterfriese: LMK if this is what you had in mind:

```
// Make this class discoverable from Objective-C. Don't instantiate directly.
```

ryanwilson (approved):
Thanks!

PR #9568, opened 2022-04-04
[Heartbeat Logging] Migrate Messaging from using `FIRHeartbeatInfo` to `FIRHeartbeatLogger`

ncooke3:
#no-changelog

### DIFF
--- a/FirebaseMessaging/Sources/FIRMessaging.m
+++ b/FirebaseMessaging/Sources/FIRMessaging.m
@@ -104,7 +104,8 @@ BOOL FIRMessagingIsContextManagerMessage(NSDictionary *message) {
 // FIRApp properties
 @property(nonatomic, readwrite, strong) NSData *apnsTokenData;
 @property(nonatomic, readwrite, strong) FIRMessagingClient *client;
-@property(nonatomic, readwrite, strong) GULReachabilityChecker *reachability;
+
+@property(nonatomic, readwrite, strong, nullable) GULReachabilityChecker *reachability;
 @property(nonatomic, readwrite, strong) FIRMessagingPubSub *pubsub;
 @property(nonatomic, readwrite, strong) FIRMessagingRmqManager *rmq2Manager;
 @property(nonatomic, readwrite, strong) FIRMessagingSyncMessageManager *syncMessageManager;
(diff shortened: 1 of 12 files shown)


ncooke3:
I had to re-run the `messaging` CI a few times because the integration tests were flaky. To check if this PR's changes were the cause, I ran triggered the messaging CI on `master` (#9579) and they flaked there as well.

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -143,14 +144,16 @@ + (FIRMessagingExtensionHelper *)extensionHelper {
 }
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-- (instancetype)initWithAnalytics:(nullable id<FIRAnalyticsInterop>)analytics
-                 withUserDefaults:(GULUserDefaults *)defaults {
+- (instancetype)initForApp:(FIRApp *)app
```
ryanwilson, line 147: Instead of injecting the app here - it's likely better / less setup for tests to pass in the `heartbeatLogger` directly. Although seems like this API isn't used in tests since no tests changed that depend on it?

ryanwilson: Also, `initWith` is preferred over `initFor`

FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -143,14 +144,16 @@ + (FIRMessagingExtensionHelper *)extensionHelper {
...
+             withAnalytics:(nullable id<FIRAnalyticsInterop>)analytics
```
ryanwilson, line 148: `with` isn't necessary on the second argument if it's used for the first

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
```diff
@@ -67,6 +69,39 @@ - (void)performTokenOperation {
...
+
+@implementation FIRHeartbeatLoggerFake
+
+- (nonnull FIRHeartbeatsPayload *)flushHeartbeatsIntoPayload {
+  // This API should not be used by the below tests because the Messaging
+  // SDK uses only the V1 heartbeat API (`heartbeatCodeForToday`) for
+  // getting a single heartbeat.
+  [self doesNotRecognizeSelector:_cmd];
+  return nil;
+}
```
ryanwilson, line 85: Just noticed this pattern across multiple SDKs - a potential improvement for the future would be to mark this and the `log` API as `@optional` in the Protocol (possible in ObjC but not in Swift). You'd need to check if the instance responds to it in the implementation but would reduce the amount of manual code you have to write and extra messaging.

FirebaseMessaging/Sources/FIRMessaging.m
ncooke3: > @ryanwilson: Instead of injecting the app here - it's likely better / less setup for tests to pass in the heartbeatLogger directly. Although seems like this API isn't used in tests since no tests changed that depend on it?

Correct. But I can change the initializer to `initWithHeartbeatLogger:analytics:userDefaults:` anyhow.

ncooke3: Reworked in a3dc31a

FirebaseMessaging/Sources/FIRMessaging.m
ncooke3: Reworked in a3dc31a

FirebaseMessaging/Tests/UnitTests/FIRMessagingTokenOperationsTest.m
ncooke3: Thanks! Added a TODO to revisit (a3dc31a).

ryanwilson (approved):

PR #9569, opened 2022-04-04
Functions and Storage deprecated Error variables

paulb777:
Deprecate Error variables in advance of Swift-first implementations that do not support globals.

API review internally at [go/storage-function-error-variable-deprecations](http://go/storage-function-error-variable-deprecations)


sipersso:
I was confused when I saw this after upgrading. So FunctionsErrorDetailsKey is deprecated, but what I am supposed to replace it with? There is no mention in the deprecation message, just that it will go away.

sipersso:
Thanks for the clarification @paulb777

PR #9573, opened 2022-04-05
[Minor] Fixed a typo in The InAppMessaging Swift test

ptxmac:
Without this change the dismess button was never added, creating a
modal dialog that cannot be closed if the message have no action

### DIFF
--- a/FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
+++ b/FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
@@ -62,7 +62,7 @@ struct ModalInAppMessageView: View {
   @ViewBuilder
   func dismissButton(modalMessage: InAppMessagingModalDisplay,
                      delegate: InAppMessagingDisplayDelegate) -> some View {
-    if let _ = modalMessage.actionButton, modalMessage.actionURL != nil {
+    if let _ = modalMessage.actionButton, modalMessage.actionURL == nil {
       Button(action: {
         delegate.messageDismissed?(modalMessage, dismissType: .typeUserTapClose)
       }) {
--- a/FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
+++ b/FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
@@ -58,7 +58,8 @@ struct ModalInAppMessageView: View {
   }
 
   // Need a dismiss button for the case where there's an action button with an action URL. Otherwise
-  // user would be forced into a clickthrough.
+  // user would be forced into a clickthrough. If there's No action button at all we also need a
+  // dismissButton, otherwise the user will not be able to close the modal!
   @ViewBuilder
   func dismissButton(modalMessage: InAppMessagingModalDisplay,
                      delegate: InAppMessagingDisplayDelegate) -> some View {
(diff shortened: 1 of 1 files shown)


ptxmac:
@christibbs I guess this is your area

paulb777:
@ptxmac Thanks for the PR! Would you describe the issue that the PR addresses and why this PR solves it?

FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
```diff
@@ -62,7 +62,7 @@ struct ModalInAppMessageView: View {
   @ViewBuilder
   func dismissButton(modalMessage: InAppMessagingModalDisplay,
                      delegate: InAppMessagingDisplayDelegate) -> some View {
-    if let _ = modalMessage.actionButton, modalMessage.actionURL != nil {
+    if let _ = modalMessage.actionButton, modalMessage.actionURL == nil {
```
morganchen12, line 65: nit: Though you didn't originally write the code this way, can you fix the check here so it's stylistically consistent?
```swift
if modalMessage.actionButton != nil && modalMessage.actionURL == nil
```

morganchen12 (approved):
Thanks!

ptxmac:
rereading the original code and comments, I'm actually not sure if my fix makes sense. I think perhaps the original test was never written with button-less in-app modals, thus there's a 3rd code path that has never been added.

The original code is "correct" under the assumption that all in-app messages have an action URL

FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
ptxmac: I've changed the style, and also made it a little more correct

FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
```diff
@@ -58,11 +58,12 @@ struct ModalInAppMessageView: View {
...
 
   // Need a dismiss button for the case where there's an action button with an action URL. Otherwise
-  // user would be forced into a clickthrough.
+  // user would be forced into a clickthrough. If there's No action button at all we also need a
+  // dismissButton, otherwise the user will not be able to close the modal!
   @ViewBuilder
   func dismissButton(modalMessage: InAppMessagingModalDisplay,
                      delegate: InAppMessagingDisplayDelegate) -> some View {
-    if let _ = modalMessage.actionButton, modalMessage.actionURL != nil {
+    if (modalMessage.actionButton != nil && modalMessage.actionURL != nil) || modelMesssage.actionButton == nil {
```
morganchen12, line 66: There's a typo in the last `modalMessage`.

morganchen12 (commented):
Can you add a test for the new behavior?

FirebaseInAppMessaging/Swift/Tests/Integration/FIAMSwiftUI/Shared/ModalInAppMessageView.swift
```diff
@@ -58,11 +58,12 @@ struct ModalInAppMessageView: View {
   }
 
   // Need a dismiss button for the case where there's an action button with an action URL. Otherwise
-  // user would be forced into a clickthrough.
+  // user would be forced into a clickthrough. If there's No action button at all we also need a
```
ncooke3, line 61: nit: I think "No" should be lowercase.
```suggestion
  // user would be forced into a clickthrough. If there's no action button at all we also need a
```

ptxmac:
> Can you add a test for the new behavior?

This code is part of a sample project - the whole code is the test :)

morganchen12 (approved):
LGTM, thanks for the thoroughness!

paulb777:
Please fix the formatting. See https://github.com/firebase/firebase-ios-sdk#code-formatting

ptxmac:
I think this problem is better solved by the firebase team itself, perhaps by adding documentation for in-app-messaging with swiftui, which is currently only present as the example I attempted to improve here

morganchen12:
@ptxmac, do you mind if I reopen this PR and take on the rest of the changes necessary? Your existing commits will end up in the SDK, but you won't be responsible for subsequent updates.

ptxmac:
Please do!

PR #9579, opened 2022-04-05
[Don't Merge] Run `messaging` CI

ncooke3:
The messaging CI is broken in #9568 but I'm not convinced the changes in that PR are causing the workflow to fail.

I ran the workflow locally on `master` and `v9b` and it failed as well.

Opening this PR to run on GHA.

#no-changelog

### DIFF
--- a/.github/workflows/messaging.yml
+++ b/.github/workflows/messaging.yml
@@ -10,6 +10,7 @@ on:
     # Podspec
     - 'FirebaseMessaging.podspec'
     # This file
+    # TODO(ncooke3): Remove this line.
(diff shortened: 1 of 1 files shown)


ncooke3:
Hmm, well the first run of `messaging` did fail: https://github.com/firebase/firebase-ios-sdk/runs/5838542113?check_suite_focus=true#step:7:669

It passed on the retry. I still cant get them to pass on a local master.

ncooke3:
Looks like the test was just flaky as I was able to get it to pass by retrying the job in https://github.com/firebase/firebase-ios-sdk/pull/9568 third time.

PR #9586, opened 2022-04-06
[Heartbeat Logging] Add logging within `FirebaseCore`

ncooke3:
#no-changelog

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -375,6 +375,7 @@ static FIRApp *sDefaultApp;
     }
   }
 
+  [self.heartbeatLogger log];
   [self subscribeForAppDidBecomeActiveNotifications];
 
   return YES;
@@ -866,8 +867,6 @@ static FIRApp *sDefaultApp;
 
 #pragma mark - App Life Cycle
 
-// TODO(ncooke3): Add heartbeat logging.
-
 - (void)subscribeForAppDidBecomeActiveNotifications {
 #if TARGET_OS_IOS || TARGET_OS_TV
   NSNotificationName notificationName = UIApplicationDidBecomeActiveNotification;
@@ -889,6 +888,8 @@ static FIRApp *sDefaultApp;
 
 - (void)logCoreTelemetryIfEnabled {
   if ([self isDataCollectionDefaultEnabled]) {
+    [self.heartbeatLogger log];
+    // TODO(ncooke3): Remove below code when CoreDiagnostics is removed.
     dispatch_async(dispatch_get_global_queue(QOS_CLASS_UTILITY, 0), ^{
       [FIRCoreDiagnosticsConnector logCoreTelemetryWithOptions:[self options]];
     });
--- a/FirebaseCore/Tests/Unit/FIRAppTest.m
+++ b/FirebaseCore/Tests/Unit/FIRAppTest.m
@@ -19,6 +19,7 @@
 #import "FirebaseCore/Extension/FIRAppInternal.h"
 #import "FirebaseCore/Extension/FIRComponentType.h"
 #import "FirebaseCore/Extension/FIRCoreDiagnosticsConnector.h"
+#import "FirebaseCore/Extension/FIRHeartbeatLogger.h"
 #import "FirebaseCore/Extension/FIROptionsInternal.h"
 #import "FirebaseCore/Sources/FIRAnalyticsConfiguration.h"
 #import "SharedTestUtilities/FIROptionsMock.h"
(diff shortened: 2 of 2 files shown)


FirebaseCore/Tests/Unit/FIRAppTest.m
```diff
@@ -50,6 +83,20 @@ + (nullable NSNumber *)readDataCollectionSwitchFromUserDefaultsForApp:(FIRApp *)
...
+
+@implementation FIRApp (Internal)
+
++ (id<FIRHeartbeatLoggerProtocol> (^)(NSString *appID))heartbeatLoggerProviderForAppID {
+  return ^id<FIRHeartbeatLoggerProtocol>(NSString *_) {
+    return heartbeatLoggerFake;
+  };
+}
+
+@end
```
ncooke3, line 96: I think an improvement here would be to expose a setter for FIRApp's `heartbeatLoggerProviderForAppID` and then set it to return a fake one in the `-[FIRAppTest setUp]`

FirebaseCore/Sources/FIRApp.m
```diff
@@ -339,7 +345,7 @@ - (instancetype)initInstanceWithName:(NSString *)name options:(FIROptions *)opti
     _options.editingLocked = YES;
     _isDefaultApp = [name isEqualToString:kFIRDefaultAppName];
     _container = [[FIRComponentContainer alloc] initWithApp:self];
-    _heartbeatLogger = [[FIRHeartbeatLogger alloc] initWithAppID:self.options.googleAppID];
+    _heartbeatLogger = [[self class] heartbeatLoggerProviderForAppID](options.googleAppID);
```
ncooke3, line 348: Since clients don't directly initialize a FIRApp, I need to find another way to be able to statically inject a fake heartbeat logger.

My current idea is to create a static property `heartbeatLoggerProviderForAppID` that is a block that is called by the internal implementation (line above) to return an initialized heartbeat logger. This block could then be overridden or set by the unit tests to inject a fake heartbeat logger.
```objc
[[self class] heartbeatLoggerProviderForAppID](options.googleAppID)
```
does look a bit wonky though.

Another (similar and maybe more readable?) approach would be to swap the heartbeatLoggerProvider block for a new factory type (`FIRHeartbeatLoggerFactory`) that conforms to a `FIRHeartbeatLoggerFactoryProtocol`. this line would then turn into something like:
```objc
_heartbeatLogger = 
      [[[self class] heartbeatLoggerFactory] makeHeartbeatLoggerWithAppID:options.googleAppID];
```
Similarly, a fake `heartbeatLoggerFactory` could be statially injected during tests to return a predefined fake heartbeat logger.

ncooke3: Hmm, I will experiment with second approach.

ncooke3: Discussed approach with Ryan offline.

FirebaseCore/Tests/Unit/FIRAppTest.m
ncooke3: Discussed approach with Ryan offline.

paulb777 (approved):

FirebaseCore/Tests/Unit/FIRAppTest.m
```diff
@@ -98,7 +105,20 @@ - (void)tearDown {
...
 
++ (void)tearDown {
+  // We stop mocking `FIRHeartbeatLogger` in the class `tearDown` method to
+  // prevent interfering with other tests that use the real `FIRHeartbeatLogger`.
+  // Doing this in the instance `tearDown` causes test failures due to a race
+  // condition between `NSNoticationCenter` and `OCMVerifyAllWithDelay`.
+  // Affected tests:
+  // - testCoreDiagnosticsLoggedWhenAppDidBecomeActive
+  // - testHeartbeatLogIsAttemptedWhenAppDidBecomeActive
+  [OCMClassMock([FIRHeartbeatLogger class]) stopMocking];
```
ryanwilson, line 121: I think this should be `[_mockHeartbeatLogger stopMocking]`, no?

ncooke3: No, because this is the _class_ `tearDown` method.

ryanwilson: Hm, I wonder if this is required at all. It _should_ stop mocking when it deallocates on its own, I know we've added it before on its own though. This seems like we're mocking it again then stopping. Have you tried to see if tests succeed as expected without this?

ncooke3: Yes, agreed. Without it, this is what happens:
image

The `FIRHeartbeatLoggerTests` fail because `OCMock` is still mocking the class (I'm not sure why because as you said: should stop mocking when it deallocates on its own).

When I first came across the issue, I tried to move the `stopMocking` call to the instance `tearDown` method. When doing this, OCMock unregisters the class for mocking so the other test files (`FIRHeartbeatLoggerTests`) aren't affected. But then, a race condition emerges between between OCMock and the way we wait for the `appDidBecomeActive` notification in: 
- testCoreDiagnosticsLoggedWhenAppDidBecomeActive
- testHeartbeatLogIsAttemptedWhenAppDidBecomeActive

Fwiw, I tried waiting for the notification in various ways with [XCTNSNotificationExpectation](https://developer.apple.com/documentation/xctest/xctnsnotificationexpectation?language=objc) but I couldnt get it to work.

image

ryanwilson: Strange, thanks a bunch for the info! LGTM

PR #9599, opened 2022-04-07
[Perf] Add missing `FirebaseCoreInternal` pod to testing Podfile

ncooke3:
#no-changelog

### DIFF
--- a/.github/workflows/performance-integration-tests.yml
+++ b/.github/workflows/performance-integration-tests.yml
@@ -3,6 +3,10 @@
 name: performance-integration-tests
 
 on:
+  pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/performance-integration-tests.yml
```diff
@@ -3,6 +3,10 @@
 name: performance-integration-tests
 
 on:
+  pull_request:
+    paths:
+    # This configuration file.
+    - '.github/workflows/performance-integration-tests.yml'
```
ncooke3, line 9: Adding this so perf's int tests run on changes to this workflow file.

visumickey (approved):

PR #9603, opened 2022-04-07
[v9] End beta period and update release tooling for select products

### LINKED ISSUE #9367: End Beta period for Swift libraries
We plan to end the beta period for the Firebase Swift libraries in the 9.0.0 release.  Please let us know if there is any feedback.

ncooke3:
Fixes #9367


scripts/build_non_firebase_sdks.sh
```diff
@@ -28,11 +28,11 @@ echo "[" >> "${ZIP_POD_JSON}"
...
 do
   if [ ${sdk} == "FirebaseFirestoreSwift" ]; then
-    echo "{\"name\":\"FirebaseFirestoreSwift\", \"version\" : \"> 8.2-beta\"}" >>  "${ZIP_POD_JSON}"
+    echo "{\"name\":\"FirebaseFirestoreSwift\", \"version\" : \"> 8.2\"}" >>  "${ZIP_POD_JSON}"
   elif [ ${sdk} == "FirebaseStorageSwift" ]; then
-    echo "{\"name\":\"FirebaseStorageSwift\", \"version\" : \"> 8.12-beta\"}" >>  "${ZIP_POD_JSON}"
+    echo "{\"name\":\"FirebaseStorageSwift\", \"version\" : \"> 8.12\"}" >>  "${ZIP_POD_JSON}"
   elif [ ${sdk} == "FirebaseRemoteConfigSwift" ]; then
-    echo "{\"name\":\"FirebaseRemoteConfigSwift\", \"version\" : \"> 8.12-beta\"}" >>  "${ZIP_POD_JSON}"
+    echo "{\"name\":\"FirebaseRemoteConfigSwift\", \"version\" : \"> 8.12\"}" >>  "${ZIP_POD_JSON}"
```
ryanwilson, line 35: cc @paulb777 are these needed anymore with the zip builder building the Swift pods?

ryanwilson (approved):

scripts/build_non_firebase_sdks.sh
paulb777: Yes, these should all be deleted. The ci scripts should have already been updated not to request them.

ncooke3: @ryanwilson @paulb777 , so this chunk should be removed?

https://github.com/firebase/firebase-ios-sdk/blob/fcd651bf35ec4a96f1ca743f496e33c82d321e43/scripts/build_non_firebase_sdks.sh#L30-L38

paulb777: I believe line 37 should remain.

ncooke3: Ahh i see now. Thanks!

paulb777 (approved):
LGTM on CI run as good as master

morganchen12 (approved):
LGTM, but please label breaking changes in the release notes (and, if applicable, actions users should take to resolve breakages).

paulb777:
Good point, we should document the necessary changes to the Podfile's and how to update an existing SPM project using one of the beta targets.

ncooke3:
> Good point, we should document the necessary changes to the Podfile's and how to update an existing SPM project using one of the beta targets.

Thanks Paul for mentioning SPM. So, for CocoaPods users, moving products out of beta will not affect them, correct?
But for SPM users, it may if the name of the library changed?

paulb777:
It does affect existing CocoaPods users because they need to remove the explicit beta specification from their Podfile. See https://stackoverflow.com/a/64671117/556617 just like we'll need to do in our quickstarts.

Yes, SPM users will be specifying a no longer existing product.

ncooke3:
I've updated the CHANGELOGs to reflect that moving out of beta is a breaking change (see above discussion).

There were two `zip / quickstart` failures that I believe are being addressed by #9607.

ncooke3:
cc: @weixifan

PR #9607, opened 2022-04-08
Add FirebaseCoreInternal for Firebase 9

granluo:
This is to fix 
https://github.com/firebase/firebase-ios-sdk/runs/5867959723?check_suite_focus=true#step:7:600

The Quickstart can be built without FirebaseCore, which does not need to be added explicitly.

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -116,6 +116,7 @@ jobs:
         LEGACY: true
       run: SAMPLE="$SDK" TARGET="${SDK}Example" scripts/setup_quickstart_framework.sh \
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseRemoteConfig/* \
+                                               "${HOME}"/ios_frameworks/Firebase/FirebaseAnalytics/FirebaseCore.xcframework \
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseAnalytics/FirebaseCoreInternal.xcframework \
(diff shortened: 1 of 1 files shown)


.github/workflows/zip.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         LEGACY: true
       run: SAMPLE="$SDK" TARGET="${SDK}Example" scripts/setup_quickstart_framework.sh \
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseRemoteConfig/* \
-                                               "${HOME}"/ios_frameworks/Firebase/FirebaseAnalytics/FirebaseCore.xcframework \
```
paulb777, line 119: I'm surprised this line can be deleted.

paulb777 (approved):
LGTM on successful zip test of ABTesting

.github/workflows/zip.yml
granluo: Looks like FirebaseCore is still needed here based on the [failed job](https://github.com/firebase/firebase-ios-sdk/runs/5882134634?check_suite_focus=true)
Will try another run after adding `FirebaseCore.framework`.

granluo: Also, the reason other Quickstarts do not have `FirebaseCore.xcframework` is because their have imported `FirebaseAnalytics/*` which includes FirebaseCore framework.

paulb777:
Tests look right. ABTesting is passing and only Storage is still failing.

PR #9622, opened 2022-04-08
Fix keychain access issues caused by prewarming

### LINKED ISSUE #8695: Firebase Auth - Signing users out after 30 minutes (As of Xcode 13, iOS 15 SDK).

  * Xcode version: 13.0
  * Firebase SDK version: 8.7.0
  * Firebase Component: Auth/Core

  (Added by me): 
  * iOS Version: 15.0 
  
After about 30 min., users are being signed out by Firebase automatically. This behavior did not occur prior to iOS 15. My company has in place, multiple safeguards via the iOS SDK & Cloud Functions for Auth. This issue is only occurring as of iOS 15.0.

#### Steps to reproduce:

Tested on both phone auth & Sign in with Apple: 

Test 1: 
Sign in --> Terminate application -> Open app after 5 minutes -->  still logged in. 

Test 2:
Sign in --> Terminate application -> Open app after 25 minutes --> still logged in. 

Test 3:
Sign in -> Terminate application --> Open app after 31 minutes --> logged out of firebase. 

This just started happening as of iOS 15 & Xcode 13. 
We have tests in place in our development environment to ensure things like this are not our fault. 
And, our application in the store is not experiencing this behavior. 

#### Relevant Code:

N/A.

rosalyntan:
Fix keychain access issues in iOS 15 that are caused by [prewarming](https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence).

For SwiftUI developers, the prewarming sequence means that `init` may be called before the device is unlocked, with the result that protected data in the keychain is unable to be accessed.

For UIKit developers, while [according to the documentation](https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence?language=objc) `didFinishLaunchingWithOptions` should not be called as part of the prewarming sequence, these Apple forum answers [[1](https://developer.apple.com/forums/thread/667959?answerId=695668022#695668022)][[2](https://developer.apple.com/forums/thread/696539)] indicate that `didFinishLaunchingWithOptions` is sometimes called during prewarming, before protected data in the keychain is accessible.

To solve this issue, we can check `isProtectedDataAvailable` before accessing the keychain, and if it is unavailable, listen for the `UIApplicationProtectedDataDidBecomeAvailable` notification.

Fixes #8695


FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -500,70 +506,88 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey
...
-      }
-      NSString *keychainServiceName =
-          [FIRAuth keychainServiceNameForAppName:strongSelf->_firebaseAppName];
-      if (keychainServiceName) {
-        strongSelf->_keychainServices =
-            [[FIRAuthKeychainServices alloc] initWithService:keychainServiceName];
-        strongSelf.storedUserManager =
-            [[FIRAuthStoredUserManager alloc] initWithServiceName:keychainServiceName];
-      }
+    if ([application isProtectedDataAvailable]) {
```
ryanwilson, line 509: This needs to be gated as only on iOS/tvOS/Catalyst still, but looks like this property was recently added to AppKit: 

https://developer.apple.com/documentation/appkit/nsapplication/3752992-isprotecteddataavailable

Not worrying about the new AppKit API yet, I think the logic should check if this API is available (it was only introduced in Catalyst in 13.1, but 13.0 exists) and if the instance doesn't respond to the selector use the old logic of just trying to read from the keychain.

WDYT?

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -500,70 +506,88 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey
...
-          FIRUser *user;
-          if ([strongSelf getUser:&user error:&error]) {
-            strongSelf.tenantID = user.tenantID;
-            [strongSelf updateCurrentUser:user byForce:NO savingToDisk:NO error:&error];
-            self->_lastNotifiedUserToken = user.rawAccessToken;
-          } else {
-            FIRLogError(kFIRLoggerAuth, @"I-AUT000001",
-                        @"Error loading saved user when starting up: %@", error);
-          }
+- (void)protectedDataInitialization:(UIApplication *)application {
```
ryanwilson, line 525: Since UIApplication isn't available on macOS, you may need to `#if TARGET_OS` this to provide two separate functions, one that takes a parameter and one that doesn't.

ryanwilson (commented):
Overall LGTM, just a comment about platform availability!

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: Done -- is there a macro for checking Catalyst version? I see examples of checking iOS version with a macro, but couldn't find one for Catalyst. Or is it enough to have the macro-level checks for the platform, and then also have the `respondsToSelector` check?

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: Done -- did some refactoring to get around this.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -485,85 +491,124 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey
...
-            strongSelf.tenantID = user.tenantID;
-            [strongSelf updateCurrentUser:user byForce:NO savingToDisk:NO error:&error];
-            self->_lastNotifiedUserToken = user.rawAccessToken;
-          } else {
-            FIRLogError(kFIRLoggerAuth, @"I-AUT000001",
-                        @"Error loading saved user when starting up: %@", error);
-          }
+- (void)protectedDataInitialization {
+#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_MACCATALYST
+  [[NSNotificationCenter defaultCenter] removeObserver:_protectedDataDidBecomeAvailableObserver
```
rosalyntan, line 535: @ryanwilson Is it fine to remove an observer that was never registered (since this observer only gets registered if protected data is unavailable)? I read that it is essentially a no-op, but couldn't find a definitive answer.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -485,85 +491,124 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey
...
+    if (![GULAppEnvironmentUtil isAppExtension]) {
+      Class cls = NSClassFromString(@"UIApplication");
+      if (cls && [cls respondsToSelector:@selector(sharedApplication)]) {
+        applicationClass = cls;
+      }
+    }
+    UIApplication *application = [applicationClass sharedApplication];
+
+    // Initialize for phone number auth.
+    strongSelf->_tokenManager = [[FIRAuthAPNSTokenManager alloc] initWithApplication:application];
```
rosalyntan, line 595: @ryanwilson This line seems to be failing static analysis now (since `application` can be nil if it's an iOS app extension and `initWithApplication` requires a non-nil argument). However, this is not a behavior change from how it was before, from what I can tell, since all I did was move this logic around. Is there a reason that it would fail static analysis now, but not before?

FirebaseAuth/Sources/Auth/FIRAuth.m
ryanwilson: Not a compile time one - I think checking via `respondsToSelector` on Catalyst makes sense to me.

FirebaseAuth/Sources/Auth/FIRAuth.m
ryanwilson: Great question! I imagine if it was never registered it would be `nil`, so you could gate it with a `nil` check. Although sounds like removing observer isn't overly necessary, but still good practice.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -485,85 +491,124 @@ - (nullable instancetype)initWithAPIKey:(NSString *)APIKey
...
+#endif  // TARGET_OS_IOS
+
+#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_MACCATALYST
     static Class applicationClass = nil;
     // iOS App extensions should not call [UIApplication sharedApplication], even if UIApplication
     // responds to it.
     if (![GULAppEnvironmentUtil isAppExtension]) {
       Class cls = NSClassFromString(@"UIApplication");
-      if (cls && [cls respondsToSelector:NSSelectorFromString(@"sharedApplication")]) {
+      if (cls && [cls respondsToSelector:@selector(sharedApplication)]) {
```
ryanwilson, line 505: I'm curious if there's a reason for switching away from `NSSelectorFromString` here

FirebaseAuth/Sources/Auth/FIRAuth.m
ryanwilson: Hmm, I agree it doesn't seem like a behavior change but maybe the analyzer couldn't pick it up before for some reason. Are these properties that require `application` optional, and could we skip creating them if `application` is `nil`?

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: No big reason -- is one preferred over the other? I see more instances of `@selector` in the Firebase Auth codebase than `NSSelectorFromString`, so I went with `@selector` (and it seems less error-prone, since it doesn't use a string-match), but can definitely change it back.

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: Done -- tested passing `nil` directly and it was fine!

FirebaseAuth/Sources/Auth/FIRAuth.m
rosalyntan: Done.

paulb777:
Marking for consideration for Firebase 9 milestone.

ryanwilson (approved):

PR #9624, opened 2022-04-08
Fix import bug

### LINKED ISSUE #8599: Obj-C to Swift auto translation Cocoa violation
For Cocoa API in Objective-C there's a convention, that an API that returns a nullable value and updates an NSError reference only returns nil when it also sets the error value.
(I haven't found an official source for this, but it's discussed here: https://stackoverflow.com/questions/34187188/using-an-objective-c-function-that-returns-an-optional-or-throw-in-swift)

The API: 
```
- (nullable FIRUser *)getStoredUserForAccessGroup:(NSString *_Nullable)accessGroup
```
is translated to:
```
func getStoredUser(forAccessGroup accessGroup: String?) throws -> User
```
following those Cocoa conventions - but in the code you can see that the intention is to return an optional `User` (meaning that it's perfectly fine and not an error to return nil).

This unfortunately can't be expressed through the automatic Objective-C/Cocoa bridging, so the API ends up being a bit weird, where the absence of a user (even though not really an error) ends up throwing a `nilError`.

https://github.com/firebase/firebase-ios-sdk/blob/0d90f2f12d18235dc6ac5131de5c8326243f0d33/FirebaseAuth/Sources/Auth/FIRAuth.m#L2233

I don't know of any good way to clean up such API, but perhaps it would be worthwhile documenting this bridging behavior somewhere.

morganchen12:
Fixes #8599.

### DIFF
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# v9.0.0
 - [fixed] **Breaking change:** Fixed an ObjC-to-Swift API conversion error where `getStoredUser(forAccessGroup:)` returned a non-optional type. This change is breaking for Swift users only (#8599).
 
 # v8.14.0
--- a/FirebaseAuth/CHANGELOG.md
+++ b/FirebaseAuth/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] **Breaking change:** Fixed an ObjC-to-Swift API conversion error where `getStoredUser(forAccessGroup:)` returned a non-optional type. This change is breaking for Swift users only (#8599).
+
 # v8.14.0
 - [added] Started to collect the Firebase user agent for Firebase Auth. (#9066)
 
--- a/FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift
+++ b/FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift
@@ -88,8 +88,12 @@ class AuthAPI_hOnlyTests: XCTestCase {
     auth.canHandleNotification([:])
     try auth.useUserAccessGroup("abc")
     let nilUser = try auth.getStoredUser(forAccessGroup: "def")
-    // This comparison doesn't need to execute, but it should compile.
-    _ = nilUser == nil
+    // If nilUser is not optional, this will raise a compiler error.
+    // This condition does not need to execute, and may not if prior
+    // functions throw.
+    if let value = nilUser {
+      XCTAssert(true)
+    }
   }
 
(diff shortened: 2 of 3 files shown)


paulb777:
Can you add a test to https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift?

paulb777:
Thanks Morgan. Is this the right fix versus making sure there is an error when nil is returned?

It seems that if an API claiming to return a FIRUser doesn't, it should provide an explanation as to why?

cc: @rosalyntan

rosalyntan:
> Is this the right fix versus making sure there is an error when nil is returned?
> 
> It seems that if an API claiming to return a FIRUser doesn't, it should provide an explanation as to why?

I think this seems right to me -- `nil` can be returned in the success case, it just means there is no user stored in the access group. [There is an example](https://firebase.google.com/docs/auth/ios/single-sign-on#migrate_a_signed-in_user_to_a_shared_keychain) in our docs that highlights this case.

paulb777:
Thanks for the explanation Rosalyn. That makes sense to me. :)

FirebaseAuth/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
```
paulb777, line 1: v 9.0.0

paulb777 (approved):

rosalyntan (approved):

FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift
```diff
@@ -87,7 +87,9 @@ class AuthAPI_hOnlyTests: XCTestCase {
     auth.setAPNSToken(Data(), type: AuthAPNSTokenType(rawValue: 2)!)
     auth.canHandleNotification([:])
     try auth.useUserAccessGroup("abc")
-    try auth.getStoredUser(forAccessGroup: "def")
+    let nilUser = try auth.getStoredUser(forAccessGroup: "def")
```
morganchen12, line 90: We discussed making this an explicit optional type in standup, but after testing it doesn't work because any type `T` can always be cast to `T?`. What this test _should_ express is `nilUser` should not be castable to type `User` without unwrapping.

FirebaseAuth/CHANGELOG.md
morganchen12: Done.

FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift
morganchen12: Made this slightly better.

PR #9625, opened 2022-04-08
[HeartbeatLogging] Remove heartbeat trigger for when app is configured

ncooke3:
Because of the fact that the `appDidBecomeActive` notification comes after the time when a Firebase app should be configured, it's reasonable to _only_ log a heartbeat upon the `appDidBecomeActive` notification.

#no-changelog

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -375,7 +375,6 @@ static FIRApp *sDefaultApp;
     }
   }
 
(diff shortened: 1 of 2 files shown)


paulb777 (approved):
LGTM on green

ncooke3:
I'm working on a sample app that can include a test to assert that `appDidBecomeActive` always comes after the point of Firebase app configuration. This would serve to catch any future changes in app lifecycle.

ryanwilson (approved):

PR #9630, opened 2022-04-09
Update binary install instructions for disabling AdId

### LINKED ISSUE #9609: Feature: Support dynamic targets
### Feature proposal

* Firebase Component: **FirebaseCrashlytics** (Auth, Core, Database, Firestore, Messaging, Storage, etc)

Can we support dynamic targets for Crashlytcs / Analytics by adding an additional library target in `Package.swift` such as this?

```
.library(
      name: "FirebaseCrashlyticsDynamic",
      type: .dynamic,
      targets: ["FirebaseCrashlytics"]
    ),
```

This way we'll be able to ship a single dynamic framework for all the targets of the app that wish to link with Crashlytics (if crash reporting will still work when it's loaded dynamically?) This would considerably shrink the size of our app bundle as currently each target that links with crashlytics, does so statically, resulting in a much larger overall .app bundle.

Clients could then wish to link against `FirebaseCrashlyticsDynamic` instead. 

Most Swift Packages support dynamic targets now, such as RxSwift: https://github.com/ReactiveX/RxSwift/blob/main/Package.swift

paulb777:
Fix #9609 

#no-changelog

### DIFF
--- a/Carthage.md
+++ b/Carthage.md
@@ -53,6 +53,10 @@ binary "https://dl.google.com/dl/firebase/ios/carthage/FirebaseStorageBinary.jso
 - Use Finder to open `Carthage/Build/iOS`.
 - Copy the contents into the top level of your Xcode project and make sure
     they're added to the right build target(s).
+    - To disable AdId support, do not copy
+   `GoogleAppMeasurementIdentitySupport.xcframework`.
+    - If the app does not use any Firebase Swift specific APIs, you do not need
+   to copy any xcframeworks whose name includes "Swift".
 - Add `$(OTHER_LDFLAGS) -ObjC` flag to "Other Linker Flags" in "Build Settings".
 - Make sure that the build target(s) includes your project's `GoogleService-Info.plist`
  ([how to download config file](https://support.google.com/firebase/answer/7015592))
--- a/ReleaseTooling/Template/README.md
+++ b/ReleaseTooling/Template/README.md
@@ -23,6 +23,9 @@ To integrate a Firebase SDK with your app:
    want the framework to be added to has a checkmark next to it, and that
    you've selected "Copy items if needed".
 
+   *To disable AdId support, do not copy
+   `GoogleAppMeasurementIdentitySupport.xcframework`.*
+
    *If the app does not use any Firebase Swift specific APIs, you do not need
    to copy any xcframeworks whose name includes "Swift" for this and the next
    step.*


guidedways:
@paulb777 I find this instruction confusing:

```
*Do not add the Firebase frameworks to the "Embed Frameworks" Xcode build
   phase. The Firebase frameworks are not embedded dynamic frameworks, but are
   [static frameworks](https://www.raywenderlich.com/65964/create-a-framework-for-ios)
   which cannot be embedded into your application's bundle.*
```

To me this sounded like "do not embed any of the Firebase frameworks the instructions asked us to copy" - and so the app crashed at runtime since we had marked "GoogleUtilities" with "do not embed". I think it should clearly mention that only the `FirebaseXXXXXX.xcframeworks` are not to be embedded.

paulb777:
@guidedways Sorry, I'm not following. The GoogleUtilities.xcframework in the FirebaseAnalytics directory should have the same characteristics as the `FirebaseXXXXXX.xcframeworks`

guidedways:
@paulb777 Our app seems to crash:

```
8.14.0 - [Firebase/Analytics][I-ACS030000] Exception on worker queue. Block ID=APMMeasurement:1651. Exception: -[APMDatabase insertBundle:isRealtime:error:]: +[NSData gul_dataByGzippingData:error:]: unrecognized selector sent to class 0x210eee4e8
```

We've got two targets in our macOS app. I added all the xcframeworks and then set them to "Do Not embed". The app crashed a few seconds into the launch with:

`+[NSData gul_dataByGzippingData:error:]: unrecognized selector sent to class 0x210eee4e8`

I then set GoogleUtilities.xcframework to "Embed & Sign", this prevented the crash but now our plugin is crashing. Even with "Embed & Sign" it seems to be constantly crashing with the error above. Strangely it seems it can't find GoogleUtilities at runtime.

guidedways:
I'm seeing other messages on the console:

```
+[FBLPromise onQueue:do:]: unrecognized selector sent to class 0x100fc39f8
```

So far we've followed the instructions to the T, I'll go over these again to see if we've missed something. The app compiles and launches but crashes at runtime with missing symbols from frameworks / libraries used by Firebase.

paulb777:
Unrecognized selector errors can happen if you miss the instruction to add `-ObjC` to the linker options.

guidedways:
@paulb777 I could have sworn I did just that - in fact we had that set already, but now that I checked a second time, it wasn't set! I'm so sorry... 😅 Thank you for your help!

ryanwilson (approved):

PR #9632, opened 2022-04-10
Interop libs should be included in the zip

paulb777:
The zip distro SHOULD include the interop libraries since the libraries built with Swift depend on them.

I've confirmed that the StorageSwiftExample test successfully builds when the Interop libs are linked in (After removing teh `import FirebaseStorageSwift` from the sources.

#no-changelog


paulb777:
Green tests still require the rest of #9632 to be solved, but this PR is an incremental improvement that gets Swift apps working with the zip that use Functions or Storage.

ryanwilson (approved):

PR #9633, opened 2022-04-11
Bump zip minimum versions

### LINKED ISSUE #9631: Zip issues with Swift-implemented libraries - Fix Zip Storage test failures
Reference https://github.com/firebase/firebase-ios-sdk/runs/5954330599?check_suite_focus=true

The Swift Sample is failing because the Interop xcframeworks are not included in the zip build:

Screen Shot 2022-04-09 at 6 58 01 AM

The Objective C Sample is failing because `FIRStorageReference` is not recognized, even though `FIRStorageMetadata` is:

Screen Shot 2022-04-09 at 7 00 19 AM

paulb777:
Bump minimum iOS versions and associated other versions to drop support for 32 bit arm IOS OS's.

With Xcode 13.2.1, when building for iOS 10, the FirebaseStorage-Swift.h bridging header gets created with architecture ifdef's that fail to trigger when testing an Objective C test app with libraries built from a Swift library.

The ifdefs are not generated when building for iOS 11 and the test app builds fine.

Fix #9631

### DIFF
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Firebase 9.0.0
 - [changed] Firebase now requires at least Xcode 13.2.1.
 - [added] The zip and Carthage distibutions now include the Swift extension frameworks. (#7819)
+- [changed] Update the minimum supported versions for the zip and Carthage distributions to
+  iOS 11.0, tvOS 11.0 and macOS 10.13.
 
 # Firebase 8.10.0
 - [fixed] Fixed platform availability checks in Swift Package Manager that may prevent code
--- a/ReleaseTooling/Sources/ZipBuilder/main.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/main.swift
@@ -100,15 +100,15 @@ struct ZipBuilderTool: ParsableCommand {
   // MARK: - Platform Arguments
 
   /// The minimum iOS Version to build for.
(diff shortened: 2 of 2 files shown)


FirebaseCore/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Firebase 9.0.0
 - [changed] Firebase now requires at least Xcode 13.2.1.
 - [added] The zip and Carthage distibutions now include the Swift extension frameworks. (#7819)
+- [changed] Update the minimum supported versions for the zip and Carthage distributions to
```
ncooke3, line 4: Should this entry be marked as a breaking change?
```suggestion
- [changed] **Breaking change**: update the minimum supported versions for the zip and Carthage distributions to
```

paulb777:
The Objective C test now passes https://github.com/firebase/firebase-ios-sdk/runs/5980503942?check_suite_focus=true. The Swift test is pending quickstart source changes defined by #9635

paulb777:
The zip build sped up by 40 minutes by dropping the 32 bit architecture builds.

ryanwilson (approved):
LGTM - I wonder if this is an Xcode 13.2.1 specific thing or if Xcode 13.3 resolves it (may be better to drop support for a single Xcode version vs bump min supported versions). Not going to block on this though, we can investigate later.

paulb777:
Trying Xcode 13.3 in #9638

Issue #9604, opened 2022-04-07
Investigate async/await implications for AppDistro's fetch releases


ryanwilson:
I think this completion handler should return an error instead of `nil, nil`.

https://github.com/firebase/firebase-ios-sdk/blob/d8bd124e06f15fa730f2f6e04b97e8bced41e7d3/FirebaseAppDistribution/Sources/FIRAppDistribution.m#L289

PR #9634, opened 2022-04-11
App Distro: fix async/await issue

ryanwilson:
Right now we're calling a completion handler with `nil, nil` when
releases are fetched but none are marked as "latest". This is an expected result - no new releases but no errors. When translated to async/await, the result is not an optional, which will crash. This turns the return type into an Optional.

### DIFF
--- a/FirebaseAppDistribution.podspec
+++ b/FirebaseAppDistribution.podspec
@@ -43,7 +43,10 @@ iOS SDK for App Distribution for Firebase.
 
   s.test_spec 'unit' do |unit_tests|
    unit_tests.scheme = { :code_coverage => true }
-   unit_tests.source_files = 'FirebaseAppDistribution/Tests/Unit*/*.[mh]'
+   unit_tests.source_files = [
+     'FirebaseAppDistribution/Tests/Unit*/*.[mh]',
+     'FirebaseAppDistribution/Tests/Unit/Swift*/*.swift',
+   ]
    unit_tests.resources = 'FirebaseAppDistribution/Tests/Unit/Resources/*'
    unit_tests.dependency 'OCMock'
   end
--- a/FirebaseAppDistribution/Sources/FIRAppDistribution.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistribution.m
@@ -245,49 +245,52 @@ NSString *const kFIRFADSignInStateKey = @"FIRFADSignInState";
 
 - (void)fetchNewLatestRelease:(void (^)(FIRAppDistributionRelease *_Nullable release,
                                         NSError *_Nullable error))completion {
-  [FIRFADApiService
-      fetchReleasesWithCompletion:^(NSArray *_Nullable releases, NSError *_Nullable error) {
-        if (error) {
-          if ([error code] == FIRFADApiErrorUnauthenticated) {
-            FIRFADErrorLog(@"Tester authentication failed when fetching releases. Tester will need "
-                           @"to sign in again.");
-            [self signOutTester];
-          } else if ([error code] == FIRFADApiErrorUnauthorized) {
-            FIRFADErrorLog(@"Tester is not authorized to view releases for this app. Tester will "
-                           @"need to sign in again.");
-            [self signOutTester];
-          }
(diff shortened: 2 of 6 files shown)


FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -245,49 +245,52 @@ - (NSError *_Nullable)mapFetchReleasesError:(NSError *)error {
...
-                              [release buildVersion]);
-                completion(release, nil);
-              });
-              return;
-            }
-          }
-        }
-
-        completion(nil, nil);
-      }];
```
ryanwilson, line 290: These are purely indentation fixes from go/style

FirebaseAppDistribution/Sources/FIRAppDistribution.m
```diff
@@ -245,49 +245,52 @@ - (NSError *_Nullable)mapFetchReleasesError:(NSError *)error {
...
+    // At this point, no release matches the "latest release". An error should be returned.
+    NSError *notFound = [self NSErrorForErrorCodeAndMessage:FIRAppDistributionErrorUnknown
+                                                    message:@"Could not find the latest release."];
+    completion(nil, notFound);
+  }];
```
ryanwilson, line 293: This is the only new code

ryanwilson:
Chatting with the App Distro team about the expected behaviour here - originally they said it should return an error but now seeing the test failing, it's worth confirming.

ryanwilson:
Should fix #9604

FirebaseAppDistribution.podspec
```diff
@@ -43,7 +43,10 @@ iOS SDK for App Distribution for Firebase.
 
   s.test_spec 'unit' do |unit_tests|
    unit_tests.scheme = { :code_coverage => true }
-   unit_tests.source_files = 'FirebaseAppDistribution/Tests/Unit*/*.[mh]'
+   unit_tests.source_files = [
+     'FirebaseAppDistribution/Tests/Unit*/*.[mh]',
+     'FirebaseAppDistribution/Tests/Unit/Swift*/*.swift',
```
paulb777, line 48: Don't forget to `git add` the new tests.

FirebaseAppDistribution/Sources/Public/FirebaseAppDistribution/FIRAppDistribution.h
```diff
@@ -47,8 +47,8 @@ NS_SWIFT_NAME(AppDistribution)
  * Check to see whether a new distribution is available
  */
 - (void)checkForUpdateWithCompletion:
-    (void (^)(FIRAppDistributionRelease *_Nullable release, NSError *_Nullable error))completion
-    NS_SWIFT_NAME(checkForUpdate(completion:));
+    (void (^)(FIRAppDistributionRelease *_Nullable_result release, NSError *_Nullable error))completion
+    NS_SWIFT_NAME(checkForUpdate(completion:)) ;
```
paulb777, line 51: extra space?

FirebaseAppDistribution.podspec
ryanwilson: Whoops! Thanks

paulb777 (approved):

Package.swift
```diff
@@ -410,11 +410,20 @@ let package = Package(
...
+      exclude: ["Swift/"],
       resources: [.process("Resources")],
       cSettings: [
         .headerSearchPath("../../.."),
       ]
     ),
+    .testTarget(
+      name: "AppDistributionUnitSwift",
+      dependencies: ["FirebaseAppDistribution"],
+      path: "FirebaseAppDistribution/Tests/Unit/Swift",
```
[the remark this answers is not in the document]
ryanwilson: Mind if I follow up in a separate PR? Looks like Core and Functions use `SwiftUnit`, and AppCheck, Database use `Unit/Swift`. Would be good to have a standard structure.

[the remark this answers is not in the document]
paulb777: SGTM, although that may be just the start of a slippery slope of standardizing test directory structure :)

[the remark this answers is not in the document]
ryanwilson: Maybe a slippery slope that needs to be slid down 😝

PR #9638, opened 2022-04-12
Try zip build with Xcode 13.3

paulb777:
See discussion in #9633


ryanwilson:
I this this can just be closed and we can investigate at a later date.

PR #9576, opened 2022-04-05
Adding Custom-Time to Firebase Storage

NathanFallet:
Hey there! So you want to contribute to a Firebase SDK?
Before you file this pull request, please read these guidelines:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here.
    If not, go file an issue about this **before creating a pull request** to discuss.

Fixes #9575 

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help
    us make Firebase APIs better, please propose your change in a feature request so that we
    can discuss it together.

PR #9648, opened 2022-04-13
Use the v9 quickstart branch until the v9 release

paulb777:
#no-changelog


paulb777:
The storage failure is a regression or exposure from #9576 that wasn't caught because that PR was from a fork.

ncooke3 (approved):

PR #9650, opened 2022-04-14
Run queries through index manager

wu-hui:


Firestore/core/src/local/leveldb_index_manager.h
```diff
@@ -89,6 +103,91 @@ class LevelDbIndexManager : public IndexManager {
...
+  /**
+   * Updates the index entries for the provided document by deleting entries
+   * that are no longer referenced in `new_entries` and adding all newly added
+   * entries.
+   */
+  void UpdateEntries(const model::Document& document,
+                     const model::FieldIndex& index,
+                     const std::set<index::IndexEntry>& existing_entries,
+                     const std::set<index::IndexEntry>& new_entries);
+  void AddIndexEntry(const model::Document& document,
```
schmidt-sebastian, line 122: Add newline as the method comment above only applies to `UpdateEntries`.

Firestore/core/src/local/leveldb_index_manager.h
```diff
@@ -89,6 +103,91 @@ class LevelDbIndexManager : public IndexManager {
...
+   * Encodes the given field values according to the specification in `target`.
+   * For IN queries, a list of possible values is returned.
+   */
+  std::vector<std::string> EncodeValues(const model::FieldIndex& index,
+                                        const core::Target& target,
+                                        core::IndexedValues values);
+
+  /**
+   * Constructs a vector of LevelDb key ranges that unions all bounds.
+   * Representing the ranges in the index entry table satisfying the given
```
schmidt-sebastian, line 168:
```suggestion
   * These ranges represent the sections in the index entry table that contain the given
```

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -1240,18 +1297,34 @@ std::string LevelDbIndexEntryKey::KeyPrefix(int32_t index_id) {
...
+                                      absl::string_view document_key) {
   Writer writer;
   writer.WriteTableName(kIndexEntriesTable);
   writer.WriteIndexId(index_id);
   writer.WriteUserId(user_id);
   writer.WriteIndexArrayValue(array_value);
   writer.WriteIndexDirectionalValue(directional_value);
-  writer.WriteDocumentId(document_name);
+  writer.WriteOrderedDocumentId(ordered_document_key);
+  writer.WriteDocumentId(document_key);
```
schmidt-sebastian, line 1327: These methods should end in "Key".

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -682,6 +723,10 @@ class Writer {
     WriteLabeledString(ComponentLabel::DocumentId, document_id);
   }
 
+  void WriteOrderedDocumentId(absl::string_view document_id) {
```
schmidt-sebastian, line 726: s/id/key

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -139,6 +141,15 @@ enum ComponentLabel {
   /** A component containing a collection group name. */
   CollectionGroup = 22,
 
+  /** A general component to differentiate indexes with same prefixes. */
+  SequenceNumber = 23,
+
+  /**
+   * A component containing an encoded document ID with either ASC or DESC
```
schmidt-sebastian, line 148: s/ID/key

Firestore/core/src/local/leveldb_key.cc
```diff
@@ -139,6 +141,15 @@ enum ComponentLabel {
...
+   * ordering.
+   */
+  OrderedDocumentId = 24,
```
schmidt-sebastian, line 151: s/Id/Key

Firestore/core/src/local/leveldb_index_manager.cc
```diff
@@ -380,7 +698,236 @@ void LevelDbIndexManager::UpdateCollectionGroup(
...
+      entry.index_id(), uid_, document_key, 0);
+  if (!raw_key.empty()) {
+    bool decoded = document_key_index_key.Decode(raw_key);
+    HARD_ASSERT(decoded,
+                "LevelDbIndexEntryDocumentKeyIndexKey cannot be decoded from "
+                "document key index table.");
+    document_key_index_key.IncreaseSeqNumber();
+  }
+
+  db_->current_transaction()->Put(document_key_index_key.Key(), entry_key);
```
schmidt-sebastian, line 892: There is probably no better way, but we are storing the index value twice just so that we can delete it later. Just something I am 99% sure you are aware of, but I want to make sure that you are 100% aware.

Firestore/core/src/model/value_util.cc
```diff
@@ -448,19 +449,17 @@ std::string CanonifyObject(const google_firestore_v1_Value& value) {
...
+    std::string field;
+    absl::StrAppend(&field, nanopb::MakeStringView(fields[i].key), ":",
                     CanonicalId(fields[i].value));
-    if (i != fields_count - 1) {
-      absl::StrAppend(&result, ",");
-    }
+    fields_string.push_back(std::move(field));
   }
-  result += "}";
+  std::sort(fields_string.begin(), fields_string.end());
```
schmidt-sebastian, line 459: Why did this have to change?

The sort order is also slightly off in some cases as you are sorting by "field:value" rather than just field, which will yield a different sort order for fields that contain special characters - and that could be considered a breaking change as these queries will not be able to served from cache right after the SDK upgrade (probably not a problem in practice).

Firestore/core/test/unit/local/leveldb_key_test.cc
```diff
@@ -658,9 +665,10 @@ TEST(IndexEntryKeyTest, EncodeDecodeCycle) {
 TEST(IndexEntryKeyTest, Description) {
   AssertExpectedKeyDescription(
       "[index_entries: index_id=1 user_id=user array_value=array "
-      "directional_value=directional document_id=foo-bar?baz!quux]",
+      "directional_value=directional ordered_document_id=<skipped> "
```
schmidt-sebastian, line 668: Why `<skipped>`?

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+  persistence->Run("AddsDocuments", [&]() {
+    index_manager->Start();
+    index_manager->AddFieldIndex(
+        MakeFieldIndex("coll", "exists", model::Segment::kAscending));
+    AddDoc("coll/doc1", Map("exists", 1));
+    AddDoc("coll/doc2", Map());
+  });
+}
+
+TEST_F(LevelDbIndexManagerTest, TestOrderByFilter) {
```
schmidt-sebastian, line 149: You should drop the Test prefix from all test names

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+        original_query
+            .StartingAt(Bound::FromValue(Array(2), /* inclusive= */ false))
+            .EndingAt(Bound::FromValue(Array(4), /* inclusive= */ false));
+
+    {
+      SCOPED_TRACE("Verifing original");
+      VerifyResults(original_query, {"coll/val2", "coll/val3", "coll/val4"});
+    }
+    {
+      SCOPED_TRACE("Verifing non restricted bound");
```
schmidt-sebastian, line 211:
```suggestion
      SCOPED_TRACE("Verifing non-restricted bound");
```

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+        original_query
+            .StartingAt(Bound::FromValue(Array(4), /* inclusive= */ false))
+            .EndingAt(Bound::FromValue(Array(2), /* inclusive= */ false));
+
+    {
+      SCOPED_TRACE("Verifying original");
+      VerifyResults(original_query, {"coll/val4", "coll/val3", "coll/val2"});
+    }
+    {
+      SCOPED_TRACE("Verifying non restricted bound");
```
schmidt-sebastian, line 246:
```suggestion
      SCOPED_TRACE("Verifying non-restricted bound");
```

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+  persistence->Run("TestAscendingOrderWithLessThanFilter", [&]() {
+    index_manager->Start();
+    SetUpMultipleOrderBys();
+
+    auto original_query = Query("coll")
+                              .AddingFilter(Filter("a", "==", 2))
+                              .AddingFilter(Filter("b", "==", 2))
+                              .AddingFilter(Filter("c", "<", 5))
+                              .AddingOrderBy(OrderBy("c", "asc"));
+    auto query_with_non_restricted_bound =
```
schmidt-sebastian, line 197: Move these queries to the scope you created?

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
schmidt-sebastian: Or non-restrictive

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+              Filter("count", "!=", 2)},
+             {"coll/val3"}}};
+
+    size_t counter = 0;
+    for (const auto& filter_result_pair : filtersAndResults) {
+      auto query = Query("coll");
+      for (const auto& filter : filter_result_pair.first) {
+        query = query.AddingFilter(filter);
+      }
+      {
```
schmidt-sebastian, line 690: We don't need the extra scope here. We can use the scope from the loop.

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+    AddDoc("coll/doc1", Map("value", true));
+    {
+      SCOPED_TRACE("With doc1");
+      VerifyResults(query, {"coll/doc1"});
+    }
+
+    AddDocs(
+        {Doc("coll/doc1", 1, Map()), Doc("coll/doc2", 1, Map("value", true))});
+    {
+      SCOPED_TRACE("With doc1 and doc2");
```
schmidt-sebastian, line 817: With non-matching doc1 and doc2

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+                                                model::Segment::kAscending));
+
+    std::vector<nanopb::Message<google_firestore_v1_Value>> data;
+    data.push_back(Map());
+    data.push_back(Map("int", 1, "array", Array(1, "foo")));
+    data.push_back(Map("array", Array(2, "foo")));
+    data.push_back(Map("int", 3, "array", Array(3, "foo")));
+    data.push_back(Map("array", "foo"));
+    data.push_back(Map("array", Array(1)));
+    data.push_back(Map("float", -0.0, "string", "a"));
```
schmidt-sebastian, line 899: I have a feeling that you had to change the canonical id because the Maps here aren't sorted. Can we sort them? Either manually or via `Map`.
If we sort manually we should probably also assert that `Map`'s input is sorted (to prevent headache for other people in the future).

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+        {q.AddingFilter(Filter("map", "==", Map())), {"coll/{map:{}}"}},
+        {q.AddingFilter(Filter("map", "==", Map("field", true))),
+         {"coll/{map:{field:true}}"}},
+        {q.AddingFilter(Filter("map.field", "==", true)),
+         {"coll/{map:{field:true}}"}},
+        {q.AddingOrderBy(OrderBy("map")),
+         {"coll/{map:{}}", "coll/{map:{field:false}}",
+          "coll/{map:{field:true}}"}},
+        {q.AddingOrderBy(OrderBy("map.field")),
+         {"coll/{map:{field:false}}", "coll/{map:{field:true}}"}}};
```
schmidt-sebastian, line 1099: I totally reviewed this. I promise :)

Firestore/core/test/unit/local/leveldb_index_manager_test.cc
```diff
@@ -62,11 +71,1042 @@ class LevelDbIndexManagerTest : public ::testing::Test {
...
+         {"coll/{multi:1}", "coll/{multi:string}", "coll/{multi:[]}"}},
+        {q.AddingFilter(Filter("multi", "in", Array(true, 1))),
+         {"coll/{multi:true}", "coll/{multi:1}"}},
+        {q.AddingFilter(Filter("multi", "not-in", Array(true, 1))),
+         {"coll/{multi:string}", "coll/{multi:[]}"}},
+        {q.AddingOrderBy(OrderBy("array"))
+             .StartingAt(Bound::FromValue(Array(Array(2)), true)),
+         {"coll/{array:[2,foo]}", "coll/{array:[3,foo],int:3}"}},
+
+        {q.AddingOrderBy(OrderBy("array", "desc"))
```
schmidt-sebastian, line 977: I am not quite able to decipher the pattern you used for adding newlines, but if there is one, feel free to keep as is:)

schmidt-sebastian (commented):
You should be proud of this PR. It's outstanding. 

Enjoy your easter break!

Firestore/core/src/local/leveldb_index_manager.cc
```diff
@@ -380,7 +698,236 @@ void LevelDbIndexManager::UpdateCollectionGroup(
...
+      entry.index_id(), uid_, entry.array_value(), entry.directional_value(),
+      EncodedDirectionalKey(index, document_key), document_key);
+  db_->current_transaction()->Delete(entry_key);
+
+  auto document_key_index_prefix =
+      LevelDbIndexEntryDocumentKeyIndexKey::KeyPrefix(entry.index_id(), uid_,
+                                                      document_key);
+  LevelDbIndexEntryDocumentKeyIndexKey document_key_index_key;
+  auto iter = db_->current_transaction()->NewIterator();
+  for (iter->Seek(document_key_index_prefix); iter->Valid(); iter->Next()) {
```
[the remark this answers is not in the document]
wu-hui: Maybe, I want to keep it for now, because it is at least correct. I'll revisit this when I port the backfiller.

ehsannas (commented):
I still have to review leveldb_index_manager_test.cc. Other than that it looks good!

Firestore/core/src/local/leveldb_index_manager.cc
wu-hui: The indexed values are saved twice in leveldb, one in the index entry table and one in `LevelDbIndexEntryDocumentKeyIndexKey` as an index table.

Firestore/core/test/unit/local/leveldb_key_test.cc
wu-hui: It is encoded string, which does not make a lot sense to print.

ehsannas (approved):

PR #9662, opened 2022-04-15
Create individual Python virtualenv's in cmake builds

dconeybe:
During cmake builds, instead of using the python interpreter installed locally, use that python interpreter to create a virtualenv for each python script run by cmake. Each virtualenv can then have its own isolated dependencies installed. This removes the dependence on having the correct python libraries installed locally. All you need is a vanilla python 3 interpreter. This also helps fix some unusual python dependency issues when this repository is incorporated into the firebase-cpp-sdk.

Here is an example of the build error in the C++ SDK that this fixes (https://github.com/firebase/firebase-cpp-sdk/runs/6065960193):

```
[ 83%] Generating nanopb sources
google/firestore/v1/document.proto:20:1: warning: Import google/api/annotations.proto is unused.
google/firestore/v1/query.proto:20:1: warning: Import google/api/annotations.proto is unused.
google/firestore/v1/common.proto:20:1: warning: Import google/api/annotations.proto is unused.
google/firestore/v1/write.proto:20:1: warning: Import google/api/annotations.proto is unused.
google/firestore/admin/index.proto:21:1: warning: Import google/api/annotations.proto is unused.

         *************************************************************
         *** Could not import the Google protobuf Python libraries ***
         *** Try installing package 'python-protobuf' or similar.  ***
         *************************************************************
    
Traceback (most recent call last):
  File "external/src/firestore/Firestore/Protos/tmpuwkOue.py", line 25
    import nanopb_generator as nanopb
  File "external/src/firestore-build/external/src/nanopb/generator/nanopb_generator.py", line 25
    import google.protobuf.text_format as text_format
  File "external/src/firestore-build/external/src/protobuf/python/google/protobuf/text_format.py", line 51
    import six
ImportError: No module named six
--nanopb_out: protoc-gen-nanopb: Plugin failed with status code 1.
```

For some reason the build is using a Python interpreter that does not have the `six` module installed; however, the `six` module is installed during setup of the runner. So it _must_ be choosing a different Python interpreter. This PR fixes this by setting up a dedicated virtualenv with `six` installed for running the Python script.

#no-changelog

### DIFF
--- a/Firestore/Protos/CMakeLists.txt
+++ b/Firestore/Protos/CMakeLists.txt
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include(FindPythonInterp)
+include(python_setup)
+FirebaseSetupPythonInterpreter(
+  OUTVAR MY_PYTHON_EXECUTABLE
+  KEY FirestoreProtos
+  REQUIREMENTS six
+)
 
 # Generate output in-place. So long as the build is idempotent this helps
 # verify that the protoc-generated output isn't changing.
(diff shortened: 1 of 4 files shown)


wu-hui (approved):
Thank you for doing this!

dconeybe:
FYI This script has been copied into the firebase-cpp-sdk as well: https://github.com/firebase/firebase-cpp-sdk/pull/908

PR #9665, opened 2022-04-19
Analytics 9.0.0

tsunghung:
Analytics 9.0.0 release for CocoaPods and Swift Package Manager.

### DIFF
--- a/FirebaseAnalytics.podspec.json
+++ b/FirebaseAnalytics.podspec.json
@@ -22,7 +22,7 @@
         "z"
     ],
     "license": {
-        "text": "Copyright 2021 Google",
+        "text": "Copyright 2022 Google",
         "type": "Copyright"
     },
     "name": "FirebaseAnalytics",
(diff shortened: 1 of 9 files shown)


FirebaseAnalytics.podspec.json
```diff
@@ -22,7 +22,7 @@
         "z"
     ],
     "license": {
-        "text": "Copyright 2021 Google",
+        "text": "Copyright 2022 Google",
```
paulb777, line 25: Technically Copyrights only need to be updated when the file is created, but we have been updating them for podspecs - so either way.

SwiftPMTests/analytics-import-test/analytics-import.swift
```diff
@@ -21,7 +21,7 @@ import FirebaseAnalytics
```
paulb777, line 21: Consider adding an import to the new library in this test and any other possible tests.

paulb777 (approved):

FirebaseAnalytics.podspec.json
tsunghung: Interesting, thanks for the info.

SwiftPMTests/analytics-import-test/analytics-import.swift
tsunghung: The new library test is tricky. The API is exported in FirebaseAnalytics. Without the OnDeviceConversion library, the new API will still be visible, but just an no-op. To verify the internal behavior is also difficult here. Will discuss internally to see how to create tests here.

PR #9669, opened 2022-04-20
python_setup.cmake: add FIREBASE_PYTHON_PYVENV_DEST_DIR cmake cache var

dconeybe:
Add the cmake cache variable `FIREBASE_PYTHON_PYVENV_DEST_DIR`. This variable will be used by the `FirebaseSetupPythonInterpreter()` function added in https://github.com/firebase/firebase-ios-sdk/pull/9662 to allow a custom directory to be specified for the root directory of the python virtualenvs it creates.

This can be used as a workaround for virtualenv paths that exceed the Windows maximum path length of 260. For example, this happened in the [`build-desktop-windows-latest-openssl`](https://github.com/firebase/firebase-cpp-sdk/runs/6087699107) workflow in the firebase-cpp-sdk repository.

#no-changelog

### DIFF
--- a/cmake/python_setup.cmake
+++ b/cmake/python_setup.cmake
@@ -91,12 +91,24 @@ function(FirebaseSetupPythonInterpreter)
     "The Python interpreter on the host system to use"
   )
 
+  # Use a custom directory for the pyvenv directories, if specified.
+  set(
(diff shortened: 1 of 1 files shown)


dconeybe:
Turns out this workaround isn't needed. The Windows problem was using a Python interpreter without a good venv.

PR #9670, opened 2022-04-20
Fix platforms and zip attribute on new pods

paulb777:
#no-changelog


ryanwilson (approved):

ryanwilson:
Carthage failed: can you create a blank JSON manifest? We could update the logic in the zip builder to create a new one if it's not found, but I think we may have done that before and pushed a JSON manifest that only contained the most recent version so I think this is fine. We can update our instructions for adding a new pod / framework too.

PR #9673, opened 2022-04-20
[v9] Update  `FirebaseSharedSwift/CHANGELOG.md` for v9

ncooke3:
Following offline discussion, it was decided that FirebaseSharedSwift should not have had `-beta` tagging in it's CHANGELOG.

### DIFF
--- a/FirebaseSharedSwift/CHANGELOG.md
+++ b/FirebaseSharedSwift/CHANGELOG.md
@@ -1,3 +1,3 @@
-# 8.11.0-beta
+# 8.11.0
 - Introduced shared Codable implementation. Initially used by Firebase Functions
   and Firebase Database. (#9091)


FirebaseSharedSwift/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 9.0.0
+- [added] **Breaking change:** `FirebaseSharedSwift` has exited beta.
```
ryanwilson, line 2: Did we actually make any breaking API changes?

ncooke3: Hmm no. I thought the "breaking change" would be due to removing the `-beta` suffix from the tag but it doesn't look like we ever had a `-beta` suffix for this library.

ncooke3:
```suggestion
- [added] `FirebaseSharedSwift` has exited beta.
```

paulb777 (commented):
As discussed, delete this addition and remove the `-beta` below.

morganchen12:
For the release notes on devsite, these should go at the top like FirebaseCore right?

paulb777:
Probably not, any release notes implications should be captured in the release notes for the product depending on FirebaseSharedSwift.

PR #9675, opened 2022-04-20
Fix a bug with limitToLast and cursors

wu-hui:


cherylEnkidu (approved):
Thank you for your fix!

paulb777:
Added to 9.0.0 release candidate

PR #9680, opened 2022-04-20
[Infra] Rotate `ServiceAccount.json` secret used by RC integration tests

ncooke3:
b/229016700

#no-changelog


.github/workflows/remoteconfig.yml
```diff
@@ -1,5 +1,5 @@
 name: remoteconfig
-
+# Kick off CI
```
paulb777, line 2: Instead, add [scripts/gha-encrypted/RemoteConfigSwiftAPI/ServiceAccount.json.gpg](https://github.com/firebase/firebase-ios-sdk/pull/9680/files#diff-c9730ec1200bc92df7f712478f35814e3facd55cd1c5b62ca9023d8a991d594c) to paths below

ncooke3: good idea. we probably could refactor the gha-encrypted/ directory so that each product has it's own secrets directory.

Then, all workflows could have something like:

```yaml
paths:
    - .github/workflows/product_workflow.yml
    - 'scripts/gha-encrypted/product/**
    # other paths
```

ncooke3:
The console tests passed in CI so this is good to go.

.github/workflows/remoteconfig.yml
ncooke3: Done in [4d7cd11](https://github.com/firebase/firebase-ios-sdk/pull/9680/commits/4d7cd114996c5d01c526336af49d831b2b76398e)

paulb777 (approved):

PR #9683, opened 2022-04-21
[FirebaseCoreInternal] Sanitize app ID used in file system resource

ncooke3:
### Context
- The name of an app's heartbeats storage file includes the Firebase app ID. Since the Firebase app ID includes the ":" character, the filename also contains a ":". 

  This works on Apple's latest filesystem but it's safer to replace each ":" with a "_"

- In practice, this code change will cause the below diff in an app's container:
  ```diff
  📂 Application Support/
  └── 📂 google-heartbeat-storage/
  -    └── 📄 heartbeats-1:123456789000:ios:abcdefghijklmnop
  +    └── 📄 heartbeats-1_123456789000_ios_abcdefghijklmnop
  ```
- Also, resolved outstanding `TODO`s in FirebaseCoreInternal.

#no-changelog

### DIFF
--- a/FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
+++ b/FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
@@ -22,7 +22,6 @@ public final class HeartbeatController {
   private let heartbeatsStorageCapacity: Int = 30
   /// Current date provider. It is used for testability.
   private let dateProvider: () -> Date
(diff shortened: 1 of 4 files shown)


paulb777 (approved):
Should there be a test for no colon's in the id?

Should this PR go into Firebase 9?

paulb777 (approved):
Thanks!

chliangGoogle (approved):

PR #9690, opened 2022-04-22
[Crashlytics] Investigate Crashlytics nightly failures

ncooke3:
Locally, I was able to run see the flaky failures when running setting `-test-iterations 1000`.

My hunch is this can be fixed by draining the operation queue between tests. Unfortunately, the tests set some local state I
couldn't figure out how to reset so I'm run against a clean GHA container in each commit.

#no-changelog

### DIFF
--- a/Crashlytics/UnitTests/FIRCLSOnDemandModelTests.m
+++ b/Crashlytics/UnitTests/FIRCLSOnDemandModelTests.m
@@ -98,6 +98,9 @@
 }
 
 - (void)tearDown {
+  dispatch_sync(self.existingReportManager.operationQueue.underlyingQueue, ^{
+                    // Drain queue.
+                });
   dispatch_sync(self.managerData.dispatchQueue, ^{
                     // Drain queue.
                 });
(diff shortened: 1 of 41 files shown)


Crashlytics/UnitTests/FIRCLSOnDemandModelTests.m
```diff
@@ -98,6 +98,9 @@ - (void)setUp {
 }
 
 - (void)tearDown {
+  dispatch_sync(self.onDemandModel.operationQueue.underlyingQueue, ^{
+                    // Drain queue.
+                });
```
ncooke3, line 106: This seemed to help. It's not the only queue at play though

ncooke3:
At this point, I suspect it's the file manager being used. A fake file manager should probably be used and injected accordingly.

ncooke3:
Paul took a look into the issue and I wanted to add some of the discoveries here so this PR has all of the available context:

This addition in `- setup` made the tests run better:
image

But, running multiple times (for detecting flakiness) yielded failures here: 
image

Additionally, this was necessary to prevent a GDT error.
image

ncooke3:
Closing because this issue was addressed in another PR.

PR #9707, opened 2022-04-26
Fix RTDB query `getData` caching behavior

IanWyszynski:
This is an attempt at fixing https://github.com/firebase/firebase-ios-sdk/issues/8286. The root cause is that get doesn't update cache correctly, always calling FSyncTree [applyServerOverwriteAtPath](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDatabase/Sources/Core/FRepo.m#L572-L575), even when `getData` was called on a query.

There is an alternative method for updating cache called [applyTaggedQueryOverwriteAtPath](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseDatabase/Sources/Core/FSyncTree.m#L381). This method is [called in onDataUpdate](https://github.com/firebase/firebase-ios-sdk/blob/2eefff0239392b6fdc462ee52ebe95fc21880eee/FirebaseDatabase/Sources/Core/FRepo.m#L727-L729) when the server sends a listener overwrite for a query.

In order to use this method, we need to create an `FView` for the `getData` query. I added 4 methods:
- FSyncTree registerQuery, unregisterQuery
- FSyncPoint registerQuery, unregisterQuery
To maintain the view without adding an event registration.


FirebaseDatabase/Sources/Core/FRepo.m
```diff
@@ -533,7 +533,7 @@ - (void)getData:(FIRDatabaseQuery *)query
         }];
         return;
     }
-    [self.persistenceManager setQueryActive:querySpec];
+    NSNumber *tag = [self.serverSyncTree registerQuery:[query querySpec]];
```
IanWyszynski, line 536: I don't think get queries are tagged on the backend, but it looks like the tag is [only used retrieve the query spec](https://github.com/firebase/firebase-ios-sdk/blob/eb30b6fdd809cbfe3a9eb7f3ccf1dd46326bef2d/FirebaseDatabase/Sources/Core/FSyncTree.m#L384) in `applyTaggedQueryOverwriteAtPath`.

FirebaseDatabase/Sources/Core/FSyncPoint.m
```diff
@@ -161,14 +161,9 @@ - (FView *)getView:(FQuerySpec *)query
...
 
-/**
- * Add an event callback for the specified query
- * Returns an array of events to raise.
- */
-- (NSArray *)addEventRegistration:(id<FEventRegistration>)eventRegistration
-       forNonExistingViewForQuery:(FQuerySpec *)query
-                      writesCache:(FWriteTreeRef *)writesCache
-                      serverCache:(FCacheNode *)serverCache {
+- (void)registerQuery:(FQuerySpec *)query
```
IanWyszynski, line 164: This method does everything that `addEventRegistration` did before actually adding an event registration to the view.

FirebaseDatabase/Sources/Core/FSyncPoint.m
```diff
@@ -199,6 +205,15 @@ - (NSArray *)addEventRegistration:(id<FEventRegistration>)eventRegistration
     return [view initialEvents:eventRegistration];
 }
 
+- (BOOL)unregisterQuery:(FQuerySpec *)query {
```
IanWyszynski, line 208: This method just removes the query from the view map if no other query added an event registration to it.

There is quite a bit of logic that isn't brought over from `removeEventRegistration`. In particular, we are not:

- We are not removing all empty views from a syncpoint if the getData query is default.
- We are not generating any `FCancelEvent`s for the query since there is no event registration associated with it.

I'm least sure about this function.

IanWyszynski:
@jsdt I am hoping you could give me some early feedback on this change. I added a test that checks for the reported issue. More in progress.

FirebaseDatabase/Sources/Core/FRepo.m
```diff
@@ -533,7 +533,7 @@ - (void)getData:(FIRDatabaseQuery *)query
...
     [self.connection
```
mortenbekditlevsen, line 537: Hi @IanWyszynski ,
I just had a small question about the deleted line 536:

```
[self.persistenceManager setQueryActive:querySpec];
```

There is still a 'matching':
```
[self.persistenceManager setQueryInactive:querySpec];
```
On the new line 595.

I am a bit confused by the previous code since there is a case of an 'early return' (old line number 564) where the query would NOT be set inactive again. This could be a bug.

All in all: I don't know what the correct behavior is, but I do think that there was already something fishy about the setQueryActive and setQueryInactive and the early return, and after this change there seems to be even more of an asymmetry here. :-)

FirebaseDatabase/Sources/Core/FSyncTree.m
```diff
@@ -575,6 +620,23 @@ - (FCacheNode *)serverCacheForQuery:(FQuerySpec *)query {
...
+
+    if (maybeSyncPoint &&
+        ([query isDefault] || [maybeSyncPoint viewExistsForQuery:query])) {
+        BOOL removed = [maybeSyncPoint unregisterQuery:query];
+        if ([maybeSyncPoint isEmpty]) {
+            self.syncPointTree = [self.syncPointTree removeValueAtPath:path];
+        }
+        if (removed) {
+            [self removeTags:@[ query ]];
+            [self.persistenceManager setQueryInactive:query];
```
IanWyszynski, line 635: Note: only call setQueryInactive when we just removed the last query in the view. This is consistent with how `removeEventRegistration` works.

FirebaseDatabase/Sources/Core/FSyncPoint.h
```diff
@@ -70,5 +70,9 @@
 - (BOOL)viewExistsForQuery:(FQuerySpec *)query;
 - (BOOL)hasCompleteView;
 - (FView *)completeView;
+- (void)registerQuery:(FQuerySpec *)query
```
jsdt, line 73: Can you add comments for any new functions?

FirebaseDatabase/Tests/Integration/FIRDatabaseGetTests.m
```diff
@@ -0,0 +1,68 @@
...
+#import "FirebaseDatabase/Sources/Api/Private/FIRDatabaseQuery_Private.h"
+#import "FirebaseDatabase/Sources/Constants/FConstants.h"
+#import "FirebaseDatabase/Sources/Core/FQuerySpec.h"
+#import "FirebaseDatabase/Sources/Utilities/FUtilities.h"
+#import "FirebaseDatabase/Tests/Helpers/FEventTester.h"
+#import "FirebaseDatabase/Tests/Helpers/FIRFakeApp.h"
+#import "FirebaseDatabase/Tests/Helpers/FTestExpectations.h"
+#import "FirebaseDatabase/Tests/Helpers/FTupleEventTypeString.h"
+
+@implementation FIRDatabaseGetTests
```
jsdt, line 28: Do we have a test where we perform a get, then perform a listen before the get is finished?

FirebaseDatabase/Sources/Core/FSyncTree.m
```diff
@@ -448,6 +457,42 @@ - (NSArray *)applyTaggedServerRangeMergeAtPath:(FPath *)path
     }
 }
 
+- (NSNumber *)registerQuery:(FQuerySpec *)query {
```
jsdt, line 460: Query is used a lot, including for listeners. Should we call this something like "registerGet"?

FirebaseDatabase/Sources/Core/FSyncTree.m
```diff
@@ -448,6 +457,42 @@ - (NSArray *)applyTaggedServerRangeMergeAtPath:(FPath *)path
...
+    FPath *path = query.path;
+
+    __block BOOL foundAncestorDefaultView = NO;
+    [self.syncPointTree
+        forEachOnPath:query.path
+           whileBlock:^BOOL(FPath *pathToSyncPoint, FSyncPoint *syncPoint) {
+             foundAncestorDefaultView =
+                 foundAncestorDefaultView || [syncPoint hasCompleteView];
+             return !foundAncestorDefaultView;
```
maneesht, line 469: I don't actually think we're using this block, since we aren't checking `foundAncestorDefaultView` later in the function

IanWyszynski: Yes, will remove this.

FirebaseDatabase/Sources/Core/FSyncTree.m
```diff
@@ -448,6 +457,42 @@ - (NSArray *)applyTaggedServerRangeMergeAtPath:(FPath *)path
...
+    BOOL viewAlreadyExists = [syncPoint viewExistsForQuery:query];
+    if (!viewAlreadyExists) {
+        FWriteTreeRef *writesCache =
+            [self.pendingWriteTree childWritesForPath:path];
+        FCacheNode *serverCache = [self serverCacheForQuery:query];
+        [syncPoint registerQuery:query
+                     writesCache:writesCache
+                     serverCache:serverCache];
+    } else if (![query loadsAllData]) {
+        tag = [self tagQuery:query];
```
IanWyszynski, line 491: This needs to be moved into the 
```
if (!viewAlreadyExists) {
  ...
}
```
bock above.

PR #9731, opened 2022-04-30
Overlays integrate with mutations and queries, and migration

wu-hui:
NOTE: This PR has to be merged together with overlay migration.


Firestore/core/src/local/local_store.cc
```diff
@@ -76,6 +79,21 @@ using remote::TargetChange;
  */
 const int64_t kResumeTokenMaxAgeSeconds = 5 * 60;  // 5 minutes
 
+DocumentKeySet GetKeysWithTransformResults(
+    const MutationBatchResult& batch_result) {
+  DocumentKeySet result;
+
+  for (size_t i = 0; i < batch_result.mutation_results().size(); ++i) {
+    if (batch_result.mutation_results()[i]
```
dconeybe, line 87: nit: Consider saving `batch_result.mutation_results()[i]` to a local variable to avoid retrieving it twice, and improving readability a little bit.

Firestore/core/src/local/local_store.cc
```diff
@@ -161,22 +180,36 @@ LocalWriteResult LocalStore::WriteLocally(std::vector<Mutation>&& mutations) {
...
     // record the base state in a separate patch mutation. This is later used to
     // guarantee consistent values and prevents flicker even if the backend
     // sends us an update that already includes our transform.
     std::vector<Mutation> base_mutations;
     for (const Mutation& mutation : mutations) {
-      absl::optional<Document> base_document =
-          existing_documents.get(mutation.key());
-
+      auto it = overlayed_documents.find(mutation.key());
+      HARD_ASSERT(it != overlayed_documents.end(),
```
dconeybe, line 209: Consider including the key value into the assertion message; this will help with debugging if this assertion were to ever fail.

Firestore/core/src/local/local_write_result.h
```diff
@@ -31,6 +31,17 @@ namespace local {
 /** The result of a write to the local store. */
 class LocalWriteResult {
  public:
+  static LocalWriteResult FromOverlayedDocuments(
+      model::BatchId batch_id, model::OverlayedDocumentMap&& overlayed_docs) {
+    model::DocumentMap docs;
+    for (const auto& overlayed_doc : overlayed_docs) {
+      docs = docs.insert(overlayed_doc.first,
+                         std::move(overlayed_doc.second.document()));
```
dconeybe, line 39: Remove the `std::move` since it has no effects because OverlayedDocumentMap only supports const iterators.

Firestore/core/src/model/mutation.cc
```diff
@@ -99,6 +99,16 @@ Mutation::Rep::Rep(DocumentKey&& key,
...
 }
 
+Mutation::Rep::Rep(DocumentKey&& key,
+                   Precondition&& precondition,
+                   std::vector<FieldTransform>&& field_transforms,
+                   absl::optional<FieldMask>&& mask)
+    : key_(std::move(key)),
+      precondition_(std::move(precondition)),
+      field_transforms_(std::move(field_transforms)),
+      mask_(mask) {
```
dconeybe, line 109: You should `std::move` the `mask` as well.

dconeybe (changes requested):

Firestore/core/src/local/document_overlay_cache.cc
```diff
@@ -0,0 +1,46 @@
...
+
+namespace firebase {
+namespace firestore {
+namespace local {
+
+using model::DocumentKey;
+using model::DocumentKeySet;
+using model::Overlay;
+
+void DocumentOverlayCache::GetOverlays(
```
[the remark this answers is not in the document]
wu-hui: Or three.

Firestore/core/src/local/memory_remote_document_cache.cc
```diff
@@ -87,7 +87,8 @@ MutableDocumentMap MemoryRemoteDocumentCache::GetAll(
       break;
     }
     const MutableDocument& document = it->second;
-    if (key.path().size() > path.size() + 1) {
+    size_t immediate_children_path_length = path.size() + 1;
```
dconeybe, line 90: Please define `immediate_children_path_length` outside of the `for` loop to avoid recalculating it all the time. This also improves readability because its "const-ness" is correctly implied by being outside of the `for` loop.

Firestore/core/src/model/overlayed_document.h
```diff
@@ -40,6 +40,10 @@ class OverlayedDocument {
     return document_;
   }
 
+  model::Document consume_document() {
```
dconeybe, line 43: Was there a reason not to just overload `document()` on rvalue reference?

Firestore/core/src/local/local_store.cc
```diff
@@ -207,7 +207,8 @@ LocalWriteResult LocalStore::WriteLocally(std::vector<Mutation>&& mutations) {
     for (const Mutation& mutation : mutations) {
       auto it = overlayed_documents.find(mutation.key());
       HARD_ASSERT(it != overlayed_documents.end(),
-                  "Failed to find overlayed document with a mutation key");
+                  "Failed to find overlayed document with mutation key: ",
```
dconeybe, line 210: nit: Add `%s` to the assertion message so that the string will actually be included (I think HARD_ASSERT takes a printf-style formatting string, but I could be wrong about that).

dconeybe (changes requested):
I still need to do that "high-level" review based on our meeting today, but here are some small follow-up nits.

Firestore/core/src/model/overlayed_document.h
wu-hui: The truth is I had problems to make it work when I tried. I tried again today and it worked..but it got me curious so I checked style guide: go/cstyle#Rvalue_references.

My read is that we should favor pass/return values instead of rvalues unless we can prove this helps with performance? I also think `consume_document()` states the intent better.

Firestore/core/src/local/local_store.cc
wu-hui: Fixed.

Firestore/core/src/model/overlayed_document.h
dconeybe: I tend to disagree. I'll just give my rationale and you can make the final decision.

go/cstyle#Rvalue_references doesn't address overloading based on whether or not the method is invoked on an rvalue reference. It's talking more about method _arguments_.

Overloading on an rvalue reference makes it _automatic_ to move the object instead of requiring the caller at the call site to be sure to invoke the correct method. The compiler will pick the `&&` overload if invoked on a temporary object and the `const&` method if invoked on a non-temporary object.

Example 1: This will call the `const&` overload of `document()`:

```
void MyFunc(const OverlayedDocument& overlayed_document) {
  printf("The overlayed document is: %s", overlayed_document.document());
}
```

Example 2: This will call the `&&` overload of `document()`:
```
void MyFunc(OverlayedDocument&& overlayed_document) {
  overlayed_documents_.push_back(std::move(overlayed_document).document());
}
```

Example 2 is effectively the same as calling

```
overlayed_documents_.push_back(overlayed_document.consume_document());
```

because `consume_document()` will return a temporary object; however, the std::move notation is standard C++11 notation instead of writing a specially-named method. This same notation can be used to move objects out of absl::optional objects:

```
void SnapshotCreator::UpdateSnapshot(absl::optional<OverlayedDocument> doc) {
  if (doc.has_value()) {
    docs_.push_back(std::move(doc).value());
  }
}
```

In this example, `absl::optional` overloads the `value()` method on `&&` to move the object out of the optional if invoked on a temporary object. There is no special "consume_value()" method.

Firestore/core/src/remote/grpc_root_certificates_generated.h
```diff
@@ -3,7 +3,7 @@
 #ifndef FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 #define FIRESTORE_CORE_SRC_REMOTE_GRPC_ROOT_CERTIFICATES_GENERATED_H_
 
-#include <stddef.h>
+#include <cstdlib>
```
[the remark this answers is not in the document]
dconeybe: It appears that this unrelated change has come back. Please make sure it's NOT part of your final merge.

Firestore/core/test/unit/local/counting_query_engine.h
```diff
@@ -77,8 +77,8 @@ class CountingQueryEngine : public QueryEngine {
    * `getAllMutationBatchesAffectingQuery()` API (since the last call to
    * `ResetCounts()`)
    */
-  size_t mutations_read_by_query() const {
-    return mutations_read_by_query_;
+  size_t overlays_read_by_query() const {
```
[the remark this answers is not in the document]
dconeybe: Indeed.

Firestore/core/src/local/local_documents_view.cc
```diff
@@ -141,79 +117,210 @@ model::DocumentMap LocalDocumentsView::GetDocumentsMatchingCollectionGroupQuery(
...
-    if (query.Matches(kv.second)) {
-      results = results.insert(key, kv.second);
+model::OverlayedDocumentMap LocalDocumentsView::ComputeViews(
+    MutableDocumentMap docs,
+    DocumentOverlayCache::OverlayByDocumentKeyMap&& overlays,
+    const DocumentKeySet& existence_state_changed) {
+  model::MutableDocumentPtrMap recalculate_documents;
+  model::FieldMaskMap mutated_fields;
+  for (const auto& docs_entry : docs) {
+    auto* doc = const_cast<MutableDocument*>(&(docs_entry.second));
```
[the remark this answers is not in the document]
dconeybe: Ahh okay, makes sense.

Firestore/core/src/local/local_documents_view.cc
```diff
@@ -141,79 +119,212 @@ model::DocumentMap LocalDocumentsView::GetDocumentsMatchingCollectionGroupQuery(
...
+    const DocumentKeySet& existence_state_changed) {
+  OverlayByDocumentKeyMap overlays;
+  PopulateOverlays(overlays, DocumentKeySet::FromKeysOf(base_docs));
+  auto overlayed_documents =
+      ComputeViews(base_docs, std::move(overlays), existence_state_changed);
+
+  DocumentMap result;
+  for (auto& entry : overlayed_documents) {
+    result =
+        result.insert(entry.first, std::move(entry.second.consume_document()));
```
dconeybe, line 185: Remove std::move here... it is not necessary because `consume_document()` returns a temporary.

Just for completeness, I'd like to reiterate my preference for `std::move(entry.second).document()` here :)

Firestore/core/src/local/local_documents_view.cc
```diff
@@ -141,79 +117,210 @@ model::DocumentMap LocalDocumentsView::GetDocumentsMatchingCollectionGroupQuery(
...
-    for (const Mutation& mutation : batch.mutations()) {
-      // Only process documents belonging to the collection.
-      if (!query.path().IsImmediateParentOf(mutation.key().path())) {
-        continue;
-      }
+  // Apply the overlays and match against the query.
+  DocumentMap results;
+  for (const auto& entry : remote_documents) {
+    const auto& key = entry.first;
+    auto doc = std::move(entry.second);
```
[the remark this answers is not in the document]
dconeybe: I just chatted with @schmidt-sebastian about this as well. He also supports the idea of moving away from SortedSet/SortedMap. I'm considering a hackweek project to migrate away from SortedSet/SortedMap for this reason.

dconeybe (changes requested):

Firestore/core/src/local/local_documents_view.cc
wu-hui: I really do not have a strong preference, but I see the similarities with optional now, so let's make it rvalue ref.

Firestore/core/src/model/overlayed_document.h
dconeybe: I just pushed up a commit to do this.

Firestore/core/src/local/local_documents_view.cc
dconeybe: I pushed up a commit to tweak to add the proper lvalue/rvalue ref-qualified overloads of `document()`.

dconeybe (approved):

Firestore/core/src/local/leveldb_persistence.cc
```diff
@@ -259,9 +275,33 @@ LevelDbBundleCache* LevelDbPersistence::bundle_cache() {
...
+  if (document_overlay_caches_.find(user.uid()) ==
+      document_overlay_caches_.end()) {
+    document_overlay_caches_.insert(
+        {user.uid(), absl::make_unique<LevelDbDocumentOverlayCache>(
+                         user, this, &serializer_)});
+  }
+  return document_overlay_caches_[user.uid()].get();
+}
+
+LevelDbOverlayMigrationManager* LevelDbPersistence::GetOverlayMigrationManager(
```
dconeybe, line 287: If `GetOverlayMigrationManager()` is invoked with a different user than a previous invocation it's going to return the `LevelDbOverlayMigrationManager` instance associated with the _old_ user (the caller likely requires that the returned `LevelDbOverlayMigrationManager` instance is associated with the _new_ user).

All of the other methods in this class that take a `const credentials::User& user` argument also have a dictionary mapping user to the instance of the type being returned. Maybe this method should do that too?

Firestore/core/src/local/memory_overlay_migration_manager.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "Firestore/core/src/local/overlay_migration_manager.h"
+
+namespace firebase {
+namespace firestore {
+namespace local {
+
+class MemoryOverlayMigrationManager : public OverlayMigrationManager {
+ public:
+  void Run() override {
```
dconeybe, line 28: Optional: Can this just be inlined into the class definition in the header file?

Firestore/core/src/local/persistence.h
```diff
@@ -133,6 +142,10 @@ class Persistence {
    */
   virtual ReferenceDelegate* reference_delegate() = 0;
 
+  /** Releases components created for users other than given `target_uid`. */
```
dconeybe, line 145: Please update this documentation to also specify under which conditions it is appropriate to invoke this new `ReleaseOtherUserSpecificComponents()` method.

dconeybe (changes requested):

Firestore/core/src/local/persistence.h
wu-hui: Done.

dconeybe (approved):

PR #9734, opened 2022-05-01
Implement the Composite Filter class

cherylEnkidu:
#no-changelog

- Implement the Composite Filter class.
- Perform parsing of AND/OR filters to create Composite Filter objects.
- Add filter validation for composite filters.
- Update core/query and api/query_core to handle composite filters.

### DIFF
--- a/Firestore/Source/API/FIRFilter+Internal.h
+++ b/Firestore/Source/API/FIRFilter+Internal.h
@@ -15,7 +15,8 @@
  */
 
 #import <Foundation/Foundation.h>
-#include "Firestore/core/src/core/field_filter.h"
+#include "Firestore/Protos/nanopb/google/firestore/v1/query.nanopb.h"
+#import "Firestore/core/src/core/field_filter.h"
 
 // TODO(orquery): This class will become public API. Change visibility and add documentation.
 
@@ -93,14 +94,26 @@ NS_SWIFT_NAME(Filter)
                               notIn:(nonnull NSArray<id> *)values
     NS_SWIFT_NAME(whereField(_:notIn:));
 
++ (FIRFilter *)orFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(orFilter(_:));
+
++ (FIRFilter *)andFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(andFilter(_:));
+
 @end
 
-/** Internal FIRFilter properties we don't want exposed in our public header files. */
-@interface FIRFilter (Internal)
+/** Exposed internally */
+@interface FSTUnaryFilter : FIRFilter
 
 @property(nonatomic, strong, readonly) FIRFieldPath *fieldPath;
-@property(nonatomic, readonly) firebase::firestore::core::FieldFilter::Operator op;
+@property(nonatomic, readonly) firebase::firestore::core::FieldFilter::Operator unaryOp;
 @property(nonatomic, strong, readonly) id value;
+
+@end
+
+@interface FSTCompositeFilter : FIRFilter
+
+@property(nonatomic, strong, readonly) NSArray<FIRFilter *> *filters;
+@property(nonatomic, readonly)
+    firebase::firestore::_google_firestore_v1_StructuredQuery_CompositeFilter_Operator compOp;
 
 @end
 
--- a/Firestore/Source/API/FIRFilter+Internal.h
+++ b/Firestore/Source/API/FIRFilter+Internal.h
@@ -16,6 +16,7 @@
 
 #import <Foundation/Foundation.h>
 #include "Firestore/Protos/nanopb/google/firestore/v1/query.nanopb.h"
+#import "Firestore/core/src/core/composite_filter.h"
 #import "Firestore/core/src/core/field_filter.h"
 
 // TODO(orquery): This class will become public API. Change visibility and add documentation.
@@ -94,9 +95,9 @@ NS_SWIFT_NAME(Filter)
                               notIn:(nonnull NSArray<id> *)values
     NS_SWIFT_NAME(whereField(_:notIn:));
 
-+ (FIRFilter *)orFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(orFilter(_:));
++ (FIRFilter *)orFilterWithFilters:(NSArray<FIRFilter *> *)filters NS_SWIFT_NAME(orFilter(_:));
 
-+ (FIRFilter *)andFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(andFilter(_:));
++ (FIRFilter *)andFilterWithFilters:(NSArray<FIRFilter *> *)filters NS_SWIFT_NAME(andFilter(_:));
 
 @end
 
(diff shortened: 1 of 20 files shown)


Firestore/core/src/core/composite_filter.cc
```diff
@@ -0,0 +1,162 @@
...
+  for (const auto& field_filter : GetFlattenedFilters()) {
+    if (condition(field_filter)) {
+      return &field_filter;
+    }
+  }
+  return nullptr;
+}
+
+const model::FieldPath* CompositeFilter::Rep::GetFirstInequalityField() const {
+  CheckingFun condition = [](const FieldFilter& field_filter_ptr) {
```
ehsannas, line 138: it's not really a pointer. so `field_filter` is probably a better variable name.

Firestore/core/src/core/composite_filter.cc
```diff
@@ -0,0 +1,162 @@
...
+  // and it does not remove duplicate subfilters from each composite filter.
+  // It is therefore way less expensive.
+  // TODO(orquery): Consider removing duplicates and ignoring order of filters
+  // in the list.
+  return op_ == other_rep.op_ && filters_ == other_rep.filters_;
+}
+
+const FieldFilter* CompositeFilter::Rep::FindFirstMatchingFilter(
+    CheckingFun& condition) const {
+  for (const auto& field_filter : GetFlattenedFilters()) {
```
ehsannas, line 129: We should update the Android SDK to use `getFlattenedFilters`.

Firestore/core/src/core/filter.h
```diff
@@ -120,10 +143,20 @@ class Filter {
...
+
+    virtual const model::FieldPath* GetFirstInequalityField() const = 0;
+
+    virtual const std::vector<FieldFilter>& GetFlattenedFilters() const = 0;
+
+    /**
+     * Memoized list of all field filters that can be found by
+     * traversing the tree of filters contained in this composite filter.
+     */
+    mutable std::vector<FieldFilter> memoized_flatten_filters_;
```
ehsannas, line 159: s/`memoized_flatten_filters_`/`memoized_flattened_filters_`

Firestore/core/src/core/composite_filter.h
```diff
@@ -0,0 +1,138 @@
...
+
+    bool IsEmpty() const override {
+      return filters_.empty();
+    }
+
+    const std::vector<FieldFilter>& GetFlattenedFilters() const override;
+
+    const model::FieldPath* GetFirstInequalityField() const override;
+
+    const FieldFilter* FindFirstMatchingFilter(CheckingFun& condition) const;
```
ehsannas, line 118: Please add the comment from the Android SDK.

Firestore/core/src/core/filter.h
```diff
@@ -88,8 +92,24 @@ class Filter {
     return rep_->ToString();
   }
 
-  size_t Hash() const {
-    return rep_->Hash();
+  /** Check whether the filter is meaningful */
```
ehsannas, line 95: `Returns true if and only if the filter is a composite filter that doesn't contain any field filters.`

Firestore/core/src/core/filter.h
```diff
@@ -88,8 +92,24 @@ class Filter {
...
+  /**
+   * Returns the first inequality filter contained within this filter.
+   * Returns nullptr if it does not contain any inequalities.
+   */
+  const model::FieldPath* GetFirstInequalityField() const {
+    return rep_->GetFirstInequalityField();
+  }
+
+  /**
+   * Returns a list of all field filters that are contained within this filter
```
ehsannas, line 109: add `.` at the end of the line.

Firestore/core/src/core/query.cc
```diff
@@ -52,6 +52,15 @@ Query::Query(ResourcePath path, std::string collection_group)
...
 }
 
+Query Query::Clone() const {
+  Query copy(*this);
+  copy.filters_ = immutable::AppendOnlyList<Filter>();
+  for (const Filter& filter : filters_) {
+    copy.filters_ = copy.filters_.push_back(filter);
+  }
+  return copy;
+}
```
ehsannas, line 62: is this function really needed? what does the copy c'tor and the assignment operator do for `immutable::AppendOnlyList`?

Firestore/core/src/core/target.cc
```diff
@@ -68,10 +68,13 @@ bool Target::IsDocumentQuery() const {
...
-
   for (const Filter& filter : filters_) {
-    if (filter.IsAFieldFilter() && filter.field() == path) {
-      result.push_back(FieldFilter(filter));
+    if (filter.IsAFieldFilter()) {
+      FieldFilter field_filter(filter);
+      if (field_filter.field() != path) {
+        continue;
+      }
+      result.push_back(std::move(field_filter));
```
ehsannas, line 77: why not:
```cpp
fo(...) {
  if (filter.IsAFieldFilter()) {
    FieldFilter field_filter(filter);
    if (field_filter.field() == path) {
      result.push_back(std::move(field_filter));
    }
  }
}
```

ehsannas (changes requested):
Good job! Generally looks good to me.

There is this Android PR (https://github.com/firebase/firebase-android-sdk/pull/3609/files) that was added to Android to fix a bug. You've already captured it in your PR here, except for the tests. Can you please add the tests too?

Firestore/core/src/core/composite_filter.cc
ehsannas: I filed an issue for this.

ehsannas:
The way the "operator" for composite filters was defined was by re-using the proto definition. It was cumbersome and inconsistent with what was done for field filters. I've made a PR (https://github.com/firebase/firebase-android-sdk/pull/3824) to fix this. Please update this PR to do the same thing. I should have made this fix earlier. Sorry for the delay.

Firestore/Source/API/FIRFilter+Internal.h
```diff
@@ -93,15 +94,27 @@ NS_SWIFT_NAME(Filter)
...
++ (FIRFilter *)orFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(orFilter(_:));
+
++ (FIRFilter *)andFilterWithFilters:(NSArray *)filters NS_SWIFT_NAME(andFilter(_:));
+
 @end
 
-/** Internal FIRFilter properties we don't want exposed in our public header files. */
-@interface FIRFilter (Internal)
+/** Exposed internally */
+@interface FSTUnaryFilter : FIRFilter
```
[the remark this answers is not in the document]
cherylEnkidu: `FIR` is used as the prefix of public class and `FST` is the prefix of internal class. Since `FSTUnaryFilter` and `FSTCompositeFilter` is hidden from public, I use `FST` as prefix.

Firestore/core/src/core/query.cc
cherylEnkidu: Use tests to verify the behaviour of  `immutable::AppendOnlyList`. Changing the filter inside `test_query` won't affect `query_`. Will delete this method and use copy c'tor instead.

ehsannas (approved):

PR #9740, opened 2022-05-02
[Infra] Run `check` when `check.yml` changes

ncooke3:
### Context
- It's a good practice to use the `on / pull_request / paths` attribute to trigger a workflow when it's `yml` file changes.
- Brought up in https://github.com/google/GoogleDataTransport/pull/58#discussion_r862960389

#no-changelog

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,8 +2,11 @@ name: check
 
 on:
   pull_request:
+    paths:
+    - '.github/workflows/check.yml'
     paths-ignore:
     - 'Firestore/**'
+
   push:
     branches: master
 


.github/workflows/check.yml
```diff
@@ -2,8 +2,11 @@ name: check
 
 on:
   pull_request:
+    paths:
+    - '.github/workflows/check.yml'
     paths-ignore:
     - 'Firestore/**'
```
ncooke3, line 8: Oh, so it looks like `check` was already set up to run `check` on `check.yml` changes.

GHA failed with: `you may only define one of `paths` and `paths-ignore` for a single event`

PR #9750, opened 2022-05-03
Fix memory leak from ASAN checks

wu-hui:
Newer version of ASAN started to support checking memory leaks. 

* `MaxValue()` was leaking some heap memory because the special field value is allocated on heap and no one frees it. It's a negligible leak, but I fixed so we have clean ASAN by allocating the entire `kMaxValue` in static memory.
* Some tests in local_serializer_tests.cc are leaking memory, they happen to be tests we intend to remove, so I simply removed them. There are some production code we need to remove as well (local_store->DecodeMutationBatch), but I'll leave that for later.


Firestore/core/src/model/value_util.cc
```diff
@@ -52,6 +52,16 @@ pb_bytes_array_s* kMaxValueFieldKey =
 const char* kRawMaxValueFieldValue = "__max__";
 pb_bytes_array_s* kMaxValueFieldValue =
     nanopb::MakeBytesArray(kRawMaxValueFieldValue);
+google_firestore_v1_MapValue_FieldsEntry kMaxValueFieldEntry = {
```
ehsannas, line 55: Nit: Add an empty line above with a comment.

ehsannas (approved):

PR #9751, opened 2022-05-03
Restore check_imports GDT exception

paulb777:
Restore exception removed in #9660 that is still needed in the GDT repo


ncooke3 (approved):
thanks for taking a closer look!

PR #9761, opened 2022-05-04
Add check for zipfile to "update C++ SDK" workflow.

jonsimantov:
Because the C++ SDK's update script now requires the zipfile to be present, modify the "update C++ SDK" workflow to ensure that the zip file is present. This should almost always be the case, except in a rare occasion when the GitHub release is created and the zip file added to it afterwards. This will wait up to an hour for the zip file to be available, similar to how it waits up to an hour for the Cocoapod repo to update.

Also, following Postel's Law (accept liberally, emit conservatively), the workflow now handles iOS SDK release version numbers with or without the leading 'v'.

### DIFF
--- a/.github/workflows/update-cpp-sdk-on-release.yml
+++ b/.github/workflows/update-cpp-sdk-on-release.yml
@@ -41,24 +41,41 @@ jobs:
 
       - name: Ensure Cocoapods repo has been updated
         run: |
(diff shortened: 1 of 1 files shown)


ncooke3 (approved):

ncooke3:
Hi @jonsimantov, is this good to merge?

jonsimantov:
Oh whoops, I thought this got merged already. :)

PR #9765, opened 2022-05-04
Update to nanopb 2.30909.0

### LINKED ISSUE #9587: Analyze error: Dereference of null pointer - Update Firebase when nanopb releases fix
I have installed via SPM Analytics and Crashlytics. There are 2 warnings coming up:

image

They are flagged as logic error:

image

paulb777:
Update to nanopb 0.3.9.9 (semantic 2.30909.0) 

Fix #9587

### DIFF
--- a/GoogleAppMeasurement.podspec.json
+++ b/GoogleAppMeasurement.podspec.json
@@ -7,7 +7,7 @@
         "GoogleUtilities/MethodSwizzler": "~> 7.7",
         "GoogleUtilities/NSData+zlib": "~> 7.7",
         "GoogleUtilities/Network": "~> 7.7",
-        "nanopb": "~> 2.30908.0"
+        "nanopb": "~> 2.30909.0"
     },
     "description": "Measurement methods that are shared between Google libraries. This pod does not expose any headers and isn't intended for direct use, but rather as a dependency of some Google libraries.",
     "frameworks": [
--- a/GoogleAppMeasurement.podspec.json
+++ b/GoogleAppMeasurement.podspec.json
@@ -7,7 +7,7 @@
         "GoogleUtilities/MethodSwizzler": "~> 7.7",
         "GoogleUtilities/NSData+zlib": "~> 7.7",
         "GoogleUtilities/Network": "~> 7.7",
-        "nanopb": "~> 2.30909.0"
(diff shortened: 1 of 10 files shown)


paulb777:
@ncooke3 The remaining  CI failures should be addressed by publishing 9.1.0 podspecs to SpecsStaging. Please merge before doing the version update.

ncooke3:
I was thinking I wait to merge #9775 and #9774. Then merge this PR. And then stage 9.1.0 podspecs to SpecsStaging. WDYT? @paulb777

paulb777:
SGTM

GoogleAppMeasurement.podspec.json
```diff
@@ -7,7 +7,7 @@
         "GoogleUtilities/MethodSwizzler": "~> 7.7",
         "GoogleUtilities/NSData+zlib": "~> 7.7",
         "GoogleUtilities/Network": "~> 7.7",
-        "nanopb": "~> 2.30908.0"
+        "nanopb": "~> 2.30909.0"
```
ncooke3, line 10: https://github.com/google/GoogleAppMeasurement/pull/27 made it so that the SPM supports `2.30908.0` and `2.30909.0`.

Would we want to do the same here?

paulb777: Yes. Good catch! 

A `git grep` on the CocoaPods Specs repo shows ARCore is also tied to 2.30908.* so this would break compatibility with them.

Package.swift
```diff
@@ -170,7 +170,7 @@ let package = Package(
     .package(
       name: "nanopb",
       url: "https://github.com/firebase/nanopb.git",
-      "2.30908.0" ..< "2.30909.0"
+      "2.30909.0" ..< "2.30910.0"
```
ncooke3, line 173: Sorry, one last question. Just double checking. Is it intended to be slightly stricter for SPM. (We aren't strict in GoogleAppMeasurement manifest)?

paulb777: Yes. We want to force SPM users to get the fix, since SPM would otherwise be more indeterministic about updating. Also I would be very surprised if there are any other SDKs that would need to couple with Firebase and an old version of nanopb

ncooke3 (approved):

ncooke3:
Approved, merged, and staging specs now

PR #9774, opened 2022-05-06
[v9.1] Bump GTMSessionFetcher dependency to >1.7.2

ncooke3:
Pinning to a patch version means we cannot use CocoaPod's `~>` operator anymore for GTMSessionFetcher.

```diff
-  s.dependency 'GTMSessionFetcher/Core', '~> 1.7'
+  s.dependency 'GTMSessionFetcher/Core', '>= 1.7.2', '< 1.8.0'
```

#no-changelog

Fixes #9753

### DIFF
--- a/FirebaseAuth.podspec
+++ b/FirebaseAuth.podspec
@@ -54,7 +54,7 @@ supports email and password accounts, as well as several 3rd party authenticatio
   s.dependency 'FirebaseCore', '~> 9.0'
   s.dependency 'GoogleUtilities/AppDelegateSwizzler', '~> 7.7'
   s.dependency 'GoogleUtilities/Environment', '~> 7.7'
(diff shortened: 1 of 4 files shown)


ryanwilson (approved):

ncooke3:
CI is green so I'm merging

paulb777 (commented):
I'm not sure we should change the podspecs on this one since all 1.7.x versions should continue to work - with only a log spam issue.

We should change the Package.swift to make sure that we force the update for SPM users since it doesn't update as seamlessly as CocoaPods.

ncooke3:
@paulb777, sorry missed your review by a hair! 

> since all 1.7.x versions should continue to work

That's a good point. Let me rollback.

PR #9776, opened 2022-05-06
[v9.1] Partial revert "[v9.1] Bump GTMSessionFetcher dep to >1.7.2 (#9774)"

ncooke3:
### Context
- Addresses https://github.com/firebase/firebase-ios-sdk/pull/9774#pullrequestreview-965220567
- `Package.swift` in `master`: https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L165-L169

#no-changelog


paulb777 (approved):

ncooke3:
Merging before CI finishes because this is a revert.

PR #9777, opened 2022-05-06
[Infra] Standardize CHANGELOG header format

ncooke3:
### Context
- The idea here is to standardize the CHANGELOG header format across the repo for two reasons.
1. The CHANGELOGs look nicer when they all have same format. (less important reason) 
2. Since all CHANGELOGs now have same format, we can easily run a `sed` command during release process to replace `# Unreleased` with the shared format (more important reason)

I verified with @morganchen12 that changing old CHANGELOG entries will not mess with docs tooling.

> The FirebaseCore CHANGELOG uses the `# Firebase X.Y.Z` format and the other formats use `# X.Y.Z`.

Example command where I transitioned from `vX.Y.Z` → `X.Y.Z`
```shell
find ./ -name '*CHANGELOG*' -not -path '*/FirebaseCore/*' -not -path '*/build/*' -not -path '*/.build/*' -not -path '*/Pods/*' | xargs sed -i "" "s/# v/# /g"  
```

#no-changelog

### DIFF
--- a/FirebaseAppDistribution/CHANGELOG.md
+++ b/FirebaseAppDistribution/CHANGELOG.md
@@ -1,24 +1,24 @@
-# v9.0.0-beta
+# 9.0.0-beta
 - [fixed] Marked `releaseNotes` as `nullable` as they don't always exist (#8602).
 - [fixed] **Breaking change:** Fixed an ObjC-to-Swift API conversion error where`checkForUpdate()`
   returned a non-optional type. This change is breaking for Swift users only (#9604).
 
-# v8.3.0-beta
+# 8.3.0-beta
 - [changed] Sign out the Tester when the call to fetch releases fails with an unauthorized error (#8270).
 
-# v7.3.0-beta
+# 7.3.0-beta
 - [changed] Sign out the Tester when the call to fetch releases fails with an unauthenticated error.
 - [fixed] Crash caused by trying to parse response as JSON when response is nil (#6996).
 
-# v0.9.3-beta
+# 0.9.3-beta
 - [changed] Updated error log for non-200 API Service calls.
 
-# v0.9.2-beta
+# 0.9.2-beta
 - [fixed] Made bug fixes (#6346) available in Zip build and Carthage.
 
-# v0.9.1-beta
+# 0.9.1-beta
 - [changed] Updated header comments (#6321).
 - [fixed] Bug for customers with restricted API keys unable to fetch releases (#6346).
 
-# v0.9.0-beta
+# 0.9.0-beta
 - Initial beta release.
--- a/FirebaseAppDistribution/CHANGELOG.md
+++ b/FirebaseAppDistribution/CHANGELOG.md
@@ -1,24 +1,24 @@
-# Firebase 9.0.0-beta
+# v9.0.0-beta
 - [fixed] Marked `releaseNotes` as `nullable` as they don't always exist (#8602).
 - [fixed] **Breaking change:** Fixed an ObjC-to-Swift API conversion error where`checkForUpdate()`
   returned a non-optional type. This change is breaking for Swift users only (#9604).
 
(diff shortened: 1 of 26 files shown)


paulb777 (commented):
Thanks for taking this on! It will be really nice to standardize these.

I'm not sure about using the Firebase prefix everywhere. We were previously using Firebase in the FirebaseCore Changelog to distinguish Firebase-wide changes versus FirebaseCore specific ones. 

Also I'm not sure "Firebase" adds anything elsewhere since we already know we're in Firebase when we get here.

ncooke3:
Thanks @paulb777! Agreed, I prefer `X.Y.Z` or `vX.Y.Z` because they are shorter.

I've changed them all but the `FirebaseCore/CHANGELOG` to use the `vX.Y.Z` format. It's very trivial to change them now that I have them all standardized.

I think it's reasonable to have all CHANGELOGs use one format with the exception of the `FirebaseCore/CHANGELOG` that uses the `Firebase X.Y.Z` format.

paulb777 (approved):

FirebaseCore/CHANGELOG.md
```diff
@@ -170,50 +170,50 @@
 # Firebase 6.31.1
 - [fixed] Sporadic missing FirebaseApp symbol build issue introduced in Firebase 6.28.0. (#6341)
 
-# Firebase 6.31.0 FirebaseCore 6.10.1 -- M78
```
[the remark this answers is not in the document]
ncooke3: SG. Done in [f8208d9](https://github.com/firebase/firebase-ios-sdk/pull/9777/commits/f8208d9de6cf0c1bec1c30f812cdbb13dfd9cc1f)

FirebaseAppDistribution/CHANGELOG.md
```diff
@@ -10,15 +10,15 @@
 - [changed] Sign out the Tester when the call to fetch releases fails with an unauthenticated error.
 - [fixed] Crash caused by trying to parse response as JSON when response is nil (#6996).
 
-# v0.9.3
+# v0.9.3-beta
```
ryanwilson, line 13: Hmm, here and below this tag doesn't actually exist (it was `AppDistribution-0.9.3`). I think it's reasonable to remove the `beta` and just leave it as is since the `0` indicates beta already.

FirebaseInAppMessaging/CHANGELOG.md
```diff
@@ -1,114 +1,114 @@
...
+# v8.6.0
 - [changed] Replaced conditionally-compiled APIs with `API_UNAVAILABLE` annotations on unsupported platforms (#8480).
 
-# 2021-8 -- v.8.5.0
+# v.8.5.0
 - [added] Added support for unit testing with in-app message data objects (#8351).
 - [added] Added support for prototyping custom in-app message views in SwiftUI (#8351).
 
-# 2021-7 --- v8.4.0
+# 7 --- v8.4.0
```
ryanwilson, line 15: 7 --- should be removed

ryanwilson (commented):
I also +1 not having the `v` prefix so it matches our tags moving forward.

FirebaseInAppMessaging/CHANGELOG.md
ncooke3: Done here and below in [f16b82e](https://github.com/firebase/firebase-ios-sdk/pull/9777/commits/f16b82eafaa5bc7323b37ff85727173e980f76c6)

ryanwilson (approved):

FirebaseAppDistribution/CHANGELOG.md
ncooke3: Done in [60b3089](https://github.com/firebase/firebase-ios-sdk/pull/9777/commits/60b30895f4e5b2d68dda1657dba38903730bfe4f)

PR #9783, opened 2022-05-09
[Release Tooling] Re-order pod manifest

ncooke3:
### Context
- FirebaseCore pods were being pushed in incorrect order.
- `FirebaseCoreExtension` depends on `FirebaseCore`, which depends on `FirebaseCoreInternal` so they so should be pushed in reverse order
- The interop pods don't depend on `FirebaseCore*` pods and I have them being pushed after `FirebaseCore*` pods. I think this should a buffer for the `FirebaseCore*` to be fully published by the time the product pods start being pushed

#no-changelog


paulb777 (approved):
Thanks!

PR #9787, opened 2022-05-09
[v9.1.0] Update versions for Release 9.1.0

ncooke3:
#no-changelog

### DIFF
--- a/FirebaseAnalyticsOnDeviceConversion.podspec
+++ b/FirebaseAnalyticsOnDeviceConversion.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
     s.name             = 'FirebaseAnalyticsOnDeviceConversion'
-    s.version          = '9.0.0'
+    s.version          = '9.1.0'
     s.summary          = 'On device conversion measurement plugin for FirebaseAnalytics. Not intended for direct use.'
 
     s.description      = <<-DESC
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
     s.cocoapods_version = '>= 1.10.2'
 
-    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '9.0'
+    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '~> 9.0'
 
     s.static_framework = true
 
--- a/FirebaseAnalyticsOnDeviceConversion.podspec
+++ b/FirebaseAnalyticsOnDeviceConversion.podspec
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
     s.cocoapods_version = '>= 1.10.2'
 
-    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '~> 9.0'
+    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '9.0'
 
     s.static_framework = true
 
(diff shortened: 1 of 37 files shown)


FirebaseAnalyticsOnDeviceConversion.podspec
```diff
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
     s.cocoapods_version = '>= 1.10.2'
 
-    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '9.0'
+    s.dependency 'GoogleAppMeasurementOnDeviceConversion', '~> 9.0'
```
ncooke3, line 21: Question, @paulb777. Should I add the `~>` operator? 

I was thinking yes because, for example, FirebaseAuth 9.1.0 has a dependency on FirebaseCore ~> 9.0.

paulb777: Actually FirebaseAnalytics requires matching GoogleAppMeasurement versions, so we should update the dependency version with every release, just as we do the GoogleAppMeasurement versions in the FirebaseAnalytics.podspec.json.

cc: @htcgh @tsunghung

ncooke3: Noted and fixed (55b44be). Unfortunately, this will require a workaround in our tooling. We couldn't keep `[FirebaseAnalyticsOnDeviceConversion.podspec](https://github.com/firebase/firebase-ios-sdk/pull/9787/files/8f54ea297b881ac3030143c391c63d60bcb1e0cf#diff-08afef50b0f0fc9cfd1744148a886f614950cdad0e3afae3fb91c9441ce3c2e5)` as a `json` spec, because of its `source` attribute. Now, as a Ruby spec, it's `s.dependency 'GoogleAppMeasurementOnDeviceConversion', '9.0'` will need to be bumped every release.

Can the Analytics** and GoogleAppMeasurement** podspecs be converted to Ruby files? That would simplify release tooling since both `json` and `podspec` formats would need to be maintained?

paulb777: I'd be ok with making all the podspecs ruby.

FirebaseAnalyticsOnDeviceConversion.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
     s.name             = 'FirebaseAnalyticsOnDeviceConversion'
-    s.version          = '9.0.0'
+    s.version          = '9.1.0'
     s.summary          = 'On device conversion measurement plugin for FirebaseAnalytics. Not intended for direct use.'
 
     s.description      = <<-DESC
```
paulb777, line 6: Line 21 below should be 9.1.0

paulb777 (approved):
LGTM after the one change noted.

The messaging failure looks like the recurring flake we've been seeing.

The zip release build will fail until the podspecs are pushed to SpecsStaging

FirebaseAnalyticsOnDeviceConversion.podspec
ncooke3: Ok, sounds good.

FirebaseAnalyticsOnDeviceConversion.podspec
ncooke3: Good catch, ty! Done.

ncooke3:
Merging because previous CI was green (apart from noted failures 👆) and latest commit should be low risk.

ncooke3:
All specs have been staged on https://github.com/firebase/SpecsStaging

PR #9792, opened 2022-05-11
Overlay Migration

wu-hui:


Firestore/core/src/local/leveldb_persistence.h
```diff
@@ -123,11 +134,15 @@ class LevelDbPersistence : public Persistence {
...
+  std::unordered_map<std::string, std::unique_ptr<LevelDbDocumentOverlayCache>>
+      document_overlay_caches_;
+  std::unique_ptr<LevelDbOverlayMigrationManager> overlay_migration_manager_;
+  std::unordered_map<std::string, std::unique_ptr<LevelDbMutationQueue>>
+      mutation_queues_;
   std::unique_ptr<LevelDbTargetCache> target_cache_;
   std::unique_ptr<LevelDbRemoteDocumentCache> document_cache_;
-  std::unique_ptr<LevelDbIndexManager> index_manager_;
+  std::unordered_map<std::string, std::unique_ptr<LevelDbIndexManager>>
+      index_managers_;
```
ehsannas, line 145: IIUC you're keeping these 4 tables in memory for all users because the overlay migration code is iterating over all users. Should we be concerned about the amount of data that's kept in memory here?

IIUC the caching done here is not necessary, but a performance improvement _for only when overlay migration happens_. OTOH this will increase the memory footprint for all use cases when user switches. Is my understanding correct? Is it worth to take the performance hit during the overlay migration, but not take a memory hit for regular operations?

Firestore/core/src/local/local_store.cc
```diff
@@ -120,6 +121,7 @@ LocalStore::LocalStore(Persistence* persistence,
 LocalStore::~LocalStore() = default;
 
 void LocalStore::Start() {
+  persistence_->GetOverlayMigrationManager()->Run();
```
ehsannas, line 124: Should this be done after `StartMutationQueue()` call below?

Firestore/core/src/local/local_store.h
```diff
@@ -276,6 +276,8 @@ class LocalStore : public bundle::BundleCallback {
 
  private:
   friend class LocalStoreTest;  // for `GetTargetData()`
+  // for reading remote documents and overlays.
```
ehsannas, line 279: nit: no comment needed for adding a test class as friend.

Firestore/core/src/local/persistence.h
```diff
@@ -118,6 +119,8 @@ class Persistence {
   virtual DocumentOverlayCache* GetDocumentOverlayCache(
       const credentials::User& user) = 0;
 
+  virtual OverlayMigrationManager* GetOverlayMigrationManager() = 0;
```
ehsannas, line 122: Add method comment.
`  /** Returns the migration manager responsible for calculating and saving overlays. */` ?

Firestore/core/test/unit/local/leveldb_overlay_migration_manager_test.cc
```diff
@@ -0,0 +1,339 @@
...
+}
+
+void LevelDbOverlayMigrationManagerTest::TearDown() {
+  persistence_->Shutdown();
+}
+
+void LevelDbOverlayMigrationManagerTest::WriteRemoteDocument(
+    const MutableDocument& doc) {
+  persistence_->Run("WriteRemoteDocument", [&] {
+    persistence_->remote_document_cache()->Add(doc, doc.read_time());
```
ehsannas, line 105: should you also do this:
(Java):
```java
    // Set read time to update time.
    document.setReadTime(document.getVersion());
```
before calling Add ?

Firestore/core/test/unit/local/leveldb_overlay_migration_manager_test.cc
```diff
@@ -0,0 +1,339 @@
...
+
+void LevelDbOverlayMigrationManagerTest::WriteMutation(
+    const Mutation& mutation) {
+  WriteMutations({std::move(mutation)});
+}
+
+void LevelDbOverlayMigrationManagerTest::WriteMutations(
+    std::vector<Mutation>&& mutations) {
+  auto result = local_store_->WriteLocally(std::move(mutations));
+  persistence_->Run("Delete Overlays For Testing", [&] {
```
ehsannas, line 117: why deleting overlays here?

ehsannas (commented):
Looks good. some comments below. Also let me know if/when you do manual testing.

Firestore/core/src/local/leveldb_persistence.h
wu-hui: What prompted me to make this change is: persistence handles out a bunch of raw pointers to document_overlay_cache, mutation_queue, etc. When MigrationManager enumerate users and call `GetDocumentOverlayCache(user)`, it invalidates the previous unique_ptr, and leaves local_store holding dangling pointers.

This is not a problem with Java because of GC, the change I made basically is a GC simulation (where the collection happens when SDK exits..). I can add more logic to do this when migration is done however. Did not think it's worth it but now it seems necessary.

ehsannas: SGTM. Thanks

ehsannas (approved):

Firestore/core/src/local/memory_persistence.cc
```diff
@@ -117,7 +117,9 @@ MemoryDocumentOverlayCache* MemoryPersistence::GetDocumentOverlayCache(
   }
 }
 
-OverlayMigrationManager* MemoryPersistence::GetOverlayMigrationManager() {
+OverlayMigrationManager* MemoryPersistence::GetOverlayMigrationManager(
+    const credentials::User& user) {
+  (void)user;
```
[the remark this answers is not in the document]
wu-hui: Done.

wu-hui:
Google internal link for manual testing: go/ios-overlay-migration-test

PR #9796, opened 2022-05-12
Ruby setup for prerlease perf test

paulb777:
Missing ruby setup should fix last macos 12 issue with prerelease workflow

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -553,6 +553,9 @@ jobs:
     runs-on: macos-12
     steps:
     - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@359bebbc29cbe6c87da6bc9ea3bc930432750108
+      with:
+        ruby-version: '2.7'
     - name: Get token
       run: scripts/decrypt_gha_secret.sh scripts/gha-encrypted/bot-access.txt.gpg \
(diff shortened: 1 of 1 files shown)


ncooke3 (approved):

.github/workflows/prerelease.yml
```diff
@@ -553,6 +553,9 @@ jobs:
     runs-on: macos-12
     steps:
     - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@359bebbc29cbe6c87da6bc9ea3bc930432750108
+      with:
+        ruby-version: '2.7'
```
ncooke3, line 558: I've been learning more about GHA to learn about ways we can broadly refactor out workflows. W.r.t. to the recent ruby versioning changes. I read more about the this action: https://github.com/ruby/setup-ruby and learned two things that could be helpful:

1. Set a global ruby version in a `.ruby-version` file in the root of repo to set a global ruby version for use across workflows.
```diff
      with:
-        ruby-version: '2.7'
+       ruby-version: .ruby-version
```
2. `ruby/setup-ruby` has a `bundler-cache` field that can be set to true to cache bundler installs

paulb777: + @DellaBitta who is thinking about ways to improve sharing and best-practices across Firebase SDK GHA infra

PR #9797, opened 2022-05-12
Fix Catalyst Flaky Test

elenadoty:
#no-changelog


ncooke3 (commented):
It looks like the `crashlytics / catalyst` workflow is still failing in the [latest workflow](https://github.com/firebase/firebase-ios-sdk/runs/6473542979?check_suite_focus=true). Was it failing locally?

elenadoty:
They're still failing locally too. Fixed one of the issues but still have to figure out another.

ncooke3 (approved):

elenadoty:
Fixed some of the flakes but there may be more work to do in the existing report manager tests.

PR #9807, opened 2022-05-16
Realtime rc dev

qdpham13:
Creates Realtime Http client file. Includes CompletionHandler & Registration classes. Also includes empty helper methods that will be filled out in future PRs.
As part of bug/230368934


FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
```diff
@@ -30,6 +30,10 @@ extern NSString *const _Nonnull FIRNamespaceGoogleMobilePlatform NS_SWIFT_NAME(
 extern NSString *const _Nonnull FIRRemoteConfigThrottledEndTimeInSecondsKey NS_SWIFT_NAME(
     RemoteConfigThrottledEndTimeInSecondsKey);
 
+extern @interface FIRConfigUpdateListenerRegistration : NSObject
```
chliangGoogle, line 33: Please add a comment explain what it does, you can refer to firestore documentation as well for format similarity.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,76 @@
...
+- (void)beginRealtimeStream {
+}
+
+- (void)pauseRealtimeStream {
+}
+
+- (FIRConfigUpdateListenerRegistration *)addConfigUpdateListener:
+    (FIRConfigUpdateCompletion)completionHandler {
+  NSString *listenerName = [@(listenerCount) stringValue];
+  [_listeners setValue:completionHandler forKey:listenerName];
```
chliangGoogle, line 63: I wonder if you need a serial queue to manage this dictionary to ensure to conflicts, you can refer to other config source code to see how it was implemented: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L152

It's better to have a separate serial queue that is only managing this class.

chliangGoogle (changes requested):

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
qdpham13: Thanks! Added comments for the interface as well as the function. SImilar to Firestores'
https://github.com/firebase/firebase-ios-sdk/blob/78cc446bb68d34b7b54b1b8f6a62f37ad540fe83/Firestore/Source/Public/FirebaseFirestore/FIRListenerRegistration.h#L21

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Wrapped dictinary calls with serial queue

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef RCNConfigRealtime_h
+#define RCNConfigRealtime_h
+
+#import <Foundation/Foundation.h>
+#import "FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h"
+
+@interface RCNConfigRealtime : NSObject <NSURLSessionDataDelegate>
+
+@property(strong, atomic, nonnull) NSMutableDictionary<NSString *, id> *listeners;
+@property(strong, atomic, nonnull) dispatch_queue_t lockQueue;
```
chliangGoogle, line 26: Rename the queue to something meaningful like listenersSerialQueue

chliangGoogle: Also all the private properties should probably move to RCNConfigRealtime.m file instead of RCNConfigRealtime.h.

Like this way: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/Sources/FIRMessaging.m#L84

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,93 @@
...
+
+- (void)remove {
+  [self->_realtimeClient removeConfigUpdateListener:listenerName];
+}
+
+@end
+
+@implementation RCNConfigRealtime
+
+- (instancetype)initWithClass {
```
chliangGoogle, line 46: Can just be init instead of initWithClass (Objective C convention)

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,93 @@
...
+#import "FirebaseRemoteConfig/Sources/RCNConfigConstants.h"
+
+NSInteger listenerCount = 1;
+
+@implementation FIRConfigUpdateListenerRegistration {
+  RCNConfigRealtime *_realtimeClient;
+  NSString *listenerName;
+}
+
+- (instancetype)initWithClass:(RCNConfigRealtime *)realtimeClient
```
chliangGoogle, line 28: initWithClient

(objective c convention)

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,93 @@
...
+
+#import "FirebaseRemoteConfig/Sources/RCNConfigRealtime.h"
+#import <Foundation/Foundation.h>
+#import "FirebaseRemoteConfig/Sources/RCNConfigConstants.h"
+
+NSInteger listenerCount = 1;
+
+@implementation FIRConfigUpdateListenerRegistration {
+  RCNConfigRealtime *_realtimeClient;
+  NSString *listenerName;
```
chliangGoogle, line 25: ivar should be prefixed with "_" so this one is _listenerName

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,93 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "FirebaseRemoteConfig/Sources/RCNConfigRealtime.h"
+#import <Foundation/Foundation.h>
+#import "FirebaseRemoteConfig/Sources/RCNConfigConstants.h"
+
+NSInteger listenerCount = 1;
```
chliangGoogle, line 21: Move this into ivar below as well and use the same serial queue to guard it when write/read it as well.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,93 @@
...
+
+@implementation FIRConfigUpdateListenerRegistration {
```
chliangGoogle, line 23: Should this be RCNConfigRealtime?

chliangGoogle (changes requested):
A few more changes as I read through more carefully.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: No. I just have the registration implementation here while I define it in FIRRemoteConfig. The RCNConfigRealtime implementation is right under this implementation.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Changed

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
qdpham13: Thanks for the review! Updated.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
```diff
@@ -82,6 +91,12 @@ typedef void (^FIRRemoteConfigFetchCompletion)(FIRRemoteConfigFetchStatus status
                                                NSError *_Nullable error)
     NS_SWIFT_UNAVAILABLE("Use Swift's closure syntax instead.");
 
+/// Completion handler invoked by config update methods when they get a response from the server.
+///
+/// @param error  Error message on failure.
+typedef void (^FIRConfigUpdateCompletion)(NSError *_Nullable error)
```
chliangGoogle, line 97: I don't see addOnConfigUpdateListener in the public API as the API review suggested. Is this still working progress?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,102 @@
...
+  [self beginRealtimeStream];
+  return [[FIRConfigUpdateListenerRegistration alloc] initWithClient:self
+                                                        listenerName:listenerName];
+}
+
+- (void)removeConfigUpdateListener:(NSString *)listenerName {
+  dispatch_async(_lockQueue, ^{
+    [self->_listeners removeObjectForKey:listenerName];
+  });
+  if ([self->_listeners count] == 0) {
```
chliangGoogle, line 97: nit use .count property is preferred for obcC

chliangGoogle (approved):
This is a good starting point but would be great to add some unit tests either in this PR or next.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,102 @@
...
+}
+
+- (void)pauseRealtimeStream {
+}
+
+- (FIRConfigUpdateListenerRegistration *)addConfigUpdateListener:
+    (FIRConfigUpdateCompletion)completionHandler {
+  NSString *listenerName = [@(self->_listenerCount) stringValue];
+  dispatch_async(_lockQueue, ^{
+    [self->_listeners setValue:completionHandler forKey:listenerName];
```
chliangGoogle, line 84: Hm I'm a bit concerned about this approach. do we need to key them by number? since we don't need to query it.
Also would it be possible in the future that the listener count decreases and cause conflicts?

Either way the line that set the listnerName should also be guaded inside the dispatch queue.

chliangGoogle (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: We do need at least a reference to the completion handler so that it can be removed when users call remove(). However, I changed from dictionary to a set so now we don't need a listener name, only the listener itself.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Done, I'll add unit tests in the next PR when I have the HTTP stream logic as well.

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
qdpham13: Yeah, I wanted to create the client class first and fill out some of the logic before I exposed the public API in FIRRemoteConfig. Thanks!

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
```diff
@@ -30,6 +30,15 @@ extern NSString *const _Nonnull FIRNamespaceGoogleMobilePlatform NS_SWIFT_NAME(
 extern NSString *const _Nonnull FIRRemoteConfigThrottledEndTimeInSecondsKey NS_SWIFT_NAME(
     RemoteConfigThrottledEndTimeInSecondsKey);
 
+/** Represents a listener that can be removed by calling remove. */
```
karenyz, line 33: can you add a bit more information here so developers can understand what this can be used for

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,102 @@
...
+  return realtimeRemoteConfigQueue;
+}
+
+- (void)beginRealtimeStream {
+}
+
+- (void)pauseRealtimeStream {
+}
+
+- (FIRConfigUpdateListenerRegistration *)addConfigUpdateListener:
```
karenyz, line 78: i was under the impression that `addConfigUpdateListener` would only be called once, but here it looks like we have a set of listeners, so does that mean we will allow developers to register multiple listeners for the same config update signal?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -0,0 +1,102 @@
...
+    (FIRConfigUpdateCompletion)completionHandler {
+  __weak RCNConfigRealtime *weakSelf = self;
+  dispatch_async(_realtimeLockQueue, ^{
+    __strong RCNConfigRealtime *strongSelf = weakSelf;
+    [strongSelf->_listeners addObject:completionHandler];
+    [strongSelf beginRealtimeStream];
```
karenyz, line 84: follow up to the above, if we allow multiple listeners then we would only want to call `beginRealtimeStream` for the first one? (similar to how you handle `pauseRealtimeStream`)

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
qdpham13: Added some more info to the comment.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Yeah, we decided to allow for multiple listeners so that it would be easier to implement features in the wrapper languages like flows for Kotlin or publish in Swift. It's also not a lot of overhead to handle multiple listeners.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Yeah we would only want to call beginRealtimeStream when there is at least 1 listener present, there is no current stream connections, and the app is in foreground. In the next PRs I was going to have a helper function that checks this and place it in the beginRealtimeStream method. I want to place the check there because beginRealtimeStream will also be called by retry functions which will also need to have this check.

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
paulb777: If possible, please avoid adding typedefs to the public API.  typedef's don't port to Swift and we don't want to increase the tech debt for a Swift transition.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2019 Google
```
paulb777, line 2: New files should be `Copyright 2022 Google LLC`

qdpham13: Updated the copyright and is there an alternative that would be preferred over typedef? I don't think I'll be adding any other typedefs to the public API.

paulb777: Use the actual types instead of the typedef - See http://goshdarnblocksyntax.com/. 

Or declare the typedef somewhere else besides a public header

paulb777:
The [commit log](https://github.com/firebase/firebase-ios-sdk/pull/9807/commits) shows commits from GitHub id - donpham13 which is likely causing the CLA problem.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
qdpham13: Moved the typedef to a private header file and fixed CLA problem

paulb777 (commented):
Thanks for the quick followups on my comments!

I'll defer to @chliangGoogle for a more thorough review and approval for the iCore team.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
chliangGoogle: But when users call remove(), they don't specify which listener, I imagine we just remove all of them.
How does Firestore handle this?

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
chliangGoogle: agree, you can simply make it a callback with NSError parameter in it like this one:
https://github.com/firebase/firebase-ios-sdk/blob/eb30b6fdd809cbfe3a9eb7f3ccf1dd46326bef2d/FirebaseAppDistribution/Tests/Unit/FIRAppDistributionTests.m#L39

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Yeah so the registration object will keep a reference to that listener for the user so they don't need to specify it. Calling remove will remove the listener that was passed into the call.

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
qdpham13: I updated this already to just a callback as described above

qdpham13:
@chliangGoogle other than the code freeze, was there any other changes that would prevent me from merging?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef RCNConfigRealtime_h
+#define RCNConfigRealtime_h
+
+#import <Foundation/Foundation.h>
+#import "FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h"
+
+/// Completion handler invoked by config update methods when they get a response from the server.
+///
+/// @param error  Error message on failure.
+typedef void (^FIRConfigUpdateCompletion)(NSError *_Nullable error)
```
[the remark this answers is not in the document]
qdpham13: Removed, will add unit tests in next PR.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
karenyz: I see. There's not much mention of this in the API proposal, maybe you could document this decision and mention how developers should handle the case where they have multiple `FIRConfigUpdateListenerRegistration`s

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
```diff
@@ -30,6 +30,19 @@ extern NSString *const _Nonnull FIRNamespaceGoogleMobilePlatform NS_SWIFT_NAME(
 extern NSString *const _Nonnull FIRRemoteConfigThrottledEndTimeInSecondsKey NS_SWIFT_NAME(
     RemoteConfigThrottledEndTimeInSecondsKey);
 
+/**
+ * Represents a listener that can be removed by calling remove. This is returned when calling
+ * addOnConfigUpdateListener and should be used when you no longer want to listen for new config
```
karenyz, line 35: "should be **removed** when you no longer want to listen..."?

karenyz (approved):

PR #9809, opened 2022-05-17
Split release job to sub tasks.

granluo:


granluo:
The release workflow is passing in [the run](https://github.com/firebase/firebase-ios-sdk/actions/runs/2359871487)

.github/workflows/prerelease.yml
```diff
@@ -2,8 +2,6 @@ name: prerelease
 
 on:
   pull_request:
-    # closed will be triggered when a pull request is merged. This is to keep https://github.com/firebase/SpecsTesting up to date.
-    types: [closed]
```
paulb777, line 6: Why is this deleted?

.github/workflows/prerelease.yml
```diff
@@ -38,13 +45,59 @@ jobs:
       run: |
          ossbotaccess=`cat oss-bot-access.txt`
          BOT_TOKEN="${ossbotaccess}" test_version="${nightly_version}" sdk_version_config="${GITHUB_WORKSPACE}/scripts/create_spec_repo/RC_firebase_sdk.textproto" local_sdk_repo_dir="${local_sdk_repo_dir}" podspec_repo_branch="${podspec_repo_branch}" scripts/release_testing_setup.sh prerelease_testing
+    - name: Clean spec repo
+      run: |
+         ossbotaccess=`cat oss-bot-access.txt`
+         git clone --quiet https://${ossbotaccess}@github.com/Firebase/SpecsTesting.git "${local_repo}"
+         cd "${local_repo}"
+         # Remove all unhided dirs, i.e. all podspec dir from the spec repo.
```
paulb777, line 53: s/unhided/unhidden/

.github/workflows/prerelease.yml
```diff
@@ -38,13 +45,59 @@ jobs:
...
+         oss-bot-access.txt "$bot_token_secret"
+         scripts/decrypt_gha_secret.sh scripts/gha-encrypted/bot-access.txt.gpg \
+         bot-access.txt "$bot_token_secret"
     - name: Update SpecsTesting repo
       run: |
         botaccess=`cat bot-access.txt`
         cd scripts/create_spec_repo/
         swift build
         pod repo add --silent "${local_repo}" https://"$botaccess"@github.com/Firebase/SpecsTesting.git
-        BOT_TOKEN="${botaccess}" .build/debug/spec-repo-builder --sdk-repo "${local_sdk_repo_dir}" --local-spec-repo-name "${local_repo}" --sdk-repo-name SpecsTesting --github-account Firebase --pod-sources 'https://${BOT_TOKEN}@github.com/Firebase/SpecsTesting' "https://github.com/firebase/SpecsDev.git" "https://github.com/firebase/SpecsStaging.git" "https://cdn.cocoapods.org/" --exclude-pods "FirebaseFirestoreTestingSupport" "FirebaseAuthTestingSupport" "FirebaseCombineSwift"
```
paulb777, line 47: Why don't we need ` --exclude-pods` any longer?

.github/workflows/release.yml
```diff
@@ -15,40 +15,92 @@ concurrency:
...
+         sdk_version_config="${GITHUB_WORKSPACE}/scripts/create_spec_repo/RC_firebase_sdk.textproto" \
+         local_sdk_repo_dir="${local_sdk_repo_dir}" \
+         podspec_repo_branch="${podspec_repo_branch}" \
+         scripts/release_testing_setup.sh prerelease_testing
+    - name: Clean spec repo
+      run: |
+         botaccess=`cat bot-access.txt`
+         git clone --quiet https://"$botaccess"@github.com/FirebasePrivate/SpecsTesting.git "${local_repo}"
+         cd "${local_repo}"
+         # Remove all unhided dirs, i.e. all podspec dir from the spec repo.
```
paulb777, line 58: unhidden

scripts/create_spec_repo/Package.swift
```diff
@@ -20,6 +20,9 @@ import PackageDescription
 
 let package = Package(
   name: "PodSpecBuilder",
+  platforms: [
+    .macOS(.v10_13),
```
paulb777, line 24: Why this change?

paulb777 (commented):
From the logs, it looks like we might save 25-30 minutes from every sub task by pushing FirebaseCore and its dependencies before starting the others.

scripts/create_spec_repo/Package.swift
granluo: This is due to the [Process/Shell update](https://github.com/firebase/firebase-ios-sdk/pull/9809/files#diff-56d9e831e359a173355a51a81b46b99cd8282fdab7b615b1f34c66c12637a3eaL93-R101) in the change.
[Launch](https://developer.apple.com/documentation/foundation/process/1414189-launch) and [LaunchPath](https://developer.apple.com/documentation/foundation/process/1414221-launchpath) in  [Process](https://developer.apple.com/documentation/foundation/process))

granluo: Considering the two previous APIs were deprecated and the tool is mainly running on GHA, which has a consistent version above 10.13, I set the version of the executable instead of using `#if available`.

.github/workflows/prerelease.yml
granluo: The previous tool was searching targeted pods based on the file extensions, so we need to exclude some specs. In this update, we are using the [ManifestParser](https://github.com/firebase/firebase-ios-sdk/pull/9809/files#diff-72d9c3ffc2e64ae904528ac81111ab7de90c1e23e11a968000c97cb1d58135cdR53-R60)  to collect only podspecs matching specs in the Manifest file, so we don't need to manually exclude specs, which are not dependencies of any pods in the Manifest. 
One thing to notice is the `HeartbeatLoggingTestUtils` which depends on[ `FirebaseCoreInternal`](https://github.com/firebase/firebase-ios-sdk/blob/master/HeartbeatLoggingTestUtils.podspec#L37), but is also [a test dependency of `FirebaseCoreInternal`](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCoreInternal.podspec#L54), so a circular dependency warning of either `HeartbeatLoggingTestUtils`  or `FirebaseCoreInternal` could be found  in the log and the `HeartbeatLoggingTestUtils` will be pushed as well.

.github/workflows/prerelease.yml
granluo: This is for testing purpose, since each test takes hours. Updating the condition will terminate the current testing run and the update will not trigger the test. Have pushed another commit to revert the triggering event of both prerelease and release workflow.

paulb777 (approved):
Thanks! I'm ok to consider the FirebaseCore refactor for the future.

granluo:
> Thanks! I'm ok to consider the FirebaseCore refactor for the future.



> Thanks! I'm ok to consider the FirebaseCore refactor for the future.

SGTM, will update the workflow in another pr.

PR #9812, opened 2022-05-17
[FirebaseCore] Mark duplicate internal API as unavailable in Swift 

ncooke3:
### Context
- `FirebaseCore/Sources/FIRComponentContainerInternal.h` was edited in cl/448585537 after the initial copybara to fix a build issue.
- This PR brings that change over to change the code at the source-of-truth so future copybaras don't undo this workaround in g3.
- I'm not expecting this to be a problem since Swift clients (of FirebaseCore) will get this API from `FirebaseCoreExtension` here: [`FirebaseCore/Extension/FIRComponentContainer.h`](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Extension/FIRComponentContainer.h#L38).

#no-changelog

### DIFF
--- a/.github/workflows/functions.yml
+++ b/.github/workflows/functions.yml
@@ -1,5 +1,5 @@
 name: functions
-
+# Touch CI
 on:
   pull_request:
     paths:
--- a/.github/workflows/functions.yml
+++ b/.github/workflows/functions.yml
@@ -1,4 +1,5 @@
 name: functions
+
 on:
   pull_request:
     paths:
--- a/.github/workflows/functions.yml
+++ b/.github/workflows/functions.yml
@@ -1,5 +1,4 @@
 name: functions
-# Touch CI
 on:
   pull_request:
     paths:
--- a/.github/workflows/storage.yml
+++ b/.github/workflows/storage.yml
@@ -1,5 +1,4 @@
 name: storage
-# Touch CI
 on:
   pull_request:
     paths:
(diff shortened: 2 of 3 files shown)


.github/workflows/storage.yml
```diff
@@ -1,5 +1,5 @@
 name: storage
-
+# Touch CI
```
ncooke3, line 2: Revert 
```suggestion
```

.github/workflows/functions.yml
```diff
@@ -1,5 +1,5 @@
 name: functions
-
+# Touch CI
```
ncooke3, line 2: Revert
```suggestion
```

ncooke3:
CI is green for FirebaseCore, FirebaseStorage, and FirebaseFunctions.

I will revert and resolve the CI triggers after after review.

ryanwilson (approved):
LGTM with the workflows reverted

.github/workflows/functions.yml
```diff
@@ -1,5 +1,4 @@
 name: functions
-
 on:
```
ncooke3, line 2:
```suggestion

on:
```

.github/workflows/storage.yml
```diff
@@ -1,5 +1,4 @@
 name: storage
-
 on:
```
ncooke3, line 2:
```suggestion

on:
```

PR #9820, opened 2022-05-19
Add async/await support to OAuthProvider

frboulais:
Add async/await support to `OAuthProvider` so you can write this code:

```swift
let credential = try await OAuthProvider(providerID: "twitter.com").getCredentialWith(nil)
try await Auth.auth().signIn(with: credential)
```

### DIFF
--- a/FirebaseCombineSwift/Sources/Auth/OAuthProvider+AsyncAwait.swift
+++ b/FirebaseCombineSwift/Sources/Auth/OAuthProvider+AsyncAwait.swift
@@ -38,4 +38,4 @@ import Foundation
       }
     }
   }
-#endif // canImport(Combine) && swift(>=5.0)
+#endif // compiler(>=5.5.2) && canImport(_Concurrency)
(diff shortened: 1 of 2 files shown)


paulb777:
Thanks for the PR! Is there any reason that the async/await functionality is added to the Combine sources?  I suspect that it should be part of a new FirebaseAuthSwift library, comparable to the Swift extension libraries we have for other Firebase products.

paulb777:
On another look, this is really just a change to the Combine API. Since Combine is community-supported, we should be able to merge without an API review. I've fixed the style issue in #10610 and will push it forward there.

paulb777:
It turns out that this is not necessary. The Swift compiler's automatic async/await API generation drops the `get` from the API and generates the async/await API. See the example at https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Tests/Sample/SwiftApiTests/SwiftAPI.swift#L264.

I'll make a PR for the typo fix separately at #10610

PR #9835, opened 2022-05-23
Realtime rc dev

qdpham13:
Adds:

- New Realtime RC based exceptions
- Http stream connection via NSURLSession and NSURLSessionDataTask
- Http delegates to receive messages and handle any incoming errors
- Http helper methods that creates request and fills in headers (Taken from RCNConfigFetch)
- Template Version Number to RCNConfigContent
- Unit tests
- Auto fetch - Fetch until we receive new template version or we run out of attempts


FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
```diff
@@ -53,6 +53,8 @@ typedef void (^RCNConfigFetcherCompletion)(NSData *data, NSURLResponse *response
 /// Add the ability to update NSURLSession's timeout after a session has already been created.
 - (void)recreateNetworkSession;
 
+- (NSString *)getTemplateVersionNumber;
```
chliangGoogle, line 56: Better define this as a ivar _templateVersionNumber internally and here as a property.

Then internally use getTemplateVersionNumber to update the ivar _templateVersionNumber

FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
```diff
@@ -56,6 +56,9 @@ typedef void (^RCNConfigFetcherCompletion)(NSData *data, NSURLResponse *response
 /// Provide fetchSession for tests to override.
 @property(nonatomic, readwrite, strong, nonnull) NSURLSession *fetchSession;
 
+/// Provide config template version number for Realtime config client.
+@property(nonatomic, strong, nonnull) NSString *templateVersionNumber;
```
chliangGoogle, line 60: use copy instead of strong for NSString

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -577,4 +584,13 @@ - (NSURLSessionDataTask *)URLSessionDataTaskWithContent:(NSData *)content
   return [_fetchSession dataTaskWithRequest:URLRequest completionHandler:fetcherCompletion];
 }
 
+- (NSString *)getTemplateVersionNumber {
+  if (_content.fetchedConfig != nil &&
+      [_content.fetchedConfig objectForKey:templateVersionNumberKey]) {
```
chliangGoogle, line 589: stringForKey:
and below

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
```diff
@@ -1464,6 +1500,48 @@ - (void)testConfigureConfigWithValidInput {
...
 }
 
+#pragma mark - Realtime tests
+
+- (void)testRealtimeAddConfigUpdateListenerWithValidListener {
+  for (int i = 0; i < RCNTestRCNumTotalInstances; i++) {
+    OCMStub([_configRealtime[i] beginRealtimeStream]).andDo(nil);
+
+    id completion = ^void(NSError *_Nullable error) {
+      if (error != nil) {
```
chliangGoogle, line 1510: hmm is this block async? If it is you will need to setup XCTestExpectation to ask the test to wait until the async method is finished so that it won't finish too early to cause test to fail:

https://developer.apple.com/documentation/xctest/asynchronous_tests_and_expectations?language=objc

chliangGoogle (changes requested):
I will be out tmr and I add @ncooke3 for any objC cmds in case karen and I both out and you are blocked by it.

FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
qdpham13: Updated

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
qdpham13: Oh that makes sense, I updated the tests to do this.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.h
```diff
@@ -14,18 +14,22 @@
  * limitations under the License.
  */
 
-#ifndef RCNConfigRealtime_h
 #define RCNConfigRealtime_h
```
ncooke3, line 17: This line can be removed too?

qdpham13: Removed

qdpham13: Updated, PTAL @ncooke3

qdpham13:
Please take a look when you get the chance, @karenyz @ncooke3. Thanks!

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
```diff
@@ -78,6 +78,16 @@ typedef NS_ERROR_ENUM(FIRRemoteConfigErrorDomain, FIRRemoteConfigError){
     FIRRemoteConfigErrorInternalError = 8003,
 } NS_SWIFT_NAME(RemoteConfigError);
 
+/// Remote Config error domain that handles errors for Realtime.
+extern NSString *const _Nonnull FIRRemoteConfigRealtimeErrorDomain NS_SWIFT_NAME(RemoteConfigRealtimeErrorDomain);
+/// Firebase Remote Config service Realtime error.
+typedef NS_ERROR_ENUM(FIRRemoteConfigRealtimeErrorDomain, FIRRemoteConfigRealtimeError){
+    /// Can't establish config update stream.
+    FIRRemoteConfigRealtimeErrorStream = 8004,
```
chliangGoogle, line 86: I think it makes more sense to merge the error with the ones above. There's no need to make seperate domain errors, you can just merge them under RemoteConfigError

chliangGoogle: @ryanwilson

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -577,4 +584,15 @@ - (NSURLSessionDataTask *)URLSessionDataTaskWithContent:(NSData *)content
   return [_fetchSession dataTaskWithRequest:URLRequest completionHandler:fetcherCompletion];
 }
 
+- (NSString *)getTemplateVersionNumber {
+  if (_content.fetchedConfig != nil &&
+      [_content.fetchedConfig objectForKey:templateVersionNumberKey] &&
+      ![[_content.fetchedConfig objectForKey:templateVersionNumberKey]
+          isKindOfClass:[NSNull class]]) {
```
chliangGoogle, line 591: I will like [_content.fetchedConfig objectForKey:templateVersionNumberKey] will be able to detect if it's null so 
![[_content.fetchedConfig objectForKey:templateVersionNumberKey]
          isKindOfClass:[NSNull class]]
is not needed.

you might want to check sth like 

[[_content.fetchedConfig objectForKey:templateVersionNumberKey]
          isKindOfClass:[NSString class]] or NSNumber depends on how you store it.

chliangGoogle (changes requested):

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
qdpham13: I think in our meetings before the API proposal we talked about creating a new error domain for realtime. But I would have no problems moving it into the original RemoteConfigError domain

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -75,6 +78,7 @@ @implementation RCNConfigFetch {
   NSURLSession *_fetchSession;  /// Managed internally by the fetch instance.
   NSString *_FIRNamespace;
   FIROptions *_options;
+  NSString *_templateVersionNumber;
```
karenyz, line 81: any particular reason `templateVersionNumber` is a string and not a number?

qdpham13: It's returned as a string in the fetch response and so we keep it like that and then convert it later in the Realtime client when we need it for version checking.

FirebaseRemoteConfig/Sources/Public/FirebaseRemoteConfig/FIRRemoteConfig.h
ryanwilson: Unfortunately we can't use the same domain when using `NS_ERROR_ENUM` since it uses that to assemble the error enum on the Swift side

ryanwilson: Sorry I didn't quite realize the errors started at 8004 - there's no reason for this to be the case if we're using a different error enum.

Re: why we can't/ shouldn't add them to `FIRRemoteConfigErrorDomain`, this is an effort to make our errors more locally scoped. It's not possible to get a `FIRRemoteConfigRealtimeErrorStream` error when calling other RC APIs, so that error shouldn't exist in the regular `FIRRemoteConfigError`. 

This is also a limitation imposed by `NS_ERROR_ENUM`, since the error types have a 1 to 1 mapping of domain to Swift error, so that's why we have a second domain. Happy to chat further about it if needed!

qdpham13:
Hey everyone, please take a look at the PR and give a review when you get the chance. If everyone feels like it would be easier to go over it in a meeting we can do that too. Thanks! @chliangGoogle @karenyz

PR #9846, opened 2022-05-26
fix doc comments in auth header

### LINKED ISSUE #9808: FIRAuthErrors.h still using ObjC naming
This should be migrated to Swift, here and below in the same file.

https://github.com/firebase/firebase-ios-sdk/blob/8d65537335cf4b6746a37298f3334128a4c6a421/FirebaseAuth/Sources/Public/FirebaseAuth/FIRAuthErrors.h#L21-L38

morganchen12:
Fixes #9808.


rosalyntan (approved):

morganchen12:
Merging even though some tests are failing because this is a comments-only change

Issue #9849, opened 2022-05-26
Cannot unset/remove customMetadata keys on StorageReference objects - regression from v8


mikehardy:
### Step 0: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a GitHub issue.
    * If this is a __feature request__ please use the Feature Request template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the `firebase` tag.
  * For general (non-iOS) Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For backend issues, console issues, and other non-SDK help that does not fall under one
    of the above categories, reach out to
    [Firebase Support](https://firebase.google.com/support/).
  * Once you've read this section and determined that your issue is appropriate for
    this repository, please delete this section.

### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 13.4
  * Firebase SDK version: 9.1.0
  * Installation method: `CocoaPods`
  * Firebase Component: Storage
  * Target platform(s): `iOS`

### [REQUIRED] Step 2: Describe the problem

I'm forward-porting react-native-firebase to the v9+ SDK release here, and I've run into a problem with Storage.

Specifically, we attempt to mirror the firebase-js-sdk API surface area to react-native library consumers (for interoperability with web code) but we call down to underlying firebase-ios-sdk and firebase-android-sdk

firebase-js-sdk documents that to remove a customMetadata key/value pair on a StorageReference, you send in a null value to the updateMetadata call.

This worked fine with firebase-ios-sdk v8 and lower, but now it appears there are some Swift-y types / casts happening, and nulls will crash an app

I will attempt to highlight the important frames 7-16 from the boilerplate with a lot of newlines:

```

    updateMetadata
16:13:27.013 detox[39601] ERROR: [WS_ERROR] The app has crashed, see the details below:

Signal 6 was raised
(
        0   Detox                               0x0000000102125c15 +[NSThread(DetoxUtils) dtx_demangledCallStackSymbols] + 37
        1   Detox                               0x0000000102128670 __DTXHandleCrash + 464
        2   Detox                               0x0000000102128db1 __DTXHandleSignal + 59
        3   libsystem_platform.dylib            0x00007fff701bddfd _sigtramp + 29
        4   ???                                 0x00007fc7a2723a60 0x0 + 140495400614496
[...]

PR #9870, opened 2022-05-31
Use nil to clear customMetadata

paulb777:
Update test to be consistent with docs.  Verified locally.

See #9849


morganchen12 (approved):
LGTM on greens

PR #9880, opened 2022-06-04
Convert Storage Task implementation to Swift

paulb777:
Migrates the Storage task related implementations and unit tests to Swift.

Since is most of the rest of the library, will merge to a branch until Storage is completely migrated to Swift.


FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Prepares a task and begins execution.
+   */
+  internal func enqueue() {
+    weak var weakSelf = self
+    DispatchQueue.global(qos: .background).async {
+      guard let strongSelf = weakSelf else {
+        return
+      }
```
andrewheard, line 52: Would it be ever so slightly cleaner to use the closure capture list syntax (not sure what the actual name is)? I.e.,
```
DispatchQueue.global(qos: .background).async { [weak self] in
  guard let strongSelf = self else { return }
```
I _think_ that would ensure that any weak references to self below would end up being optional, which could be a nice safety check.

paulb777: Good thought and I learned something from investigating. However, the self references in the completion block need to be retained to make sure that the completion block always gets called.

FirebaseStorage/Sources/StorageTask.swift
```diff
@@ -29,15 +34,66 @@ import FirebaseStorageInternal
...
+  internal var metadata: FIRIMPLStorageMetadata?
+
+//  /**
+//   * Error which occurred during task execution, or nil if no error occurred.
+//   */
+  internal var error: NSError?
+//
+//  /**
+//   * NSProgress object which tracks the progress of an observable task.
+//   */
```
andrewheard, line 70: nit: Please remove extra `//`

FirebaseStorage/Sources/StorageTask.swift
```diff
@@ -29,15 +34,66 @@ import FirebaseStorageInternal
    * An immutable view of the task and associated metadata, progress, error, etc.
    */
   @objc public var snapshot: StorageTaskSnapshot {
-    return StorageTaskSnapshot(task: self)
+    objc_sync_enter(StorageTask.self)
```
andrewheard, line 37: It'd be interesting to see if there's a Swift-y alternative to `objc_sync_` or `@synchronized` but this is fine too.

andrewheard (approved):
LGTM -- only optional style nits and questions. Sorry for the delay!

FirebaseStorage/Sources/StorageDownloadTask.swift
```diff
@@ -26,37 +31,173 @@ import FirebaseStorageInternal
...
+          fetcherCompletion(data, error as? NSError)
+        }
+      }
+    }
+  }
+
+  internal func cancel(withError error: NSError) {
+    weak var weakSelf = self
+    DispatchQueue.global(qos: .background).async {
+      weakSelf?.state = .cancelled
```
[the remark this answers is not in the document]
paulb777: Good question. This is a straight port. I suppose that since everything is a message send to weakSelf, if it goes nil, then the rest of the block will be a no-op.

FirebaseStorage/Sources/StorageTask.swift
paulb777: Done.

FirebaseStorage/Sources/StorageTask.swift
paulb777: I suspect there is

paulb777 (commented):
Thanks for the thorough review and great suggestions!

PR #9885, opened 2022-06-07
Firestore: Add NoDestructor to suppress destructors on exit

dconeybe:
#no-changelog


ehsannas (approved):
Nice! I think this is an improvement in the way global/static scope handling was done in the past.

PR #9887, opened 2022-06-07
Use `@_implementationOnly import` for FirebaseCoreExtension

ryanwilson:
Using a regular import caused internal Core APIs to be surfaced to end
users.


paulb777 (approved):
Good find! Did you confirm the interop pods and FirebaseSharedSwift APIs are not exposed also?

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -25,6 +24,9 @@ import FirebaseSharedSwift
   import GTMSessionFetcherCore
 #endif
 
+// Avoids exposing internal FirebaseCore APIs to Swift users.
+@_implementationOnly import FirebaseCoreExtension
```
ncooke3, line 28: Awesome!

ncooke3 (approved):

FirebaseFunctions/Sources/Functions.swift
ryanwilson: The leading underscore makes me nervous that it'll break at some point unexpectedly... but it's better than the alternative at least 🤷 .

ryanwilson:
> Good find! Did you confirm the interop pods and FirebaseSharedSwift APIs are not exposed also?

Confirmed that interop pods are not exposed, neither is StorageInternal. 

Thanks for bringing up SharedSwift: the SharedSwift APIs are actually required to be exposed but currently aren't, so auto complete isn't working as expected. I'll see if we can address that with `@_exported` in a follow up PR.

PR #9914, opened 2022-06-16
Realtime rc dev

qdpham13:
- Expose public realtime method and wire up the client in FIRRemoteConfig
- Add retries with back off and limited attempts on every failed http stream connection
- Add http stream retry for every background to foreground app movement
- Update tests


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -406,9 +409,9 @@ - (void)fetchLatestConfig:(NSInteger)remainingAttempts targetVersion:(NSInteger)
 }
 
 - (void)scheduleFetch:(NSInteger)remainingAttempts targetVersion:(NSInteger)targetVersion {
-  /// Needs fetch to occur between 1 - 4 seconds. Randomize to not cause ddos alerts in backend
+  /// Needs fetch to occur between 0 - 3 seconds. Randomize to not cause ddos alerts in backend
```
chliangGoogle, line 412: nit: DDoS

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -117,6 +117,9 @@ - (instancetype)init:(RCNConfigFetch *)configFetch
     _options = options;
     _namespace = namespace;
 
+    /// Set retry seconds to a random number between 1 and 6 seconds.
```
chliangGoogle, line 120: Any reason this is 1-6 seconds while below there's a 1-4+1 and 1-3 for another use case?

I wonder if we should just unify here and also create a constant for the threshold number

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -333,7 +336,7 @@ - (void)retryHTTPConnection {
     __strong RCNConfigRealtime *strongSelf = weakSelf;
     if ([strongSelf canMakeConnection] && MAX_RETRY_COUNT > 0 && !isRetrying) {
       if (MAX_RETRY_COUNT < MAX_RETRY) {
-        double RETRY_MULTIPLIER = arc4random_uniform(60) + 10;
+        double RETRY_MULTIPLIER = arc4random_uniform(3) + 2;
```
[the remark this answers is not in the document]
qdpham13: Because this is a number I want to multiply the retry seconds by and I want it to be between 2 and 5 times as the previous attempt.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: They are all different b/c they're for different mechanisms. One is for when the fetch should occur so that we prevent a DDoS attack and we do it from 0 - 3 seconds. Another is for the base http retry time which we want from 1 - 6 seconds so that all app instances don't retry at the same time. And the last one is 2 - 5 for how much we want to multiply base retry time by, this is for exponential back-off if the last attempt failed.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -53,6 +53,12 @@
 NSTimeInterval timeoutSeconds = 4320;
 NSInteger FETCH_ATTEMPTS = 3;
```
paulb777, line 54: Please check the Objective C style guide. I think all caps is only for #define constants.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -53,6 +53,12 @@
...
 
+// Retry parameters
+NSInteger MAX_RETRY = 7;
+NSInteger MAX_RETRY_COUNT = 7;
+NSInteger RETRY_SECONDS = 3;
```
paulb777, line 59: Why is this initialized here but then set below?

paulb777 (commented):
Probably outside the scope of this PR: is integration testing FirebasePerformance with Realtime RC part of the test plan?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated variables to follow obj-c style guide.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: gMaxRetryCount keeps track of retries remaining and gMaxRetries is the initial count that gMaxRetryCount will be restarted to. Changed gRetrySeconds to being initialized in the constructor.

paulb777:
@qdpham13 What about apps that are using FirePerf and want to use real-time RC.  Will the same RC run real-time for the app RC values, but not real-time for Fireperf RC values?

paulb777 (approved):
Thanks for the clarifications!

PR #9921, opened 2022-06-17
[Infra] Add `.ruby-version` file for GHA

ncooke3:
### Context
- Pin the monorepo's Ruby version to 2.7 (for use by GHA)
- From https://github.com/ruby/setup-ruby#usage:
```YML
- uses: ruby/setup-ruby@v1
      with:
        ruby-version: '3.0' # Not needed with a .ruby-version file
```

#no-changelog


ncooke3:
@paulb777, after some experiments in #9919, I think the .ruby-version file needs to be checked into `master` to be picked up by GHA. It should be possible based on https://github.com/actions/setup-ruby/issues/31#issuecomment-609531499

paulb777 (approved):

ncooke3:
Cool, this did the trick. Verified here: https://github.com/firebase/firebase-ios-sdk/runs/6943204414?check_suite_focus=true#step:4:14

ryanwilson:
As a heads up: this broke my local CocoaPods flow since I had version 2.7.1 installed and not 2.7.0:

```
$ pod repo update
rbenv: version `ruby-2.7' is not installed (set by <REDACTED_PATH_TO_REPO>/.ruby-version)
```

I'm not sure the best solution, but wanted to call  it out in case others run into this.

ncooke3:
Sorry @ryanwilson, I had a feeling something like that would happen! Were you able to easily pin your version to 2.7.0?

ryanwilson:
> Sorry @ryanwilson, I had a feeling something like that would happen! Were you able to easily pin your version to 2.7.0?

It was easier for me to just change that value to my local version (2.7.1) 😝

PR #9829, opened 2022-05-22
[Infra] Convert `*.podspec.json`s to Ruby podspecs

ncooke3:
### Context
- Used only Ruby podspecs (as opposed to JSON ones) to help streamline release tooling's spec updating logic. 
  - There were some implicit assumptions that `podspec.json` specs were closed source so I had to update them so the source of truth is the the manually set `isClosedSource` property on the `Pod` struct.
- Updated the `pod_lib_lint.rb` script to 1. reject linting a binary pod and 2. avoid linting local binary deps. These changes are because `pod lib lint` does not work for local binary pods.

<details>
<summary>Issues (see below thread for resolution)</summary>
<br>

The spec is not validating with `pod lib lint` but it is with `pod spec lint`. When **lib** linting, I'm getting the following error:
```
 -> FirebaseAnalytics (9.1.0)
    - ERROR | [FirebaseAnalytics/AdIdSupport,FirebaseAnalytics/WithoutAdIdSupport] file patterns: The `vendored_frameworks` pattern did not match any file.

[!] FirebaseAnalytics did not pass validation, due to 1 error.
```
This error doesn't make sense to me because if I download and open the zipped source, the xcframework is at `Frameworks/FirebaseAnalytics.xcframework` path... 🤔 

I'm running the following command:
```console
bundle exec pod lib lint \
    --sources=https://github.com/firebase/SpecsDev.git,https://github.com/firebase/SpecsStaging.git,https://cdn.cocoapods.org/ \
    --include-podspecs=\ {FirebaseCore.podspec,FirebaseCoreDiagnostics.podspec,FirebaseCoreInternal.podspec,FirebaseInstallations.podspec\} \
    --analyze FirebaseAnalytics.podspec \
    --platforms=macos \
    --skip-import-validation \
    --no-clean
```

BUT, if **spec** lint, it validates:
```console
bundle exec pod spec lint 
    --sources=https://github.com/firebase/SpecsDev.git,https://github.com/firebase/SpecsStaging.git,https://cdn.cocoapods.org/ \
    --analyze FirebaseAnalytics.podspec \ 
    --platforms=macos \
    --skip-import-validation \
    --no-clean
```

</details>

#no-changelog

PR #9922, opened 2022-06-17
[Infra] Remove explicit Ruby versioning from workflows

ncooke3:
### Context
Following #9921, we no longer have to specify the ruby version for each job as it will be inferred from the `.ruby-version` file at root.

Done by piping all workflows into a sed where the two lines _after_ the pattern are removed.
```console
find .github/workflows -name '*.yml' | xargs sed -i '' -e '/setup-ruby/{n;N;d;}'
```

You can see it in action in the GHA below: [example](https://github.com/firebase/firebase-ios-sdk/runs/6943399440?check_suite_focus=true#step:3:9)

cc: @granluo, @dellabitta, @sunmou99 for GHA related improvement

#no-changelog

### DIFF
--- a/.github/workflows/abtesting.yml
+++ b/.github/workflows/abtesting.yml
@@ -27,8 +27,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: ruby/setup-ruby@359bebbc29cbe6c87da6bc9ea3bc930432750108
-      with:
-        ruby-version: '2.7'
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: Build and test
@@ -64,8 +62,6 @@ jobs:
       with:
         cache_key: ${{ matrix.os }}
     - uses: ruby/setup-ruby@359bebbc29cbe6c87da6bc9ea3bc930432750108
-      with:
-        ruby-version: '2.7'
     - name: Setup Bundler
       run: scripts/setup_bundler.sh
     - name: Setup project and Build for Catalyst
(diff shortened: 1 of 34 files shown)


.github/workflows/abtesting.yml
```diff
@@ -27,8 +27,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: ruby/setup-ruby@359bebbc29cbe6c87da6bc9ea3bc930432750108
```
paulb777, line 29: Do we still need the `setup-ruby` with the commit hash?

paulb777 (approved):
Nice cleanup!

LGTM on good CI run

.github/workflows/abtesting.yml
ncooke3: Hmm good question, testing in #9923

ncooke3: So using the ruby/setup-ruby action is still needed: https://github.com/firebase/firebase-ios-sdk/runs/6943893771?check_suite_focus=true#step:4:18

But we can use `@v1` instead of the hash

ncooke3:
[`inappmessaging / quickstart (pull_request)`](https://github.com/firebase/firebase-ios-sdk/runs/6943400215?check_suite_focus=true) had a failure... I'm investigating whether it is a regression caused by #9829.

ncooke3:
CI failures will be addressed by #9924

.github/workflows/abtesting.yml
ncooke3: Resolved in a3a1bd0

paulb777 (approved):

PR #9923, opened 2022-06-17
[DO NOT SUBMIT] Test #9921

ncooke3:
#no-changelog


ncooke3:
tl;dr is that we still need to use the `ruby/setup-ruby` action because https://github.com/firebase/firebase-ios-sdk/runs/6943893771?check_suite_focus=true

PR #9926, opened 2022-06-19
Add a StorageMetadata integration test

paulb777:
For investigations related to #9849 discussions, add a test for initializing StorageMetadata with `initWithDictionary` and then clearing it. Also test the name field.

The test passes in both Firebase 8 and 9.


andrewheard:
> 1 Warning
> ⚠️	Did you forget to add a changelog entry? (Add #no-changelog to the PR description to silence this warning.)
> Generated by 🚫 [Danger](https://danger.systems/)

Just double checking, is the changelog entry only for things we'd want to show up in release notes (which wouldn't be necessary here)? Otherwise, this LGTM.

PR #9944, opened 2022-06-24
external_rules.cmake: fail fast when preparing the external projects

dconeybe:
This PR modifies the cmake logic that downloads and configures external projects (e.g. snappy, leveldb) so that it fails fast when the downloading and/or configuring fails. Previously, if, for example, the snappy "patch" step were to fail then cmake would happily continue on until much later things went badly, often causing difficult to diagnose errors.

This PR makes the following improvements:
1. Adds logging about the commands that are being run so that it's easier to debug when things go poorly.
2. Checks the result of the `execute_process()` cmake function calls and fails immediately if they fail.

#no-changelog

### DIFF
--- a/cmake/external_rules.cmake
+++ b/cmake/external_rules.cmake
@@ -37,29 +37,63 @@ function(download_external_sources)
     )
   endif()
 
-  execute_process(
-    COMMAND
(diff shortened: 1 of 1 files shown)


dconeybe:
obolsete

PR #9947, opened 2022-06-24
Patch nanopb CMakeLists.txt to look for python 3.7 instead of python 2.7

dconeybe:
This fixes the following cmake build error on machines that lack a python 2.7 interpreter (e.g. a fresh macbook):

```
CMake Error at Cellar/cmake/3.23.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE) (Required is at
  least version "2.7")
Call Stack (most recent call first):
  Cellar/cmake/3.23.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  Cellar/cmake/3.23.2/share/cmake/Modules/FindPythonInterp.cmake:169 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  build/external/src/nanopb/CMakeLists.txt:44 (find_package)
```

when the following command is run:

```
cmake -S . -B build
```

#no-changelog

### DIFF
--- a/cmake/external/nanopb.cmake
+++ b/cmake/external/nanopb.cmake
@@ -34,5 +34,6 @@ ExternalProject_Add(
   BUILD_COMMAND ""
   INSTALL_COMMAND ""
   TEST_COMMAND ""
+  PATCH_COMMAND patch -Np1 --binary -i ${CMAKE_CURRENT_LIST_DIR}/nanopb.patch
   HTTP_HEADER "${EXTERNAL_PROJECT_HTTP_HEADER}"
 )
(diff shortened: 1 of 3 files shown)


cherylEnkidu (approved):
Thank you for your work!

ehsannas (commented):
Drive-by review: newer nanopb versions don't rely on python 2.7. Is there a reason we're not upgrading to a newer nanopb version?

paulb777:
Drive-by answer: Firestore's nanopb version needs to match the rest of Firebase and since Firebase supports 1P clients needs to be compatible with the rest of Google. Updating Firestore's nanopb version requires updating all Google nanopb clients.

ehsannas:
I see. Thanks @paulb777

PR #9962, opened 2022-06-28
Discontinue bitcode inclusion

paulb777:
Apple announced the discontinuation of bitcode support.

This drops the zip distribution size from 367MB to 290MB


paulb777:
Hmm, based on this statement in the Xcode release notes

_Starting with Xcode 14, bitcode is no longer required for watchOS and tvOS applications, and the App Store no longer accepts bitcode submissions from Xcode 14._

it looks like bitcode might still be required for tvOS with Xcode 13 builds?

We can hold off on merging until we get more clarity.

cc: @htcgh

paulb777:
Based on some other Googler conversations with Apple, it looks like bitcode will be required for tvOS with Xcode 13 forever.  

So the plan is to hold off on this PR until we drop support for Xcode 13 likely Spring 2023 when Apple stops accepting Xcode 13 App Store submissions.

PR #7604, opened 2021-02-24
Only update experiments for firebase namespace

karenyz:
Currently `updateExperimentsWithResponse` is being called on fetches for all namespaces, causing issues with experiment exposures as `fireperf` fetch requests are incorrectly updating experiment data

Tracked internally in b/179461119

PR #9972, opened 2022-06-30
Modify activate path to only update experiments for firebase namespace

karenyz:
Similar fix to https://github.com/firebase/firebase-ios-sdk/pull/7604, which only updated the fetch path but not activate

Tracked internally in b/211879339


paulb777 (approved):
LGTM!

PR #10012, opened 2022-07-14
Update versions for Release 9.4.0

ryanwilson:


Package.swift
```diff
@@ -19,7 +19,7 @@
 import PackageDescription
 import class Foundation.ProcessInfo
 
-let firebaseVersion = "9.3.0"
+let firebaseVersion = "9.4.0"
```
andrewheard, line 22: Do we have a process in place to ensure Analytics is changed to 9.4.0 before release since it can't be changed here until the [GoogleAppMeasurement](https://github.com/google/GoogleAppMeasurement.git) repo is updated?

andrewheard (approved):

Package.swift
paulb777: This looks like a regression from #9829. The GoogleAppMeasurement versions should not be updated every release. cc; @ncooke3

paulb777: Disregard my previous statement. The GoogleAppMeasurement podspec versions should be updated, but not the SPM versions or pod archives. The reason for the CI failures is the podspecs need to be pushed to SpecsStaging

ncooke3: Pods are currently being published to SpecsStaging.

ncooke3:
I'm investigating why the pods did not push so I'm going to delete this branch and restart the process to see if there was a https://github.com/firebase/firebase-ios-sdk/pull/9829 regression.

ncooke3:
I cleaned up all 9.4 artifacts and was able to get the script working when trying myself.

PR #10014, opened 2022-07-14
Get rid of objc blocks inside dyld callback

wowlocal:
Objective-C blocks may invoke certain methods that attempt to acquire the Objective-C runtime lock. Simultaneously, _dyld_register_func_for_add_image is called once for each image currently part of the program, with its critical section protected by the dyld internal lock. Moreover, the Objective-C runtime frequently accesses dyld, as seen in objc_copyImageNames calling dyld_image_path_containing_address with the Objective-C runtime lock already held https://github.com/apple-oss-distributions/objc4/blob/rel/objc4-838/runtime/objc-runtime-new.mm#L5540.
The lock/unlock policies between dyld and Objective-C runtime appear uncertain and potentially dangerous, making it preferable to avoid using the Objective-C runtime inside dyld callbacks.

These changes aim to enhance the stability and safety of the code by preventing potential deadlocks caused by the unclear lock policies between dyld and Objective-C runtime.

### DIFF
--- a/Crashlytics/Crashlytics/Components/FIRCLSBinaryImage.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSBinaryImage.h
@@ -64,8 +64,7 @@ typedef struct {
   FIRCLSBinaryImageRuntimeNode nodes[CLS_BINARY_IMAGE_RUNTIME_NODE_COUNT];
 } FIRCLSBinaryImageReadWriteContext;
 
-void FIRCLSBinaryImageInit(FIRCLSBinaryImageReadOnlyContext* roContext,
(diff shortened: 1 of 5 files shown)


wowlocal:
It's been a while since I submitted the pull request, and I was wondering if someone could take a moment to review it. I believe that the changes I've proposed could benefit the project, and I'm eager to hear your feedback or address any concerns that you may have. Thank you for your time and consideration. I look forward to your feedback.

samedson (approved):
Hey @wowlocal, apologies for the delay here. We tested out the PR and it looks good. I'm just going to follow up with a few comments.

Issue #9985, opened 2022-07-06
FIRESTORE INTERNAL ASSERTION FAILED: value.has_value() Fatal crash on App Start


MauScheff:
### Describe your environment

  * Xcode version: 13.4.1
  * Firebase SDK version: 9.2.0
  * Installation method: CocoaPods (via Flutter)
  * Firebase Component: Firestore
  * Target platform(s): iOS

###  Describe the problem
When the App gets into a certain state, it crashes on Start every time, even after App reinstall.

#### Steps to reproduce:

This is the error that Crashlytics reports.
```
Fatal Exception: NSInternalInconsistencyException
FIRESTORE INTERNAL ASSERTION FAILED: value.has_value()

Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x99d78 __exceptionPreprocess
1  libobjc.A.dylib                0x16734 objc_exception_throw
2  Foundation                     0x131358 -[NSMutableDictionary(NSMutableDictionary) initWithContentsOfFile:]
3  FirebaseFirestore              0x46828 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
4  FirebaseFirestore              0x462d4 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
5  FirebaseFirestore              0x956a0 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
6  FirebaseFirestore              0x95730 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
7  FirebaseFirestore              0xface0 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
8  FirebaseFirestore              0xd36ac (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
9  FirebaseFirestore              0xd3e2c (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
10 FirebaseFirestore              0xde96c (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
11 FirebaseFirestore              0xc3e78 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
12 FirebaseFirestore              0xd9e04 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
13 FirebaseFirestore              0x137ea8 (Missing UUID b1c971767c493d6baaca5c69dfe3fc5b)
[...]

PR #10016, opened 2022-07-15
Fix Overlay bug with nested field deletion

wu-hui:
Fix for https://github.com/firebase/firebase-ios-sdk/issues/9985

Beside this particular fix, this PR also has a fix for MutationTest.VerifyOverlayRoundTrips itself. It has been passing the wrong field mask, and all tests are treating overlay as set mutations as a result. This test fix also uncovered another bug, which is also fixed in this PR.

### DIFF
--- a/Firestore/core/test/unit/model/mutation_test.cc
+++ b/Firestore/core/test/unit/model/mutation_test.cc
@@ -73,7 +73,7 @@ void VerifyOverlayRoundTrips(const MutableDocument& doc,
   MutableDocument doc_for_mutations = doc.Clone();
   MutableDocument doc_for_overlay = doc.Clone();
 
-  absl::optional<FieldMask> mask;
+  absl::optional<FieldMask> mask = FieldMask();
   for (const Mutation& mutation : mutations) {
     mask = mutation.ApplyToLocalView(doc_for_mutations, mask, now);
   }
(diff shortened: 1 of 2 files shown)


Firestore/core/test/unit/model/mutation_test.cc
```diff
@@ -712,6 +712,25 @@ TEST(MutationTest, OverlayWithFieldDeletionOfNestedField) {
...
+      PatchMutation("collection/key", Map("foo", "foo-patched-value"),
+                    {{"bar.baz", ServerTimestamp()}});
+  Mutation patch3 =
+      PatchMutation("collection/key", Map("foo", "foo-patched-value"),
+                    {Field("bar.baz")}, {});
+
+  Mutation patch4 = PatchMutation(
+      "collection/key", Map("foo", "foo-patched-value"), {Field("bar")}, {});
+
+  VerifyOverlayRoundTrips(doc, {patch1, patch2, patch3, patch4});
```
ehsannas, line 731: can we also try multiple levels as well? e.g. `foo.bar.baz` ?

wu-hui: Done.

Theunodb (approved):

Theunodb (approved):

ehsannas (approved):

PR #10027, opened 2022-07-19
Create dictionary with compactMap

paulb777:
Follow up on a review request from #9979

### DIFF
--- a/FirebaseStorage/Sources/StorageMetadata.swift
+++ b/FirebaseStorage/Sources/StorageMetadata.swift
@@ -107,50 +107,22 @@ import FirebaseStorageInternal
    * @return A Dictionary that represents the contents of the metadata.
    */
   @objc open func dictionaryRepresentation() -> [String: AnyHashable] {
-    var dictionary: [String: AnyHashable] = [:]
-    if let bucket = bucket {
-      dictionary["bucket"] = bucket
-    }
-    if let cacheControl = cacheControl {
-      dictionary["cacheControl"] = cacheControl
-    }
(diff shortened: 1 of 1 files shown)


FirebaseStorage/Sources/StorageMetadata.swift
```diff
@@ -107,50 +107,22 @@ import FirebaseStorageInternal
...
+      "contentLanguage": contentLanguage,
+      "contentType": contentType,
+      "md5Hash": md5Hash,
+      "size": size != 0 ? size : nil,
+      "generation": generation != 0 ? "\(generation)" : nil,
+      "metageneration": metageneration != 0 ? "\(metageneration)" : nil,
+      "timeCreated": timeCreated.map(stringFromDate),
+      "updated": updated.map(stringFromDate),
+      "name": path,
+    ].compactMapValues { $0 }.merging(["metadata": customMetadata]) { current, _ in current }
```
paulb777, line 125: If `customMetadata` is treated like the other fields, the other fields end up like AnyHashable(nil) instead of nil.

morganchen12: My guess is that behavior could be worked around with a clever type cast of `customMetadata`, but no need to spend too much time massaging this code.

morganchen12 (approved):

FirebaseStorage/Sources/StorageMetadata.swift
paulb777: Agreed, but I couldn't find it and this gets almost all of the cleaner code benefit.

PR #9838, opened 2022-05-24
Firestore TransactionOptions added, to specify maxAttempts

dconeybe:

PR #10033, opened 2022-07-20
Fix build warning

paulb777:
Fix build warning introduced in #9838


paulb777:
Already fixed in #9987

PR #10046, opened 2022-07-27
Fix RC Codable to handle arrays and dictionaries from plist defaults

### LINKED ISSUE #9980: RemoteConfig invalid array decoding

  * Xcode version: 13.4.1
  * Firebase SDK version: 9.2.0
  * Installation method: `Swift Package Manage`
  * Firebase Component: `Remote Config`
  * Target platform(s): `iOS`

#### Steps to reproduce:
Example XML:
RemoteConfigDefaults.plist
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
	<dict>
		<key>state</key>
		<string>["yes", "no", "maybe"]</string>
	</dict>
</plist>
```

Fetching the data is fine however decoding runs into the following error:

> DecodingError
  ▿ typeMismatch : 2 elements
    - .0 : Swift.Array<Any>
    ▿ .1 : Context
      - codingPath : 0 elements
      - debugDescription : "Expected to decode Array<Any> but found FirebaseRemoteConfigValueDecoderHelper instead."
      - underlyingError : nil

#### Relevant Code:

```
remoteConfig = RemoteConfig.remoteConfig()
remoteConfig.configSettings = settings
remoteConfig.setDefaults(fromPlist: "RemoteConfigDefaults")
do {
	config = try remoteConfig.decoded(asType: [String].self)
} catch {
	print(error)
}
```

paulb777:
Fix #9980

The Remote Config Value Codable implementation did not expect arrays or dictionaries other than from json. However, arrays and dictionaries can be specified in defaults set from plists.

This PR adds tests and fixes the implementation.

### DIFF
--- a/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
+++ b/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
@@ -18,6 +18,11 @@ import Foundation
 import FirebaseRemoteConfig
 import FirebaseSharedSwift
 
+public enum RemoteConfigValueDecoderError: Error {
+  case arrayDecodeFailure
+  case dictionaryDecodeFailure
+}
+
 /// Implement the FirebaseRemoteConfigValueDecoding protocol for the shared Firebase decoder to
 /// decode Remote Config Values. It returns the four different kinds of values from
 /// a RemoteConfigValue object.
@@ -44,6 +49,20 @@ struct FirebaseRemoteConfigValueDecoderHelper: FirebaseRemoteConfigValueDecoding
     guard let value = value.jsonValue as? [String: AnyHashable] else {
       // nil is the historical behavior for failing to extract JSON.
       return nil
+    }
+    return value
+  }
+
+  func arrayValue() throws -> [AnyHashable] {
+    guard let value = value.jsonValue as? [AnyHashable] else {
+      throw RemoteConfigValueDecoderError.arrayDecodeFailure
+    }
+    return value
+  }
+
+  func dictionaryValue() throws -> [String: AnyHashable] {
+    guard let value = value.jsonValue as? [String: AnyHashable] else {
+      throw RemoteConfigValueDecoderError.dictionaryDecodeFailure
     }
     return value
   }
--- a/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
+++ b/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
@@ -45,14 +45,6 @@ struct FirebaseRemoteConfigValueDecoderHelper: FirebaseRemoteConfigValueDecoding
     return value.dataValue
   }
 
-  func jsonValue() -> [String: AnyHashable]? {
-    guard let value = value.jsonValue as? [String: AnyHashable] else {
-      // nil is the historical behavior for failing to extract JSON.
-      return nil
-    }
-    return value
-  }
-
   func arrayValue() throws -> [AnyHashable] {
     guard let value = value.jsonValue as? [AnyHashable] else {
       throw RemoteConfigValueDecoderError.arrayDecodeFailure
--- a/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
+++ b/FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
@@ -18,11 +18,6 @@ import Foundation
 import FirebaseRemoteConfig
 import FirebaseSharedSwift
 
-enum RemoteConfigValueDecoderError: Error {
-  case arrayDecodeFailure
-  case dictionaryDecodeFailure
-}
-
 /// Implement the FirebaseRemoteConfigValueDecoding protocol for the shared Firebase decoder to
 /// decode Remote Config Values. It returns the four different kinds of values from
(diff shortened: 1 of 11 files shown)


FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
```diff
@@ -47,4 +52,18 @@ struct FirebaseRemoteConfigValueDecoderHelper: FirebaseRemoteConfigValueDecoding
     }
```
ncooke3, line 52: Should we consider throwing an error here to make behavior more consistent with the API you have added? 

If so, we could create an issue to track this for Firebase 10

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
```diff
@@ -47,4 +52,18 @@ struct FirebaseRemoteConfigValueDecoderHelper: FirebaseRemoteConfigValueDecoding
...
   }
+
+  func arrayValue() throws -> [AnyHashable] {
+    guard let value = value.jsonValue as? [AnyHashable] else {
+      throw RemoteConfigValueDecoderError.arrayDecodeFailure
+    }
+    return value
+  }
+
+  func dictionaryValue() throws -> [String: AnyHashable] {
```
ncooke3, line 63: Are just constrained to `String` keys? Could the plist dictionary have integer keys that are decoded as such?

FirebaseRemoteConfigSwift/Tests/Defaults-testInfo.plist
```diff
@@ -0,0 +1,42 @@
...
+	<key>arrayIntValue</key>
+	<array>
+		<integer>1</integer>
+		<integer>2</integer>
+		<integer>0</integer>
+		<integer>3</integer>
+	</array>
+	<key>dictValue</key>
+	<dict>
+		<key>foo</key>
```
ncooke3, line 34: My above comment refers to if this was a `1` instead of `"foo"`...

ncooke3 (approved):
I have a few questions, but this LGTM

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
```diff
@@ -18,6 +18,11 @@ import Foundation
 import FirebaseRemoteConfig
 import FirebaseSharedSwift
 
+public enum RemoteConfigValueDecoderError: Error {
+  case arrayDecodeFailure
```
chliangGoogle, line 22: Just double check for the error should the first letter be capitalized?

Do we need API review for this?

chliangGoogle (approved):

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
paulb777: Maybe? Returning nil is the historical behavior for failing to extract json. The array and dictionary methods are new and don't have historical precedent.

We might want to wait to we do a more fundamental refactor of RemoteConfigValue.

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
paulb777: I believe the keys must be `String`. See [FirebaseDataEncoder.swift](https://github.com/firebase/firebase-ios-sdk/pull/10046/files#diff-db170d9b3f2582a4ba0a083c21fddd7da24eca78abaa33a0f330a3df8184eb19) line 2479 below.

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
paulb777: Good catch! Removed from the public API surface

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
karenyz: ^ would be interested in discussing this a bit more! feels like `arrayValue` and `dictionaryValue` would be preferred over `jsonValue`

karenyz (approved):

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
paulb777: And now completely removed in favor of using the existing DecodingError

FirebaseRemoteConfigSwift/Sources/FirebaseRemoteConfigValueDecoderHelper.swift
paulb777: Great point! There's no reason for `jsonValue` any longer and I replaced its usage with `dictionaryValue`.

paulb777:
A future task could be updating `DecodingError._typeMismatch` with more customized error messages for RCValue decoding failures for json, dictionaries and arrays.

PR #10050, opened 2022-07-28
update auth changelog to test copybara

chliangGoogle:
I suspect because no changelog cause the script not detect any changes. 

I'm thinking updating this not for the release, but for testing copybara.


paulb777 (commented):
This shouldn't be necessary. There may be something wrong with your copybara config, since it works for me. See cl/463973759.

I can help you track it down tomorrow.

PR #10051, opened 2022-07-29
RC Dynamic property wrapper feature 

chliangGoogle:
This feature is built on top of real time config so please just look at the RemoteConfigSwift folder for the code change by selecting FirebaseRemoteConfigSwift on the top left nav bar.

Also add a SwiftUI Sample test app to visualize the API experience.


paulb777:
Please change the base branch so the PR only shows the property wrapper changes.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -0,0 +1,49 @@
...
+internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
+  @Published var configValue: T
+  private let key: String
+  private let remoteConfig: RemoteConfig
+  private var registration: FIRConfigUpdateListenerRegistration?
+
+  init(key: String, remoteConfig: RemoteConfig) {
+    self.key = key
+    self.remoteConfig = remoteConfig
+    configValue = try! remoteConfig.configValue(forKey: key).decoded()
```
ncooke3, line 30: I don't have any opinions about error handling right now but will think about it.

ncooke3 (commented):
Looking good– left some review!

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
paulb777: Likely it should be switched to `guard let configValue = try? ....` or a `try` wrapped in a `do catch`.

paulb777 (commented):
Looking good. Can we see a demo an upcoming standup?

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    public var wrappedValue: T {
+      configValueObserver.configValue
+    }
+
+    /// Creates an instance by defining key.
+    /// This property depends on default remote config.
+    ///
+    /// - Parameter key: key name
+    public init(forKey key: String) {
```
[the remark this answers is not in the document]
chliangGoogle: Good catch! Fixed it.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift
```diff
@@ -0,0 +1,53 @@
...
+      self.key = key
+      remoteConfig = RemoteConfig.remoteConfig()
```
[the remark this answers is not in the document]
chliangGoogle: This is outdated since we are not supporting multiple FIRApp instance for this feature.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift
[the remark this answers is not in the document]
chliangGoogle: This is a good idea, we should mention this in API review to evaluate with RC folks.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -27,20 +27,23 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
   init(key: String, remoteConfig: RemoteConfig) {
     self.key = key
     self.remoteConfig = remoteConfig
-    configValue = try! remoteConfig.configValue(forKey: key).decoded()
-    registration = remoteConfig.add(onConfigUpdateListener: { error in
+    self.configValue = try! remoteConfig.configValue(forKey: key).decoded(asType: T.self)
```
paulb777, line 30: How about?
```
do {
 self.configValue =  try remoteConfig.configValue(forKey: key).decoded(asType: T.self)
} catch {
 fatalError("RemoteConfig Internal Error: Failed to decode value for key:\(key) with error \(error)")
}
```

ncooke3: I don't think we should raise a fatal error because there is no way for the developer to handle it. Property wrappers don't have a good mechanism for handling errors so we'll probably have to follow what was done with the Firestore property wrapper.
1. Wrap the property in a result type
2. Make the property an optional
3. Extend the API to take a default value to fall back to

Firestore's PW supports 1 & 2. I think supporting all three could be good here.

paulb777: Good point! I was addressing the immediate issue of improving the `!` crash potential, but your suggestion is much better.

paulb777: Changed to optional in #10128.  More thought about error handling is probably needed.

chliangGoogle: For 3, it's an API decision we will talk about in the API review.

For 1 and 2, @ncooke3 i'm not sure what you meant, can you give an example for each? Keep in mind this is a dynamic property wrapper that's binding with a UI component, so we can't change the way it's defined (The configValue proeprty)

As for now @paulb777 should I keep what I have and update my PR with https://github.com/firebase/firebase-ios-sdk/pull/10128?

chliangGoogle: Also by make the property optional, @paulb777  would https://github.com/firebase/firebase-ios-sdk/pull/10128 make  developers have to pass in a default value?  As we discussed, it would be impossible for the current implementation to pass in a default value given the generic type is T.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift
```diff
@@ -0,0 +1,53 @@
...
+  public struct RemoteConfigProperty<T: Decodable>: DynamicProperty {
+    @StateObject private var configValueObserver: RemoteConfigValueObservable<T>
+
+    /// Remote Config key name for this property
+    public let key: String
+
+    /// Remote Config instance for this property
+    public let remoteConfig: RemoteConfig
+
+    public var wrappedValue: T {
```
paulb777, line 33: Should this be T?

paulb777:
On idea to eliminate the force unwraps would be to have type specific property wrappers like RemoteConfigPropertyInt, RemoteConfigPropertyString, RemoteConfigPropertyArray, RemoteConfigPropertyStruct, etc.

Then they could all use `guard try? ... else set default value` except for RemoteConfigPropertyStruct which would need the property wrapper to take a default value parameter.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -19,16 +19,16 @@ import FirebaseRemoteConfig
 
 @available(iOS 14.0, macOS 11.0, macCatalyst 14.0, tvOS 14.0, watchOS 7.0, *)
 internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
-  @Published var configValue: T
+  @Published var configValue: T!
```
ncooke3, line 22: Accesses this implicitly unwrapped property will crash when decoding fails.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
ncooke3: @chliangGoogle, here are some examples of 1 & 2. They both shift the handling of the success (non-nil) and failure (nil) states to the API consumer.

1. `@RemoteConfigProperty(key: "award_message") var awardMessage: Result<String, _>`
The type of the above property is instead a result type so the required type of the property wrapper is `Result<T, Error> where T: Decodable`.
3. `@RemoteConfigProperty(key: "award_message") var awardMessage: String?`
The type of the above property has a `?` so the required type of the property wrapper is `T? where T: Decodable`.

ncooke3:
> On idea to eliminate the force unwraps would be to have type specific property wrappers like RemoteConfigPropertyInt, RemoteConfigPropertyString, RemoteConfigPropertyArray, RemoteConfigPropertyStruct, etc.
> 
> Then they could all use `guard try? ... else set default value` except for RemoteConfigPropertyStruct which would need the property wrapper to take a default value parameter.

This should be able to be done without needed these wrapper types.

We have access to the T type which means we should be able to do something like this:
```swift
 init(key: String, remoteConfig: RemoteConfig) {
    self.key = key
    self.remoteConfig = remoteConfig

    try {
      self.configValue = try remoteConfig[key].decoded()
    } catch {
      print("Decoding of remote config value to given type failed.")
      if T.self is Bool.Type {
        // we are working with a bool and can provide a default value on our own
      } else if T.self is Int.Type {
        // ...
      } else if ... { ... }
      } else {
        // We are working with a custom struct and need to have been given a default to use. 
      }
    }
```

Still though, I still support the idea of enforcing the developer to give default value for _all_ types. The reason being is for two reasons:
- If we don't enforce a default value in the API's signature, the API is guessing really just making an assumption about the "default" behavior of the console-defined remote config schema. For example, would `Bool` default to `false`? What's to say the developer's remote config isn't configured to where they would prefer `true` as the default value. 
- I think of properties as just syntactic sugar. It is my hunch that as the API surface currently is, developers have to have a default value ready if they are using remote config to determine what a UI should look like. They may use the `setDefaults(from:)` API or doing something like:
```swift
struct ContentView: View {
    var body: some View {
        NavigationView {
            VStack {
                // ...
            }
            .onAppear {
                try { 
                    self.welcomeMessage = try remoteConfig["welcome_message"].decoded()
                } catch {
                    self.welcomeMessage = "Welcome!"
                } 
            }
        }
    }
}
```

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+import FirebaseRemoteConfig
+
+@available(iOS 14.0, macOS 11.0, macCatalyst 14.0, tvOS 14.0, watchOS 7.0, *)
+internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
+  @Published var configValue: T!
```
ncooke3, line 22: Hi @chliangGoogle and @paulb777, left some more thoughts below.

> On idea to eliminate the force unwraps would be to have type specific property wrappers like RemoteConfigPropertyInt, RemoteConfigPropertyString, RemoteConfigPropertyArray, RemoteConfigPropertyStruct, etc.
> 
> Then they could all use `guard try? ... else set default value` except for RemoteConfigPropertyStruct which would need the property wrapper to take a default value parameter.

This should be able to be done without needed these wrapper types.

We have access to the T type which means we should be able to do something like this:
```swift
 init(key: String, remoteConfig: RemoteConfig) {
    self.key = key
    self.remoteConfig = remoteConfig

    try {
      self.configValue = try remoteConfig[key].decoded()
    } catch {
      print("Decoding of remote config value to given type failed.")
      if T.self is Bool.Type {
        // we are working with a bool and can provide a default value on our own
      } else if T.self is Int.Type {
        // ...
      } else if ... { ... }
      } else {
        // We are working with a custom struct and need to have been given a default to use. 
      }
    }
```

Still though, I still support the idea of enforcing the developer to give default value for _all_ types. The reason being is for two reasons:
- If we don't enforce a default value in the API's signature, the API is guessing really just making an assumption about the "default" behavior of the console-defined remote config schema. For example, would `Bool` default to `false`? What's to say the developer's remote config isn't configured to where they would prefer `true` as the default value. 
- I think of properties as just syntactic sugar. It is my hunch that as the API surface currently is, developers have to have a default value ready if they are using remote config to determine what a UI should look like. They may use the `setDefaults(from:)` API or doing something like:
```swift
struct ContentView: View {
    var body: some View {
        NavigationView {
            VStack {
                // ...
            }
            .onAppear {
                try { 
                    self.welcomeMessage = try remoteConfig["welcome_message"].decoded()
                } catch {
                    self.welcomeMessage = "Welcome!"
                } 
            }
        }
    }
}
```
  Either way, I feel like a default value is going to be needed to fall back on _when using remote config to configure the UI_. 
  Without one, the only other option I see is either hiding the UI element or presenting an error message to the user.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import SwiftUI
+import FirebaseRemoteConfig
+
+@available(iOS 14.0, macOS 11.0, macCatalyst 14.0, tvOS 14.0, watchOS 7.0, *)
+internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
+  @Published var configValue: T
+  private let key: String
+  private let remoteConfig: RemoteConfig
+  private var registration: FIRConfigUpdateListenerRegistration?
```
ryanwilson, line 25: Random comment: this sticks out as there shouldn't be a `FIR` prefix. Should be fixed in the ObjC source.

chliangGoogle: @karenyz @qdpham13 We should adjust the realtime API on this

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -0,0 +1,73 @@
...
+        default:
+          self.configValue = value
+        }
+      }
+    } catch {
+      self.configValue = fallbackValue
+    }
+
+    // This will turn on real time update for all configs
+    let currentRegister = self.remoteConfig.add(onConfigUpdateListener: { error in
```
karenyz, line 50: (i think you are already aware of this based on your comment), but just calling out that the listener listens for updates to **any** key, not just the key of this particular property.. looks like the perfect use-case for a per-key listener in the future 🧐

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
karenyz: I was actually thinking about whether `placeholder` is needed since RC already supports in-app defaults.. so this was helpful to read! 

However I *think* the way RC works is even with no in-app default OR server-side value we still return an "empty" value (0, false, etc), so not entirely sure when we would fallback to the placeholder.. Still agree we should have it though!

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
ryanwilson: The `FIR` prefix should never be used in Swift APIs, and always used in Objective-C. Yep, just dropping the FIR prefix for Swift works.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
chliangGoogle: The fallback placeholder is 4th mechanism after the static. It only reflects on the property wrapper parameter that is not included in the original SDK.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
chliangGoogle: @karenyz let me know if you want to include this to the RC API review.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
chliangGoogle: that's correct

FirebaseRemoteConfig/Sources/FIRRemoteConfig.m
```diff
@@ -43,6 +43,8 @@
 static NSString *const kRemoteConfigMinimumFetchIntervalKey = @"_rcn_minimum_fetch_interval";
 /// Timeout value for waiting on a fetch response.
 static NSString *const kRemoteConfigFetchTimeoutKey = @"_rcn_fetch_timeout";
+/// Notification when config is successfully activated
+const NSNotificationName FIRRemoteConfigChangeNotification = @"FIRRemoteConfigChangeNotification";
```
erikeldridge, line 47: Minor (naming): could/should this be `FIRRemoteConfigActivateNotification` to disambiguate from realtime's `onConfigUpdateListener` naming?

chliangGoogle: Good call. Done.

PR #10067, opened 2022-08-03
Realtime rc failsafe

qdpham13:
Add back off flag to Realtime. If enabled, this will prevent Realtime from working until the app restarts.
Backend will have a new field called featureDisabled added to stream response. If this field is enabled, this will set stop Realtime from working until the next time the app is restarted. The current back off state in the SDK will be represented by the flag gIsBackoffEnabled.

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
@@ -146,6 +146,16 @@ static bool gIsRealtimeDisabled;
   return realtimeRemoteConfigQueue;
 }
 
+- (void)propogateErrors:(NSError *)error {
+  __weak RCNConfigRealtime *weakSelf = self;
+  dispatch_async(_realtimeLockQueue, ^{
+    __strong RCNConfigRealtime *strongSelf = weakSelf;
+    for (RCNConfigUpdateCompletion listener in strongSelf->_listeners) {
+      listener(error);
+    }
+  });
+}
+
 #pragma mark - Http Helpers
 
 - (NSString *)constructServerURL {
@@ -364,9 +374,7 @@ static bool gIsRealtimeDisabled;
                  }];
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000014", @"Cannot establish connection. Error: %@",
                   error);
-      for (RCNConfigUpdateCompletion listener in self->_listeners) {
-        listener(error);
-      }
+      [self propogateErrors:error];
     }
   });
 }
@@ -409,9 +417,7 @@ static bool gIsRealtimeDisabled;
                               FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000010",
                                           @"Failed to retrive config due to fetch error. Error: %@",
                                           error);
-                              for (RCNConfigUpdateCompletion listener in strongSelf->_listeners) {
-                                listener(error);
-                              }
+                              [self propogateErrors:error];
                             }
                           }
                         }];
@@ -442,9 +448,7 @@ static bool gIsRealtimeDisabled;
                           }];
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000011",
                   @"Ran out of fetch attempts, cannot find target config version.");
-      for (RCNConfigUpdateCompletion listener in strongSelf->_listeners) {
-        listener(error);
-      }
+      [self propogateErrors:error];
       return;
     }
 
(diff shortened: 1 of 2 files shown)


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -458,7 +464,15 @@ - (void)URLSession:(NSURLSession *)session
...
   NSString *targetTemplateVersion = _configFetch.templateVersionNumber;
   if (dataError == nil) {
-    targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    if ([response objectForKey:kTemplateVersionNumberKey]) {
+      targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    }
+    if ([response objectForKey:kIsFeatureDisabled]) {
+      gIsFeatureDisabled = [response objectForKey:kIsFeatureDisabled];
+      if (gIsFeatureDisabled) {
+        [self pauseRealtimeStream];
```
karenyz, line 473: do we still want to call autoFetch after pausing the stream? i thought featureDisabled would turn off realtime entirely

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -458,7 +464,15 @@ - (void)URLSession:(NSURLSession *)session
                                                              error:&dataError];
   NSString *targetTemplateVersion = _configFetch.templateVersionNumber;
   if (dataError == nil) {
-    targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    if ([response objectForKey:kTemplateVersionNumberKey]) {
+      targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    }
+    if ([response objectForKey:kIsFeatureDisabled]) {
```
karenyz, line 470: is there a scenario where these fields will not be populated?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -122,6 +125,7 @@ - (instancetype)init:(RCNConfigFetch *)configFetch
 
     gMaxRetryCount = gMaxRetries;
     gIsRetrying = false;
+    gIsFeatureDisabled = false;
```
karenyz, line 128: would suggest naming this `gIsRealtimeDisabled`

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -458,7 +464,15 @@ - (void)URLSession:(NSURLSession *)session
                                                              error:&dataError];
   NSString *targetTemplateVersion = _configFetch.templateVersionNumber;
   if (dataError == nil) {
-    targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    if ([response objectForKey:kTemplateVersionNumberKey]) {
+      targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    }
+    if ([response objectForKey:kIsFeatureDisabled]) {
+      gIsFeatureDisabled = [response objectForKey:kIsFeatureDisabled];
+      if (gIsFeatureDisabled) {
```
karenyz, line 472: can we add a test for this?

karenyz (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
karenyz: also should we return a RealtimeError to the listener when realtime is disabled?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Good catch, added a check to see of flag is enabled before autofetching

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: In theory there shouldn't be but the response may be mangled in some way. So added checks just in case.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Added a test for the flag and bubble error through listener

karenyz (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -448,6 +454,39 @@ - (void)autoFetch:(NSInteger)remainingAttempts targetVersion:(NSInteger)targetVe
...
+                              }];
+          for (RCNConfigUpdateCompletion listener in strongSelf->_listeners) {
+            listener(error);
+          }
+        });
+      }
+    }
+  }
+  if (!gIsRealtimeDisabled) {
+    [self autoFetch:gFetchAttempts targetVersion:[targetTemplateVersion integerValue]];
```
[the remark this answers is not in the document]
qdpham13: Moved it inside the nil check.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -448,6 +454,39 @@ - (void)autoFetch:(NSInteger)remainingAttempts targetVersion:(NSInteger)targetVe
...
 
+- (void)evaluateStreamResponse:(NSDictionary *)response error:(NSError *)dataError {
+  NSString *targetTemplateVersion = _configFetch.templateVersionNumber;
+  if (dataError == nil) {
+    if ([response objectForKey:kTemplateVersionNumberKey]) {
+      targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+    }
+    if ([response objectForKey:kIsFeatureDisabled]) {
+      gIsRealtimeDisabled = [response objectForKey:kIsFeatureDisabled];
+      if (gIsRealtimeDisabled) {
```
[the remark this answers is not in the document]
qdpham13: Updated.

FirebaseRemoteConfig/Tests/Unit/RCNRemoteConfigTest.m
```diff
@@ -1691,6 +1692,33 @@ - (void)testAddOnConfigUpdateMethodFail {
   }
 }
 
+- (void)testRealtimeBackoffDisabled {
```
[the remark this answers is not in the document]
qdpham13: Updated.

karenyz (approved):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -58,6 +60,7 @@
 static NSInteger gMaxRetryCount;
 static NSInteger gRetrySeconds;
 static bool gIsRetrying;
+static bool gIsRealtimeDisabled;
```
chliangGoogle, line 63: Not sure where my comments go, but these values' read and write should be guarded by a serial queue to ensure accuracy.

qdpham13: Do this in the patch PR plus realtimeDisabled flag only goes to true if activated, never back to false.

PR #10073, opened 2022-08-05
Fix module name for Promises and Protobuf

### LINKED ISSUE #10071: Linking FirebaseInstallation fails with lld

  * Xcode version: 13.4.1
  * Firebase SDK version: 8.7.0
  * Installation method: `Zip file` (select one)
  * Firebase Component: Analytics
  * Target platform(s): `All`

When linking our app using `lld`, we receive the following error: 
```
error: framework not found for -framework FBLPromises
```

We discovered that this is coming from `FirebaseInstallation` whose `LC_LINKER_OPTIONS` includes `FBLPromises` as a load command
```
...
  string #1 -framework
  string #2 FBLPromises
Load command 5
     cmd LC_LINKER_OPTION
...
```

It appears that `FBLPromises` was renamed to `PromisesObjC` and is a dependency of `FirebaseInstallation` & included in the zipped release of frameworks. We do link `PromisesObjC` into our bundle, but my guess is that the distribution step still links against the old `FBLPromises` name?

This doesn't throw an error in `ld` though since it'll ignore invalid frameworks included by the `LC_LINKER_OPTIONS` flag (https://github.com/llvm/llvm-project/issues/56939).
#### Steps to reproduce:
Link an app using `lld` against `FirebaseInstallation` + dependencies.

`error: framework not found for -framework FBLPromises`

paulb777:
Fix #10071

### DIFF
--- a/.github/workflows/zip.yml
+++ b/.github/workflows/zip.yml
@@ -123,7 +123,7 @@ jobs:
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseRemoteConfig/* \
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseAnalytics/FirebaseCore.xcframework \
                                                "${HOME}"/ios_frameworks/Firebase/FirebaseAnalytics/FirebaseCoreInternal.xcframework \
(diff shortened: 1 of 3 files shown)


ncooke3 (approved):
LGTM. Thanks for digging into this one!

PR #10077, opened 2022-08-06
Storage and StorageReference to Swift

paulb777:
This is the (probably) penultimate PR for the Storage migration. It migrates the Storage, StorageReference, and StoragePath classes and unit tests to Swift.  StorageTokenAuthorizer remains in Objective C.

One APIish change is changing APIs that threw Objective C exception to fatalError's since that should be much less disruptive than adding all clients to update to a `throws` API.

#no-changelog


FirebaseStorage/Sources/Internal/StoragePath.swift
```diff
@@ -0,0 +1,209 @@
...
+      return nil
+    }
+    var output = string
+    while true {
+      let newOutput = output.replacingOccurrences(of: "//", with: "/")
+      if newOutput == output {
+        break
+      }
+      output = newOutput
+    }
```
morganchen12, line 117: Kind of an obtuse way of writing this, but you could replace this `while true` loop with a reduce.

```swift
let slash: Character = "/"
output = string.reduce(into: "") { partialResult, next in
  guard partialResult.last.map({ $0 == slash }) ?? false else {
    return partialResult.append(next)
  }
  return next == slash ? () : partialResult.append(next) // this line here is especially code-golfy
}
```

Also, note that the method signature could be simplified from `String? -> String?` to `String -> String`, and we can instead lean on language constructs for dealing with optional values outside the function implementation.

FirebaseStorage/Sources/Internal/StoragePath.swift
```diff
@@ -0,0 +1,209 @@
...
+    if output.hasSuffix("/") {
+      output = String(output.dropLast())
+    }
+    if output.hasPrefix("/") {
+      return String(output.dropFirst())
+    } else {
+      return output
+    }
```
morganchen12, line 125: These ifs can be replaced by a single

```swift
return output.trimmingCharacters(in: ["/"])
```

FirebaseStorage/Sources/Storage.swift
```diff
@@ -164,7 +150,25 @@ import FirebaseAuthInterop
...
+      // If there exists a default bucket, throw if provided a different bucket.
+      if path.bucket != storageBucket {
+        fatalError("Provided bucket:\(path.bucket) does not match the Storage bucket of the current " +
+          "instance: \(storageBucket)")
+      }
+      return StorageReference(storage: self, path: path)
+    } catch let StoragePathError.storagePathError(message) {
+      fatalError(message)
+    } catch {
+      fatalError("Internal error finding StoragePath")
```
morganchen12, line 170: nit: Print the error in this case.

morganchen12 (approved):
LGTM with a handful of implementation suggestions.

FirebaseStorage/Sources/Internal/StoragePath.swift
```diff
@@ -0,0 +1,209 @@
...
+   */
+  static func path(string: String) throws -> StoragePath {
+    if string.hasPrefix("gs://") {
+      // "gs://bucket/path/to/object"
+      return try path(GSURI: string)
+    } else if string.hasPrefix("http://") || string.hasPrefix("https://") {
+      // "http[s]://firebasestorage.googleapis.com/bucket/path/to/object?signed_url_params"
+      return try path(HTTPURL: string)
+    } else {
+      // Invalid scheme, raise an exception!
```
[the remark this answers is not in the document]
paulb777: Fixed here and other places.

FirebaseStorage/Sources/Internal/StoragePath.swift
paulb777: Fixed the signature. The other part is interesting, but at least to me is significantly more obtuse, so I'll skip/.

FirebaseStorage/Sources/Internal/StoragePath.swift
paulb777: Nice!

FirebaseStorage/Sources/Storage.swift
paulb777: Done.

paulb777:
Merging, I'll address any additional comments in subsequent PRs

PR #10078, opened 2022-08-07
[Infra] Fix incorrectly named GHA job

ncooke3:
#no-changelog


ncooke3:
Based on https://github.com/google/GoogleDataTransport/pull/82#issuecomment-1207303748, I'm expecting this workflow to fail because the destination that the job's underlying xcodebuild command uses doesn't exist on Xcode that GHA is using.

paulb777 (approved):

ncooke3:
Ok, I was wrong. The watchOS workflow succeeded.

PR #10082, opened 2022-08-08
[Infra] Use watchOS device when using `scripts/build.sh` for watchOS

ncooke3:
### Context
Originally, [when the watchOS option in `scripts/build.sh`](https://github.com/firebase/firebase-ios-sdk/blob/a61f7b367b013b8fd9596aa85d6e0de0fad9d5d4/scripts/build.sh#L168-L170) was written, standalone watchOS apps were not yet a thing. So instead, the `--destination` of watchOS schemes ran on an iPhone as a companion target. This works but hid some failures that running on a watchOS destination would have surfaced (see #10085).

#no-changelog

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,5 +15,3 @@ source 'https://rubygems.org'
 gem 'cocoapods', '1.11.3'
 gem 'cocoapods-generate', '2.0.1'
 gem 'danger', '8.4.5'
-
-# Run CI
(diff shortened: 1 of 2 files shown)


ncooke3:
CI failures from [6a609f7](https://github.com/firebase/firebase-ios-sdk/pull/10082/commits/6a609f7a7fc9ab3bb7feaefd055b9525c34b3249) are addressed separately by #10085

paulb777 (approved):

andrewheard (approved):

PR #10088, opened 2022-08-09
Update versions for Release 9.5.0

chliangGoogle:


ncooke3 (approved):
LGTM on green.

paulb777 (approved):

paulb777:
Looks like an unrelated infra issue in the firestore tests - https://github.com/firebase/firebase-ios-sdk/runs/7755695980?check_suite_focus=true#step:5:42

chliangGoogle:
Yeah, will submit if the rest is green.

ncooke3:
I believe the failing watchOS workflow is due to #10082. It should have been surfaced by #10082 but wasn't so I have sent #10091 to fix that.

I'll look into fixing the watchOS workflow.

PR #10089, opened 2022-08-09
Realtime rc merge patch v1

qdpham13:
Fixes bugs from the bug bash. Also adds check for background so connection is not made while app is in the background.

- bug/241812063 - Too many config updates when reconnecting stream
- bug/241811829 - Multiple config updates during start up
- bug/241810568 - Duplicate Fetches

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
@@ -58,6 +58,7 @@ static NSInteger const gMaxRetries = 7;
 static NSInteger gMaxRetryCount;
 static NSInteger gRetrySeconds;
 static bool gIsRetrying;
+static bool gIsInBackground;
 
 @interface FIRConfigUpdateListenerRegistration ()
 @property(strong, atomic, nonnull) RCNConfigUpdateCompletion completionHandler;
@@ -122,6 +123,7 @@ static bool gIsRetrying;
 
     gMaxRetryCount = gMaxRetries;
     gIsRetrying = false;
+    gIsInBackground = false;
 
     [self setUpHttpRequest];
     [self setUpHttpSession];
@@ -327,7 +329,7 @@ static bool gIsRetrying;
 }
 
 - (bool)canMakeConnection {
-  return [self noRunningConnection] && [self->_listeners count] > 0;
+  return [self noRunningConnection] && [self->_listeners count] > 0 && !gIsInBackground;
 }
 
 #pragma mark - Retry Helpers
@@ -368,12 +370,28 @@ static bool gIsRetrying;
 - (void)backgroundChangeListener {
   [_notificationCenter addObserver:self
                           selector:@selector(isInForeground)
+                              name:@"UIApplicationWillEnterForegroundNotification"
+                            object:nil];
+
+  [_notificationCenter addObserver:self
+                          selector:@selector(isInBackground)
                               name:@"UIApplicationDidEnterBackgroundNotification"
                             object:nil];
 }
 
 - (void)isInForeground {
   [self beginRealtimeStream];
+  __weak RCNConfigRealtime *weakSelf = self;
+  dispatch_async(_realtimeLockQueue, ^{
+    __strong RCNConfigRealtime *strongSelf = weakSelf;
+    gIsInBackground = false;
+  });
+}
+
+- (void)isInBackground {
+  dispatch_async(_realtimeLockQueue, ^{
+    gIsInBackground = true;
+  });
 }
 
 #pragma mark - Autofetch Helpers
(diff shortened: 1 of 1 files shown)


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -455,17 +484,27 @@ - (void)URLSession:(NSURLSession *)session
...
+    FIRLogDebug(kFIRLoggerRemoteConfig, @"I-RCN000015",
+                @"Received config update message on stream.");
+    NSRange msgRange =
+        NSMakeRange(beginRange.location, endRange.location - beginRange.location + 1);
+    strData = [strData substringWithRange:msgRange];
+    data = [strData dataUsingEncoding:NSUTF8StringEncoding];
+    NSDictionary *response = [NSJSONSerialization JSONObjectWithData:data
+                                                             options:NSJSONReadingMutableContainers
+                                                               error:&dataError];
+    NSInteger oldTemplateVersion = [_configFetch.templateVersionNumber integerValue];
```
karenyz, line 499: "clientTemplateVersion"?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -455,17 +484,27 @@ - (void)URLSession:(NSURLSession *)session
...
+    if (dataError == nil) {
+      NSInteger updateTemplateVersion =
+          [[response objectForKey:kTemplateVersionNumberKey] integerValue];
+      if (updateTemplateVersion > oldTemplateVersion) {
```
karenyz, line 503: so when an experiment is updated the template version does not increment. should this be ">=" to handle that case?

and should we add a check to not autofetch if "templateVersion" is not in the response?

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -382,33 +400,44 @@ - (void)fetchLatestConfig:(NSInteger)remainingAttempts targetVersion:(NSInteger)
...
-    [strongSelf->_configFetch
-        fetchConfigWithExpirationDuration:0
-                        completionHandler:^(FIRRemoteConfigFetchStatus status, NSError *error) {
-                          if (status == FIRRemoteConfigFetchStatusSuccess) {
-                            if ([strongSelf->_configFetch.templateVersionNumber integerValue] >=
-                                targetVersion) {
-                              for (RCNConfigUpdateCompletion listener in strongSelf->_listeners) {
-                                listener(nil);
+    if (strongSelf->_settings.isFetchInProgress) {
+      if (strongSelf->_settings.lastFetchStatus == FIRRemoteConfigFetchStatusSuccess) {
```
karenyz, line 404: what happens if fetch status is not success or version < target version? for those cases I think we should still attempt to fetch?

karenyz (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -327,7 +329,7 @@ - (bool)noRunningConnection {
 }
 
 - (bool)canMakeConnection {
-  return [self noRunningConnection] && [self->_listeners count] > 0;
+  return [self noRunningConnection] && [self->_listeners count] > 0 && !gIsInBackground;
```
chliangGoogle, line 332: reading needs to be guarded by a serial queue as well

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -58,6 +58,7 @@
 static NSInteger gMaxRetryCount;
 static NSInteger gRetrySeconds;
 static bool gIsRetrying;
```
chliangGoogle, line 60: I think these parameters are safer to be set as ivar and set/get it with a function guarded by a serial queue.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -455,17 +484,27 @@ - (void)URLSession:(NSURLSession *)session
...
-  data = [strData dataUsingEncoding:NSUTF8StringEncoding];
-  NSDictionary *response = [NSJSONSerialization JSONObjectWithData:data
-                                                           options:NSJSONReadingMutableContainers
-                                                             error:&dataError];
-  NSString *targetTemplateVersion = _configFetch.templateVersionNumber;
-  if (dataError == nil) {
-    targetTemplateVersion = [response objectForKey:kTemplateVersionNumberKey];
+  NSRange endRange = [strData rangeOfString:@"}"];
+  NSRange beginRange = [strData rangeOfString:@"{"];
+  if (beginRange.location != NSNotFound && endRange.location != NSNotFound) {
```
chliangGoogle, line 489: what if beginRange is found and end is not? would that continue as expected?

chliangGoogle (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Changed to >= and added check for templateVersion in message

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated for this case.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Moved all of these vars to be wrapped by serial queue.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: moved to ivar and wrapped with serial queue

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: If end is not found we will just not regard it as a message. endRange is needed in order to parse it.

qdpham13:
PTAL when you get the chance. @chliangGoogle @karenyz

karenyz (approved):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -103,6 +99,11 @@ @implementation RCNConfigRealtime {
   RCNConfigSettings *_settings;
   FIROptions *_options;
   NSString *_namespace;
+  NSInteger gMaxRetryCount;
```
chliangGoogle, line 102: all the ivar should start with prefix '_'

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -103,6 +99,11 @@ @implementation RCNConfigRealtime {
...
+  NSInteger gRetrySeconds;
```
chliangGoogle, line 103: fyi, for all the ivar read, you can use dispatch_sync to guard it with a serial queue so it can returns the value right away.

I can't see the whole code base so just in case if there's reading from those ivars make sure you also guard it with serial queue.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
chliangGoogle: then you can remove the g if it's a iVar

chliangGoogle (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Yeah I made sure that all of these vars are guarded with serial queues

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -518,7 +518,7 @@ - (void)evaluateStreamResponse:(NSDictionary *)response error:(NSError *)dataErr
       [self propogateErrors:error];
     } else {
       NSInteger clientTemplateVersion = [_configFetch.templateVersionNumber integerValue];
-      if (updateTemplateVersion >= clientTemplateVersion) {
+      if (updateTemplateVersion > clientTemplateVersion) {
```
karenyz, line 521: just to confirm - we have the same behavior on Android?

qdpham13: Yeah it's like this already on Android, https://github.com/firebase/firebase-android-sdk/blob/cd8e0af2e8f462d7b3c88200b637580b18771ec8/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java#L160

chliangGoogle (approved):

PR #10090, opened 2022-08-10
Add setIndexConfiguration for client side indexing

cherylEnkidu:
Expose client side indexing feature with `FIRFirestore.setIndexConfigurationFromJSON` and `FIRFirestore.setIndexConfigurationFromStream`

### DIFF
--- a/Firestore/core/src/api/firestore.cc
+++ b/Firestore/core/src/api/firestore.cc
@@ -17,6 +17,7 @@
 #include "Firestore/core/src/api/firestore.h"
 
 #include <utility>
+#include <vector>
 
 #include "Firestore/core/src/api/collection_reference.h"
 #include "Firestore/core/src/api/document_reference.h"
@@ -31,13 +32,16 @@
 #include "Firestore/core/src/credentials/empty_credentials_provider.h"
 #include "Firestore/core/src/local/leveldb_persistence.h"
 #include "Firestore/core/src/model/document_key.h"
+#include "Firestore/core/src/model/field_path.h"
 #include "Firestore/core/src/model/resource_path.h"
 #include "Firestore/core/src/remote/firebase_metadata_provider.h"
 #include "Firestore/core/src/remote/grpc_connection.h"
 #include "Firestore/core/src/util/async_queue.h"
+#include "Firestore/core/src/util/exception.h"
 #include "Firestore/core/src/util/executor.h"
 #include "Firestore/core/src/util/hard_assert.h"
 #include "Firestore/core/src/util/status.h"
+#include "Firestore/third_party/nlohmann_json/json.hpp"
 #include "absl/memory/memory.h"
 
 namespace firebase {
@@ -49,13 +53,18 @@ using core::DatabaseInfo;
 using core::FirestoreClient;
 using credentials::AuthCredentialsProvider;
 using local::LevelDbPersistence;
+using model::FieldIndex;
+using model::FieldPath;
 using model::ResourcePath;
+using model::Segment;
+using nlohmann::json;
 using remote::FirebaseMetadataProvider;
 using remote::GrpcConnection;
 using util::AsyncQueue;
 using util::Empty;
 using util::Executor;
 using util::Status;
+using util::ThrowInvalidArgument;
 
 const int kDefaultTransactionMaxAttempts = 5;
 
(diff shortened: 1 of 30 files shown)


wu-hui (commented):
I have not finished with review yet, but thought it'd good to give some early feedback.

Firestore/core/src/api/firestore.cc
```diff
@@ -242,6 +251,68 @@ DatabaseInfo Firestore::MakeDatabaseInfo() const {
...
+    callback(Status(Error::kErrorFailedPrecondition,
+                    "Cannot enable indexes when persistence is disabled."));
+    return;
+  }
+
+  auto json_object =
+      nlohmann::json::parse(config.begin(), config.end(),
+                            /*callback=*/nullptr, /*allow_exceptions=*/false);
+  if (json_object.is_discarded()) {
+    callback(Status(Error::kErrorDataLoss, "Invalid Json format."));
```
wu-hui, line 269: I see you copied `kErrorDataLoss` from bundles, but I still think invalid argument makes more sense in this context.

Firestore/core/src/api/firestore.cc
```diff
@@ -242,6 +251,68 @@ DatabaseInfo Firestore::MakeDatabaseInfo() const {
...
+    return;
+  }
+
+  std::vector<FieldIndex> parsed_indexes;
+  const auto& json_indexes = reader.RequiredArray("indexes", json_object);
```
wu-hui, line 274: This should be OptionalArray?

Firestore/core/src/api/firestore.cc
```diff
@@ -242,6 +251,68 @@ DatabaseInfo Firestore::MakeDatabaseInfo() const {
...
+            Segment(std::move(field_path), Segment::Kind::kContains));
+      } else if ("ASCENDING" ==
+                 reader.OptionalString("order", json_field, default_string)) {
+        segments.emplace_back(
+            Segment(std::move(field_path), Segment::Kind::kAscending));
+      } else {
+        segments.emplace_back(
+            Segment(std::move(field_path), Segment::Kind::kDescending));
+      }
+    }
```
wu-hui, line 299: I wonder if we should add a check like line 305 here so we return immediately instead of moving to the next index.

Firestore/core/test/unit/local/local_store_test.cc
```diff
@@ -282,140 +230,52 @@ void LocalStoreTest::AcknowledgeMutationWithVersion(
...
       local_store_.ApplyBundledDocuments(DocVectorToMap(documents), "");
 }
 
-void LocalStoreTest::ResetPersistenceStats() {
+void LocalStoreTestBase::ResetPersistenceStats() {
   query_engine_.ResetCounts();
 }
 
-/** Asserts that the last target ID is the given number. */
-#define FSTAssertTargetID(target_id)       \
```
wu-hui, line 321: I promise I reviewed every line of these code movement :-)

Firestore/core/test/unit/local/leveldb_local_store_test.cc
```diff
@@ -46,6 +72,233 @@ INSTANTIATE_TEST_SUITE_P(LevelDbLocalStoreTest,
...
+  FieldIndex index_c =
+      FieldIndex{2,
+                 "coll",
+                 {model::Segment{Field("c1"), model::Segment::Kind::kAscending},
+                  model::Segment{Field("c2"), model::Segment::Kind::kContains}},
+                 model::FieldIndex::InitialState()};
+
+  ConfigureFieldIndexes({index_a, index_b});
+  auto field_indexes = GetFieldIndexes();
+  ASSERT_EQ(field_indexes, Vector(FieldIndex(index_b), FieldIndex(index_a)));
```
wu-hui, line 95: Android's `assertThat(fieldIndexes).containsExactly(indexA, indexB);` works regardless of the ordering of elements. You can probably use a `std::set` here since `FieldIndex` can be ordered.

Otherwise this test might break sometime in the future when nothing is actually wrong.

wu-hui (commented):
Finished review, overall good work, we are close to the finish line!

wu-hui (approved):
I think you forgot a comment about using `std::move`, but LGTM!

Congrats to making it to the finish line!

PR #10091, opened 2022-08-10
[Infra] Some workflows haven't been triggered on Gemfile changes

ncooke3:
### Context
I think the failing watchOS workflow in #10088 may be attributed to #10082. I noticed that triggering a global CI run in #10082 did not trigger this workflow. Further investigation revealed that the AppCheck workflow has not triggered on `Gemfile*` changes. This PR adds it as a trigger.

I also did an audit of the other workflows and added `Gemfile*` triggering where necessary so that all workflows that have at least one non-cron job trigger on the `Gemfile*` changes. (I excluded irrelevant workflows like notice generation and code coverage).

#no-changelog

### DIFF
--- a/.github/workflows/app_check.yml
+++ b/.github/workflows/app_check.yml
@@ -5,6 +5,7 @@ on:
     paths:
     - 'FirebaseAppCheck**'
     - '.github/workflows/app_check.yml'
+    - 'Gemfile*'
   schedule:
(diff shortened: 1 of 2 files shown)


ncooke3:
The `app_check / spm (watchOS)` failure is expected as that will be resolved separately.

paulb777 (approved):

PR #10093, opened 2022-08-11
[AppCheck] Update supported targets for AppAttestProvider

ncooke3:


paulb777 (approved):

FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
```diff
@@ -46,29 +46,9 @@
...
+#define FIR_APP_ATTEST_SUPPORTED_TARGETS TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_TV
 
 // `AppAttestProvider` availability annotations
 #define FIR_APP_ATTEST_PROVIDER_AVAILABILITY \
-  API_AVAILABLE(ios(14.0)) API_UNAVAILABLE(macos, tvos, watchos)
-
-#endif  // (defined(__IPHONE_14_5) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_14_5) || \
-          (defined(__MAC_11_3) && __MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_11_3) ||        \
-          (defined(__TVOS_14_5) && __TV_OS_VERSION_MAX_ALLOWED >= __TVOS_14_5)
+  API_AVAILABLE(macos(11.0), ios(14.0), tvos(15.0)) API_UNAVAILABLE(watchos)
```
andrewheard, line 54: non-blocking: This is similar to the DeviceCheck situation where App Attest isn't actually functional on macOS 11.0. Apple's documentation says macOS 11.0+ on the [DCAppAttestService](https://developer.apple.com/documentation/devicecheck/dcappattestservice) page but all of the methods/properties list 11.3+ (e.g., [`shared`](https://developer.apple.com/documentation/devicecheck/dcappattestservice/3573914-shared), [`isSupported`](https://developer.apple.com/documentation/devicecheck/dcappattestservice/3573915-issupported), and [`generateKey(completionHandler:)`](https://developer.apple.com/documentation/devicecheck/dcappattestservice/3573913-generatekey)). The availability values are also different for Mac Catalyst than iOS as well ([`14.5+`](https://developer.apple.com/documentation/devicecheck/dcappattestservice/3573915-issupported)).

We may want to audit these as part of [FR#10095](https://github.com/firebase/firebase-ios-sdk/issues/10095) for Firebase 10.

andrewheard, line 54: Sorry if this is a dumb question, do we need an annotation for `macCatalyst` now that `&& !TARGET_OS_MACCATALYST` has been removed or is it determined implicitly based on the `macos` and `ios` values for `API_AVAILABLE(...)`?

andrewheard (approved):
LGTM. All my comments are just suggestions for future work.

FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
ncooke3: Hmm good question. I don't think we need to. 

If you look at the `TargetConditionals.h` header, it has a nice diagram:
```
 *    +---------------------------------------------------------------------------+
 *    |                             TARGET_OS_MAC                                 |
 *    | +-----+ +-------------------------------------------------+ +-----------+ |
 *    | |     | |                  TARGET_OS_IPHONE               | |           | |
 *    | |     | | +-----------------+ +----+ +-------+ +--------+ | |           | |
 *    | |     | | |       IOS       | |    | |       | |        | | |           | |
 *    | | OSX | | | +-------------+ | | TV | | WATCH | | BRIDGE | | | DRIVERKIT | |
 *    | |     | | | | MACCATALYST | | |    | |       | |        | | |           | |
 *    | |     | | | +-------------+ | |    | |       | |        | | |           | |
 *    | |     | | +-----------------+ +----+ +-------+ +--------+ | |           | |
 *    | +-----+ +-------------------------------------------------+ +-----------+ |
 *    +---------------------------------------------------------------------------+
```

MacCatalyst is a subset of iOS so I typically don't include it unless I'm excluding it.

As for the `API_AVAILABLE` annotation, it doesn't look like we have any precedent in the codebase for using it. So I guess it is implicitly derived based on the iOS version. In the case of App Attest support, iOS the macCatalyst support versions are in sync (iOS/macCatalyst 14.0+).

ncooke3: Added a comment in #10095

andrewheard: That's super helpful, thank you!

ncooke3: Marking as resolved to merge, but feel free to comment and I can address any feedback in next PR!

PR #10094, opened 2022-08-11
[AppCheck] DeviceCheck provider is supported by watchOS 9.0+

ncooke3:
### Context
This PR achieves two things:
1. Resolves recent [AppCheck / watchOS CI issue](https://github.com/firebase/firebase-ios-sdk/pull/10088#issuecomment-1210908096). 
2. `DeviceCheck` now has [beta support for watchOS 9.0+](https://developer.apple.com/documentation/devicecheck) in Xcode 14. This means that we can update AppCheck's DeviceCheck provider to reflect the new availability.

I validated that `app_check / spm (watchOS)`'s underlying script passed locally on Xcode 13.3.1.

### DIFF
--- a/FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
+++ b/FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
@@ -16,9 +16,24 @@
 
 // Availability conditions for different App Check SDK components.
 
+#import <Foundation/Foundation.h>
 #import <TargetConditionals.h>
 
 #pragma mark - DeviceCheck
+
+// DeviceCheck availability was extended to watchOS in Xcode 14.
+#if defined(__WATCHOS_9_0) && __WATCH_OS_VERSION_MAX_ALLOWED >= __WATCHOS_9_0
+
+// Targets where DeviceCheck framework is available to be used in preprocessor conditions.
+#define FIR_DEVICE_CHECK_SUPPORTED_TARGETS \
+  TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_TV || TARGET_OS_WATCH
+
+// `DeviceCheckProvider` availability.
+#define FIR_DEVICE_CHECK_PROVIDER_AVAILABILITY \
+  API_AVAILABLE(ios(11.0), macos(10.15), tvos(11.0), watchos(9.0))
+
+// TODO(ncooke3): Remove `#else` clause when Xcode 14 is the minimum supported Xcode.
+#else  // defined(__WATCHOS_9_0) && __WATCH_OS_VERSION_MAX_ALLOWED >= __WATCHOS_9_0
 
 // Targets where DeviceCheck framework is available to be used in preprocessor conditions.
 #define FIR_DEVICE_CHECK_SUPPORTED_TARGETS TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_TV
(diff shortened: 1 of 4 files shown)


FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
```diff
@@ -16,17 +16,34 @@
 
 // Availability conditions for different App Check SDK components.
 
+#import <Foundation/Foundation.h>
```
ncooke3, line 19: I struggled on this for awhile before realizing the Foundation import was needed to evaluate `defined(__WATCHOS_9_0)` as truthy. I don't know why because it **should** be in the prelude but the `Foundation` import definitely makes a difference... sigh

FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
```diff
@@ -16,17 +16,34 @@
...
+// DeviceCheck availability was extended to watchOS in Xcode 14.
+#if defined(__WATCHOS_9_0) && __WATCH_OS_VERSION_MAX_ALLOWED >= __WATCHOS_9_0
+
+// Targets where DeviceCheck framework is available to be used in preprocessor conditions.
+#define FIR_DEVICE_CHECK_SUPPORTED_TARGETS \
+  TARGET_OS_IOS || TARGET_OS_OSX || TARGET_OS_TV || TARGET_OS_WATCH
+
+// `DeviceCheckProvider` availability.
+#define FIR_DEVICE_CHECK_PROVIDER_AVAILABILITY \
+  API_AVAILABLE(ios(11.0), macos(10.15), tvos(11.0), watchos(9.0))
```
andrewheard, line 33: non-blocking: This is outside the scope of the review but I wonder if we should revisit the choice to specify `macos(10.15)` as the minimum version. It is the version specified by Apple but essentially it's a no-op on Catalina (`DCDevice.current.isSupported` returns false). I worry that it could mislead users into thinking they can use App Attest (using DeviceCheck) on macOS 10.15.

FirebaseAppCheck/Tests/Unit/Core/FIRAppCheckTests.m
```diff
@@ -38,6 +38,10 @@
 
 #import "FirebaseCore/Extension/FirebaseCoreInternal.h"
 
+// Because the DeviceCheck provider is the default provider for AppCheck. There
+// may be test cases that are dependent on DeviceCheck being available.
```
andrewheard, line 42: super nit: Maybe something like 
> Since DeviceCheck is the default attestation provider for AppCheck, disable test cases that may be dependent on DeviceCheck being available.

I think changing the period to a comma would also totally work too 😀

andrewheard (approved):
Awesome, I'm glad Apple is adding DeviceCheck support to more platforms since it's fairly lightweight. Thanks for getting the SDK ready for release!

FirebaseAppCheck/Tests/Unit/Core/FIRAppCheckTests.m
ncooke3: Done in d7bff0b

FirebaseAppCheck/Sources/Public/FirebaseAppCheck/FIRAppCheckAvailability.h
ncooke3: Hmm. Interesting, I did not know that. Thanks for bringing that up. We could create an issue to track that for Firebase 10

andrewheard: No problem! I've added an [FR](https://github.com/firebase/firebase-ios-sdk/issues/10095#issue-1336561148) with the details I know.

Note: IIRC, I did test DeviceCheck on iOS 11 but I haven't personally tried it on tvOS 11 (and I can't downgrade to test it). I would assume the non-macOS availability labels are accurate though.

paulb777 (approved):

chliangGoogle (approved):

PR #10112, opened 2022-08-18
[Core] Support watchOS lifecycle notifications

ncooke3:
### Context
Lifecycle notifications have not been observed on watchOS so watchOS heartbeats are not created. 

The [`WKApplicationDidBecomeActiveNotification`](https://developer.apple.com/documentation/watchkit/wkapplicationdidbecomeactivenotification?language=objc) notification that is being observed it only available on `watchOS 7.0+`.

Our [SPM distribution has watchOS `7.0`](https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L26) as the minimum version but our [CocoaPods specs have `6.0`](https://github.com/firebase/firebase-ios-sdk/search?l=Ruby&q=watchos).

In an effort to maintain compatibility with `watchOS 6.0`, this PR adds support for the following behavior:
- On watchOS 7.0+, heartbeats are captured when the `WKApplicationDidBecomeActiveNotification` is fired.
- On watchOS 6.0, heartbeats are captured on app configuration. This is triggered by the already existing [`kFIRAppReadyToConfigureSDKNotification`](https://github.com/firebase/firebase-ios-sdk/blob/f8ee68bfc7873efbfdfcea831a20435c4b89f875/FirebaseCore/Sources/FIRApp.m#L459) notification.

#no-changelog

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -876,7 +876,16 @@ static FIRApp *sDefaultApp;
 #elif TARGET_OS_OSX
   NSNotificationName notificationName = NSApplicationDidBecomeActiveNotification;
 #elif TARGET_OS_WATCH
-  NSNotificationName notificationName = WKApplicationDidBecomeActiveNotification;
+  // TODO(ncooke3): Remove when minimum supported watchOS version is watchOS 7.0.
+  // On watchOS 7.0+, heartbeats are logged when the watch app becomes active.
+  // On watchOS 6.0, heartbeats are logged when the Firebase app is configuring.
+  // While it does not cover all use cases, logging when the Firebase app is
+  // configuring is done because watchOS lifecycle notifications are a
+  // watchOS 7.0+ feature.
+  NSNotificationName notificationName = kFIRAppReadyToConfigureSDKNotification;
+  if (@available(watchOS 7.0, *)) {
+    notificationName = WKApplicationDidBecomeActiveNotification;
+  }
 #endif
 
   [[NSNotificationCenter defaultCenter] addObserver:self
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -22,6 +22,10 @@
 #import <AppKit/AppKit.h>
 #endif
 
+#if __has_include(<WatchKit/WatchKit.h>)
+#import <WatchKit/WatchKit.h>
+#endif
+
 #import "FirebaseCore/Sources/Public/FirebaseCore/FIRApp.h"
 
 #import "FirebaseCore/Sources/FIRAnalyticsConfiguration.h"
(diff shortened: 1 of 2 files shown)


paulb777 (commented):
Please update the podspecs to 7 and add a release note about the minimum version change.

FirebaseCore/Sources/FIRApp.m
```diff
@@ -871,14 +879,16 @@ - (void)subscribeForAppDidBecomeActiveNotifications {
   NSNotificationName notificationName = UIApplicationDidBecomeActiveNotification;
 #elif TARGET_OS_OSX
   NSNotificationName notificationName = NSApplicationDidBecomeActiveNotification;
+#elif TARGET_OS_WATCH_7_OR_LATER
+  NSNotificationName notificationName = WKApplicationDidBecomeActiveNotification;
 #endif
```
chliangGoogle, line 884: Here add
#elif TARGET_OS_WATCH_6 or before?
[self logCoreTelemetryIfEnabled]

ncooke3: I understand. I actually need to change the implementation to do runtime checks because build time checking for the OS version will not work as expected.

paulb777: @ncooke3 See https://github.com/google/GoogleUtilities/blob/main/GoogleUtilities/Environment/third_party/GULAppEnvironmentUtil.m#L245 for references on runtime version checking

ncooke3: Thanks, I looked into it but ultimately used `if (@available(watchOS 7.0, *))` since it was more lightweight. LMK what you think...

paulb777: Agreed. LGTM

paulb777 (approved):

ncooke3:
I'm going to do a round of manual testing in an actual app before merging. Just to double check everything is as expected.

chliangGoogle (approved):

FirebaseCore/Sources/FIRApp.m
```diff
@@ -871,14 +875,23 @@ - (void)subscribeForAppDidBecomeActiveNotifications {
...
 #elif TARGET_OS_OSX
   NSNotificationName notificationName = NSApplicationDidBecomeActiveNotification;
+#elif TARGET_OS_WATCH
+  // TODO(ncooke3): Remove when minimum supported watchOS version is watchOS 7.0.
+  // On watchOS 7.0+, heartbeats are logged when the watch app becomes active.
+  // On watchOS 6.0, heartbeats are logged when the Firebase app is configuring.
+  // While it does not cover all use cases, logging when the Firebase app is
+  // configuring is done because watchOS lifecycle notifications are a
+  // watchOS 7.0+ feature.
+  NSNotificationName notificationName = kFIRAppReadyToConfigureSDKNotification;
```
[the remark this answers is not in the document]
ncooke3: We rely on this notification because cannot observe app launches on watchOS 6.

ncooke3:
I manually tested the following cases:
1. `WKApplicationDidBecomeActiveNotification` was available and used
2. `WKApplicationDidBecomeActiveNotification` was not available and kFIRAppReadyToConfigureSDKNotification was used instead

Both cases successfully triggered the logging of a heartbeat.

PR #10114, opened 2022-08-18
[watchOS] Bump watchOS minimum version support to watchOS 7.0

ncooke3:
### Context
This will unblock use of watchOS 7.0+ APIs

```
git grep --name-only "'6.0'" | xargs sed -i '' 's/'6.0'/'7.0'/g'
```


ncooke3:
CI was green. Closing though due to change in plans.

PR #10131, opened 2022-08-23
Reenable any GTMSessionFetcher 2.x dependency

paulb777:
More term long-term sustainable solution than #10123 to the deprecation warning introduced by GTMSessionFetcher 2.1


ncooke3 (approved):

petea (approved):

andrewheard (commented):
Thanks, Paul!

PR #10134, opened 2022-08-24
Fix Analytics source build

paulb777:
See internal b/243449951

### DIFF
--- a/ReleaseTooling/Sources/ZipBuilder/CocoaPodUtils.swift
+++ b/ReleaseTooling/Sources/ZipBuilder/CocoaPodUtils.swift
@@ -551,6 +551,7 @@ enum CocoaPodUtils {
   private static func isSourcePodspec(_ podspecPath: String) -> Bool {
     do {
       let contents = try String(contentsOfFile: podspecPath, encoding: .utf8)
+      // The presence of ".vendored_frameworks" in a podspec indicates a binary pod.
       return contents.range(of: ".vendored_frameworks") == nil
     } catch {
       fatalError("Could not read \(podspecPath): \(error)")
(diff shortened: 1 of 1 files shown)


ReleaseTooling/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -482,16 +482,11 @@ enum CocoaPodUtils {
...
-        .map(\.name)
-        .contains(rootPod)
-
       // Check if we want to use a local version of the podspec.
-      if !podIsClosedSource,
-         let localURL = localPodspecPath,
-         FileManager.default
-         .fileExists(atPath: localURL.appendingPathComponent("\(rootPod).podspec").path) {
+      if let localURL = localPodspecPath,
+         let pathURL = localPodspecPath?.appendingPathComponent("\(rootPod).podspec").path,
```
ncooke3, line 487: optional nit: the `localURL` could just be reused since you know it's there by the time `pathURL` is assigned. not a big deal though.

```suggestion
         let pathURL = localURL.appendingPathComponent("\(rootPod).podspec").path,
```

ReleaseTooling/Sources/ZipBuilder/CocoaPodUtils.swift
```diff
@@ -553,6 +548,15 @@ enum CocoaPodUtils {
     return podfile
   }
 
+  private static func isSourcePodspec(_ podspecPath: String) -> Bool {
+    do {
+      let contents = try String(contentsOfFile: podspecPath, encoding: .utf8)
+      return contents.range(of: ".vendored_frameworks") == nil
```
ncooke3, line 554: optional: if you feel it may be helpful, add a comment explaining why the code checks for the presence of `vendored_frameworks`

ncooke3 (approved):
LGTM! Learned something here. Let two comments that are totally optional.

ReleaseTooling/Sources/ZipBuilder/CocoaPodUtils.swift
paulb777: Would be nice, but

Screen Shot 2022-08-24 at 9 24 55 PM

It feels like there should be a way to do this, but I'm not sure what it would be ....

PR #10139, opened 2022-08-25
[Infra] Add SPM debug info to issue template

ncooke3:
### Context
For CocoaPods-related issues, we request the `Podfile.lock`.

Now, for SwiftPM-related issues, we request the `Package.resolved`. Thanks @rizafran for the idea!

See render template at [here](https://github.com/firebase/firebase-ios-sdk/blob/nc/swiftpm-issue-template/.github/ISSUE_TEMPLATE/BUG_REPORT.yml)

#no-changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -144,7 +144,7 @@ body:
         <br>
 
         ```yml
-        
+
         Replace this line with the contents of your Podfile.lock!
 
         ```
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -124,7 +124,7 @@ body:
         <br>
 
         ```json
-        
+
         Replace this line with the contents of your Package.resolved.
 
         ```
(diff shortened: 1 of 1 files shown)


paulb777 (approved):

morganchen12 (approved):
LGTM

Devs using SPM through Xcode may not know how to find their Package.resolved (in Xcode 14, it's at `Project.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`), so we'll likely need to add this to a playbook somewhere. Or maybe directly to this template?

rizafran (approved):
LGTM

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -112,6 +112,26 @@ body:
       render: shell
```
ncooke3, line 112: Thanks @morganchen12 

> Devs using SPM through Xcode may not know how to find their Package.resolved (in Xcode 14, it's at `Project.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`), so we'll likely need to add this to a playbook somewhere. Or maybe directly to this template?

This is new behavior, right? Is the Package.resolved not dumped in the same directory as the Package.swift?

peterfriese: > This is new behavior, right?

Pretty sure it is.

ncooke3: Thanks @morganchen12 and @peterfriese. 

I'll see if devs are including it or not in their new issues and go from there. 

I wonder what the motivation was to hiding the Package.resolved in .build/... I think I prefer it in the same directory at the Package.swift.

PR #10143, opened 2022-08-26
[RC] Fix retry logic to respect max retry count

karenyz:
Currently we are optimistically resetting the retry count in `beginRealtimeStream` immediately after starting the data task, however the request may fail, and the reset retry count causes us to retry infinitely (see internal b/241811665)

- Moving the logic to be handled in the data task delegates and only reset retry count when request has succeeded.
- Removed some duplicate checks in `retryHTTPConnection` because `beginRealtimeStream` has the same checks
- Some minor renaming

Able to repro/verify in the sample app by turning off the device network to generate an error

Ideally this should be covered in a unit test, however there is no existing testing for any of the data task interactions, so it's slightly tricky to add and will be done in a following PR


danasilver (approved):
:shipit: 🚀

chliangGoogle (approved):

PR #10145, opened 2022-08-26
Update versions for Release 9.6.0

andrewheard:
Updated version numbers in preparation for the next Firebase release (9.6.0).

#no-changelog

PR #10146, opened 2022-08-26
[AppCheck] Disable tests that use keychain

ncooke3:
### Context
- Intentionally caused by https://github.com/google/GoogleUtilities/pull/75
- While we do lose some coverage on CI, [GoogleUtilities/#75](https://github.com/google/GoogleUtilities/pull/75) makes the macOS keychain behave like the iOS style keychain (which is tested on CI).
- Also cleaned up the `AppCheckUnit` test target in the Package manifest. I think prefer excluding at the test file level to keep things consistent between build systems and the source of truth in the test file.
- Unblocks some of the CI in #10145

### Future work
- [ ] Open a PR to update AppCheck CHANGELOG with note mentioning new macOS constraint (need to use Keychain Sharing capability)

#no-changelog

### DIFF
--- a/FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
+++ b/FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#import <TargetConditionals.h>
+
 // Tests that use the Keychain require a host app and Swift Package Manager
 // does not support adding a host app to test targets.
 #if !SWIFT_PACKAGE
--- a/FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
+++ b/FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+// Tests that use the Keychain require a host app and Swift Package Manager
+// does not support adding a host app to test targets.
+#if !SWIFT_PACKAGE
+
+// Skip keychain tests on Catalyst and macOS. Tests are skipped because they
+// involve interactions with the keychain that require a provisioning profile.
+// See go/firebase-macos-keychain-popups for more details.
+#if !TARGET_OS_MACCATALYST && !TARGET_OS_OSX
+
 #import <XCTest/XCTest.h>
 
 #import "FBLPromise+Testing.h"
@@ -84,3 +93,7 @@
 }
 
 @end
+
+#endif  // !TARGET_OS_MACCATALYST && !TARGET_OS_OSX
+
+#endif  // !SWIFT_PACKAGE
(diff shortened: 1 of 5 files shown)


FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
```diff
@@ -14,6 +14,17 @@
  * limitations under the License.
  */
 
+#import <TargetConditionals.h>
+
+// Tests that use the Keychain require a host app and Swift Package Manager
+// does not support adding a host app to test targets.
+#if !SWIFT_PACKAGE
```
paulb777, line 21: Why does SWIFT_PACKAGE need to be excluded?  What's changed about this for iOS and tvOS?

Package.swift
```diff
@@ -1256,12 +1256,6 @@ let package = Package(
       exclude: [
         // Disable Swift tests as mixed targets are not supported (Xcode 12.3).
         "Unit/Swift",
-
-        // Disable Keychain dependent tests as they require a host application on iOS.
```
paulb777, line 1260: Why add the requirement for a host application on iOS?

ncooke3: The requirement of the host application is a SwiftPM limitation: https://forums.swift.org/t/host-application-for-spm-tests/24363/12

FirebaseAppCheck/Tests/Integration/FIRDeviceCheckAPIServiceE2ETests.m
ncooke3: So this file has one test and it has been disabled for some time. I just saw it was included in the Package.manifest under the exclusions due to host app requirement.

After noticing the tests were disabled, I added a TODO for myself a few months ago.

ncooke3: It's not urgent because App Check has excellent test coverage but it would be nice to get this file fixed so the test runs.

Package.swift
paulb777: I see, I read this backwards - as a new exclusion instead of an exclusion delete.

paulb777 (approved):
LGTM on green. thanks!

FirebaseAppCheck/Tests/Unit/Core/FIRAppCheckIntegrationTests.m
```diff
@@ -62,6 +64,8 @@ @interface FIRAppCheckIntegrationTests : XCTestCase
 @property(nonatomic, nullable) id mockAppCheckProvider;
 @property(nonatomic, nullable) id mockTokenRefresher;
 
+- (void)testDefaultAppCheckProvider FIR_DEVICE_CHECK_PROVIDER_AVAILABILITY;
```
ncooke3, line 67: @paulb777, I thought this was pretty clever. This test required DeviceCheck so rather than do a runtime check in the test method, do a compile time check. Now, you can tell whether DeviceCheck was available based on the test log.

paulb777: Maybe too clever for me. 😛 Might need to chat about this one.

Issue #5973, opened 2020-07-04
FR: Storage - use NSProgress objects properly when reporting progress, not re-initting them every time


xaphod:
### Feature proposal

* Firebase Component: Storage

In `FIRStorageTask.m` there's this code for reporting progress:
```
- (FIRStorageTaskSnapshot *)snapshot {
  @synchronized(self) {
    NSProgress *progress = [NSProgress progressWithTotalUnitCount:self.progress.totalUnitCount];
```

A new NSProgress object is created every time there's a snapshot. This makes it impossible to use NSProgress objects how they are often used: they get added as a child of another NSProgress object, via `NSProgress.addChild(...)`. That's because often one has lots of download tasks and wants to report overall/aggregate progress to the user -- and rather than having to calculate this every time something changes, one just observes a parent NSProgress object.

Ask: instead of making a new NSProgress object every time, please reuse one NSProgress object per task, specifically so we can add it as a child of a "parent" NSProgress object.

Bonus ask, worth 10,000 points: also support NSProgress' pause/unpause/cancel functionality, so we can easily tie in pausing multiple tasks at once

Thanks!

Dionne-Rodriguez:
spent hours trying to implement my own logic to handle multiple image uploads, but it is very hard to track the progress of each unique upload because NSProgress gets recreated everytime. Hopefully the firebase devs can fix this issue

paulb777:
Sorry about the trouble. We're happy to review a PR that addresses this. Otherwise, we'll see if we can include better NSProgress support in this year's [Swift API improvement initiative](https://github.com/firebase/firebase-ios-sdk/blob/master/ROADMAP.md).

xaphod:
@Dionne-Rodriguez Here's how I handled it in case it helps. I'm in a rush so if you have questions just ask.
Key learning: use the magical `Progress.current()`.
A `Download` is a struct containing a `Progress`.

```
            var download = Download.init(storagePath: storagePath)
            download.progress = Progress.init(parent: Progress.current(), userInfo: nil)
            let path = subfolder + "/" + String(localFilename)
            download.pathRelativeToFolder = path
            self.downloads[storagePath] = download
            let fullPath = self.folderUrl.appendingPathComponent(path)
            self.printDebugProgressInfo(for: download)

            let storageRef = Storage.storage().reference(withPath: storagePath)
            // get the metadata up front so we see the total before the download finishes for small files
            storageRef.getMetadata { (meta, error) in
[...]

PR #10161, opened 2022-09-01
Storage v2 API proof of concept

paulb777:
Proof of concept of a Swift-only API for the StorageReference APIs available for iOS 13+. This PR provides an example implementation and integration tests for `putFile` and `putData` .

- The v2 API is now an async/await API that returns on completion and throws on errors
- If task control(cancel, pause, resume) is wanted, it can be controlled by a passed in `Progress` instance. See https://developer.apple.com/documentation/foundation/progress

For example, the new API for putFile are :

-       @discardableResult
        func putFileV2(from fileURL: URL,
                 metadata: StorageMetadata? = nil,
                 progress: Progress? = nil,
                 progressBlock: ((Progress) -> Void)? = nil) async throws -> StorageMetadata {

- The public StorageUploadTask class is replaced by the internal helper StorageUploadTaskV2 class.
- progress tracking is available via an optional callback instead of a state management system
- An immediate `cancel` works, but `cancel` from a progress callback causes 
```
SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation!
2022-09-02 13:47:25.949984-0700 AppHost-FirebaseStorage-Unit-Tests[38388:17164810] SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation!
```
I believe it's because GTMSessionFetcher deallocates its callbacks after a call to `stopFetching`.

### DIFF
--- a/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
+++ b/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
@@ -81,6 +81,7 @@ internal class StorageUploadTaskV2: StorageTask {
       uploadFetcher.comment = "File UploadTask"
     }
     uploadFetcher.maxRetryInterval = reference.storage.maxUploadRetryInterval
+    uploadFetcher.stopFetchingTriggersCompletionHandler = true
 
     if let progressBlock = progressBlock {
       uploadFetcher.sendProgressBlock = { (bytesSent: Int64, totalBytesSent: Int64,
--- a/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
+++ b/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
@@ -35,6 +35,9 @@ internal class StorageUploadTaskV2: StorageTask {
    * Prepares a GTMSessionFetcher task and does an upload.
    */
   internal func upload() async throws -> StorageMetadata {
+    if progress.isCancelled {
+      throw StorageError.cancelled
+    }
     if let contentValidationError = isContentToUploadInvalid() {
       throw StorageError.swiftConvert(objcError: contentValidationError)
     }
--- a/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
+++ b/FirebaseStorage/Sources/Internal/StorageUploadTaskV2.swift
@@ -89,6 +89,9 @@ internal class StorageUploadTaskV2: StorageTask {
           self.progress.totalUnitCount = totalBytesExpectedToSend
           self.metadata = self.uploadMetadata
           progressBlock(self.progress)
+          if self.progress.isCancelled {
+            uploadFetcher.stopFetching()
+          }
       }
     }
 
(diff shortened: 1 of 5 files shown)


mortenbekditlevsen:
Really exciting work!

Regarding the second api:
`putFileHandle`

Can't the returned task be generic over the `StorageMetadata` rather than `Void` and then return the metadata inside the task, rather than calling the completion?

It seems to me that the completion could then be made redundant - and the task could be used both for cancellation handling or it could be awaited for the metadata.

I only gave it a quick read so I could be misunderstanding some detail...

paulb777:
Great point @mortenbekditlevsen ! I'll see if I can get it to work.

Also, @ryanwilson pointed out that a more correct implementation of Progress may address the pause/resume question as well as #5973. See https://developer.apple.com/documentation/foundation/progress/1412377-pause

paulb777:
The code is cleaner after following @mortenbekditlevsen's suggestion, but it broke cancellation. I'll revisit when digging deeper in to Progress management.

mortenbekditlevsen:
Regarding task cancellation: Note that even if you don't explicitly return a `Task`, but instead have an async function, any async function are always called in the context of a `Task`, so _that_ task can be cancelled, and the cancellation state will be propagated to async code.

If you want to bridge task cancellation from the Swift concurrency world to previous completion base code, you can use the `withTaskCancellationHandler` API.

```
        await withTaskCancellationHandler {
            self.cancel() // <- a method to cancel the work happening in the gtmsessionfetcher
        } operation: {
            await withCheckedContinuation { continuation in
                 ... 
            }
        }
```

paulb777:
Thanks @mortenbekditlevsen 

When changing to 

```
  private func beginFetch(uploadFetcher: GTMSessionUploadFetcher) async -> (Data?, Error?) {
    return await withTaskCancellationHandler {
      uploadFetcher.stopFetching() // <- a method to cancel the work happening in the gtmsessionfetcher
    } operation: {
      await withCheckedContinuation { continuation in
        uploadFetcher.beginFetch { data, error in
          continuation.resume(returning: (data, error))
        }
      }
    }
  }
```

I see:

```
SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation!
2022-09-02 09:01:48.428477-0700 AppHost-FirebaseStorage-Unit-Tests[7137:16856931] SWIFT TASK CONTINUATION MISUSE: beginFetch(uploadFetcher:) leaked its continuation!
```

I did fix `testSimplePutFileWithCancel` by adding `let _ = try await task.value` in the `do`.

Still failing `testSimplePutFileWithCancelFromProgress` though. I need to find a way to check Cancellation once the upload has already started ...

paulb777:
GTMSessionFetcher now supports cancellation from an async/await `beginFetch` call from https://github.com/google/gtm-session-fetcher/pull/328

paulb777:
Going to close this. On further thought, it's probably fine to have separate APIs for async/await behavior and APIs to manage task run control and progress monitoring.

PR #10162, opened 2022-09-01
Save template version in settings

qdpham13:
Save template version in settings

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigFetch.m
@@ -64,9 +64,6 @@ static NSInteger const kRCNFetchResponseHTTPStatusCodeGatewayTimeout = 504;
 // Deprecated error code previously from FirebaseCore
 static const NSInteger sFIRErrorCodeConfigFailed = -114;
 
-// Key to retrive fetch config template version number
-static NSString *const kTemplateVersionNumberKey = @"templateVersion";
-
 #pragma mark - RCNConfig
 
 @implementation RCNConfigFetch {
@@ -105,7 +102,7 @@ static NSString *const kTemplateVersionNumberKey = @"templateVersion";
     _content = content;
     _fetchSession = [self newFetchSession];
     _options = options;
-    _templateVersionNumber = [self getTemplateVersionNumber:_content.fetchedConfig];
+    _templateVersionNumber = [self->_settings lastTemplateVersion];
   }
   return self;
 }
@@ -374,7 +371,7 @@ static NSString *const kTemplateVersionNumberKey = @"templateVersion";
 
       if (error || (statusCode != kRCNFetchResponseHTTPStatusCodeOK)) {
         // Update metadata about fetch failure.
-        [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO];
+        [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO templateVersion:@"1"];
         if (error) {
           if (strongSelf->_settings.lastFetchStatus == FIRRemoteConfigFetchStatusSuccess) {
             FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000025",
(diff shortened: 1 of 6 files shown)


FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -585,9 +584,10 @@ - (NSURLSessionDataTask *)URLSessionDataTaskWithContent:(NSData *)content
 }
 
 - (NSString *)getTemplateVersionNumber:(NSDictionary *)fetchedConfig {
```
karenyz, line 586: here it looks like we are still trying to read from the actual config.. however it doesn't look like we need this method anymore?

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
```diff
@@ -110,6 +110,7 @@ - (instancetype)initWithDatabaseManager:(RCNConfigDBManager *)manager
     }
 
     _isFetchInProgress = NO;
+    _lastTemplateVersion = @"1";
```
karenyz, line 113: can we store this in `_userDefaultsManager`? (see `lastFetchTimeInterval` for example). i think this will allow us to avoid any DB interactions (we longer have to add templateVersion to device context)

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
```diff
@@ -374,7 +371,7 @@ - (void)fetchWithUserProperties:(NSDictionary *)userProperties
 
       if (error || (statusCode != kRCNFetchResponseHTTPStatusCodeOK)) {
         // Update metadata about fetch failure.
-        [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO];
+        [strongSelf->_settings updateMetadataWithFetchSuccessStatus:NO templateVersion:@"1"];
```
karenyz, line 374: it looks like templateVersion isn't updated if fetch fails, but i would pass templateVersion = 0 or nil just to be clear.

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
karenyz: another thought is this shouldn't be initialized as 1 because if the developer publishes their first version then the client won't fetch?

karenyz (changes requested):

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
qdpham13: The template version variable is stored in memory in this file. It would be initiated from disk using RCNConfigSettings and then updated every fetch request as well as saved to disk. This function is used to extract it from the fetch response.

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
qdpham13: I don't think so, it doesn't look like it's saved to disk (https://github.com/firebase/firebase-ios-sdk/blob/96450bd48f3ca6ec7e7854a45f95d2a1449e1296/FirebaseRemoteConfig/Sources/RCNConfigSettings.m#L101). 
And that's a good point, I'll initialize it to 0.

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
qdpham13: Done.

FirebaseRemoteConfig/Sources/RCNConfigSettings.m
karenyz: hmm I think NSUserDefaults does save to disk? https://developer.apple.com/documentation/foundation/nsuserdefaults I would be surprised if we reset something like `lastFetchTime` every app restart

qdpham13: Oh I thought you meant the manager, I updated to use userDefaults

FirebaseRemoteConfig/Sources/RCNConfigFetch.m
qdpham13: No there's no need, we already call it to set _templateVersionNumber on line 494

karenyz: 👍

karenyz (approved):
thanks for making the change!!

PR #10166, opened 2022-09-02
[Messaging] macOS keychain auth prompt fix

aashishpatil-g:
This is fix for Messaging to match change made to core here -> https://github.com/google/GoogleUtilities/pull/75

The fix is to avoid auth alerts (popups) on macOS when using Keychain.

Background: go/firebase-macos-keychain-popups

### DIFF
--- a/FirebaseMessaging/Tests/UnitTests/FIRMessagingAuthKeychainTest.m
+++ b/FirebaseMessaging/Tests/UnitTests/FIRMessagingAuthKeychainTest.m
@@ -215,6 +215,11 @@ static NSString *const kBundleID2 = @"com.google.abtesting.dev";
 }
 
 - (void)testQueryCachedKeychainItems {
+// Skip keychain tests on Catalyst and macOS. Tests are skipped because they
+// involve interactions with the keychain that require a provisioning profile.
+// See go/firebase-macos-keychain-popups for more details.
+#if !TARGET_OS_MACCATALYST && !TARGET_OS_OSX
+
   XCTestExpectation *addItemToKeychainExpectation =
       [self expectationWithDescription:@"Test added item should be cached properly"];
   // A wildcard query should return empty data when there's nothing in keychain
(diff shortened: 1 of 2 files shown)


ncooke3 (approved):
LGTM when CI finished and is all green. Feel free to merge then!

No need to update the changelog in this PR. I'm going to do a sweep of all the affected product's changelogs and send a PR for them altogether.

Thanks!

ncooke3:
And pls wait for @chliangGoogle's review before merging as well!

chliangGoogle (approved):
LGTM, make sure the issue is not reproducible after the change.

chliangGoogle:
@aashishpatil-g Can you update the [changelog](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseMessaging/CHANGELOG.md) too?    Use "Unreleased" at the top if you are not sure which version to go out.

ncooke3:
> No need to update the changelog in this PR. I'm going to do a sweep of all the affected product's changelogs and send a PR for them altogether.

@chliangGoogle, I told aashishpatil-g@ to hold off on CHANGELOG update since I'm going to send a PR to update all affected CHANGELOGs after this merges. I want to coordinate the wording with morganchen12@ since we likely will include in our internal docs.

ncooke3:
@aashishpatil-g, the failing CI is expected. The workflow in question is failing because we are testing on [all platforms](https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/messaging.yml#L50). 

One inevitable downside to this solution is that we cannot test macOS test cases that directly use the keychain. The fix here is to `#if` out the offending tests on macOS like what was done here: https://github.com/firebase/firebase-ios-sdk/pull/10146

FirebaseMessaging/Tests/UnitTests/FIRMessagingAuthKeychainTest.m
```diff
@@ -215,6 +215,11 @@ - (void)testKeyChainNoCorruptionWithUniqueService {
 }
 
 - (void)testQueryCachedKeychainItems {
```
ncooke3, line 217: Rather than `#if` out the test implementation, could you regroup the tests so they can be excluded with a single  `#if`:
```objc
// Skip keychain tests on Catalyst and macOS. Tests are skipped because they
// involve interactions with the keychain that require a provisioning profile.
// See go/firebase-macos-keychain-popups for more details.
#if !TARGET_OS_MACCATALYST && !TARGET_OS_OSX

- (void)testQueryCachedKeychainItems { ... }

- (void)testCachedKeychainOverwrite { ... }

- (void)testSetKeychainItemShouldDeleteOldEntry { ... } 

- (void)testQueryAndAddEntry { ... }


#endif  // !TARGET_OS_MACCATALYST && !TARGET_OS_OSX
```

This is for two reasons, one to reduce the number of `#ifs` we use and two so, in the event we are ever debugging test logs, the only tests that show up were actually run.

aashishpatil-g: Made the change in the latest commit - afcd2c7dbc6b980127b716ac159089ef06f48612

paulb777 (approved):

ncooke3 (approved):

Issue #10171, opened 2022-09-05
RingBuffer.swift Reference to Generic type Array Requires arguments in <...>


Speuce:
### Description

After updating my dependencies and re-building my flutter & firebase project, I get the following issue:

`Swift Compiler Error (Xcode): Reference to generic type 'Array' requires arguments in <...>
.../RingBuffer.swift:18:36`

More specifically, this is caused because the line `case outOfBoundsPush(pushIndex: Array.Index, endIndex: Array.Index)` does not specify a type on the Arrays.

The quick-fix, as discussed in [stackoverflow](https://stackoverflow.com/questions/73523004/installing-react-native-firebase-fails-at-compiling-ringbuffer-swift) is to override the file and change the line to `case outOfBoundsPush(pushIndex: Array<Any>.Index, endIndex: Array<Any>.Index)`, specifying a type and resolving the issue. 

My intuition tells me that this is likely a Swift 4 vs Swift 5 problem, even though I had specified to use swift 5.3 in my podfile.
I understand that firebase may no longer be supporting swift 4, regardless, I don't think missing a _single_ type specification should warrant breaking projects that haven't updated.

### Reproducing the issue

Minimum Reproducible Issue: https://swiftfiddle.com/qhguupd5uvgyhfla6th46ucmxm

### Firebase SDK Version

9.4.0

### Xcode Version

12.3

### Installation Method

CocoaPods

### Firebase Product(s)

Authentication, Database, DynamicLinks, Firestore, Storage

### Targeted Platforms

iOS

### Relevant Log Output

```shell
Swift Compiler Error (Xcode): Reference to generic type 'Array' requires arguments in <...>
/Users/redacted/src/ios/Pods/FirebaseCoreInternal/FirebaseCore/Internal/Sources/HeartbeatLogging/RingBuffer.swift:18:36

Swift Compiler Error (Xcode): Reference to generic type 'Array' requires arguments in <...>
/Users/redacted/src/ios/Pods/FirebaseCoreInternal/FirebaseCore/Internal/Sources/HeartbeatLogging/RingBuffer.swift:18:59

Swift Compiler Error (Xcode): Cannot infer contextual base in reference to member 'outOfBoundsPush'
/Users/redacted/src/ios/Pods/FirebaseCoreInternal/FirebaseCore/Internal/Sources/HeartbeatLogging/RingBuffer.swift:22:14
```

### If using Swift Package Manager, the project's Package.resolved

<details>
[...]

PR #10176, opened 2022-09-06
[CoreInternal] Add explicit generics typing for Array.Index usage (2)

ncooke3:
### Context
Follow-up to #10075. Older Swift versions (< 5.3) do not like referencing the
`Array` type without specifying the generic type for the array. This can result
in a buildtime error when building with a Swift version under 5.3.

Prevents #10171

### DIFF
--- a/FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
+++ b/FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
@@ -103,12 +103,14 @@ public final class HeartbeatController {
       )
     }
 
-    // Synchronously gets and returns the stored heartbeats, resetting storage
-    // using the given transform. If the operation throws, assume no
-    // heartbeat(s) were retrieved or set.
-    if let heartbeatsBundle = try? storage.getAndSet(using: resetTransform) {
-      return heartbeatsBundle.makeHeartbeatsPayload()
-    } else {
+    do {
+      // Synchronously gets and returns the stored heartbeats, resetting storage
+      // using the given transform.
+      let heartbeatsBundle = try storage.getAndSet(using: resetTransform)
+      // If no heartbeats bundle was stored, return an empty payload.
+      return heartbeatsBundle?.makeHeartbeatsPayload() ?? HeartbeatsPayload.emptyPayload
+    } catch {
+      // If the operation throws, assume no heartbeat(s) were retrieved or set.
       return HeartbeatsPayload.emptyPayload
     }
   }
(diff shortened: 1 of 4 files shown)


paulb777 (approved):
Please add a `pod lib lint` CI  test that sets the swift-format to 5.3 or 5.0 or 4.2, after verifying that it fails without your fix.

ncooke3:
> Please add a `pod lib lint` CI test that sets the swift-format to 5.3 or 5.0 or 4.2, after verifying that it fails without your fix.

Ugh, I'm locally linting with the Swift version specifier and not reproducing the original build failure.

paulb777:
Hmm, I wonder if that behavior is related to compiler version and not Swift version?

paulb777 (approved):
OK to merge this today ahead of code freeze and work on testing later.

ncooke3:
> Hmm, I wonder if that behavior is related to compiler version and not Swift version?

Agreed, I don't think I understand yet where the problem is.

> OK to merge this today ahead of code freeze and work on testing later.

Thanks, will do 👍

FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
```diff
@@ -103,12 +103,14 @@ public final class HeartbeatController {
       )
     }
 
-    // Synchronously gets and returns the stored heartbeats, resetting storage
-    // using the given transform. If the operation throws, assume no
-    // heartbeat(s) were retrieved or set.
-    if let heartbeatsBundle = try? storage.getAndSet(using: resetTransform) {
```
ncooke3, line 109: This was problematic because `getAndSet(using:)` returns an optional. In the latest Swift versions, the `try?` would do a "double-unwrap": it would convert the throwing result to an optional, unwrap it, and then unwrap the result of the API itself. It's pretty handy and cleans up the unwrapping logic. But, this behavior is not recognized on older Swift versions so i refactored things to be more backwards compatible. The new diff is perhaps easier to understand because the "double-unwrapping" behavior of `try?` may not be too obvious (I think i originally stumbled on it by mistake).

ncooke3:
@paulb777, this is good for another pass. Feel free to merge after approving!

paulb777 (approved):

FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift
paulb777: LGTM

PR #10148, opened 2022-08-27
[MLModelDownloader] Disable keychain-dependent tests on macCatalyst and macOS

ncooke3:
### Context
- Intentionally caused by https://github.com/google/GoogleUtilities/pull/75
  - MLModelDownloader is affected because it depends on Installations (which depends on GoogleUtilities's keychain abstraction)
- While we do lose some coverage on CI, [GoogleUtilities/#75](https://github.com/google/GoogleUtilities/pull/75) makes the macOS keychain behave like the iOS style keychain (which is tested on CI).
- Unblocks some of the CI in #10145

### Future work
- [ ] Open a PR to update MLModelDownloader CHANGELOG with note mentioning new macOS constraint (need to use Keychain Sharing capability)

#no-changelog

PR #10181, opened 2022-09-06
[v9.6.0] Update CHANGELOGs for macOS keychain fix

### LINKED ISSUE #9392: Vidline wants to access key "com.firebase.FiRInstallations.installations.1:19 312161103:ios:513816e35120d976bf153a" in your keychain.
How to stop this popup from appearin !!!

34241645540838_ pic
g

ncooke3:
### Context
Completely addressing #9392 required several PRs to eliminate the permission pop-ups on macOS:
- https://github.com/google/GoogleUtilities/pull/75
- https://github.com/firebase/firebase-ios-sdk/pull/10148
- https://github.com/firebase/firebase-ios-sdk/pull/10146
- https://github.com/firebase/firebase-ios-sdk/pull/10166

The fix involved making the macOS keychain behave like the iOS-style keychain. The iOS-style keychain is sandboxed per application by default. This does not happen by default on macOS and requires using the `kSecUseDataProtectionKeychain` attribute when interacting with the Keychain. The `kSecUseDataProtectionKeychain`  attribute tells macOS to treat a given keychain item like it is part of a keychain access group that the app knows and trusts. To synthesize and officiate this access group, it is required that macOS apps enable the `Keychain Sharing` capability for targets that depend on Firebase.

Fixes #9392

### DIFF
--- a/FirebaseCore/CHANGELOG.md
+++ b/FirebaseCore/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Firebase 9.6.0
+- [fixed] Mac apps using Firebase products that store SDK data in the keychain
+  will no longer prompt the user for permission to access the keychain. This
+  requires that Mac apps using Firebase be signed with a provisioning profile
+  that has the Keychain Sharing capability enabled. (#9392)
 - [fixed] Update dependency specification for GTMSessionFetcher to allow all 2.x versions. (#10131)
 
 # Firebase 9.5.0
--- a/FirebaseAppCheck/CHANGELOG.md
+++ b/FirebaseAppCheck/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 9.6.0
+- [fixed] Mac apps using this product will no longer prompt the user for
+  permission when storing/retrieving SDK data in the keychain. Using this
+  product on macOS now requires that the target be signed with a
+  provisioning profile that has the Keychain Sharing capability
+  enabled. (#9392)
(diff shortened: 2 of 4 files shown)


FirebaseCore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Firebase 9.6.0
```
ncooke3, line 1: I added this central CHANGELOG entry and ones on a per-product basis. I realize that the dependency on Installations means I could technically add an entry for every product that depends on Installations. Should I do that or this is fine or just keep the Firebase-wide entry in FirebaseCore? @paulb777 and @morganchen12

paulb777: My preference would be to only have this one and not duplicate for the products, but I'll defer to @morganchen12 to decide.

paulb777 (approved):

FirebaseCore/CHANGELOG.md
morganchen12: +1, let's not duplicate. It is technically a pan-Firebase change, so if it regresses in any product, it's considered a bug.

morganchen12 (approved):

FirebaseCore/CHANGELOG.md
ncooke3: Thanks y'all. SGTM

PR #10188, opened 2022-09-08
Realtime rc no change

qdpham13:
Fix NO_CHANGE issue by removing the Etag whenever the template version is the default value (0).
https://docs.google.com/document/d/14C1z60RlKwri12pnton8noDlITH45sO6OQxCq7jSTbs/edit#heading=h.xgjl2srtytjt

### DIFF
--- a/FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
+++ b/FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
@@ -49,7 +49,7 @@ typedef void (^RCNConfigFetcherCompletion)(NSData *data, NSURLResponse *response
 /// @param completionHandler   Callback handler.
 - (void)fetchConfigWithExpirationDuration:(NSTimeInterval)expirationDuration
                         completionHandler:(FIRRemoteConfigFetchCompletion)completionHandler
-                        addEtagToHeader: (bool)addEtagToHeader;
+             excludeEtagHeaderForRealtime:(bool)excludeEtagHeaderForRealtime;
 
 /// Add the ability to update NSURLSession's timeout after a session has already been created.
 - (void)recreateNetworkSession;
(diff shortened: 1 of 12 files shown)


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -451,7 +453,8 @@ - (void)fetchLatestConfig:(NSInteger)remainingAttempts targetVersion:(NSInteger)
                                         targetVersion:targetVersion];
                               }
                             }
-                          }];
+                          }
+               excludeEtagHeaderForRealtime:excludeEtagHeaderForRealtime];
```
karenyz, line 457: nit: can we update the method signature to have `excludeEtagHeaderForRealtime:` before `completionHandler:`?

FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
```diff
@@ -48,7 +48,8 @@ typedef void (^RCNConfigFetcherCompletion)(NSData *data, NSURLResponse *response
 /// @param expirationDuration  Expiration duration, in seconds.
 /// @param completionHandler   Callback handler.
```
karenyz, line 49: we should briefly explain here why `excludeEtagHeaderForRealtime` is needed

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -420,6 +420,8 @@ - (void)fetchLatestConfig:(NSInteger)remainingAttempts targetVersion:(NSInteger)
         [strongSelf autoFetch:remainingAttempts - 1 targetVersion:targetVersion];
       }
     } else {
+      bool excludeEtagHeaderForRealtime =
```
karenyz, line 423: maybe also worth adding a comment here explaining when template version would be 0

qdpham13: Done

FirebaseRemoteConfig/Sources/Private/RCNConfigFetch.h
qdpham13: Done

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
qdpham13: Think it would make more sense to just add it at the end for the method overall so that current calls to this method don't change too much.

karenyz (approved):

PR #10202, opened 2022-09-10
GoogleAppMeasurement versioning issue

Matthew-G-Jennings:
Version 9.6.0 is not tagged, it should be 9.3.0 or issues with CI on firebase using projects arise. Alternatively v9.6.0 on GoogleAppMeasurement could be tagged

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -150,7 +150,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Note that CI changes the version to the head of main for CI.
       // See scripts/setup_spm_tests.sh.
-      .exact("9.6.0")
+      .exact("9.3.0")
     ),
     .package(
       name: "GoogleDataTransport",


Matthew-G-Jennings:
"Dependencies could not be resolved because no versions of 'GoogleAppMeasurement' match the requirement 9.6.0 and root depends on 'GoogleAppMeasurement' 9.6.0."

paulb777:
Thanks for the PR, but ...

9.6.0 is currently being tested and not yet released. 

We work around SPM in CI with this: https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/setup_spm_tests.sh#L18

In general, we don't recommend external developers relying on the `master` branch.

I'll close, but feel free to continue the discussion.

Matthew-G-Jennings:
Thanks for your reply. I had a look at that, are you sure it's working? 

As of the latest FB release our CI was broken giving the message "Dependencies could not be resolved because no versions of 'GoogleAppMeasurement' match the requirement 9.6.0 and root depends on 'GoogleAppMeasurement' 9.6.0."

It is entirely possible this is an issue with our CI (probably more likely) but it is pretty close to default github action stuff. Using specific version of firebase 9.4.1 works fine

Matthew-G-Jennings:
image

Matthew-G-Jennings:
I see what's going on, this is my mistake, had it set to use master/main when it should have been latest release. Apologies for the time wasting.

PR #10236, opened 2022-09-20
FDL: Mark FIRDynamicLink's init as NS_UNAVAILABLE

### LINKED ISSUE #10000: DynamicLink initializer shouldn't be available from Swift
There's no `init` marked as `NS_UNAVAILABLE`, so a bare initializer is currently available in Swift but shouldn't be. 

https://github.com/firebase/firebase-ios-sdk/blob/8467858fc4cee858511380ac0f9ea5a17d007b5d/FirebaseDynamicLinks/Sources/Public/FirebaseDynamicLinks/FIRDynamicLink.h#L55-L86

We'll need to determine the right strategy for this: it's a breaking change to remove it, but it doesn't work as is. Introducing an initializer that's deprecated will introduce the API for ObjC users, which we don't want either.

andrewheard:
DynamicLink's initializer was not intended to be publicly available. Marking it as NS_UNAVAILABLE hides it from both ObjC and Swift callers; resolves #10000.

### DIFF
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 10.0.0
+- [removed] Removed bare initializer from `DynamicLink`. (#10236)
+
 # 9.0.0
 - [fixed] Fixed async/await crash when retrieving a dynamic link from a universal link fails. (#9612)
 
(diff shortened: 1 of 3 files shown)


FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 10.0.0
+- [removed] Removed bare initializer from `DynamicLink`. (#10236)
```
paulb777, line 2: Should be the issue number, not the PR number.

paulb777 (approved):

FirebaseDynamicLinks/CHANGELOG.md
andrewheard: Fixed. Thanks!

PR #10239, opened 2022-09-21
Address 9963 review feedback

paulb777:
- Use `callback?` instead of `if let callback ...`
- Better StorageDownloadTask `enqueue` signature
- Restore non-optional `bucket` property in StorageMetadata

Initializing an unspecified `bucket` property as `""` instead of `nil` when creating `StorageMetadata` with a dictionary was an inadvertent Firebase 9 change from Firebase 8. This PR keeps the Firebase 9 behavior onto Firebase 10.


ncooke3 (approved):
Thanks for investigating the bucket name API issue.

PR #9963, opened 2022-06-29
Migrate Storage implementation from Objective-C to Swift

paulb777:
`storage-swift-phase2` is a long-lived branch for migrating the Storage implementation from Objective-C to Swift.

Almost all of the changes are a straight port from Objective C to Swift. One exception is the Authorizer code implements a new GTMSessionFetcher protocol to avoid complexity around using `NSInvocation` from Swift required with the old protocol.

Key commits:

- Convert Storage Task implementation to Swift (#9880)
- Convert StorageMetadata to Swift (#9979)
- Storage and StorageReference to Swift (#10077)
- StorageAuthorizer to Swift (#10117)
- Remove FirebaseStorageInternal (#10118)
- Storage Changelog for Firebase 10 (#10215)
- Add new throwing reference(for url:URL) API (#10216)

PR #10245, opened 2022-09-22
Fix inadvertent API change

paulb777:
Address post-merge #9963 comment

#no-changelog

### DIFF
--- a/FirebaseStorage/Sources/StorageReference.swift
+++ b/FirebaseStorage/Sources/StorageReference.swift
@@ -410,15 +410,15 @@ import Foundation
    *       the current `StorageReference`.
    */
   @objc(listAllWithCompletion:)
-  open func listAll(completion: @escaping ((_: StorageListResult?, _: NSError?) -> Void)) {
+  open func listAll(completion: @escaping ((_: StorageListResult?, _: Error?) -> Void)) {
     let fetcherService = storage.fetcherServiceForApp
     var prefixes = [StorageReference]()
     var items = [StorageReference]()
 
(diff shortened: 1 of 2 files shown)


FirebaseStorage/Sources/StorageReference.swift
```diff
@@ -410,15 +410,15 @@ import Foundation
    *       the current `StorageReference`.
    */
   @objc(listAllWithCompletion:)
-  open func listAll(completion: @escaping ((_: StorageListResult?, _: NSError?) -> Void)) {
+  open func listAll(completion: @escaping ((_: StorageListResult?, _: Error?) -> Void)) {
```
ncooke3, line 413: I think to have caught this, we would have needed explicit typing in the API tests: 

https://github.com/firebase/firebase-ios-sdk/blob/3d51280713332af3f023abc7e305541871c1e005/FirebaseStorage/Tests/Unit/StorageAPITests.swift#L86-L87

```swift
ref.listAll { listResult: StorageListResult?, error: Error? in 
 } 
```

ncooke3 (approved):
LGTM. Based on CI, it doesn't look like any tests were affected.

FirebaseStorage/Sources/StorageReference.swift
ncooke3: We'd have to do this for all Swift closures to get complete coverage.

paulb777: Good point! Done for the StorageAPI tests in this PR.

PR #10249, opened 2022-09-22
App google domain support

eldhosembabu:
Adding app.google (1p domain) support in FDL SDK which was missing.

Internal reference bug for Android: b/235838240. For iOS, it was not an issue since we had a workaround telling developers to add the app.google domain to the list of custom domains in plist. Now with this change, they don't need to do it anymore.


FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 10.0.1
+- [fixed] Added app.google (1p domain) support in FDL SDK which was missing.
```
paulb777, line 2: Include in 10.0.0 which won't code freeze until next week.

paulb777 (approved):
LGTM on green CI and changelog fix

FirebaseDynamicLinks/CHANGELOG.md

PR #10258, opened 2022-09-24
Firestore: Re-write API docs for COUNT API

dconeybe:
The wording is copied from https://github.com/firebase/firebase-android-sdk/pull/4143

#no-changelog


wu-hui (approved):
I don't have good suggestions, but I think some of the comments can be more concise. Maybe we should run this by a tech writer too.

dconeybe:
> I don't have good suggestions, but I think some of the comments can be more concise. Maybe we should run this by a tech writer too.

Thanks for the feedback. I do tend to be kind of wordy. I'll engage a tech writer for review, but I'll merge this for now.

PR #10261, opened 2022-09-26
Add isEqual and hash for aggregate classes

wu-hui:


Firestore/Example/Tests/Integration/API/FIRCountTests.mm
```diff
@@ -28,6 +28,38 @@ @interface FIRCountTests : FSTIntegrationTestCase
 
 @implementation FIRCountTests
 
+- (void)testEquals {
```
dconeybe, line 31: nit: Name this method `testAggregateQueryEquals`

Firestore/Example/Tests/Integration/API/FIRCountTests.mm
```diff
@@ -77,6 +109,41 @@ - (void)testCanRunCountWithOrderBys {
...
+      [self readSnapshotForAggregate:[[testCollection queryWhereField:@"k" isEqualTo:@"b"] count]];
+  FIRAggregateQuerySnapshot* snapshot1Same =
+      [self readSnapshotForAggregate:[[testCollection queryWhereField:@"k" isEqualTo:@"b"] count]];
+
+  FIRAggregateQuerySnapshot* snapshot2 =
+      [self readSnapshotForAggregate:[[testCollection queryWhereField:@"k" isEqualTo:@"a"] count]];
+  [self writeDocumentRef:[testCollection documentWithPath:@"d"] data:@{@"k" : @"a"}];
+  FIRAggregateQuerySnapshot* snapshot2Different =
+      [self readSnapshotForAggregate:[[testCollection queryWhereField:@"k" isEqualTo:@"a"] count]];
+
```
dconeybe, line 133: Add a test case for two different `FIRAggregateQuerySnapshot` objects with the same result set but different queries. So far you've tested snapshots with the same AggregateQuery and same results, and the same AggregateQuery with different results. That third one is the only one missing to fill out the test coverage.

dconeybe (changes requested):
Just a couple of minor comments. Otherwise, LGTM. Thanks for filling out this implementation!

Firestore/Example/Tests/Integration/API/FIRCountTests.mm
wu-hui: Done.

dconeybe (approved):
LGTM

PR #10259, opened 2022-09-24
[Infra] Remove CoreDiagnostics library and references

ncooke3:
### Context
- Remove CoreDiagnostics as it is no longer needed.
- There are no references when grepping the repo for  'CoreDiagnostics'.
   ```shell
   ➜  firebase-ios-sdk git:(nc/remove-cd) 
   ‣ git grep 'CoreDiagnostics'  
   ➜  firebase-ios-sdk git:(nc/remove-cd) 
   ```

#no-changelog

Fixes #10192

PR #10262, opened 2022-09-26
[Cleanup] Follow-up to #10259

ncooke3:
# Context
- Cleanup-oriented follow-up to #10259
- Removed `FIRHeartbeatInfo.[hm]`
- Removed `FIRHeartbeatInfoTest.m`
- Moved and renamed `FIRHeartbeatInfoCode` enum to `FIRDailyHeartbeatCode`. This is because the enum only supports daily heartbeats in an effort to be compatible with platform logging V1 clients.
- No more references to `FIRHeartbeatInfo`:
   ```shell
   ➜  firebase-ios-sdk git:(nc/heartbeat-cleanup) 
   ‣ git grep FIRHeartbeatInfo
   ➜  firebase-ios-sdk git:(nc/heartbeat-cleanup) 
   ```
- Revisited a TODO and decided to take no action after experimenting with it.

#no-changelog


paulb777 (approved):
LGTM on green

PR #10266, opened 2022-09-26
add a readme for analytics swift sdk and add analytics view modifier in rc swift test app

chliangGoogle:
Testing out a analytics view modifier feature with RC property wrapper. Updated the app and documentation.

#no-changelog


FirebaseRemoteConfigSwift/Apps/SwiftUISample/SwiftUISample/ContentView.swift
```diff
@@ -35,10 +36,9 @@ struct ContentView: View {
   @RemoteConfigProperty(
     key: "recipe", fallback: Recipe(recipeName: "banana bread", cookTime: 40, notes: "yum!")
   )
-  var recipe: Recipe
 
-  @State private var realtimeSwitch = false
-  var realtimeToggle: Bool
+  var recipe: Recipe
+  @State private var isChecked = false
```
karenyz, line 41: does `isChecked` control something in the UI? looks like the button image only depends on `toggleValue`

chliangGoogle: Good catch! I made it more practical case where isCheck guard the value of the toggle and a toggleStyleSquare config value controls the checkbox style to be square or circle.

karenyz (approved):

PR #10268, opened 2022-09-27
[AppCheck][WIP] Add isSupported to providers

andrewheard:
Added an `isSupported` method to the `FIRAppCheckProvider` protocol. This would allow developers to easily check if a provider is supported on the device/OS and to fallback to other providers if it's not available.

[skip ci] - just an idea, not ready for review


paulb777:
Closing, since code is no longer in this repo

PR #10273, opened 2022-09-27
[Firestore] Remove encoding/decoding strategy type aliases

andrewheard:
Removed the `Firestore.Encoder` and `Firestore.Decoder` encoding/decoding strategy type aliases and directly used the types from `FirebaseDataEncoder` and `FirebaseDataDecoder`.

Note: This requires the `TimestampEncoding/DecodingStrategy` extension to be added to `FirebaseDataEncoder/Decoder` instead of `Firestore.Encoder` and `Firestore.Decoder`. This will make the `timestamp` strategy visible to all users of the `FirebaseDataEncoder/Decoder` (that have FirestoreSwift installed), even though `Timestamp` is a Firestore-specific type.


paulb777 (approved):
LGTM on green

ncooke3 (approved):

PR #10246, opened 2022-09-22
Public count

wu-hui:

PR #10278, opened 2022-09-28
[Firestore] Update CHANGELOG for new Count API (#10246)

ncooke3:
### Context
Meant to roll this into the other Firestore PRs. The count API PR has the changelog entry under `# Unreleased`; moving it under `# 10.0.0`.


ncooke3:
I used [skip ci] on the last few commits of https://github.com/firebase/firebase-ios-sdk/pull/10167. While I think that was safe, CI on this PR ultimately confirmed that Firestone's latest HEAD is ok.

paulb777 (approved):

PR #10282, opened 2022-09-28
Update Firestore sample for iOS 16

peterfriese:
Just some general housekeeping on the Firestore sample:

- use new Navigation API
- use single app icon
- improve menu structure
- fix security rules

It might also be worth adding some demos to show how to use Firestore's Codable support


paulb777:

PR #10285, opened 2022-09-28
Initial commit of Firebase Sessions SDK

samedson:
Paired with: @jeremyjiang-dev 

# Notes
 - We haven't setup any SDKs with the Sessions SDK as a dependency, so this should not go out to users (unless they specify it directly)
 - We've only setup CocoaPods so far (for development), not SPM

# Questions
 - We were using [FunctionsComponent.swift](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseFunctions/Sources/Internal/FunctionsComponent.swift#L24-L33) as a reference. We see there is a protocol and initializer there. Our protocol isn't being called, and I'm not sure where their `functions(...` protocol method is being called
   - Is the protocol needed?
   - It looks like it's needed as something to pass into the `Component` constructor

#no-changelog
#sessions

### DIFF
--- a/FirebaseCore/Sources/FIRApp.m
+++ b/FirebaseCore/Sources/FIRApp.m
@@ -822,6 +822,7 @@ static FIRApp *sDefaultApp;
   // Dictionary of class names that conform to `FIRLibrary` and their user agents. These should only
   // be SDKs that are written in Swift but still visible to ObjC.
   NSDictionary<NSString *, NSString *> *swiftComponents = @{
+    @"FIRSessions" : @"fire-sessions",
     @"FIRFunctionsComponent" : @"fire-fun",
     @"FIRStorageComponent" : @"fire-str",
   };
--- a/FirebaseSessions.podspec
+++ b/FirebaseSessions.podspec
@@ -1,0 +1,45 @@
+Pod::Spec.new do |s|
+  s.name             = 'FirebaseSessions'
+  s.version          = '10.0.0'
+  s.summary          = 'Firebase Sessions'
+
+  s.description      = <<-DESC
+  Firebase Sessions.
+                       DESC
+
+  s.homepage         = 'https://firebase.google.com'
+  s.license          = { :type => 'Apache-2.0', :file => 'LICENSE' }
+  s.authors          = 'Google, Inc.'
+
+  s.source           = {
+    :git => 'https://github.com/firebase/firebase-ios-sdk.git',
+    :tag => 'CocoaPods-' + s.version.to_s
+  }
+  s.social_media_url = 'https://twitter.com/Firebase'
+
+  ios_deployment_target = '11.0'
+  osx_deployment_target = '10.13'
+  tvos_deployment_target = '12.0'
+  watchos_deployment_target = '6.0'
+
+  s.swift_version = '5.3'
+
+  s.ios.deployment_target = ios_deployment_target
+  s.osx.deployment_target = osx_deployment_target
+  s.tvos.deployment_target = tvos_deployment_target
+  s.watchos.deployment_target = watchos_deployment_target
+
+  s.cocoapods_version = '>= 1.4.0'
+  s.prefix_header_file = false
+
+  s.source_files = "FirebaseSessions/Sources/**/*.swift"
+
+  s.dependency 'FirebaseCore', '~> 10.0'
+  s.dependency 'FirebaseCoreExtension', '~> 10.0'
+  s.dependency 'FirebaseInstallations', '~> 10.0'
+
+  s.pod_target_xcconfig = {
+    'GCC_C_LANGUAGE_STANDARD' => 'c99',
+    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
+  }
+end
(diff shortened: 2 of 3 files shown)


FirebaseSessions.podspec
```diff
@@ -0,0 +1,50 @@
...
+  s.dependency 'FirebaseCore', '~> 10.0'
+  s.dependency 'FirebaseCoreExtension', '~> 10.0'
+  s.dependency 'FirebaseInstallations', '~> 10.0'
+
+  preprocessor_definitions = ''
+  s.pod_target_xcconfig = {
+    'GCC_C_LANGUAGE_STANDARD' => 'c99',
+    'GCC_PREPROCESSOR_DEFINITIONS' => preprocessor_definitions,
+    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
+  }
```
ncooke3, line 49: Optional, but could trim this down to: 
```suggestion
  s.pod_target_xcconfig = {
    'GCC_C_LANGUAGE_STANDARD' => 'c99',
    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
  }
```

ncooke3 (commented):
Still working through the DD... but here's a first pass!

FirebaseCore/Sources/FIRApp.m
```diff
@@ -822,6 +822,7 @@ + (void)registerSwiftComponents {
   // Dictionary of class names that conform to `FIRLibrary` and their user agents. These should only
   // be SDKs that are written in Swift but still visible to ObjC.
   NSDictionary<NSString *, NSString *> *swiftComponents = @{
+    @"FIRSessions" : @"fire-sessions",
```
ncooke3, line 825: @samedson, this answers the question in the PR description, right?

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -0,0 +1,55 @@
...
+
+  required init(app: FirebaseApp) {
+    self.app = app
+  }
+
+  // MARK: - Library conformance
+
+  static func componentsToRegister() -> [Component] {
+    return [Component(SessionsProvider.self,
+                      instantiationTiming: .eagerInDefaultApp,
```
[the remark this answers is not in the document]
samedson: That's a good question - so I had assumed it would try and start it up again for the 2nd app if it was alwaysEager - is that the case, or is that controlled by something else? We only want the Sessions SDK to startup for the default app (like Crashlytics and Performance do).

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -0,0 +1,55 @@
...
+        isCacheable.pointee = true
+        return self.init(app: app)
+      }]
+  }
+
+  // MARK: - SessionsProvider conformance
+
+  func sessions(for app: FirebaseApp) -> Sessions {
+    return self
+  }
```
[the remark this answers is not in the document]
samedson: Yes this is temporary

FirebaseCore/Sources/FIRApp.m
samedson: It's kindof part of the question. I have this SessionProvider but it doesn't seem to do anything besides satisfy the constraint that the Component must be passed a protocol. We use the init method in the Sessions class to start the SDK, not the `sessions(for app)` method that's defined by the protocol.

FirebaseSessions/Sources/FirebaseSessions.swift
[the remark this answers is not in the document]
samedson: See my response in FIRApp.m

samedson:
Thanks for the suggestions @ncooke3 ! Added my responses.

ncooke3:
CI should be fixed by a rebase on top of `master`.

FirebaseSessions/Sources/FirebaseSessions.swift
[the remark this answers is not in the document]
ncooke3: I spoke with Ryan about this and I think since the long term design is to make this available to other SDKs, we could use `.alwaysEager` and then, for now, check in the block whether to instantiate based on if it's a default app. See below suggestion.

ncooke3:
@samedson, sorry– I thought the rebase would surely fix CI. It's not a blocker though since the failure is unrelated.

samedson:
Sounds good - thanks for the advice! I made one change to include a reference to the Google App ID because we'll need that

FirebaseSessions.podspec
```diff
@@ -0,0 +1,45 @@
+Pod::Spec.new do |s|
+  s.name             = 'FirebaseSessions'
+  s.version          = '10.0.0'
+  s.summary          = 'Firebase Sessions'
+
+  s.description      = <<-DESC
+  Firebase Sessions.
```
paulb777, line 7: Should have a comment about internal use only.  See https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCoreInternal.podspec#L7

paulb777 (commented):
While starting with CocoaPods is ok, it might be better to do early development in Swift Package Manager, since it's the future.

FirebaseSessions.podspec
samedson: Sounds good 👍

samedson: Yeah re: CocoaPods - I totally did CocoaPods first cause it's what I'm comfortable with. I'll spend some time setting up SPM

paulb777 (approved):

PR #10281, opened 2022-09-28
[Do Not Submit] Run a global tap after code freeze

ncooke3:
The nightlies kicked off before Firebase 10 PR were all merged. Triggering this PR as a global tap. 

I'm confident everything is alright as-is but this won't hurt.

#no-changelog

PR #10286, opened 2022-09-28
[Firebase 10] Revamp FIRLogger for Swift use

ncooke3:
See https://github.com/firebase/firebase-ios-sdk/pull/10281#discussion_r982503642 for context!

#no-changelog


Gemfile
```diff
@@ -15,3 +15,5 @@ source 'https://rubygems.org'
 gem 'cocoapods', '1.11.3'
 gem 'cocoapods-generate', '2.0.1'
 gem 'danger', '8.4.5'
+
+# Global TAP
```
ncooke3, line 19: Revert before merging.

FirebaseCore/Extension/FIRLogger.h
```diff
@@ -147,26 +147,40 @@ extern void FIRLogWarningSwift(FIRLoggerService service, NSString *messageCode,
...
+///
+/// - Parameters:
+///   - level: The log level to use (defined by `FirebaseLoggerLevel` enum values).
+///   - service: The service name of type `FirebaseLoggerService`.
+///   - code: The mesage code. Starting with "I-" which means iOS, followed by a capitalized
+///   three-character service identifier and a six digit integer message ID that is unique within
+///   the service. An example of the message code is @"I-COR000001".
+///   - message: Formatted string to be used as the log's message.
+///   - args: Arguments list obtained from calling `va_start`, used when message is a format string.
 + (void)logWithLevel:(FIRLoggerLevel)level
```
ncooke3, line 179: FYI: I couldn't find any usage of this API

FirebaseCore/Sources/FIRLogger.m
```diff
@@ -157,27 +157,44 @@ void FIRLogBasic(FIRLoggerLevel level,
...
 + (void)logWithLevel:(FIRLoggerLevel)level
-         withService:(FIRLoggerService)service
-            withCode:(NSString *)messageCode
-         withMessage:(NSString *)message
-            withArgs:(va_list)args {
-  FIRLogBasic(level, service, messageCode, message, args);
+             service:(FIRLoggerService)service
+                code:(NSString *)code
+             message:(NSString *)message {
+  if (level == FIRLoggerLevelMin) {
```
ncooke3, line 178: This implementation could be cleaned up if it just called `FIRLogBasic` but that would use the GULs logger implementation.

The `FIRLog*Swift` functions called into the macro function defined in FIRLogger.m.

ncooke3: Is there a reason we use the GUL implementation vs: https://github.com/firebase/firebase-ios-sdk/blob/e236acacafffa30de4537c9a7a3028e16f29b9a5/FirebaseCore/Sources/FIRLogger.m#L138-L158

FirebaseCore/Sources/FIRLogger.m
```diff
@@ -157,27 +157,44 @@ void FIRLogBasic(FIRLoggerLevel level,
...
+    case FIRLoggerLevelWarning:
+      return FIRLogWarning(service, code, @"%@", message);
+    case FIRLoggerLevelNotice:
+      return FIRLogNotice(service, code, @"%@", message);
+    case FIRLoggerLevelInfo:
+      return FIRLogInfo(service, code, @"%@", message);
+    case FIRLoggerLevelDebug:
+      return FIRLogDebug(service, code, @"%@", message);
+    default:
+      return FIRLogDebug(service, code, @"%@", message);
```
morganchen12, line 196: nit: Should an unknown value log as `Error` instead of `Debug`?

morganchen12 (commented):
LGTM, but please wait for other approvers before merging.

FirebaseCore/Extension/FIRLogger.h
andrewheard: The NSString/va_list approach isn't as useful in Swift either since string interpolation is concise.

paulb777: Fine to remove, if also confirmed with an internal global tap.

FirebaseCore/Sources/FIRLogger.m
paulb777: Maybe because FIRLogger was intended to be a wrapper over GULLogger?

paulb777 (approved):

FirebaseCore/Sources/FIRLogger.m
andrewheard: It looks like the generated macros end up calling `FIRLogBasic` anyway too.

ncooke3: @paulb777, so you are OK with the current implementation or you think I should remove the above macro and just log through GULs?

FirebaseCore/Sources/FIRLogger.m
ncooke3: Good idea. I will let CI finish and push a fix.

Firestore/Swift/Source/Codable/DocumentID.swift
```diff
@@ -118,17 +118,18 @@ public struct DocumentID<Value: DocumentIDWrappable & Codable>:
...
       value = newValue
     }
   }
 
-  private func logWarning(for value: Value) {
-    FIRLogWarningSwift(
-      "[FirebaseFirestoreSwift]",
-      "I-FST000002",
-      """
+  private func logWarning(with value: Value) {
```
andrewheard, line 127: nit: I don't feel strongly about it but potentially something like `logIgnoredValueWarning` since that function is for a specific logging case.

andrewheard (approved):
This LGTM as-is. I'm wondering if it would be best not to worry too much about a cleanup at this point since we might want to look into using [Logger](https://developer.apple.com/documentation/os/logger) when running on iOS 14+ or some other Swifty approach.

FirebaseCore/Sources/FIRLogger.m
ncooke3: > It looks like the generated macros end up calling `FIRLogBasic` anyway too.

@andrewheard, I see it now. I'm going to update my new API's implementation to call directly to `FIRLogBasic`

Firestore/Swift/Source/Codable/DocumentID.swift
ncooke3: Thanks, I prefer that. Done in 4feb8fc071d23a004f55d9ad619022d071595e33

FirebaseCore/Sources/FIRLogger.m
ncooke3: Thanks, I prefer that. Done in 4feb8fc071d23a004f55d9ad619022d071595e33

FirebaseCore/Extension/FIRLogger.h
ncooke3: Ehh global TAP is not looking good and I dont want to spend time looking into it now. I'm going to revert some the API breaks.

FirebaseCore/Sources/FIRLogger.m
ncooke3: I changed the implementation so this is actually obsolete

ncooke3:
[`auth / pod-lib-lint (watchos)` is failing ](https://github.com/firebase/firebase-ios-sdk/actions/runs/3146851345/jobs/5115778295) but I have no idea why. Investigating.....

ncooke3:
> [`auth / pod-lib-lint (watchos)` is failing ](https://github.com/firebase/firebase-ios-sdk/actions/runs/3146851345/jobs/5115778295) but I have no idea why. Investigating.....

Thanks Paul for the tip. The issue is like what was seen in https://github.com/firebase/firebase-ios-sdk/pull/9732. I've fixed the issue locally. I think I will wait for CI to finish to avoid the churn of a restart. 

Assuming no other real failures occur (I expect a flake or two) in the current CI jobs, I will push my fix with a `skip ci` flag after they are done. I have tested my local fix and it looks gtg.

Gemfile
ncooke3: Done.

ncooke3:
The **latest** rc0 zip is building at https://github.com/firebase/firebase-ios-sdk/actions/runs/3148200619

I merged this before the nightlies started but can diff it with the nightly zip to make sure everything is as expected

PR #10288, opened 2022-09-29
Fix sign in alert shown inconsistently when triggered in viewDidAppear

tejasd:
This fixes the issue (#8096) where a scene may not be considered foreground in viewDidAppear.

There can still be issues for apps that have multiple scenes, and the right approach for that would involve using `UISceneDidActivateNotification`.

An example is https://github.com/firebase/firebase-ios-sdk/pull/9169.

### DIFF
--- a/FirebaseAppDistribution/CHANGELOG.md
+++ b/FirebaseAppDistribution/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 9.0.1
-- [fixed] Fixed inconsistent sign in prompts in single scenes (#8096).
+- [fixed] Fixed inconsistent sign in prompts in single scene apps (#8096).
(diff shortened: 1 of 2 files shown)


lfkellogg (approved):

FirebaseAppDistribution/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 9.0.1
```
ncooke3, line 1: This will actually release in Firebase 10.
```suggestion
# 10.1.0
```

ncooke3: @paulb777, or should we do a patch release for this?

PR #10296, opened 2022-10-01
Persist FIRPhoneMultiFactorInfo (instead of FIRMultiFactorInfo) so that full phone number is always cached

renkelvin:
Fix a bug where multi factor phone number returns NULL.


prameshj (approved):

paulb777 (commented):
Fix tests and if client visible, add a Changelog

renkelvin:
Rerun fixed the CI failures. See the change log changes here: https://github.com/firebase/firebase-ios-sdk/pull/10303

paulb777 (approved):

PR #10297, opened 2022-10-03
[Sessions] Setup Sessions SDK Class Structure and Tests

samedson:
This establishes the patterns we will follow to structure and test the Sessions SDK.

To make our Classes testable:
 - Classes should only interact with other Classes via Protocols
 - Classes should only get instances to other Classes via arguments passed in to their constructor
 - All top-level Classes should be instantiated at the start of the app in FirebaseSessions.swift or some other helper class

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/Coordinator.swift
+++ b/FirebaseSessions/Sources/Coordinator.swift
@@ -24,7 +24,7 @@ class Coordinator {
     self.identifiers = identifiers
   }
 
-  public func runMain() {
+  func runMain() {
     // TODO:
   }
 }
--- a/FirebaseSessions/Sources/Coordinator.swift
+++ b/FirebaseSessions/Sources/Coordinator.swift
@@ -1,0 +1,30 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+///
+/// Main is responsible for coordinating the systems in this SDK involved with sending a Session Start event.
+///
+class Coordinator {
+  private let identifiers: IdentifierProvider
+
+  init(identifiers: IdentifierProvider) {
+    self.identifiers = identifiers
+  }
+
+  public func runMain() {
+    // TODO:
+  }
+}
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 import Foundation
+
 import FirebaseCore
+import FirebaseInstallations
 
 // Avoids exposing internal FirebaseCore APIs to Swift users.
 @_implementationOnly import FirebaseCoreExtension
@@ -23,16 +25,43 @@ protocol SessionsProvider {
   @objc static func sessions() -> Void
 }
 
-@objc(FIRSessions) class Sessions: NSObject, Library, SessionsProvider {
+@objc(FIRSessions) final class Sessions: NSObject, Library, SessionsProvider {
   // MARK: - Private Variables
 
-  /// The app associated with all sessions.
-  private let googleAppID: String
+  /// The Firebase App ID associated with Sessions.
+  private let appID: String
+
+  /// Top-level Classes in the Sessions SDK
+  private let coordinator: Coordinator
+  private let initiator: Initiator
+  private let identifiers: Identifiers
 
   // MARK: - Initializers
 
-  required init(app: FirebaseApp) {
-    googleAppID = app.options.googleAppID
+  required convenience init(appID: String, installationIDProvider: InstallationIDProvider) {
+    let identifiers = Identifiers(installationIDProvider: installationIDProvider)
+    let coordinator = Coordinator(identifiers: identifiers)
+    let initiator = Initiator()
+
+    self.init(appID: appID,
+              identifiers: identifiers,
+              coordinator: coordinator,
+              initiator: initiator)
+  }
+
+  init(appID: String, identifiers: Identifiers, coordinator: Coordinator, initiator: Initiator) {
+    self.appID = appID
+
+    self.identifiers = identifiers
+    self.coordinator = coordinator
+    self.initiator = initiator
+
+    super.init()
+
+    self.initiator.beginListening {
+      self.identifiers.generateNewSessionID()
+      self.coordinator.runMain()
+    }
   }
 
   // MARK: - Library conformance
@@ -44,7 +73,8 @@ protocol SessionsProvider {
         // Sessions SDK only works for the default app
         guard let app = container.app, app.isDefaultApp else { return nil }
         isCacheable.pointee = true
-        return self.init(app: app)
+        let installations = Installations.installations(app: app)
+        return self.init(appID: app.options.googleAppID, installationIDProvider: installations)
       }]
   }
 
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -32,16 +32,16 @@ protocol SessionsProvider {
   private let appID: String
 
   /// Top-level Classes in the Sessions SDK
-  private let coordinator: Coordinator
-  private let initiator: Initiator
+  private let coordinator: SessionCoordinator
+  private let initiator: SessionInitiator
   private let identifiers: Identifiers
 
   // MARK: - Initializers
 
   required convenience init(appID: String, installations: InstallationsProtocol) {
     let identifiers = Identifiers(installations: installations)
-    let coordinator = Coordinator(identifiers: identifiers)
-    let initiator = Initiator()
+    let coordinator = SessionCoordinator(identifiers: identifiers)
+    let initiator = SessionInitiator()
 
     self.init(appID: appID,
               identifiers: identifiers,
@@ -49,7 +49,8 @@ protocol SessionsProvider {
               initiator: initiator)
   }
 
-  init(appID: String, identifiers: Identifiers, coordinator: Coordinator, initiator: Initiator) {
+  init(appID: String, identifiers: Identifiers, coordinator: SessionCoordinator,
+       initiator: SessionInitiator) {
     self.appID = appID
 
     self.identifiers = identifiers
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -38,8 +38,8 @@ protocol SessionsProvider {
 
   // MARK: - Initializers
 
-  required convenience init(appID: String, installationIDProvider: InstallationIDProvider) {
-    let identifiers = Identifiers(installationIDProvider: installationIDProvider)
+  required convenience init(appID: String, installations: InstallationsProtocol) {
+    let identifiers = Identifiers(installations: installations)
     let coordinator = Coordinator(identifiers: identifiers)
     let initiator = Initiator()
 
@@ -74,7 +74,7 @@ protocol SessionsProvider {
         guard let app = container.app, app.isDefaultApp else { return nil }
         isCacheable.pointee = true
         let installations = Installations.installations(app: app)
-        return self.init(appID: app.options.googleAppID, installationIDProvider: installations)
+        return self.init(appID: app.options.googleAppID, installations: installations)
       }]
   }
 
--- a/FirebaseSessions/Sources/Identifiers.swift
+++ b/FirebaseSessions/Sources/Identifiers.swift
@@ -41,12 +41,12 @@ protocol IdentifierProvider {
 ///   (Maybe) 4) Persisting, reading, and incrementing an increasing index
 ///
 class Identifiers: IdentifierProvider {
-  private let installationIDProvider: InstallationIDProvider
+  private let installations: InstallationsProtocol
 
   private var uuid: UUID
 
-  init(installationIDProvider: InstallationIDProvider) {
-    self.installationIDProvider = installationIDProvider
+  init(installations: InstallationsProtocol) {
+    self.installations = installations
     uuid = UUID()
   }
 
@@ -75,7 +75,7 @@ class Identifiers: IdentifierProvider {
 
     let semaphore = DispatchSemaphore(value: 0)
 
-    installationIDProvider.provideInstallationID { fiid, error in
+    installations.provideInstallationID { fiid, error in
       if let fiid = fiid {
         localInstallationID = fiid
       } else if let error = error {
--- a/FirebaseSessions/Sources/Identifiers.swift
+++ b/FirebaseSessions/Sources/Identifiers.swift
@@ -50,7 +50,7 @@ class Identifiers: IdentifierProvider {
     uuid = UUID()
   }
 
-  public func generateNewSessionID() {
+  func generateNewSessionID() {
     uuid = UUID()
 
     let lastStoredSessionID = UserDefaults.standard.string(forKey: sessionIDUserDefaultsKey) ?? ""
@@ -62,7 +62,7 @@ class Identifiers: IdentifierProvider {
 
   // This method must be run on a background thread due to how Firebase Installations
   // handles threading.
-  public var installationID: String {
+  var installationID: String {
     if Thread.isMainThread {
       Logger
         .logError(
@@ -96,11 +96,11 @@ class Identifiers: IdentifierProvider {
     return localInstallationID
   }
 
-  public var sessionID: String {
+  var sessionID: String {
     return UserDefaults.standard.string(forKey: sessionIDUserDefaultsKey) ?? ""
   }
 
-  public var lastSessionID: String {
+  var lastSessionID: String {
     return UserDefaults.standard.string(forKey: lastSessionIDUserDefaultsKey) ?? ""
   }
 }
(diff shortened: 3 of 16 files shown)


FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -26,13 +28,34 @@ protocol SessionsProvider {
...
+    identifiers = Identifiers(installationIDProvider: installations)
+    main = Main(identifiers: identifiers)
+    initiator = Initiator()
+
+    super.init()
+
+    initiator.beginListening {
+      self.identifiers.generateNewSessionID()
+      self.main.runMain()
+    }
```
ncooke3, line 58: 1. Who will be calling `required init(app: FirebaseApp)`? SDKs that use the Sessions SDK?

---

2.  I think I would treat the `init(app:)` as a convenience initializer and call into an internal initializer to inject the class members. This will help will testability. 

Something like:

```swift
  required convenience init(app: FirebaseApp) {
    // ... 
  }
    
  init(appID: String, installations: Installations, initiator: Initiator, /* ... */) {  
    // ... 
  }
    
```

--- 

3. Additionally, unrelated to this PR, but why is `init(app:)` marked as `required`? Are there plans to subclass the `Sessions` class?

samedson: Here's where it's called, and where I got the idea to add required:

Screen Shot 2022-10-04 at 11 04 54 AM

samedson: Updated to match the recommendations

samedson:
Weird because pod lib lint is passing locally:

```
➜  firebase-ios-sdk git:(sessions-deps2) bundle exec pod lib lint --sources='https://github.com/firebase/SpecsDev.git,https://github.com/firebase/SpecsStaging.git,https://cdn.cocoapods.org/' --include-podspecs='{FirebaseCore.podspec,FirebaseCoreInternal.podspec,FirebaseCoreExtension.podspec,FirebaseInstallations.podspec}' --analyze FirebaseSessions.podspec --platforms=watchos --skip-tests

 -> FirebaseSessions (10.0.0)
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | xcodebuild:  note: Build preparation complete
    - NOTE  | [watchOS] xcodebuild:  note: Planning
    - NOTE  | [watchOS] xcodebuild:  note: Building targets in dependency order

FirebaseSessions passed validation.
```

samedson:
I wonder if it's due to the second scheme issue:

```
 -> FirebaseSessions (10.0.0)
    - ERROR | [watchOS] xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - NOTE  | [watchOS] xcodebuild:  xcodebuild: error: Unable to find a destination matching the provided destination specifier:
```

paulb777:
Something fundamental about watchos has changed lately. We had several nightlies fail last night. My initial thought is the default GHA Xcode no longer includes a watchos simulator

FirebaseSessions/generate_project.sh
```diff
@@ -0,0 +1,27 @@
...
+
+readonly DIR="$( git rev-parse --show-toplevel )"
+
+#
+# Use the following command in this comment if you're iterating on the
+# podspec. This will prevent settings from changing when you reload the project.
+#
+# pod gen FirebaseSessions.podspec --local-sources=./ --auto-open --platforms=ios,macos,tvos
+#
+pod gen "${DIR}/FirebaseSessions.podspec" --auto-open --gen-directory="${DIR}/gen" --local-sources="${DIR}" --platforms=ios,macos,tvos --clean
```
jeremyjiang-dev, line 26: Should we add watchOS?

jeremyjiang-dev (approved):
Great work! Thanks Sam!

FirebaseSessions/generate_project.sh
samedson: So I originally had it in there, but unfortunately it doesn't work on CI right now. The way Cocoapods generates watchOS tests is broken: https://github.com/CocoaPods/CocoaPods/issues/8283

FirebaseSessions/Sources/Logger.swift
```diff
@@ -0,0 +1,57 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+@_implementationOnly import FirebaseCoreExtension
+
```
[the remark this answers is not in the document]
samedson: Awesome that makes sense.

FirebaseSessions/Sources/Main.swift
```diff
@@ -0,0 +1,30 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+///
+/// Main is responsible for coordinating the systems in this SDK involved with sending a Session Start event.
+///
+class Main: NSObject {
```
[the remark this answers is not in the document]
samedson: This class is going to expand much more later, because I'm trying to keep business logic out of FirebaseSessions.swift. I've renamed it to Coordinator based on the responsibilities it has

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -26,13 +28,42 @@ protocol SessionsProvider {
...
+  /// Top-level Classes in the Sessions SDK
+  private let main: Main
+  private let initiator: Initiator
+  private let identifiers: Identifiers
 
   // MARK: - Initializers
 
-  required init(app: FirebaseApp) {
-    googleAppID = app.options.googleAppID
+  required convenience init(app: FirebaseApp) {
```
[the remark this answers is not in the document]
samedson: I like the idea of reducing our dependency surface to just the functions we need. One thing I've done instead of passing a function is kept the protocol in there because we'll be using that pattern all over the SDK and I am trying to reuse patterns as much as possible.

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -26,13 +28,42 @@ protocol SessionsProvider {
 @objc(FIRSessions) class Sessions: NSObject, Library, SessionsProvider {
```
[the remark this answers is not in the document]
samedson: I don't think it's necessary either, but the Component constructor only takes a protocol, so it seems I needed to make one up

FirebaseSessions/Sources/Identifiers.swift
```diff
@@ -0,0 +1,81 @@
...
+  public var installationID: String {
+    var localInstallationID = ""
+    installationIDProvider.provideInstallationID { fiid, error in
+      if error != nil {
+        logError("Error getting Firebase Installation ID: \(error.debugDescription)")
+      }
+      localInstallationID = fiid ?? ""
+    }
+    return localInstallationID
+  }
```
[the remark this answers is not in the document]
samedson: Oof let me fix that

FirebaseSessions/Sources/Logger.swift
```diff
@@ -0,0 +1,59 @@
...
+let logServiceTag = "[FirebaseSessions]"
+let logCode = "I-SES000000"
+
+enum Logger {
```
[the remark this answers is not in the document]
samedson: That's a good call - also somehow I made that an enum instead of a class so I've fixed that

FirebaseSessions/Sources/Initiator.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+///
+/// The Initiator is responsible for:
+///   1) Running the initiate callback whenever a Session Start Event should
+///      begin sending. This can happen at a cold start of the app, and when it
+///      been in the background for a period of time (originally set at 30 mins)
+///      and comes to the foreground.
+///
+class Initiator: NSObject {
```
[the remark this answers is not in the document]
samedson: This class should stay internal to the Sessions SDK. We'll have an explicit private API that other SDKs read from.

FirebaseSessions/Sources/Identifiers.swift
```diff
@@ -0,0 +1,81 @@
...
+}
+
+///
+/// Identifiers is responsible for:
+///   1) Getting the Installation ID from Installations
+///   2) Generating the Session ID
+///   3) Persisting and reading the Session ID from the last session
+///   (Maybe) 4) Persisting, reading, and incrementing an increasing index
+///
+class Identifiers: NSObject, IdentifierProvider {
```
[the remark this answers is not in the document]
samedson: Same: This class should stay internal to the Sessions SDK. We'll have an explicit private API that other SDKs read from.

FirebaseSessions/Sources/Logger.swift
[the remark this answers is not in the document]
samedson: Oh cool ok then we'll go with enum

FirebaseSessions/Sources/Identifiers.swift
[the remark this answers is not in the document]
samedson: Fixed. Wish I could use async await

FirebaseSessions/Sources/Identifiers.swift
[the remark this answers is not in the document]
samedson: Added some for the InstallationID

FirebaseSessions/Sources/Initiator.swift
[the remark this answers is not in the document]
ncooke3: Ok, I commented because this class had an API defined with public visibility. The `public` attribute should be removed then?

FirebaseSessions/Sources/Coordinator.swift
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+///
+/// Main is responsible for coordinating the systems in this SDK involved with sending a Session Start event.
+///
```
ncooke3, line 19: Xcode supercut: select anywhere in the symbol/function name and do: ⌥ + ⌘ + /  to generate a formatted doc comment.

samedson: That's awesome - gonna use that from now on

FirebaseSessions/Sources/Identifiers.swift
```diff
@@ -0,0 +1,106 @@
...
+    let lastStoredSessionID = UserDefaults.standard.string(forKey: sessionIDUserDefaultsKey) ?? ""
+    UserDefaults.standard.set(lastStoredSessionID, forKey: lastSessionIDUserDefaultsKey)
+
+    let newSessionID = uuid.uuidString.replacingOccurrences(of: "-", with: "").lowercased()
+    UserDefaults.standard.set(newSessionID, forKey: sessionIDUserDefaultsKey)
+  }
+
+  // This method must be run on a background thread due to how Firebase Installations
+  // handles threading.
+  var installationID: String {
```
ncooke3, line 65: I'm not sure we should try to make this getter behave synchronously. I think we should instead replace this property with a method that takes a completion handler to get the block.

FirebaseSessions/Sources/FirebaseSessions.swift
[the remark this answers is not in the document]
ncooke3: Discussed offline. Will hold off.

FirebaseSessions/Sources/Identifiers.swift
samedson: That would make using the api much harder though no? At some point I need an ugly semaphor and I'd rather not offload that to other classes

FirebaseSessions/Sources/Installations+InstallationIDProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+
+protocol InstallationIDProvider {
+  func provideInstallationID(provide: @escaping (String?, Error?) -> Void)
+}
+
+extension Installations: InstallationIDProvider {
+  func provideInstallationID(provide: @escaping (String?, Error?) -> Void) {
+    installationID(completion: provide)
+  }
+}
```
ncooke3, line 27: I think we should do three things if we go the protocol route:
1. Hide away the optional completion handler arguments by using `Result` type. It'll make usage easier and more consistent throughout the implementation and tests
2. This one is more personal opinion so feel free to push back: name the protocol `InstallationsProtocol` because it's existence is only to abstract away the `Installations` type for testability's sake. This was a consensus the team arrived at while working on another Swift project. If the protocol is meant to be conformed too by multiple objects, a generic name inline with Swift Protocol's naming guidelines is preferred. If the protocol is mean to abstract away a concrete type for testability then just name the protocol the _concrete type_ + _Protocol_... so here would be _InstallationsProtocol_.

```suggestion
protocol InstallationsProtocol {
  func installationID(completion: @escaping (Result<String, Error>) -> Void)
}

extension Installations: InstallationsProtocol {
    func installationID(completion: @escaping (Result<String, Error>) -> Void) {
        self.installationID { (installationID: String?, error: Error?) in
            if let installationID = installationID {
              completion(.success(installationID))
            } else if let error = error {
              completion(.failure(error))
            }
        }
    }
}

```

samedson: 2 sounds good. My only issue with 1 is it means I can't test that extension - maybe I can just improve the logic in `Identifiers` to use the `if let ...` flow?

FirebaseSessions/Sources/Identifiers.swift
ncooke3: I would just keep the API surface asynchronous with completion handlers and avoid using semaphores throughout the SDK.

Currently, accessing this property will call to installations every time. How many times in the SDK's lifecycle will we need an installation ID? Only once at every startup of the SDK?

FirebaseSessions/Sources/Installations+InstallationIDProvider.swift
ncooke3: > My only issue with 1 is it means I can't test that extension

Hmm, I'm not 100% following– there were no visibility changes?

ncooke3: Meaning the `InstallationsProtocol` should be visible during testing when doing `@testable import FirebaseSessions`

samedson: Just pushed up my suggestion

samedson: Also pushed up the rename

ncooke3: Thanks Sam, I still don't follow why we can't use the `func installationID(completion: @escaping (Result<String, Error>) -> Void)` signature?

samedson: Oh cause I can't test that part of the code

ncooke3: Oh there _isn't_ anything to test. We are just wrapping the Installations API with a new API to return the `Result`-type.

ncooke3: I guess I don't see how it's any different from the `func provideInstallationID(provide: @escaping (String?, Error?) -> Void) ` except that it has the `Result` parameter as syntactic sugar.

samedson: Ok updated it

ncooke3 (approved):

FirebaseSessions/Sources/Initiator.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+///
+/// The Initiator is responsible for:
+///   1) Running the initiate callback whenever a Session Start Event should
+///      begin sending. This can happen at a cold start of the app, and when it
+///      been in the background for a period of time (originally set at 30 mins)
+///      and comes to the foreground.
+///
+class Initiator {
```
ncooke3, line 24: Maybe `SessionListener` or `SessionLifecycle(Manager|Controller)`?

FirebaseSessions/Sources/Coordinator.swift
```diff
@@ -0,0 +1,30 @@
...
+class Coordinator {
```
ncooke3, line 20: I think I'd opt to be more specific and say `SessionCoordinator`. They are all internal types but it's not clear on the name alone what is being coordinated.

ncooke3 (approved):
Thanks for working through comments. I have a few remaining that can be addressed now or in next PR

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -23,16 +25,43 @@ protocol SessionsProvider {
...
+  private let identifiers: Identifiers
 
   // MARK: - Initializers
 
-  required init(app: FirebaseApp) {
-    googleAppID = app.options.googleAppID
+  required convenience init(appID: String, installations: InstallationsProtocol) {
+    let identifiers = Identifiers(installations: installations)
+    let coordinator = Coordinator(identifiers: identifiers)
+    let initiator = Initiator()
```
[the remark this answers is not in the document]
samedson: I don't think I want to - Identifiers will likely be passed around to multiple classes and I'd like to test the Coordinator while mocking Identifiers

FirebaseSessions/Sources/Identifiers.swift
```diff
@@ -0,0 +1,109 @@
...
+}
+
+///
+/// Identifiers is responsible for:
+///   1) Getting the Installation ID from Installations
+///   2) Generating the Session ID
+///   3) Persisting and reading the Session ID from the last session
+///   (Maybe) 4) Persisting, reading, and incrementing an increasing index
+///
+class Identifiers: IdentifierProvider {
```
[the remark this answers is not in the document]
samedson: But this class is a collection of identifiers - it also has the installation ID

FirebaseSessions/Sources/Initiator.swift
samedson: Eh I don't like -Manager names though cause it's an overloaded term (and Controller insinuates MVC). I'm thinking SessionInitiator because it initiates the Session and that's how I've been describing it in the design

ncooke3: SessionInitiator - i like it

PR #10304, opened 2022-10-04
FirestoreSwift: Make DocumentID setter internal

andrewheard:
**This is an alternative to the warning-only approach proposed in the latest `DocumentID` API review.**

Made the `@DocumentID` property wrapper's value setter internal. `DocumentID` is currently read-only and is only populated when reading from Firestore. This PR limits access to the setter since the value is ignored when writing to Firestore.

Updated the constructor to always set `nil`, regardless of the `wrappedValue` provided. Developers are notified of this via an existing log message warning them that the value is ignored.

Added conformance tests to verify (and demonstrate) the functionality change.

### DIFF
--- a/Firestore/Swift/Source/Codable/DocumentID.swift
+++ b/Firestore/Swift/Source/Codable/DocumentID.swift
@@ -125,7 +125,6 @@ public struct DocumentID<Value: DocumentIDWrappable & Codable>:
     self.value = nil
   }
 
-  
   /// Gets the document identifier if populated; `nil` otherwise.
   ///
   /// - Important: This property is read-only; the setter is `internal`.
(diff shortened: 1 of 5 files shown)


Firestore/Swift/Source/Codable/DocumentID.swift
```diff
@@ -105,21 +105,33 @@ internal protocol DocumentIDProtocol {
...
   public init(wrappedValue value: Value?) {
     if let value = value {
       logIgnoredValueWarning(value: value)
     }
-    self.value = value
+    self.value = nil
   }
 
-  public var wrappedValue: Value? {
+  /// Gets the document identifier if populated; `nil` otherwise.
```
peterfriese, line 128: I think we always call it document ID (not document identifier)

PR #10305, opened 2022-10-05
FirestoreSwift: Remove DocumentID breaking change from changelog

andrewheard:
Updated the FirestoreSwift CHANGELOG to match the current state of the repo. Removed the note about the breaking change for `DocumentID` since we are only logging warnings for this release.

### DIFF
--- a/Firestore/Swift/CHANGELOG.md
+++ b/Firestore/Swift/CHANGELOG.md
@@ -1,9 +1,7 @@
 # 10.0.0
-- [changed] **Breaking Change:** Made the `@DocumentID` property wrapper value
-  setter internal to clarify that the value is ignored during writes. (#9368)
-- [changed] Initializing a `@DocumentID` property wrapper with a non-nil value
-  or using the `@DocumentID` property wrapper value setter will log a warning.
-  This is because the set value will be ignored. (#9368)
+- [changed] Passing a non-nil value to the `@DocumentID` property wrapper's
+  constructor or setter will log a warning. This is because the set value will
+  be ignored. (#9368)
 - [changed] `Firestore.Encoder` and `Firestore.Decoder` now wraps the shared
   `FirebaseDataEncoder` and `FirebaseDataDecoder` types, which provides new
   customization options for encoding and decoding data to and from Firestore


Firestore/Swift/CHANGELOG.md
```diff
@@ -1,9 +1,7 @@
 # 10.0.0
-- [changed] **Breaking Change:** Made the `@DocumentID` property wrapper value
-  setter internal to clarify that the value is ignored during writes. (#9368)
-- [changed] Initializing a `@DocumentID` property wrapper with a non-nil value
-  or using the `@DocumentID` property wrapper value setter will log a warning.
-  This is because the set value will be ignored. (#9368)
+- [changed] Passing a non-nil value to the `@DocumentID` property wrapper's
+  constructor or setter will log a warning. This is because the set value will
```
morganchen12, line 3: nit: `will log a warning and the set value will be ignored` is more concise.

morganchen12 (approved):

Firestore/Swift/CHANGELOG.md
andrewheard: Sounds better too. Thanks!

Firestore/Swift/CHANGELOG.md
```diff
@@ -1,9 +1,7 @@
 # 10.0.0
-- [changed] **Breaking Change:** Made the `@DocumentID` property wrapper value
-  setter internal to clarify that the value is ignored during writes. (#9368)
```
paulb777, line 3: Isn't this still true that we made the setter internal?

Firestore/Swift/CHANGELOG.md
```diff
@@ -1,9 +1,7 @@
...
-- [changed] Initializing a `@DocumentID` property wrapper with a non-nil value
-  or using the `@DocumentID` property wrapper value setter will log a warning.
-  This is because the set value will be ignored. (#9368)
+- [changed] Passing a non-nil value to the `@DocumentID` property wrapper's
+  constructor or setter will log a warning and the set value will be ignored.
+  (#9368)
```
ncooke3, line 4: Isn't the breaking change that we removed the public visibility from this init:

Screen Shot 2022-10-05 at 1 00 37 PM

_(Diff from #10167)_

andrewheard: I'd argue that this isn't a breaking change since it was only visible when `DocumentID` _wasn't_ used as a property wrapper, i.e., `DocumentID<DocumentReference>(from: docRef)`.

The following is no longer possible (though I can't think of a use case):
```
let userDoc: DocumentReference = Firestore.firestore().document("users/foo")
do {
  let docID = try DocumentID<DocumentReference>(from: userDoc)
  if let docRef = docID.wrappedValue {
    print("Document ID: \(docRef.documentID)")
  }
} catch {
  print("DocumentID Error: \(error)")
}
```

The visibility change doesn't affect usage of at `@DocumentID` though. The following is still possible (though it will trigger our warnings at runtime:
```
struct User: Codable {
  @DocumentID var docRef: DocumentReference? = Firestore.firestore().document("users/foo")
}
```

We could list the above as a breaking change but I think it could cause confusion since DocumentID has always been intended to be used as a property wrapper. Let me know if you feel differently, of course, though.

Firestore/Swift/CHANGELOG.md
andrewheard: Sorry, I responded in the [thread](https://github.com/firebase/firebase-ios-sdk/pull/10305#discussion_r988149779) below.

Firestore/Swift/CHANGELOG.md
ncooke3: Thanks for the thorough answer! I think that, if something is no longer possible (even if it doesn't have a use case), it's still a breaking change. Curious what @morganchen12 and @paulb777  think though.

paulb777: I've seen plenty of examples of people using tooling in completely unexpected ways, so that would be an argument to call it breaking. On the flip side, if we're convinced it won't actually break anything, I'm fine to remove the breaking notification. I'll defer to Andrew's judgement since he's looked at it the most, unless Morgan feels strongly one way or the other.

paulb777 (approved):

ncooke3 (approved):

Firestore/Swift/CHANGELOG.md
andrewheard: Let's err on the side of caution. I reworded the changelog and specifically mentioned that `@DocumentID` is unchanged to avoid any confusion.

morganchen12 (approved):

paulb777 (approved):

Issue #8200, opened 2021-06-05
Building SwiftProtobuf with BUILD_LIBRARY_FOR_DISTRIBUTION=YES is not supported


acecilia:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: `12.3`
  * Firebase SDK version: `8.1.0`
  * Installation method: `Carthage`
  * Firebase Component: `FirebaseMLModelDownloader`

### [REQUIRED] Step 2: Describe the problem

* [This PR](https://github.com/firebase/firebase-ios-sdk/pull/7281) added SwiftProtobuf as a dependency of `FirebaseMLModelDownloader`
* The carthage distribution seems like it [uses BUILD_LIBRARY_FOR_DISTRIBUTION=YES](https://github.com/firebase/firebase-ios-sdk/blob/4f00074ad71b17be146d6941b414936b96bf2e45/ReleaseTooling/Sources/ZipBuilder/FrameworkBuilder.swift#L215)
* But the problem is that SwiftProtobuf [does not support being built with BUILD_LIBRARY_FOR_DISTRIBUTION=YES](https://github.com/apple/swift-protobuf/issues/1101#issuecomment-752030103). From the linked comment:

> Swift Protobuf does not support itself being built in library evolution mode, so it cannot be present in the public ABI of any framework that is built in library evolution mode. That means it can only ever be imported @_implementationOnly, and its types can never be public.

I have other third parties in my app that depend on `SwiftProtobuf` ([grpc](https://github.com/grpc/grpc-swift/blob/main/gRPC-Swift.podspec#L27)), so when using `FirebaseMLModelDownloader` I am building my own `SwiftProtobuf` xcframework using Carthage, and that uses `BUILD_LIBRARY_FOR_DISTRIBUTION=NO`. When doing that, multiple link errors show up (Undefined symbols). See the following issue for an extensive description of the problem: https://github.com/apple/swift-protobuf/issues/1101

#### Steps to reproduce:

See: https://github.com/apple/swift-protobuf/issues/1101#issuecomment-855076010

paulb777:
@acecilia Thanks for the report and sharing the thorough investigation.

We'll have to investigate if there are any alternatives for the `FirebaseMLModelDownloader` binary distributions.

[...]

PR #10311, opened 2022-10-06
[Sessions] Add SessionStartEvent wrapper for building the nanopb proto

samedson:
- Adds functions for interacting with nanopb. Some parts of this were difficult to do in Swift, so I had to do them in ObjC.
 - Adds SessionStartEvent as an abstraction over the proto
 - Adds FireLog dependency, and logs events to FireLog

#no-changelog

### DIFF
--- a/FirebaseSessions.podspec
+++ b/FirebaseSessions.podspec
@@ -35,16 +35,22 @@ Pod::Spec.new do |s|
 
   base_dir = "FirebaseSessions/"
   s.source_files = [
-    base_dir + 'Sources/**/*.swift',
+    base_dir + 'Sources/**/*.{swift,h,m}',
+    base_dir + 'Protogen/**/*.{c,h,m,mm}',
   ]
 
   s.dependency 'FirebaseCore', '~> 10.0'
   s.dependency 'FirebaseCoreExtension', '~> 10.0'
   s.dependency 'FirebaseInstallations', '~> 10.0'
+  s.dependency 'GoogleDataTransport', '~> 9.2'
+  s.dependency 'nanopb', '>= 2.30908.0', '< 2.30910.0'
 
   s.pod_target_xcconfig = {
     'GCC_C_LANGUAGE_STANDARD' => 'c99',
-    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"'
+    'HEADER_SEARCH_PATHS' => '"${PODS_TARGET_SRCROOT}"',
+    'GCC_PREPROCESSOR_DEFINITIONS' =>
+      # For nanopb:
+      'PB_FIELD_32BIT=1 PB_NO_PACKED_STRUCTS=1 PB_ENABLE_MALLOC=1',
   }
 
   s.test_spec 'unit' do |unit_tests|
--- a/FirebaseSessions/Sources/EventGDTLogger.swift
+++ b/FirebaseSessions/Sources/EventGDTLogger.swift
@@ -1,0 +1,49 @@
+//
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+@_implementationOnly import GoogleDataTransport
+
+protocol EventGDTLoggerProtocol {
+  func logEvent(event: SessionStartEvent, completion: @escaping (Result<Void, Error>) -> Void)
+}
+
+///
+/// EventGDTLogger is responsible for
+///   1) Creating GDT Events and logging them to the GoogleDataTransport SDK
+///   2) Handling debugging situations (eg. running in Simulator or printing the event to console)
+///
+class EventGDTLogger: EventGDTLoggerProtocol {
+  let googleDataTransport: GoogleDataTransportProtocol
+
+  init(googleDataTransport: GoogleDataTransportProtocol) {
+    self.googleDataTransport = googleDataTransport
+  }
+
+  /// Logs the event to FireLog, taking into account debugging cases such as running
+  /// in simulator.
+  func logEvent(event: SessionStartEvent, completion: @escaping (Result<Void, Error>) -> Void) {
+    let gdtEvent = googleDataTransport.eventForTransport()
+    gdtEvent.dataObject = event
+    gdtEvent.qosTier = GDTCOREventQoS.qosDefault
+    if ProcessInfo.processInfo.environment["SIMULATOR_UDID"] != nil {
+      Logger.logDebug("Logging events using fast QOS due to running on a simulator")
+      gdtEvent.qosTier = GDTCOREventQoS.qoSFast
+    }
+
+    googleDataTransport.logGDTEvent(event: gdtEvent, completion: completion)
+  }
+}
--- a/FirebaseSessions/Sources/EventGDTLogger.swift
+++ b/FirebaseSessions/Sources/EventGDTLogger.swift
@@ -39,10 +39,10 @@ class EventGDTLogger: EventGDTLoggerProtocol {
     let gdtEvent = googleDataTransport.eventForTransport()
     gdtEvent.dataObject = event
     gdtEvent.qosTier = GDTCOREventQoS.qosDefault
-    if ProcessInfo.processInfo.environment["SIMULATOR_UDID"] != nil {
+    #if targetEnvironment(simulator)
       Logger.logDebug("Logging events using fast QOS due to running on a simulator")
       gdtEvent.qosTier = GDTCOREventQoS.qoSFast
-    }
+    #endif  // targetEnvironment(simulator)
 
     googleDataTransport.logGDTEvent(event: gdtEvent, completion: completion)
   }
--- a/FirebaseSessions/Sources/EventGDTLogger.swift
+++ b/FirebaseSessions/Sources/EventGDTLogger.swift
@@ -42,7 +42,7 @@ class EventGDTLogger: EventGDTLoggerProtocol {
     #if targetEnvironment(simulator)
       Logger.logDebug("Logging events using fast QOS due to running on a simulator")
       gdtEvent.qosTier = GDTCOREventQoS.qoSFast
-    #endif  // targetEnvironment(simulator)
+    #endif // targetEnvironment(simulator)
 
     googleDataTransport.logGDTEvent(event: gdtEvent, completion: completion)
   }
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -77,7 +77,7 @@ protocol SessionsProvider {
       self.identifiers.generateNewSessionID()
       let event = SessionStartEvent(identifiers: self.identifiers)
       DispatchQueue.global().async {
-        self.coordinator.beginLoggingSessionStart(event: event) { result in
+        self.coordinator.attemptLoggingSessionStart(event: event) { result in
         }
       }
     }
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -14,11 +14,15 @@
 
 import Foundation
 
-import FirebaseCore
-import FirebaseInstallations
-
 // Avoids exposing internal FirebaseCore APIs to Swift users.
 @_implementationOnly import FirebaseCoreExtension
+@_implementationOnly import FirebaseInstallations
+@_implementationOnly import GoogleDataTransport
+
+private enum GoogleDataTransportConfig {
+  static let sessionsLogSource = "1974"
+  static let sessionsTarget = GDTCORTarget.FLL
+}
 
 @objc(FIRSessionsProvider)
 protocol SessionsProvider {
@@ -38,9 +42,18 @@ protocol SessionsProvider {
 
   // MARK: - Initializers
 
+  // Initializes the SDK and top-level classes
   required convenience init(appID: String, installations: InstallationsProtocol) {
+    let googleDataTransport = GDTCORTransport(
+      mappingID: GoogleDataTransportConfig.sessionsLogSource,
+      transformers: nil,
+      target: GoogleDataTransportConfig.sessionsTarget
+    )
+
+    let fireLogger = EventGDTLogger(googleDataTransport: googleDataTransport!)
+
     let identifiers = Identifiers(installations: installations)
-    let coordinator = SessionCoordinator(identifiers: identifiers)
+    let coordinator = SessionCoordinator(identifiers: identifiers, fireLogger: fireLogger)
     let initiator = SessionInitiator()
 
     self.init(appID: appID,
@@ -49,6 +62,7 @@ protocol SessionsProvider {
               initiator: initiator)
   }
 
+  // Initializes the SDK and begines the process of listening for lifecycle events and logging events
   init(appID: String, identifiers: Identifiers, coordinator: SessionCoordinator,
        initiator: SessionInitiator) {
     self.appID = appID
@@ -61,7 +75,11 @@ protocol SessionsProvider {
 
     self.initiator.beginListening {
       self.identifiers.generateNewSessionID()
-      self.coordinator.runMain()
+      let event = SessionStartEvent(identifiers: self.identifiers)
+      DispatchQueue.global().async {
+        self.coordinator.beginLoggingSessionStart(event: event) { result in
+        }
+      }
     }
   }
 
--- a/FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
+++ b/FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
@@ -17,8 +17,8 @@ import Foundation
 
 @_implementationOnly import GoogleDataTransport
 
-struct GoogleDataTransportProtocolErrors {
-  static let writeFailureError = NSError(domain: "GoogleDataTransportProtocol", code: -1)
+enum GoogleDataTransportProtocolErrors: Error {
+  case writeFailure
 }
 
 protocol GoogleDataTransportProtocol {
@@ -32,7 +32,7 @@ extension GDTCORTransport: GoogleDataTransportProtocol {
       if let error = error {
         completion(.failure(error))
       } else if !wasWritten {
-        completion(.failure(GoogleDataTransportProtocolErrors.writeFailureError))
+        completion(.failure(GoogleDataTransportProtocolErrors.writeFailure))
       } else {
         completion(.success(()))
       }
--- a/FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
+++ b/FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
@@ -1,0 +1,41 @@
+//
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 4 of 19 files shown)


FirebaseSessions.podspec
```diff
@@ -35,12 +35,15 @@ Pod::Spec.new do |s|
 
   base_dir = "FirebaseSessions/"
   s.source_files = [
-    base_dir + 'Sources/**/*.swift',
+    base_dir + 'Sources/**/*.{swift,h,m}',
+    base_dir + 'Protogen/**/*.{c,h,m,mm}',
```
ncooke3, line 39: Heads up that this will unfortunately not be as straightforward when setting up SwiftPM support due to the no mixed languages constraint. There are some workarounds, however.

And also, we should be able to call C from Swift... that might be able to help us.

samedson: Ooh that's a really good callout. Do you know of these workarounds? I would love to learn this power. Also, this code could easily be written in C if that works (in fact I believe it is valid C).

ncooke3: > Do you know of these workarounds? I would love to learn this power. 

I could be way off but I'm getting prequels Palpatine x Anakin vibes 😄 

But yes! We could create a target called something like `SessionsProtoLib` that the main Sessions SwiftPM target depends on. That should work. So basically just separate the Sessions SDK into a Swift target that depends on a C based target with the proto stuff. 

An alternative idea that should avoid all of this (and is probably easier) is to use: https://github.com/apple/swift-protobuf which is Apple's official Swift proto library.

paulb777: I would like the idea of using SwiftProtobuf for Swift libraries, but SwiftProtobuf doesn't support binary distributions. See https://github.com/firebase/firebase-ios-sdk/issues/8200 and https://github.com/apple/swift-protobuf/issues/1101

ncooke3: @samedson, looked into and we are going to continue to use nanopb. LMK if you need help structuring the SPM stuff

FirebaseSessions/Sources/Time.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol TimeProvider {
```
leotianlizhan, line 18: Since this is a "provider" pattern only used by session start event, could you take a look at https://github.com/firebase/firebase-ios-sdk/pull/10360 and see if, in this case, would it make sense to inject function instead of object?

FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -0,0 +1,60 @@
...
+
+///
+/// SessionStartEvent is responsible for:
+///   1) Writing fields to the Session proto
+///   2) Synthesizing itself for persisting to disk and logging to GoogleDataTransport
+///
+class SessionStartEvent: NSObject, GDTCOREventDataObject {
+  var proto: firebase_appquality_sessions_SessionEvent
+
+  init(identifiers: IdentifierProvider, time: TimeProvider) {
```
leotianlizhan, line 28: nit: give `time` param a default argument of `Time`, so `FirebaseSessions` doesn't need to have extra code.

samedson: That's a great idea

FirebaseSessions/Sources/Time.swift
samedson: That's a good question - I've been favoring the Provider pattern for 2 reasons:
 - Consistency
 - If there's another class that needs Time, then it's easy to move it around.

In this case I don't see another use case for time, so I'm down to go with the function approach.

samedson: Ehh actually I might keep the Provider. After writing up the functional solution it seems less readable to me

FirebaseSessions/Sources/EventFireLogger.swift
```diff
@@ -0,0 +1,42 @@
...
+
+import Foundation
+
+@_implementationOnly import GoogleDataTransport
+
+protocol EventFireLoggerProtocol {
+  func logEvent(event: SessionStartEvent, completion: @escaping (Result<Void, Error>) -> Void)
+}
+
+class EventFireLogger: EventFireLoggerProtocol {
```
visumickey, line 24: A design suggestion: Can we set this class in such a way that we bring in an option (dispatch mode) to just dump the event to the console rather than dispatch to Firelog? This would be helpful for our local testing where we don't have to wait for long to see the events to reach Firelog to ascertain that the events are as expected. FWIW: Default is Dispatch to Firelog.

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -19,6 +19,12 @@ import FirebaseInstallations
 
 // Avoids exposing internal FirebaseCore APIs to Swift users.
 @_implementationOnly import FirebaseCoreExtension
+@_implementationOnly import GoogleDataTransport
+
+private enum GoogleDataTransportConfig {
+  static let sessionsMappingFileID = "1974"
```
visumickey, line 25: Is this the log source? If so, can we call it a LOG SOURCE instead of calling it Mapping File ID?

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -61,7 +75,11 @@ protocol SessionsProvider {
 
     self.initiator.beginListening {
       self.identifiers.generateNewSessionID()
-      self.coordinator.runMain()
+      let event = SessionStartEvent(identifiers: self.identifiers)
+      DispatchQueue.global().async {
+        self.coordinator.beginLoggingSessionStart(event: event) { result in
```
visumickey, line 80: Should we call this method as `logSessionStartEvent`? Prefix of `begin` is hard to understand what to expect with that method. I'm generally in favor of using methods with a Verb as a prefix.

FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
```diff
@@ -0,0 +1,38 @@
...
+  func eventForTransport() -> GDTCOREvent
+}
+
+extension GDTCORTransport: GoogleDataTransportProtocol {
+  func logGDTEvent(event: GDTCOREvent, completion: @escaping (Result<Void, Error>) -> Void) {
+    sendDataEvent(event) { wasWritten, error in
+      if let error = error {
+        completion(.failure(error))
+      } else if !wasWritten {
+        let writeFailureError = NSError(domain: "GoogleDataTransportProtocol", code: -1)
```
visumickey, line 31: Do we plan to have all the constants defined in a common place? Eg: GoogleDataTransportProtocol

FirebaseSessions/Sources/NanoPB/FIRSESNanoPBHelpers.h
```diff
@@ -0,0 +1,55 @@
...
+
+/** Mallocs a pb_bytes_array and copies the given NSString's bytes into the bytes array.
+ * @note Memory needs to be freed manually, through pb_free or pb_release.
+ * @param string The string to encode as pb_bytes.
+ */
+pb_bytes_array_t* _Nullable FIRSESEncodeString(NSString* _Nullable string);
+
+BOOL FIRSESIsPBArrayEqual(pb_bytes_array_t* _Nullable array, pb_bytes_array_t* _Nullable expected);
+
+BOOL FIRSESIsPBStringEqual(pb_bytes_array_t* _Nullable pbString, NSString* _Nullable str);
```
visumickey, line 49: I'm missing comments for each of these functions. Can you clarify them with some details?

FirebaseSessions/Sources/NanoPB/FIRSESNanoPBHelpers.m
```diff
@@ -0,0 +1,129 @@
...
+  return [NSError errorWithDomain:@"FIRSESEncodeError"
+                             code:-1
+                         userInfo:@{@"NSLocalizedDescriptionKey" : description}];
+}
+
+// It seems impossible to specify the nullability of the `fields` parameter below,
+// yet the compiler complains that it's missing a nullability specifier. Google
+// yields no results at this time.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnullability-completeness"
```
visumickey, line 36: `nullability-completeness` was used for older version of XCode where there was a big. This got fixed in 10.1 (I think). Do we still need this?

FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -0,0 +1,60 @@
...
+import Foundation
+
+@_implementationOnly import GoogleDataTransport
+
+///
+/// SessionStartEvent is responsible for:
+///   1) Writing fields to the Session proto
+///   2) Synthesizing itself for persisting to disk and logging to GoogleDataTransport
+///
+class SessionStartEvent: NSObject, GDTCOREventDataObject {
```
visumickey, line 25: Design suggestion: I would recommend keeping proto conversion away from the actual event. SessionStartEvent as a class can have all the essentials for the event and a separate layer can be used to convert the SessionStartEvent to a SessionEventProto and can be used for dispatch.

This would help in moving the session start event across the library until we are ready for dispatch. I would look at this event as meta data preparation in this phase and a dispatch is when we need to turn them to  proto.

visumickey (commented):
Apart from the comments: 3 major suggestions:

1. We need a clear/module class interaction defined and name the classes according to them. Some generic terms used for classes would grow in time to become hard to maintain. 
2. Lets add comments to every class/function/variable to ensure that it becomes easier to understand that the class/function is supposed to do. Right now, I find it hard to understand the purpose of them
3. Logging levels for log messages - We need to have a good defined rule about the logging level for messages. Typically, `error` level is set for only something that the user has set something wrong as. For eg: bad app Id name. Anything else, should have a log level of `warning` or below. It might be good to even define them in the execution doc.

FirebaseSessions/Sources/NanoPB/FIRSESNanoPBHelpers.m
samedson: Yeah the comment above explains the problem

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: I used `begin` to convey the fact that it's not always going to log the event. This method needs to take into account settings, sampling, etc.

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: I'm down to switch it. They call it mapping file ID in the GDT API so that's why I went with that

FirebaseSessions/Sources/EventFireLogger.swift
samedson: Added it to Taskflow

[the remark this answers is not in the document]
samedson: I worry that EventDispatcher would be confusing because on iOS Dispatch is used to convey threading, where this class is responsible for logging to FireLog. The method on GDT is also called `logGDTEvent`. I understand that logging also pertains to printing to console logs, but in this case I do think EventFireLogger is very clear.

Maybe `EventGDTLogger`?

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: Also, this was confusing to me, but constants are lower camel case in the style guide: go/swift-style#global-constants

samedson:
> Apart from the comments: 3 major suggestions:
> 
> 1. We need a clear/module class interaction defined and name the classes according to them. Some generic terms used for classes would grow in time to become hard to maintain.
> 2. Lets add comments to every class/function/variable to ensure that it becomes easier to understand that the class/function is supposed to do. Right now, I find it hard to understand the purpose of them
> 3. Logging levels for log messages - We need to have a good defined rule about the logging level for messages. Typically, `error` level is set for only something that the user has set something wrong as. For eg: bad app Id name. Anything else, should have a log level of `warning` or below. It might be good to even define them in the execution doc.

1. The way I think about naming classes is to look at what they do. I often find class names that are vague using overloaded terms like -Manager -Controller, etc. and I tend to move away from suffixes like those because they have no meaning (or overloaded meanings that we aren't conforming to). If a class name can be described by 1 word I go for 1 word because the extra words are meaningless.

2. I think commenting every class is good, I don't think commenting every variable or method is worth the time though - it actually takes a lot of time to do, and doesn't provide value in the long run. I subscribe to the idea that if the method does more than one thing it's worth a comment, but most variables or methods should be self-documenting.

3. In terms of logging levels, I generally subscribe to the following: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

Warning is recoverable, Error is not. I'm interested if the core folks have advice they try to get everyone to follow.

FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
samedson: Ooh yeah let me add a constant at the top here

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: I've been having trouble naming this method - I had `main` before which also has issues

FirebaseSessions/Sources/EventGDTLogger.swift
```diff
@@ -0,0 +1,49 @@
...
+  /// Logs the event to FireLog, taking into account debugging cases such as running
+  /// in simulator.
+  func logEvent(event: SessionStartEvent, completion: @escaping (Result<Void, Error>) -> Void) {
+    let gdtEvent = googleDataTransport.eventForTransport()
+    gdtEvent.dataObject = event
+    gdtEvent.qosTier = GDTCOREventQoS.qosDefault
+    if ProcessInfo.processInfo.environment["SIMULATOR_UDID"] != nil {
+      Logger.logDebug("Logging events using fast QOS due to running on a simulator")
+      gdtEvent.qosTier = GDTCOREventQoS.qoSFast
+    }
```
ncooke3, line 45: I think it'd be better to shift this to build time. `TARGET_OS_SIMULATOR` should work.

```suggestion
    #if TARGET_OS_SIMULATOR
      Logger.logDebug("Logging events using fast QOS due to running on a simulator")
      gdtEvent.qosTier = GDTCOREventQoS.qoSFast
    #endif  // TARGET_OS_SIMULATOR
```

FirebaseSessions/Sources/FirebaseSessions.swift
ncooke3: Maybe `attemptLoggingSessionStartEvent`

FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
```diff
@@ -0,0 +1,41 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+@_implementationOnly import GoogleDataTransport
+
+struct GoogleDataTransportProtocolErrors {
+  static let writeFailureError = NSError(domain: "GoogleDataTransportProtocol", code: -1)
+}
```
ncooke3, line 22: Did you consider a Swift error enum? https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html

FirebaseSessions/Sources/SessionStartEvent.swift
visumickey: Separation is always hard for such situations. If we can, I would recommend doing the separation now.

FirebaseSessions/Tests/Unit/Mocks/MockGDTLogger.swift
```diff
@@ -0,0 +1,30 @@
...
+@testable import FirebaseSessions
+
+class MockGDTLogger: EventGDTLoggerProtocol {
+  var loggedEvent: SessionStartEvent?
+  var didLog = false
+  var result: Result<Void, Error> = .success(())
+
+  func logEvent(event: SessionStartEvent, completion: @escaping (Result<Void, Error>) -> Void) {
+    loggedEvent = event
+    didLog = true
```
visumickey, line 27: What is the purpose of didLog here?

visumickey (approved):

FirebaseSessions/Tests/Unit/IdentifiersTests.swift
```diff
@@ -51,23 +51,24 @@ class IdentifiersTests: XCTestCase {
...
   }
 
+  /// Ensures that generating a Session ID multiple times results in the last Session ID being set in the previousSessionID field
   func testRotateSessionID() throws {
     identifiers.generateNewSessionID()
 
     let firstSessionID = identifiers.sessionID
     assert(isValidSessionID(identifiers.sessionID))
-    assert(identifiers.lastSessionID.count == 0)
+    assert(identifiers.previousSessionID.count == 0)
```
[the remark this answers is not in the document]
samedson: This was an oversight - XCTAssert is better

FirebaseSessions/Sources/EventGDTLogger.swift
samedson: Will that work for frameworks?

FirebaseSessions/Sources/GoogleDataTransport+GoogleDataTransportProtocol.swift
samedson: That's much better

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -38,9 +42,18 @@ protocol SessionsProvider {
 
   // MARK: - Initializers
 
+  // Initializes the SDK and top-level classes
   required convenience init(appID: String, installations: InstallationsProtocol) {
+    let googleDataTransport = GDTCORTransport(
```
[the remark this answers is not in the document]
samedson: Yeah so we talked offline - the reason for this is to test it against a mock GDT

FirebaseSessions/Sources/Identifiers.swift
```diff
@@ -50,6 +51,8 @@ class Identifiers: IdentifierProvider {
     uuid = UUID()
```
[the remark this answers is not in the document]
samedson: I need to fix this class so that it doesn't save every ID

FirebaseSessions/Tests/Unit/Mocks/MockGDTLogger.swift
samedson: The purpose is to check in the test. I guess we can use loggedEvent...

FirebaseSessions/Sources/EventGDTLogger.swift
ncooke3: I'm sorry, that was an objc macro 😅 . `targetEnvironment(simulator)` should do the trick and should work for frameworks.

```suggestion
    #if targetEnvironment(simulator)
      Logger.logDebug("Logging events using fast QOS due to running on a simulator")
      gdtEvent.qosTier = GDTCOREventQoS.qoSFast
    #endif  // targetEnvironment(simulator)
```

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -38,9 +42,18 @@ protocol SessionsProvider {
...
+      mappingID: GoogleDataTransportConfig.sessionsLogSource,
+      transformers: nil,
+      target: GoogleDataTransportConfig.sessionsTarget
+    )
+
+    let fireLogger = EventGDTLogger(googleDataTransport: googleDataTransport!)
+
     let identifiers = Identifiers(installations: installations)
-    let coordinator = SessionCoordinator(identifiers: identifiers)
+    let coordinator = SessionCoordinator(identifiers: identifiers, fireLogger: fireLogger)
```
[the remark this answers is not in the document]
samedson: Similar thing - it's there to pass in and test.

samedson:
Gonna submit this to unblock Leo. I'm gonna follow up with a refactor to the Identifiers class

PR #10314, opened 2022-10-06
Investigate watchOS CI

paulb777:
Fix SPM watchOS tests for Xcode 14 CI

The `pod lib lint` watchOS tests remain broken. Details with `--verbose` here https://github.com/firebase/firebase-ios-sdk/actions/runs/3200880464/jobs/5228288507

#no-changelog

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -166,7 +166,7 @@ tvos_flags=(
   -destination 'platform=tvOS Simulator,name=Apple TV'
 )
 watchos_flags=(
-  -destination 'platform=WatchOS Simulator,name=Apple Watch Series 7 - 45mm'
+  -destination 'platform=WatchOS Simulator,name=Apple Watch Series 7 (45mm)'
 )
 catalyst_flags=(
   ARCHS=x86_64 VALID_ARCHS=x86_64 SUPPORTS_MACCATALYST=YES -sdk macosx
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -166,7 +166,7 @@ tvos_flags=(
   -destination 'platform=tvOS Simulator,name=Apple TV'
 )
 watchos_flags=(
-  -destination 'platform=WatchOS Simulator,name=Apple Watch Series 7 - 45mm'
+  -destination 'platform=WatchOS Simulator,name=Apple Watch Series 7 (45mm)'
```
ncooke3, line 169: very sneaky but it looks like it's also now "**w**atchOS Simulator". 
```diff
- platform=WatchOS Simulator
+ platform=watchOS Simulator
```

paulb777: Fixed. It does seem case-insensitive though since the change was sufficient to start passing the spm watchOS tests

paulb777:
This PR fixes the SPM watchOS tests on Xcode 14.

`--verbose` shows there is still an unresolved problem with `pod lib lint` tests.  See https://github.com/firebase/firebase-ios-sdk/actions/runs/3200880464/jobs/5228288507

A follow up PR should either investigate those further or disable them for the near term.

ncooke3 (approved):

Issue #10147, opened 2022-08-27
New Watch App project configuration from Xcode14 does not receive push notifications using Firebase Cloud Messaging


AkiraHayakawaJMAS:
### Description

Xcode14 also eliminates the WatchKit Extension and allows everything to be integrated into the WatchKit App, as shown in the following Apple document.
https://developer.apple.com/documentation/watchkit/wkapplication

This will also affect the processing of bundle identifiers, and there will be cases where the bundle identifier that can be retrieved programmatically will no longer have the watchkitextension.

Example: [[NSBundle mainBundle] bundleIdentifier]
Xcode13:com.sample.bundle.app.watchkitapp.watchkitextension
Xcode14:com.sample.bundle.app.watchkitapp

There is some functions in the Firebase Cloud Messaging framework that falls under this issue. It is FIRMessagingAppIdentifier and FIRMessagingBundleIDByRemovingLastPartFrom.
This functions handles removing the last element of the bundleID. (I understand the purpose is to remove ".watchkitextension").
As noted in the source code comments, this is done in the watchKitExtension, so the
This processing is performed to obtain the bundleId of watchKitApp.

However, starting with Xcode14, there will be cases where this process itself will no longer be necessary.
If you adopt the new WatchApp project structure from Xcode14, you will need to use
Messaging may not be able to send push notifications to the Apple Watch app.

We would appreciate your consideration to address this issue as soon as possible.

### Reproducing the issue

Create a new Watch App project in Xcode14.
Develop a Watch App to use and receive Firebadse Cloud messaging.
Attempt to send messages through Firebase console or API, etc.

### Firebase SDK Version

9.5.0

### Xcode Version

14.0

### Installation Method

Swift Package Manager

### Firebase Product(s)

Messaging

### Targeted Platforms

watchOS

### If using Swift Package Manager, the project's Package.resolved

<details>
<summary>Expand <code>Package.resolved</code> snippet</summary>
<br>

```json

Replace this line with the contents of your Package.resolved.

```

</details>

### If using CocoaPods, the project's Podfile.lock

<details>
<summary>Expand <code>Podfile.lock</code> snippet</summary>
<br>

```yml

Replace this line with the contents of your Podfile.lock!

```

</details>

aashishpatil-g:
This should be fixed with https://github.com/firebase/firebase-ios-sdk/pull/10324.

PR #10324, opened 2022-10-07
[Messaging] Xcode 14 watchkit extension bundle id fix

aashishpatil-g:
### Discussion

Fix for https://github.com/firebase/firebase-ios-sdk/issues/10147

In versions of Xcode prior to Xcode 14, whenever a watch app was created, Xcode always created two targets 
Watch App (contained assets for the app but no code)
Watch Kit Extension (contained the code / logic for the app)

Xcode 14 onwards, no watch kit extension target is created for watchOS apps. There is a consolidated Watch App target that contains both code and assets. 

Firebase Messaging code used to strip out the watch extension suffix and use the watch app bundle ID for FCM registration. Since there is no extension anymore, the code is inadvertently stripping out the last component of the app bundle identifier itself, resulting in an incorrect bundle ID being used for FCM registration. 

This change detects if code is running in an extension by looking for the watch kit NSExtensionPointIdentifier in the Info.plist before stripping out the last component of the bundle identifier. 

### Testing

- Created a sample app with watch kit extension on Xcode 13 and confirmed that function returned YES (true) for extension. 
- Created a sample watch app (with no extension) on Xcode 14 and confirmed that function returned NO.

### API Changes
None

### DIFF
--- a/FirebaseMessaging/Sources/FIRMessagingUtilities.m
+++ b/FirebaseMessaging/Sources/FIRMessagingUtilities.m
@@ -27,6 +27,8 @@ NSString *const kFIRMessagingInstanceIDUserDefaultsKeyLocale =
 static NSString *const kFIRMessagingAPNSSandboxPrefix = @"s_";
 static NSString *const kFIRMessagingAPNSProdPrefix = @"p_";
 
+static NSString *const kFIRMessagingWatchKitExtensionPoint = @"com.apple.watchkit";
+
 #if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH
 static NSString *const kEntitlementsAPSEnvironmentKey = @"Entitlements.aps-environment";
 #else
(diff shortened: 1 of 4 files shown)


paulb777 (approved):
Thanks!

FirebaseMessaging/Sources/FIRMessagingUtilities.m
```diff
@@ -89,6 +95,25 @@ int64_t FIRMessagingCurrentTimestampInMilliseconds(void) {
   return [FIROptions defaultOptions].googleAppID;
 }
 
+BOOL FIRMessagingIsWatchKitExtension(void) {
```
chliangGoogle, line 98: would this be able to differentiate from old version because it would be nice to keep the older version working as well.

Also just checking to see if you have tested on real watch yet? If not, I can help test it on my device.

chliangGoogle (changes requested):

FirebaseMessaging/Sources/FIRMessagingUtilities.m
```diff
@@ -76,10 +78,14 @@ int64_t FIRMessagingCurrentTimestampInMilliseconds(void) {
 NSString *FIRMessagingAppIdentifier(void) {
   NSString *bundleID = [[NSBundle mainBundle] bundleIdentifier];
 #if TARGET_OS_WATCH
-  // The code is running in watchKit extension target but the actually bundleID is in the watchKit
-  // target. So we need to remove the last part of the bundle ID in watchKit extension to match
-  // the one in watchKit target.
-  return FIRMessagingBundleIDByRemovingLastPartFrom(bundleID);
+  if (FIRMessagingIsWatchKitExtension()) {
```
chliangGoogle, line 81: we should keep this logic for older xcode/ios version

chliangGoogle (commented):
Pending approval on testing on a real device.

Also please address the comment for older version for backward compatibility.

chliangGoogle (approved):

aashishpatil-g:
Tested on device for both old and new scenarios and confirmed that logic works.

PR #10325, opened 2022-10-07
Try fixing watchOS tests

morganchen12:


morganchen12:
Creating a new simulator with the old name fixes the SPM tests, but not pod lib lint:

https://github.com/firebase/firebase-ios-sdk/actions/runs/3181200877/jobs/5185674943
https://github.com/firebase/firebase-ios-sdk/actions/runs/3207972174/jobs/5243403550

paulb777:
The SPM tests were fixed by #10314

PR #10301, opened 2022-10-03
Fix missing aggregate query headers in umbrella

morganchen12:
#no-changelog

PR #10338, opened 2022-10-11
Add new danger check

morganchen12:
Prevents regression of #10301.

Note that this PR doesn't check which umbrella header is modified, so a change that updates multiple SDKs could still not warn if not all of the requisite umbrella headers were updated.


paulb777 (approved):
Thanks!

PR #10351, opened 2022-10-13
[Sessions] Add Sessions SDK nanopb and proto

samedson:
Adds just the proto and proto setup. SDK integration will be in a future PR


FirebaseSessions/Protogen/nanopb/clientanalytics.nanopb.h
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright 2019 Google
```
ncooke3, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

FirebaseSessions/Protogen/nanopb/sessions.nanopb.c
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright 2019 Google
```
ncooke3, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

FirebaseSessions/Protogen/nanopb/sessions.nanopb.h
```diff
@@ -0,0 +1,185 @@
+/*
+ * Copyright 2019 Google
```
ncooke3, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

FirebaseSessions/ProtoSupport/Protos/clientanalytics.proto
```diff
@@ -0,0 +1,67 @@
+
+syntax = "proto2";
+
+// This is in a separate file so it can be used as proto2.
```
ncooke3, line 4: Question (I didn't look too hard) why is the clientanalytics proto included? Who will use it? I figured the sessions proto will be a field on it but I'm not seeing a field below where that would be the case.

FirebaseSessions/ProtoSupport/nanopb_objc_generator.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
ncooke3: Nvm, I see your answer in the description.

ncooke3:
Please fix the `check` workflow as well!

FirebaseSessions/ProtoSupport/Protos/sessions.proto
```diff
@@ -0,0 +1,160 @@
...
+  string sdk_version = 2;
+  // Describes the network connectivity of the device
+  NetworkConnectionInfo network_connection_info = 3;
+  // Collects the OS running. Eg. iOS, iPadOs, tvOS, etc.
+  OsName os_name = 4;
+}
+
+// The name of the operating system on which the application is running.
+// Currently Apple only.
+enum OsName {
```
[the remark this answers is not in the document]
samedson: We don't currently have plans to split out Catalyst and Mac. These are shown as filters in the Crashlytics dashboard, and we felt that you as a developer would know whether your app was a macOS app or an iOS app running on mac (because by bundle ID, your app can only be one or the other IIRC).

Screen Shot 2022-10-13 at 4 07 29 PM
54e-4c9b-87c4-9afc9c1d9cd1.png)

FirebaseSessions/ProtoSupport/Protos/clientanalytics.proto
samedson: We use the `NetworkConnectionInfo` in the Session proto. Nanopb does not like the fully qualified namespace so that may have made it less clear to the eye

samedson:
I wonder if we can skip some import checks: 

```
+ /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/scripts/check_imports.swift
Import Error: /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/FirebaseSessions/Protogen/nanopb/sessions.nanopb.c:20 Import sessions.nanopb.h does not exist.
Import Error: /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/FirebaseSessions/Protogen/nanopb/clientanalytics.nanopb.c:20 Import clientanalytics.nanopb.h does not exist.
Import Error: /Users/runner/work/firebase-ios-sdk/firebase-ios-sdk/FirebaseSessions/Protogen/nanopb/sessions.nanopb.h:24 Import clientanalytics.nanopb.h does not exist.
```

FirebaseSessions/ProtoSupport/Protos/clientanalytics.options
```diff
@@ -0,0 +1,23 @@
+# Copyright 2022 Google LLC
```
visumickey, line 1: Do we need this file? All the lines seem to be commented out

FirebaseSessions/ProtoSupport/Protos/clientanalytics.proto
visumickey: This does not need to be in a separate proto file. In firebase performance, we packed these enums to the same parent proto to avoid referencing multiple proto files. Can we do something similar here?

https://github.com/firebase/firebase-ios-sdk/tree/master/FirebasePerformance/ProtoSupport/Protos

https://github.com/firebase/firebase-ios-sdk/blob/master/FirebasePerformance/ProtoSupport/Protos/perf_metric.proto#L387-L445

FirebaseSessions/ProtoSupport/Protos/sessions.proto
```diff
@@ -0,0 +1,160 @@
...
+// 3. Remove all the datapol.semantic_type tags
+// 4. Remove "wireless_android_play_playlog." from
+//    "wireless_android_play_playlog.NetworkConnectionInfo"
+// 5. Any new proto fields of type string, repeated, or bytes must be specified
+//    in the sessions.options file as "type:FT_POINTER"
+//
+
+syntax = "proto3";
+
+import "clientanalytics.proto";
```
visumickey, line 30: This can be avoided by including the definitions here.

FirebaseSessions/ProtoSupport/nanopb_objc_generator.py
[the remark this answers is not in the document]
visumickey: Unfortunately this script is a repeat for every product.

@ncooke3 Is there a way we can abstract this script to the top and make it usable across different products that need it? Having this copied over across products seem to be an overkill.

FirebaseSessions/ProtoSupport/Protos/clientanalytics.proto
samedson: This file is proto2, our main proto is proto3. This file won't build as proto3 because it has non-zero defaults (eg. NetworkType.NONE has a value of -1). We might be able to combine these files if we convert our sessions.proto to proto2, but that seems risky if our backend is proto3, and it will be confusing for updating it (because you'll have to remember to swap it from proto3 to proto2).

Fireperf is a different case cause the main proto is proto2

FirebaseSessions/ProtoSupport/Protos/sessions.proto
samedson: See my comment on proto2 vs 3

FirebaseSessions/ProtoSupport/Protos/sessions.proto
```diff
@@ -0,0 +1,160 @@
...
+
+package firebase.appquality;
```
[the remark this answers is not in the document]
samedson: We use `package firebase.appquality.sessions;` in G3 - I think the super long one is the Java namespace. I can update it to match the package, but if we use the Java one it's going to the generated code have super long prefixes

FirebaseSessions/ProtoSupport/Protos/clientanalytics.options
samedson: Looks like it'll still generate without it

FirebaseSessions/ProtoSupport/nanopb_objc_generator.py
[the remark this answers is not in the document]
ncooke3: It should be fine to abstract it to the top based on its current usage. I can look into it after this PR is merged.

samedson:
Let me know if there's any more changes needed on this

scripts/check_imports.swift
```diff
@@ -34,6 +34,7 @@ let skipDirPatterns = ["/Sample/", "/Pods/",
   [
     "CoreOnly/Sources", // Skip Firebase.h.
     "SwiftPMTests", // The SwiftPM imports test module imports.
+    "FirebaseSessions/Protogen/", // Generated nanopb code with imports
```
ncooke3, line 37: @paulb777, is there a way to do this without excluding from the `check_import` check?

ncooke3 (approved):

PR #10355, opened 2022-10-14
Revert "Add new danger check (#10338)"

morganchen12:
The change does not correctly flag additions to the umbrella header.


cherylEnkidu (approved):
Thank you!

PR #10360, opened 2022-10-18
`SessionsInitiator` triggers new session for every foregrounding after 30 minutes of background

leotianlizhan:
b/254304035
This change makes `SessionsInitiator` call the callback passed in whenever going from background to foreground after 30 minutes. WatchOS only works on 7.0+ because old versions do not support lifecycle events.

## Sidenote: dependency inject a "provider" function
For dependency injection, instead of using a `Protocol`, I end up just using a function reference inspired by:

- https://betterprogramming.pub/use-functions-for-dependency-injection-in-swift-a885f38d8ed0
- https://www.swiftbysundell.com/articles/time-traveling-in-swift-unit-tests/

This is because, our use-case matches the 1st article's point: all we are injecting is a "Provider" object, that has a single `func provide() -> Object`. TLDR of 1st article: injecting a function reference has many benefits: **less code = no longer need a brand new `DateProvider` file**, and test mocks just look like inline functions! I thought this was very cool and efficient.

However, after writing tests and realizing I can't just use an inline function, I still need a new `ShadowDate` object to manipulate time in tests, I start doubting back and forth wondering if this is one of those cases where `Protocol` is a better fit. Let me know what you think @samedson


FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -22,8 +26,43 @@ import Foundation
...
+    #endif
+  }
+  
+  @objc func appBackgrounded() {
+    print("hello")
+    backgroundTime = now()
+  }
+  
+  @objc func appForegrounded() {
+    print("foregrounded")
```
samedson, line 62: Can leave out the prints here and in backgrounded

FirebaseSessions/Tests/Unit/Mocks/ShadowDate.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  ShadowDate.swift
+//  Pods
+//
+//  Created by Leo Zhan on 2022-10-17.
+//
```
samedson, line 6: We'll wanna put the Google license header here. You can configure Xcode to add it automatically by adding a file to `/Users/<YOURUSER>/Library/Developer/Xcode/UserData`. I'll send you the file in chat

FirebaseSessions/Tests/Unit/Mocks/ShadowDate.swift
```diff
@@ -0,0 +1,21 @@
...
+
+import Foundation
+
+// A paused Date that can only be advanced through functions
+class ShadowDate {
```
samedson, line 11: I'm curious why Shadow- instead of Mock-? As far as I know about the terms, Shadow inspects the inside of the class vs mock just mocks functionality.

FirebaseSessions/Tests/Unit/InitiatorTests.swift
```diff
@@ -24,4 +24,26 @@ class InitiatorTests: XCTestCase {
...
+    initiator.appBackgrounded()
+    pausedClock.advance(by: 60 * 30 + 1)
+    initiator.appForegrounded()
+    assert(sessionCount == 2)
+    
+    // Simulate only 30 minutes of backgrounding
+    initiator.appBackgrounded()
+    pausedClock.advance(by: 60 * 30)
+    initiator.appForegrounded()
+    assert(sessionCount == 2)
```
samedson, line 47: Could we add a comment explaining that the sessioncount should not increase because we incremented the time by less than the session background timeout?

FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -22,8 +26,43 @@ import Foundation
 ///      and comes to the foreground.
 ///
 class SessionInitiator {
+  let sessionTimeout: Double = 1800
```
samedson, line 29: Can write this as `= 30 * 60 // 30 minutes` to increase clarity

FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -22,8 +26,43 @@ import Foundation
...
+  let now: () -> Date
```
samedson, line 30: I might rename `now` to something more function-ey. Maybe `getNow`?

FirebaseSessions/Sources/SessionInitiator.swift
visumickey: Also, can we replace Double with time measurements? Eg: TimeInterval rather than a Double. Though technically both are same, calling it TimeInterval will enforce the context more clear.

FirebaseSessions/Sources/SessionInitiator.swift
visumickey: I would make it more clearer and unambiguous - Can we call it `getCurrentTime`?

visumickey (commented):
Will add more thoughts once the PR gets towards a completion phase.

FirebaseSessions/Tests/Unit/InitiatorTests.swift
leotianlizhan: Done

FirebaseSessions/Sources/SessionInitiator.swift
leotianlizhan: Done

FirebaseSessions/Sources/SessionInitiator.swift
leotianlizhan: Done

FirebaseSessions/Tests/Unit/Mocks/ShadowDate.swift
leotianlizhan: Done, changed to Mock-

ncooke3:
If it helps, we have some precedent for doing this in the Core SDK. You can see an example unit test [here](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Internal/Tests/Unit/HeartbeatControllerTests.swift#L62-L109). Because the provider block is `@escaping`, you actually have the reference semantics that make needing `MockDate` unnecessary. 

Additionally, I personally like referring to `() -> ?`-typed property's as `somethingProvider`. So, for example, `currentTimeProvider` that is then called like: `let currentTime = currentTimeProvider()`. This is just my opinion though.

leotianlizhan:
> If it helps, we have some precedent for doing this in the Core SDK. You can see an example unit test [here](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Internal/Tests/Unit/HeartbeatControllerTests.swift#L62-L109). Because the provider block is `@escaping`, you actually have the reference semantics that make needing `MockDate` unnecessary.
> 
> Additionally, I personally like referring to `() -> ?`-typed property's as `somethingProvider`. So, for example, `currentTimeProvider` that is then called like: `let currentTime = currentTimeProvider()`. This is just my opinion though.

Thanks for the code reference, I'm gonna steal that! Do you recommend then to inject a "provider function" rather than a separate "provider class"? @ncooke3

FirebaseSessions/Sources/SessionInitiator.swift
leotianlizhan: How do y'all feel about @ncooke3 's comment of just something like `currentTimeProvider` or `dateProvider`? This would be consistent with iCore's style like this iCore class [HeartbeatController](https://github.com/firebase/firebase-ios-sdk/blob/f2e44defd3886a735199f1ec6768377adbaa5a4d/FirebaseCore/Internal/Sources/HeartbeatLogging/HeartbeatController.swift#L53:L64)

FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -22,8 +28,72 @@ import Foundation
 ///      and comes to the foreground.
 ///
 class SessionInitiator {
+  let sessionTimeout: TimeInterval = 30 * 60 // 30 minutes
+  let currentTime: () -> Date
+  var backgroundTime = Date.distantFuture
+  var initiateSessionStart: () -> Void = {}
+
+  init(dateProvider: @escaping () -> Date = Date.init) {
```
samedson, line 36: Slight nit - what do you think of calling this currentDateProvider to convey it should represent the current time?

FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -22,8 +28,72 @@ import Foundation
...
+        object: nil
+      )
+      notificationCenter.addObserver(
+        self,
+        selector: #selector(appForegrounded),
+        name: NSApplication.didBecomeActiveNotification,
+        object: nil
+      )
+    #elseif os(watchOS)
+      if #available(watchOSApplicationExtension 7.0, *) {
```
samedson, line 72: Can we add a comment here indicating the current situation? Eg. "watchOS versions less than 7.0 don't support lifecycle events"

leotianlizhan: Done

PR #10376, opened 2022-10-19
[Sessions] Add ApplicationInfo for filling in the proto

samedson:
Adds ApplicationInfo class and fills in parts of ApplicationInfo in the proto

A few fields remain because they need to be moved into a centralized place. These will likely be moved to GoogleUtilities so that Crash, Perf, and Sessions can all share the same code. For example, we moved osName to GoogleUtilities in this PR: https://github.com/google/GoogleUtilities/pull/89

We will attempt to move the following to GoogleUtilities:
 - device_model
 - network_connection_info
 - mcc_mnc - I did fill this in because the code is small, but I will attempt to move it to GoogleUtilities as well

For development_platform_name/version, I had originally thought we would need to fetch these from the Crashlytics SDK because they are set via an API and not read from the environment. This does come with a big problem though, mainly that if just FirePerf is installed, then development_platform won't be correct. We may need to come up with an approach that both Sessions and Crashlytics can duplicate.

 - Added taskflow items for these

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -39,6 +39,7 @@ protocol SessionsProvider {
   private let coordinator: SessionCoordinator
   private let initiator: SessionInitiator
   private let identifiers: Identifiers
+  private let appInfo: ApplicationInfo
 
   // MARK: - Initializers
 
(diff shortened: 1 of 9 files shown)


FirebaseSessions/Tests/Unit/Mocks/MockApplicationInfo.swift
```diff
@@ -26,15 +26,19 @@ class MockApplicationInfo: ApplicationInfoProtocol {
 
   var osName: String = ""
 
+  var mccMNC: String = ""
+
   static let testAppID = "testAppID"
   static let testBundleID = "testBundleID"
   static let testSDKVersion = "testSDKVersion"
   static let testOSName = "ios"
+  static let testMCCMNC = "testMCCMNC"
```
visumickey, line 35: Can we do this as the proto structure like the NetworkConnectionInfo and see how this gets converted to a string as we want?

visumickey (approved):

FirebaseSessions/Tests/Unit/Mocks/MockApplicationInfo.swift
samedson: Let me follow up with this and see if it makes sense to you

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -55,27 +56,30 @@ protocol SessionsProvider {
...
     let coordinator = SessionCoordinator(identifiers: identifiers, fireLogger: fireLogger)
     let initiator = SessionInitiator()
+    let appInfo = ApplicationInfo(appID: appID)
 
     self.init(appID: appID,
               identifiers: identifiers,
               coordinator: coordinator,
-              initiator: initiator)
+              initiator: initiator,
+              appInfo: appInfo)
```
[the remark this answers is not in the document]
samedson: It's not strictly necessary, but Nick has been recommending this so we can test FirebaseSessions and mock out the parameters like `appInfo`.

Issue #10374, opened 2022-10-19
A Few FirebaseStorage bugs on iOS SDK 10


russellwheatley:
### Description

I've created a repository with a full reproduction of the issues we experienced trying to integrate FlutterFire with iOS SDK 10. Please see issue reproduction details for the bugs we found. The downloadURL issue may actually be a problem with firebase-tools, not sure.

### Reproducing the issue

Here's the repository which illustrates the issues: https://github.com/invertase/ios-sdk-10-issues

Ripped from the README.md of that repository:

To reproduce and what to expect:

1. Clone repo and pod install. Open xcworkspace in Xcode.
2. Run Storage emulator and allow all read/write operations in Storage rules.
3. Run the app.
4. Press the "downloadUrl emulator failure" button to see the error produced. The response includes a metadata object not stringified which I think is the problem.
Comment out the [emulator initialization](https://github.com/invertase/ios-sdk-10-issues/blob/main/storage-issues/AppDelegate.m#L23) and run app again to see a successful response. It does not contain a 
metadata object.
5. Press the "putFile failure" button to see this error: "Fatal error: Internal error enqueueing a Storage task". It's because path does not exist on
metadata which is a read only property. It can be worked around using [intializeWithDictionary](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/StorageMetadata.swift#L156) and adding "name" property as it puts "name" on "path" property [here](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/StorageMetadata.swift#L174)
6. Press the "updateMetadata failure" button to see response does not include updated custom metadata (i.e. "foo":"bar" in this example). If you comment out
the way FIRStorageMetadata is initialised [here](https://github.com/invertase/ios-sdk-10-issues/blob/main/storage-issues/ViewController.m#L98-L99) and 
uncomment [here](https://github.com/invertase/ios-sdk-10-issues/blob/main/storage-issues/ViewController.m#L102-L103), you will see custom metadata in response.

I have more detailed notes on the PR that landed for FlutterFire [here](https://github.com/firebase/flutterfire/pull/9708#issue-1404360916)

### Firebase SDK Version

10.0.0

### Xcode Version

14.0.1

### Installation Method

CocoaPods

### Firebase Product(s)

Storage

### Targeted Platforms

iOS

### If using Swift Package Manager, the project's Package.resolved

<details>
[...]

PR #10385, opened 2022-10-20
[Storage] Debug updateMetadata customMetadata iniWithDictionary issue

andrewheard:
Debugging the issue `updateMetadata` method fails to update `customMetadata` when you initialize metadata with `initWithDictionary`. See #10374 issue 6 for further details.

### DIFF
--- a/FirebaseStorage/Sources/StorageMetadata.swift
+++ b/FirebaseStorage/Sources/StorageMetadata.swift
@@ -146,6 +146,7 @@ import Foundation
   // MARK: - Public Initializers
 
   @objc override public convenience init() {
+    // Approach 2: `initialMetadata` is `[:]`
     self.init(dictionary: [:])
   }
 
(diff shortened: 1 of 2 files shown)


andrewheard:
Nick found some important details:
> Debugging further, I isolated the difference in behavior through this call stack. The `StorageMetadata.remove` API functions differently depending on the initialization pattern because when you use initWithDictionary, both the `matchingMetadata` and `oldMetadata` are the same and the `StorageMetadata.remove` cancels them out (removing the custom metadata).

andrewheard:
I added some inline comments to the files for debugging. @paulb777 made a good suggestion to check if the issue existed in the old implementation. A quick look at [FIRStorageMetadata.m](https://github.com/firebase/firebase-ios-sdk/blob/111d8d6ad1a1afd6c8e9561d26e55ab1e74fcb42/FirebaseStorage/Sources/FIRStorageMetadata.m#L169-L195) makes me think that it would but I haven't tested to verify.

paulb777:
Both Firebase 8 and 9 fail the test in the same way.

paulb777:
Closing. See the explanation in #10374

PR #10386, opened 2022-10-20
[Sessions] Update Sessions proto with newest changes

samedson:
Updates the proto to match the backend

#no-changelog

### DIFF
--- a/FirebaseSessions/ProtoSupport/Protos/sessions.proto
+++ b/FirebaseSessions/ProtoSupport/Protos/sessions.proto
@@ -23,6 +23,7 @@
 //    "wireless_android_play_playlog.NetworkConnectionInfo"
 // 5. Any new proto fields of type string, repeated, or bytes must be specified
 //    in the sessions.options file as "type:FT_POINTER"
+// 6. Remove any Lint comments (eg. IfChanged lines)
 //
 
 syntax = "proto3";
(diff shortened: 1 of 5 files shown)


leotianlizhan (approved):

FirebaseSessions/ProtoSupport/Protos/sessions.proto
```diff
@@ -120,11 +131,10 @@ message ApplicationInfo {
...
 // - network_connection_info
+// Next tag: 4
 message AndroidApplicationInfo {
   // The applications package name. Eg. com.google.search
   string package_name = 1;
-  // The version of the Android Firebase-Sessions SDK in use
-  string sdk_version = 2;
   // The version name as specified in the app's manifest
   // https://developer.android.com/guide/topics/manifest/manifest-element#vname
   string version_name = 3;
```
leotianlizhan, line 140: Curious: are these supposed to skip tags? e.x. going from 1 to 3, skipping 2.

samedson: Yeah so if you look at Raymond's backend PR he removed that field, but proto has strong recommendations to never re-use tags (because essentially the fields are just names, and the tags actually identify the data over the wire).

So for example if we were to use tag 2 for version_name to reuse sdk_version's tag then if we had a client with the old proto we would be setting sdk_version as version_name (because that client would have sdk_version = 2; in its proto).

PR #10391, opened 2022-10-21
Sync spec tests from web SDK to iOS SDK

milaGGL:


Firestore/Example/Tests/SpecTests/json/bundle_spec_test.json
```diff
@@ -3,7 +3,8 @@
     "describeName": "Bundles:",
     "itName": "Bundles query can be loaded and resumed from different tabs",
     "tags": [
-      "multi-client"
+      "multi-client",
+      "no-ios"
```
milaGGL, line 7: It has no-ios tag, but still get parsed and extracted to ios sdk 🤔

dconeybe: It will get extracted, but should not get executed by the ios spec test runner.

dconeybe (approved):
Thanks for syncing up the spec tests!

PR #10396, opened 2022-10-24
[Sessions] Add Test App and Integration tests for the Sessions SDK

leotianlizhan:
b/253049457

Screen Shot 2022-11-01 at 9 48 58 AM

#no-changelog


FirebaseSessions/Tests/TestApp/Podfile
```diff
@@ -0,0 +1,28 @@
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
+
+def shared_pods
+  pod 'FirebaseCore', :path => '../../../'
+  pod 'FirebaseInstallations', :path => '../../../'
+  pod 'FirebaseCoreInternal', :path => '../../../'
```
visumickey, line 7: Curious if we need these internal pods? If we don't use (which I presume it is), can we remove them from the list of dependencies?

visumickey: Infact, I'm thinking if we can depend just on FirebaseSessions here and all the other dependencies should naturally be brought in.

FirebaseSessions/Tests/TestApp/Podfile
```diff
@@ -0,0 +1,28 @@
...
+  pod 'FirebaseCoreExtension', :path => '../../../'
+  pod 'FirebaseSessions', :path => '../../../'
+end
+
+target 'AppQualityDevApp_iOS' do
+  platform :ios, '15.0'
```
visumickey, line 13: Should we allow for older versions of iOS at all? May be a good reference for minimum compatible version for us?

FirebaseSessions/Tests/TestApp/UITestsiOS/UITestsiOS.swift
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+
+final class UITestsiOS: XCTestCase {
+  override func setUpWithError() throws {
```
visumickey, line 19: Both the `setup` and `tearDown` methods can be cleaned up as we are not using them.

FirebaseSessions/Tests/TestApp/macOS/macOS.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
visumickey, line 1: Do we need to push in entitlements file to the repository?

FirebaseSessions/generate_testapp.sh
```diff
@@ -0,0 +1,28 @@
...
+readonly DIR="$( git rev-parse --show-toplevel )"
+
+#
+# Use the following command in this comment if you're iterating on the
+# podspec. This will prevent settings from changing when you reload the project.
+#
+# pod gen FirebaseSessions.podspec --local-sources=./ --auto-open --platforms=ios,macos,tvos
+#
+cd $DIR/FirebaseSessions/Tests/TestApp
+pod install
```
visumickey, line 27: Would we have steps to bring in the secret file `GoogleServiceInfo.plist` before opening up the project?

scripts/gha-encrypted/Sessions/GoogleService-Info_TestApp_autopush.plist.gpg
```diff
@@ -0,0 +1,6 @@
+�	�����q���:42G6�M��_VA��"7�]�pm6��]����0ʽc��Ð�7��Ɔ7
```
visumickey, line 1: Should the filename have a suffix of `autopush`?

FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import FirebaseCore
+import FirebaseSessions
```
visumickey, line 18: Is this needed? I meant the import of `FirebaseSessions`

FirebaseSessions/Tests/TestApp/Podfile
paulb777: You need all the recursive dependencies. Otherwise, you'll get the latest release versions instead of the ones from the repo.

visumickey: Aah! Got it. Makes sense. Its been a while, got it.

FirebaseSessions/Tests/TestApp/Podfile
samedson: That's a good idea - set it out our deployment target in the cocoapod

FirebaseSessions/generate_testapp.sh
samedson: So I'm not sure we're able to do this. The password is stored in GitHub actions and passed in as a variable to our yaml file https://github.com/firebase/firebase-ios-sdk/pull/10396/files#diff-09baab8602f4dabcc3ffaa06efbdf84abaecd1f2bed8523fe833b189308172a3R42

scripts/gha-encrypted/Sessions/GoogleService-Info_TestApp_autopush.plist.gpg
samedson: Yeah I was thinking this would be the autopush plist - do you think it should be something else?

FirebaseSessions/Tests/TestApp/macOS/macOS.entitlements
samedson: Yeah I think we need it - it's kindof an info.plist for the macOS app without having a separate plist for just macOS

visumickey: Got it. Thanks for clarifying.

FirebaseSessions/generate_testapp.sh
visumickey: @paulb777 Is there a way to make it happen in our shell scripts for the project generation?

scripts/gha-encrypted/Sessions/GoogleService-Info_TestApp_autopush.plist.gpg
visumickey: All the environments can have the same plist file. So, would recommend removing hte environment name in the suffix.

FirebaseSessions/generate_testapp.sh
paulb777: A Googler only script could copy it from its CitC location

samedson: Added a script
Screen Shot 2022-11-01 at 9 48 31 AM

visumickey: Awesome. This is great! Thanks for doing this @samedson . Thanks @paulb777 .

visumickey (approved):

PR #10398, opened 2022-10-24
Replace FilterList with std::vector<Filter> for future maintenance

cherylEnkidu:
#no-changelog

### DIFF
--- a/Firestore/core/test/unit/core/query_test.cc
+++ b/Firestore/core/test/unit/core/query_test.cc
@@ -20,6 +20,7 @@
 
 #include "Firestore/core/src/core/bound.h"
 #include "Firestore/core/src/core/field_filter.h"
+#include "Firestore/core/src/core/filter.h"
 #include "Firestore/core/src/model/document_set.h"
 #include "Firestore/core/src/model/field_path.h"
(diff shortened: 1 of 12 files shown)


Firestore/core/test/unit/core/query_test.cc
```diff
@@ -48,7 +49,6 @@ using testutil::CollectionGroupQuery;
 using testutil::DbId;
 using testutil::Doc;
 using testutil::Field;
-using testutil::Filter;
```
ehsannas, line 51: why did you remove this?

ehsannas (commented):
Generally LGTM. Not clear why the changes were made to query_test.cc (adding `testutil::` everywhere)?

Firestore/core/test/unit/core/query_test.cc
cherylEnkidu: Since we introduce `std::vector<Filter>` to this file. Here Filter refers to `core::Filter` (class name) instead of `testutil::Filter` (function name). With `using testutil::Filter;`, the compiler will not work.

ehsannas (approved):

PR #10401, opened 2022-10-25
Support GetFilters() in core::Filter

cherylEnkidu:
#no-changelog

### DIFF
--- a/Firestore/core/src/core/field_filter.cc
+++ b/Firestore/core/src/core/field_filter.cc
@@ -131,6 +131,12 @@ const std::vector<FieldFilter>& FieldFilter::Rep::GetFlattenedFilters() const {
   return Filter::Rep::memoized_flattened_filters_;
 }
 
+std::vector<Filter> FieldFilter::Rep::GetFilters() const {
+  // This is the only filter within this object, so we return a list of size
+  // one.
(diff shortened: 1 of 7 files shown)


Firestore/core/src/core/field_filter.cc
```diff
@@ -131,6 +131,12 @@ const std::vector<FieldFilter>& FieldFilter::Rep::GetFlattenedFilters() const {
   return Filter::Rep::memoized_flattened_filters_;
 }
 
+std::vector<Filter> FieldFilter::Rep::GetFilters() const {
+  // This is the only filter within this object, so we return a list of size
+  // one.
+  return std::vector<Filter>{FieldFilter(std::make_shared<const Rep>(*this))};
```
ehsannas, line 137: why not `return {*this}` ?

cherylEnkidu: `core::Filter` class itself is not an abstract class and doesn't contain any virtual functions. All the virtual functions are implemented through `Filter::Rep`. So `*this` is a dereference to `FieldFilter::Rep` instead of `FieldFilter`.

ehsannas: 👍

ehsannas (approved):

PR #10402, opened 2022-10-25
Replace OrderByList by std::vector<OrderBy>

cherylEnkidu:
#no-changelog


Firestore/core/test/unit/local/target_cache_test.cc
```diff
@@ -18,6 +18,7 @@
 
 #include <set>
 #include <string>
+#include <unordered_set>
```
ehsannas, line 21: is this needed?

Firestore/core/test/unit/core/query_test.cc
```diff
@@ -50,7 +50,6 @@ using testutil::DbId;
 using testutil::Doc;
 using testutil::Field;
 using testutil::Map;
-using testutil::OrderBy;
```
ehsannas, line 53: why is this being removed?

Firestore/core/test/unit/local/target_cache_test.cc
cherylEnkidu: There are use cases of `unordered_set` inside `target_cache_test.cc`. Not sure why the compiler doesn't complain before, but it complains in this pr without include statement.

ehsannas (approved):

PR #10403, opened 2022-10-25
Firestore: Fix data race in credentials provider

dconeybe:
Fixes: #10393


Firestore/core/test/unit/credentials/firebase_auth_credentials_provider_test.mm
```diff
@@ -153,6 +153,29 @@ - (void)getTokenForcingRefresh:(BOOL)forceRefresh
...
 
+// Catch race conditions in FirebaseAuthCredentialsProvider::GetToken() when
+// run under thread sanitizer.
+// See https://github.com/firebase/firebase-ios-sdk/issues/10393
+TEST(FirebaseAuthCredentialsProviderTest, GetTokenCalledByAnotherThread) {
+  FIRApp* app = testutil::AppForUnitTesting();
+  FSTAuthFake* auth = [[FSTAuthFake alloc] initWithToken:@"token for fake uid"
+                                                     uid:@"fake uid"];
+  FirebaseAuthCredentialsProvider credentials_provider(app, auth);
+  credentials_provider.InvalidateToken();
```
cherylEnkidu, line 164: Could you please tell me why we need to call InvalidateToken()?

Firestore/core/test/unit/credentials/firebase_auth_credentials_provider_test.mm
```diff
@@ -153,6 +153,29 @@ - (void)getTokenForcingRefresh:(BOOL)forceRefresh
...
+    credentials_provider.GetToken(
+        [](util::StatusOr<AuthToken> result) { EXPECT_TRUE(result.ok()); });
+  });
+  std::thread thread2([&credentials_provider] {
+    credentials_provider.GetToken(
+        [](util::StatusOr<AuthToken> result) { EXPECT_TRUE(result.ok()); });
+  });
+
+  thread1.join();
+  thread2.join();
```
cherylEnkidu, line 176: What would happen to this test if we used previous nonatomic implementation?

cherylEnkidu (commented):
Thanks for your fix! Just a few questions

Firestore/core/test/unit/credentials/firebase_auth_credentials_provider_test.mm
dconeybe: I honestly don't know. I just copied a test from above and tweaked it. If it blocks your code review, I can investigate further though.

cherylEnkidu: Thank you!

Firestore/core/test/unit/credentials/firebase_auth_credentials_provider_test.mm
dconeybe: Changing `atomic` back to `nonatomic` causes the Thread Sanitizer tests to fail with a race condition. I don't recall exactly what the race condition was though. So, I've pushed up a commit to this PR, https://github.com/firebase/firebase-ios-sdk/pull/10403/commits/e9cc0f9a6d7c32ab14ce6a9927b1a3e06ca4cc18, that changes `atomic` back to `nonatomic`. We'll see how/if this test fails then I'll reply back to this comment.

Firestore/core/test/unit/credentials/firebase_auth_credentials_provider_test.mm
dconeybe: Actually... you're right. I had copied the code above, which is specifically testing InvalidateToken(). It probably isn't relevant for this test. I'll remove it.

cherylEnkidu (approved):
LGTM!

tsahi-deri:
@dconeybe Hey, is this fix exist in release version 10.1.0?

paulb777:
It will go into 10.2.0. I updated the milestone on the issue.

PR #10406, opened 2022-10-25
Add Http status check to Realtime stream

qdpham13:
Checks HTTP status before retrying. Matching Android PR: https://github.com/firebase/firebase-android-sdk/pull/4121
- Http status codes being used to retry are:
Ok 200
Client Timeout 408
Bad Gateway 502
Unavailable 503
Gateway timeout 504
Changes timeout connection to 5 and a half minutes.


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -552,18 +562,30 @@ - (void)URLSession:(NSURLSession *)session
...
     didReceiveResponse:(NSURLResponse *)response
      completionHandler:(void (^)(NSURLSessionResponseDisposition disposition))completionHandler {
   _isRequestInProgress = false;
   NSHTTPURLResponse *_httpURLResponse = (NSHTTPURLResponse *)response;
-  if ([_httpURLResponse statusCode] != 200) {
+  NSInteger statusCode = [_httpURLResponse statusCode];
+  if (statusCode != kRCNFetchResponseHTTPStatusOk) {
     [self pauseRealtimeStream];
-    [self retryHTTPConnection];
+    if ([self isStatusCodeRetryable:statusCode]) {
```
karenyz, line 584: for cases where we don't retry, should we return an error to the client?

qdpham13: Yeah it'd make sense to return a StreamError if we can't retry, updated.

PR #10407, opened 2022-10-26
Fix a bug where the linking/reauth flows errors are not caught

renkelvin:


renkelvin:
Changelog: https://github.com/firebase/firebase-ios-sdk/pull/10408

paulb777 (approved):
Can you add a test?

rosalyntan (approved):

PR #10410, opened 2022-10-26
Performing IN expansion

cherylEnkidu:
#no-changelog

### DIFF
--- a/Firestore/core/src/local/leveldb_index_manager.cc
+++ b/Firestore/core/src/local/leveldb_index_manager.cc
@@ -536,14 +536,20 @@ IndexManager::IndexType LevelDbIndexManager::GetIndexType(
 
 absl::optional<std::vector<model::DocumentKey>>
 LevelDbIndexManager::GetDocumentsMatchingTarget(const core::Target& target) {
-  std::vector<DocumentKey> result;
-  std::unordered_set<std::string> existing_keys;
+  std::vector<std::pair<core::Target, model::FieldIndex>> indexes;
   for (const auto& sub_target : GetSubTargets(target)) {
     auto index_opt = GetFieldIndex(sub_target);
     if (!index_opt.has_value()) {
       return absl::nullopt;
     }
-    const FieldIndex& index = index_opt.value();
+    indexes.emplace_back(sub_target, index_opt.value());
+  }
+
+  std::vector<DocumentKey> result;
+  std::unordered_set<std::string> existing_keys;
+  for (const auto& entry : indexes) {
+    const Target& sub_target = entry.first;
+    const FieldIndex& index = entry.second;
 
     LOG_DEBUG("Using index %s to execute target %s", index.collection_group(),
               sub_target.CanonicalId());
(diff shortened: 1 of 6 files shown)


Firestore/core/src/local/leveldb_index_manager.cc
```diff
@@ -536,21 +536,14 @@ IndexManager::IndexType LevelDbIndexManager::GetIndexType(
...
-
-    indexes.insert({sub_target, index_opt.value()});
-  }
-
-  std::vector<DocumentKey> result;
-  std::unordered_set<std::string> existing_keys;
-  for (const auto& entry : indexes) {
-    const Target& sub_target = entry.first;
-    const FieldIndex& index = entry.second;
+    const FieldIndex& index = index_opt.value();
```
cherylEnkidu, line 546: In the preview implementation, the order of result is decided by the index order in `unordered_map` (which is unknown). I change this to match the behaviour of Android SDK, so the sequence will be the same as inside sub_targets.

Firestore/core/test/unit/local/leveldb_query_engine_test.cc
```diff
@@ -55,6 +54,15 @@ std::unique_ptr<Persistence> PersistenceFactory() {
...
 
+model::DocumentMap DocumentMap(
+    const std::vector<model::MutableDocument>& docs) {
+  model::DocumentMap doc_map;
+  for (const auto& doc : docs) {
+    doc_map = doc_map.insert(doc.key(), doc);
+  }
+  return doc_map;
+}
+
```
ehsannas, line 65: Nice :)

Firestore/core/src/local/leveldb_index_manager.cc
ehsannas: 👍

ehsannas, line 546: Thanks for your explanation of why you made this change. It's true that it's now a deterministic order, but I think this has a negative performance impact.

The code essentially wants to achieve this: "if any of the disjunction terms doesn't have a client-side index, we cannot serve the query using the client-side index". 

Now imagine that you have a query like
`a==1 || b==2 || c==3 || d==4 || e==5`. And that we _do_ have a client-side index for `a` and `b` and `c` and `d` (but we don't have a client-side index for `e`).

The code now uses the indexes to perform the first 4 queries (including all the work on line 551 to 581) using the client-side index, and then when it reaches `e`, it realizes `e` doesn't have a client-side index and it'll have to throw away all the work it did for the first 4 branches.

In other words, "early exit" (returning nullopt) used to happen much earlier in such cases.

ehsannas: Can we use a ordered_map instead?

ehsannas (commented):
This is great Cheryl. Just one concern below.

Firestore/core/src/local/leveldb_index_manager.cc
cherylEnkidu: Thank you for pointing that out! I totally agree with your points. However `ordered_map` will compare the value of  `Target` to decide the sequence. So I choose to use `vector<pair<..., ...>>` to keep the order return from `GetSubTargets`.

ehsannas: 👍

ehsannas (approved):
LGTM!

dconeybe:
Note: This PR introduced a subtle bug where a nested array or map in an "IN" condition causes a double-free bug and may crash the app. Googlers see b/260406277 for details.

Issue #10417, opened 2022-10-28
SwiftUI and Objective-C inter-op with Swizzling


fabrizioCorut:
### Description

First of all, let's define some terminology. I will refer to:
- a "SwiftUI app" an app that declare its entry point as being the implementation of the `SwiftUI.App` protocol, it being marked by the new `@main` annotation; This app can still have `AppDelegate` and `SceneDelegate` implementations, but those will not be the entry point of the application;

Using `FirebaseApp` in a "SwiftUI app" is very confusing and there is no guidance as to where to _correctly_ place the `FirebaseApp.configure()` method call and more importantly, **why** place it where it should be placed.

Some of my observations from placing the `FirebaseApp.configure()` call in different places:

### `FirebaseApp.configure()` in the `App's` `init` method

Works, but some of the components from `Firebase` are reliant on the `UIApplication.shared` instance, which is **not initialized** at the time of the call; ❌
Setup:
```swift
@main
struct FirebaseApp: App {
	
	init() {
		FirebaseApp.configure()
	}
		
	var body: some Scene {
		WindowGroup {
			ContentView()
		}
	}
}
```

Concrete **issue** example from `FIRAuth.protectedDataInitialization`: 
```Objective-C
UIApplication *application = [applicationClass sharedApplication];

if (application) {
	// Initialize for phone number auth.
	strongSelf->_tokenManager = [[FIRAuthAPNSTokenManager alloc] initWithApplication:application];

	strongSelf->_appCredentialManager =
		[[FIRAuthAppCredentialManager alloc] initWithKeychain:strongSelf->_keychainServices];

	strongSelf->_notificationManager = [[FIRAuthNotificationManager alloc]
		initWithApplication:application
	appCredentialManager:strongSelf->_appCredentialManager];
}
```
at this time and point, as far as I could investigate and test, the `application is NULL` and that doesn't seem to change during the lifetime of the app in this setup.
**Obs:** swizzling on/ off **does not solve** the behaviour described.

### `FirebaseApp.configure()` in `UIApplicationDelegate.application(_, didFinishLaunchingWithOptions:)` 

Solves the problem described above since at init time the `UIApplication.shared` exists. But this is where the discussion gets interesting.

- with swizzling **disabled** everything works as expected; ✅
[...]

PR #10418, opened 2022-10-28
Make App Distribution compatible with swizzling-disabled implementations

morganchen12:
Required to mitigate #10417.


paulb777 (approved):
Thanks!

PR #10423, opened 2022-10-31
[DNM] Experiment with style.sh

ncooke3:
#no-changelog


scripts/style.sh
```diff
@@ -156,7 +156,7 @@ s%^./%%
 \%^/?vendor/bundle/% d
 
 # Sources pulled in by the Mint package manager
-\%^Mint% d
+\%^mint% d
```
ncooke3, line 159: i wouldn't expect this to be case-sensitive but maybe it is?

paulb777:
Closing since #10425 landed

PR #10424, opened 2022-10-31
[v10.2.0] Update versions for Release 10.2.0

andrewheard:
Updated versions for the next release.

#no-changelog


paulb777 (approved):
The test failures look like an infra issue that may clear up after this merges.

andrewheard:
> The test failures look like an infra issue that may clear up after this merges.

Thanks for the heads up, @paulb777!

Issue #9968, opened 2022-06-30
FR: Support iOS 16 Simulator on Firebase Messaging


ispiropoulos:
### Feature proposal

* Firebase Component: Messaging

Trying to test remote notifications via Firebase Messaging on an Apple Silicon Mac with macOS Ventura and Xcode 14.
Getting this log message:

`2022-06-30 15:56:00.487199+0300 my-project[75680:3976914] 9.2.0 - [FirebaseMessaging][I-FCM023014] Running InstanceID on a simulator doesn't have APNS. Use prod profile by default.`

And notification sending fails with this error:
`"errorCode": "UNREGISTERED"`

The proposed change is to support iOS 16 simulators.

paulb777:
Thanks for the feature request.  From the Xcode 14 Release notes:

_Simulator now supports remote notifications in iOS 16 when running in macOS 13 on Mac computers with Apple silicon or T2 processors. Simulator supports the Apple Push Notification Service Sandbox environment. Your server can send a remote notification to your app running in that simulator by connecting to the APNS Sandbox (api.sandbox.push.apple.com). Each simulator generates registration tokens unique to the combination of that simulator and the Mac hardware it’s running on. See [User Notifications](https://developer.apple.com/documentation/usernotifications/) for more information._

_Remote Notifications support more features (like Notification Service Extensions) than locally simulated notifications using .apns payload files or the simctl push command._

_Device Registration Tokens are of variable length. Tokens in Simulator may be larger than current physical device tokens. Don’t hardcode any specific length or format for these tokens. (60974170_)

KoCMoHaBTa:
If you are interested about a temporary workaround, while we wait for this to be implemented, you can do the following:
1. Build and run your app on a device trough Xcode
2. Go to your app's DerivedData folder -> `Build/Products/Dev-iphoneos/{YOUR_APP_NAME}.app`
3. Copy the `embedded.mobileprovision` to your project and add it as a resource to the app - we need this, because the [code here](https://github.com/firebase/firebase-ios-sdk/blob/abf29403d99d907ab980036944c3a5208f0388a0/FirebaseMessaging/Sources/FIRMessagingUtilities.m#L319) is trying to identify the environment form the provisioning profile which defaults to production if no profile is found , but the Simulator supports only the sandbox environment.
4. Declare the following Objective-C Category in your code:

**GULAppEnvironmentUtil+SimulatorPN.h**
```
@import GoogleUtilities_Environment;
[...]

Issue #10214, opened 2022-09-14
Phone Auth not working in iOS 16 simulator.


Archetapp:
### Description

For our app, in order to login, we use Phone Authentication. When testing on any OS below iOS16 in the iOS Simulator, it works perfectly for both test numbers and live phone numbers.

But now with iOS16, it throws the following error:

"Error Domain=FIRAuthErrorDomain Code=17048 "Token mismatch" UserInfo={NSLocalizedDescription=Token mismatch, FIRAuthErrorUserInfoNameKey=INVALID_APP_CREDENTIAL}"

I've tried it in both Sandbox, Production & Unknown APNs environments, but nothing fixes it.

### Reproducing the issue

Submit a number to Phone Authentication 

```
 finalPhone = "+" + "12345678910"
        PhoneAuthProvider.provider(auth: Auth.auth()).verifyPhoneNumber(finalPhone, uiDelegate: nil) {
            verificationID, error in
            if let error = error {
                print(error)
                completion(false)
                return
            } else {
                UserDefaults.standard.set(verificationID, forKey: "authVerificationID")
                completion(true)
            }
        }
```

### Firebase SDK Version

master

### Xcode Version

14.0

### Installation Method

Swift Package Manager

### Firebase Product(s)

Authentication

### Targeted Platforms

iOS

### If using Swift Package Manager, the project's Package.resolved

<details>
<summary>Expand <code>Package.resolved</code> snippet</summary>
<br>

```json

Replace this line with the contents of your Package.resolved.

```

</details>

### If using CocoaPods, the project's Podfile.lock

<details>
<summary>Expand <code>Podfile.lock</code> snippet</summary>
<br>

```yml

Replace this line with the contents of your Podfile.lock!

```

</details>

morganchen12:
@Archetapp are you able to use phone auth successfully via our [sample app](https://github.com/firebase/quickstart-ios/tree/master/authentication)?

tomas-petrovcin:
Having the same issue (using via react-native-firebase).

As @Archetapp described, it works fine with previous iOS versions - doesn't work since developer beta iOS 16, now tested on stable release with same results

It works on devices, but I think it might be because it uses verification using notification, whereas simulator defaults to recaptcha. Seems to me it might be that recaptcha isn't able to open on new version of Safari or something like that. 

We could test that by enforcing recaptcha on device as well as see if it fails, but I am not sure how I could do that.

[...]

PR #10426, opened 2022-10-31
Force to recaptcha verification flow for phone auth in Simulators

renkelvin:
In the latest Xcode and macOS versions, APNs is now supported in simulators. Before that, phone auth will always fallback to the recaptcha verification flow.
Since FCM doesn't support iOS 16 simulators yet, https://github.com/firebase/firebase-ios-sdk/issues/9968, this change force the phone auth flow fallback to recaptcha in simulators.

### DIFF
--- a/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
+++ b/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
+#import <GoogleUtilities/GULAppEnvironmentUtil.h>
 #import <TargetConditionals.h>
+
 #if TARGET_OS_IOS
 
 #import "FirebaseAuth/Sources/Public/FirebaseAuth/FIRAuthSettings.h"
@@ -593,6 +595,10 @@ extern NSString *const FIRPhoneMultiFactorID;
  */
 - (void)verifyClientWithUIDelegate:(nullable id<FIRAuthUIDelegate>)UIDelegate
                         completion:(FIRVerifyClientCallback)completion {
+  if ([GULAppEnvironmentUtil isSimulator]) {
+    [self reCAPTCHAFlowWithUIDelegate:UIDelegate completion:completion];
+    return;
+  }
   if (_auth.appCredentialManager.credential) {
     completion(_auth.appCredentialManager.credential, nil, nil);
     return;
--- a/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
+++ b/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
@@ -595,16 +595,17 @@ extern NSString *const FIRPhoneMultiFactorID;
  */
 - (void)verifyClientWithUIDelegate:(nullable id<FIRAuthUIDelegate>)UIDelegate
                         completion:(FIRVerifyClientCallback)completion {
-  // Remove the simulator check below after FCM supports APNs in simulators
+// Remove the simulator check below after FCM supports APNs in simulators
+#if TARGET_OS_SIMULATOR
   if (@available(iOS 16, *)) {
-    if ([GULAppEnvironmentUtil isSimulator]) {
-      NSDictionary *environment = [[NSProcessInfo processInfo] environment];
-      if ((environment[@"XCTestConfigurationFilePath"] == nil)) {
-        [self reCAPTCHAFlowWithUIDelegate:UIDelegate completion:completion];
-        return;
-      }
+    NSDictionary *environment = [[NSProcessInfo processInfo] environment];
+    if ((environment[@"XCTestConfigurationFilePath"] == nil)) {
+      [self reCAPTCHAFlowWithUIDelegate:UIDelegate completion:completion];
+      return;
     }
   }
+#endif
+
   if (_auth.appCredentialManager.credential) {
     completion(_auth.appCredentialManager.credential, nil, nil);
     return;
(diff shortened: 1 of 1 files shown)


FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -593,6 +595,10 @@ - (void)verifyClientAndSendVerificationCodeToPhoneNumber:(NSString *)phoneNumber
  */
 - (void)verifyClientWithUIDelegate:(nullable id<FIRAuthUIDelegate>)UIDelegate
                         completion:(FIRVerifyClientCallback)completion {
+  if ([GULAppEnvironmentUtil isSimulator]) {
```
paulb777, line 598: Should there be an iOS version check for at least 16 as well?

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -593,6 +595,10 @@ - (void)verifyClientAndSendVerificationCodeToPhoneNumber:(NSString *)phoneNumber
...
+    [self reCAPTCHAFlowWithUIDelegate:UIDelegate completion:completion];
```
rosalyntan, line 599: I'm wondering if there's actually something going wrong in our fallback logic:

As per the reported issue, APNS is failing with INVALID_APP_CREDENTIAL: https://github.com/firebase/firebase-ios-sdk/issues/10214#issue-1373420565. However, we should be falling back to reCAPTCHA automatically in that case: https://github.com/firebase/firebase-ios-sdk/blob/bba19ad013bdd08eae551636e9e345642afb24a9/FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m#L624

That might be worth looking into a little before hard-coding simulators to always use reCAPTCHA.

renkelvin: I don't think we want to fallback to recaptcha in that case. If APNs is available but the app credential is invalid, there is a good chance the the app is not legitimate, so we should just error out.

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
renkelvin: Yeah, good point.

renkelvin: Added a check to not include this change in unit tests (as they're run in the simulators). Please take a look. Thanks!

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -593,6 +595,16 @@ - (void)verifyClientAndSendVerificationCodeToPhoneNumber:(NSString *)phoneNumber
  */
 - (void)verifyClientWithUIDelegate:(nullable id<FIRAuthUIDelegate>)UIDelegate
                         completion:(FIRVerifyClientCallback)completion {
+  // Remove the simulator check below after FCM supports APNs in simulators
```
paulb777, line 598: Why does Auth depend on FCM?

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -593,6 +595,16 @@ - (void)verifyClientAndSendVerificationCodeToPhoneNumber:(NSString *)phoneNumber
...
+  if (@available(iOS 17, *)) {
```
paulb777, line 599: How will this change make any difference for iOS 16 simulator users?

paulb777 (commented):
It looks like the integration test Firebase project is in a bad state for the signin tests. Would you fix?

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
renkelvin: Auth uses FCM to send a silent push notification which contains a secret for verification. It's not delivered in the simulators for now.

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
renkelvin: Ops, wrong commit. Just fixed.

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -593,6 +595,16 @@ - (void)verifyClientAndSendVerificationCodeToPhoneNumber:(NSString *)phoneNumber
  */
 - (void)verifyClientWithUIDelegate:(nullable id<FIRAuthUIDelegate>)UIDelegate
                         completion:(FIRVerifyClientCallback)completion {
+  // Remove the simulator check below after FCM supports APNs in simulators
+  if (@available(iOS 16, *)) {
+    if ([GULAppEnvironmentUtil isSimulator]) {
```
morganchen12, line 600: You can use `#if TARGET_OS_SIMULATOR` so this bit of code doesn't get compiled into non-simulator target destination builds.

morganchen12 (approved):

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
+#import <GoogleUtilities/GULAppEnvironmentUtil.h>
```
paulb777, line 17: This line may not be needed any longer.

paulb777 (approved):

FirebaseAuth/Sources/AuthProvider/Phone/FIRPhoneAuthProvider.m
renkelvin: Yeah, good point

renkelvin:
> It looks like the integration test Firebase project is in a bad state for the signin tests. Would you fix?

Yeah, it's fixed now.

PR #10430, opened 2022-10-31
[Sessions] Fix how Session ID is generated

samedson:
- Before, Session IDs would persist forever. So the previousSessionID would be set to the last sessionID, even if the app had cold started.
 - Now, previousSessionID is only set if the sessionID rolls over in a background-foreground initiation. It is not persisted across runs of the app

#no-changelog

### DIFF
--- a/FirebaseSessions/Tests/Unit/IdentifiersTests.swift
+++ b/FirebaseSessions/Tests/Unit/IdentifiersTests.swift
@@ -48,10 +48,18 @@ class IdentifiersTests: XCTestCase {
     return true
   }
 
+  // This test case isn't important behavior. When Crash and Perf integrate
+  // with the Sessions SDK, we may want to move to a lazy solution where
+  // sessionID can never be empty
+  func test_sessionID_beforeGenerateReturnsNothing() throws {
+    XCTAssert(identifiers.sessionID.count == 0)
+    XCTAssertNil(identifiers.previousSessionID)
+  }
+
   func test_generateNewSessionID_generatesValidID() throws {
     identifiers.generateNewSessionID()
     XCTAssert(isValidSessionID(identifiers.sessionID))
-    XCTAssert(identifiers.previousSessionID.count == 0)
+    XCTAssertNil(identifiers.previousSessionID)
   }
 
   /// Ensures that generating a Session ID multiple times results in the last Session ID being set in the previousSessionID field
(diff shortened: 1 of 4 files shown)


FirebaseSessions/Tests/Unit/IdentifiersTests.swift
```diff
@@ -60,15 +68,16 @@ class IdentifiersTests: XCTestCase {
...
     let firstSessionID = identifiers.sessionID
     XCTAssert(isValidSessionID(identifiers.sessionID))
-    XCTAssert(identifiers.previousSessionID.count == 0)
+    XCTAssertNil(identifiers.previousSessionID)
 
     identifiers.generateNewSessionID()
 
     XCTAssert(isValidSessionID(identifiers.sessionID))
-    XCTAssert(isValidSessionID(identifiers.previousSessionID))
+    XCTAssert(isValidSessionID(identifiers.previousSessionID ?? ""))
```
visumickey, line 76: Can we do something like below?
```
let previousSessionID = identifiers.previousSessionID
XCTAssert(isValidSessionID(previousSessionID!))
```
With the above statement, the actual error might get muddled in.

FirebaseSessions/Tests/Unit/IdentifiersTests.swift
```diff
@@ -60,15 +68,16 @@ class IdentifiersTests: XCTestCase {
...
 
     // Ensure the new lastSessionID is equal to the sessionID from earlier
-    XCTAssert(identifiers.previousSessionID.compare(firstSessionID) == ComparisonResult.orderedSame)
+    XCTAssert((identifiers.previousSessionID ?? "").compare(firstSessionID) == ComparisonResult
```
visumickey, line 79: Same as above.

FirebaseSessions/Tests/Unit/IdentifiersTests.swift
```diff
@@ -60,15 +68,16 @@ class IdentifiersTests: XCTestCase {
...
+      .orderedSame)
```
visumickey, line 80: General comment across - Can we use `XCTAssertEqual` for object comparison? It is a bit more comprehensive in comparison that direct comparison of strings.

samedson: Huh it seems they fixed String equality in Swift - I always had problems using AssertEqual for strings in ObjC

visumickey (approved):

PR #10437, opened 2022-11-01
Fix Firestore failing to return empty results from the local cache

dconeybe:
Fix a bug where Firestore fails to return results from the local cache if the result set was empty.

This bug is due to the following logic in `query_listener.cc`:

https://github.com/firebase/firebase-ios-sdk/blob/a1b78b247b11e5be77836c6ac86b66b453cb101d/Firestore/core/src/core/query_listener.cc#L150-L151

which assumes that if the result set from the local cache is empty that there is no cached result; however, if the result set of the query is indeed empty then it should be returning that empty result set from the cache.

This fix improves the logic to use the presence of a hasCachedResults flag to indicate that the empty result set is cached data and should be raised to the client.

This bug fix is ported from https://github.com/firebase/firebase-js-sdk/pull/6624 and https://github.com/firebase/firebase-android-sdk/pull/4207, which fixed https://github.com/firebase/firebase-js-sdk/issues/5873.


Firestore/Example/Tests/API/FSTAPIHelpers.mm
```diff
@@ -154,7 +155,8 @@
                             mutatedKeys,
                             static_cast<bool>(fromCache),
                             /*sync_state_changed=*/true,
-                            /*excludes_metadata_changes=*/false};
+                            /*excludes_metadata_changes=*/false,
+                            hasCachedResults ? true : false};
```
milaGGL, line 159: if hasCachedResults type is boolean, why we need to check it here? Are inputs like `BOOL hasCachedResults` nullable and thats why we add `hasCachedResults ? true : false` to check existence?

dconeybe: I thought the same thing. It turns out that Objective-C's `BOOL` type cannot be converted to C++'s `bool` type. That being said, I see a few lines above (line 156) they use `static_cast` to convert `fromCache` from `BOOL` to `bool`; therefore, I'll change the code to do _that_.

Firestore/Example/Tests/SpecTests/json/listen_spec_test.json
```diff
@@ -3858,6 +3856,25 @@
...
+                  "==",
+                  true
+                ]
+              ],
+              "orderBys": [
+              ],
+              "path": "collection"
+            }
+          }
+        ],
```
ehsannas, line 3877: I trust that these json files were generated correctly :)

Firestore/core/src/core/query_listener.cc
```diff
@@ -147,8 +148,10 @@ bool QueryListener::ShouldRaiseInitialEvent(const ViewSnapshot& snapshot,
     return false;
   }
 
-  // Raise data from cache if we have any documents or we are offline
-  return !snapshot.documents().empty() || online_state == OnlineState::Offline;
+  // Raise data from cache if we have any documents, have cached results before,
+  // or we are offline.
+  return !snapshot.documents().empty() || snapshot.has_cached_results() ||
+         online_state == OnlineState::Offline;
```
ehsannas, line 154: It's perhaps "safer" the way it's written now, but I'm curious why we didn't use:

```cpp
return snapshot.has_cached_results() || online_state == OnlineState::Offline;
```

in other words, if `has_cached_results` is true, it doesn't matter what the size of `snapshot.documents()` is. right?

ehsannas (approved):
LGTM!

Firestore/core/src/core/query_listener.cc
ehsannas: hmm... perhpas there's a scenario where snapshot.documents() is not empty, but snapshot.has_cached_results is false? (e.g. some other query downloaded some documents that overlap with this query?)

Firestore/Example/Tests/SpecTests/json/listen_spec_test.json
dconeybe: Lol! You _can_ verify by running the export yourself ;)

Firestore/core/src/core/query_listener.cc
dconeybe: This is a bit nuanced indeed. `has_cached_results()` returns `true` if there _are_ cached results, whether or not this particular snapshot is being raised from cache. Namely, `has_cached_results()` could be `true` but `fromCache()` could be `false` if the snapshot is not being raised from cache (but it _could_ be).

Also, the scenario where `!documents().empty()` is `true` but `has_cached_results()` is `false` is when the query has never been run before BUT the client performed a write that was written into the local cache, in addition to the scenario you outlined where there is a coincidentally query that was previously run.

dconeybe:
> LGTM!

Thanks for the review, @ehsannas!

Firestore/core/src/core/query_listener.cc
ehsannas: Thanks for clarifying 👍

paulb777:
@dconeybe Note that you can mark the fix version in the milestone field of the PR - I'll do this one now ...

dconeybe:
@paulb777 Ahh good to know. I'll use that in the future.

PR #10439, opened 2022-11-01
Combine nanopb scripts into a single location

samedson:
This PR does 2 things:
1) Unites all nanopb generation scripts into one folder, with the exception of Firestore.
2) Fixes the generator to always use repo-relative includes, eg. `#include "x.nanopb.h" => FirebaseX/Nanopb/x.nanopb.h`. This fixed a problem in FirebaseSessions include

Firestore has a much more complicated setup that I didn't have time to figure out. Most products have 1 or 2 protos, but Firestore has many, and they seem to be hooked up to CMake in some way.

## Validation
To validate all these scripts produce the same code, I deleted all the resulting `.nanopb.c/h` files, and ran the generators for all the affected SDKs. The only changes should be in copyright headers and the version of nanopb used. The generated source should stay the same.

Screen Shot 2022-11-01 at 3 01 49 PM

```
# Ran the following:

./Crashlytics/ProtoSupport/generate_crashlytics_protos.sh && 
./FirebaseInAppMessaging/ProtoSupport/generate_nanopb_protos.sh && 
FirebaseMessaging/ProtoSupport/generate_protos.sh &&
FirebasePerformance/ProtoSupport/generate_protos.sh &&
FirebaseSessions/generate_protos.sh
```

*The resulting changes have been included in this PR*

There is one change in the FIAM proto, that I've commented about below

#no-changelog


samedson:
Need to do `Firestore` too

Crashlytics/Protogen/nanopb/crashlytics.nanopb.h
```diff
@@ -15,7 +15,7 @@
  */
 
 /* Automatically generated nanopb header */
-/* Generated by nanopb-0.3.9.7 */
+/* Generated by nanopb-0.3.9.8 */
```
paulb777, line 18: There is no difference in generation, but technically we should be on 0.3.9.9 now.

FirebaseInAppMessaging/Sources/Analytics/Protogen/nanopb/fiam.nanopb.c
```diff
@@ -31,10 +31,10 @@ const pb_field_t logs_proto_firebase_inappmessaging_CampaignAnalytics_fields[10]
     PB_FIELD(  2, BYTES   , OPTIONAL, POINTER , OTHER, logs_proto_firebase_inappmessaging_CampaignAnalytics, campaign_id, project_number, 0),
     PB_FIELD(  3, MESSAGE , OPTIONAL, STATIC  , OTHER, logs_proto_firebase_inappmessaging_CampaignAnalytics, client_app, campaign_id, &logs_proto_firebase_inappmessaging_ClientAppInfo_fields),
     PB_FIELD(  4, INT64   , OPTIONAL, STATIC  , OTHER, logs_proto_firebase_inappmessaging_CampaignAnalytics, client_timestamp_millis, client_app, 0),
-    PB_ANONYMOUS_ONEOF_FIELD(event,   5, ENUM    , ONEOF, STATIC  , OTHER, logs_proto_firebase_inappmessaging_CampaignAnalytics, event_type, client_timestamp_millis, 0),
-    PB_ANONYMOUS_ONEOF_FIELD(event,   6, ENUM    , ONEOF, STATIC  , UNION, logs_proto_firebase_inappmessaging_CampaignAnalytics, dismiss_type, client_timestamp_millis, 0),
-    PB_ANONYMOUS_ONEOF_FIELD(event,   7, ENUM    , ONEOF, STATIC  , UNION, logs_proto_firebase_inappmessaging_CampaignAnalytics, render_error_reason, client_timestamp_millis, 0),
-    PB_ANONYMOUS_ONEOF_FIELD(event,   8, ENUM    , ONEOF, STATIC  , UNION, logs_proto_firebase_inappmessaging_CampaignAnalytics, fetch_error_reason, client_timestamp_millis, 0),
```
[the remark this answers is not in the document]
paulb777: Looks like FIAM wasn't updated when others were.

paulb777 (approved):
This is great! Thanks for consolidating this process.

scripts/nanopb/generate_protos.sh
```diff
@@ -24,41 +24,48 @@ readonly DIR="$( git rev-parse --show-toplevel )"
...
+  --include_prefix="${INCLUDE_PREFIX}"
 
 rm -rf "${NANOPB_TEMPDIR}"
 
 RED='\033[0;31m'
 NC='\033[0m'
 echo ""
 echo ""
-echo -e "${RED}Important: Any new proto fields of type string, repeated, or bytes must be specified in the sessions.options file${NC}"
+echo -e "${RED}Important: Any new proto fields of type string, repeated, or bytes must be specified in the proto's .options file with type:FT_POINTER${NC}"
```
ncooke3, line 69: Curious, could you explain why the `FT_POINTER` type is needed? It sounds like at one point that it wasn't. I've seen and used it before but didn't know why.

ncooke3 (approved):
Thanks Sam! Awesome cleanup 👏

scripts/nanopb/generate_protos.sh
samedson: This change is more of a clarification to the original intent of this log. I added this log around a week ago.

NanoPB has a few options for this, and it defaults to FT_CALLBACK. FT_CALLBACK requires you to implement an encode and decode function for each field, versus FT_POINTER just generates a field that's a `void *` that you fill in.

Eg., I changed `session_id` to FT_CALLBACK, and here's the diff:

```
 typedef struct _firebase_appquality_sessions_SessionInfo {
-    pb_bytes_array_t *session_id;
+    pb_callback_t session_id;
     pb_bytes_array_t *previous_session_id;
     pb_bytes_array_t *firebase_installation_id;
```

pb_callback_t looks like, which is extra work to implement:
```
struct pb_callback_s {
#ifdef PB_OLD_CALLBACK_STYLE
    /* Deprecated since nanopb-0.2.1 */
    union {
        bool (*decode)(pb_istream_t *stream, const pb_field_t *field, void *arg);
        bool (*encode)(pb_ostream_t *stream, const pb_field_t *field, const void *arg);
    } funcs;
#else
    /* New function signature, which allows modifying arg contents in callback. */
    union {
        bool (*decode)(pb_istream_t *stream, const pb_field_t *field, void **arg);
        bool (*encode)(pb_ostream_t *stream, const pb_field_t *field, void * const *arg);
    } funcs;
#endif    
    
    /* Free arg for use by callback */
    void *arg;
};
```

samedson:
> Looks like FIAM wasn't updated when others were.

@paulb777 do you mean that it wasn't updated to a new nanopb version? Do you have any worries with that ENUM => UENUM change?

paulb777:
No concerns. Your change aligns the proto generation with the library being linked with.

scripts/nanopb/generate_protos.sh
ncooke3: thanks Sam

Crashlytics/Protogen/nanopb/crashlytics.nanopb.h
samedson: Lemme follow up with this

PR #10442, opened 2022-11-02
Settings cache reader

leotianlizhan:
b/257074690
#no-changelog
## Goal
Add Crashlytics Settings to `FirebaseSessions`. 
## This PR
* Adds the "read local file" functionality of settings. 
* This also includes reading a so-called "cache-key" which provides meta-data to determine if current cache is expired. 
## Tests
Tests are split into 3 categories:
* Test settings API return correct values
  * default settings without cache
  * valid cache
* Test cache key expiration causes `isCacheExpired` to be marked as true
  * Can expire by TTL
  * Can expire by app version update
  * Can expire by GoogleID change
* Test corrupted cache file
  * Corrupted settings
  * Corrupted cache-key


FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,167 @@
...
+      Logger.logDebug("[Sessions:Settings] No settings were cached")
+      return
+    }
+    guard let parsedDictionary = cacheData.dictionaryValue else {
+      removeCache()
+      return
+    }
+    do {
+      objc_sync_enter(lock)
+      defer { objc_sync_exit(lock) }
```
samedson, line 96: This is smart - if I understand correctly you're using `do {...}` to create a block so that the `defer {objc_sync_exit...}` can evaluate at the end of the `do`, instead of at the end of the function.

I think you may be able to improve readability by making a lock function with a callback.

Eg.

```
func locked(callback: () -> ()) {
      objc_sync_enter(lock)
      defer { objc_sync_exit(lock) }
      callback()
}
```

Then use it with
```
locked() {
  /// do threading stuff
}
```

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,167 @@
...
+    if currentTime.timeIntervalSince(cacheKey.createdAt) > cacheDurationSeconds {
+      Logger.logDebug("[Sessions:Settings] Settings TTL expired")
+      do {
+        objc_sync_enter(lock)
+        defer { objc_sync_exit(lock) }
+        isCacheKeyExpired = true
+      }
+    }
+    if appInfo.synthesizedVersion != cacheKey.appVersion {
+      Logger.logDebug("[Sessions:Settings] Settings expired because app version changed")
```
samedson, line 120: Kindof a nit - I believe we can leave off the `[Sessions:Settings]` and either have no prefix, or just use `[Settings]` as the prefix. Our logger by default adds a prefix to logs, eg:

```
 10.2.0 - [FirebaseSessions][I-SES000000] Logging events using fast QOS due to running on a simulator
```

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,167 @@
...
+    isCacheKeyExpired = true
+    _settingsDictionary = nil
+  }
+}
+
+extension ApplicationInfoProtocol {
+  var synthesizedVersion: String { return "\(appDisplayVersion) (\(appBuildVersion))" }
+}
+
+private extension Data {
```
samedson, line 142: I believe it is customary to put extensions at the top of the file

FirebaseSessions/Sources/SettingsFileManager.swift
```diff
@@ -0,0 +1,76 @@
...
+    guard let cachesDirectory = fileManager.urls(for: .cachesDirectory, in: .userDomainMask).first
+    else {
+      directoryUrl = URL(fileURLWithPath: "/")
+      return
+    }
+    directoryUrl = cachesDirectory.appendingCompatible(path: SettingsFileManager.directoryName)
+    do {
+      try fileManager.createDirectory(at: directoryUrl, withIntermediateDirectories: true)
+    } catch {
+      Logger.logDebug("SettingsFileManager: \(error)")
```
samedson, line 40: I might log this as a `logError` and add some context, eg.: `SettingsFileManager failed to create caches directory with error: \(error)`

FirebaseSessions/Sources/SettingsFileManager.swift
```diff
@@ -0,0 +1,76 @@
...
+  func removeCacheFiles() {
+    DispatchQueue.global(qos: .background).async { [weak self] in
+      guard let strongSelf = self else {
+        return
+      }
+      do {
+        try strongSelf.fileManager.removeItem(at: strongSelf.settingsCacheKeyPath)
+        try strongSelf.fileManager.removeItem(at: strongSelf.settingsCacheContentPath)
+      } catch {
+        Logger.logDebug("removeItem failed: \(error)")
```
samedson, line 61: Debug is good here, but similarly I might add more context to the message for our future selves.

FirebaseSessions/Tests/Unit/SettingsTests.swift
```diff
@@ -0,0 +1,212 @@
...
+  var appInfo: MockApplicationInfo!
+
+  override func setUp() {
+    appInfo = MockApplicationInfo()
+    settingsFileManager = SettingsFileManager(fileManager: fileManager)
+    settings = Settings(fileManager: settingsFileManager, appInfo: appInfo)
+  }
+
+  func test_noCacheSaved_returnsDefaultSettings() {
+    assert(settings.isCacheExpired == true)
```
ncooke3, line 41: nit– I would favor XCTAssert for unit test assertions.

ncooke3 (commented):
easy-to-read tests 👍

FirebaseSessions/Tests/Unit/SettingsTests.swift
ncooke3:
```suggestion
    XCTAssertTrue(settings.isCacheExpired)
```

FirebaseSessions/Sources/Settings.swift
leotianlizhan: That's a great idea, I love this syntactic sugar. Updated the PR to do basically this, but using an extension so the syntax now look like:
```
lock.synchronized {
...
}
```

FirebaseSessions/Sources/Settings.swift
leotianlizhan: Done

FirebaseSessions/Sources/Settings.swift
leotianlizhan: Done

FirebaseSessions/Sources/SettingsFileManager.swift
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+  }
+}
+
+extension URL {
```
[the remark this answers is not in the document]
leotianlizhan: Done

FirebaseSessions/Sources/SettingsFileManager.swift
leotianlizhan: Done

FirebaseSessions/Sources/SettingsFileManager.swift
leotianlizhan: Done

FirebaseSessions/Tests/Unit/SettingsTests.swift
leotianlizhan: Done

leotianlizhan:
@ncooke3 @samedson Thanks for the review, addressed all the comments. However, I also added a `MockSettingsFileManager` to fix a test-flakiness issue with `removeCacheFilesAsync()`. Sometimes tests fail randomly because that method is running async in a dispatch queue, and it is fixed when I make it blocking/synchronous. However the way I did it feels kind of weird to me, is there a better way to mock this?

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+// limitations under the License.
+
+import Foundation
+
+extension ApplicationInfoProtocol {
+  var synthesizedVersion: String { return "\(appDisplayVersion) (\(appBuildVersion))" }
+}
+
+private extension Data {
+  var dictionaryValue: [String: AnyObject]? {
```
visumickey, line 23: What is the purpose of this dictionaryValue conversion?

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+
+extension NSLocking {
+  func synchronized<T>(_ closure: () throws -> T) rethrows -> T {
+    lock()
+    defer { unlock() }
+    return try closure()
+  }
+}
+
+protocol SettingsProtocol {
```
visumickey, line 57: Recommend adding some comments here for functions.

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+  var sessionsEnabled: Bool { get }
+  var samplingRate: Double { get }
+  var sessionTimeout: TimeInterval { get }
+}
+
+class Settings: SettingsProtocol {
+  private static let cacheDurationSecondsDefault: TimeInterval = 60 * 60
+  private let fileManager: SettingsFileManagerProtocol
+  private let appInfo: ApplicationInfoProtocol
+  // Underscored variables may be accessed in different threads,
```
visumickey, line 68: Is this a convention you are following?

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  var settingsDictionary: [String: AnyObject]? {
+    lock.synchronized {
+      _settingsDictionary
+    }
+  }
+
+  var sessionsEnabled: Bool {
+    guard let enabled = settingsDictionary?["sessions_enabled"]?.boolValue else {
```
visumickey, line 85: Lets define the Settings Flag Keys as Constants. Having them inline is hard to parse.

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+      return true
+    }
+    return enabled
+  }
+
+  var samplingRate: Double {
+    guard let rate = settingsDictionary?["sampling_rate"]?.doubleValue else {
+      return 1
```
visumickey, line 93: return "1.0"

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+  var sessionTimeout: TimeInterval {
+    guard let timeout = settingsDictionary?["session_timeout"]?.doubleValue else {
+      return 30 * 60
+    }
+    return timeout
+  }
+
+  var isCacheExpired: Bool {
+    guard settingsDictionary != nil else {
+      return true
```
visumickey, line 107: Default is `true`? Does that mean we try to expire the cache every time the dictionary is empty?

When would the settingsDictionary be empty?

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+    lock.synchronized {
+      _settingsDictionary = parsedDictionary
+    }
+    guard let cacheKeyData = fileManager.data(contentsOf: fileManager.settingsCacheKeyPath),
+          let cacheKey = cacheKeyData.cacheKeyValue else {
+      Logger.logError("[Settings] Could not load settings cache key")
+      removeCache()
+      return
+    }
+    guard cacheKey.googleAppID == googleAppID else {
```
visumickey, line 146: Is this for the support for multi-apps? Does Crashlytics support multi-apps situation?

visumickey (commented):
I think we need a general design of this service. I'm confused with that we are trying to accomplish here. Is the general approach that we will have a separate fetching service, storing the configs and this PR is about taking the configs from the file to the memory? We need to align on the design here.

Secondly, most of the classes/functions/variables have zero to less comments. But can we ensure that we have enough comments to qualify the intent of the those variables/functions?

FirebaseSessions/Sources/SettingsCacheClient.swift
```diff
@@ -0,0 +1,63 @@
...
+  private static let settingsVersion: Int = 1
+  private static let content: String = "settings"
+  private static let key: String = "cache-key"
+  /// UserDefaults holds values in memory, making access O(1) and synchronous within the app, while abstracting away async disk IO.
+  private let cache: UserDefaults = .standard
+
+  init() {}
+
+  /// Converting to dictionary is O(1) because object conversion is O(1)
+  var cacheContent: [String: Any]? {
```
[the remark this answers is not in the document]
leotianlizhan: cacheKey is the meta-data associated with current cache, only used to determine if current cache is still valid (is it expired, did app version update, did developer register a new app id). It's like a "real life key" to the cache

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,94 @@
...
+      cache.removeCache()
+      return true
+    }
+    guard cacheKey.googleAppID == appInfo.appID else {
+      Logger
+        .logDebug("[Settings] Cache expired because Google App ID changed")
+      cache.removeCache()
+      return true
+    }
+    if currentTime.timeIntervalSince(cacheKey.createdAt) > cacheDurationSeconds {
```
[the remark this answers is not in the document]
leotianlizhan: `cachekey.createdAt` is last fetched time. cache duration is part of the settings/cache content.

FirebaseSessions/Sources/Settings.swift
leotianlizhan: Done

FirebaseSessions/Sources/Settings.swift
leotianlizhan: removed

FirebaseSessions/Sources/Settings.swift
leotianlizhan: removed

FirebaseSessions/Sources/Settings.swift
leotianlizhan: outdated

FirebaseSessions/Sources/Settings.swift
leotianlizhan: It's for developers registering a new app id, not for multi-apps.

FirebaseSessions/Sources/SettingsCacheClient.swift
```diff
@@ -0,0 +1,63 @@
...
+}
+
+class SettingsCache: SettingsCacheClient {
+  private static let settingsVersion: Int = 1
+  private static let content: String = "settings"
+  private static let key: String = "cache-key"
+  /// UserDefaults holds values in memory, making access O(1) and synchronous within the app, while abstracting away async disk IO.
+  private let cache: UserDefaults = .standard
+
+  init() {}
```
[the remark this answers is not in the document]
leotianlizhan: Done

FirebaseSessions/Sources/SettingsCacheClient.swift
```diff
@@ -0,0 +1,63 @@
...
+  /// Returns in-memory cache content in O(1) time
+  var cacheContent: [String: Any]? { get }
+  /// Returns in-memory cache key, no performance guarantee because conversion depends on size of CacheKey
+  var cacheKey: CacheKey? { get }
+  func removeCache()
+}
+
+class SettingsCache: SettingsCacheClient {
+  private static let settingsVersion: Int = 1
+  private static let content: String = "settings"
```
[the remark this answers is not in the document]
leotianlizhan: done

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,173 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+extension ApplicationInfoProtocol {
```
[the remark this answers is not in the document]
leotianlizhan: Added some comments

FirebaseSessions/Sources/Settings.swift
leotianlizhan: Added some comments

FirebaseSessions/Sources/SettingsCacheClient.swift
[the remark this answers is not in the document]
visumickey: Got it. Thanks for clarifying. Now it makes a lot more sense.

Can we call the variable `cachedContent` - CacheContent sounded like a verb - meaning the action of caching. We are referring to only the cachedContent here.

visumickey (approved):

FirebaseSessions/Sources/Settings.swift
```diff
@@ -0,0 +1,96 @@
...
+  var sessionsEnabled: Bool {
+    guard let enabled = cache.cacheContent?["sessions_enabled"] as? Bool else {
+      return true
+    }
+    return enabled
+  }
+
+  var samplingRate: Double {
+    guard let rate = cache.cacheContent?["sampling_rate"] as? Double else {
+      return 1.0
```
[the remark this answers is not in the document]
leotianlizhan: Done

PR #10445, opened 2022-11-02
Firestore: Update spec tests

dconeybe:
Update the spec tests from https://github.com/firebase/firebase-js-sdk/tree/master/packages/firestore/test/unit/specs/ at commit https://github.com/firebase/firebase-js-sdk/commit/9e9ee7ba3383237e0b92ad8183dd0f12640f4f3a.

Note: The spec test changes from https://github.com/firebase/firebase-js-sdk/pull/6624 were manually reverted in this PR because otherwise the tests will fail. This will be fixed, and the spec tests updated, by https://github.com/firebase/firebase-ios-sdk/pull/10437.

#no-changelog

### DIFF
--- a/Firestore/Example/Tests/SpecTests/json/listen_spec_test.json
+++ b/Firestore/Example/Tests/SpecTests/json/listen_spec_test.json
@@ -2257,6 +2257,892 @@
       }
     ]
   },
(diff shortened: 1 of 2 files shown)


milaGGL (approved):
LGTM

PR #10447, opened 2022-11-02
Upgrade nanopb to 0.3.9.9 for collection of SDKs

samedson:
Follow up from https://github.com/firebase/firebase-ios-sdk/pull/10439

#no-changelog


paulb777 (approved):
Thanks!

PR #10448, opened 2022-11-02
Format nanopb generator scripts

samedson:
Follow up from https://github.com/firebase/firebase-ios-sdk/pull/10439

Validated this works by re-running generators and ensuring there's no changes.


paulb777 (approved):
Thanks! CI failure looks like unrelated flake

PR #10450, opened 2022-11-03
Update nanopb generator script to check for python and help

samedson:
Getting a working `python` command on macOS is non-trivial. This error message helps with that.

Screen Shot 2022-11-03 at 10 29 07 AM

Proof the technique works:

Screen Shot 2022-11-03 at 10 29 27 AM


samedson:
Fixing style in https://github.com/firebase/firebase-ios-sdk/pull/10448

visumickey (approved):

PR #10451, opened 2022-11-03
[Infra] Bump clang-format to v15

dconeybe:
Related to clang-format not managing brew versioning properly. See https://github.com/llvm/llvm-project/issues/49340

#no-changelog


ncooke3 (approved):
`check` is green so LGTM. Thanks!

PR #10452, opened 2022-11-03
Add development environment to the proto message ApplicationInfo.

visumickey:
Add the proto field containing the development environment that can be used to differentiate Autopush/Staging/Prod for Firelog/backend to differentiate.

#no-changelog


samedson (approved):
🚢 🚢 🚢

PR #10456, opened 2022-11-03
[IGNORE] Apply the session environment to the Session start event proto

visumickey:
#no-changelog


FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -113,4 +113,24 @@ class SessionStartEventTests: XCTestCase {
       XCTAssertEqual(event.proto.application_info.apple_app_info.os_name, expected)
     }
   }
+
+  func test_convertLogEnvironment_convertsCorrectly() {
+    let expectations: [(given: String, expected: firebase_appquality_sessions_LogEnvironment)] = [
+      ("prod", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_PROD),
```
samedson, line 119: Could we add an empty case?

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -113,4 +113,24 @@ class SessionStartEventTests: XCTestCase {
...
+      ("staging", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_STAGING),
+      ("autopush", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_AUTOPUSH),
+      ("PROD", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_PROD),
+      ("STAGING", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_STAGING),
+      ("AUTOPUSH", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_AUTOPUSH),
+      ("random", firebase_appquality_sessions_LogEnvironment_LOG_ENVIRONMENT_UNKNOWN),
+    ]
+
+    expectations.forEach { (given: String, expected: firebase_appquality_sessions_LogEnvironment) in
+      appInfo.environment = given
```
samedson, line 129: For this, I worry we're not testing the `environment` getter itself. We can instead set `ProcessInfo.processInfo.environment` and leave off the mock for appInfo.environment.

Alternatively, we can add a test to ApplicationInfoTest and do the `ProcessInfo.processInfo.environment` testing there

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -63,4 +66,15 @@ class ApplicationInfo: ApplicationInfoProtocol {
...
+
+  var environment: String {
+    if let env = ProcessInfo.processInfo.environment["FIREBASE_RUN_ENVIRONMENT"] {
+      // If the variable is set without any value, assume it is PROD.
+      if (env.trimmingCharacters(in: .whitespaces).isEmpty) {
+        return "PROD"
+      }
+      return env
+    }
+    return "PROD"
```
samedson, line 78: I think we can use an enum here to increase clarity:

```
    enum DevEnvironment: String {
        case prod   = "prod"
        case staging   = "staging"
        case autopush = "autopush"
    }
```

The `DevEnvironment(rawValue: String)` constructor will return a `DevEnvironment?`, so we can construct simply, and use a default value in case there's no match:

```
  let env = ProcessInfo...
  return DevEnvironment(rawValue: env.lowercased()) ?? DevEnvironment.prod
```


I think this will also ensure if we add or remove an env, switch statements will complain that we haven't exhausted all possibilities, like in `convertLogEnvironment`.

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -63,4 +66,15 @@ class ApplicationInfo: ApplicationInfoProtocol {
   var mccMNC: String {
     return FIRSESValidateMccMnc(networkInfo.mobileCountryCode, networkInfo.mobileNetworkCode) ?? ""
   }
+
+  var environment: String {
+    if let env = ProcessInfo.processInfo.environment["FIREBASE_RUN_ENVIRONMENT"] {
```
samedson, line 71: `ProcessInfo.processInfo.environment` are environment variables passed to the app. `ProcessInfo.processInfo.arguments` is how we get command line arguments. I think it'd be good for us to use command line arguments instead for consistency.

visumickey: I don't think arguments is a great fit here. Arguments is an array of Strings, whereas Environment is a map of String to String. With a MAP, it is easy to get a key-value pair. With an array it is usually beneficial only for the availability or unavailability of a flag.  Our usecase needs finding the value rather than just checking availability of the value.

https://developer.apple.com/documentation/foundation/processinfo/1417911-environment
https://developer.apple.com/documentation/foundation/processinfo/1415596-arguments

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
visumickey: Done.

Issue #10267, opened 2022-09-27
Error when using `linkWithProvider`


Lyokone:
### Description

When using the `linkWithProvider` method and clicking on the "OK" button on the top left corner, the app crash with the following error:
```
Thread 34: "One of IDToken, accessToken, pendingToken, or requestURI must be supplied."
```

![simulator_screenshot_F832AC73-4A77-4405-98BA-97298BE2D5FD](https://user-images.githubusercontent.com/3680002/192483213-270550b5-ff3c-4042-a417-17324633317c.png)

If I don't click on "OK" there are no issues

First reported in https://github.com/firebase/flutterfire/issues/9615

### Reproducing the issue

```
  [currentUser
      linkWithProvider:self.authProvider
            UIDelegate:nil
            completion:^(FIRAuthDataResult *authResult, NSError *error) {
              handleAppleAuthResult(self, arguments, auth, authResult.credential, error, result);
            }];
```

The completion is never called (checked by putting a breakpoint in it

### Firebase SDK Version

9.6.0

### Xcode Version

13.4.1

### Installation Method

CocoaPods

### Firebase Product(s)

Authentication

### Targeted Platforms

iOS

### Relevant Log Output

```shell
Thread 34: "One of IDToken, accessToken, pendingToken, or requestURI must be supplied."
```

### If using Swift Package Manager, the project's Package.resolved

<details>
<summary>Expand <code>Package.resolved</code> snippet</summary>
<br>

```json

Replace this line with the contents of your Package.resolved.

```

</details>

### If using CocoaPods, the project's Podfile.lock

<details>
<summary>Expand <code>Podfile.lock</code> snippet</summary>
<br>

```yml

Replace this line with the contents of your Podfile.lock!

```

</details>

Albertbol:
Hey, pretty crucial bug for production, is there any news on this one, basically, your app crashes every time when the user cancels authentication.

rizafran:
Thanks for reporting this. I've tried reproducing the issue and I was able to get the same behavior. Let me check this with the team and see what we can do here.

Albertbol:
Hey, any progress on this one? Pretty crucial bug.
Thanks

Albertbol:
Hey, any action on this?
There was a couple of pull requests already.
https://github.com/flutter/flutter/issues/85785

rizafran:
Sorry for the trouble, @Albertbol. Our engineers are already looking into this issue, but I'll make a follow up about this.

[...]

PR #10457, opened 2022-11-04
Update CHANGELOG.md

renkelvin:


FirebaseAuth/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
-Fix a bug where the linking/reauth flows errors are not caught. (#10407)
+- [fixed] Fix a bug where the linking/reauth flows errors are not caught. (#10407)
```
paulb777, line 2: If there's an issue number, please use that instead of the PR. #10267 in this case?

paulb777 (approved):

FirebaseAuth/CHANGELOG.md
renkelvin: Yeah, make sense.

PR #10468, opened 2022-11-08
Storage lock fix

morganchen12:
Experimental fix (maybe) for #10463.


FirebaseStorage/Sources/Storage.swift
```diff
@@ -302,6 +302,7 @@ import FirebaseAuthInterop
   }
 
   /// Map of apps to a dictionary of buckets to GTMSessionFetcherService.
+  private static let fetcherServiceLock = NSObject()
```
ncooke3, line 305: just a thought– in case this has similar behavior as an empty dictionary, maybe try creating a new UUID

morganchen12: `objc_sync_enter` uses the address so the object itself shouldn't matter. Dictionary being a value type with runtime optimizations messes with this according to [this blog post](https://straypixels.net/swift-dictionary-locking/), though the error we're running into isn't exactly the same.

paulb777: An NSObject should not have the same Swift runtime optimizations as a Swift dictionary.

This change fixes my repro example.

paulb777 (approved):
LGTM with a changelog.

The test failure is a common occasional flake

FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -137,6 +138,24 @@ class StorageResultTests: StorageIntegrationCommon {
     waitForExpectations()
   }
 
+  func testNoDeadlocks() throws {
```
paulb777, line 141: Nice!

FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -137,6 +138,24 @@ class StorageResultTests: StorageIntegrationCommon {
...
+    let storage2 = Storage.storage(url: "")
+
```
paulb777, line 143: Run style

PR #10474, opened 2022-11-09
Apply development environment to the Protos.

visumickey:
This is a replacement to the PR #10456 because of the state it got into.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/ApplicationInfo.swift
+++ b/FirebaseSessions/Sources/ApplicationInfo.swift
@@ -77,7 +77,7 @@ class ApplicationInfo: ApplicationInfoProtocol {
   }
   
   var environment: DevEnvironment {
-    let environment = envParams["FIREBASE_RUN_ENVIRONMENT"]
+    let environment = envParams["FirebaseSessionsRunEnvironment"]
     if (environment != nil) {
       return DevEnvironment(rawValue: environment!.trimmingCharacters(in: .whitespaces).lowercased()) ?? DevEnvironment.prod
     }
--- a/FirebaseSessions/Sources/ApplicationInfo.swift
+++ b/FirebaseSessions/Sources/ApplicationInfo.swift
@@ -18,6 +18,13 @@ import Foundation
 @_implementationOnly import FirebaseCore
 @_implementationOnly import GoogleUtilities
 
+/// Development environment for the application.
+enum DevEnvironment: String {
+  case prod       = "prod" // Prod environment
+  case staging    = "staging" // Staging environment
+  case autopush   = "autopush" // Autopush environment
+}
+
 protocol ApplicationInfoProtocol {
   /// Google App ID / GMP App ID
   var appID: String { get }
@@ -33,16 +40,21 @@ protocol ApplicationInfoProtocol {
 
   /// Validated Mobile Country Code and Mobile Network Code
   var mccMNC: String { get }
+  
+  /// Development environment on which the application is running.
+  var environment: DevEnvironment { get }
 }
 
 class ApplicationInfo: ApplicationInfoProtocol {
   let appID: String
 
   private let networkInfo: NetworkInfoProtocol
+  private let envParams: [String : String]
 
-  init(appID: String, networkInfo: NetworkInfoProtocol = NetworkInfo()) {
+  init(appID: String, networkInfo: NetworkInfoProtocol = NetworkInfo(), envParams: [String : String] = ProcessInfo.processInfo.environment) {
     self.appID = appID
     self.networkInfo = networkInfo
+    self.envParams = envParams
   }
 
   var bundleID: String {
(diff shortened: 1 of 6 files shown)


FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -63,4 +75,12 @@ class ApplicationInfo: ApplicationInfoProtocol {
   var mccMNC: String {
     return FIRSESValidateMccMnc(networkInfo.mobileCountryCode, networkInfo.mobileNetworkCode) ?? ""
   }
+  
+  var environment: DevEnvironment {
+    let environment = envParams["FIREBASE_RUN_ENVIRONMENT"]
+    if (environment != nil) {
+      return DevEnvironment(rawValue: environment!.trimmingCharacters(in: .whitespaces).lowercased()) ?? DevEnvironment.prod
```
samedson, line 82: Can we change `environment!` to `environment?` - Using an `!` can crash the app if it's null

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -33,16 +40,21 @@ protocol ApplicationInfoProtocol {
...
 }
 
 class ApplicationInfo: ApplicationInfoProtocol {
   let appID: String
 
   private let networkInfo: NetworkInfoProtocol
+  private let envParams: [String : String]
 
-  init(appID: String, networkInfo: NetworkInfoProtocol = NetworkInfo()) {
+  init(appID: String, networkInfo: NetworkInfoProtocol = NetworkInfo(), envParams: [String : String] = ProcessInfo.processInfo.environment) {
```
samedson, line 54: For this, you can use command line arguments as a dictionary by accessing UserDefaults (It might also match the envParams type)

If I setup a command line argument `-Hello 123`:
Screen Shot 2022-11-09 at 12 39 43 PM

Then I can access it in `UserDefaults.standard` or if you want a dictionary: `UserDefaults.standard.dictionaryRepresentation()`. The one thing is you have to leave off the `-` in the argument.

Screen Shot 2022-11-09 at 12 40 00 PM

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -63,4 +75,12 @@ class ApplicationInfo: ApplicationInfoProtocol {
   var mccMNC: String {
     return FIRSESValidateMccMnc(networkInfo.mobileCountryCode, networkInfo.mobileNetworkCode) ?? ""
   }
+  
+  var environment: DevEnvironment {
+    let environment = envParams["FIREBASE_RUN_ENVIRONMENT"]
```
samedson, line 80: Maybe make this a Sessions specific name? Eg. `-FIRSessionsRunEnvironment` or `FIREBASE_SESSIONS_RUN_ENVIRONMENT` if you want to keep it a plist.

Could we also make this a constant, eg:   `private let commandLineArgument = "-FIRSessionsRunEnvironment"` in this class

FirebaseSessions/Sources/ApplicationInfo.swift
samedson, line 82: Maybe in a follow up, could we add a log here or before logging to FireLog to tell us when we're using a non-prod environment?

visumickey: With a nil check above, it is safe to force unwrap it here. No necessary risks here.

paulb777: `if let environment = envParams....`

FirebaseSessions/Sources/ApplicationInfo.swift
visumickey: UserDefaults might not be a great fit for this. UserDefaults are typically used for storing users preferences and using that for capturing log environment does not look appropriate to me. So, for now, I will go ahead use this as an environment variable (like we discussed offline).

FirebaseSessions/Sources/ApplicationInfo.swift
visumickey: I like this idea. Will go ahead with "FIRSessionsRunEnvironment" as the name of the environment variable.

FirebaseSessions/Sources/ApplicationInfo.swift
visumickey: Done. in the DevLogger.

samedson (approved):
🚢 🚢 🚢  Ship it!

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -63,4 +76,12 @@ class ApplicationInfo: ApplicationInfoProtocol {
   var mccMNC: String {
     return FIRSESValidateMccMnc(networkInfo.mobileCountryCode, networkInfo.mobileNetworkCode) ?? ""
   }
+
+  var environment: DevEnvironment {
+    if let environment = envParams["FirebaseSessionsRunEnvironment"] {
```
samedson, line 81: This is great - I forgot about `if let` and `guard let`

PR #10479, opened 2022-11-10
[Sessions] Revert broken encoding fix and fix decoding

samedson:
Follow on from https://github.com/firebase/firebase-ios-sdk/pull/10428, we found that the encoding "fix" resulted in explicit null terminators in the backend, which broke parsing and created extra characters in the string. 

 - Turns out encoding was correct, and decoding was the problem
 - The fix is to revert the encoding change that adds null terminators, and instead ad null terminators to the decoding.

#no-changelog


samedson:
I want to improve the code that adds the null terminator to decoding, so this is WIP right now

visumickey (approved):
The changes look good to me. Is there any way we could catch this with tests? This is intentionally gnarly that debugging such issues would be tricky!

samedson:
Yes I added some tests in https://github.com/firebase/firebase-ios-sdk/pull/10428 - they just don't seem to acknowledge that there's a null terminator at the end of the decoded string. I think maybe I can improve them by checking the proto bytes directly - I'll think about that

PR #10413, opened 2022-10-27
Fdl/fixes issues parsing utm parameters from universal links

cavalos0086:
Fix https://github.com/firebase/firebase-ios-sdk/issues/10341
- Fixes issue where utm_parameters are not returned back when using universal links
- Adds `utm_term` and `utm_content` as part of the utm parameters

PR #10498, opened 2022-11-16
Fdl/removes nsnull values

cavalos0086:
On this PR: [10413](https://github.com/firebase/firebase-ios-sdk/pull/10413), we use `NSNull` for values inside the dictionary. Some tests were broken because of that.

### DIFF
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,5 +1,5 @@
 # unreleased
-- [fixed] Fixes utm params breaking certaing flows
+- [fixed] Fixes issue where `utmParametersDictionary` / `minimumAppVersion` were not provided and their value were set to `[NSNull null]` instead of `nil`. 
 
 # 10.2.0
 - [fixed] Fixes utm parameters not being returned to dynamic link when using universal links (#10341)
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,3 +1,6 @@
+# unreleased
+- [fixed] Fixes utm params breaking certaing flows
+
 # 10.2.0
 - [fixed] Fixes utm parameters not being returned to dynamic link when using universal links (#10341)
 
--- a/FirebaseDynamicLinks/CHANGELOG.md
+++ b/FirebaseDynamicLinks/CHANGELOG.md
@@ -1,5 +1,5 @@
 # unreleased
-- [fixed] Fixes issue where `utmParametersDictionary` / `minimumAppVersion` were not provided and their value were set to `[NSNull null]` instead of `nil`. 
+- [fixed] Fixes issue where `utmParametersDictionary` / `minimumAppVersion` were not provided and their value were set to `[NSNull null]` instead of `nil`.
 
 # 10.2.0
 - [fixed] Fixes utm parameters not being returned to dynamic link when using universal links (#10341)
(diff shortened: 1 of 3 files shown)


FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
```diff
@@ -894,6 +894,63 @@ - (void)testDynamicLinkFromUniversalLinkURLReturnsDLMinimumVersion {
...
+      [NSString stringWithFormat:kStructuredUniversalLinkFmtSubdomainDeepLink, urlSuffix];
+  NSURL *url = [NSURL URLWithString:urlString];
+
+  [self.service setUpWithLaunchOptions:nil
+                                apiKey:kAPIKey
+                             urlScheme:kURLScheme
+                          userDefaults:self.userDefaults];
+  XCTestExpectation *expectation = [self expectationWithDescription:@"completion called"];
+  [self.service
+      dynamicLinkFromUniversalLinkURL:url
```
paulb777, line 921: Does one of these tests cover the escape that was caught by the 1P app?

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
```diff
@@ -417,17 +417,37 @@ - (nullable FIRDynamicLink *)dynamicLinkFromCustomSchemeURL:(NSURL *)url {
...
+
+          if (parameters[kFIRDLParameterContent] != nil) {
+            [paramsDictionary setValue:parameters[kFIRDLParameterContent]
+                                forKey:kFIRDLParameterContent];
+          }
+
+          FIRDynamicLink *dynamicLink =
+              [[FIRDynamicLink alloc] initWithParametersDictionary:paramsDictionary];
           dynamicLink.matchType = FIRDLMatchTypeUnique;
+          dynamicLink.minimumAppVersion = parameters[kFIRDLParameterMinimumAppVersion];
```
paulb777, line 450: Why doesn't this have the same `nil` checking? It looks like it can be nil from its documentation: `@param minAppVersion The minimum app version string, if any, found in the response from a server`

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
paulb777: I mean - did you add a test so we don't need to rely on 1P tests to regression test?

FirebaseDynamicLinks/Sources/FIRDynamicLinks.m
eldhosembabu: Yeah thats true. 

But it seems like @cavalos0086 reverted it back to the old code: https://github.com/firebase/firebase-ios-sdk/pull/10413/files#diff-38f8dcf2a819794228e1caf6b2b19c7432ceadfbd049fb77437a86c08b47cd09L423

FirebaseDynamicLinks/Tests/Unit/FIRDynamicLinksTest.m
paulb777: I see. Probably not necessary.Thanks.

paulb777 (approved):

eldhosembabu (approved):

FirebaseDynamicLinks/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# unreleased
+- [fixed] Fixes utm params breaking certaing flows
```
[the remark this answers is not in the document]
cavalos0086: Done!

paulb777:
The fix is staged for the next 10.3.0 release, due out next release.

In the meantime, this PR can be manually applied to workaround.

PR #10501, opened 2022-11-17
Rename the class "Indentifiers" to "Session"

visumickey:
Rename the class `Indentifiers` to `Session` since the class caters to all the necessary details of a session.

This is a precursor to bring in the "Sampling" changes.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/Session.swift
+++ b/FirebaseSessions/Sources/Session.swift
@@ -17,7 +17,7 @@ import Foundation
 
 @_implementationOnly import FirebaseInstallations
 
-protocol IdentifierProvider {
+protocol SessionProtocol {
   var installationID: String {
     get
   }
@@ -32,13 +32,13 @@ protocol IdentifierProvider {
 }
 
 ///
-/// Identifiers is responsible for:
+/// Session is responsible for:
 ///   1) Getting the Installation ID from Installations
 ///   2) Generating the Session ID
 ///   3) Persisting and reading the Session ID from the last session
 ///   (Maybe) 4) Persisting, reading, and incrementing an increasing index
 ///
-class Identifiers: IdentifierProvider {
+class Session: SessionProtocol {
   private let installations: InstallationsProtocol
 
   private var _sessionID: String?
@@ -61,7 +61,7 @@ class Identifiers: IdentifierProvider {
     if Thread.isMainThread {
       Logger
         .logError(
-          "Error: Identifiers.installationID getter must be called on a background thread. Using an empty ID"
+          "Error: Session.installationID getter must be called on a background thread. Using an empty ID"
         )
       return ""
     }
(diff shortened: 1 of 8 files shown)


leotianlizhan:
I'm not sure what's the purpose of doing this. I think this name is confusing with other classes such as `SessionsStartEvent` and `FirebaseSessions`. I also worry about adding [too much responsibility to 1 class](https://moma.corp.google.com/search?q=single+responsibility).

leotianlizhan (commented):
I'm not sure what's the purpose of doing this. I think this name is confusing with other classes such as `SessionsStartEvent` and `FirebaseSessions`. I also worry about adding [too much responsibility to 1 class](https://moma.corp.google.com/search?q=single+responsibility).

visumickey:
> I'm not sure what's the purpose of doing this. I think this name is confusing with other classes such as `SessionsStartEvent` and `FirebaseSessions`. I also worry about adding [too much responsibility to 1 class].

Session is a data construct on its own (containing its own attributes like sessionID, previous sessionId, should sample and more) - The current class `identifiers` is a generic class containing all the identifiers which does not provide separation of concerns.

Session is just the construct - SessionStartEvent is the event that gets dispatched to Firelog. So, they are separate. I don't think that would be confusing.

About FirebaseSessions, I think that would need some renaming, but since it is namespaced with Firebase - I would not be worried as that is how the library is named and it is needed to identify the starting class of the library.

FirebaseSessions/Sources/Session.swift
```diff
@@ -32,13 +32,13 @@ protocol IdentifierProvider {
...
 ///
-/// Identifiers is responsible for:
+/// Session is responsible for:
 ///   1) Getting the Installation ID from Installations
 ///   2) Generating the Session ID
 ///   3) Persisting and reading the Session ID from the last session
 ///   (Maybe) 4) Persisting, reading, and incrementing an increasing index
 ///
-class Identifiers: IdentifierProvider {
+class Session: SessionProtocol {
```
ncooke3, line 41: As someone not familiar with all the details, the `Identifiers` name confused me because it sounds like a collection type. What is the purpose of this class? Is it to represent a "session" or just provide a caller with an installation ID?

leotianlizhan: I think this class is supposed to be a provider of all the IDs we need, including `sessionID` and `installationID`, and is responsible for generating them and making them available.

ncooke3: Thanks @leotianlizhan. Hmm, maybe then something like:
- `IdentityManager` 
- `IdentifierManager`
- `IdentityController` 
- `IdentifierController`
- `IdentityProvider`
- `IdentifierProvider`
- `IdentifierGenerator`
- `IdentifierCreator`

One of these might be a better fit? @visumickey

PR #10505, opened 2022-11-17
Update versions for Release 10.3.0

chliangGoogle:


paulb777:
Based on the zip build failure, InAppMessaging and other pods are missing from https://github.com/firebase/SpecsStaging

PR #10508, opened 2022-11-18
Depend on Google Utlities for device platform + Add Model information to session events.

visumickey:
#no-changelog

### DIFF
--- a/FirebaseSessions/Tests/TestApp/Podfile.lock
+++ b/FirebaseSessions/Tests/TestApp/Podfile.lock
@@ -1,0 +1,79 @@
+PODS:
+  - FirebaseCore (10.2.0):
+    - FirebaseCoreInternal (~> 10.0)
+    - GoogleUtilities/Environment (~> 7.8)
+    - GoogleUtilities/Logger (~> 7.8)
+  - FirebaseCoreExtension (10.2.0):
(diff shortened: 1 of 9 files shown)


FirebaseSessions/Tests/TestApp/Podfile.lock
```diff
@@ -0,0 +1,79 @@
+PODS:
```
leotianlizhan, line 1: Is this file meant to be added? Otherwise the rest of the code LGTM!

leotianlizhan (commented):
LGTM except I'm not sure if the `Podfile.lock` is meant to be added

FirebaseSessions/Tests/TestApp/Podfile.lock
visumickey: Podfile.lock is the file that contains the version of the dependencies that is used for referencing dependencies and their version number. Without this, Cocoapods will continue to use the older version of the dependency. Since, we are updating the version of GoogleUtilities, I chose to add this file in to say that we work only with a particular version of GoogleUtilities and above.

leotianlizhan: Sounds good!

leotianlizhan (approved):

Issue #10010, opened 2022-07-14
FirebaseStorage - reference(forUrl: URL) Invalid url format


willshelley:
### [REQUIRED] Step 1: Describe your environment

  * Xcode version: 14 beta
  * Firebase SDK version: 9.2
  * Installation method: Swift Package Manager
  * Firebase Component: Storage
  * Target platform(s): iOS 16 beta only

### [REQUIRED] Step 2: Describe the problem

After fetching the `downloadUrl` from a `storageRef` i'm receiving a correctly formatted download URL with %2F for the path parameters. When fetching the `storageRef` for this url using `Storage.storage().reference(forUrl: URL)`. The returned `storageRef` does not properly decode the path where the object is stored in Storage i.e. `gs://someapp.appspot.com/images%2FsomeImage`, Expecting `gs://someapp.appspot.com/images/someImage`. I'm then receiving a correct permissions error as this is not the correct path to the object.

This is only occurring when testing a real device with iOS16 beta. All other versions seem to be working. 

Current workaround is to use: `url.removingPercentEncoding`

```
let downloadUrl = try await storageRef.downloadURL().absoluteString

let storageRef = db.reference(forURL: url)
let data = try await storageRef.data(maxSize: 10485760)
// Correct permissions error for trying to access a object that doesn't exist - gs://someapp.appspot.com/images%2FsomeImage

let storageRef = db.reference(forURL: url.removingPercentEncoding!)
let data = try await storageRef.data(maxSize: 10485760)
// This produces expected result - gs://someapp.appspot.com/images/someImage

```

andrewheard:
Thanks for pointing this out! I'm actually able to reproduce what you're seeing both on the iOS 16 simulator and a real device. I've traced it down to a change in [`URL.pathComponents`](https://developer.apple.com/documentation/foundation/url/1780324-pathcomponents?changes=latest_major)'s behaviour in iOS 16.

### Example
```
let url = URL(string: "https://example.com/images%2Fexample.png")!
print("URL Components: \(url.pathComponents)")
```

- iOS 15.5: `URL Components: ["/", "images", "example.png"]`
- iOS 16.0 Beta 3: `URL Components: ["/", "images%2Fexample.png"]`

The documentation for [NSURL](https://developer.apple.com/documentation/foundation/nsurl/1407365-pathcomponents) states:
[...]

PR #10511, opened 2022-11-19
[Crashlytics] Adding unit test scheme to SPM

themiswang:
### Discussion
Right now Crashlytics doesn't support unit test from SPM, adding the Craslytics unit tests scheme in for SPM

### Testing
Will run CI/CD to make sure nothing broken

#no-changelog

### DIFF
--- a/Crashlytics/UnitTests/FABURLBuilderTests.m
+++ b/Crashlytics/UnitTests/FABURLBuilderTests.m
@@ -34,7 +34,7 @@
 - (void)tearDown {
   [super tearDown];
 }
-
+#if !TARGET_OS_WATCH
 - (void)testTheRightUrlEncodingIsUsed {
   NSString *actual =
       @"https://settings.crashlytics.com/spi/v2/platforms/ios/apps/"
@@ -52,5 +52,5 @@
   NSString *encoded = [urlBuilder escapeString:actual];
   XCTAssertEqualObjects(encoded, expected);
 }
-
+#endif
 @end
--- a/Package.swift
+++ b/Package.swift
@@ -531,7 +531,23 @@ let package = Package(
         .linkedFramework("SystemConfiguration", .when(platforms: [.iOS, .macOS, .tvOS])),
       ]
     ),
-
+    .testTarget(
+      name: "FirebaseCrashlyticsUnit",
+      dependencies: ["FirebaseCrashlytics", "OCMock"],
+      path: "Crashlytics/UnitTests",
+      resources: [
+        .process("FIRCLSMachO/data"),
+        .copy("Data"),
+      ],
+      cSettings: [
+        .headerSearchPath("../.."),
+        .define("DISPLAY_VERSION", to: firebaseVersion),
+        .define("CLS_SDK_NAME", to: "Crashlytics iOS SDK", .when(platforms: [.iOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics macOS SDK", .when(platforms: [.macOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics tvOS SDK", .when(platforms: [.tvOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics watchOS SDK", .when(platforms: [.watchOS])),
+      ]
+    ),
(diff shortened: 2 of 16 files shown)


Crashlytics/UnitTests/FABURLBuilderTests.m
```diff
@@ -34,7 +34,7 @@ - (void)setUp {
 - (void)tearDown {
   [super tearDown];
 }
-
+#if !TARGET_OS_WATCH
```
themiswang, line 37: Skip this unit test for watchOS. It has a different encoding strategy and it is broken on my local watchOS unit tests.
https://github.com/firebase/firebase-ios-sdk/blob/17cc9723c8fdf469f89f65be70aca7902ec4e4f0/Crashlytics/Shared/FIRCLSNetworking/FIRCLSURLBuilder.m#L52 
Screen Shot 2022-11-18 at 11 52 48 PM

Package.swift
```diff
@@ -531,7 +531,26 @@ let package = Package(
...
       ]
     ),
-
+    .testTarget(
+      name: "FirebaseCrashlyticsUnit",
+      dependencies: ["FirebaseCrashlytics", "OCMock"],
+      path: "Crashlytics/UnitTests",
+      resources: [
+        .process("FIRCLSMachO/data"),
+        .copy("Data"),
```
themiswang, line 540: Retains the directory’s structure since in there are files have same names under this directory. Example: `metadata.clsrecord`
Screen Shot 2022-11-21 at 9 55 54 AM

Crashlytics/UnitTests/FABURLBuilderTests.m
paulb777: I wonder if this is an Xcode 14 regression like #10010

Package.swift
```diff
@@ -531,7 +531,26 @@ let package = Package(
...
+      ],
+      cSettings: [
+        .headerSearchPath("../.."),
+        .define("DISPLAY_VERSION", to: firebaseVersion),
+        .define("CLS_SDK_NAME", to: "Crashlytics iOS SDK", .when(platforms: [.iOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics macOS SDK", .when(platforms: [.macOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics tvOS SDK", .when(platforms: [.tvOS])),
+        .define("CLS_SDK_NAME", to: "Crashlytics watchOS SDK", .when(platforms: [.watchOS])),
+        .define("PB_FIELD_32BIT", to: "1"),
```
paulb777, line 549: I don't think the `PB` defines are needed since all of the nanopb references should be from the Crashlytics library and not the tests.

paulb777 (approved):
Great work! Thanks @themiswang

Crashlytics/UnitTests/FABURLBuilderTests.m
themiswang: Seems like this is a more of an issue on how paths are percent encoded in `NSString `: https://stackoverflow.com/questions/44381108/inconsistencies-in-url-encoding-methods-across-objective-c-and-swift
I agree with the highest scored answer there `It is attempting to percent encode a URL as if it was just a path`, but as the comment we have I am not sure who and why we use NSCharacterSet.URLPathAllowedCharacterSet for watchOS. Let me remove the environment condition and rerun the CI/CD again see if it can pass the unit test on SPM.

Package.swift
themiswang: Removed all PB defines!

Crashlytics/UnitTests/FABURLBuilderTests.m
themiswang: @paulb777 It failed :(
https://github.com/firebase/firebase-ios-sdk/actions/runs/3516805883/jobs/5893825977
Screen Shot 2022-11-21 at 12 54 54 PM

themiswang: Disable this unit test for watchOS now. I'll sync up with @samedson later once he comes back from OOO.

PR #10516, opened 2022-11-22
Introduce sampling to session events.

visumickey:
Implement sampling for session related events.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessionsError.swift
+++ b/FirebaseSessions/Sources/FirebaseSessionsError.swift
@@ -1,0 +1,21 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Contains the list of errors that are localized for Firebase Sessions Library
+enum FirebaseSessionsError: Error {
+  /// Event sampled related error
+  case SessionSampledError
+}
--- a/FirebaseSessions/Sources/FirebaseSessionsError.swift
+++ b/FirebaseSessions/Sources/FirebaseSessionsError.swift
@@ -16,6 +16,6 @@ import Foundation
 
 /// Contains the list of errors that are localized for Firebase Sessions Library
 enum FirebaseSessionsError: Error {
-  /// Event sampled related error
-  case SessionSampledError
+  /// Event sampling related error
+  case SessionSamplingError
 }
--- a/FirebaseSessions/Sources/SessionCoordinator.swift
+++ b/FirebaseSessions/Sources/SessionCoordinator.swift
@@ -33,7 +33,7 @@ class SessionCoordinator {
   // Begins the process of logging a SessionStartEvent to FireLog, while taking into account Data Collection, Sampling, and fetching Settings
   func attemptLoggingSessionStart(event: SessionStartEvent,
                                   callback: @escaping (Result<Void, Error>) -> Void) {
-    if sampler.shouldSendEventForSession(sessionId: identifiers.sessionID) == true {
+    if sampler.shouldSendEventForSession(sessionId: identifiers.sessionID) {
       event.setInstallationID(identifiers: identifiers)
 
       fireLogger.logEvent(event: event) { result in
--- a/FirebaseSessions/Sources/SessionCoordinator.swift
+++ b/FirebaseSessions/Sources/SessionCoordinator.swift
@@ -21,29 +21,40 @@ import Foundation
 class SessionCoordinator {
   let identifiers: IdentifierProvider
   let fireLogger: EventGDTLoggerProtocol
+  let sampler: SessionSamplerProtocol
 
-  init(identifiers: IdentifierProvider, fireLogger: EventGDTLoggerProtocol) {
+  init(identifiers: IdentifierProvider, fireLogger: EventGDTLoggerProtocol, sampler: SessionSamplerProtocol) {
     self.identifiers = identifiers
     self.fireLogger = fireLogger
+    self.sampler = sampler
   }
 
   // Begins the process of logging a SessionStartEvent to FireLog, while taking into account Data Collection, Sampling, and fetching Settings
   func attemptLoggingSessionStart(event: SessionStartEvent,
                                   callback: @escaping (Result<Void, Error>) -> Void) {
-    event.setInstallationID(identifiers: identifiers)
-
-    fireLogger.logEvent(event: event) { result in
-      switch result {
-      case .success():
-        Logger.logInfo("Successfully logged Session Start event to GoogleDataTransport")
-        callback(.success(()))
-      case let .failure(error):
-        Logger
-          .logError(
-            "Error logging Session Start event to GoogleDataTransport: \(error)."
-          )
-        callback(.failure(error))
+    
+    if (self.sampler.shouldSendEventForSession(sessionId: identifiers.sessionID) == true) {
+      event.setInstallationID(identifiers: identifiers)
+      
+      fireLogger.logEvent(event: event) { result in
+        switch result {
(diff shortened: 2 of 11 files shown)


FirebaseSessions/Sources/SessionSampler.swift
```diff
@@ -0,0 +1,43 @@
...
+  var sessionSamplingRate: Double
+
+  /// TODO: Update this to a sampling logic once we have the configuration flags in place.
+  /// Currently defaulted to 1.0 where no events are dropped.
+  init(sessionSamplingRate: Double = 1.0) {
+    self.sessionSamplingRate = sessionSamplingRate
+  }
+
+  func shouldSendEventForSession(sessionId: String) -> Bool {
+    let randomFloat = Double.random(in: 0 ..< 1)
```
ncooke3, line 37: We want `1` to be a possible value, correct? If so:
```suggestion
    let randomFloat = Double.random(in: 0 ... 1)
```

FirebaseSessions/Sources/SessionSampler.swift
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol SessionSamplerProtocol {
+  /// Sampling rate that has to be applied across sessions.
+  /// Ranges from 0 to 1 in Double.
+  var sessionSamplingRate: Double { get set }
```
ncooke3, line 20: Does this need to be exposed to clients of types conforming to `SessionSamplerProtocol`? My hunch is no because a `SessionSamplerProtocol` should tell me whether or not to send an event– the sampling rate is an implementation detail.

FirebaseSessions/Sources/SessionCoordinator.swift
```diff
@@ -21,29 +21,40 @@ import Foundation
...
+        switch result {
+        case .success():
+          Logger.logInfo("Successfully logged Session Start event to GoogleDataTransport")
+          callback(.success(()))
+        case let .failure(error):
+          Logger
+            .logError(
+              "Error logging Session Start event to GoogleDataTransport: \(error)."
+            )
+          callback(.failure(error))
```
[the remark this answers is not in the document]
visumickey: I totally like this idea. All the errors from the FirebaseSessions library should be scoped to SessionsError type. But I would recommend this refactor to be done in a later PR separate from this change.

FirebaseSessions/Sources/SessionSampler.swift
visumickey: Ha! Thanks for the catch and suggestion. You are right, we want 1 to be allowed.

FirebaseSessions/Sources/SessionSampler.swift
```diff
@@ -0,0 +1,43 @@
...
+  init(sessionSamplingRate: Double = 1.0) {
+    self.sessionSamplingRate = sessionSamplingRate
+  }
+
+  func shouldSendEventForSession(sessionId: String) -> Bool {
+    let randomFloat = Double.random(in: 0 ..< 1)
+    if randomFloat > sessionSamplingRate {
+      return false
+    }
+    return true
```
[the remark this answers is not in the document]
visumickey: I really wanted to do this succinct. But in the interest of easy understanding, chose to be a bit more explicit in the if checks here.

FirebaseSessions/Sources/SessionCoordinator.swift
```diff
@@ -21,29 +21,40 @@ import Foundation
...
     self.identifiers = identifiers
     self.fireLogger = fireLogger
+    self.sampler = sampler
   }
 
   // Begins the process of logging a SessionStartEvent to FireLog, while taking into account Data Collection, Sampling, and fetching Settings
   func attemptLoggingSessionStart(event: SessionStartEvent,
                                   callback: @escaping (Result<Void, Error>) -> Void) {
-    event.setInstallationID(identifiers: identifiers)
+    if sampler.shouldSendEventForSession(sessionId: identifiers.sessionID) == true {
```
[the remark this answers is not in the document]
visumickey: Like the suggestion, done.

FirebaseSessions/Sources/SessionSampler.swift
visumickey: It is as per the design.

SessionSampler decides if an event needs to be sampled based on 2 factors - Event, Sampling rate.

Event is passed as an input and Sampling Rate is configurable (since that is an external factor that determines it - Configuration System).

So it is per the design to make it configurable.

FirebaseSessions/Sources/FirebaseSessionsError.swift
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Contains the list of errors that are localized for Firebase Sessions Library
+enum FirebaseSessionsError: Error {
+  /// Event sampled related error
+  case SessionSampledError
```
leotianlizhan, line 20: Nit: "Sampled" by definition means to grab a subset from a set of data, so this could be confusing to think "error, because the session is being sent to backend?". I wonder if there's a better word for "this-session-is-being-dropped-error".

FirebaseSessions/Sources/SessionSampler.swift
```diff
@@ -0,0 +1,43 @@
...
+  init(sessionSamplingRate: Double = 1.0) {
+    self.sessionSamplingRate = sessionSamplingRate
+  }
+
+  func shouldSendEventForSession(sessionId: String) -> Bool {
+    let randomFloat = Double.random(in: 0 ... 1)
+    if randomFloat > sessionSamplingRate {
+      return false
+    }
+    return true
```
leotianlizhan, line 41: Conciseness nit: 
```
return randomFloat <= sessionSamplingRate
```

visumickey: This is similar to the comment above. I chose to avoid conciseness in favor of easy to understand.

FirebaseSessions/Tests/Unit/SessionCoordinatorTests.swift
```diff
@@ -84,4 +90,43 @@ class SessionCoordinatorTests: XCTestCase {
...
+        resultSuccess = true
+      case .failure:
+        resultSuccess = false
+      }
+    }
+
+    XCTAssertTrue(resultSuccess)
+  }
+
+  func test_eventDropped_EventsSampled() throws {
```
leotianlizhan, line 113: Similar to earlier comment, `sample` is a confusing word, and to follow the same pattern as the previous test, perhaps `test_eventDropped_handlesZeroSamplingRate`?

visumickey: Good point. Updated it.

FirebaseSessions/Tests/Unit/SessionSamplerTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+  /// Validates if the events are disabled when the sampling rate is Zero.
+  func test_DisablesEventCollection_samplingRateZero() {
+    let localSampler = SessionSampler(sessionSamplingRate: 0.0)
+    XCTAssertEqual(localSampler.shouldSendEventForSession(sessionId: "random"), false)
+    XCTAssertEqual(localSampler.shouldSendEventForSession(sessionId: "anyEvent"), false)
+  }
+
+  /// Validates if the events are allowed when the sampling rate is One.
+  func test_AllowsEventCollection_samplingRateOne() {
```
leotianlizhan, line 35: Camel-case to follow same convention as other tests.

leotianlizhan (commented):
Overall looks good, just some nitpicks!

FirebaseSessions/Tests/Unit/SessionSamplerTests.swift
visumickey: Done.

FirebaseSessions/Sources/FirebaseSessionsError.swift
visumickey: This Error is just an error theme. Every error will have a module information and error description. This class defines just the module and not the actual message. So, this is just to reference that it is a "Sampling" related error. So, I would not worry about "event dropped message" in this context.

ncooke3 (approved):

leotianlizhan (approved):

PR #10518, opened 2022-11-23
Fix strict concurrency checking errors

paulb777:
- Fix Swift 6 errors generated with Build Setting: `Strict Concurrency Checking -> Complete`
- Consistent strong and weak `self` accessing
- Explicit `self` retains where necessary to ensure operations, including server uploads and completion blocks execute

### DIFF
--- a/FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
+++ b/FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
@@ -44,34 +44,34 @@ internal class StorageDeleteTask: StorageTask, StorageTaskManagement {
    * Prepares a task and begins execution.
    */
   internal func enqueue() {
-    weak var weakSelf = self
-    dispatchQueue.async {
-      guard let strongSelf = weakSelf else { return }
-      strongSelf.state = .queueing
-      var request = strongSelf.baseRequest
+    let completion = taskCompletion
+    taskCompletion = { (error: Error?) in
+      completion?(error)
+      // Reference self in completion handler in order to retain self until completion is called.
+      self.taskCompletion = nil
+    }
+    dispatchQueue.async { [weak self] in
+      guard let self = self else { return }
+      self.state = .queueing
+      var request = self.baseRequest
       request.httpMethod = "DELETE"
-      request.timeoutInterval = strongSelf.reference.storage.maxOperationRetryTime
+      request.timeoutInterval = self.reference.storage.maxOperationRetryTime
 
-      let callback = strongSelf.taskCompletion
-      strongSelf.taskCompletion = nil
-
-      let fetcher = strongSelf.fetcherService.fetcher(with: request)
+      let fetcher = self.fetcherService.fetcher(with: request)
       fetcher.comment = "DeleteTask"
-      strongSelf.fetcher = fetcher
+      self.fetcher = fetcher
 
-      strongSelf.fetcherCompletion = { (data: Data?, error: NSError?) in
+      self.fetcherCompletion = { [weak self] (data: Data?, error: NSError?) in
+        guard let self = self else { return }
         if let error = error, self.error == nil {
-          self.error = StorageErrorCode.error(withServerError: error, ref: strongSelf.reference)
+          self.error = StorageErrorCode.error(withServerError: error, ref: self.reference)
         }
-        callback?(self.error)
+        self.taskCompletion?(self.error)
         self.fetcherCompletion = nil
       }
 
-      strongSelf.fetcher?.beginFetch { data, error in
-        let strongSelf = weakSelf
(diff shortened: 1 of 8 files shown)


FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
```diff
@@ -44,34 +44,34 @@ internal class StorageDeleteTask: StorageTask, StorageTaskManagement {
...
-    dispatchQueue.async {
-      guard let strongSelf = weakSelf else { return }
-      strongSelf.state = .queueing
-      var request = strongSelf.baseRequest
+    let completion = taskCompletion
+    taskCompletion = { (error: Error?) in
+      completion?(error)
+      // Reference self in completion handler in order to retain self until completion is called.
+      self.taskCompletion = nil
+    }
```
morganchen12, line 52: Instead of doing this, why not just let the `dispatchQueue.async` closure capture `self` strongly? (Here and throughout)

FirebaseStorage/Sources/StorageReference.swift
```diff
@@ -705,4 +653,26 @@ import Foundation
     }
     return nil
   }
+
+  private func startAndObserveUploadTask(task: StorageUploadTask,
+                                         completion: ((_: StorageMetadata?, _: Error?) -> Void)?) {
+    if let completion = completion {
```
morganchen12, line 659: I realize this is not a behavior change from the previous code, but should we enqueue the task regardless of whether or not there's an observation callback? As the code currently is, if a user tries an upload with no observers, the upload will never actually begin.

FirebaseStorage/Sources/StorageUploadTask.swift
```diff
@@ -36,21 +36,20 @@ import Foundation
...
    */
   @objc open func enqueue() {
-    weak var weakSelf = self
-    dispatchQueue.async {
-      guard let strongSelf = weakSelf else { return }
-      if let contentValidationError = strongSelf.contentUploadError() {
-        strongSelf.error = contentValidationError
-        strongSelf.finishTaskWithStatus(status: .failure, snapshot: strongSelf.snapshot)
+    // Capturing self so that the upload is done whether or not there is a callback.
+    dispatchQueue.async { [self] in
```
morganchen12, line 40: optional nit: `[self] in` is the default capture and can be omitted. I'm fine with leaving this as-is though.

morganchen12 (commented):
Overall LGTM, had a few small questions.

FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
paulb777: I tried that, but it's not sufficient for self to get deallocated before entering outer `enqueue` `dispatchAsync`

FirebaseStorage/Sources/StorageReference.swift
paulb777: It does get called whether or not there is a completion handler.

The subsequent braces are end of function and end of class. 

The end of class one makes it confusing...

FirebaseStorage/Sources/StorageUploadTask.swift
paulb777: Understood. I wanted to be explicit to contrast with the other files.

paulb777 (commented):
Thanks for the review!

FirebaseStorage/Sources/Internal/StorageDeleteTask.swift
andrewheard: I _think_ that would create a retain-cycle because `self` has a strong reference to `dispatchQueue` in [`StorageTask`](https://github.com/firebase/firebase-ios-sdk/blob/3d51280713332af3f023abc7e305541871c1e005/FirebaseStorage/Sources/StorageTask.swift#L80) and `dispatchQueue` would then hold a strong reference back to `self` as well.

paulb777: Here's the backtrace from the `deinit`: 

Screen Shot 2022-11-23 at 3 56 38 PM

andrewheard, line 52: Comment (no action required): Although this workaround for retaining the completion handler is arguably less readable at first glance, I couldn't come up with a cleaner alternative and think we should merge with this approach.

Your inline comments "`Reference self in completion handler in order to retain self until completion is called`", and using the same pattern throughout, address my readability concerns.

andrewheard (approved):
LGTM, Paul! Thank you for preparing us for the inevitable day when Strict Concurrency Checking is enabled by default and also using a consistent approach in all the files.

PR #10519, opened 2022-11-23
Add jobs to build quickstart apps and run tests on FTL

sunmou99:
Added jobs that build quickstart apps and run test on [Firebase Test Lab](https://firebase.google.com/docs/test-lab/ios/run-xctest).

Other fixes:
Removed redundant checked out steps.
Kept `auth` and `messaging` not run quickstart tests on simulators (because there are failures)
Enabled performance quickstart swift test.

TODO:
Will remove `pull_request` trigger in this PR.
Some quickstart apps have issue running on FTL. They are currently disabled. Will create ticket and investigate the issue.

### Testing
Please refer the checks (workflow jobs) named `${product}-ftl`
Test are running on FTL's default iOS devices. And it's possible to set a different devices to run tests. ([Available devices](https://firebase.google.com/docs/test-lab/ios/available-testing-devices))

### DIFF
--- a/.github/workflows/abtesting.yml
+++ b/.github/workflows/abtesting.yml
@@ -77,7 +77,6 @@ jobs:
     runs-on: macos-12
     steps:
     - uses: actions/checkout@v3
-    - uses: actions/checkout@v3
     - uses: ruby/setup-ruby@v1
     - name: Setup quickstart
       env:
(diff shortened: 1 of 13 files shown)


scripts/test_quickstart_ftl.sh
```diff
@@ -0,0 +1,36 @@
+# Copyright 2019 Google
```
paulb777, line 1: 2022 Google LLC

.github/workflows/remoteconfig.yml
```diff
@@ -123,6 +122,30 @@ jobs:
     - name: Test Swift Quickstart
       run: ([ -z $plist_secret ] || scripts/third_party/travis/retry.sh scripts/test_quickstart.sh Config true)
 
+  # quickstart-ftl:
```
paulb777, line 125: Please add a comment about why tests are commented out

.github/workflows/abtesting.yml
```diff
@@ -91,6 +90,35 @@ jobs:
         LEGACY: true
       run: ([ -z $plist_secret ] || scripts/third_party/travis/retry.sh scripts/test_quickstart.sh ABTesting true)
 
+  quickstart-ftl:
+    # Don't run on private repo unless it is a PR.
+    if: (github.repository == 'Firebase/firebase-ios-sdk' && github.event_name == 'schedule') || github.event_name == 'pull_request'
```
paulb777, line 95: I believe this test should be 
```
 if: github.repository == 'Firebase/firebase-ios-sdk' 
```
since I don't think it will work from a forked repo since secrets won't be available

sunmou99: Updated the comment to `# Don't run on private repo.`
I will remove the `pull_request` trigger later. Currently, I keep it to show the test result.
Another question, should I change the job name to `quickstart-ftl-cron-only`?

paulb777: Yes, making that name change would help with clarity.

Please remove the `pull_request` trigger before merging.

paulb777 (approved):

PR #10530, opened 2022-11-29
Settings downloader

leotianlizhan:
#no-changelog


FirebaseSessions/Sources/SettingsDownloadClient.swift
```diff
@@ -0,0 +1,57 @@
...
+  private func buildURL(appInfo: ApplicationInfoProtocol) -> URL? {
+    var components = URLComponents()
+    components.scheme = "https"
+    components.host = "firebase-settings.crashlytics.com"
+    components.path = "/spi/v2/platforms/android/gmp/\(appInfo.appID)/settings"
+    components.queryItems = [
+      URLQueryItem(name: "build_version", value: appInfo.appBuildVersion),
+      URLQueryItem(name: "display_version", value: appInfo.appDisplayVersion),
+      URLQueryItem(name: "instance", value: "abadc0de"),
+      URLQueryItem(name: "source", value: "4"), // 4 is enum value for AppStore
```
samedson, line 53: So here's the info on `instance` and `source`.

`instance` is the hash of the app. If you change the app's source code and rebuild without chainging the app version, this value will change. The code to do this is complex and poorly written in Crashlytics [code](https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/Crashlytics/Settings/Models/FIRCLSApplicationIdentifierModel.m#L85-L136)

 - I would recommend we leave this out for now and consider deprecating it in Crashlytics Settings. As far as I can tell we don't use it to respond with Settings. For this PR feel free to leave it off and I'll add a bug for us to dig deeper with the Crashlytics team
 
 `source` is the install source. It's [super limited](https://github.com/firebase/firebase-ios-sdk/blob/master/Crashlytics/Crashlytics/Components/FIRCLSApplication.h#L28-L32). iCore tracks this data separately via platform logging, so I don't think it's worth us uploading it here.

 - For this PR, also feel free to leave this off

FirebaseSessions/Sources/SettingsDownloadClient.swift
```diff
@@ -0,0 +1,99 @@
...
+        var request = URLRequest(url: url)
+        request.setValue("application/json", forHTTPHeaderField: "Accept")
+        request.setValue(fiid, forHTTPHeaderField: "X-Crashlytics-Installation-ID")
+        request.setValue(appInfo.deviceModel, forHTTPHeaderField: "X-Crashlytics-Device-Model")
+        request.setValue(
+          FIRSESGetSysctlEntry("kern.osversion"),
+          forHTTPHeaderField: "X-Crashlytics-OS-Build-Version"
+        )
+        request.setValue(
+          GULAppEnvironmentUtil.systemVersion(),
```
samedson, line 89: To make these available in case they are useful to other parts of our code, I would recommend putting `GULAppEnvironmentUtil.systemVersion()` and `FIRSESGetSysctlEntry("kern.osversion")` into ApplicationInfo.

FirebaseSessions/Sources/FirebaseSessionsError.swift
```diff
@@ -20,4 +20,6 @@ enum FirebaseSessionsError: Error {
   case SessionSamplingError
   /// Firebase Installation ID related error
   case SessionInstallationsError
+  /// Settings related error
+  case SettingsError(String)
```
samedson, line 24: It probably isn't applicable yet, but I believe normally Swift recommends having a separate case for each error, so it's a bit more granular than Java where you have a class of errors with an error message that differentiates them. The reason Swift recommends granular enum cases is so you can add a switch on the error later on. We don't have a use case for that yet, but we could see a situation where we want to retry a 400-level error, but not a parsing error, for example.

I don't think it's necessary to make that change now but thought I'd explain.

FirebaseSessions/Sources/SettingsDownloadClient.swift
```diff
@@ -0,0 +1,99 @@
...
+              completion(.success(dict))
+            } else {
+              completion(.failure(FirebaseSessionsError
+                  .SettingsError("Failed to parse JSON to dictionary")))
+            }
+          } else if let error = error {
+            completion(.failure(FirebaseSessionsError.SettingsError(error.localizedDescription)))
+          }
+        }
+        task.resume()
```
samedson, line 57: I'm curious what this line is for. Could we add a comment too?

samedson: Ah I understand now: https://developer.apple.com/documentation/foundation/nsurlsessiontask/1411121-resume - this seems to just start the download

If we could add a comment that might help our future selves

FirebaseSessions/Sources/SettingsCacheClient.swift
```diff
@@ -25,9 +25,9 @@ struct CacheKey: Codable {
 /// SettingsCacheClient is responsible for accessing the cache that Settings are stored in.
 protocol SettingsCacheClient {
   /// Returns in-memory cache content in O(1) time
-  var cacheContent: [String: Any]? { get }
+  var cacheContent: [String: Any]? { get set }
```
samedson, line 28: I'm curious if we can pull this question mark off of this type by returning a default value if the `cache.dictionary` is null. It might save us some question marks in other places? Just an idea.

FirebaseSessions/Sources/SettingsDownloadClient.swift
```diff
@@ -0,0 +1,99 @@
...
+
+  init(appInfo: ApplicationInfoProtocol, identifiers: IdentifierProvider,
+       installations: InstallationsProtocol) {
+    self.appInfo = appInfo
+    self.identifiers = identifiers
+    self.installations = installations
+  }
+
+  func fetch(completion: @escaping (Result<[String: Any], Error>) -> Void) {
+    guard let url = url else {
```
samedson, line 37: For readability purposes, I might recommend not shadowing the `url` private getter with the url variable. Maybe name the variable `validURL`?

FirebaseSessions/Sources/SettingsDownloadClient.swift
```diff
@@ -0,0 +1,99 @@
...
+    components.path = "/spi/v2/platforms/android/gmp/\(appInfo.appID)/settings"
+    components.queryItems = [
+      URLQueryItem(name: "build_version", value: appInfo.appBuildVersion),
+      URLQueryItem(name: "display_version", value: appInfo.appDisplayVersion),
+    ]
+    return components.url
+  }
+
+  private func buildRequest(url: URL, completion: @escaping (Result<URLRequest, Error>) -> Void) {
+    installations.installationID { [self] result in
```
samedson, line 77: I might recommend pulling this out into the `fetch` method. This may simplify our code so we can make `buildRequest` synchronous and have it just return a value.  It might also be a bit more readable - I had not envisioned `buildRequest` would be getting FIIDs so moving FIID to `fetch` kindof puts it in our face more so to speak so we stay aware of it.

FirebaseSessions/Sources/SettingsDownloadClient.swift
leotianlizhan: Done

FirebaseSessions/Sources/SettingsCacheClient.swift
leotianlizhan: Good point! Changed to return empty dictionary.

FirebaseSessions/Sources/SettingsDownloadClient.swift
leotianlizhan: I'm not entirely sure what you mean, is it something like 
```
func fetch(...) {
  guard let url = url else {
    return
  }
  
  installations.installationID { result in
    switch result {
    case let .success(fiid):
      let request = buildRequest(url: url, installationID: fiid)
      let task = URLSession.shared.dataTask(with: request) { data, response, error in
          if let data = data {
            if let dict = try? JSONSerialization.jsonObject(with: data) as? [String: Any] {
              completion(.success(dict))
            } else {
              completion(.failure(FirebaseSessionsError
                  .SettingsError("Failed to parse JSON to dictionary")))
            }
          } else if let error = error {
            completion(.failure(FirebaseSessionsError.SettingsError(error.localizedDescription)))
          }
      }
      task.resume()
    case let .failure(error): 
      completion(.failure(FirebaseSessionsError.SettingsError(error.localizedDescription)))
    }
  }
}
```

FirebaseSessions/Sources/FirebaseSessionsError.swift
leotianlizhan: That's helpful to know. I guess I added the separation between each error in the `String` description passed in when creating the error. Is it preferable then to create individual errors for each case and not add any additional description?

FirebaseSessions/Sources/SettingsDownloadClient.swift
leotianlizhan: Done

FirebaseSessions/Sources/SettingsDownloadClient.swift
leotianlizhan: Done

FirebaseSessions/Sources/Settings.swift
```diff
@@ -32,47 +33,77 @@ class Settings: SettingsProtocol {
...
+       cache: SettingsCacheClient = SettingsCache()) {
     self.appInfo = appInfo
+    self.cache = cache
+    self.downloader = downloader
+  }
+
+  func fetchAndCacheSettings(currentTime: Date) {
+    // Only fetch if cache is expired, otherwise do nothing
+    guard isCacheExpired(currentTime: currentTime) else {
+      Logger.logDebug("[Settings] Cache is not expired, no fetch will be made.")
```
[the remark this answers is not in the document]
samedson: I think we can leave it as debug. In Crashlytics we don't print it every run and it's been fine

FirebaseSessions/Sources/SettingsCacheClient.swift
```diff
@@ -41,20 +41,34 @@ class SettingsCache: SettingsCacheClient {
   private let cache: UserDefaults = .standard
 
   /// Converting to dictionary is O(1) because object conversion is O(1)
-  var cacheContent: [String: Any]? {
-    return cache.dictionary(forKey: SettingsCache.content)
+  var cacheContent: [String: Any] {
+    get {
+      return cache.dictionary(forKey: SettingsCache.content) ?? [:]
```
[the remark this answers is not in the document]
samedson: So I believe the reason it's named content is cause there's a key that we cache and the settings content itself that we cache. We may want to call this `contentKey` and the other `cacheKeyKey` or something less clunky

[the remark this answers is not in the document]
samedson: Another way we could maybe name this is by moving the keys into their own enum, eg:

```
enum SettingsCacheKey: String {
  static let forContent = ...
  static let forKey = ...
}
```

FirebaseSessions/Sources/SettingsDownloadClient.swift
samedson: Yes this is exactly it 👍 To me the advantage is we've moved all our async code to one spot so it's easier to see the different paths to calling `completion`

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -86,6 +93,8 @@ protocol SessionsProvider {
 
     super.init()
 
+    self.settings.fetchAndCacheSettings(currentTime: Date())
```
samedson, line 96: Yeah to Visu's point, we probably want this to attempt each time a session starts. We can do this by just moving this call to beginListening

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -100,4 +104,12 @@ class ApplicationInfo: ApplicationInfoProtocol {
   var appDisplayVersion: String {
     return Bundle.main.infoDictionary?["CFBundleVersion"] as? String ?? ""
   }
+
+  var osBuildVersion: String {
+    return FIRSESGetSysctlEntry("kern.osversion") ?? ""
+  }
```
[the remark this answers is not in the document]
leotianlizhan: This is in `GULAppEnvironmentUtil`, but it's not a public method, that's why I had to copy it.

FirebaseSessions/Sources/Settings.swift
```diff
@@ -32,47 +33,77 @@ class Settings: SettingsProtocol {
...
+        )
+      case let .failure(error):
+        Logger.logError("[Settings] Fetching newest settings failed with error: \(error)")
+      }
+    }
   }
 
-  func isCacheExpired(currentTime: Date) -> Bool {
-    guard cache.cacheContent != nil else {
+  private func isCacheExpired(currentTime: Date) -> Bool {
```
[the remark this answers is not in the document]
leotianlizhan: Tried changing it but it breaks tests because this is a private method, tests needs to pass down a fake date to every method. Considering that and since this is a private method only used by `fetch` I think it's fine to keep it no-defaults.

FirebaseSessions/Sources/Settings.swift
```diff
@@ -32,47 +33,77 @@ class Settings: SettingsProtocol {
...
+
+  init(appInfo: ApplicationInfoProtocol,
+       downloader: SettingsDownloadClient,
+       cache: SettingsCacheClient = SettingsCache()) {
     self.appInfo = appInfo
+    self.cache = cache
+    self.downloader = downloader
+  }
+
+  func fetchAndCacheSettings(currentTime: Date) {
```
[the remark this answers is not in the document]
leotianlizhan: Done, but it was a bit tricky to do due to the protocol. I had to add an extension.

FirebaseSessions/Sources/SettingsDownloadClient.swift
leotianlizhan: Done 👍

FirebaseSessions/Sources/SettingsCacheClient.swift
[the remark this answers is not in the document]
leotianlizhan: Love the suggestion to use enum to namespace it! Changed to `UserDefaultsKeys.forContent` and `UserDefaultsKeys.forCacheKey`.

FirebaseSessions/Tests/Unit/SettingsTests.swift
```diff
@@ -21,176 +21,237 @@ class SettingsTests: XCTestCase {
...
+  var settings: SettingsProtocol!
 
   override func setUp() {
+    super.setUp()
     appInfo = MockApplicationInfo()
+    downloader = MockSettingsDownloader(successResponse: validSettings)
     cache = SettingsCache()
-    cache.removeCache()
-    settings = Settings(cache: cache, appInfo: appInfo)
+    cache.removeCache() // just resintantiating cache isn't enough because of persistence
```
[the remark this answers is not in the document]
leotianlizhan: Done

leotianlizhan:
> 1. When do we initiate a fetch? Currently this seems to be initiated during SDK initialization. Ideally this check should happen during the start of every session.
Good point, changed to attempt fetch at the start of each session.
> 2. We need to bring in the logic to the fetcher to see if the cache is too old (Cache TTL) before initiating a fetch.
Cache TTL check is done in the highest layer (`Settings.swift`, almost like "resolver layer" in Fireperf), this is because `SettingsCacheClient` is **only** responsible for persisting cache as well as the cache's TTL metadata (`CacheKey`); while `SettingsDownloadClient` is **only** responsible for sending and receiving the network-request. I think this has better separation-of-concerns than moving cache TTL to "downloader layer".

FirebaseSessions/Sources/FirebaseSessionsError.swift
samedson: I think you'd do both. That way you get both a nice description and a way to do a switch on it.


Not something necessary for this PR

samedson (approved):
LGTM! Awesome job on a big* PR

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -87,6 +94,7 @@ protocol SessionsProvider {
     super.init()
 
     self.initiator.beginListening {
+      self.settings.fetchAndCacheSettings()
```
visumickey, line 97: NIT: Thanks for bringing it in. Can you add a comment to state that we start with updating the remote configurations?

FirebaseSessions/Sources/Settings.swift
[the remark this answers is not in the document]
visumickey: Aah! Got it. I see the constraint now. Makes sense. Thanks for doing it.

FirebaseSessions/Sources/Settings.swift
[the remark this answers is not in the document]
visumickey: Sounds good. We can look at this holistically across Firefit products. Added a task to attend on this.

FirebaseSessions/Sources/Settings.swift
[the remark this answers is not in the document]
visumickey: Sounds good. In that my only (nit) suggestion would be to name the parameter as just `time` instead of `currentTime`.

visumickey (approved):

FirebaseSessions/Sources/Settings.swift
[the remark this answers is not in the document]
leotianlizhan: Done

FirebaseSessions/Sources/FirebaseSessions.swift
leotianlizhan: Done

visumickey (approved):

PR #10536, opened 2022-11-29
Fix Firestore Changelog for 10.3.0

paulb777:
It looks like MultiDB support was added last week and missed the 10.2.0 release.


tom-andersen (approved):
Thanks!

PR #10539, opened 2022-11-30
Add NetworkInformation to session start events.

visumickey:
#no-changelog


leotianlizhan (commented):
Overall LGTM, but "`AppInfo` having a 1-to-1 copy of `NetworkInfo` fields, exposing properties of an object that it owns" feels kind of weird to me. I don't know if this is a well known pattern and if it's good practice or not. I tried to Google/Moma this to find guidance but couldn't find anything. Any suggestions @ncooke3 ?

visumickey:
> Overall LGTM, but "`AppInfo` having a 1-to-1 copy of `NetworkInfo` fields, exposing properties of an object that it owns" feels kind of weird to me. I don't know if this is a well known pattern and if it's good practice or not. I tried to Google/Moma this to find guidance but couldn't find anything. Any suggestions @ncooke3 ?

I like this approach. Look at NetworkInfo as a network information data provider to ApplicationInfo. ApplicationInfo is a packager of many application information including network state and packaging in SessionEvents. Secondly, this is also inline with the already existing pattern that I do not want to change it in this PR.

FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -147,4 +151,65 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
...
+    @unknown default:
+      return firebase_appquality_sessions_NetworkConnectionInfo_NetworkType_DUMMY
+    }
+  }
+
+  private func convertMobileSubtype(mobileSubtype: String)
+    -> firebase_appquality_sessions_NetworkConnectionInfo_MobileSubtype {
+    var subtype: firebase_appquality_sessions_NetworkConnectionInfo_MobileSubtype
+
+    #if TARGET_OS_IOS
```
samedson, line 173: Watch out - this is an Objective-C macro, so these tests aren't being run. The equivalent macro for Swift is `#if os(iOS)`

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -183,4 +185,200 @@ class SessionStartEventTests: XCTestCase {
...
+        XCTAssertEqual(
+          event.proto.application_info.apple_app_info.network_connection_info.network_type,
+          expected
+        )
+      }
+    }
+  }
+
+  /// Following tests can be run only in iOS environment
+  #if TARGET_OS_IOS
```
samedson, line 226: Same here and for the next test about the ObjC vs Swift macros. I think these tests are failing because of the of the macro issue in `SessionStartEvent.swift`

FirebaseSessions/Sources/SessionStartEvent.swift
visumickey: I tried applying the the Swift Macro - But that did not really work. So, had to settle down with the macro `TARGET_OS_IOS`.

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -183,4 +185,200 @@ class SessionStartEventTests: XCTestCase {
...
+        .WIFI,
+        firebase_appquality_sessions_NetworkConnectionInfo_NetworkType_WIFI
+      ),
+      (
+        .mobile,
+        firebase_appquality_sessions_NetworkConnectionInfo_NetworkType_MOBILE
+      ),
+      (
+        .none,
+        firebase_appquality_sessions_NetworkConnectionInfo_NetworkType_DUMMY
```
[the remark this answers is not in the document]
visumickey: Interesting reasons:

1. In the encodeDecode method we use, we have a list of fields that are used for decoding. Currently the decode method uses the sessions_proto fields and so the network information (currently served by the clientanalytics proto) was never decoded. So, the option was to add the fields of both the proto to the decode logic, but that does not seem trivial.
2. ClientAnalytics proto has very minimal changes - just the network message (Nothing related to sessions). It had just the enums for NetworkType, MobileSubtype - so felt it does not need to have two separate protos.

For the above reasons, I decided to unify the protos to be the same. But the cons on this approach is that the proto2 messages support optionals "NONE" as a value. With proto3, there are no optional values - So, chose to use DUMMY as a value.

Happy to hear your thoughts!

FirebaseSessions/Sources/SessionStartEvent.swift
ncooke3: what issue did you run into with Sam's suggestion? I don't think TARGET_OS_IOS expands correctly in a Swift context because, as Sam said, it's an Objective-C macro.

ncooke3: nevermind-- saw they were changed in 505a108!

ncooke3:
> > Overall LGTM, but "`AppInfo` having a 1-to-1 copy of `NetworkInfo` fields, exposing properties of an object that it owns" feels kind of weird to me. I don't know if this is a well known pattern and if it's good practice or not. I tried to Google/Moma this to find guidance but couldn't find anything. Any suggestions @ncooke3 ?
> 
> I like this approach. Look at NetworkInfo as a network information data provider to ApplicationInfo. ApplicationInfo is a packager of many application information including network state and packaging in SessionEvents. Secondly, this is also inline with the already existing pattern that I do not want to change it in this PR.

I think instead of re-exposing the properties of of an object it owns, we should instead just add a property `networkInfo` on `ApplicationInfo`. Since we are constructing model data here, I think we should leverage creating a hierarchy instead bubbling re-exposing fields. 

```diff
protocol ApplicationInfoProtocol {
-  /// Type of network the device is connected to
-  var networkType: GULNetworkType { get }
-
-  /// Mobile subtype the device is connected to
-  var mobileSubtype: String { get }
-
-  var networkType: GULNetworkType {
-    return networkInfo.networkType
-  }
-
-  var mobileSubtype: String {
-    return networkInfo.mobileSubtype
-  }
+  /// The application's network info.
+  var network: NetworkInfo { get } 
}
```


Additionally, while it wasn't added in this PR, `NetworkInfo` is a reference type. This seems like a good candidate for a value type?

visumickey:
> > > @leotianlizhan Overall LGTM, but "`AppInfo` having a 1-to-1 copy of `NetworkInfo` fields, exposing properties of an object that it owns" feels kind of weird to me. I don't know if this is a well known pattern and if it's good practice or not. I tried to Google/Moma this to find guidance but couldn't find anything. Any suggestions @ncooke3 ?
> > 
> > 
> > I like this approach. Look at NetworkInfo as a network information data provider to ApplicationInfo. ApplicationInfo is a packager of many application information including network state and packaging in SessionEvents. Secondly, this is also inline with the already existing pattern that I do not want to change it in this PR.
> 
> I think instead of re-exposing the properties of of an object it owns, we should instead just add a property `networkInfo` on `ApplicationInfo`. Since we are constructing model data here, I think we should leverage creating a hierarchy instead of re-exposing fields.
> 
> ```diff
> protocol ApplicationInfoProtocol {
> -  /// Type of network the device is connected to
> -  var networkType: GULNetworkType { get }
> -
> -  /// Mobile subtype the device is connected to
> -  var mobileSubtype: String { get }
> -
> +  /// The application's network info.
> +  var network: NetworkInfo { get } 
> }
> ```
> 
> Additionally, while it wasn't added in this PR, `NetworkInfo` is a reference type. This seems like a good candidate for a value type? Was this considered for `ApplicationInfo` as well?

That is a good suggestion @ncooke3 . I made the change. About the value type vs reference type, I totally agree - We just use reference type in all the places which is less than ideal. I'm working on a separate PR to start using value types more than reference types. I've kept it out of scope of the current PR.

ncooke3 (approved):

samedson (approved):
:shipit:

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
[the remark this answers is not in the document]
samedson: Thank you for the info - this all makes sense 👍

PR #10541, opened 2022-11-30
[Sessions] Use deviceSimulatorModel only for compatibility with Crashlytics

samedson:
- `deviceSimulatorModel` can work on simulators and devices.
 - It will also have similar outputs as `deviceModel` on a device.

Crashlytics will only use `deviceSimulatorModel`, so this PR updates Sessions to do the exact same thing.

#no-changelog


visumickey (approved):
Thanks for doing this change. In that case, is the deviceModel API even necessary with GULAppEnvironmentUtil?

samedson:
That API actually predates AQS - I believe it's used by other SDKs.

PR #10545, opened 2022-11-30
Remove installations ID from Identifiers and move it to SessionCoordinator

visumickey:
Currently FirebaseInstallationsId is generated in Identifiers synchronously. But, this has lead to situations where FirebaseInstallationsID missing in the session start event (since the current time out of 1.0 does not add predictability).

The current PR is attempting to make the logEvent action asynchronous where we wait for the installationID to be available before we decide to send the event to Firelog. SessionsCoordinator receives the session start event, waits for installations ID until the callback from InstallationsService happens, then ships the event to Firelog.

#no-changelog


FirebaseSessions/Sources/SessionCoordinator.swift
```diff
@@ -20,33 +20,50 @@ import Foundation
...
+                  "Error logging Session Start event to GoogleDataTransport: \(error)."
+                )
+              callback(.failure(error))
+            }
+          }
         case let .failure(error):
           Logger
             .logError(
-              "Error logging Session Start event to GoogleDataTransport: \(error)."
+              "Error getting Firebase Installation ID: \(error). Skip sending event."
```
samedson, line 65: Can we `callback` the failure here so it gets bubbled to the top?

visumickey: Ha, got it. Good catch. Added the callback.

samedson (approved):
Ship it!

FirebaseSessions/Sources/SessionCoordinator.swift
```diff
@@ -20,33 +20,51 @@ import Foundation
...
     self.sampler = sampler
   }
 
   // Begins the process of logging a SessionStartEvent to FireLog, while taking into account Data Collection, Sampling, and fetching Settings
   func attemptLoggingSessionStart(event: SessionStartEvent,
                                   callback: @escaping (Result<Void, Error>) -> Void) {
+    /// Order of execution
+    /// 1. Check if the session can be sent. If yes, move to 2. Else, drop the event.
+    /// 2. Fetch the installations Id. If successful, move to 3. Else, drop sending the event.
+    /// 3. Log the event. If successful, all is good. Else, log the message with error.
```
samedson, line 43: Love the comment 😎

leotianlizhan (approved):
LGTM!

PR #10550, opened 2022-12-02
Run Performance Swift QS tests in CI

paulb777:
Fix a typo that was running the ObjC tests twice and never the Swift tests

Thanks to @sunmou99 for pointing this out


visumickey (approved):
Thanks for catching this @sunmou99 and thanks to @paulb777 for fixing it.

PR #10552, opened 2022-12-02
[Messaging] Standalone Watch App Sample

aashishpatil-g:

### Discussion

Sample Standalone Watch App for testing Firebase Messaging on watchOS


.gitignore
```diff
@@ -154,3 +154,4 @@ FirebaseAppCheck/Apps/AppCheckCustomProvideApp/AppCheckCustomProvideApp/GoogleSe
 /Example/FirestoreSample/FirestoreSample/GoogleService-Info.plist
 /Example/FirestoreSample/ui-debug.log
 /Example/FirestoreSample/firestore-debug.log
+FirebaseMessaging/Apps/SampleStandaloneWatchApp/SampleStandaloneWatchApp Watch App/GoogleService-Info.plist
```
charlotteliang, line 157: Can we just use this one?

https://github.com/firebase/firebase-ios-sdk/blob/master/.gitignore#L27

FirebaseMessaging/Apps/SampleStandaloneWatchApp/SampleStandaloneWatchAppWatchApp/SampleStandaloneWatchAppApp.swift
```diff
@@ -0,0 +1,55 @@
...
+    notifCenter.requestAuthorization(options: [.alert, .sound]) { granted, error in
+      if granted {
+        WKApplication.shared().registerForRemoteNotifications()
+      }
+    }
+    Messaging.messaging().delegate = self
+  }
+
+  func didRegisterForRemoteNotifications(withDeviceToken deviceToken: Data) {
+    print("APNS didRegisterForRemoteNotifications. Got device token \(deviceToken)")
```
charlotteliang, line 43: Nit: cant we add a comment here:
    /// Swizzling should be disabled in Messaging for watchOS, set APNS token manually.

charlotteliang (approved):

FirebaseMessaging/Apps/SampleStandaloneWatchApp/SampleStandaloneWatchAppWatchApp/SampleStandaloneWatchAppApp.swift
peterfriese: +1 to that. Let's also please link to the instructions for disabling Swizzling, or alternatively write them down in the comment.

aashishpatil-g: Done - added the comment and a link to our documentation page.

.gitignore
aashishpatil-g: The path for this sample is different so had to add this entry. However, I've moved this line within the file so its in same place as (colocated with)  other Messaging gitignores.

charlotteliang (approved):
Looks good. Thanks!

FirebaseMessaging/Apps/SampleStandaloneWatchApp/SampleStandaloneWatchAppWatchApp/SampleStandaloneWatchAppApp.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+}
+
+// MARK: AppDelegate
+
+class FCMWatchAppDelegate: NSObject, WKApplicationDelegate, MessagingDelegate {
+  func applicationDidFinishLaunching() {
+    FirebaseApp.configure()
+    let notifCenter = UNUserNotificationCenter.current()
```
peterfriese, line 33: Let's use `notificationCenter` instead.

peterfriese (commented):
I think we should provide some guidance around the bundle ID, particularly whether or not developers need to use the `.watchkitapp` suffix, and whether or not this is case-sensitive.

I also added a README with some instructions - not all of these are part of the [docs](https://firebase.google.com/docs/cloud-messaging/ios/client).

paulb777 (commented):
Thanks for adding. 

Please add a build regression test to CI like https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/messaging.yml#L187

FirebaseMessaging/Apps/SampleStandaloneWatchApp/SampleStandaloneWatchAppWatchApp/SampleStandaloneWatchAppApp.swift
aashishpatil-g: Closing this since @peterfriese has already made the change

aashishpatil-g:
> Thanks for adding.
> 
> Please add a build regression test to CI like https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/messaging.yml#L187

Added the build regression test.

PR #10566, opened 2022-12-12
cmake: fail fast when execute_process() fails

dconeybe:
This PR modifies cmake builds to fail fast when calling cmake's built-in `execute_process()` function.

By default, `execute_process()` does not check the exit code of the child process. Therefore, if the child process fails then the cmake build process silently ignores the failure and carries on, usually causing strange errors later on. This PR makes it so that the build fails immediately upon execute_process() failing, making it much easier to diagnose the root problem.

### DIFF
--- a/cmake/external_rules.cmake
+++ b/cmake/external_rules.cmake
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


cherylEnkidu (approved):
LGTM! Thanks for the error message improving!

PR #10572, opened 2022-12-13
[Sessions] Add Swift Package Manager tests for Sessions

samedson:
### Changes to our Workflow

Going forward, we're going to have to change how we import some external libraries.

#### 1) Swift Importing ObjC
For Swift code that is importing any ObjC code, we'll have to import our Swift target when building under SPM:
```
#if SWIFT_PACKAGE
  import FirebaseSessionsObjC
#endif // SWIFT_PACKAGE
```

#### 2) Importing GoogleUtilities
When importing GoogleUtilities, we'll need the following import. This is because SPM seems to define different imports than Cocoapods.
```
#if SWIFT_PACKAGE
  import GoogleUtilities_Environment
#else
  import GoogleUtilities
#endif // SWIFT_PACKAGE
```

#### 3) Maintaining the Exclude Field:
We'll need to add any files here that aren't ObjC files whenever we add them:
https://github.com/firebase/firebase-ios-sdk/pull/10572/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR1119-R1127

### Multiple Targets, and new SourcesObjC folder
 - Swift package manager can't support targets with both ObjC and Swift in them. In addition, it's clunky to maintain the excludes list if you specify the target as our SDK's root folder `FirebaseSessions/`. Crashlytics does it and it's a lot of excludes to maintain: https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L499-L510 . Therefore, this PR moves all ObjC / C code to SourcesObjC, included the generated NanoPB files. 
     - I've updated the NanoPB generator script to use this new path

### check_imports Problem
I'm running into the following import check failure:

Screen Shot 2022-12-14 at 12 06 26 PM

I went and found the PR that it's from, and it seems to be from a time when we needed two different imports for SPM and Cocoapods for other libraries: https://github.com/firebase/firebase-ios-sdk/pull/5843

My problem is I think I need to have different import paths for SPM and Cocoapods *within my own code*, so this advice from the checker to use `<` for my imports doesn't seem applicable. This is because I need to define a separate target for the ObjC code, and the root of that target is `FirebaseSessions/SourcesObjC`. Therefore, my SPM imports leave off `FirebaseSessions` eg: https://github.com/firebase/firebase-ios-sdk/pull/10572/files#diff-7353704f2b1e643e6fe7b10a6586e1b05e36a4eca95a208f5afbf09351732a36R20-R30

**Resolution:** I've resolved this by moving the root of the ObjC target to the root of the SDK `FirebaseSessions/SourcesObjC` => `FirebaseSessions`: https://github.com/firebase/firebase-ios-sdk/pull/10572/commits/d91550c4e74c87dca7998c21936131c9dcc1efff, but it feels bad to have to maintain the long list of excludes.

### TODO
- [x] Regenerate the nanopb files and make sure they work under SPM

#no-changelog


FirebaseSessions/SourcesObjC/Protogen/nanopb/sessions.nanopb.c
```diff
@@ -17,8 +17,11 @@
 /* Automatically generated nanopb constant definitions */
 /* Generated by nanopb-0.3.9.9 */
 
+
 #include "sessions.nanopb.h"
 
+//#include "FirebaseSessions/SourcesObjC/Protogen/nanopb/sessions.nanopb.h"
```
samedson, line 23: This needs to be resolved - do we generate #ifs?

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -15,7 +15,15 @@
...
-@_implementationOnly import GoogleUtilities
+#if SWIFT_PACKAGE
+  import FirebaseSessionsObjC
+#endif // SWIFT_PACKAGE
+
+#if SWIFT_PACKAGE
+  import GoogleUtilities_Environment
+#else
+  import GoogleUtilities
+#endif // SWIFT_PACKAGE
```
samedson, line 26: The imports for SPM are a little clunky - I wonder if there's a cleaner way to do this

Package.swift
```diff
@@ -129,6 +129,10 @@ let package = Package(
       name: "FirebaseRemoteConfigSwift",
       targets: ["FirebaseRemoteConfigSwift"]
     ),
+    .library(
+      name: "FirebaseSessions",
+      targets: ["FirebaseSessions"]
+    ),
```
samedson, line 135: Will this library go out for release? If we'd like to keep the SDK unreleased, can we leave off the library but keep the target?

Maybe it's ok to leave the library because customers will have to depend on it directly?

FirebaseSessions/Sources/ApplicationInfo.swift
```diff
@@ -16,7 +16,16 @@
 import Foundation
 
 @_implementationOnly import FirebaseCore
-@_implementationOnly import GoogleUtilities
```
samedson, line 19: We've been adding `@_implementationOnly` to our imports of other libraries. Necessary or no? I don't see others doing it when I search for `import GoogleUtilities_` in swift code

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
visumickey: Can we extract some of these common inclusions to a separate file and import that? Would be helpful to centralize this clunky code?

FirebaseSessions/Sources/ApplicationInfo.swift
samedson: Gonna fix that in this PR for Sessions SDK

samedson: Interesting - it gets unhappy when I do that for the ObjC target
Screen Shot 2022-12-14 at 1 17 13 PM

Package.swift
```diff
@@ -129,6 +129,10 @@ let package = Package(
...
     .library(
       name: "FirebaseStorage",
       targets: ["FirebaseStorage"]
```
paulb777, line 138: Line 167 below: GoogleUtilities 7.10.0 is now required.

samedson: Ooh good catch - updated.

On a similar vein, I've been trying to figure out a way to get it to build for the lowest iOS / watchOS version we support instead of the highest.

FirebaseSessions/Sources/ApplicationInfo.swift
ncooke3: I think it is because using `@_implementationOnly XYZ` implies you do not intend to import the API from module XYZ and expose it publicly. Does `description` need to be public? If `firebase_appaquality_sessions_eventType` needs to be used outside of the `Sessions` target, than don't worry about using `@_implementationOnly` for this particular import.

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
ncooke3: I don't think that is feasible. Swift doesn't have macros-- something that would help here.

Package.swift
ncooke3: This SDK is for internal use only, right? If so, it shouldn't have a library– only a target.

FirebaseSessions/SourcesObjC/NanoPB/FIRSESNanoPBHelpers.h
```diff
@@ -16,6 +16,8 @@
 #ifndef FIRSESNanoPBHelpers_h
 #define FIRSESNanoPBHelpers_h
 
+#import <Foundation/Foundation.h>
```
visumickey, line 19: Why are we adding this now? Is this something we added manually?

FirebaseSessions/Tests/Unit/Mocks/MockInstallationsProtocol.swift
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import FirebaseInstallations
+@_implementationOnly import FirebaseInstallations
```
visumickey, line 16: Why is installations implementationOnly? Is this because it is purely internal and not exposed as public?

Package.swift
visumickey: AQS Library is internal - yes. Similar to Firelog, other libraries would depend on this similar to Crashlytics/Fireperf. Probably, we might have to do something like this. https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L155-L158

visumickey: Also, we might have have to bring in dependencies for this package? Like Installations and GoogleUtilities?

FirebaseSessions/SourcesObjC/NanoPB/FIRSESNanoPBHelpers.h
samedson: Yeah I was surprised by this, but for this to build under SPM it needed this import. I was getting a failure that like NSObject doesn't exist before adding this. I believe Cocoapods may be implicitly including it.

Package.swift
samedson: These dependencies are specified in the target, so we're already gtg there: https://github.com/firebase/firebase-ios-sdk/pull/10572/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR1093-R1097

To your comment above, we are ok leaving that out because `.package(` specifications are for libraries outside of the repo. Since we've put the `.target(` in for FirebaseSessions other SDKs can just depend on that target.

samedson: Going to remove this library specification

Package.swift
```diff
@@ -1080,6 +1080,66 @@ let package = Package(
...
+        .define("PB_FIELD_32BIT", to: "1"),
+        .define("PB_NO_PACKED_STRUCTS", to: "1"),
+        .define("PB_ENABLE_MALLOC", to: "1"),
+      ],
+      linkerSettings: [
+        .linkedFramework("Security"),
+        .linkedFramework("SystemConfiguration", .when(platforms: [.iOS, .macOS, .tvOS])),
+      ]
+    ),
+    .target(
```
visumickey, line 1108: This is an interesting way to bring in ObjC dependencies to Swift world as a separate target. Like the idea. Can we add comments about this for future references.

visumickey (approved):

leotianlizhan (approved):

FirebaseSessions/Tests/Unit/Mocks/MockInstallationsProtocol.swift
samedson: Yeah exactly - so it's not exposed to our users. Not a huge deal because we're just a dependency of Crash / Perf but is good practice.

FirebaseSessions/Sources/ApplicationInfo.swift
samedson: `description` needs to be public to the Sessions SDK, but not outside it. In fact, nothing in the Sessions SDK (yet) needs to be public because it's just going to be a dependency of Crash / Perf. So `firebase_appaquality_sessions_eventType` should only be public to the Sessions SDK, but not users of the Sessions SDK.

I'm going to submit this PR for now and maybe we can follow up later on tightening up access controls to the SDK.

Package.swift
samedson: Good call - added a comment!

PR #10573, opened 2022-12-14
Apply settings configs

leotianlizhan:
Apply the Settings config only once at the beginning of app-start to the rest of `FirebaseSessions` SDK:
1. enable/disable logging (b/257076205)
2. sampling rate (b/262250981)
3. session timeout (b/257076041)

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -55,13 +55,6 @@ protocol SessionsProvider {
     let fireLogger = EventGDTLogger(googleDataTransport: googleDataTransport!)
 
     let identifiers = Identifiers()
-    let coordinator = SessionCoordinator(
-      identifiers: identifiers,
-      installations: installations,
-      fireLogger: fireLogger,
-      sampler: SessionSampler()
-    )
-    let initiator = SessionInitiator()
     let appInfo = ApplicationInfo(appID: appID)
     let settings = Settings(
       appInfo: appInfo,
@@ -71,6 +64,14 @@ protocol SessionsProvider {
         installations: installations
       )
     )
+    let coordinator = SessionCoordinator(
+      identifiers: identifiers,
+      installations: installations,
+      fireLogger: fireLogger,
+      sampler: SessionSampler(sessionSamplingRate: settings.samplingRate),
+      loggingEnabled: settings.sessionsEnabled
+    )
+    let initiator = SessionInitiator(sessionTimeout: settings.sessionTimeout)
 
     self.init(appID: appID,
               identifiers: identifiers,
--- a/FirebaseSessions/Tests/Unit/InitiatorTests.swift
+++ b/FirebaseSessions/Tests/Unit/InitiatorTests.swift
@@ -18,9 +18,10 @@ import XCTest
 class InitiatorTests: XCTestCase {
   // 2021-11-01 @ 00:00:00 (EST)
   let date = Date(timeIntervalSince1970: 1_635_739_200)
+  let sessionTimeout: TimeInterval = 30 * 60 // 30 minutes
 
   func test_beginListening_initiatesColdStart() throws {
-    let initiator = SessionInitiator()
+    let initiator = SessionInitiator(sessionTimeout: sessionTimeout)
     var initiateCalled = false
     initiator.beginListening {
       initiateCalled = true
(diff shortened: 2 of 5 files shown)


FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -71,6 +64,14 @@ protocol SessionsProvider {
...
       )
     )
+    let coordinator = SessionCoordinator(
+      identifiers: identifiers,
+      installations: installations,
+      fireLogger: fireLogger,
+      sampler: SessionSampler(sessionSamplingRate: settings.samplingRate),
+      loggingEnabled: settings.sessionsEnabled
+    )
+    let initiator = SessionInitiator(sessionTimeout: settings.sessionTimeout)
```
visumickey, line 74: WIth `timeout` being passed in the initialization phase, we might be stuck at situations where the remote config changes would be applied only in subsequent cold start. I would recommend the configuration change being applied more instantaneously rather than delay. Can we pass the settings service here to the initiator and the initiator can apply the config value at the time of deciding a new session start?

FirebaseSessions/Tests/Unit/InitiatorTests.swift
```diff
@@ -31,7 +32,10 @@ class InitiatorTests: XCTestCase {
   func test_appForegrounded_initiatesNewSession() throws {
```
visumickey, line 32: Can we create a testcase where the timeout interval is used to generate a new session (the timeout interval can be set to a low value, say 1 second)?

FirebaseSessions/Tests/Unit/SessionCoordinatorTests.swift
```diff
@@ -34,11 +34,38 @@ class SessionCoordinatorTests: XCTestCase {
...
       installations: installations,
       fireLogger: fireLogger,
-      sampler: sampler
+      sampler: sampler,
+      loggingEnabled: true
     )
     sampler.sessionSamplingRate = 1.0
   }
 
+  func test_attemptLoggingSessionStart_logsNothingWhenDisabled() throws {
```
visumickey, line 43: Can we create test cases where the loggingEnabled value is toggled? Basically wanted to validate the situation where disabling logging, followed by enabling logging makes sure we are sending events? and vice versa?

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -71,6 +64,14 @@ protocol SessionsProvider {
         installations: installations
       )
     )
+    let coordinator = SessionCoordinator(
+      identifiers: identifiers,
+      installations: installations,
+      fireLogger: fireLogger,
+      sampler: SessionSampler(sessionSamplingRate: settings.samplingRate),
+      loggingEnabled: settings.sessionsEnabled
```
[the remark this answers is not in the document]
leotianlizhan: I think that depends on the behavior we want. Currently, by passing in individual primitive values in constructor, configs are only changed at the start of an app. Alternatively, we can pass in `Settings` and query the value when using it, which can lead to the values changing every session.

FirebaseSessions/Sources/FirebaseSessions.swift
leotianlizhan: That's what I'm unsure about. I believe our design stated updating configs at cold start, so I did it this way. Passing settings is totally possible too, but it needs to be that we intentionally want to possibly change configs every session.

FirebaseSessions/Tests/Unit/SessionCoordinatorTests.swift
leotianlizhan: All other test cases in this file should have `loggingEnabled` set to true

FirebaseSessions/Sources/FirebaseSessions.swift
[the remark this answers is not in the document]
visumickey: Letting @samedson to weigh-in on this. My recommendation would be to apply the most recent configs values downloaded rather than wait until the next cold start.

FirebaseSessions/Tests/Unit/SessionCoordinatorTests.swift
visumickey: Sure, that is true. I'm looking for test cases where the logging state is toggled at run-time. Eg: start with false, and then changes to true and then changes back to false and vice versa.

PR #10576, opened 2022-12-14
[Sessions] Use Notifications for Initiator tests

samedson:
Use the underlying notification, instead of calling the selector. 

This depends on https://github.com/firebase/firebase-ios-sdk/pull/10572

#no-changelog


leotianlizhan (approved):
LGTM, this test is much better in that it won't be broken when we make changes in `SessionInitiator`!

FirebaseSessions/Sources/SessionInitiator.swift
```diff
@@ -89,11 +89,11 @@ class SessionInitiator {
     #endif
   }
 
-  @objc func appBackgrounded() {
+  @objc private func appBackgrounded() {
```
visumickey, line 92: Thanks for restricting the access to this API. Curious if we need to do such an audit to the rest of the library to make sure we are drawing restrictions around classes clearly? Not a blocking comment - but something for following up.

visumickey (approved):

FirebaseSessions/Sources/SessionInitiator.swift
samedson: Yeah we should totally do it - let me add a task

PR #10579, opened 2022-12-15
Fix overlay patch mutation bug

wu-hui:


Firestore/core/src/model/overlayed_document.h
```diff
@@ -26,8 +26,10 @@ namespace firebase {
 namespace firestore {
 namespace model {
 
-/** Represents a local view (overlay) of a document, and the fields that are
- * locally mutated. */
+/**
+ * Represents a local view (overlay) of a document, and the fields that are //
```
ehsannas, line 30: extra `//` at the end of the line

Firestore/core/test/unit/testutil/testutil.cc
```diff
@@ -648,6 +648,21 @@ model::FieldIndex MakeFieldIndex(const std::string& collection_group,
...
+OverlayTypeMap(
+    std::vector<std::pair<model::DocumentKey, model::Mutation::Type>> pairs) {
+  std::unordered_map<model::DocumentKey, model::Mutation::Type,
+                     model::DocumentKeyHash>
+      result;
+  for (const auto& p : pairs) {
+    result.emplace(p.first, p.second);
+  }
+
+  return result;
```
ehsannas, line 663: Is this function necessary? I think you should be able to directly initialize the `unordered_map` using the initializer list syntax. See examples here: https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map

ehsannas (commented):
Just a couple of nits below

ehsannas (approved):

PR #10586, opened 2022-12-16
Introduce LocalConfigOverrides for settings.

visumickey:
Note: There are no tests here yet since everything depends on Bundle and there is no point testing them yet.

Next PR would be about integrating the local config overrides with Settings layer and at that time tests will be introduced.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/SessionLocalConfig.swift
+++ b/FirebaseSessions/Sources/SessionLocalConfig.swift
@@ -1,0 +1,53 @@
+//
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


FirebaseSessions/Sources/SessionLocalConfig.swift
```diff
@@ -0,0 +1,53 @@
...
+
+  /// Session timeout flag value.
+  var sessionSamplingRate: Double? { get }
+}
+
+/// Class that manages the local overrides configs related to the library.
+class SessionLocalConfig: SessionLocalConfigProtocol {
+  static let PlistKey_sessions_enabled = "firebase_sessions_enabled"
+  static let PlistKey_sessions_timeout = "firebase_sessions_timeout"
+  static let PlistKey_sessions_samplingRate = "firebase_sessions_samplingRate"
```
samedson, line 34: Optional: I usually see Plist Keys in Upper Camel Case (eg. FirebaseSessionsEnabled), but totally up to you if you wanna switch

samedson (approved):

FirebaseSessions/Sources/SessionLocalConfig.swift
visumickey: Sounds good. I'm fine with that approach.

PR #10590, opened 2022-12-16
[Messaging] - Decline Registration when there is no APNS Token

aashishpatil-g:
### Discussion
FCM SDK should not vend an FCM Token when there is no APNS token available or set. This means SDK will no trigger a registration till an APNS Token is provided. 

### Testing
Added unit test to check for NSError when declining FCM Token request. 

Tested following scenarios to check for regression - 
- Fresh install of app with keychain cleared.
- Reinstall of app via Xcode
- Fresh install of app with keychain not cleared. 

For each scenario, confirmed that a notification was sent as expected to the vended FCM Token.

### DIFF
--- a/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
+++ b/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
@@ -180,13 +180,15 @@
     [tokenOptions addEntriesFromDictionary:options];
   }
 
-  //ensure we have an APNS Token
+  // ensure we have an APNS Token
   if (tokenOptions[kFIRMessagingTokenOptionsAPNSKey] == nil) {
-    //we don't have an APNS token. Don't fetch or return a FCM Token
+    // we don't have an APNS token. Don't fetch or return a FCM Token
     FIRMessagingLoggerWarn(kFIRMessagingMessageCodeAPNSTokenNotAvailableDuringTokenFetch,
                            @"Declining request for FCM Token since no APNS Token specified");
     dispatch_async(dispatch_get_main_queue(), ^{
-      NSError *missingAPNSTokenError = [NSError messagingErrorWithCode:kFIRMessagingErrorCodeMissingDeviceToken failureReason:@"No APNS token specified before fetching FCM Token"];
+      NSError *missingAPNSTokenError =
+          [NSError messagingErrorWithCode:kFIRMessagingErrorCodeMissingDeviceToken
+                            failureReason:@"No APNS token specified before fetching FCM Token"];
       handler(nil, missingAPNSTokenError);
     });
     return;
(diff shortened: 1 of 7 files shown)


FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -180,6 +180,18 @@ - (void)tokenWithAuthorizedEntity:(NSString *)authorizedEntity
     [tokenOptions addEntriesFromDictionary:options];
   }
 
+  //ensure we have an APNS Token
+  if (tokenOptions[kFIRMessagingTokenOptionsAPNSKey] == nil) {
```
charlotteliang, line 184: Have you tested this with fresh install? I believe the first request is almost always not including a APNS token. 

Would this show as a critical error to developers or just warning  on the API surface level? I want to make sure they are not thinking things stop working even though the second request will go through

aashishpatil-g: I tested three different scenarios (also included in PR description) to confirm we don't regress - 

- Fresh install of app with keychain cleared.
- Reinstall of app via Xcode
- Fresh install of app with keychain not cleared.

For each scenario, confirmed that a notification was sent as expected to the vended FCM Token.

It gets 
- logged as a WARNING. I am on the fence if this should be logged as ERROR. 
- the completion block returns an NSError with a new code kFIRMessagingErrorCodeMissingDeviceToken

aashishpatil-g: Also added a unit test that checks for an NSError when device token is missing.

paulb777:
Please fix the CI failures

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
charlotteliang: Thanks for explaining! I think it should be warning because apple might return the apns token with a delay then the first token request will fail.
Unless we changed the logic that only send the first token request when apple returns apns token. I don't remember if we made the change yet.

charlotteliang (approved):

paulb777 (approved):

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
aashishpatil-g: Ok cool - I'll leave it as WARN. 

>>Unless we changed the logic that only send the first token request when apple returns apns token. I don't remember if we made the change yet.

The actual flow (I think its part of swizzling logic) hasn't been changed here - but we won't make a register call if there is no APNS token.

PR #10591, opened 2022-12-16
Add Pause/resume swift integration tests

paulb777:
The pause and resume tests were never ported from Objective C to Swift. Here they are ...

#no-changelog

### DIFF
--- a/FirebaseStorage/Tests/Integration/StorageIntegration.swift
+++ b/FirebaseStorage/Tests/Integration/StorageIntegration.swift
@@ -587,6 +587,88 @@ class StorageResultTests: StorageIntegrationCommon {
     waitForExpectations()
   }
 
+  func testResumeGetFile() {
+    let expectation = self.expectation(description: #function)
+    let expectationPause = self.expectation(description: "pause")
+    let expectationResume = self.expectation(description: "resume")
+    let ref = storage.reference().child("ios/public/1mb")
+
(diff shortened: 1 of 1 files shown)


FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -587,6 +587,88 @@ class StorageResultTests: StorageIntegrationCommon {
...
+        description.contains("State: Resume"))
+      let progress = snapshot.progress
+      if let completed = progress?.completedUnitCount {
+        XCTAssertGreaterThanOrEqual(completed, downloadedBytes)
+        downloadedBytes = completed
+        if completed > resumeAtBytes {
+          task.pause()
+          expectationPause.fulfill()
+          resumeAtBytes = Int.max
+        }
```
ncooke3, line 617: Sorry, I'm not grasping the significance of `completed > resumeAtBytes`. What does that mean?

ncooke3 (approved):
LGTM. Left one question for my own understanding...

FirebaseStorage/Tests/Integration/StorageIntegration.swift
paulb777: In practice in the debugger, about three separate progress events happen before `resumeAtBytes` is exceeded and the pause gets executed.

PR #10592, opened 2022-12-16
[Firestore] Encode Data as Blob instead of String

andrewheard:
**[WIP]** For debugging, updated FirebaseDataEncoder's handling of Data values to match the previous [FirestoreEncoder](https://github.com/firebase/firebase-ios-sdk/blob/161c046ded212fe76fe68f3e8bd4e32ec462960a/Firestore/third_party/FirestoreEncoder/FirestoreEncoder.swift#L456-L457) implementation.

This results in a Data value being encoded as a Blob, as before, in Firestore.

Note: This cannot be submitted as-is since it breaks support for `DataEncodingStrategy`.

### DIFF
--- a/FirebaseSharedSwift/Sources/third_party/FirebaseDataEncoder/FirebaseDataEncoder.swift
+++ b/FirebaseSharedSwift/Sources/third_party/FirebaseDataEncoder/FirebaseDataEncoder.swift
@@ -936,8 +936,7 @@ extension __JSONEncoder {
       // Respect Date encoding strategy
       return try self.box((value as! Date))
     } else if type == Data.self || type == NSData.self {
-      // Respect Data encoding strategy
-      return try self.box((value as! Data))
+      return (value as! NSData)
     } else if type == URL.self || type == NSURL.self {
       // Encode URLs as single strings.
       return self.box((value as! URL).absoluteString)


mortenbekditlevsen:
Could this functionality be added as a custom data encoding strategy inside of the firestore encoder? And then also make this strategy be the default when using the firestore encoder. And ditto for decoding.

I am likely responsible for introducing this behavior change as I was unaware of any special handling of NSData that firestore must have.
Many apologies for that.

mortenbekditlevsen:
Right - so another idea would be to include Data in the passthrough  types, but then you lose the user customization aspect. 

I'll look at it a bit more and get back.

mortenbekditlevsen:
Two vague ideas, one a bit hacky:

1. Create a subclass of NSData. In the strategy wrap the data in an instance of this subclass. Mark this type as a passthrough type and hope that firestore accepts the subclass as if was plain NSData. Requires Darwin, so no good for a Swift-only futures 😊

2. Create a wrapper struct in Swift and mark this as passthrough. Add special unwrapping handling in the Firebase encoder. This means that the wrapper type should likely also live with the encoder. Like above, wrap in this based on strategy.

I'll see if I can find time to try it out.

mortenbekditlevsen:
I'm not sure, but could something like #10596 work? Untested as I am on the go.

paulb777:
Closing in favor of #10598. Please continue the conversation there.

PR #10596, opened 2022-12-20
Attempt at providing a 'blob' data encoding strategy for firestore

mortenbekditlevsen:
This is just a PR to participate in a discussion on PR #10592


paulb777:
See  #10598 as an alternative. The change is more targeted and avoids the complexity of the PassThrough struct.

mortenbekditlevsen:
Excellent. The extra ceremony in my PR was just an attempt to hide the blob strategy from being used with RTDB and callable functions, but the other way does the job. 😊

PR #10598, opened 2022-12-20
Restore 9.x Firestore Data Codable behavior

paulb777:
Internal b/262520379

Version 10.x changed the default behavior of coding and decoding `Data` types in Firestore. This PR restores the old behavior by implementing the `blob` data encoding and decoding strategy.


FirebaseSharedSwift/Sources/third_party/FirebaseDataEncoder/FirebaseDataEncoder.swift
```diff
@@ -142,6 +142,9 @@ public class FirebaseDataEncoder {
     /// Encoded the `Data` as a Base64-encoded string. This is the default strategy.
     case base64
 
+    /// Encoded the `Data` as an `NSData` blob.
```
andrewheard, line 145: nit: "Encode" instead of "Encoded"

andrewheard (approved):

PR #10599, opened 2022-12-20
Reorganize Sessions settings

visumickey:
#no-changelog

Pending: Unit tests for Sessions Settings.
Changes here are based out of the design here: go/sessions-settings-module-ios-visu


FirebaseSessions/Sources/Settings/SDKDefaultSettings.swift
```diff
@@ -0,0 +1,46 @@
...
+    return 30 * 60
+  }
+
+  var samplingRate: Double? {
+    // Default is all events are dispatched
+    return 1.0
+  }
+}
+
+typealias SDKDefaultSettingsProvider = SDKDefaultSettings
```
leotianlizhan, line 36: Not action needed, just curious what's this `typealias` + `extension` pattern, is it a recommended pattern in swift?

visumickey: Typealias is a technique to make the code more modular and readable. In this example, I have split the big class into sub modules depending on the action they perform (implementing different protocols). As classes grow, this is a simple yet powerful technique to make the code maintainable.

Some references: https://www.programiz.com/swift-programming/typealias, https://www.avanderlee.com/swift/typealias-usage-swift/

Happy to provide a lot more context offline.

FirebaseSessions/Sources/Settings/SessionsSettings.swift
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  var sessionTimeout: TimeInterval {
+    // Order of precendence LocalOverrides > Remote Settings > SDK Defaults
+    if !localOverrides.isSettingsStale(), localOverrides.sessionTimeout != nil {
+      return localOverrides.sessionTimeout!
+    } else if !remoteSettings.isSettingsStale(), remoteSettings.sessionTimeout != nil {
+      return remoteSettings.sessionTimeout!
+    }
+    return sdkDefaults.sessionTimeout!
```
ncooke3, line 65: nit: Favor unwrapping optional to minimize bang operator usage. 

Something like:
```suggestion
    if !localOverrides.isSettingsStale(), let sessionTimeout = localOverrides.sessionTimeout {
      return sessionTimeout
    } else if !remoteSettings.isSettingsStale(), let sessionTimeout = remoteSettings.sessionTimeout {
      return remoteSettings.sessionTimeout!
    }
    return sdkDefaults.sessionTimeout! // <– Is this guaranteed to never be nil?
```

FirebaseSessions/Sources/Settings/SessionsSettings.swift
```diff
@@ -0,0 +1,84 @@
...
+    // Order of precendence LocalOverrides > Remote Settings > SDK Defaults
+    if !localOverrides.isSettingsStale(), localOverrides.samplingRate != nil {
+      return localOverrides.samplingRate!
+    } else if !remoteSettings.isSettingsStale(), remoteSettings.samplingRate != nil {
+      return remoteSettings.samplingRate!
+    }
+    return sdkDefaults.samplingRate!
+  }
+
+  func checkAndUpdateSettings() {
```
ncooke3, line 78: optional: I don't think the `checkAnd` part adds much important info. I think `updateSettings` sufficiently described this API's function and implementation. Feel free to keep it if you prefer it though.
```suggestion
  func updateSettings() {
```

FirebaseSessions/Sources/Settings/SessionsSettings.swift
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  var samplingRate: Double {
+    // Order of precendence LocalOverrides > Remote Settings > SDK Defaults
+    if !localOverrides.isSettingsStale(), localOverrides.samplingRate != nil {
+      return localOverrides.samplingRate!
+    } else if !remoteSettings.isSettingsStale(), remoteSettings.samplingRate != nil {
+      return remoteSettings.samplingRate!
+    }
+    return sdkDefaults.samplingRate!
```
ncooke3, line 75: This can be refactored similarly to the block above!

FirebaseSessions/Tests/Unit/SessionsSettingsTests.swift
```diff
@@ -13,13 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Foundation
-@testable import FirebaseSessions
-
-class MockSessionLocalConfig: SessionLocalConfigProtocol {
-  var sessionEnabled: Bool?
+import XCTest
```
ncooke3, line 16: Are there future plans for this file?

FirebaseSessions/Sources/Settings/SessionsSettings.swift
visumickey: I like that too. Modified.

FirebaseSessions/Sources/Settings/SessionsSettings.swift
visumickey: Makes sense. Done.

FirebaseSessions/Sources/Settings/SessionsSettings.swift
visumickey: Makes sense. Done.

visumickey: Note: SDK Defaults is guaranteed to never be NIL as they are hardcoded within the SDK.

FirebaseSessions/Tests/Unit/SessionsSettingsTests.swift
visumickey: Yes, I'm adding tests to this.

visumickey: The tests are now added.

FirebaseSessions/Sources/Settings/LocalOverrideSettings.swift
```diff
@@ -38,7 +38,10 @@ class LocalOverrideSettings: SettingsProvider, SettingsProtocol {
   }
 
   private func plistValueForConfig(configName: String) -> Any? {
-    return Bundle.main.value(forKey: configName)
+    guard let value = Bundle.main.object(forInfoDictionaryKey: configName) else {
+      return nil
+    }
+    return value
```
ncooke3, line 44: `object(forInfoDictionaryKey:)` returns `Any?` so this could be simplified to just:
```suggestion
    return Bundle.main.object(forInfoDictionaryKey: configName)
```

FirebaseSessions/Sources/Settings/SettingsProtocol.swift
```diff
@@ -0,0 +1,29 @@
...
+protocol SettingsProtocol {
+  // Config to show if sessions is enabled
+  var sessionsEnabled: Bool? { get }
+
+  // Config showing the sampling rate for sessions
+
+  var samplingRate: Double? { get }
+
+  // Background timeout config value before which a new session is generated
+  var sessionTimeout: TimeInterval? { get }
```
ncooke3, line 28: Now, there may be a reason for this that I'm not aware of, so apologies and disregard if so. But all of this protocol's fields seem like they could be non-optional. For example, if I'm given a type that conforms to this protocol, I would expect that it can be in two states, either enabled or disabled. The third `nil` state is a bit harder to reason (to me, at least) and also adds an extra state that complicates client code. 

I know this protocol was defined in a previous PR but I did want to mention it in case you agree and think it is helpful. I do think reducing the use of optionals to only cases where it is absolutely necessary is a good practice and can eliminate away a lot of unnecessary code paths.

ncooke3 (approved):
LGTM

FirebaseSessions/Sources/Settings/SessionsSettings.swift
```diff
@@ -0,0 +1,86 @@
...
+      return sessionTimeout
+    }
+    return sdkDefaults.sessionTimeout!
+  }
+
+  var samplingRate: Double {
+    // Order of precendence LocalOverrides > Remote Settings > SDK Defaults
+    if !localOverrides.isSettingsStale(), let samplingRate = localOverrides.samplingRate {
+      return samplingRate
+    } else if !remoteSettings.isSettingsStale(), let samplingRate = remoteSettings.samplingRate {
```
samedson, line 74: I wonder if we can remove the `isSettingsStale` call altogether and instead just rely on the `if let ... = ...` being nil for each value.

FirebaseSessions/Sources/Settings/SettingsProtocol.swift
samedson: This is a good callout. In this case what we're doing is relying on the nil values for when the specified "layer" doesn't have a value to provide. So we're saying we have 3 Settings layers, one fetched, one configured via plist, and the 3rd is the defaults. The defaults always return a value, but the other 2 layers may not have values to provide.

Luckily, this API will be easy to use because the getters on `SessionsSettings` are not optional. The SDK will read values from `SessionsSettings` and not the other layers

FirebaseSessions/Sources/Settings/SessionsSettings.swift
visumickey: I like this. As a config settings fetcher, we don't have to worry about the staleness of configs.

FirebaseSessions/Sources/Settings/LocalOverrideSettings.swift
visumickey: I like this idea. Done!

FirebaseSessions/Sources/Settings/SettingsProtocol.swift
visumickey: Very good question and a valid one. I would typically try to be in the same state. The reason for the third state is that there is a prioritization of a config value if the value is available or not. In other words, there is a state which qualifies if a config value is available or not, in addition to if it the config value says enabled/disabled. So, had to go with this approach (a common paradigm for such situations). TBH: I first started my approach without the third state, but that would quite honestly add more complications/less-understandable code.

samedson (approved):

PR #10604, opened 2022-12-21
Handle decoding base64-encoded strings as blob data in Codables

andrewheard:
Version 10.x changed the default behaviour of encoding and decoding `Data` types in Firestore, resulting in them being stored as base64-encoded `String` fields instead of `Blob` (bytes) fields. PR #10598 restores the old coding behaviour from 9.x (defaulting to `blob`).

To prevent potential type mismatch errors when upgrading to the next version, this PR allows base64-encoded string values to be decoded as blob data when using the [`blob`](https://github.com/firebase/firebase-ios-sdk/blob/a2ea36c31796f9d71becf86c4c8b3bac49241a4b/FirebaseSharedSwift/Sources/third_party/FirebaseDataEncoder/FirebaseDataEncoder.swift#L145-L146) `DataDecodingStrategy`.

### DIFF
--- a/FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
+++ b/FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
@@ -481,6 +481,35 @@ class TestFirebaseDataEncoder: XCTestCase {
                    dataEncodingStrategy: .custom(encode),
                    dataDecodingStrategy: .custom(decode))
   }
+  
+  func testDecodingBase64StringAsBlobData() {
+    let data = "abcdef".data(using: .utf8)!
+    let base64String = "YWJjZGVm"
+    
+    let encoder = FirebaseDataEncoder()
+    encoder.dataEncodingStrategy = .base64
+    var payload: Any! = nil
+    do {
+      payload = try encoder.encode(data)
+    } catch {
+      XCTFail("Failed to encode \(Data.self): \(error)")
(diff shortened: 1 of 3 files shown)


paulb777 (approved):
LGTM on green. Thanks!

FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
```diff
@@ -482,6 +482,35 @@ class TestFirebaseDataEncoder: XCTestCase {
...
+    let base64String = "YWJjZGVm"
+
+    let encoder = FirebaseDataEncoder()
+    encoder.dataEncodingStrategy = .base64
+    var payload: Any! = nil
+    do {
+      payload = try encoder.encode(data)
+    } catch {
+      XCTFail("Failed to encode \(Data.self): \(error)")
+    }
```
ncooke3, line 496: Just throwing this out for there for future work... in case it was considered, another approach that removes the need for the `do {} catch { XCTFail(...) }` pattern is to mark the test method as `throws`. You wouldn't have the formatted error message in the failure but the throwing line should still appear in the logs– which should provide enough context as to what went wrong.

andrewheard: Good idea, I switched to that approach in PR #10638.

PR #10609, opened 2022-12-22
Add Storage API to limit upload chunk size

### LINKED ISSUE #10137: [FR]: Set chunk size on upload task
### Description

I may be missing something very obvious, but I see no way to set the upload chunk size from swift code for a 'FIRStorageUploadTask' created via FIRStorageReference putFile.
e.g. 
```swift
let uploadTask = remoteRef.putFile(from: fileURL, metadata: metadata)
```

1. Use case: allow custom chunk sizes so resumable uploads work by default
2. Currently the chunk size seems hard coded to a large value 
3. I propose allowing custom chunk sizes

In  [FIRStorageUploadTask.m](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorageInternal/Sources/FIRStorageUploadTask.m#L122) the chunk size seems to be set from the constant `kGTMSessionUploadFetcherStandardChunkSize` which is set to LLONG_MAX; 

```swift
    GTMSessionUploadFetcher *uploadFetcher =
        [GTMSessionUploadFetcher uploadFetcherWithRequest:request
                                           uploadMIMEType:strongSelf->_uploadMetadata.contentType
                                                chunkSize:kGTMSessionUploadFetcherStandardChunkSize
                                           fetcherService:self.fetcherService];
```

Perhaps there is a way to modify this somewhere, but I can't find any documentation on it.

Thanks for any support.

### API Proposal

```swift
let uploadTask = remoteRef.putFile(from: fileURL, metadata: metadata, chunkSize: myChunkSize )
```

### Firebase Product(s)

Storage

paulb777:
Fix #10137

Googlers: API proposal at [go/firebase-storage-chunk-size](http://goto.google.com/firebase-storage-chunk-size)

### DIFF
--- a/FirebaseStorage/Sources/Storage.swift
+++ b/FirebaseStorage/Sources/Storage.swift
@@ -124,6 +124,12 @@ import FirebaseAuthInterop
   }
 
   /**
+   * Specify the maximum upload chunk size. Values less than 256K (262144) will be rounded up to 256K. Values
+   * above 256K will be rounded down to the nearest 256K multiple. The default is no maximum.
+   */
+  @objc public var uploadChunkSize: Int64 = .max
+
+  /**
    * A `DispatchQueue` that all developer callbacks are fired on. Defaults to the main queue.
    */
   @objc public var callbackQueue: DispatchQueue {
--- a/FirebaseStorage/Sources/Storage.swift
+++ b/FirebaseStorage/Sources/Storage.swift
@@ -127,7 +127,7 @@ import FirebaseAuthInterop
    * Specify the maximum upload chunk size. Values less than 256K (262144) will be rounded up to 256K. Values
    * above 256K will be rounded down to the nearest 256K multiple. The default is no maximum.
    */
-  @objc public var uploadChunkSize: Int64 = .max
+  @objc public var uploadChunkSizeBytes: Int64 = .max
 
   /**
    * A `DispatchQueue` that all developer callbacks are fired on. Defaults to the main queue.
--- a/FirebaseStorage/Tests/Integration/StorageIntegration.swift
+++ b/FirebaseStorage/Tests/Integration/StorageIntegration.swift
@@ -265,6 +265,52 @@ class StorageResultTests: StorageIntegrationCommon {
     waitForExpectations()
   }
 
+  func testPutFileLimitedChunk() throws {
+    let expectation = self.expectation(description: #function)
+    let putFileExpectation = self.expectation(description: "putFile")
+    let ref = storage.reference(withPath: "ios/public/testPutFilePauseResume")
+    let bundle = Bundle(for: StorageIntegrationCommon.self)
+    let filePath = try XCTUnwrap(bundle.path(forResource: "1mb", ofType: "dat"),
+                                 "Failed to get filePath")
+    let data = try XCTUnwrap(try Data(contentsOf: URL(fileURLWithPath: filePath)),
+                             "Failed to load file")
+    let tmpDirURL = URL(fileURLWithPath: NSTemporaryDirectory())
+    let fileURL = tmpDirURL.appendingPathComponent("LargePutFile.txt")
+    var progressCount = 0
+
+    try data.write(to: fileURL, options: .atomicWrite)
+
+    // Limit the upload chunk size
+    storage.uploadChunkSize = 256_000
+
+    let task = ref.putFile(from: fileURL) { result in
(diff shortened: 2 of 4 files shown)


paulb777:
@maneesht Let me know what you think and I can write up an API proposal.

maneesht:
> @maneesht Let me know what you think and I can write up an API proposal.

LGTM

FirebaseStorage/Sources/Storage.swift
```diff
@@ -123,6 +123,12 @@ import FirebaseAuthInterop
     }
   }
 
+  /**
+   * Specify the maximum upload chunk size. Values less than 256K (262144) will be rounded up to 256K. Values
+   * above 256K will be rounded down to the nearest 256K multiple. The default is no maximum.
```
charlotteliang, line 128: the comment says default is no maximum but it seems like it's assigned a default value maximum.

paulb777: The default value is the maximum Int64 value which is billions of gigabytes, so effectively unlimited.

FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -265,6 +265,52 @@ class StorageResultTests: StorageIntegrationCommon {
...
+    let data = try XCTUnwrap(try Data(contentsOf: URL(fileURLWithPath: filePath)),
+                             "Failed to load file")
+    let tmpDirURL = URL(fileURLWithPath: NSTemporaryDirectory())
+    let fileURL = tmpDirURL.appendingPathComponent("LargePutFile.txt")
+    var progressCount = 0
+
+    try data.write(to: fileURL, options: .atomicWrite)
+
+    // Limit the upload chunk size
+    storage.uploadChunkSizeBytes = 256_000
```
charlotteliang, line 284: Will this get overwrite by Int64.max later? trying to understand the order here.

charlotteliang: should we also add a test when user sets an number smaller than 256 K?

charlotteliang (approved):

FirebaseStorage/Tests/Integration/StorageIntegration.swift
paulb777: Made the code clearer with `defer` and a comment.

Added the test for smaller.

FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -265,6 +265,106 @@ class StorageResultTests: StorageIntegrationCommon {
...
+
+    try data.write(to: fileURL, options: .atomicWrite)
+
+    // Limit the upload chunk size. This should behave exactly like the previous
+    // test since small chunk sizes are rounded up to 256K.
+    storage.uploadChunkSizeBytes = 1
+
+    let task = ref.putFile(from: fileURL) { result in
+      XCTAssertGreaterThanOrEqual(progressCount, 4)
+      XCTAssertLessThanOrEqual(progressCount, 6)
```
charlotteliang, line 342: Should we also have another test that's like a multiple of 256 then i'm guessing the progress count will be smaller than 4?

charlotteliang (approved):
LGTM!

FirebaseStorage/Tests/Integration/StorageIntegration.swift
```diff
@@ -265,6 +265,106 @@ class StorageResultTests: StorageIntegrationCommon {
...
+    let fileURL = tmpDirURL.appendingPathComponent("LargePutFile.txt")
+    var progressCount = 0
+
+    try data.write(to: fileURL, options: .atomicWrite)
+
+    // Limit the upload chunk size
+    storage.uploadChunkSizeBytes = 256_000
+
+    let task = ref.putFile(from: fileURL) { result in
+      XCTAssertGreaterThanOrEqual(progressCount, 4)
```
maneesht, line 291: Why can't we do `XCTAssertEqual` here and be more precise?

paulb777: Since 256,000 is a limit and the chunks could be smaller but not bigger.

FirebaseStorage/Tests/Integration/StorageIntegration.swift
paulb777: There's the original test above that does this.

PR #10618, opened 2022-12-28
Add TestApp targets with Crash / Perf installed

samedson:
Added these to manually interact with interop code.

Screen Shot 2022-12-28 at 2 38 07 PM


visumickey (commented):
Can you create another target where both Crashlytics and Fireperf are installed? That would cover all the usecases we wanted to test.

visumickey (approved):

PR #10619, opened 2022-12-28
[Sessions] Add optional interop between Crashlytics and Sessions

samedson:
# Description
 - I originally defined `FIRSessionsProvider` in FirebaseSessions.swift. Then, in Crashlytics I was importing `<FirebaseSessions/FirebaseSessions-Swift.h>`. This was going to be fairly messy to code because it needs to be an optional import until Firebase Sessions releases, and all uses of the `FIRSessionsProvider` would need to be protected by `#if __has_include...` or some other IfDef. 

 - Instead, I tried the AnalyticsInterop pattern, and it was fairly simple to implement, and makes this interum state easy to code by just giving us a `nil` sessions SDK when it's not installed in the customer's app.

Once Sessions SDK releases, we can make it a hard dependency by just moving `FIRSessionsProvider` to FirebaseSessions.swift, and importing `<FirebaseSessions/FirebaseSessions-Swift.h>`. I've added TODOs for this.

Example reading the session ID:

Screen Shot 2022-12-28 at 3 39 14 PM

# TODO

- [ ] Figure out how to wait for Subscribers in the Sessions SDK for the purposes of determining data collection state and filling in fields.

#no-changelog

### DIFF
--- a/FirebaseSessions/Interop/FIRSessionsProvider.h
+++ b/FirebaseSessions/Interop/FIRSessionsProvider.h
@@ -38,8 +38,9 @@ NS_SWIFT_NAME(SessionsProvider)
 
 /// Subscribes the given `subscriber` to the Notification for receiving SessionID changes.
 /// The `onSessionIDChanged` method will be called immediately with the existing Session ID
-/// to handle cases where the Sessions SDK has started and rotated before this subscription was made.
-- (void)subscribeForSessionIDchanged:(id <FIRSessionsSubscriber>)subscriber;
+/// to handle cases where the Sessions SDK has started and rotated before this subscription was
+/// made.
+- (void)subscribeForSessionIDchanged:(id<FIRSessionsSubscriber>)subscriber;
 
 @end
 
--- a/FirebaseSessions/Interop/FIRSessionsProvider.h
+++ b/FirebaseSessions/Interop/FIRSessionsProvider.h
@@ -30,7 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// FirebaseSessions.swift
 @protocol FIRSessionsProvider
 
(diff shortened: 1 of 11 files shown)


FirebaseSessions/Interop/FIRSessionsProvider.h
```diff
@@ -0,0 +1,35 @@
...
+/// FirebaseSessions APIs.
+///
+/// Normally this protocol would be defined in FirebaseSessions.swift, but
+/// we haven't yet released FirebaseSessions SDK, so it is an optional
+/// dependency. To make SDKs that depend on FirebaseSessions build,
+/// we're defining this header in Objective-C and including it in all places.
+///
+/// TODO(samedson) Remove Interop when FirebaseSessions releases and move to
+/// FirebaseSessions.swift
+@protocol FIRSessionsProvider
```
paulb777, line 29: Perhaps `NS_SWIFT_NAME(SessionsProvider)`

paulb777 (commented):
Package.swift needs updates

samedson:
> Package.swift needs updates

@paulb777  so I thought that too, but for some reason we don't specify the Analytics Interop file here:
https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#L492-L533

Even though we do in the podspec:
https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCrashlytics.podspec#L36

And weirdly, Crashlytics builds when I open the Package.swift and compile the SDK target

paulb777:
Oh right, SPM doesn't need referenced headers to be specified. There is an SPM CI problem though

samedson:
Closing in favor of https://github.com/firebase/firebase-ios-sdk/pull/10664

Issue #10623, opened 2022-12-29
Crashlytics Unit Test Directory Structure incompatible with Swift Tools version 5.6


paulb777:
### Description

The Crashlytics unit tests need to be restructured to unblock Swift Tools version migration

### Reproducing the issue

Update Package.swift line 1 to 5.6 and try to load in Xcode 

![Screen Shot 2022-12-29 at 11 30 16 AM](https://user-images.githubusercontent.com/73870/210003643-42a334c6-30df-44fd-934c-2092c1a47b9e.png)

### Firebase SDK Version

10.3.0

### Xcode Version

14.2

### Installation Method

Swift Package Manager

### Firebase Product(s)

Crashlytics

### Targeted Platforms

iOS

### If using Swift Package Manager, the project's Package.resolved

<details>
<summary>Expand <code>Package.resolved</code> snippet</summary>
<br>

```json

Replace this line with the contents of your Package.resolved.

```

</details>

### If using CocoaPods, the project's Podfile.lock

<details>
<summary>Expand <code>Podfile.lock</code> snippet</summary>
<br>

```yml

Replace this line with the contents of your Podfile.lock!

```

</details>

PR #10634, opened 2023-01-03
[Crashlytics]Fix crashlytics unit test directory

themiswang:
Change to copy for all resources to maintain the directory structure. Related to issue #10623 
- Retain the resources directory structure by using copy instead of process to avoid dup naming within same the directory.
- Renaming the group name from `data` to `machO_data` to avoid dup name for another resources group.
- Changed podspec for cocoapod as well

### Testing
Testing on both 5.3 and 5.6, passed unit tests, confirm no package resolve error for version 5.6.

 #no-changelog


paulb777 (commented):
Thanks for investigating. It looks like the structure doesn't work for CocoaPods. Is there something that will work for the three options of 5.3, 5.6 and CocoaPods?

themiswang:
Hey @paulb777, sorry tagged you before before the pr is ready. I might know the issue why pod does not work because I forget change renamed directory for pod spec. Let me play around with it more and I'll make it ready for review once I passed all unit tests

paulb777 (approved):
👍

PR #10637, opened 2023-01-04
Apply the Configs Settings to the SDK

visumickey:
Pending: Applying sampling rate. Reason: Requires a bit of refactor before applying, so will do in a separate PR.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -61,12 +61,13 @@ protocol SessionsProvider {
       fireLogger: fireLogger,
       sampler: SessionSampler()
     )
-    let initiator = SessionInitiator()
+
     let appInfo = ApplicationInfo(appID: appID)
     let settings = SessionsSettings(
       appInfo: appInfo,
       installations: installations
     )
+    let initiator = SessionInitiator(settings: settings)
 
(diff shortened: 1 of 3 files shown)


FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -93,10 +94,15 @@ protocol SessionsProvider {
       // On each session start, first update Settings if expired
       self.settings.updateSettings()
       self.identifiers.generateNewSessionID()
-      let event = SessionStartEvent(identifiers: self.identifiers, appInfo: self.appInfo)
-      DispatchQueue.global().async {
-        self.coordinator.attemptLoggingSessionStart(event: event) { result in
+      // Generate a session start event only when session data collection is enabled.
+      if self.settings.sessionsEnabled {
```
samedson, line 98: At some point I want to add tests for this code, maybe in a follow up.

samedson (approved):
LGTM!

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: Acknowledged. The current test setup does not allow for testing (modular or abstract testing) this easily. So, will differ this on a future change to make such refactors.

PR #10638, opened 2023-01-04
[FirestoreSharedSwift] Write base64 decode test as throwing XCTest

andrewheard:
Small refactor to simplify the error handling in `testDecodingBase64StringAsBlobData` by making it a throwing test. As a side-effect, this will prevent the test from continuing if an earlier step throws (similar to `continueAfterFailure` set to `false`) which may make failures easier to debug.

Updated to use `XCTUnwrap` to explicitly check for optionals when performing type coercion.

Note: Did not refactor other tests in the file to reduce diversion from the original `swift/test/stdlib/TestJSONEncoder.swift` source.

#no-changelog

### DIFF
--- a/FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
+++ b/FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
@@ -482,33 +482,33 @@ class TestFirebaseDataEncoder: XCTestCase {
                    dataDecodingStrategy: .custom(decode))
   }
 
-  func testDecodingBase64StringAsBlobData() {
-    let data = "abcdef".data(using: .utf8)!
+  func testDecodingBase64StringAsBlobData() throws {
+    // Tests implicit migration of data that was written with .base64 (String type) using
+    // Firestore 10.0 through 10.3 (see PR #10604).
+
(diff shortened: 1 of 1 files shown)


FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
```diff
@@ -482,33 +482,33 @@ class TestFirebaseDataEncoder: XCTestCase {
                    dataDecodingStrategy: .custom(decode))
   }
 
-  func testDecodingBase64StringAsBlobData() {
-    let data = "abcdef".data(using: .utf8)!
+  func testDecodingBase64StringAsBlobData() throws {
+    // Tests implicit migration of data that was written with .base64 (String type) using
```
paulb777, line 486: Nit: function comment should be above the function.

paulb777 (approved):
LGTM on nit fix and green

ncooke3 (approved):

FirebaseSharedSwift/Tests/third_party/DataEncoderTests.swift
andrewheard: Done. Thanks!

PR #10645, opened 2023-01-05
Add FIID and new proto fields to Crashlytics

samedson:


Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
```diff
@@ -94,7 +94,9 @@ - (void)prepareAndSubmitReport:(FIRCLSInternalReport *)report
         dispatch_once(&regenerateOnceToken, ^{
           // Check to see if the FID has rotated before we construct the payload
           // so that the payload has an updated value.
-          [self.installIDModel regenerateInstallIDIfNeeded];
+          [self.installIDModel regenerateInstallIDIfNeededWithBlock:^(NSString *_Nonnull newFIID) {
+            self.fiid = [newFIID copy];
```
visumickey, line 98: Why explicitly copy when the property is already qualified as "Copy"?

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
```diff
@@ -29,19 +29,22 @@
...
                  googleAppId:(NSString *)googleAppID
-              installIDModel:(FIRCLSInstallIdentifierModel *)installIDModel {
+              installIDModel:(FIRCLSInstallIdentifierModel *)installIDModel
+                        fiid:(NSString *)fiid {
   self = [super init];
   if (self) {
     _folderPath = folderPath;
     _googleAppID = googleAppID;
     _installIDModel = installIDModel;
+    _fiid = [fiid copy];
```
visumickey, line 47: Same as above

Crashlytics/ProtoSupport/Protos/crashlytics.proto
```diff
@@ -33,9 +34,16 @@ message Report {
...
-  // Unique device generated guid for application install.
+  // Unique Crashlytics-specific device generated guid for application install.
+  // Equivalent to Session.app.installation_uuid
   string installation_uuid = 5;
 
+  // Unique device generated ID from the FirebaseInstallations SDK
+  string firebase_installation_id = 16;
+
+  // The last Session ID associated with the crash report.
+  string app_quality_session_id = 17;
```
visumickey, line 45: Would the backend change be done in a separate CL?

visumickey (commented):
The changes look good to me!

Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
samedson: It's kindof a funny thing, but the style guide suggests you copy strings when you don't know where they came from: https://google.github.io/styleguide/objcguide.html#setters-copy-nsstrings. The main thing is to avoid cases of it being mutable under the hood or having different retain semantics than what you expect.

Crashlytics/ProtoSupport/Protos/crashlytics.proto
samedson: Yes that is done in cl/499957188

Crashlytics/Crashlytics/Models/Record/FIRCLSReportAdapter.m
samedson: Answered above

Crashlytics/Crashlytics/Controllers/FIRCLSReportUploader.m
visumickey: Interesting. Surprisingly their example is a on an override setter method which makes it even more weird. Since there is no harm in re-copying, I'm fine with this change.

Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
```diff
@@ -98,14 +98,17 @@ - (NSString *)generateInstallationUUID {
...
+- (BOOL)regenerateInstallIDIfNeededWithBlock:(void (^)(NSString *fiid))block {
   BOOL __block didRotate = false;
 
   dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
 
   // This runs Completion async, so wait a reasonable amount of time for it to finish.
   [self.installations
       installationIDWithCompletion:^(NSString *_Nullable currentIID, NSError *_Nullable error) {
+        // Provide the IID to the callback
+        block(currentIID);
```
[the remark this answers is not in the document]
samedson: So in this case I don't actually have any way of handling the error besides passing nil to the block and logging a message. Since logging a message is handled in `rotateCrashlyticsInstallUUIDWithIID`, I went with just simply logging what we're dealt. Does that make sense to you?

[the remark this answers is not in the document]
themiswang: In `rotateCrashlyticsInstallUUIDWithIID`, we stored a fiid hash to UserDefaults `[defaults setObject:currentIIDHash forKey:FIRCLSInstallationIIDHashKey];` and if there is error we return early. Want to make sure we are not changing the fiid if error happen to align with the way we store this fiidHash.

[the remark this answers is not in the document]
themiswang: Discussed offline, added comments for explanation.

PR #10650, opened 2023-01-06
Upgrade grpc to 150 and C++ to 14 For Cocoapods and SPM

wu-hui:


Package.swift
```diff
@@ -175,12 +175,12 @@ let package = Package(
...
       name: "abseil",
       url: "https://github.com/firebase/abseil-cpp-SwiftPM.git",
-      "0.20220203.1" ..< "0.20220204.0"
+      "0.20220623.0" ..< "0.20220624.0"
     ),
     .package(
       name: "gRPC",
       url: "https://github.com/grpc/grpc-ios.git",
-      "1.44.0-grpc" ..< "1.45.0-grpc"
+      "1.50.0-grpc" ..< "1.50.1-grpc"
```
paulb777, line 183: To allow patches like the CocoaPods spec, this should be 1.51.0-grpc

HannahShiSFB:
We have released 1.50.1, can you try if it passes

wu-hui:
> We have released 1.50.1, can you try if it passes

Thanks, I am seeing build error however: https://github.com/firebase/firebase-ios-sdk/actions/runs/4599998333/jobs/8126084582?pr=10650

@sampajano FYI

sampajano:
@HannahShiSFB Hi Hannah could you help take a look at the above error and if they ring a bell? Thanks!

cherylEnkidu (approved):

wu-hui:
> @wu-hui, FYI ICYMI, #11208 updated this for SPM

Thank you!

Issue #10654, opened 2023-01-07
[FR]: DynamicLinks should support an initializer that accepts a FirebaseApp


sergiocampama:
### Description

Use case: Running an app with integrations to multiple Firebase environments (projects)

Currently: `DynamicLinks.dynamicLinks()` is the only way to get an instance of DynamicLinks.

Proposal: Same as multiple other components, there should be a `DynamicLinks.dynamicLinks(app: firebaseApp)` API to get a DynamicLinks object not tied to the default app.

### API Proposal

```
/**
 * @method dynamicLinks
 * @abstract Shared instance of FIRDynamicLinks.
 * @param app The app for which to retrieve the associated `FIRDynamicLinks` instance.
 * @return Shared instance of FIRDynamicLinks.
 */
+ (instancetype)dynamicLinksWithApp:(FIRApp *)app NS_SWIFT_NAME(dynamicLinks(app:));
```

Implementation:

```
+ (instancetype)dynamicLinksWithApp:(FIRApp *)app {
  id<FIRDynamicLinksInstanceProvider> instance =
      FIR_COMPONENT(FIRDynamicLinksInstanceProvider, app.container);
  return (FIRDynamicLinks *)instance;
}
```

### Firebase Product(s)

DynamicLinks

paulb777:
Now that [Firebase DynamicLinks is planned to shut down](https://firebase.google.com/support/dynamic-links-faq), closing its feature requests.

PR #10655, opened 2023-01-07
DynamicLinks: Add app based initializer

sergiocampama:
Solves #10654 10654


sergiocampama:
Looks like it has a dependency on analytics, and that's why it has to use the default app, which is not ideal

paulb777:
thanks for investigating. There shouldn't be a dependency on analytics, but the FDL back end may not support multiple FirebaseApps for a single iOS app.

PR #10660, opened 2023-01-10
[Sessions] Audit logging and update README with strategy

samedson:
Updated README: https://github.com/firebase/firebase-ios-sdk/blob/bc76232a4e0bf58daf00b7c3c4640805722d1477/FirebaseSessions/README.md

Example with `-FIRDebugEnabled`

```
[FirebaseSessions][I-SES000000] Version 10.4.0
[FirebaseSessions][I-SES000000] [Settings] Cache is not expired, no fetch will be made.
[FirebaseSessions][I-SES000000] Initiated new Session with ID: 3152c81f3bad4b518c2a1e0b4459509f
[FirebaseSessions][I-SES000000] Logging events using fast QOS due to running on a simulator
[FirebaseSessions][I-SES000000] Printing Session Event due to "-FIRSessionsDebugEvents" command line argument
[FirebaseSessions][I-SES000000] Successfully logged Session Start event to GoogleDataTransport

```

#no-changelog

### DIFF
--- a/FirebaseSessions/README.md
+++ b/FirebaseSessions/README.md
@@ -6,11 +6,14 @@ Follow the [Main Firebase Readme](https://github.com/firebase/firebase-ios-sdk#d
 ## Development
 ### Generating the Project and Test Project
 
- - `generate_project.sh` uses [cocoapods-generate](https://github.com/square/cocoapods-generate) to create an Xcode Workspace that has the SDK installed for all the SDK's supported platforms. This is useful for test-based development.
+ - Test-based Development:
+    - **Option 1:** `generate_project.sh` uses [cocoapods-generate](https://github.com/square/cocoapods-generate) to create an Xcode Workspace that has the SDK installed for all the SDK's supported platforms. This is useful for test-based development.
+    - **Option 2:** `open Package.swift` in the root of the firebase-ios-sdk repo. You can run tests using the `FirebaseSessionsUnit` Scheme
  - `generate_testapp.sh` generates and opens a test app with the Sessions SDK included. This is useful for developing the Sessions SDK against a real app.
 
-### Switching dev environments - Autopush/Staging/Prod
+### Debugging Options
 
+#### Switching Dev Environments - Autopush/Staging/Prod
 SDK is configured to send events to different environments. To enforce different environments for sending events, we use an environment variable to configure the specific environment. Since environment variables are enforced in the context of the App, use the TestApp to send events to different environments after using the following configuration steps.
 
 - Enter "Edit scheme" - On the title bar menu "Product" > "Scheme" > "Edit Scheme"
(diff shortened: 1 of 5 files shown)


FirebaseSessions/README.md
```diff
@@ -42,3 +50,11 @@ brew install protobuf
...
  1. Update the SDK to use the new proto fields
+
+
+### Logging
+The Sessions SDK uses the following strategy when determining log level:
+ - **Info** should be used rarely. Because the Sessions SDK is a dependency of other products, customers will not expect regular logs from the SDK. Therefore, info events are not recommended except under circumstances where the code path is blocked by another debug parameter (eg. `-FIRSessionsDebugEvents` will log under info because we don't want to require it be paired with `-FIRDebugEnabled`)
+ - **Debug** Is recommended to be used generously in the Sessions SDK for the purposes of debugging customer issues.
+ - **Warning** Is used when the Sessions SDK runs into a recoverable issue that still results in events being sent. For example, a problem converting between values that results in an incorrect value being reported.
+ - **Error** Is used when the Sessions SDK runs into an unrecoverable issue that prevents functionality from working. If we would want customers to reach out to us when a issue happens, then error logs should be used to convey the issue.
```
visumickey, line 60: How do these levels maps to issues that are "Fixable" by the developer themselves versus the ones that the developers would be aware, but they cannot do anything about it?

For eg: Error state IMO is something that the developer cannot fix by themselves and unaware of the impact of the message. Warning is something that the developer cannot do anything about, but has a sense of the impact. Debug is something to give more details for the developers to share with us.

It its current definition, the term "unrecoverable" is something from the SDK perspective, but does not provide any signal/actionable thing from the developers perspective. Is that how you imagine this to be?

FirebaseSessions/Sources/Settings/LocalOverrideSettings.swift
```diff
@@ -32,7 +32,7 @@ class LocalOverrideSettings: SettingsProvider, SettingsProtocol {
 
   var sessionTimeout: TimeInterval? {
     let timeout = plistValueForConfig(configName: LocalOverrideSettings
-      .PlistKey_sessions_timeout) as? String
+      .PlistKey_sessions_timeout) as? Double
     if timeout != nil {
       return Double(timeout!)
```
visumickey, line 37: If we are already casting this as a Double?, we do not have to do typecasting later. Here and below.

visumickey (commented):
I'm generally fine with the changes. All the log levels definitions look good to me except for "Error". Once we have nailed in, we should be all set.

FirebaseSessions/Sources/Settings/LocalOverrideSettings.swift
samedson: Moved this to a new PR to show you: https://github.com/firebase/firebase-ios-sdk/pull/10716

FirebaseSessions/README.md
samedson: So I've been thinking about this more, and I think we need to use language to convey fixability instead of log level.

Eg. Say we decided `warning` was for fixable problems, but there was an issue that would cause the SDK to not upload events. In that situation I think I would want the log to be an error so that it's more visible to the customer.

So maybe our recommendation should be: if it's fixable, put the resolution steps in the log itself.

visumickey: That sounds good. I like the verbiage nature of these errors/concerns that can enable the developers to fix it. In that case, does an `Error` level mean that it is something the developer can fix it? Or should such issues be tagged at `Warning` level.

visumickey: As long as that is aligned - the changes look good to me.

samedson: Yeah what I meant above is to make it based on the message instead of log level.

visumickey (approved):

PR #10664, opened 2023-01-10
Add required dependency between Crashlytics and Session

samedson:
Alternative to https://github.com/firebase/firebase-ios-sdk/pull/10619 where the dependency is required.

Replaces https://github.com/firebase/firebase-ios-sdk/pull/10636

# TODO
- [x] Add tests done in https://github.com/firebase/firebase-ios-sdk/pull/10669
- [x] Ensure future backgrounds still resolve the promises done in https://github.com/firebase/firebase-ios-sdk/pull/10669

PR #10636, opened 2023-01-03
Add required dependency between Crashlytics and Sessions

samedson:
Alternative to https://github.com/firebase/firebase-ios-sdk/pull/10619 where the dependency is hard.

# TODO

- [x] Rebase on https://github.com/firebase/firebase-ios-sdk/pull/10637 and make sure Settings Fetch is outside the block that disables the SDK
- [ ] Add tests
- [ ] Ensure future backgrounds still resolve the promises

PR #10665, opened 2023-01-10
Add Sessions Internal API

samedson:
Follow on from https://github.com/firebase/firebase-ios-sdk/pull/10636 and https://github.com/firebase/firebase-ios-sdk/pull/10664

- [x] Use non-default Notification Center

# TODO
In a later PR I will add:
 - Tests
 - Testapp Fixes


FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -89,23 +98,103 @@ protocol SessionsProvider {
...
+            )
+          return
+        }
+
+        Logger.logDebug("Data Collection is enabled for at least one Subscriber")
+
+        // Fetch settings if they have expired. This muse happen after the check for
+        // data collection, but before the check for sessionsEnabled from Settings
+        // so that re-fetching can happen.
+        self.settings.updateSettings()
```
leotianlizhan, line 134: `updateSettings()` is async, not sure from the comment if this **must** be done "before the check for sessionsEnabled".

leotianlizhan (approved):
LGTM!

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: Ah let me clarify - we need to make sure we fetch before we check if sessionsEnabled because if we didn't, we could get in a situation where settings disables itself indefinitely

leotianlizhan: Got it, so `updateSettings()` doesn't necessarily need to complete before "check sessionsEnabled", it just has to be initiated?

samedson: Yeah that's right

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -42,6 +38,19 @@ protocol SessionsProvider {
   private let appInfo: ApplicationInfo
   private let settings: SessionsSettings
 
+  /// Subscribers
+  /// `subscribers` are used to determine the Data Collection state of the Sessions SDK.
+  /// If any Subscribers has Data Collection enabled, the Sessions SDK will send events
```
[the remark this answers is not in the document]
samedson: That's correct - I think that's what Michael told us to do

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
```diff
@@ -0,0 +1,56 @@
...
+public protocol SessionsSubscriber {
+  func onSessionChanged(_ session: SessionPayload)
+  var isDataCollectionEnabled: Bool { get }
+  var sessionsSubscriberName: SessionsSubscriberName { get }
+}
+
+/// Session Payload is a container for Session Data passed to Subscribers
+/// whenever the Session changes
+@objc(FIRSessionPayload)
+public class SessionPayload: NSObject {
```
visumickey, line 30: NIT: Can we rename this to "SessionDetails" or "SessionInfo" - Payload sounds more like data to upload/download over network.

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
```diff
@@ -0,0 +1,56 @@
...
+  var sessionId: String
+
+  public init(sessionId: String) {
+    self.sessionId = sessionId
+    super.init()
```
visumickey, line 35: Lines needs to be reversed here. Super.init to be called before using self (Safe practice).

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+/// Session Subscriber Names are used for identifying subscribers
+@objc(FIRSessionsSubscriberName)
+public enum SessionsSubscriberName: Int, CustomStringConvertible {
```
visumickey, line 41: Why is this a "CustomStringConvertible"?

FirebaseSessions/Sources/SessionGenerator.swift
```diff
@@ -62,7 +62,7 @@ class SessionGenerator {
     return newSession
   }
 
-  func currentSession() -> SessionInfo? {
+  var currentSession: SessionInfo? {
```
visumickey, line 65: If we are making it a var, let's enforce no-one to set the value of it.

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: I'm thinking if updateSettings should be done before waiting for all the dependencies initialization. What is the rationale to wait for all the dependencies to be initialized?

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -89,23 +98,104 @@ protocol SessionsProvider {
...
+  func addEventDataCollectionState(event: SessionStartEvent) {
+    subscribers.forEach { subscriber in
+      event.set(subscriber: subscriber.sessionsSubscriberName,
+                isDataCollectionEnabled: subscriber.isDataCollectionEnabled)
+    }
+  }
+
+  // MARK: - SessionsProvider
+
+  var currentSessionPayload: SessionPayload {
```
visumickey, line 173: Can we return NIL instead of empty string?

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -89,23 +98,104 @@ protocol SessionsProvider {
...
+    return SessionPayload(sessionId: sessionGenerator.currentSession?.sessionId ?? "")
+  }
+
+  func register(subscriber: SessionsSubscriber) {
```
visumickey, line 177: Would there be an option for the dependencies to unregister? I think we should provide the API to unsubscribe from hearing updates.

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -89,23 +98,104 @@ protocol SessionsProvider {
 
     super.init()
 
+    SessionsDependencies.dependencies.forEach { subscriberName in
```
visumickey, line 101: Where do these dependencies come from? May be I'm missing that part of the code in my eyes?

FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -78,6 +78,19 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
     proto.session_data.data_collection_status.session_sampling_rate = samplingRate
   }
 
+  func set(subscriber: SessionsSubscriberName, isDataCollectionEnabled: Bool) {
```
visumickey, line 81: Should we remove the lines 67-70 since we have a separate setter function now? Also, where is this `set:` API called?

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
samedson: I like SessionDetails

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
samedson: In Swift I believe it's actually required to set all variables before `super.init()`

Screenshot 2023-01-12 at 11 22 44 AM

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
samedson: So that we can `print` it - That is what the `description` property is implemented for

FirebaseSessions/Sources/SessionStartEvent.swift
samedson: We want to keep it in case the SDK isn't installed - but actually I need to handle that case and set them to NOT_INSTALLED

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: This part isn't called yet - when we integrate with products, they will call this eg: https://github.com/firebase/firebase-ios-sdk/pull/10664/files#diff-3c7ed577af56820c03d0e93aaff77cc09d547a85837e6e57c66401ae85330033R195

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: I don't see a need to yet - Crash and Perf all live for the lifetime of the app

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: I wanted to return the object in all cases to prevent crashes in the product SDKs. If the sessionID is empty, maybe we'd want to know via a log, but I don't want Crash and Perf to have to handle nil cases. WDYT?

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
visumickey: Interesting. Something for me to read/catchup! Thanks for verifying.

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
visumickey: Gotcha!

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: Fair point. But have a thought around how the unregister will look like. Infact, if the product data collection flags toggle, in the future, we might want to honor that by avoiding transactions across SDKs. So, an unregister would be required in that context (not immediately).

FirebaseSessions/Sources/Public/SessionsSubscriber.swift
visumickey: Ack!

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: I would prefer to send NIL and handle the case in the product SDKs. I would want to be able to differentiate situations where a sessionID is valid and invalid. With empty it takes it to an unpredictable scenario.

Aside: Even if empty, we might have to do a check on the session ID length in the product SDKs. Isn't that true? So, instead of empty check, we could as well do a NIL check?

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: > 

Yes, that is totally a valid point. Agreed! This makes sense now.

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: Perfect, that completes the loop for me. Like I mentioned, capturing this in a doc would very helpful.

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: Sounds good - just updated that

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: Sounds good - maybe we can add this in a follow up

visumickey: Sounds good.

visumickey (approved):

FirebaseSessions/Sources/FirebaseSessions.swift
visumickey: Thanks for updating it.

PR #10666, opened 2023-01-11
[Auth Exchange] Add a model for an ExchangeTokenResponse

andrewheard:
Added a model for the response to an `ExchangeInstallationAuthTokenRequest` or `ExchangeCustomTokenRequest`.

#no-changelog

### DIFF
--- a/Crashlytics/Crashlytics/Helpers/FIRCLSDefines.h
+++ b/Crashlytics/Crashlytics/Helpers/FIRCLSDefines.h
@@ -23,9 +23,8 @@
 // These macros generate a function to force a symbol for the containing .o, to work around an issue
 // where strip will not strip debug information without a symbol to strip.
 #define DUMMY_FUNCTION_NAME(x) CONCAT(fircls_strip_this_, x)
-#define INJECT_STRIP_SYMBOL(x)        \
-  void DUMMY_FUNCTION_NAME(x)(void) { \
-  }
+#define INJECT_STRIP_SYMBOL(x) \
+  void DUMMY_FUNCTION_NAME(x)(void) {}
 
 // These make some target os types available to previous versions of xcode that do not yet have them
 // in their SDKs
--- a/FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
+++ b/FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
@@ -60,3 +60,66 @@ struct ExchangeOIDCTokenRequest: Encodable {
     self.authCodeCredentials = authCodeCredentials
   }
 }
+
+// MARK: - Response Models
+
+/// Response that encapsulates an Auth Exchange access token.
+///
+/// This is the return value for an `ExchangeInstallationAuthTokenRequest` or
+/// `ExchangeCustomTokenRequest`.
+struct ExchangeTokenResponse: Decodable, Equatable {
+  /// A container for a Firebase  access token and the duration of time until it expires.
+  struct AuthExchangeToken: Decodable, Equatable {
+    /// A signed [JWT](https://tools.ietf.org/html/rfc7519) containing claims that identify a user.
+    let accessToken: String
+
+    /// The duration of time until the `accessToken` expires, approximately relative to the time
+    /// this response was received.
+    let timeToLive: ProtobufDuration
+
+    enum CodingKeys: String, CodingKey {
+      case accessToken
+      case timeToLive = "ttl"
+    }
+  }
+
+  /// An Auth Exchange access token that can be used to access Firebase services.
+  let token: AuthExchangeToken
+}
+
+/// Model of a `google.protobuf.Duration`, which represents a time span.
+///
+/// A Protocol Buffer
+/// [`Duration`](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#duration)
+/// represents a signed, fixed-length span of time represented as a count of seconds and fractions
+/// of seconds at nanosecond resolution. It is independent of any calendar and concepts like "day"
+/// or "month".
+struct ProtobufDuration: Equatable {
+  private static let nanosecondsPerSecond = 1e9 // 1 x 10⁹ nanoseconds / second
+
+  /// The number of seconds in the time span.
+  let seconds: Int64
+
+  /// The fraction of a second, at nanosecond resolution, in addition to `seconds` in the time span.
+  let nanoseconds: Int32
+
+  /// Floating-point representation of the time span in seconds.
+  var duration: TimeInterval {
+    return TimeInterval(seconds) + Double(nanoseconds) / ProtobufDuration.nanosecondsPerSecond
+  }
+
+  /// Creates a new instance from a JSON string representation of a `Duration`.
+  ///
+  /// - parameter json: a string with the format "`{seconds}.{nanoseconds}s`", with nanoseconds
+  ///   expressed as fractional seconds.
+  init(json: String) throws {
+    (seconds, nanoseconds) = try parseDuration(text: json)
+  }
+}
+
+extension ProtobufDuration: Decodable {
+  init(from decoder: Decoder) throws {
+    let durationString = try decoder.singleValueContainer().decode(String.self)
+    try self.init(json: durationString)
+  }
+}
--- a/FirebaseAuthExchange/Tests/Unit/ExchangeServiceModelTests.swift
+++ b/FirebaseAuthExchange/Tests/Unit/ExchangeServiceModelTests.swift
@@ -113,7 +113,9 @@ final class ModelTests: XCTestCase {
 
   func testDecodeExchangeTokenResponse() throws {
     let accessToken = "test-access-token"
-    let timeToLive = "3.141592653s"
+    let timeToLiveSeconds: Int64 = 3
+    let timeToLiveNanos: Int32 = 141592653
+    let timeToLive = "\(timeToLiveSeconds).\(timeToLiveNanos)s"
     let responseJSON = """
     {
       "token": {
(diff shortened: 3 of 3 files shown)


FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
```diff
@@ -60,3 +60,66 @@ struct ExchangeOIDCTokenRequest: Encodable {
...
+  }
+
+  /// Creates a new instance from a JSON string representation of a `Duration`.
+  ///
+  /// - parameter json: a string with the format "`{seconds}.{nanoseconds}s`", with nanoseconds
+  ///   expressed as fractional seconds.
+  init(json: String) throws {
+    (seconds, nanoseconds) = try parseDuration(text: json)
+  }
+}
```
rosalyntan, line 118: This might be a basic question, but would you be able to just use a class like [`UnitDuration`](https://developer.apple.com/documentation/foundation/nsunitduration?language=objc) (which seems to support nanoseconds) instead of making a new struct here?

andrewheard: To be honest I hadn't ever run across the [Units and Measurements](https://developer.apple.com/documentation/foundation/units_and_measurement) APIs in `Foundation` so thanks for pointing this out. It's nice that it handles the units when doing operations like addition, e.g., `Measurement(value: 1.0, unit: UnitDuration.hours) + Measurement(value: 5, unit: UnitDuration.minutes)`. Unfortunately, [`UnitSeconds.nanoseconds`](https://developer.apple.com/documentation/foundation/unitduration/3172540-nanoseconds) was added in iOS 13.

We may want to revisit this decision in the future when we're targeting iOS 13+ but for now I think a custom struct is simpler. In an ideal world we'd use a [`Duration`](https://developer.apple.com/documentation/swift/duration) since it's a value (struct) type and supports nanoseconds but it was only added in iOS 16 :cry:.

Crashlytics/Crashlytics/Helpers/FIRCLSDefines.h
```diff
@@ -23,9 +23,8 @@
 // These macros generate a function to force a symbol for the containing .o, to work around an issue
 // where strip will not strip debug information without a symbol to strip.
 #define DUMMY_FUNCTION_NAME(x) CONCAT(fircls_strip_this_, x)
-#define INJECT_STRIP_SYMBOL(x)        \
-  void DUMMY_FUNCTION_NAME(x)(void) { \
-  }
+#define INJECT_STRIP_SYMBOL(x) \
+  void DUMMY_FUNCTION_NAME(x)(void) {}
```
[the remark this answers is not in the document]
andrewheard: This was done to silence a formatting [error](https://github.com/firebase/firebase-ios-sdk/actions/runs/3888342893/jobs/6635556377). I wonder if this will occur more frequently as we diverge further from `master`.

paulb777 (commented):
LGTM, deferring to others for approval

FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
```diff
@@ -60,3 +60,66 @@ struct ExchangeOIDCTokenRequest: Encodable {
...
+/// of seconds at nanosecond resolution. It is independent of any calendar and concepts like "day"
+/// or "month".
+struct ProtobufDuration: Equatable {
+  private static let nanosecondsPerSecond = 1e9 // 1 x 10⁹ nanoseconds / second
+
+  /// The number of seconds in the time span.
+  let seconds: Int64
+
+  /// The fraction of a second, at nanosecond resolution, in addition to `seconds` in the time span.
+  let nanoseconds: Int32
```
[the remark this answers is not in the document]
andrewheard: I'm not actually concerned with the memory footprint of `Int32` vs. `Int`/`Int64` but just wanted to be consistent with the data types in [`google.protobuf.Duration`](https://github.com/protocolbuffers/protobuf/blob/c80e7efac72510a2bc3e9365520055f6d6656c1d/src/google/protobuf/duration.proto#L108-L114). FWIW, the `swift-protobuf` code generator also produces an `Int32` ([duration.pb.swift](https://github.com/apple/swift-protobuf/blob/b0c4726327fd5b700cc2f0531a499712efa01028/Sources/SwiftProtobufCore/duration.pb.swift#L120-L126)).

The reasoning behind 32-bit being sufficient is that the nanos portion should be at most 999,999,999 (or else it's a full second) and that'll still fit in a signed 32-bit integer type.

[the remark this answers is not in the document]
paulb777: SGTM, thanks.

FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
```diff
@@ -60,3 +60,66 @@ struct ExchangeOIDCTokenRequest: Encodable {
...
+}
+
+/// Model of a `google.protobuf.Duration`, which represents a time span.
+///
+/// A Protocol Buffer
+/// [`Duration`](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#duration)
+/// represents a signed, fixed-length span of time represented as a count of seconds and fractions
+/// of seconds at nanosecond resolution. It is independent of any calendar and concepts like "day"
+/// or "month".
+struct ProtobufDuration: Equatable {
```
daymxn, line 97: Haha we were just discussing whether to implement a type for `Duration`, or just do a quick cast on the spot. Do you intend to expose this to the public (or something similar), or is this just for the network (service) layer?

FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
```diff
@@ -60,3 +60,66 @@ struct ExchangeOIDCTokenRequest: Encodable {
...
+    /// A signed [JWT](https://tools.ietf.org/html/rfc7519) containing claims that identify a user.
+    let accessToken: String
+
+    /// The duration of time until the `accessToken` expires, approximately relative to the time
+    /// this response was received.
+    let timeToLive: ProtobufDuration
+
+    enum CodingKeys: String, CodingKey {
+      case accessToken
+      case timeToLive = "ttl"
```
daymxn, line 82: Yeah I agree, `timeToLive` makes more sense from our end as well. Gonna go ahead and add a similar notion in ours to do the same (even though it won't go beyond network calls- still looks cleaner)

:)

Crashlytics/Crashlytics/Helpers/FIRCLSDefines.h
```diff
@@ -23,9 +23,8 @@
 // These macros generate a function to force a symbol for the containing .o, to work around an issue
```
daymxn, line 23: ooo, you get macros in swift??

FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
andrewheard: This is just for the network layer so keeping it as a `String` like you did in [`AuthExchangeTokenP`](https://github.com/firebase/firebase-android-sdk/blob/1817518ab0d4286e95028126ea946c125c6153a6/firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt#L36) is also something we should consider for the Swift SDK. I liked the idea of having access to the seconds and nanos components for testing purposes (vs. floating-point values) but I do need to convert it to a [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval) (i.e., a `Double`) eventually in order to expose it as `expirationDate` of type [`Date`](https://developer.apple.com/documentation/foundation/date) in the public [`AuthExchangeToken`](https://github.com/firebase/firebase-ios-sdk/blob/8cdc6fef63310e9c861689dbb0f0d0829ef22c86/FirebaseAuthExchange/Sources/AuthExchangeToken.swift#L16-L25): https://github.com/firebase/firebase-ios-sdk/blob/8cdc6fef63310e9c861689dbb0f0d0829ef22c86/FirebaseAuthExchange/Sources/AuthExchangeToken.swift#L16-L25

Planning to use [`init(timeIntervalSince1970: TimeInterval)`](https://developer.apple.com/documentation/foundation/date/1780353-init) for that.

rosalyntan: Yeah in the[ App Check Android implementation](https://github.com/firebase/firebase-android-sdk/blob/df79e47b6d2fc14552f8ac9053bb6a190d325642/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenResponse.java#L38), we just kept it as a String, but I'm fine with either approach.

FirebaseAuthExchange/Tests/Unit/ExchangeServiceModelTests.swift
```diff
@@ -97,11 +101,41 @@ final class ModelTests: XCTestCase {
...
+    }
+    """
+    let expectedResponse = ExchangeTokenResponse(token: ExchangeTokenResponse.AuthExchangeToken(
+      accessToken: accessToken,
+      timeToLive: try ProtobufDuration(json: timeToLive)
+    ))
+
+    let response = try jsonDecoder.decode(ExchangeTokenResponse.self, from: responseJSON)
+
+    XCTAssertEqual(response, expectedResponse)
```
rosalyntan, line 132: nit: Can you also assert the expected number of seconds and nanoseconds stored in the `ProtobufDuration` object?

rosalyntan (approved):

charlotteliang (approved):

Crashlytics/Crashlytics/Helpers/FIRCLSDefines.h
andrewheard: In Swift there are [Compiler Control Statements](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#ID538) but not macros in the sense that you're likely thinking of. This is just a C header file that ended in the PR due to a change in clang-format. :stuck_out_tongue:

FirebaseAuthExchange/Sources/Models/ExchangeServiceModels.swift
andrewheard: Thanks! That'll make it easier for anyone reviewing both codebases.

PR #10669, opened 2023-01-11
Add Tests Against Sessions SDK and Internal API, Organize Errors

samedson:
This PR adds tests against the Sessions SDK itself. This adds tests for the Internal API that other integrated Product SDKs will interact with.

In addition, this adds a centralized place to log errors that come from the Sessions startup, or the SessionCoordinator when logging the event.

#no-changelog 

This adds a bunch of tests using the following function that makes it fairly easy to write tests against the SDK:

```
  /// This function forms the basis for all tests of type `FirebaseSessionsTests`. It's written
  /// so that you don't need to setup expectations for each test individually.
  ///
  /// It has 4 parts:
  ///  - `subscriberSDKs` tells the test to expect the list of subscriber SDKs
  ///  - `preSessionsInit`  is before Sessions or any Subscriber SDKs start up. This is a good
  /// place to mock variables in any dependencies (eg. Settings, or mocking any Subscribers themselves)
  ///  - `postSessionsInit` is after Sessions has started up, but before logging any events. This
  /// is a good place for Subscribers to call register on the Sessions SDK
  ///  - `postLogEvent` is called whenever an event is logged via the Sessions SDK. This is where
  /// most assertions will happen.
  func runSessionsSDK(subscriberSDKs: [SessionsSubscriber],
                      preSessionsInit: (MockSettingsProtocol) -> Void,
                      postSessionsInit: () -> Void,
                      postLogEvent: @escaping (Result<Void, FirebaseSessionsError>,
                                               [SessionsSubscriber]) -> Void) {
```

Adds tests for things like:
 - Data Collection disabling settings and Session Logging
 - Subscribers getting onSessionChanged notifications
 - Settings disabling Session logging via Sampling / sessionsEnabled


FirebaseSessions/Tests/Unit/FirebaseSessionsTests+Subscribers.swift
```diff
@@ -0,0 +1,126 @@
...
+      }, postLogEvent: { result, subscriberSDKs in
+        // Make sure we didn't do any data collection
+        self.assertFailure(result: result, expectedError: .NoDependenciesError)
+        XCTAssertFalse(self.mockSettings.updateSettingsCalled)
+        XCTAssertNil(self.mockCoordinator.loggedEvent)
+      }
+    )
+  }
+
+  func test_noSubscribersWithRegistrations_doesNotBlowUp() {
```
leotianlizhan, line 105: 😂

FirebaseSessions/Tests/Unit/Error+EquatableTests.swift
```diff
@@ -0,0 +1,80 @@
...
+    return String(reflecting: lhs) == String(reflecting: rhs)
+  }
+}
+
+enum FakeError: Error {
+  case Fake
+}
+
+final class ErrorEquatableTests: XCTestCase {
+  func test_equalErrorTypes_areEqual() throws {
```
leotianlizhan, line 38: Is the whole point of these 2 tests to verify the extension above? and the extension above is used in other tests?

leotianlizhan (commented):
Halfway done, but wanted to send these comments out.

FirebaseSessions/Tests/Unit/Error+EquatableTests.swift
samedson: Exactly. I want to use XCTAssertEqual on Errors, but Errors do not have equality checks natively, and I didn't want to put Error equality into the SDK because it's not perfect.

Let me add a comment on that

samedson: Oh nvm I have a comment at the top of the file

leotianlizhan (approved):
LGTM

FirebaseSessions/Sources/Settings/SettingsProtocol.swift
```diff
@@ -17,13 +17,15 @@ import Foundation
 
 /// Provides the APIs to access Settings and their configuration values
 protocol SettingsProtocol {
+  // Update the settings for all the settings providers
+  func updateSettings()
```
visumickey, line 21: Curious on this refactor: After the previous refactor, we had two protocols with the following needs:

1. SettingsProvider - Provides settings from any service (eg: Crashlytics settings, SDK defaults, LocalOverrides)
2. SettingsProtocol - Just a service that provides the config values.

With the new refactor, we have a new API updateSettings in the SettingsProtocol. Feel like the SettingsProtocol does more than just abstracting the configs. What do you think would be the purpose of the SettingsProtocol in this new behavior?

FirebaseSessions/Tests/Unit/FirebaseSessionsTests+BaseBehaviors.swift
```diff
@@ -0,0 +1,148 @@
...
+#if SWIFT_PACKAGE
+  import FirebaseSessionsObjC
+#endif // SWIFT_PACKAGE
+
+@testable import FirebaseSessions
+
+final class FirebaseSessionsTests_BaseBehaviors: FirebaseSessionsTests {
+  // MARK: - Test Settings & Sampling
+
+  func test_settingsDisabled_doesNotLogSessionEventButDoesFetchSettings() {
```
visumickey, line 27: These are such powerful tests to validate the behavior of the library in whole. Great work here!

FirebaseSessions/Tests/Unit/FirebaseSessionsTests.swift
```diff
@@ -0,0 +1,146 @@
...
+@testable import FirebaseSessions
+
+///
+/// This is the parent class for tests that test against the Sessions internal
+/// API or test Sessions `init` functionality.
+///
+/// Test cases should only go in subclasses because all tests in the parent
+/// class will be run by subclasses.
+///
+class FirebaseSessionsTests: XCTestCase {
```
visumickey, line 31: Why is this extended out of XCTestCase while most of the job done by this class is Utility driven? I would go to the extend of calling it FirebaseSessionsMock.

visumickey (approved):
It is a huge change with so many files impacted. The overall change looks great to me.

I would recommend doing baby steps refactor if possible. I understand that it might be tricky to split the change, but reviewing  mammoth change is equally challenging :)

Thanks a lot of making this incredible impactful change making the whole library easy and testable.

FirebaseSessions/Sources/Settings/SettingsProtocol.swift
samedson: The purpose of SettingsProtocol should be the same - just that it abstracts the configs. The only difference is before we were calling updateSettings on the SessionsSettings class directly instead of using the protocol internally: https://github.com/firebase/firebase-ios-sdk/blob/de6eb3f971ddb9c6ad0a27b78d698a95d1d77d3e/FirebaseSessions/Sources/FirebaseSessions.swift#L177

See https://github.com/firebase/firebase-ios-sdk/pull/10669/files#diff-04ac03aa6a3c3aae8fb90eb2b10c05ee112682853acbd903aa7abc7db089e94cR39

FirebaseSessions/Tests/Unit/FirebaseSessionsTests.swift
samedson: Great question - I wanted to be able to reference the instance variables, so I made this a base class. It subclasses XCTestCase so that other classes can be subclasses of it.

I think maybe renaming it to FirebaseSessionsTestsBase might help convey what I meant.

Or I might go with encapsulation over subclassing in this case as you suggested

PR #10674, opened 2023-01-12
Enable Autopush and Staging tests for E2E test apps.

visumickey:
Add staging and autopush runs of the test apps every 4 hours.

#no-changelog


leotianlizhan (approved):
lgtm

samedson (approved):
LGTM - I would double check it actually updates the dev environment. We might want to add a log here so we know when we've set a value for the env: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/Sources/ApplicationInfo.swift#L103

visumickey:
> LGTM - I would double check it actually updates the dev environment. We might want to add a log here so we know when we've set a value for the env: https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/Sources/ApplicationInfo.swift#L103

Environment is a part of the event and we list the environment value in the DevEventConsoleLogger. Do you think it makes sense to add another log statement in ApplicationInfo?

PR #10682, opened 2023-01-13
Add required dependency between Crashlytics and Session

samedson:
### History
Alternative to https://github.com/firebase/firebase-ios-sdk/pull/10619 where the dependency is required.
Replaces https://github.com/firebase/firebase-ios-sdk/pull/10636

### TODO
- [x] Add tests done in https://github.com/firebase/firebase-ios-sdk/pull/10669
- [x] Ensure future backgrounds still resolve the promises done in https://github.com/firebase/firebase-ios-sdk/pull/10669
- [ ] Send the Session ID with the crash report (going to follow up)

### DIFF
--- a/Crashlytics/CHANGELOG.md
+++ b/Crashlytics/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [added] Added dependency on Firebase Sessions SDK to power future metrics and debugging features in Crashlytics
+
 # 10.4.0
 - [added] Updated Crashlytics to include the Firebase Installation ID for consistency with other products (#10645).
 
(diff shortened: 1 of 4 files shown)


FirebaseSessions/Sources/Public/SessionsSubscriber.swift
```diff
@@ -28,7 +28,7 @@ public protocol SessionsSubscriber {
 /// whenever the Session changes
 @objc(FIRSessionDetails)
 public class SessionDetails: NSObject {
-  var sessionId: String?
+  @objc public var sessionId: String?
```
leotianlizhan, line 31: FYI I moved this to a separate PR that I hope we could merge early because this fix is needed for my PR to work as well

samedson: Awesome - thanks for doing that

PR #10684, opened 2023-01-13
Fix Sessions Test App with Fake Subscriber

samedson:
Adds a fake subscriber SDKs to our test app with data collection enabled

### DIFF
--- a/FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
+++ b/FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
@@ -20,6 +20,8 @@ import FirebaseSessions
 @main
 class AppQualityDevAppApp: App {
   required init() {
+    // In other Product SDKs, this is called via `+ load`, but
+    // we're faking that here because Swift doesn't have `+ load`
     MockSubscriberSDK.addDependency()
 
     FirebaseApp.configure()
--- a/FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
+++ b/FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
@@ -18,9 +18,7 @@ import FirebaseCore
 import FirebaseSessions
 
 @main
-class AppQualityDevAppApp: App, SessionsSubscriber {
-
-
+class AppQualityDevAppApp: App {
   required init() {
     MockSubscriberSDK.addDependency()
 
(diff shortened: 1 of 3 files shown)


visumickey (approved):
Nice. Let's make sure the hack does not stay for long! But a good solution.

FirebaseSessions/Tests/TestApp/Shared/MockSubscriberSDK.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+@objc(FIRMockSubscriberSDK) final class MockSubscriberSDK: NSObject, Library, SessionsSubscriber,
+  MockSubscriberSDKProtocol {
+  static func addDependency() {
+    FirebaseApp.registerInternalLibrary(
+      MockSubscriberSDK.self,
+      withName: "mock-firebase-sessions-subscriber-sdk"
+    )
+    SessionsDependencies.addDependency(name: SessionsSubscriberName.Performance)
```
leotianlizhan, line 34: So this mock sdk is pretending to be Fireperf right?

FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
```diff
@@ -15,10 +15,13 @@
 
 import SwiftUI
 import FirebaseCore
+import FirebaseSessions
 
 @main
-struct AppQualityDevAppApp: App {
-  init() {
+class AppQualityDevAppApp: App {
+  required init() {
```
leotianlizhan, line 22: I remember when talking about how session dependency works we said the dependencies would be added when loaded to memory, like `+ load()`, so is `required init()` equivalent to `+ load()`?

leotianlizhan (approved):
LGTM

FirebaseSessions/Tests/TestApp/Shared/MockSubscriberSDK.swift
samedson: That's right

FirebaseSessions/Tests/TestApp/Shared/AppQualityDevAppApp.swift
samedson: So I'm cheating a little bit. Swift doesn't have the same `load` function (at least I couldn't figure it out in 5 mins of Googling), so I just call this function at the start up the app, but before the Firebase SDKs start up.

So technically this solution is called later than `load`, but it's earlier than Firebase Initialization `FirebaseApp.configure()`, so it's ok.

leotianlizhan: Got it, makes sense!

PR #10696, opened 2023-01-17
[WIP] Attempt at merging own ported swift code to official repo

mortenbekditlevsen:
I finally got my own auth port and unit tests building. 5 unit tests fail, and this needs a ton of cleaning up.

But perhaps it could be a beginning at contributing to the auth port to Swift.

I just wanted to share it, now that everything is compiling. Will continue cleanup work.


FirebaseAuth/Tests/Unit/SwiftAPI.swift
```diff
@@ -31,7 +31,7 @@ class AuthAPI_hOnlyTests: XCTestCase {
   // Each function corresponds with a public header.
   func FIRActionCodeSettings_h() {
     let codeSettings = FirebaseAuth.ActionCodeSettings()
-    codeSettings.setIOSBundleID("abc")
+    codeSettings.iOSBundleID = "abc"
```
paulb777, line 34: Looks like a bug in the existing API, but we'll still need to call this out for API review.

paulb777:
I see one test failure running the unit tests in Nick's SPM:

```
/Users/paulbeusterien/gh/firebase-ios-sdk/FirebaseAuth/Sources/Auth/FIRAuth.m:417: error: -[FIRAuthBackendCreateAuthURITests testRequestAndResponseEncoding] : The default FirebaseApp instance must be configured before the default Authinstance can be initialized. One way to ensure this is to call `FirebaseApp.configure()` in the App Delegate's `application(_:didFinishLaunchingWithOptions:)` (or the `@main` struct's initializer in SwiftUI). (NSInternalInconsistencyException)
```
I suspect we need to do explicit recognition of Auth in `registerSwiftComponents` https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseCore/Sources/FIRApp.m#L820

@mortenbekditlevsen Are you using Xcode to develop? I'm not able to build your branch with `pod gen` and don't know of a way to integrate a different SPM into Xcode.

mortenbekditlevsen:
Hi Paul,
Yeah, I'm down to that same single error, and I figured that it was probably due to everything that is skipped when building.

With help from your comments, I have been able to build and test using `swift-test` from the custom spm build.
I am still using Xcode as an editor, but not for building... :-) 
Not quite as nice a way of presenting the errors, but it has been ok.

mortenbekditlevsen:
Question: In a few places (StartMFASignInResponse.swift is one), the corresponding Objective-C code did not completely follow convention, so it returned 'false', indicating an error, but the error variable was not set.

I'm pretty sure that this imports to Swift by throwing some internal 'error missing' error. But in the Swift-code we of course need an actual error to throw. Is there an appropriate error here?

mortenbekditlevsen:
One thing to note is that a lot of the code in this PR was already converted around September, 2022, so if there were any bug fixes, I would likely have missed them. I did find a few places where new API was added, which was evident from tests and related code - and I assume that any bug fixes would also have unit tests to prevent regressions, so this could indicate that the code is ok.

I haven't been completely consistent in how I copied over comments, but I also don't know if they should perhaps be formatted in a different way in Swift. Let me know if there's any changes you would like in this area.

paulb777:
We may not be able to answer your questions in a timely way, so it may be best to make an assumption.  I'll start and share a Google doc or spreadsheet later this week to capture API changes, TODOs, likely existing bugs, and any other open questions.

If I get time this week, I'll take a stab at getting your changes to build from a pod-gen Xcode project.

paulb777:
For coding style questions, see https://google.github.io/swift

paulb777:
Building in CocoaPods, I'm seeing 
Screen Shot 2023-01-19 at 3 17 10 PM

I'm not sure how this builds in SPM, since the FIRVerifyPhoneNumberRequest.m is gone and there is not a replacement yet. cc: @ncooke3

mortenbekditlevsen:
Hi Paul,
Ah, yes, when I run tests I am not building for iOS, so I've been missing iOS-specific things.
It compiles for macOS since the code is behind platform checks.
I'll add the VerifyPhoneNumberRequest and Response and try and see if I can use cocoapods for building with Xcode.

mortenbekditlevsen:
Hi again,
I added the missing types, but I would like to be able to test this myself too.
When I run `pod gen` I get the following error:

```
pod gen FirebaseAuth.podspec --local-sources=./ --auto-open --platforms=ios


Generating workspace in `gen/FirebaseAuth`
[!] CocoaPods could not find compatible versions for pod "GoogleUtilities/AppDelegateSwizzler":
  In Podfile:
    FirebaseAuth (from `../../FirebaseAuth.podspec`) was resolved to 10.4.0, which depends on
      GoogleUtilities/AppDelegateSwizzler (~> 7.8)

None of your spec sources contain a spec satisfying the dependency: `GoogleUtilities/AppDelegateSwizzler (~> 7.8)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
```
Do you have any pointers for running pod gen?

paulb777:
You might need to run `pod repo update`.  

With a few additional changes, I'm able to build your PR with `pod gen` for macos. 

I'll merge your PR to my branch and then send another PR with the fixes for macos and `pod gen`.

paulb777:
Please use the `auth-swift` branch going forward as a long-lived feature branch for the Auth Swift work.

FirebaseAuth/Sources/Swift/Utilities/AuthErrorUtils.swift
```diff
@@ -0,0 +1,540 @@
...
+    @objc public static func secondFactorRequiredError(resolver: MultiFactorResolverWrapper?) -> Error {
+        var userInfo: [String: Any] = [:]
+
+        if let resolver {
+            userInfo[FIRAuthErrorUserInfoMultiFactorResolverKey] = resolver
+        }
+
+        return error(code: .secondFactorRequired, userInfo: userInfo)
+    }
+
```
ncooke3, line 535: Hi @mortenbekditlevsen,  I'm fixing the build for iOS. [This line](https://github.com/firebase/firebase-ios-sdk/blob/b9762b438c38c05e0ffbfb324319eed2a4a43090/FirebaseAuth/Sources/Backend/FIRAuthBackend.m#L777) is failing to build on iOS. Based on the Swift signature here, I'm not seeing a direct translation. What do you think?

ncooke3: Should the API look something like?
```suggestion
secondFactorRequiredErrorWithPendingCredential:response.MFAPendingCredential
//                                                                hints
    @objc public static func secondFactorRequiredError(
      pendingCredential: // ...
      hints: // ...
    ) -> Error {
        var userInfo: [String: Any] = [:]

        if let resolver {
            userInfo[FIRAuthErrorUserInfoMultiFactorResolverKey] = resolver
        }

        return error(code: .secondFactorRequired, userInfo: userInfo)
    }

```

mortenbekditlevsen: Yeah, sorry, this looks like a work in progress. I hope to be able to build using Xcode at some point, so I can build the iOS-specific code to.

I'm not completely certain about the visibility of the  `FIRMultiFactorResolver`, but with the obj-c import it should be something like this:

```
  #if os(iOS)

    @objc public static func secondFactorRequiredError(pendingCredential: String?, hints: [MultiFactorInfo]?)
      -> Error {
      var userInfo: [String: Any] = [:]

      if let pendingCredential, let hints {
          let resolver = FIRMultiFactorResolver(MFAPendingCredential: pendingCredential, hints: hints)
          userInfo[FIRAuthErrorUserInfoMultiFactorResolverKey] = resolver
      }

      return error(code: .secondFactorRequired, userInfo: userInfo)
    }
  #endif
```

Would you like a fix on this branch or on the `auth-swift` one?

ncooke3: Ahh, got it. Makes sense. I'm working on #10755 right now which will allow us to build within Xcode when building from CocoaPods. That PR tries to expose non-public headers to the Swift part of the pod. 

Locally, I have been iterating on it and in the process have the library mostly building for iOS. And I'll open a PR after #10755 merges with those fixes to get iOS building. You can push to that branch when ready–– I'll let you know!

PR #10702, opened 2023-01-17
[Infra] Update 'actions/checkout' version in 'update-cpp-sdk-on-release.yaml'

ncooke3:
### Context

When I published 10.4.0., I got an email notification regarding the update-cpp-sdk-on-release.yam [failing](https://github.com/firebase/firebase-ios-sdk/actions/runs/3943423218).

I think that version may not be installed on the VM so I pinned the `checkout` action to the major version.

#no-changelog


paulb777:
@jonsimantov Does the workflow also need to trigger when the yml file is modified or have a manual trigger to force the update after a fail?

jonsimantov (approved):

PR #10704, opened 2023-01-17
Add required dependency between Performance and Sessions

leotianlizhan:
This PR is only the first step, as it only registers the dependency so we can use Sessions SDK and session ID is flowing to us, but not yet being used. This PR is based on https://github.com/firebase/firebase-ios-sdk/pull/10682.

- [x] Step 1: register dependency (this PR). 
- [ ] Step 2: make Fireperf's `SessionManager` use the session IDs we now get from SessionsSDK)

### DIFF
--- a/FirebasePerformance.podspec
+++ b/FirebasePerformance.podspec
@@ -62,6 +62,7 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
   s.dependency 'FirebaseCore', '~> 10.0'
   s.dependency 'FirebaseInstallations', '~> 10.0'
   s.dependency 'FirebaseRemoteConfig', '~> 10.0'
+  s.dependency 'FirebaseSessions', '~> 10.0'
   s.dependency 'GoogleDataTransport', '~> 9.2'
   s.dependency 'GoogleUtilities/Environment', '~> 7.8'
   s.dependency 'GoogleUtilities/ISASwizzler', '~> 7.8'
--- a/FirebasePerformance/Sources/FPRClient.m
+++ b/FirebasePerformance/Sources/FPRClient.m
@@ -60,7 +60,6 @@
 #pragma mark - Component registration system
 
 + (nonnull NSArray<FIRComponent *> *)componentsToRegister {
-  
   FIRComponentCreationBlock creationBlock =
       ^id _Nullable(FIRComponentContainer *container, BOOL *isCacheable) {
     if (!container.app.isDefaultApp) {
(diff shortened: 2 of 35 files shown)


FirebasePerformance/Sources/FPRClient.m
```diff
@@ -340,4 +357,18 @@ - (void)disableInstrumentation {
...
+- (void)onSessionChanged:(FIRSessionDetails *_Nonnull)session {
+  FPRLogDebug(kFPRSessionId, @"Session ID changed: %@", session.sessionId);
+}
+
+- (BOOL)isDataCollectionEnabled {
+  return self.configuration.isDataCollectionEnabled;
+}
+
+- (FIRSessionsSubscriberName)sessionsSubscriberName {
+  return FIRSessionsSubscriberNameCrashlytics;
```
samedson, line 371: Must be Performance

samedson: I wanted a way to ensure this and `addDependencyWithName:` used the same name, but couldn't think of any outside defining a constant and use it both places

leotianlizhan: Good catch, I saw "Registering Sessions SDK subscriber with name: Crashlytics, data collection enabled: true" in the logs but then I thought it's normal, Crashlytics is already integrated, turns out it's just my typo

samedson:
You can change this to merge into sessions-main now: https://github.com/firebase/firebase-ios-sdk/commits/sessions-main

visumickey (commented):
The changes look good to me (and exciting). This is great.

Would there be any tests that needs to be done to ensure that the dependencies are met?

FirebasePerformance.podspec
```diff
@@ -62,6 +62,7 @@ Firebase Performance library to measure performance of Mobile and Web Apps.
   s.dependency 'FirebaseCore', '~> 10.0'
   s.dependency 'FirebaseInstallations', '~> 10.0'
   s.dependency 'FirebaseRemoteConfig', '~> 10.0'
+  s.dependency 'FirebaseSessions', '~> 10.0'
```
samedson, line 65: We should probably actually add a changelog entry for this change. You can write a similar message to https://github.com/firebase/firebase-ios-sdk/pull/10682/files

samedson:
Said this in the other PR:

Screenshot 2023-01-18 at 3 46 33 PM

FirebasePerformance.podspec
leotianlizhan: Done

samedson (approved):
LGTM!

PR #10706, opened 2023-01-17
change sessionId from the subscriber payload to public

leotianlizhan:
Both https://github.com/firebase/firebase-ios-sdk/pull/10704 and https://github.com/firebase/firebase-ios-sdk/pull/10682 requires this change, otherwise Crashlytics and Performance can't access the session id in the payload given by Sessions SDK.

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/Public/SessionsSubscriber.swift
+++ b/FirebaseSessions/Sources/Public/SessionsSubscriber.swift
@@ -28,7 +28,7 @@ public protocol SessionsSubscriber {
 /// whenever the Session changes
 @objc(FIRSessionDetails)
 public class SessionDetails: NSObject {
-  var sessionId: String?
+  @objc public var sessionId: String?
 
   public init(sessionId: String?) {
     self.sessionId = sessionId


FirebaseSessions/Sources/Public/SessionsSubscriber.swift
```diff
@@ -28,7 +28,7 @@ public protocol SessionsSubscriber {
 /// whenever the Session changes
 @objc(FIRSessionDetails)
 public class SessionDetails: NSObject {
-  var sessionId: String?
+  @objc public var sessionId: String?
 
   public init(sessionId: String?) {
```
leotianlizhan, line 33: @samedson Is there a way we can make `sessionId` not optional? This way when payload is passed down to Crashlytics and Fireperf, we don't have to deal with handling "nil sessionId", which I don't think should even exist at this point.

visumickey: It is a tricky situation here. Initially, the implementation @samedson had was to make SessionID non-optional (Where in case where the session ID cannot be generated, the value will be set to Empty). 

But, that had some caveats. Empty sessionID is not valid either. So, the sessionID listeners have a need to validate the accuracy of the sessionID. To handle situations where the new session ID generation has issues, making it an optional would be a easiest way to handle it for future.

visumickey (approved):

samedson (approved):

PR #10711, opened 2023-01-18
Integrate sessionID from SessionsSDK into FireperfSDK

leotianlizhan:
Fireperf now consumes session updates initiated and generated by SessionsSDK. All Fireperf's events will now associate with the AQS session.

- [x] Step 1: register dependency (https://github.com/firebase/firebase-ios-sdk/pull/10704). 
- [x] Step 2: make Fireperf's `SessionManager` use the session IDs we now get from SessionsSDK (this PR).

### DIFF
--- a/FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
+++ b/FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
@@ -31,7 +31,7 @@ typedef NS_OPTIONS(NSUInteger, FPRSessionOptions) {
 @property(nonatomic, readonly) FPRSessionOptions options;
 
 /* Length of the session in minutes. */
-- (NSUInteger)sessionLengthInMinutes:(nonnull NSDate *)now;
+- (NSUInteger)sessionLengthInMinutesFromDate:(nonnull NSDate *)now;
 
 /**
  * Creates an instance of FPRSessionDetails with the provided sessionId and the list of available
--- a/FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
+++ b/FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
@@ -31,7 +31,7 @@ typedef NS_OPTIONS(NSUInteger, FPRSessionOptions) {
 @property(nonatomic, readonly) FPRSessionOptions options;
 
 /* Length of the session in minutes. */
-@property(nonatomic, readonly) NSUInteger sessionLengthInMinutes;
+- (NSUInteger)sessionLengthInMinutes:(nonnull NSDate *)now;
 
 /**
  * Creates an instance of FPRSessionDetails with the provided sessionId and the list of available
--- a/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
+++ b/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
@@ -25,10 +25,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property(atomic, nullable, readwrite) FPRSessionDetails *sessionDetails;
 
 /**
- * Checks if the currently active session is beyond maximum allowed time. If so renew the session,
+ * Checks if the currently active session is beyond maximum allowed time for gauge-collection. If so stop gauges,
  * else no-op.
  */
-- (void)renewSessionIdIfRunningTooLong;
+- (void)stopGaugesIfRunningTooLong;
+
+- (void)updateSessionId:(nonnull NSString *)sessionIdString;
 
 @end
 
--- a/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
+++ b/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
@@ -25,8 +25,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property(atomic, nullable, readwrite) FPRSessionDetails *sessionDetails;
 
 /**
- * Checks if the currently active session is beyond maximum allowed time for gauge-collection. If so stop gauges,
- * else no-op.
+ * Checks if the currently active session is beyond maximum allowed time for gauge-collection. If so
+ * stop gauges, else no-op.
  */
 - (void)stopGaugesIfRunningTooLong;
 
--- a/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
+++ b/FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
@@ -44,14 +44,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)stopGaugesIfRunningTooLong;
 
-/**
- * Checks if the provided sessionId can have gauge data collection enabled.
- *
- * @param sessionId Session Id for which the check is done.
- * @return YES if gauge collection is enabled, NO otherwise.
- */
-- (BOOL)isGaugeCollectionEnabledForSessionId:(NSString *)sessionId;
-
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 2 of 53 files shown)


samedson (commented):
This LGTM! Going to let Visu be the final say because he has more context

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
```diff
@@ -57,41 +55,25 @@ - (FPRSessionManager *)initWithNotificationCenter:(NSNotificationCenter *)notifi
   self = [super init];
   if (self) {
     _sessionNotificationCenter = notificationCenter;
-    _trackingApplicationStateChanges = NO;
-    [self updateSessionId:nil];
+    _sessionDetails = [[FPRSessionDetails alloc] initWithSessionId:@"placeholder"
```
visumickey, line 58: Would there be a functionality change here? In the past, as a part of app startup, we used to generate a sessionID and start collecting gauges.

In this new world
   * Would there be delay in starting to collect gauges?
   * Secondly, starting with placeholder and then moving to an actual sessionID could be tricky since we might end up sending events with the placeholder sessionID. Anything we can do to make sure the gauges are not incorrectly referenced. For eg: creating a cache of events until we get a sessionID, skip data collection until we have a valid sessionID or could be any other alternatives.

FirebasePerformance/Tests/Unit/FPRSessionManagerTest.m
```diff
@@ -34,45 +36,28 @@ - (void)testInstanceCreation {
...
-  XCTAssertNotNil(instance.sessionDetails.sessionId);
-
-  NSString *lowercaseSessionId = [instance.sessionDetails.sessionId lowercaseString];
-  XCTAssertEqualObjects(lowercaseSessionId, instance.sessionDetails.sessionId);
-}
-
-/** Validate that sessionId does not change when calling renew method immediately. */
-- (void)testSessionIdNotGettingRenewed {
+/** Validate that gauge collection does not change when calling renew method immediately. */
+- (void)testGaugeDoesNotStopBeforeMaxDuration {
```
visumickey, line 40: Do we have tests for stopping gauge collection after the session has been running for too long? You can try this by overriding the max duration config and see if we are stopping to collect gauge events after that.

visumickey (commented):
The changes look good to me except for one key scenario - How is the app startup scenario handled? (Default session ID).

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
leotianlizhan: There would be no functionality change because we call `[sessions registerWithSubscriber:[self sharedInstance]];` in the `FIRComponentCreationBlock`. So as soon as Firebase creates Fireperf, `FirebaseSessions.register(subscriber)` will immediately call `onSessionChanged`, giving Fireperf a session ID. This is before any trace is yet created. 
image

In my tests, `onSessionChanged` is always the first one to initialize the singleton `FPRSessionManager`, then immediately call `updateSessionId` as shown above, so actually the gauges are now collected earlier, since we did not have the creation block immediately initialize `FPRSessionManager` before.

FirebasePerformance/Tests/Unit/FPRSessionManagerTest.m
leotianlizhan: I am trying to write this right now, but it's super hard with all the singletons (`[FPRConfigurations sharedInstance]` is directly called in code, no way to pass in a fake) basically blocking me from injecting dependencies. Going with the overriding max duration in config approach, minimum value cannot be 0 minutes because it would be auto-overwritten to default 240 minutes. 1 minute would be too long to wait. This is why I prefer no-singletons, and passing in `Date` as current time, but I can't pass in time because `FPRSessionDetails.sessionLengthInMinutes` is a property.

leotianlizhan: Done, by adding a `(NSDate *)now` param to `stopGaugesIfRunningTooLong`

FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
```diff
@@ -31,7 +31,7 @@ typedef NS_OPTIONS(NSUInteger, FPRSessionOptions) {
 @property(nonatomic, readonly) FPRSessionOptions options;
 
 /* Length of the session in minutes. */
-@property(nonatomic, readonly) NSUInteger sessionLengthInMinutes;
+- (NSUInteger)sessionLengthInMinutes:(nonnull NSDate *)now;
```
visumickey, line 34: Since we are passing parameter now, recommend naming this API: `sessionLengthInMinutesFromDate:(NSDate *) now`.

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
visumickey: Let's make sure to comment these details above to give the context for any future developers.

Aside, instead of the placeholder, can we start with an empty string sessionID (instead of the text placeholder)?

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
```diff
@@ -57,51 +55,40 @@ - (FPRSessionManager *)initWithNotificationCenter:(NSNotificationCenter *)notifi
...
-- (void)updateSessionId:(NSNotification *)notification {
-  NSUUID *uuid = [NSUUID UUID];
-  NSString *sessionIdString = [uuid UUIDString];
-  sessionIdString = [sessionIdString stringByReplacingOccurrencesOfString:@"-" withString:@""];
-  sessionIdString = [sessionIdString lowercaseString];
+- (void)updateSessionId:(NSString *)sessionIdString {
+  [self updateSessionId:sessionIdString forceGauges:false];
+}
 
+- (void)updateSessionId:(NSString *)sessionIdString forceGauges:(BOOL)forceGauges {
```
visumickey, line 81: What is the purpose of the parameter `forceGauges`? Any situation where we force to collect gauges?

visumickey (commented):
Thanks for adding the new tests. Looking more complete now.

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
leotianlizhan: It's for testing-only. I've removed it now and use `[OCMockObject partialMockForObject:...]` instead.

FirebasePerformance/Sources/AppActivity/FPRSessionDetails.h
leotianlizhan: Done

FirebasePerformance/Sources/AppActivity/FPRSessionManager.m
leotianlizhan: Done

FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
```diff
@@ -25,10 +25,12 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
- * Checks if the currently active session is beyond maximum allowed time. If so renew the session,
- * else no-op.
+ * Checks if the currently active session is beyond maximum allowed time for gauge-collection. If so
+ * stop gauges, else no-op.
  */
-- (void)renewSessionIdIfRunningTooLong;
+- (void)stopGaugesIfRunningTooLong:(NSDate *)now;
+
+- (BOOL)isGaugeCollectionEnabledForSessionId:(NSString *)sessionId;
```
visumickey, line 33: Add comments for this API as to what expect of this API.

visumickey (approved):
Awesome Leo! Thanks for making this change.

FirebasePerformance/Sources/AppActivity/FPRSessionManager+Private.h
leotianlizhan: Done

samedson (approved):

PR #10715, opened 2023-01-19
Update Crashlytics SDK to send Session ID in crash reports

samedson:
#no-changelog

TODO:

- [x] Tests
- [x] Incorporate https://github.com/firebase/firebase-ios-sdk/pull/10750

### DIFF
--- a/Crashlytics/Crashlytics/Components/FIRCLSContext.h
+++ b/Crashlytics/Crashlytics/Components/FIRCLSContext.h
@@ -86,6 +86,7 @@ typedef struct {
   const char* rootPath;
   const char* previouslyCrashedFileRootPath;
   const char* sessionId;
+  const char* appQualitySessionId;
   const char* betaToken;
   bool errorsEnabled;
   bool customExceptionsEnabled;
@@ -96,10 +97,12 @@ typedef struct {
 } FIRCLSContextInitData;
 
 #ifdef __OBJC__
-bool FIRCLSContextInitialize(FIRCLSInternalReport* report,
-                             FIRCLSSettings* settings,
-                             FIRCLSFileManager* fileManager);
-
+bool FIRCLSContextInitialize(FIRCLSContextInitData* initData, FIRCLSFileManager* fileManager);
+FIRCLSContextInitData FIRCLSContextBuildInitData(FIRCLSInternalReport* report,
+                                                 FIRCLSSettings* settings,
+                                                 FIRCLSFileManager* fileManager,
+                                                 NSString* appQualitySessionId);
+bool FIRCLSContextRecordMetadata(NSString* rootPath, const FIRCLSContextInitData* initData);
 #endif
 
 void FIRCLSContextBaseInit(void);
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.h
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.h
@@ -1,0 +1,43 @@
+//
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+#import "Crashlytics/Crashlytics/Models/FIRCLSFileManager.h"
+#import "Crashlytics/Crashlytics/Models/FIRCLSInternalReport.h"
+#import "Crashlytics/Crashlytics/Models/FIRCLSSettings.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+///
+/// The ContextManager determines when to build the context object,
+/// and write its metadata. It was created because the FIRCLSContext
+/// is interacted with via functions, which makes it hard to include in tests.
+/// In addition, we this class is responsible for re-writing the Metadata object
+/// when the App Quality Session ID changes.
+///
+@interface FIRCLSContextManager : NSObject
+
+/// This should be set immediately when the FirebaseSessions SDK generates
+/// a new Session ID.
+@property(nonatomic, copy) NSString *appQualitySessionId;
+
+- (BOOL)setupContextWithReport:(FIRCLSInternalReport *)report
+                      settings:(FIRCLSSettings *)settings
+                   fileManager:(FIRCLSFileManager *)fileManager;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.m
+++ b/Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.m
@@ -1,0 +1,78 @@
+//
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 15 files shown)


Crashlytics/Crashlytics/Components/FIRCLSContext.h
```diff
@@ -96,10 +97,12 @@ typedef struct {
 } FIRCLSContextInitData;
 
 #ifdef __OBJC__
-bool FIRCLSContextInitialize(FIRCLSInternalReport* report,
-                             FIRCLSSettings* settings,
-                             FIRCLSFileManager* fileManager);
-
+bool FIRCLSContextInitialize(FIRCLSContextInitData* initData, FIRCLSFileManager* fileManager);
```
themiswang, line 100: Do we still need to pass `fileManager` as another argument here? Seems init Data also has a reference of `fileManager `

themiswang:
I have tried to rerun the failed unit tests but seems like it failed again. Double check with you for the failed ones.

samedson:
@themiswang ah you know what I think it's because we haven't released FirebaseSessions to cocoapods. We've been ignoring those failures for now, and will keep it in mind when we merge to master:

```
$ pod spec lint FirebaseCrashlytics.podspec --skip-tests --platforms=tvos,ios,macos --sources=https://github.com/firebase/SpecsTesting,https://cdn.cocoapods.org/

Cloning spec repo `firebase-specstesting` from `https://github.com/firebase/SpecsTesting`
 -> FirebaseCrashlytics (10.5.0)
    - ERROR | [tvOS] unknown: Encountered an unknown error (Unable to find a specification for `FirebaseSessions (~> 10.0)` depended upon by `FirebaseCrashlytics`

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
) during validation.

Analyzed 1 podspec.
```

themiswang (approved):
LGTM

Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.h
```diff
@@ -0,0 +1,43 @@
...
+///
+/// The ContextManager determines when to build the context object,
+/// and write its metadata. It was created because the FIRCLSContex
+/// is interacted with via functions, which makes it hard to include in tests.
+/// In addition, we this class is responsible for re-writing the Metadata object
+/// when the App Quality Session ID changes.
+///
+@interface FIRCLSContextManager : NSObject
+
+- (BOOL)initContextWithReport:(FIRCLSInternalReport *)report
```
visumickey, line 33: nitpick: Typically functions starting with `init` should return the instance and not a BOOL. If it is not a big refactor, I would recommend returning instance type rather than a BOOL.

Secondly, if we have a custom initializer, would recommend to the world that they don't use default initializer. So, marking `init` as NS_UNAVAILABLE would be a better practice.

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
```diff
@@ -416,7 +420,9 @@ - (BOOL)startCrashReporterWithProfilingMark:(FIRCLSProfileMark)mark
     return NO;
   }
 
-  if (!FIRCLSContextInitialize(report, self.settings, _fileManager)) {
+  if (![self.contextManager initContextWithReport:report
```
visumickey, line 423: When the initializer returns an instance, this would change into a NIL check rather than a bool check.

Crashlytics/Crashlytics/FIRCrashlytics.m
```diff
@@ -153,6 +145,18 @@ - (instancetype)initWithApp:(FIRApp *)app
                                                          settings:settings
                                                     onDemandModel:onDemandModel];
 
+    if (sessions) {
```
visumickey, line 148: If there is a significance in delaying this initialization, would recommend comments qualifying why delayed initialization is necessary.

Crashlytics/Crashlytics/FIRCrashlytics.m
```diff
@@ -391,6 +395,8 @@ - (void)recordOnDemandExceptionModel:(FIRExceptionModel *)exceptionModel {
 
 - (void)onSessionChanged:(FIRSessionDetails *_Nonnull)session {
   FIRCLSDebugLog(@"Session ID changed: %@", session.sessionId);
+
+  [self.managerData.contextManager updateAppQualitySessionId:session.sessionId];
```
visumickey, line 399: Do you want to send a copy of it rather than the received object?

visumickey (approved):
The changes look to me from the context of integration with AQS.

But provided some suggestions on general coding suggestions for a reliable ObjC.

Crashlytics/Crashlytics/Controllers/FIRCLSManagerData.h
```diff
@@ -25,6 +25,7 @@ NS_ASSUME_NONNULL_BEGIN
 @class FIRCLSOnDemandModel;
 @class FIRCLSSettings;
 @class FIRCLSLaunchMarkerModel;
+@class FIRCLSContextManager;
```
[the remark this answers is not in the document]
samedson: Yes - this prevents circular dependencies and (maybe) speeds up builds. In headers you don't need the implementation of a class, just its name to link to later.

Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.h
```diff
@@ -0,0 +1,43 @@
...
+
+#import "Crashlytics/Crashlytics/Models/FIRCLSFileManager.h"
+#import "Crashlytics/Crashlytics/Models/FIRCLSInternalReport.h"
+#import "Crashlytics/Crashlytics/Models/FIRCLSSettings.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+///
+/// The ContextManager determines when to build the context object,
+/// and write its metadata. It was created because the FIRCLSContex
```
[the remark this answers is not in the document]
samedson: Done

Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.m
```diff
@@ -0,0 +1,79 @@
...
+
+@end
+
+@implementation FIRCLSContextManager
+
+- (instancetype)init {
+  self = [super init];
+  if (!self) {
+    return self;
+  }
```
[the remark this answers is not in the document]
samedson: Yeah it's a funny pattern - if you check out the style guide there's a part where in init they always do `if (self) { ...set vars }`

go/objc-style#example

This is just a simplified version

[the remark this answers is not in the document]
samedson: But the advice about ivars is correct - updating that

Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.h
samedson: This is a good callout - init is a bad prefix

Crashlytics/Crashlytics/FIRCrashlytics.m
samedson: I put the comments down where I call `registerWithSubscriber`

Crashlytics/Crashlytics/FIRCrashlytics.m
samedson: That's a great idea

Crashlytics/Crashlytics/Controllers/FIRCLSContextManager.m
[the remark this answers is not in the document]
visumickey: Thanks for updating to `ivars`.

For the first part, i was referring to returning `self` in a `if` check block. Instead can we do more like below:

```
if (self) {
}
```

instead of 

```
if (!self) {
    return self
}
```

visumickey (approved):

Crashlytics/Crashlytics/Controllers/FIRCLSReportManager.m
visumickey: Thanks for making the change.

PR #10717, opened 2023-01-19
Prevent capturing of Network Info unless Perf is installed

samedson:
#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/SessionStartEvent.swift
+++ b/FirebaseSessions/Sources/SessionStartEvent.swift
@@ -57,12 +57,15 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
     proto.application_info.which_platform_info = FIRSESGetAppleApplicationInfoTag()
     proto.application_info.apple_app_info
       .bundle_short_version = makeProtoString(appInfo.appDisplayVersion)
-    proto.application_info.apple_app_info.network_connection_info
-      .network_type = convertNetworkType(networkType: appInfo.networkInfo.networkType)
-    proto.application_info.apple_app_info.network_connection_info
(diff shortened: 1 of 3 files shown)


FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -57,13 +57,16 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
     proto.application_info.which_platform_info = FIRSESGetAppleApplicationInfoTag()
     proto.application_info.apple_app_info
       .bundle_short_version = makeProtoString(appInfo.appDisplayVersion)
-    proto.application_info.apple_app_info.network_connection_info
-      .network_type = convertNetworkType(networkType: appInfo.networkInfo.networkType)
-    proto.application_info.apple_app_info.network_connection_info
-      .mobile_subtype = convertMobileSubtype(mobileSubtype: appInfo.networkInfo.mobileSubtype)
     proto.application_info.apple_app_info.os_name = convertOSName(osName: appInfo.osName)
     proto.application_info.apple_app_info.mcc_mnc = makeProtoString(appInfo.mccMNC)
```
visumickey, line 61: Should mccMNC be also added as a reserved field? I assume Crashlytics does not capture this information?

leotianlizhan (approved):
lgtm

visumickey (approved):

FirebaseSessions/Sources/SessionStartEvent.swift
samedson: Yeah that totally makes sense to do 👍

samedson: Done

PR #10723, opened 2023-01-20
Style auth-swift

paulb777:


paulb777:
cc: @mortenbekditlevsen @ncooke3 

This PR wraps up a string of commits that should leave the `auth-swift` branch buildable for macOS with `pod gen`, running the unit tests with one failure via Nick's SPM branch, and passing the `check` CI workflow.

Please base subsequent Auth Swift PR's on the `auth-swift` branch.

PR #10735, opened 2023-01-24
Fix build warnings and style

paulb777:
Fix style - primarily indentation
Fix warnings - missing header in umbrella and `outError` in doc comments


paulb777:
cc: @mortenbekditlevsen

Issue #10707, opened 2023-01-18
Possible crash if FIRMessagingTokenManager is dealloc'ed while a token fetch is in progress


kelan:
### Description

I am seeing a crash in `FIRMessagingTokenManager` which is 100% repeatable during some login/logout flows in our app, and also is fairly repeatable in our automated tests on CI that use the app as the test host (which has the Firebase SDK started as normal), but aren't doing anything directly with this class.

The issue is that `-[FIRMessagingTokenManager fetchNewTokenWithAuthorizedEntity:scope:instanceID:options:handler:]` does a "weak/strong dance" for self in its completion block, but it doesn't check for `nil` `self` (after strongifying it) before dereferencing it on line 315 (`self->_tokenStore saveTokenInfo:tokenInfo handler:…`). So it crashes if the instance has been dealloced, thus `self` is now `nil`.

I think we are releasing our strong reference to the instance right after requesting the token, because we async'ly find that another auth token is invalid, so we want to reset the app to a signed-out state. I think there is a similar race during our automated CI tests which hit this crash fairly often.

Looking at the code, it seems like it should check for nil self after calling `FIRMessaging_STRONGIFY(self)` on FIRMessagingTokenManager.m:298, and do an early return.

### Firebase SDK Version

At least in 9.5.0 and the latest (10.4.0)

### Xcode Version

14.2 (14C18)

### Installation Method

Swift Package Manager

### Firebase Product(s)

Messaging

### Targeted Platforms

iOS

PR #10740, opened 2023-01-26
[Messaging] - Fix crash for strongSelf dereference

aashishpatil-g:
### Discussion

Fix for https://github.com/firebase/firebase-ios-sdk/issues/10707

Added a check for self within the block. Found some other places in the file and fixed those too. 

#no-changelog

### DIFF
--- a/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
+++ b/FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
@@ -239,24 +239,25 @@
   }
 
   FIRMessaging_WEAKIFY(self);
-  [_authService
-      fetchCheckinInfoWithHandler:^(FIRMessagingCheckinPreferences *preferences, NSError *error) {
-        FIRMessaging_STRONGIFY(self);
-        if (error) {
-          newHandler(nil, error);
-          return;
-        }
+  [_authService fetchCheckinInfoWithHandler:^(FIRMessagingCheckinPreferences *preferences,
+                                              NSError *error) {
+    FIRMessaging_STRONGIFY(self);
+    if (error) {
+      newHandler(nil, error);
+      return;
+    }
 
-        if (!self) {
-          NSError *noSelf = [NSError messagingErrorWithCode:kFIRMessagingErrorCodeUnknown
-                                              failureReason:@"Lost Reference to TokenManager"];
-          handler(nil, noSelf);
-          return;
-        }
+    if (!self) {
+      NSError *derefErr =
+          [NSError messagingErrorWithCode:kFIRMessagingErrorCodeUnknown
+                            failureReason:@"Unable to fetch token. Lost Reference to TokenManager"];
+      handler(nil, derefErr);
+      return;
+    }
 
-        FIRMessaging_WEAKIFY(self);
-        [self->_installations installationIDWithCompletion:^(NSString *_Nullable identifier,
-                                                             NSError *_Nullable error) {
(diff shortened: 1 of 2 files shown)


FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -247,6 +247,13 @@ - (void)tokenWithAuthorizedEntity:(NSString *)authorizedEntity
           return;
         }
 
+        if (!self) {
+          NSError *noSelf = [NSError messagingErrorWithCode:kFIRMessagingErrorCodeUnknown
```
charlotteliang, line 251: nit: maybe just error or referenceError

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
```diff
@@ -300,6 +307,14 @@ - (void)fetchNewTokenWithAuthorizedEntity:(NSString *)authorizedEntity
           handler(nil, error);
           return;
         }
+
+        if (!self) {
+          NSError *noSelf = [NSError messagingErrorWithCode:kFIRMessagingErrorCodeUnknown
```
charlotteliang, line 312: Same here, just for style naming convention

charlotteliang (approved):

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
charlotteliang: Also we should give a new error code than kFIRMessagingErrorCodeUnknown

aashishpatil-g: Fixed this - renamed to lostRefError.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
aashishpatil-g: Fixed this - renamed to lostRefError.

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
aashishpatil-g: @charlotteliang - 
I was torn about the error code for the following reasons -
- All existing (other) error codes we have are about Firebase (FCM) related errors. Self reference lost is not specific to Firebase and can occur in any kind of app. So didn't want to add a generic unrelated error code to Firebase set of error codes.
- We don't know at this place in code why the self reference was lost (i.e. why was TokenManager dealloced). 

Due to these reasons, I chose to leave error code as Unknown but provided a descriptive error message (failureReason) so developers can understand what is going on from the error message / logs.

paulb777 (approved):

charlotteliang (approved):

FirebaseMessaging/Sources/Token/FIRMessagingTokenManager.m
charlotteliang: you can create a error code for all self reference errors. I wonder if there's another error you can reuse (e.g. internal error) other than Unknown.

aashishpatil-g: There isn't an internal error code. I can create one - maybe it will be available for such scenarios in the future.

aashishpatil-g: Added an internal code instead of unknown.

PR #10744, opened 2023-01-27
Prevent capturing network info when Perf data collection off

samedson:
We're treating data collection off as SDK not installed

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -209,10 +209,8 @@ private enum GoogleDataTransportConfig {
   func addSubscriberFields(event: SessionStartEvent) {
     subscribers.forEach { subscriber in
       event.set(subscriber: subscriber.sessionsSubscriberName,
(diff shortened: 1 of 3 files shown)


visumickey:
> We're treating data collection off as SDK not installed
> 
> #no-changelog

I'm not sure if I understand the intent of this PR @samedson. Can you clarify it a bit more?

visumickey (approved):
My assumption on this PR is that it is abstracting setting restricted fields inside the event object rather than exposing it to outside. If so, this change LGTM.

samedson:
@visumickey 

The main change is before we didn't care if data collection was disabled.

 - **BEFORE**: We would collect NetworkInfo / MCCMNC if FirePerf was installed regardless of data collection
 - **NOW**: We only collect if Perf installed && Data Collection Enabled for Perf

```
    // Only set restricted fields if Data Collection is enabled. If it's disabled,
    // we're treating that as if the product isn't installed.
    if isDataCollectionEnabled {
      setRestrictedFields(subscriber: subscriber,
                          appInfo: appInfo)
    }
```

PR #10745, opened 2023-01-27
Reduce memory usage by applying query check sooner

wu-hui:
see: https://github.com/firebase/firebase-android-sdk/pull/4591


Firestore/core/src/local/leveldb_remote_document_cache.cc
```diff
@@ -171,13 +173,18 @@ MutableDocumentMap LevelDbRemoteDocumentCache::GetAll(
...
 MutableDocumentMap LevelDbRemoteDocumentCache::GetAllExisting(
-    DocumentVersionMap&& remote_map) const {
+    DocumentVersionMap&& remote_map,
+    const core::Query& query,
+    const model::OverlayByDocumentKeyMap& mutated_docs) const {
   BackgroundQueue tasks(executor_.get());
   AsyncResults<std::pair<DocumentKey, MutableDocument>> results;
   for (const auto& key_version : remote_map) {
-    tasks.Execute([this, &results, key_version] {
+    tasks.Execute([this, &results, key_version, query, mutated_docs] {
```
cherylEnkidu, line 182: can we pass `query` and `mutated_docs` by reference?

cherylEnkidu (changes requested):
Thank you for the improvement!

Firestore/core/src/local/leveldb_remote_document_cache.cc
```diff
@@ -171,13 +173,18 @@ MutableDocumentMap LevelDbRemoteDocumentCache::GetAll(
...
 MutableDocumentMap LevelDbRemoteDocumentCache::GetAllExisting(
-    DocumentVersionMap&& remote_map) const {
+    DocumentVersionMap&& remote_map,
+    const core::Query& query,
+    const model::OverlayByDocumentKeyMap& mutated_docs) const {
   BackgroundQueue tasks(executor_.get());
   AsyncResults<std::pair<DocumentKey, MutableDocument>> results;
   for (const auto& key_version : remote_map) {
-    tasks.Execute([this, &results, key_version] {
+    tasks.Execute([this, &results, key_version, &query, &mutated_docs] {
```
cherylEnkidu, line 182: Sorry I didn't find out earlier, could you please also make `key_version` pass by reference?

cherylEnkidu (approved):

cherylEnkidu:
Looks like `sanitizers` complains after merge the master branch

Firestore/core/src/local/leveldb_remote_document_cache.cc
wu-hui: I do not think query can be captured by reference here, somehow it is deleted before its usage here. This is what caused sanitizer failure.

cherylEnkidu: I don't understand why `query` would be deleted :(

PR #10756, opened 2023-01-31
Compare p13n metadata values as dictionaries.

danasilver:
Compare keyed P13n metadata using `isEqualToDictionary:`. P13n metadata are [stored in dictionaries](https://github.com/firebase/firebase-ios-sdk/blob/48b529dd1598242004948919273486f7416b4b98/FirebaseRemoteConfig/Sources/RCNConfigContent.m#L35-L39), ex. `{"param_key": NSDictionary, "param_key_2": NSDictionary}`.

Comparing as strings raises an exception: `'NSInvalidArgumentException', reason: '-[__NSDictionaryI isEqualToString:]: unrecognized selector sent to instance 0x600003f63f80'
terminating with uncaught exception of type NSException`

#no-changelog


karenyz (approved):
ohh good catch

PR #10760, opened 2023-01-31
Realtime RC QA Bug Fix: Checks for type of P13N keys

qdpham13:
Fixes bug uncovered by manual QA: https://buganizer.corp.google.com/issues/267284784
TLDR: App crashes when P13N param is deleted.

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigContent.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigContent.m
@@ -428,7 +428,8 @@ const NSTimeInterval kDatabaseLoadTimeoutSecs = 30.0;
 
   // add params with new/updated p13n metadata
   for (NSString *key in [fetchedP13n allKeys]) {
-    if (activeP13n[key] == nil || ![activeP13n[key] isEqualToString:fetchedP13n[key]]) {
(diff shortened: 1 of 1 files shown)


FirebaseRemoteConfig/Sources/RCNConfigContent.m
```diff
@@ -428,7 +428,8 @@ - (FIRRemoteConfigUpdate *)getConfigUpdateForNamespace:(NSString *)FIRNamespace
 
   // add params with new/updated p13n metadata
   for (NSString *key in [fetchedP13n allKeys]) {
-    if (activeP13n[key] == nil || ![activeP13n[key] isEqualToString:fetchedP13n[key]]) {
+    if (activeP13n[key] == nil || ([fetchedP13n[key] isKindOfClass:[NSString class]] &&
```
danasilver, line 431: I think the bug here is that P13n metadata is stored as a `NSDictionary` rather a string. Just merged in a fix for that https://github.com/firebase/firebase-ios-sdk/pull/10756

qdpham13: Nice!!

PR #10762, opened 2023-02-01
Remove previous_session_id in favor of first_session_id and session_index

samedson:
- Does not implement new fields yet

#no-changelog

PR #10764, opened 2023-02-01
Update SessionGenerator for first_session_id, session_index

samedson:
Follow up from https://github.com/firebase/firebase-ios-sdk/pull/10762

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/SessionGenerator.swift
+++ b/FirebaseSessions/Sources/SessionGenerator.swift
@@ -41,7 +41,7 @@ class SessionGenerator {
   private var thisSession: SessionInfo?
   private var settings: SettingsProtocol
 
-  private var firstSessionId: String?
+  private var firstSessionId = ""
   private var sessionIndex: Int32
 
   init(settings: SettingsProtocol) {
@@ -57,8 +57,7 @@ class SessionGenerator {
 
     // If firstSessionId is set, use it. Otherwise set it to the
     // first generated Session ID
-    let firstSessionId = self.firstSessionId ?? newSessionId
-    self.firstSessionId = firstSessionId
+    firstSessionId = firstSessionId.isEmpty ? newSessionId : firstSessionId
 
     sessionIndex += 1
 
--- a/FirebaseSessions/Sources/SessionGenerator.swift
+++ b/FirebaseSessions/Sources/SessionGenerator.swift
@@ -58,9 +58,9 @@ class SessionGenerator {
     // If firstSessionId is set, use it. Otherwise set it to the
     // first generated Session ID
     let firstSessionId = self.firstSessionId ?? newSessionId
+    self.firstSessionId = firstSessionId
 
     sessionIndex += 1
-    self.firstSessionId = firstSessionId
 
     var collectEvents = true
     let randomValue = Double.random(in: 0 ... 1)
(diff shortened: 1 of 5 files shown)


FirebaseSessions/Sources/SessionGenerator.swift
```diff
@@ -19,13 +19,15 @@ import Foundation
 
 struct SessionInfo {
   let sessionId: String
-  let previousSessionId: String?
+  let firstSessionId: String
   let shouldDispatchEvents: Bool
+  let sessionIndex: Int32
```
visumickey, line 24: Can we use native types like `Integer`?

FirebaseSessions/Sources/SessionGenerator.swift
```diff
@@ -39,25 +41,37 @@ class SessionGenerator {
...
 
   // Generates a new Session ID. If there was already a generated Session ID
   // from the last session during the app's lifecycle, it will also set the last Session ID
   func generateNewSession() -> SessionInfo {
     let newSessionId = UUID().uuidString.replacingOccurrences(of: "-", with: "").lowercased()
 
+    // If firstSessionId is set, use it. Otherwise set it to the
+    // first generated Session ID
+    let firstSessionId = self.firstSessionId ?? newSessionId
+    self.firstSessionId = firstSessionId
```
visumickey, line 61: Can we consolidate it to `self.firstSessionId = self.firstSessionId ?? newSessionId`. Is that an anti-pattern?

FirebaseSessions/Tests/Unit/SessionStartEventTests.swift
```diff
@@ -38,6 +38,24 @@ class SessionStartEventTests: XCTestCase {
...
+      sessionId: "test_session_id",
+      firstSessionId: "test_first_session_id",
+      dispatchEvents: true,
+      sessionIndex: 0
+    )
+  }
+
+  var thirdSessionInfo: SessionInfo {
+    return SessionInfo(
+      sessionId: "test_session_id",
```
visumickey, line 52: For completeness, lets use unique sessionId for each sessionInfo. Easy to ensure that we validate on the right session Ids.

visumickey (approved):

FirebaseSessions/Sources/SessionGenerator.swift
samedson: In Swift, `Int32` is a native type - I think `Integer` is just an interface. Using `Int32` makes it compatible with our proto which also uses int32

https://stackoverflow.com/questions/24053432/int-vs-integer-in-swift

Screenshot 2023-02-01 at 1 52 32 PM

FirebaseSessions/Sources/SessionGenerator.swift
samedson: Yeah I was trying to find a non-clunky way to write this. Since `self.firstSessionId` is a `String?`, I needed a way to define it as a `String` so I used a variable.

If I use `self.firstSessionId = self.firstSessionId ?? newSessionId`, then I need another `??` in the constructor for `SessionInfo`. I still think there's a better way to write this...

Screenshot 2023-02-01 at 1 55 02 PM

samedson: Maybe a ternary `self.firstSessionId = self.firstSessionId.isEmpty ? newSessionId : self.firstSessionId`, but I kindof hate ternarys

PR #10766, opened 2023-02-01
Fix typo in error message

mikehardy:
This PR is *almost* worthless but hey, I was in there and saw it and the github web PR ui is a pretty quick flow, so here's a typo fix

Yes, I ignored the template for PRs, apologies in advance, however if you simply click on "file changes" and check the diff that's faster than reading anything else I would type in here.

Cheers :-)


paulb777 (approved):
Thanks!

PR #10769, opened 2023-02-01
Update sampling to calculate per-run-of-app instead of per-session

samedson:
#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -70,7 +70,7 @@ private enum GoogleDataTransportConfig {
     )
 
     let sessionGenerator = SessionGenerator(collectEvents: Sessions
-      .calculateCollectEvents(settings: settings))
+      .shouldCollectEvents(settings: settings))
     let coordinator = SessionCoordinator(
       installations: installations,
       fireLogger: fireLogger
@@ -198,7 +198,7 @@ private enum GoogleDataTransportConfig {
 
   // MARK: - Sampling
 
-  static func calculateCollectEvents(settings: SettingsProtocol) -> Bool {
+  static func shouldCollectEvents(settings: SettingsProtocol) -> Bool {
     // Calculate whether we should sample events using settings data
     // Sampling rate of 1 means we do not sample.
     let randomValue = Double.random(in: 0 ... 1)
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -69,7 +69,7 @@ private enum GoogleDataTransportConfig {
       installations: installations
     )
 
-    let sessionGenerator = SessionGenerator(settings: settings)
+    let sessionGenerator = SessionGenerator(collectEvents: Sessions.calculateCollectEvents(settings: settings))
     let coordinator = SessionCoordinator(
       installations: installations,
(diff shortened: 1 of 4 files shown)


FirebaseSessions/Sources/SessionGenerator.swift
```diff
@@ -43,11 +43,23 @@ class SessionGenerator {
 
   private var firstSessionId = ""
   private var sessionIndex: Int32
+  private var collectEvents: Bool
 
-  init(settings: SettingsProtocol) {
+  init(settings: SettingsProtocol,
```
visumickey, line 48: Should this change be here or should this move to FirebaseSessions initializer? I'm thinking this logic of checking if sessions can be generated can be moved to [here](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/Sources/FirebaseSessions.swift#L145). This is before listening the lifecycle events.

The only catch is that we might have to update settings before the listening begins. If that is done, then I think that would be a much more ideal route. What do you think?

samedson: I like that idea from a simplicity perspective. My only problem with doing that is it's going to involve a pretty big refactor to the tests

samedson: Hmm actually there might be a simple way to do it

FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -195,6 +196,15 @@ private enum GoogleDataTransportConfig {
     }
   }
 
+  // MARK: - Sampling
+
+  static func calculateCollectEvents(settings: SettingsProtocol) -> Bool {
```
visumickey, line 201: Just a small nitpick: Rename this function to `shouldCollectEvents()`?

FirebaseSessions/Sources/SessionGenerator.swift
```diff
@@ -39,15 +39,16 @@ struct SessionInfo {
...
   private var sessionIndex: Int32
+  private var collectEvents: Bool
 
-  init(settings: SettingsProtocol) {
-    self.settings = settings
+  init(collectEvents: Bool) {
     // This will be incremented to 0 on the first generation
     sessionIndex = -1
+
+    self.collectEvents = collectEvents
```
visumickey, line 51: Where is this value used? I mean where do you check if collectEvents is true and so generate a new session or otherwise?

I'm also trying to validate the place where the config fetches happen irrespective of the state of data collection.

samedson: This is used in constructing the SessionInfo object below: https://github.com/firebase/firebase-ios-sdk/pull/10769/files#diff-55a6b65c4c6ce0a2d723a584879ca1d8bfb4f621e449d935a34e73e6462c6defR67

samedson: This is where config fetches happen: https://github.com/firebase/firebase-ios-sdk/blob/784c1012a22ac3a1e7eed0bbba652b8c3c6d3292/FirebaseSessions/Sources/FirebaseSessions.swift#L178

visumickey (approved):

FirebaseSessions/Sources/SessionGenerator.swift
visumickey: Sounds good. Thank you!

PR #10770, opened 2023-02-01
Realtime RC 1P Bug Fix: Add checks for retryHttpConnection method

qdpham13:
Fixes this bug: https://b.corp.google.com/issues/265943796. There was no checks for background/active connections for the retry method causing it to burn out it's retries when the app is in the background. Adding these checks should prevent this.


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -348,7 +348,11 @@ - (void)retryHTTPConnection {
   __weak RCNConfigRealtime *weakSelf = self;
   dispatch_async(_realtimeLockQueue, ^{
     __strong RCNConfigRealtime *strongSelf = weakSelf;
-    if (strongSelf->_remainingRetryCount > 0) {
```
karenyz, line 351: wdyt about adding `if (strongSelf->_isInBackground) { return; }` at the beginning instead?

karenyz (approved):

PR #10776, opened 2023-02-03
[auth-swift] Remove commented-out imports

ncooke3:
### Context
Remove unused imports (the imported types have been since re-defined in Swift)

#no-changelog


ncooke3:
Merging without review as this seems trivial.

PR #10777, opened 2023-02-03
Updated gtm session fetcher to allow for exponential backoff

maneesht:


paulb777 (commented):
If Storage is not using any new GTMSessionFetcher APIs, I'm not sure we should start requiring GTMSessionFetcher 3.1.   Anyone who wants the new capability can already get it by updating and those that have other dependencies requiring earlier GTM versions won't be impacted.

maneesht:
@paulb777 - the main reason for the upgrade is to fix situations when exponential backoff is required. For example, if there's an issue with an intermediary Google Cloud service, without this fix, we will just keep hitting the server infinitely, draining resources on people's machines.

maneesht:
Context: https://github.com/google/gtm-session-fetcher/pull/323/

paulb777:
There is nothing stopping developers from updating GTMSessonFetcher now to 3.1 if they can.

I'd prefer to avoid breaking interoperability with SDKs like [ARCore](https://github.com/CocoaPods/Specs/blob/master/Specs/4/e/a/ARCore/1.35.0/ARCore.podspec.json#L82) until a major release, when we have a process to engage with other SDK teams on dependency changes.

paulb777:
Marked for the next breaking change of Firebase 11 to update the minimum allowed version of GTMSessionFetcher.

paulb777:

PR #10778, opened 2023-02-04
[auth-swift] Remove `module.modulemap` as it is unneeded

ncooke3:
### Context
- A manual module map is unneeded. CocoaPods will generate one. It seems to work fine with the recently added private module map.
- This is nice because achieving interoperability that's on par with SwiftPM does not require any new build files–– just the build script.

#no-changelog

### DIFF
--- a/scripts/build_private_module_map.rb
+++ b/scripts/build_private_module_map.rb
@@ -75,7 +75,6 @@ def main(args)
     else
         # Overwrite the private module map with the generated contents.
         File.write(
-            # TODO(ncooke3): See if you can throw it in the public headers directory?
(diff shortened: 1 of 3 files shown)


ncooke3:
Manually verified that the pod linted successfully for iOS and macOS. These changes do not affect SwiftPM.

paulb777 (approved):

scripts/build_private_module_map.rb
```diff
@@ -75,7 +75,6 @@ def main(args)
     else
         # Overwrite the private module map with the generated contents.
         File.write(
-            # TODO(ncooke3): See if you can throw it in the public headers directory?
```
ncooke3, line 78: Removed TODO as this really isn't neccessary.

PR #10780, opened 2023-02-04
Added release note to update GTMSessionFetcher

maneesht:


FirebaseStorage/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
paulb777, line 1: There should be time to get this into next week's 10.5.0 release notes update so it can be moved into the section below. cc: @morganchen12

- [fixed] Run `pod update` or `File -> Packages -> Update to latest Packages` to update the `GTMSessionFetcher` dependency to at least version 3.1.0. This fixes an issue where it infinitely retries when FirebaseStorage returns a 500 response.

maneesht: Done.

FirebaseStorage/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # 10.5.0
 - [added] Added Storage API to limit upload chunk size. (#10137)
+- [fixed] Run pod update or File -> Packages -> Update to latest Packages to update the GTMSessionFetcher dependency to at least version 3.1.0. This fixes an issue where it infinitely retries when FirebaseStorage returns a 500 response.
```
charlotteliang, line 3: This should go to next release since code freeze day is due.

paulb777: @charlotteliang The change has already happened and this is a release notes only update.  I don't think the release notes generation has begun yet.

@maneesht Please add the backquotes back around the commands and add line breaks consistent with the rest of the file

maneesht: Done.

morganchen12 (approved):

PR #10781, opened 2023-02-04
Fix unit test failure

paulb777:
This fixes the one test failure we've had for awhile.

It seems that running the unit tests from Swift will fail with NSInternalInconsistencyException from https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseAuth/Sources/Auth/FIRAuth.m#L443 but not when run from Objective C.

This change fixes that problem by not calling `Auth.auth()` too early.

### DIFF
--- a/FirebaseAuth/Sources/Swift/Backend/IdentityToolkitRequest.swift
+++ b/FirebaseAuth/Sources/Swift/Backend/IdentityToolkitRequest.swift
@@ -54,30 +54,8 @@ private let kIdentityPlatformStagingAPIHost =
 
   let _useStaging: Bool
 
-  /** @fn initWithEndpoint:APIKey:
-   @brief Designated initializer.
-   @param endpoint The endpoint name.
-   @param requestConfiguration An object containing configurations to be added to the request.
-   */
-  @objc public init(endpoint: String, requestConfiguration: AuthRequestConfiguration) {
-    self.endpoint = endpoint
-    APIKey = requestConfiguration.APIKey
-    _requestConfiguration = requestConfiguration
-    _useIdentityPlatform = false
-    _useStaging = false
-
-    // Automatically set the tenant ID. If the request is initialized before FIRAuth is configured,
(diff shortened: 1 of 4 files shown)


FirebaseAuth/Sources/Swift/Backend/IdentityToolkitRequest.swift
```diff
@@ -67,9 +68,14 @@ private let kIdentityPlatformStagingAPIHost =
 
     // Automatically set the tenant ID. If the request is initialized before FIRAuth is configured,
     // set tenant ID to nil.
-    tenantID = Auth.auth().tenantID
+    if FirebaseApp.app() == nil {
```
ncooke3: Do you know why the request would be initialized before Auth in the first place?

FirebaseAuth/Sources/Swift/Backend/VerifyClientRequest.swift
```diff
@@ -41,8 +41,8 @@ public class VerifyClientRequest: IdentityToolkitRequest, AuthRPCRequest {
   @objc public private(set) var isSandbox: Bool
 
   // TODO(ncooke3): This API should be unneccessary...
-  @objc override public init(endpoint: String,
-                             requestConfiguration: AuthRequestConfiguration) {
+  @objc public init(endpoint: String,
+                    requestConfiguration: AuthRequestConfiguration) {
```
ncooke3, line 45: Wrt my comment above, is the idea to gradually remove these initializers in favor of a more complete one?

ncooke3 (approved):

FirebaseAuth/Sources/Swift/Backend/VerifyClientRequest.swift
paulb777: Removed now. Thanks for pointing that out.

ncooke3 (approved):

Issue #10742, opened 2023-01-27
[FirebaseMessaging] FirebaseApp.configure() triggers "APNS device token not set before retrieving FCM Token"


DanielReyesDev:
### Description

I noticed that everytime an app executes the method `FirebaseApp.configure()`, the SDK throws the following output in the console:
```
APNS device token not set before retrieving FCM Token for Sender ID '804880211453'.Be sure to re-retrieve the FCM token once the APNS device token is set.
```

- is this a valid warning message? I mean, how can we event setup the APNS device token before the whole SDK initialization?

- I noticed this behavior is also happening in older SDK version like 9.4.0

- Additionally I noticed that automatically the SDK is making an API call to `https://fcmtoken.googleapis.com/register` without the `apns` token. The exact same call that was sending a 400 when this behavior happened last week before the rollback here: https://github.com/firebase/firebase-ios-sdk/issues/10679#issuecomment-1396293848

- how can we avoid that `/register` api call at the very beginning of the SDK initialization?

### Reproducing the issue

1. Download and configure the official quick-start ios project: https://github.com/firebase/quickstart-ios
2. Run the app "MessagingExampleSwift" on a real iPhone
3. You will see this error in console right after FirebaseApp.configure() is executed:

```
APNS device token not set before retrieving FCM Token for Sender ID '676015592495'. Notifications to this FCM Token will not be delivered over APNS.Be sure to re-retrieve the FCM token once the APNS device token is set.
```

### Firebase SDK Version

10.4.0

### Xcode Version

13.4

### Installation Method

Swift Package Manager

### Firebase Product(s)

Messaging

### Targeted Platforms

iOS

### Relevant Log Output

```shell
2023-01-26 18:22:15.695691-0600 MessagingExampleSwift[1179:196158] [SceneConfiguration] Info.plist contained no UIScene configuration dictionary (looking for configuration named "(no name)")
2023-01-26 18:22:15.713204-0600 MessagingExampleSwift[1179:196158] [Assert] UINavigationBar decoded as unlocked for UINavigationController, or navigationBar delegate set up incorrectly. Inconsistent configuration may cause problems. navigationController=<UINavigationController: 0x10200ea00>, navigationBar=<UINavigationBar: 0x100c07830; frame = (0 0; 0 50); opaque = NO; autoresize = W; gestureRecognizers = <NSArray: 0x283c52400>; layer = <CALayer: 0x28324aa60>> delegate=0x10200ea00
[...]

Issue #10679, opened 2023-01-13
Messaging.messaging().token returns errors


RobinCaroff:
### Description

It has been two days since that last Messaging.messaging().token success with our app. 

It currently returns the following error :

```
Error Domain=com.google.fcm Code=0 "Invalid fetch response, expected 'token' or 'Error' key" UserInfo={NSLocalizedFailureReason=Invalid fetch response, expected 'token' or 'Error' key}
```

Here is how we retrieve the token :

```
self.getMessaging().token { token, error in
              if let error = error {
                Logger.shared.log(.app, .error, "Error fetching FCM registration token: \(error)")
                observer(.failure(error))
              } else if let token = token {
                Logger.shared.log(.app, .debug, "FCM registration token: \(token)")
                observer(.success(token))
              }
            }
            return Disposables.create()
```

Note 1: the live app has not been updated since 2 weeks and we reproduce the same issue with oldest version.

Note 2: We have a different target using the same code with a different bundle id which does not have the issue. 

### Firebase SDK Version

10.3.0

### Xcode Version

14.1

### Installation Method

CocoaPods

### Firebase Product(s)

Messaging

### Targeted Platforms

iOS

### Relevant Log Output

```shell
Error Domain=com.google.fcm Code=0 "Invalid fetch response, expected 'token' or 'Error' key" UserInfo={NSLocalizedFailureReason=Invalid fetch response, expected 'token' or 'Error' key}
```

### If using Swift Package Manager, the project's Package.resolved

<details>
<summary>Expand <code>Package.resolved</code> snippet</summary>
<br>

```json

Replace this line with the contents of your Package.resolved.

```

</details>

### If using CocoaPods, the project's Podfile.lock

<details>
<summary>Expand <code>Podfile.lock</code> snippet</summary>
<br>

```yml

PODS:
  - Alamofire (5.6.4)
  - Differentiator (5.0.0)
  - Firebase/AnalyticsWithoutAdIdSupport (10.3.0):
    - Firebase/CoreOnly
    - FirebaseAnalytics/WithoutAdIdSupport (~> 10.3.0)
  - Firebase/CoreOnly (10.3.0):
    - FirebaseCore (= 10.3.0)
  - Firebase/Crashlytics (10.3.0):
    - Firebase/CoreOnly
    - FirebaseCrashlytics (~> 10.3.0)
  - Firebase/Messaging (10.3.0):
    - Firebase/CoreOnly
[...]

PR #10789, opened 2023-02-07
[Messaging] - Validate APNS token in the configure flow

aashishpatil-g:
### Discussion

The FirebaseApp.configure() flow triggers a FCM token fetch, which can be racy and occur before we have received and set an APNS token.  The change in this PR validates that we have an APNS token in the configure flow before fetching the FCM token. 

This PR addresses the scenario 1 identified in this comment - https://github.com/firebase/firebase-ios-sdk/issues/10679#issuecomment-1402776795

Should fix https://github.com/firebase/firebase-ios-sdk/issues/10742

### DIFF
--- a/FirebaseMessaging/Sources/FIRMessaging.m
+++ b/FirebaseMessaging/Sources/FIRMessaging.m
@@ -239,11 +239,13 @@ BOOL FIRMessagingIsContextManagerMessage(NSDictionary *message) {
                                                        scope:kFIRMessagingDefaultTokenScope]
           .token;
   // When there is a cached token, do the token refresh.
+  // Before fetching FCM token, confirm there is an APNS token to avoid validation error
+  // This error is innocuous since we refetch after APNS token is set but it can seem alarming
   if (cachedToken) {
     // Clean up expired tokens by checking the token refresh policy.
     [self.installations installationIDWithCompletion:^(NSString *_Nullable identifier,
                                                        NSError *_Nullable error) {
-      if ([self.tokenManager checkTokenRefreshPolicyWithIID:identifier]) {
+      if ([self.tokenManager checkTokenRefreshPolicyWithIID:identifier] && self.APNSToken) {
         // Default token is expired, fetch default token from server.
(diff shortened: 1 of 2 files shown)


FirebaseMessaging/Sources/FIRMessaging.m
```diff
@@ -239,11 +239,13 @@ - (void)didCompleteConfigure {
...
           .token;
   // When there is a cached token, do the token refresh.
+  // Before fetching FCM token, confirm there is an APNS token to avoid validation error
+  // This error is innocuous since we refetch after APNS token is set but it can seem alarming
   if (cachedToken) {
     // Clean up expired tokens by checking the token refresh policy.
     [self.installations installationIDWithCompletion:^(NSString *_Nullable identifier,
                                                        NSError *_Nullable error) {
-      if ([self.tokenManager checkTokenRefreshPolicyWithIID:identifier]) {
+      if ([self.tokenManager checkTokenRefreshPolicyWithIID:identifier] && self.APNSToken) {
```
charlotteliang, line 248: I wonder if we should guard the check in a lower level where always check APNS token is valid before sending a get FCM token request

charlotteliang (approved):
On a higher level, should we also think about only trigger fcm token request when messaging apns token is set (I believe that should happen every time app starts) rather than trigger at configure startup time.

FirebaseMessaging/Sources/FIRMessaging.m
aashishpatil-g: Thats a good point - the lower check already exists in the TokenManager. However, that returns an Error. The error is not harmful and we refetch once APNS token is set but the Error causes an alarm for developers and I think triggers Error reports. 

This check is so we don't begin fetch it in the configure flow itself - slightly higher level than the TokenManager. This will prevent false errors.

aashishpatil-g: Also, we already have a flow that refetches once an APNS token is set. This is to prevent a default configure flow we have.

paulb777 (approved):

PR #10790, opened 2023-02-07
Fix Session Event printout for sampling, log_environment

samedson:
Before:

```
Session Event:
...
      session_sampling_rate: 0.0
...
      log_environment: _firebase_appquality_sessions_LogEnvironment(rawValue: 3)
```

After:

```
Session Event:
...
      session_sampling_rate: 1.0
...
      log_environment: PROD
```

#no-changelog

### DIFF
--- a/FirebaseSessions/Sources/FirebaseSessions.swift
+++ b/FirebaseSessions/Sources/FirebaseSessions.swift
@@ -181,6 +181,7 @@ private enum GoogleDataTransportConfig {
         self.settings.updateSettings()
 
         self.addSubscriberFields(event: event)
+        event.setSamplingRate(samplingRate: self.settings.samplingRate)
 
         guard sessionInfo.shouldDispatchEvents else {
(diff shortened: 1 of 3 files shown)


FirebaseSessions/Sources/FirebaseSessions.swift
```diff
@@ -181,6 +181,7 @@ private enum GoogleDataTransportConfig {
         self.settings.updateSettings()
 
         self.addSubscriberFields(event: event)
+        event.setSamplingRate(samplingRate: self.settings.samplingRate)
```
visumickey, line 184: Did we miss this field in the event earlier? Should I add this field to the PLX alert?

visumickey (approved):
Good catch and thanks for fixing it.

FirebaseSessions/Sources/FirebaseSessions.swift
samedson: Yeah we somehow made this method but never called it - worth adding to plx alerts

PR #10791, opened 2023-02-07
Add Sessions to release manifest

paulb777:
Stage and release Sessions for 10.6.0 and beyond

Note, the zip release build will fail until Sessions is actually staged.


samedson (approved):

paulb777:
Merged too fast. Looks like there's a zip head build issue. I'm investigating and will send a followup PR

PR #10811, opened 2023-02-13
Turn on Firebase Sessions Test App while SDK disabled

samedson:
Disabled in https://github.com/firebase/firebase-ios-sdk/pull/10796, need the testapp to continue working


visumickey (approved):
Changes look good to me. Should we have to do the same for the other non-iOS apps? like MacOS/Catalyst/WatchOS targets? Do they all use the the Info.plist?

samedson:
Great question @visumickey - tv/watch use this one it seems, but macos needed a different one. Added it

PR #10816, opened 2023-02-14
Three more RPCs to Swift

paulb777:
Swift implementations and unit tests for three RPCs - GetAccountInfo, SetAccountInfo, GetProjectConfig

Continue to defer (ifdef out) tests that rely on mocks.

Also add TODO's for @mortenbekditlevsen's review comments in #10806


mortenbekditlevsen:
Awesome work! I love the net change in number of lines! 😊

PR #10825, opened 2023-02-15
[auth-swift] Re-enable logging code (pt. 2)

ncooke3:
Follow-up to #10817 

#no-changelog

### DIFF
--- a/FirebaseCore/Extension/FIRHeartbeatLogger.h
+++ b/FirebaseCore/Extension/FIRHeartbeatLogger.h
@@ -17,6 +17,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 #ifndef FIREBASE_BUILD_CMAKE
+//@import FirebaseCoreInternal;
 @class FIRHeartbeatsPayload;
 #endif  // FIREBASE_BUILD_CMAKE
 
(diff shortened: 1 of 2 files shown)


FirebaseCore/Extension/FIRHeartbeatLogger.h
```diff
@@ -17,7 +17,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 #ifndef FIREBASE_BUILD_CMAKE
-@class FIRHeartbeatsPayload;
+@import FirebaseCoreInternal;
```
ncooke3, line 20: The reason the API used in the above file wasn't previously available to Swift was that the API signatures referenced forward declared types. This doesn't work well (hence, this [proposal](https://forums.swift.org/t/pitch-importing-forward-declared-objective-c-classes-and-protocols/61926)), and the way around it is to import the type's module instead of forward declaring it. Then, when the header is parsed for it's Swift-compatible API, all API shows in Swift.

ncooke3: The unfortunate part is that `FirebaseCoreExtension` now exports `FirebaseCoreInternal`, but that seems unavoidable.

paulb777: FirebaseCoreInternal is already a dependency of FirebaseCore, so in practice should have no impact.

paulb777 (approved):

PR #10819, opened 2023-02-14
WIP: auth-swift Converted a few more files

mortenbekditlevsen:
Hi @paulb777 ,
I converted a few more classes last night. I didn’t yet get around to copying over comments and copyright notices and won’t get around to that the next few days since I’m going on vacation. 
I thought I’d still open the wip PR to prevent duplicate work. I’ll clean it up in the beginning of next week. 😊

PR #10829, opened 2023-02-16
[auth-swift] Test improvements

paulb777:
- Build SwiftAPI Tests from CocoaPods
- Fix (Undo) breaking API changes discovered in iOS specific API tests
- API weirdness include `IDToken` and `UIDelegate` parameters that have historically been converted to `idToken` and `uiDelegate` for Swift (noted in spreadsheet)
- Update tests using `MultiFactorAssertion` and `MultiFactorInfo` that were relying on default initializers that should have been unavailable in Objective C (noted in spreadsheet)
- Restore disabled tests from #10819 in FIRAuthSerialTaskQueueTests.m that run for me in both iOS and macOS cc: @mortenbekditlevsen 
- Fixed decoder build warning in AdditionalUserInfo.swift

### DIFF
--- a/FirebaseAuth.podspec
+++ b/FirebaseAuth.podspec
@@ -83,7 +83,10 @@ supports email and password accounts, as well as several 3rd party authenticatio
         :osx => osx_deployment_target,
         :tvos => tvos_deployment_target
       }
-      unit_tests.source_files = 'FirebaseAuth/Tests/Unit/*.[mh]'
+      unit_tests.source_files = [
+        'FirebaseAuth/Tests/Unit/*.[mh]',
+        'FirebaseAuth/Tests/Unit/*.swift'
+      ]
       unit_tests.osx.exclude_files = [
         'FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenManagerTests.m',
         'FirebaseAuth/Tests/Unit/FIRAuthAPNSTokenTests.m',
--- a/FirebaseAuth.podspec
+++ b/FirebaseAuth.podspec
@@ -83,10 +83,7 @@ supports email and password accounts, as well as several 3rd party authenticatio
         :osx => osx_deployment_target,
         :tvos => tvos_deployment_target
       }
-      unit_tests.source_files = [
-        'FirebaseAuth/Tests/Unit/*.[mh]',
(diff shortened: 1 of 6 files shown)


FirebaseAuth.podspec
```diff
@@ -83,7 +83,10 @@ supports email and password accounts, as well as several 3rd party authenticatio
         :osx => osx_deployment_target,
         :tvos => tvos_deployment_target
       }
-      unit_tests.source_files = 'FirebaseAuth/Tests/Unit/*.[mh]'
+      unit_tests.source_files = [
+        'FirebaseAuth/Tests/Unit/*.[mh]',
+        'FirebaseAuth/Tests/Unit/*.swift'
+      ]
```
ncooke3, line 89: optional:
```suggestion
      unit_tests.source_files = 'FirebaseAuth/Tests/Unit/*.{m,h,swift}'
```

FirebaseAuth/Tests/Unit/SwiftAPI.swift
```diff
@@ -262,23 +262,23 @@ class AuthAPI_hOnlyTests: XCTestCase {
   #if !os(macOS)
     func FIRFedederatedAuthProvider_h() {
       class FederatedAuthImplementation: NSObject, FederatedAuthProvider {
+        @available(iOS 13, tvOS 13, macOS 10.15, macCatalyst 13, watchOS 7, *)
```
ncooke3, line 265: Is this file's diff reflective of the API surface in `master` or has there been a divergence?

ncooke3 (approved):
Nice!

FirebaseAuth/Tests/Unit/SwiftAPI.swift
paulb777: This specific change is because with the Swift implementation, the async version of the API needs to be also implemented. I'll add another note to the spreadsheet to track the open question of how to deal with public protocols with previously auto-generated async apis.

There are a few other divergences because auth-swift ran these tests on mac for the first time and also, Swift is strict about assignments when there is not a `@discardableResult`

ncooke3: Understood-- thanks 👍

PR #10844, opened 2023-02-17
Return Server produced error if Http status code is 403 

qdpham13:
https://buganizer.corp.google.com/issues/268398841
Pass back server error if it is the 403 API enablement response. didReceiveResponse only has statusCode response while didReceiveData has the actual error message. So in order to know if the error message is the API enablement one, I check if the URL in is the data response.

### DIFF
--- a/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
+++ b/FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
@@ -36,10 +36,6 @@ static NSString *const kServerURLKey = @"key=";
 static NSString *const kEnableRealtimeURL =
     @"https://console.developers.google.com/apis/api/firebaseremoteconfigrealtime.googleapis.com/"
     @"overview?project=";
-static NSString *const kApiDisabledErrorMessage =
-    @"The Remote Config Realtime API is disabled. It must be enabled in the Google Cloud Console. "
-    @"If you enabled this API recently, wait a few minutes for the action to propagate to our "
-    @"systems and retry. You can enable the API by following this URL: %@";
 
 /// Header names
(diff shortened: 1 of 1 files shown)


FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -585,7 +602,9 @@ - (void)URLSession:(NSURLSession *)session
                  }];
       FIRLogError(kFIRLoggerRemoteConfig, @"I-RCN000021", @"Cannot establish connection. Error: %@",
                   error);
-      [self propogateErrors:error];
+      if (statusCode != 403) {
```
karenyz, line 605: is there no other info in the response that indicates what kind of error it is? feels fairly brittle to be checking for a literal string.. 
i might prefer just returning the full message for all 403s, since that will likely require user intervention to be fixed

qdpham13: The status code and response body are processed in separate delegate methods so we can't just return all 403s. However all server errors have ' code: <HTTP_CODE> ' in the response so I think we can match on the string 
' code:403 ' instead. Updated.

FirebaseRemoteConfig/Sources/RCNConfigRealtime.m
```diff
@@ -532,6 +535,18 @@ - (void)URLSession:(NSURLSession *)session
     didReceiveData:(NSData *)data {
   NSError *dataError;
   NSString *strData = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+
+  /// If response data contains the API enablement link, return the entire message to the user in
```
karenyz, line 539: nit: update comment since we're no longer checking for enablement link

karenyz (approved):

PR #10853, opened 2023-02-21
Analytics 10.6.0

htcgh:
Analytics 10.6.0 #no-changelog

Note: No source changes to GoogleAppMeasurementOnDeviceConversion

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -150,7 +150,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Note that CI changes the version to the head of main for CI.
       // See scripts/setup_spm_tests.sh.
-      .exact("10.4.0")
+      .exact("10.6.0")
     ),
     .package(
       name: "GoogleDataTransport",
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -150,7 +150,7 @@ let package = Package(
       url: "https://github.com/google/GoogleAppMeasurement.git",
       // Note that CI changes the version to the head of main for CI.
       // See scripts/setup_spm_tests.sh.
-      .exact("10.4.0")
+      .exact("10.6.0")
```
andrewheard, line 153: I _think_ you'll need to add a tag named `10.6.0` in the [GoogleAppMeasurement](https://github.com/google/GoogleAppMeasurement.git) repo before we merge this change (so that Swift Package Manager can find it).

andrewheard (changes requested):
Looks good other than the tag on GoogleAppMeasurement. Thanks!

Package.swift
paulb777: That doesn't happen until release day. Until then, CI builds the `main` branch

andrewheard: Thanks for the clarification that it's known to break non-CI builds targeting master.

Linking to the issue / possible workaround you shared: https://github.com/google/GoogleAppMeasurement/issues/24#issuecomment-1097109812

andrewheard (approved):

WChoy:
You changes the dependency for GoogleAppMeasurement to a non-existent version
The bad version code in https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift line 153

Current stable version is 10.1.0
Prior versions include: 10.0.0 and 9.6.0

see https://github.com/google/GoogleAppMeasurement/releases

PS: I would like a Sparky Plushie for my desk
```

   .package(
      name: "GoogleAppMeasurement",
      .exact("10.6.0")
    ),
```

Package.swift
WChoy: Getting the error in the firebase-ios-sdk:
 
Failed to resolve dependencies Dependencies could not be resolved because no versions of 'googleappmeasurement' match the requirement 10.6.0 and 'firebase-ios-sdk' depends on 'googleappmeasurement' 10.6.0.

The public doesn't have access to  https://dl.google.com/firebase/ios/analytics/93d139ae9f688e9c/GoogleAppMeasurement-10.6.0.tar.gz

paulb777: That's expected behavior. 10.6.0 has not yet released.

PR #10854, opened 2023-02-22
Firestore: Add CalculateMd5Digest() function

dconeybe:
This PR adds the C++ function `CalculateMd5Digest()` to the internal "util" directory. This is _not_ part of the public API; it is just an implementation detail. The implementation of the md5 algorithm was copied from Chromium.


Firestore/core/src/util/md5_commoncrypto.cc
```diff
@@ -0,0 +1,68 @@
...
+#if FIRESTORE_MD5_IMPL_COMMONCRYPTO
+
+// NOLINTNEXTLINE(build/include)
+#include <CommonCrypto/CommonDigest.h>
+
+namespace firebase {
+namespace firestore {
+namespace util {
+
+// Suppress deprecation warnings because due to CommonCrypto's md5 functions.
```
dconeybe, line 32: nit: Remove extraneous word: "because"

Firestore/core/src/util/md5_openssl.cc
```diff
@@ -0,0 +1,63 @@
...
+
+  {
+    std::array<unsigned char, 16> digest;
+    int md5_final_result = MD5_Final(digest.data(), &ctx);
+    HARD_ASSERT(md5_final_result == 1, "MD5_Final() returned %s but expected 1",
+                md5_final_result);
+    return digest;
+  }
+}
+
```
dconeybe, line 58: nit: Add `SUPPRESS_END()`

Firestore/core/src/util/md5_commoncrypto.cc
```diff
@@ -0,0 +1,68 @@
...
+
+  {
+    int md5_update_result = CC_MD5_Update(&ctx, s.data(), s.length());
+    HARD_ASSERT(md5_update_result == 1,
+                "CC_MD5_Update() returned %s, but expected 1",
+                md5_update_result);
+  }
+
+  {
+    std::array<unsigned char, 16> digest;
```
dconeybe, line 55: nit: Add a `static_assert` that `CC_MD5_DIGEST_LENGTH==16`

Firestore/core/src/util/md5_openssl.cc
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  {
+    int md5_update_result = MD5_Update(&ctx, s.data(), s.length());
+    HARD_ASSERT(md5_update_result == 1,
+                "MD5_Update() returned %s, but expected 1", md5_update_result);
+  }
+
+  {
+    std::array<unsigned char, 16> digest;
```
dconeybe, line 51: nit: Add a `static_assert` that `MD5_DIGEST_LENGTH==16`

Firestore/core/test/unit/util/md5_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  std::array<unsigned char, 16> result;
+  for (int i = 0; i < 16; ++i) {
+    unsigned char c1 = UnsignedCharFromHexDigit(s[i * 2]);
+    unsigned char c2 = UnsignedCharFromHexDigit(s[(i * 2) + 1]);
+    result[i] = (c1 << 4) | c2;
+  }
+  return result;
+}
+
+std::array<unsigned char, 16> GetPreComputedMd5Digest(const std::string& s) {
```
milaGGL, line 140: Can we add special characters test case as well: https://github.com/firebase/firebase-js-sdk/blob/97ca3ee1580ca8b90c3293da2652eccf02709030/packages/firestore/test/unit/core/webchannel_wrapper.test.ts#L62

milaGGL: But GetStringWithAllPossibleCharacters should have covered special characters here. 🤔

dconeybe: Great idea. But in C++ strings are just "byte strings" that know nothing about Unicode code points/code units and surrogate characters and stuff like that. Since MD5 only processes bytes (not characters) there probably isn't much value in adding tests for UTF-8 encoding of Unicode characters.

wu-hui (approved):
Thanks for making the change!

milaGGL (approved):
Looks great!

PR #10860, opened 2023-02-23
Add Swift unit tests and utilities for AppDistributionApiService

tejasd:
Add Swift unit tests and utils for adding swift tests for networking.

#no-changelog


FirebaseAppDistributionInternal/Sources/AppDistributionApiService.swift
```diff
@@ -86,17 +104,29 @@ enum AppDistributionApiError: NSInteger {
   }
 
   @objc(fetchReleasesWithCompletion:) public static func fetchReleases(completion: @escaping AppDistributionFetchReleasesCompletion) {
+    fetchReleases(
+      app: FirebaseApp.app()!,
```
tagboola, line 108: Should we use a guard here instead of a forced unwrapping?

FirebaseAppDistributionInternal/Sources/AppDistributionApiService.swift
```diff
@@ -105,14 +135,14 @@ enum AppDistributionApiError: NSInteger {
         format: "Url: %@ Auth token: %@ Api Key: %@",
         urlString,
         authTokenResult?.authToken ?? "",
-        FirebaseApp.app()?.options.apiKey ?? "unknown"
+        app.options.apiKey!
```
tagboola, line 138: Should we use a guard here instead of a forced unwrapping?

FirebaseAppDistributionInternal/Tests/Unit/AppDistributionApiServiceTests.swift
```diff
@@ -13,6 +13,156 @@
...
+    let expectation = XCTestExpectation(description: "Fetch releases succeeds with two releases.")
+
+    AppDistributionApiService.fetchReleases(
+      app: app,
+      installations: installations,
+      urlSession: urlSession,
+      completion: { releases, error in
+        XCTAssertNotNil(releases)
+        XCTAssertNil(error)
+        XCTAssertEqual(releases?.count, 2)
```
tagboola, line 112: Should we also assert the value of the releases variable?

tagboola (approved):

FirebaseAppDistributionInternal/Tests/Unit/AppDistributionApiServiceTests.swift
tejasd: Added

FirebaseAppDistributionInternal/Sources/AppDistributionApiService.swift
tejasd: Done

PR #10874, opened 2023-02-28
Add golden tests for BloomFilter

milaGGL:
#no-changelog
- [x] Import the util::CalculateMd5Digest
- [x] Add Golden tests


Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
 
+class BloomFilterGoldenTest : public ::testing::Test {
+ public:
+  void RunGoldenTest(std::string& test_file) {
+    size_t start_pos = test_file.find("bloom_filter_proto");
+    if (start_pos == std::string::npos) {
+      return;
+    }
+    std::string result_file = test_file;
+    result_file.replace(start_pos, sizeof("bloom_filter_proto") - 1,
```
dconeybe, line 153: Don't use `sizeof` here. Although technically it works, it's not idiomatic C++ and won't work for other types.

Instead, create a local `const std::string xxx = "bloom_filter_proto"` and use `xxx.size()` instead. Also, use this new variable in the call to `test_file.find()` above.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+    }
+  }
+
+ private:
+  std::string GOLDEN_DOCUMENT_PREFIX_ =
+      "projects/project-1/databases/database-1/documents/coll/doc";
+  Path GOLDEN_TEST_FOLDER_ = (Path::FromUtf8(__FILE__).Dirname())
+                                 .AppendUtf8("bloom_filter_golden_test_data/");
+
+  json ReadFile(std::string& file_name) {
```
dconeybe, line 186: nit `std::string&` -> `const std::string&`

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+      bool expectedResult = membership_result[i] == '1';
+      bool mightContainResult = bloom_filter.MightContain(
+          GOLDEN_DOCUMENT_PREFIX_ + std::to_string(i));
+
+      EXPECT_EQ(mightContainResult, expectedResult);
+    }
+  }
+
+ private:
+  std::string GOLDEN_DOCUMENT_PREFIX_ =
```
dconeybe, line 181: The `GOLDEN_DOCUMENT_PREFIX_` and `GOLDEN_TEST_FOLDER_` member variables should be declared `const` and `static` since they are the same for every instance of the class. Also, the naming convention would have their names be `kGoldenDocumentPrefix` and `kGoldenTestFolder`, respectively.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+  Path GOLDEN_TEST_FOLDER_ = (Path::FromUtf8(__FILE__).Dirname())
+                                 .AppendUtf8("bloom_filter_golden_test_data/");
+
+  json ReadFile(std::string& file_name) {
+    Path file_path = GOLDEN_TEST_FOLDER_.AppendUtf8(file_name);
+    std::ifstream stream(file_path.native_value());
+    return nlohmann::json::parse(stream);
+  }
+
+  JsonReader reader;
```
dconeybe, line 192: Would it make sense for this `JsonReader` member variable to instead be a local variable in the `RunGoldenTest()` method?

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+  std::string GOLDEN_DOCUMENT_PREFIX_ =
+      "projects/project-1/databases/database-1/documents/coll/doc";
+  Path GOLDEN_TEST_FOLDER_ = (Path::FromUtf8(__FILE__).Dirname())
+                                 .AppendUtf8("bloom_filter_golden_test_data/");
+
+  json ReadFile(std::string& file_name) {
+    Path file_path = GOLDEN_TEST_FOLDER_.AppendUtf8(file_name);
+    std::ifstream stream(file_path.native_value());
+    return nlohmann::json::parse(stream);
+  }
```
dconeybe, line 190: Consider adding `HARD_ASSERT(stream.good())` before returning the parsed json, just to make sure that there weren't any errors reading from the stream.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+ private:
+  std::string GOLDEN_DOCUMENT_PREFIX_ =
+      "projects/project-1/databases/database-1/documents/coll/doc";
+  Path GOLDEN_TEST_FOLDER_ = (Path::FromUtf8(__FILE__).Dirname())
+                                 .AppendUtf8("bloom_filter_golden_test_data/");
+
+  json ReadFile(std::string& file_name) {
+    Path file_path = GOLDEN_TEST_FOLDER_.AppendUtf8(file_name);
+    std::ifstream stream(file_path.native_value());
+    return nlohmann::json::parse(stream);
```
dconeybe, line 189: How does `nlohmann::json::parse()` behave if parsing fails?

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+    json result_file_json = ReadFile(result_file);
+
+    nlohmann::json bits = reader.OptionalObject("bits", test_file_json, {});
+    std::string bitmap = reader.OptionalString("bitmap", bits, "");
+    int padding = reader.OptionalInt("padding", bits, 0);
+    int hash_count = reader.OptionalInt("hashCount", test_file_json, 0);
+    std::string decoded;
+    absl::Base64Unescape(bitmap, &decoded);
+    std::vector<uint8_t> decoded_map(decoded.begin(), decoded.end());
+    BloomFilter bloom_filter = BloomFilter(decoded_map, padding, hash_count);
```
dconeybe, line 166: nit: `BloomFilter bloom_filter = BloomFilter(...)` -> `BloomFilter bloom_filter(...)`

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+ * <p>Full document path is generated by concatenating documentPrefix and number
+ * n, eg, projects/project-1/databases/database-1/documents/coll/doc12.
+ *
+ * <p>The test result is generated by checking the membership of documents from
+ * documentPrefix+0 to documentPrefix+2n. The membership results from 0 to n is
+ * expected to be true, and the membership results from n to 2n is expected to
+ * be false with some false positive results.
+ */
+TEST_F(BloomFilterGoldenTest, GoldenTest1Document1FalsePositiveRate) {
+  std::string test_file =
```
dconeybe, line 208: Remove the `test_file` local variable and just specify the string literal directly to the `RunGoldenTest()` method.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+      "Validation_BloomFilterTest_MD5_1_1_bloom_filter_proto.json";
+  RunGoldenTest(test_file);
+}
+TEST_F(BloomFilterGoldenTest, GoldenTest1Document01FalsePositiveRate) {
```
dconeybe, line 212: nit: Please put a blank line between each test function.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +142,131 @@ TEST(BloomFilterTest,
...
+    std::string bitmap = reader.OptionalString("bitmap", bits, "");
+    int padding = reader.OptionalInt("padding", bits, 0);
+    int hash_count = reader.OptionalInt("hashCount", test_file_json, 0);
+    std::string decoded;
+    absl::Base64Unescape(bitmap, &decoded);
+    std::vector<uint8_t> decoded_map(decoded.begin(), decoded.end());
+    BloomFilter bloom_filter = BloomFilter(decoded_map, padding, hash_count);
+
+    std::string membership_result =
+        reader.OptionalString("membershipTestResults", result_file_json, "");
```
dconeybe, line 169: Consider using a default value of `"[invalid]"` or something like that, then HARD_ASSERT that the `membership_result` string is not equal to that special string. I'm nervous that if there is a typo in the field name that `membership_result` would just be empty and the test would succeed despite not actually testing anything.

dconeybe (changes requested):

Firestore/core/test/unit/remote/bloom_filter_test.cc
milaGGL: Does't allow me to do "static const". is it because I am not doing this in header file?
Screenshot 2023-02-28 at 3 35 21 PM

Firestore/core/test/unit/remote/bloom_filter_test.cc
milaGGL: I am thinking instantiate one reader, instead of repeating the creation for each RunGoldenTest method, but there is not big difference as we only run the method once in each test, and only the RunGoldenTest using the reader. I will move it.

Firestore/core/test/unit/remote/bloom_filter_test.cc
milaGGL: It will throw an exception: https://json.nlohmann.me/features/parsing/parse_exceptions/ 
I will add a try catch block here.

Firestore/core/test/unit/remote/bloom_filter_test.cc
milaGGL: if the member type is build in types like int, char, its okay to do "static" tho. 🤔

Firestore/core/test/unit/remote/bloom_filter_test.cc
dconeybe: No need to add a try/catch block. That just adds noise to the code. Just let the exception propagate and fail the test. I just wanted to make sure that it wasn't silently ignored.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -144,54 +145,60 @@ TEST(BloomFilterUnitTest,
 
 class BloomFilterGoldenTest : public ::testing::Test {
  public:
-  void RunGoldenTest(std::string& test_file) {
-    size_t start_pos = test_file.find("bloom_filter_proto");
-    if (start_pos == std::string::npos) {
-      return;
-    }
+  void RunGoldenTest(const std::string& test_file) {
+    const std::string substring = "bloom_filter_proto";
```
dconeybe, line 149: I think it would help readability if the logic to calculate the result file from the test file were moved into a helper method. Could you do that and see if it improves readability in your opinion? Note that the HARD_ASSERT can be part of that helper method. Also, the helper method can take the test file string _by value_ (as opposed to _by reference_), mutate it, and return it. Then you don't need to explicitly make a copy.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -144,54 +145,60 @@ TEST(BloomFilterUnitTest,
...
     std::string decoded;
     absl::Base64Unescape(bitmap, &decoded);
     std::vector<uint8_t> decoded_map(decoded.begin(), decoded.end());
-    BloomFilter bloom_filter = BloomFilter(decoded_map, padding, hash_count);
+    BloomFilter bloom_filter(decoded_map, padding, hash_count);
 
-    std::string membership_result =
-        reader.OptionalString("membershipTestResults", result_file_json, "");
+    std::string membership_result = reader.OptionalString(
+        "membershipTestResults", result_file_json, "[invalid]");
```
dconeybe, line 173: Add a HARD_ASSERT that `membership_result` is not equal to `[invalid]`.

Firestore/core/test/unit/remote/bloom_filter_test.cc
dconeybe: Discussed offline.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,131 @@ TEST(BloomFilterTest,
...
+
+      EXPECT_EQ(mightContainResult, expectedResult);
+    }
+  }
+
+ private:
+  static const char* kGoldenDocumentPrefix;
+  
+  static Path GetGoldenTestFolder() {
+    return (Path::FromUtf8(__FILE__).Dirname())
```
dconeybe, line 187: nit: Is it possible to remove the parentheses around `Path::FromUtf8(__FILE__).Dirname()`

dconeybe (changes requested):

Firestore/core/test/unit/remote/bloom_filter_test.cc
milaGGL: Re-arranged the code in RunGoldenTest with helper functions.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -15,23 +15,34 @@
...
+#include "Firestore/core/src/util/path.h"
+#include "absl/strings/escaping.h"
 #include "gtest/gtest.h"
 
 namespace firebase {
 namespace firestore {
 namespace remote {
-namespace {
 
+namespace {
```
dconeybe, line 33: ultranit: add a blank line after `namespace {`

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+      bool mightContainResult =
+          bloom_filter.MightContain(kGoldenDocumentPrefix + std::to_string(i));
+
+      EXPECT_EQ(mightContainResult, expectedResult);
+    }
+  }
+
+ private:
+  static const char* kGoldenDocumentPrefix;
+  JsonReader reader;
```
dconeybe, line 163: Please make `JsonReader reader` a local variable in the method(s) that use it, rather than a member variable. We want to avoid any possible "bleeding" of information between method invocations and the performance gain of re-using an existing instance is negligible.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
   }
 }
 
+class BloomFilterGoldenTest : public ::testing::Test {
+ public:
+  void RunGoldenTest(const std::string& test_file) {
```
dconeybe, line 148: It looks like `RunGoldenTest()` can be declared `static` (after moving `JsonReader reader` to a local variable, as recommended below). Please make this change, if possible.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+  }
+
+  json ReadFile(const std::string& file_name) {
+    Path file_path = GetGoldenTestFolder().AppendUtf8(file_name);
+    std::ifstream stream(file_path.native_value());
+    HARD_ASSERT(stream.good());
+    return nlohmann::json::parse(stream);
+  }
+
+  BloomFilter GetBloomFilter(const std::string& file_name) {
```
dconeybe, line 177: nit: Rename `GetBloomFilter()` to `ReadBloomFilter()` or `LoadBloomFilter()` or something that more clearly implies that it is loading the bloom filter from an external resource.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+    nlohmann::json bits = reader.OptionalObject("bits", test_file, {});
+    std::string bitmap = reader.OptionalString("bitmap", bits, "");
+    int padding = reader.OptionalInt("padding", bits, 0);
+    int hash_count = reader.OptionalInt("hashCount", test_file, 0);
+    std::string decoded;
+    absl::Base64Unescape(bitmap, &decoded);
+    std::vector<uint8_t> decoded_map(decoded.begin(), decoded.end());
+
+    StatusOr<BloomFilter> maybe_bloom_filter =
+        BloomFilter::Create(std::move(decoded_map), padding, hash_count);
```
dconeybe, line 188: Nice use of std::move ;)

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+    HARD_ASSERT(maybe_bloom_filter.ok(),
```
dconeybe, line 189: Include the error message from `maybe_bloom_filter` in the assertion failure message. This will greatly assist with debugging.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+                "Bloom filter input file %s has invalid values.", file_name);
+    BloomFilter bloom_filter = maybe_bloom_filter.ValueOrDie();
```
dconeybe, line 191: Avoid assigning the return value of `maybe_bloom_filter.ValueOrDie()` to a local variable since the only purpose of that local variable is to immediately return it. Instead, just `return maybe_bloom_filter.ValueOrDie()`.

Firestore/core/test/unit/remote/bloom_filter_test.cc
```diff
@@ -132,6 +143,153 @@ TEST(BloomFilterTest,
...
+
+    return bloom_filter;
+  }
+
+  std::string LocateResultFile(std::string file_name) {
+    const std::string substring = "bloom_filter_proto";
+    size_t start_pos = file_name.find(substring);
+    HARD_ASSERT(start_pos != std::string::npos,
+                "Test file name %s is not valid, expected to include "
+                "\"bloom_filter_proto\".",
```
dconeybe, line 201: Don't copy the string "bloom_filter_proto" here. Instead, use "%s" in the message and `substring` as an argument to the HARD_ASSERT.

dconeybe (changes requested):

dconeybe (approved):

PR #10899, opened 2023-03-06
[auth-swift] Backend now completely in Swift

paulb777:
- AuthBackendRPCImplentationTests now in Swift
- All ObjC backend code gone
- Next step is to implement disabled mock tests in AuthTests and UserTests


paulb777:
@ncooke3 This includes two heartbeat request tests to Swift

PR #10903, opened 2023-03-07
[auth-swift] Swift AuthDispatcher, two fakes, one test

paulb777:
- AuthDispatcher in Swift
- One AuthTest to Swift
- Two new fakes to replace mocking

Note, the new test fails in SPM - likely because of the known keychain issues with Auth SPM testing.


paulb777:
Rebased here to get #10906

PR #10911, opened 2023-03-08
Refactor localstore.ShouldPersistTargetData

wu-hui:


Firestore/core/src/local/local_store.cc
```diff
@@ -465,6 +475,10 @@ void LocalStore::NotifyLocalViewChanges(
             target_data.WithLastLimboFreeSnapshotVersion(
                 last_limbo_free_snapshot_version);
         target_data_by_target_[target_id] = updated_target_data;
+
+        if (ShouldPersistTargetData(updated_target_data, target_data, {})) {
+          target_cache_->UpdateTarget(updated_target_data);
+        }
```
ehsannas, line 481: isn't `updated_target_data` taking its last limbo-free snapshot version from `target_data`, and hence the difference between them will always be zero?

Firestore/core/src/local/local_store.cc
```diff
@@ -417,6 +417,16 @@ bool LocalStore::ShouldPersistTargetData(const TargetData& new_target_data,
...
 
+  // Update the target cache if sufficient time has passed since the last
+  // LastLimboFreeSnapshotVersion
+  int64_t new_limbo_free_seconds =
+      new_target_data.last_limbo_free_snapshot_version().timestamp().seconds();
+  int64_t old_limbo_free_seconds =
+      old_target_data.last_limbo_free_snapshot_version().timestamp().seconds();
+  int64_t limbo_free_time_delta =
+      new_limbo_free_seconds - old_limbo_free_seconds;
+  if (limbo_free_time_delta >= kResumeTokenMaxAgeSeconds) return true;
```
ehsannas, line 428: this sounds reasonable. For Android, this logic is placed inside `NotifyLocalViewChanges`. Arguably it makes more sense for the logic to be inside `ShouldPersistTargetData`.

ehsannas (approved):

PR #10937, opened 2023-03-10
Export *just* the AppDistributionInternal source code to master to fix CI issues

tejasd:
There are CI failures which seem to need this source code in master.

Eg. https://github.com/firebase/firebase-ios-sdk/actions/runs/4387975314/jobs/7683918159

This PR is an attempt at that.

#no-changelog


paulb777:
I believe two other alternatives to this would be pushing AppDistribution-internal to SpecsDev or temporarily disabling the failing CI job in your branch.

tejasd:
Pushing it SpecsDev using `pod repo push https://github.com/firebase/SpecsDev FirebaseAppDistributionInternal.podspec` errors out.

Validating spec
 -> FirebaseAppDistributionInternal (10.5.0-beta)
    - ERROR | [iOS] file patterns: The `source_files` pattern did not match any file.
    - ERROR | [iOS] file patterns: The `resource_bundles` pattern for `AppDistributionInternalResources` did not match any file.
    - ERROR | [iOS] unknown: Encountered an unknown error (Unable to install the `FirebaseAppDistributionInternal` pod, because the `FirebaseAppDistributionInternal-Unit-unit` target in Xcode would have no sources to compile.) during validation.

It's possible to temporarily disable the CI tasks - but this seems like a harmless change (it won't be included in `spm`, it's not added as a dependency to `FirebaseAppDistribution`.

The only thing I'm not sure if anything is needed, is to prevent it from being pushed to `https://cdn.cocoapods.org/` from master?

FirebaseAppDistributionInternal.podspec
```diff
@@ -0,0 +1,64 @@
...
+    SDK for Swift code in Firebase App Distribution.
+                         DESC
+
+    s.homepage         = 'https://firebase.google.com'
+    s.license          = { :type => 'Apache-2.0', :file => 'LICENSE' }
+    s.authors          = 'Google, Inc.'
+
+    s.source           = {
+      :git => 'https://github.com/firebase/firebase-ios-sdk.git',
+      :tag => 'CocoaPods-' + s.version.to_s
```
paulb777, line 22: This line would need to change to successfully push to SpecsDev

paulb777 (approved):
I see.  You'd need to change the podspec to successfully push to SpecsDev.

I agree it should be safe to do it this other way. We won't try to publish until it gets added to FirebaseManifest.swift

FirebaseAppDistributionInternal.podspec
tejasd: Would it make sense to avoid pushing the podspec to master then? And just push it to SpecsDev?

paulb777: I'm fine either way. Note that "push" means two different things. For master it's a git push. For Specs Dev, it's a CocoaPods publish.

tejasd: Makes sense! Reverted the `podspec` from master for now. I'll do a repo push to SpecsDev to fix the issue!

PR #10944, opened 2023-03-13
Fix ApiService

tagboola:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/FirebaseAppDistributionInternal/Sources/ApiService.swift
+++ b/FirebaseAppDistributionInternal/Sources/ApiService.swift
@@ -206,7 +206,7 @@ struct CreateFeedbackReportRequest: Codable {
       // we'll have to figure out a way to get the project number
       let urlString = String(
         format: Strings.findReleaseEndpointUrlTemplate,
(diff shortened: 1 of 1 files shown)


FirebaseAppDistributionInternal/Sources/ApiService.swift
```diff
@@ -206,7 +206,7 @@ struct CreateFeedbackReportRequest: Codable {
       // we'll have to figure out a way to get the project number
       let urlString = String(
         format: Strings.findReleaseEndpointUrlTemplate,
-        app,
```
tejasd, line 209: If we don't want to pass the appId, I think this change should be in the Strings enum?

tejasd (approved):

FirebaseAppDistributionInternal/Sources/ApiService.swift
tagboola: Removing this change. The dash won't actually work. Will fix this in a follow up CL.

PR #10946, opened 2023-03-13
Add missing readtime as resume type

wu-hui:


Firestore/core/src/remote/serializer.cc
dconeybe, line 1: Is there a unit test somewhere to verify this behavior?

dconeybe (changes requested):

PR #10953, opened 2023-03-14
Apply bloom filter on existence filter mismatch

milaGGL:
#no-changelog


Firestore/core/src/remote/datastore.cc
```diff
@@ -108,6 +108,7 @@ Datastore::Datastore(
       connectivity_monitor_{connectivity_monitor},
       grpc_connection_{database_info, worker_queue, &grpc_queue_,
                        connectivity_monitor_, firebase_metadata_provider},
+      database_info_{database_info},
```
milaGGL, line 111: save the parameter database_info as class attribute to read datastore.databaseId out of it later.

Firestore/core/src/remote/datastore.h
```diff
@@ -212,6 +216,7 @@ class Datastore : public std::enable_shared_from_this<Datastore> {
   GrpcConnection grpc_connection_;
 
   std::vector<std::unique_ptr<GrpcCall>> active_calls_;
+  core::DatabaseInfo database_info_;
```
milaGGL, line 219: save database_info as class attribute to read datastore.databaseId out of it later.

Firestore/core/src/remote/remote_event.cc
```diff
@@ -235,15 +237,68 @@ void WatchChangeAggregator::HandleExistenceFilter(
...
     }
   }
 }
 
+/** Returns whether a bloom filter removed the deleted documents successfully.
+ */
+bool WatchChangeAggregator::ApplyBloomFilter(
+    ExistenceFilterWatchChange existence_filter, int current_count) {
+  int expected_count = existence_filter.filter().count();
+  absl::optional<BloomFilter> bloom_filter =
```
milaGGL, line 261: should I move this? This should be the only place that actually uses bloom_filter in existence_filter, so stealing it should not create a problem, unless, in the future, we have decided to re-run bloom filter application or use it somewhere else.

Firestore/core/test/unit/remote/remote_event_test.cc
```diff
@@ -564,6 +563,108 @@ TEST_F(RemoteEventTest, ExistenceFilterMismatchClearsTarget) {
...
+                              DocumentKeySet{}};
+  ASSERT_TRUE(event.target_changes().at(1) == target_change1);
+
+  TargetChange target_change2{resume_token1_, false, DocumentKeySet{},
+                              DocumentKeySet{}, DocumentKeySet{}};
+  ASSERT_TRUE(event.target_changes().at(2) == target_change2);
+
+  // The given BloomFilter will return false on MightContain(doc1) and true on
+  // MightContain(doc2).
+  ExistenceFilterWatchChange change4{
```
milaGGL, line 601: this bloom filter proto is generated using the helper in web SDK, and copy paste the result here. Should i create a helper in iOS as well or this is okay?

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -330,9 +330,26 @@ - (SnapshotVersion)parseVersion:(NSNumber *_Nullable)version {
 }
 
 - (absl::optional<BloomFilter>)parseBloomFilter:(NSDictionary *_Nullable)bloomFilterProto {
```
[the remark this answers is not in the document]
dconeybe: Objective-C uses camelCase, like Java. C++ uses snake_case. This file is Objective-C++, which is why you see a mixture of the two. I'd just recommend to use whatever convention is used by the surrounding code.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -330,10 +330,29 @@ - (SnapshotVersion)parseVersion:(NSNumber *_Nullable)version {
...
+
+  // If not specified in proto, default padding and hashCount to 0.
+  int32_t padding = [bitsData[@"padding"] intValue];
+  int32_t hashCount = [bloomFilterProto[@"hashCount"] intValue];
+  StatusOr<BloomFilter> maybeBloomFilter = BloomFilter::Create(bitmap, padding, hashCount);
+
+  if (maybeBloomFilter.ok()) {
+    return std::move(maybeBloomFilter).ValueOrDie();
+  } else {
+    LOG_WARN("Parsing BloomFilterProto failed: %s", maybeBloomFilter.status().error_message());
```
dconeybe, line 353: If parsing the bloom filter fails, this should fail the test. Can you modify this code so that the test that calls this method fails? Maybe call XCTFail() and see if that works?

Firestore/core/src/remote/serializer.cc
```diff
@@ -1436,23 +1437,29 @@ std::unique_ptr<WatchChange> Serializer::DecodeExistenceFilterWatchChange(
...
-  // Create bloom filter if there is an unchanged_names present in the filter
-  // and inputs are valid, otherwise keep it null.
-  absl::optional<BloomFilter> bloom_filter;
   if (filter.has_unchanged_names) {
-    // TODO(Mila): None of the ported spec tests here actually has the bloom
-    // filter json string, so hard code an empty bloom filter for now. The
-    // actual parsing code will be written in the next PR, where we can validate
-    // the parsing result.
-    // TODO(Mila): handle bloom filter creation failure.
+    pb_bytes_array_t* bitmap_ptr = filter.unchanged_names.has_bits
```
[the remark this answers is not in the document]
dconeybe: Yes, you do need to handle the case where `filter.unchanged_names.has_bits== false`: if it is false then use an empty array instead. The way the code is currently written will crash if `filter.unchanged_names.has_bits== false` because it sets `bitmap_ptr` to `nullptr` and then immediately after `bitmap_ptr->bytes` and `bitmap_ptr->size` are accessed on that null pointer.

I'd recommend something like this:

```cpp
std::vector<uint8_t> bitmap;
if (bitmap_ptr) {
  bitmap = std::vector<uint8_t>(bitmap_ptr->bytes, bitmap_ptr->bytes + bitmap_ptr->size);
}
```

Firestore/core/test/unit/remote/serializer_test.cc
```diff
@@ -1751,6 +1750,25 @@ TEST_F(SerializerTest, DecodesListenResponseWithExistenceFilter) {
...
+  proto.mutable_filter()->set_count(555);
+  proto.mutable_filter()->set_target_id(999);
+
+  v1::BloomFilter* bloom_filter =
+      proto.mutable_filter()->mutable_unchanged_names();
+  bloom_filter->set_hash_count(33);
+  bloom_filter->mutable_bits()->set_padding(7);
+  bloom_filter->mutable_bits()->set_bitmap("\x42\xFE");
+
+  SCOPED_TRACE("DecodesListenResponseWithExistenceFilter");
```
dconeybe, line 1768: nit: Fix the `SCOPED_TRACE` to match the test name.

Firestore/core/test/unit/remote/watch_change_test.cc
```diff
@@ -40,12 +40,20 @@ TEST(WatchChangeTest, CanCreateDocumentWatchChange) {
...
-// TODO(Mila): Add test coverage for when the bloom filter is not null
 TEST(WatchChangeTest, CanCreateExistenceFilterWatchChange) {
-  ExistenceFilter filter{7, /*bloom_filter=*/absl::nullopt};
-  ExistenceFilterWatchChange change{filter, 5};
-  EXPECT_EQ(change.filter().count(), 7);
-  EXPECT_EQ(change.target_id(), 5);
+  {
+    ExistenceFilter filter{7, /*bloom_filter=*/absl::nullopt};
+    ExistenceFilterWatchChange change{filter, 5};
+    EXPECT_EQ(change.filter().count(), 7);
```
dconeybe, line 47: Can you add an "EXPECT" on `change.filter().bloom_filter()` here as well?

dconeybe (changes requested):
Looks fantastic! Mostly minor comments.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
milaGGL: Some of our spec tests have invalid bloom filter proto, to test the flow "fail to create a BloomFilter -> trigger full re-query". So some of the bloom filter creation are doomed to fail, and the test should still carry on.

Firestore/core/test/unit/remote/remote_event_test.cc
milaGGL: TODO: find a way to set databaseId here

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
dconeybe: Discussed offline. The bloom filter parsing logic will be moved to later in the stack, for parity with the web and android sdks and also so that the spec tests can exercise the logic that parses the bloom filter from the proto.

Firestore/core/src/remote/existence_filter.h
```diff
@@ -18,36 +18,52 @@
...
+struct BloomFilterParameter {
+  std::vector<uint8_t> bitmap;
+  int32_t padding;
+  int32_t hash_count;
+};
+
+inline bool operator==(const BloomFilterParameter& lhs,
+                       const BloomFilterParameter& rhs) {
+  return lhs.padding == rhs.padding && lhs.hash_count == rhs.hash_count &&
+         std::equal(lhs.bitmap.begin(), lhs.bitmap.end(), rhs.bitmap.begin());
```
dconeybe, line 38: nit: Just use `lhs.bitmap == rhs.bitmap` instead of `std::equal`.

Firestore/core/src/remote/existence_filter.h
```diff
@@ -18,36 +18,52 @@
...
 #include <utility>
+#include <vector>
 
 #include "Firestore/core/src/remote/bloom_filter.h"
 
 namespace firebase {
 namespace firestore {
 namespace remote {
 
+struct BloomFilterParameter {
```
dconeybe, line 29: nit: Rename `BloomFilterParameter` to `BloomFilterParameters` (i.e. make it plural) because it contains 3 parameters.

Firestore/core/src/remote/remote_event.cc
```diff
@@ -254,22 +255,32 @@ void WatchChangeAggregator::HandleExistenceFilter(
...
+                          bloom_filter_parameter.value().hash_count);
+  if (!maybe_bloom_filter.ok()) {
+    LOG_WARN("Creating BloomFilter failed: %s",
+             maybe_bloom_filter.status().error_message());
     return false;
   }
 
   int removed_document_count = FilterRemovedDocuments(
-      bloom_filter.value(), existence_filter.target_id());
+      std::move(maybe_bloom_filter).ValueOrDie(), existence_filter.target_id());
```
dconeybe, line 275: Remove the `std::move()` here because `FilterRemovedDocuments()` takes a const reference argument.

Firestore/core/src/remote/serializer.cc
```diff
@@ -1451,29 +1451,21 @@ std::unique_ptr<WatchChange> Serializer::DecodeExistenceFilterWatchChange(
...
+  }
+
+  int32_t hash_count = filter.unchanged_names.hash_count;
+  int32_t padding = 0;
+  std::vector<uint8_t> bitmap;
+  if (filter.unchanged_names.has_bits) {
+    padding = filter.unchanged_names.bits.padding;
+
+    pb_bytes_array_t* bitmap_ptr = filter.unchanged_names.bits.bitmap;
+    bitmap = std::vector<uint8_t>(bitmap_ptr->bytes,
```
dconeybe, line 1465: Verify that `bitmap_ptr` and `bitmap_ptr->bytes` are not `nullptr` before accessing them. Otherwise, if they are null, then the program will crash.

Firestore/core/src/remote/serializer.cc
```diff
@@ -1451,29 +1451,21 @@ std::unique_ptr<WatchChange> Serializer::DecodeExistenceFilterWatchChange(
...
+                                  bitmap_ptr->bytes + bitmap_ptr->size);
   }
-  return {filter.count, absl::nullopt};
+  return {filter.count, BloomFilterParameter{bitmap, padding, hash_count}};
```
dconeybe, line 1468: nit: `std::move()` the `bitmap` here to avoid the copy.

Firestore/core/test/unit/remote/fake_target_metadata_provider.h
```diff
@@ -75,9 +75,18 @@ class FakeTargetMetadataProvider : public TargetMetadataProvider {
       model::TargetId target_id) const override;
   model::DatabaseId GetDatabaseId() const override;
 
+  /**
+   * Sets the database_id to a custom value, used for getting Document's full
+   * path.
+   */
+  void SetDatabaseId(model::DatabaseId database_id) {
+    database_id_ = database_id;
```
dconeybe, line 83: nit: `database_id` -> `std::move(database_id)`

Firestore/core/test/unit/remote/fake_target_metadata_provider.h
```diff
@@ -75,9 +75,18 @@ class FakeTargetMetadataProvider : public TargetMetadataProvider {
...
+  }
+
  private:
   std::unordered_map<model::TargetId, model::DocumentKeySet> synced_keys_;
   std::unordered_map<model::TargetId, local::TargetData> target_data_;
+  model::DatabaseId database_id_ = model::DatabaseId("p", "d");
```
dconeybe, line 89: Please use the full values instead of "p" and "d". I think they were previously "test-project" and "test-database", or something like that.

Firestore/core/test/unit/remote/remote_event_test.cc
```diff
@@ -105,6 +105,10 @@ class RemoteEventTest : public testing::Test {
       DocumentKeySet existing_keys,
       const std::vector<std::unique_ptr<WatchChange>>& watch_changes);
 
+  void OverrideDefaultDatabaseId(model::DatabaseId database_id) {
+    target_metadata_provider_.SetDatabaseId(database_id);
```
dconeybe, line 109: nit: std::move() the database_id here.

Firestore/core/test/unit/remote/remote_event_test.cc
```diff
@@ -577,6 +581,7 @@ TEST_F(RemoteEventTest, ExistenceFilterMismatchWithBloomFilterSuccess) {
       target_map, no_outstanding_responses_,
       DocumentKeySet{doc1.key(), doc2.key()},
       Changes(std::move(change1), std::move(change2), std::move(change3)));
+  OverrideDefaultDatabaseId(model::DatabaseId("test-project", "test-database"));
```
dconeybe, line 584: Please add an inline comment above this line to document _why_ you are explicitly specifying the database ID.

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -345,14 +347,7 @@ - (SnapshotVersion)parseVersion:(NSNumber *_Nullable)version {
...
   int32_t hashCount = [bloomFilterProto[@"hashCount"] intValue];
-  StatusOr<BloomFilter> maybeBloomFilter = BloomFilter::Create(bitmap, padding, hashCount);
-
-  if (maybeBloomFilter.ok()) {
-    return std::move(maybeBloomFilter).ValueOrDie();
-  } else {
-    LOG_WARN("Parsing BloomFilterProto failed: %s", maybeBloomFilter.status().error_message());
-    return absl::nullopt;
-  }
+  return BloomFilterParameter{bitmap, padding, hashCount};
```
dconeybe, line 350: std::move(bitmap) here

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -682,6 +678,13 @@ - (void)doStep:(NSDictionary *)step {
   }
 }
 
+auto sortDocumentViewChange = [](std::vector<DocumentViewChange> &changes) {
```
dconeybe, line 681: Oh this works too. Can you just make this a "normal" function and put it into an anonymous namespace?

dconeybe (changes requested):

Firestore/core/test/unit/remote/fake_target_metadata_provider.h
milaGGL: setting `"test-project", "(default)"` as default value for testing

Firestore/core/src/remote/serializer.cc
milaGGL: c++ suggested `bitmap_ptr->bytes !=nullptr` is always true.

dconeybe (changes requested):

dconeybe (approved):

PR #10954, opened 2023-03-14
Validate QueryPurpose in spec test

milaGGL:
Ported from Web: https://github.com/firebase/firebase-js-sdk/pull/7113
Android: https://github.com/firebase/firebase-android-sdk/pull/4770

### DIFF
--- a/Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
+++ b/Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
@@ -769,10 +769,13 @@ NSString *ToTargetIdListString(const ActiveTargetMap &map) {
             std::vector<TargetData> queries;
             for (id queryJson in queriesJson) {
               Query query = [self parseQuery:queryJson];
-              // TODO(mcg): populate the purpose of the target once it's possible to encode that in
-              // the spec tests. For now, hard-code that it's a listen despite the fact that it's
-              // not always the right value.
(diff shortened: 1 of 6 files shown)


Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
```diff
@@ -887,11 +890,13 @@ - (void)validateActiveTargets {
     XCTAssertNotEqual(found, actualTargets.end(), @"Expected active target not found: %s",
                       targetData.ToString().c_str());
 
-    // TODO(mcg): validate the purpose of the target once it's possible to encode that in the
-    // spec tests. For now, only validate properties that can be validated.
+    // TODO(Mila): Replace the XCTAssertEqual() checks on the individual properties of TargetData
```
dconeybe, line 893: nit: Remove your name from this "TODO" comment. By putting your name there it implies that you are actively keeping track of this work and will do it when the time comes. That is _not_ the case, so just leave it as an "unassigned" TODO to record that _somebody_ should do this in the future.

dconeybe (approved):
Approved, with one very minor comment about a "TODO" comment. Thanks for doing this!

Firestore/Example/Tests/SpecTests/FSTSpecTests.mm
milaGGL: `// TODO:...` gives error messages like this: 
Screenshot 2023-03-16 at 9 22 12 AM

PR #10958, opened 2023-03-15
Fixing threading-related hang during initialization in urgent mode

wowlocal:
This commit fixes a threading issue in the FIRCLSInstallIdentifierModel class that caused a 10-second hang during Crashlytics initialization in urgent mode. The issue was caused by a semaphore that was waiting for a completion handler to unblock it, but the completion handler was waiting for the main thread to call dispatch_async, resulting in a deadlock. To solve this problem, the code was updated to use a runloop spinning mechanism instead of the semaphore when running on the main thread. This allows dispatch_async to notify the thread to resume execution without blocking it.

Changes include:
- Introducing a check for the main thread before using a semaphore
- Using a run loop to handle main thread execution
- Adding a test case for report preparation on the main thread
- Adding a GoogleService-Info.plist file for support initialization of the FIRApp.defaultApp in the unit tests

### DIFF
--- a/Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
+++ b/Crashlytics/Crashlytics/Models/FIRCLSInstallIdentifierModel.m
@@ -101,8 +101,14 @@ static unsigned long long FIRCLSInstallationsWaitTime = 10 * NSEC_PER_SEC;
 - (BOOL)regenerateInstallIDIfNeededWithBlock:(void (^)(NSString *fiid))block {
   BOOL __block didRotate = false;
 
-  dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+  dispatch_semaphore_t semaphore = nil;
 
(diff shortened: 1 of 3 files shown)


wowlocal:
> Hey @wowlocal, I made a change to this PR in a new one: https://github.com/firebase/firebase-ios-sdk/pull/11216/files. I put the rationale in the description.
> 
> I am for some reason unable to run your new test using Swift Package Manager. Are you able to run it?

The resourcePath behaves differently when using SPM.
I have identified and tested a solution for this issue based on similar cases. [firebase-ios-sdk.patch](https://github.com/firebase/firebase-ios-sdk/files/11377485/firebase-ios-sdk.patch)

wowlocal:
Fixed in https://github.com/firebase/firebase-ios-sdk/pull/11216

samedson:
Closing in favor of https://github.com/firebase/firebase-ios-sdk/pull/11216

PR #10972, opened 2023-03-20
Add missing readtime

wu-hui:


wu-hui:
@dconeybe This is same as https://github.com/firebase/firebase-ios-sdk/pull/10946 

I added a test, but somehow force pushed then closed the PR, I cannot reopen that one apparently.

Firestore/core/test/unit/remote/serializer_test.cc
```diff
@@ -1124,6 +1124,42 @@ TEST_F(SerializerTest, EncodesFirstLevelKeyQueries) {
...
+                                           model::SnapshotVersion::None()),
+                    proto);
+  }
+
+  {
+    SCOPED_TRACE("EncodesTargetDataWithResumeByReadTime");
+    v1::Target proto;
+    proto.mutable_documents()->add_documents(ResourceName("docs/1"));
+    proto.set_target_id(1);
+    proto.mutable_read_time()->set_seconds(1000);
```
dconeybe, line 1154: nit: Use a different value for `seconds` and `nanos` to make sure that the round trip doesn't accidentally invert them.

dconeybe (approved):
Thanks for adding the test! LGTM. Just one minor comment about a small improvement to the test.

Firestore/core/test/unit/remote/serializer_test.cc
wu-hui: Done.

PR #10975, opened 2023-03-20
Refactor feedback view controller to be scroll view + refactor release name

tejasd:
#no-changelog

### DIFF
--- a/FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
+++ b/FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
@@ -293,19 +293,21 @@ SFAuthenticationSession *_safariAuthenticationVC;
 - (void)screenshotDetected:(NSNotification *)notification {
   dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0ul), ^{
     [FIRFADInAppFeedback getManuallyCapturedScreenshotWithCompletion:^(UIImage *screenshot) {
-      dispatch_async(dispatch_get_main_queue(), ^{
         if ([[FIRAppDistribution appDistribution] isTesterSignedIn]) {
-          [self startFeedbackWithAdditionalFormText:self.additionalFormText image:screenshot];
+          dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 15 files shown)


FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -230,14 +228,17 @@ - (void)startFeedbackWithAdditionalFormText:(NSString *)additionalFormText
                                                                 // UIService to Swift.
                                                                 [self resetUIState];
                                                               }];
-  [self initializeUIState];
-  [self.hostingViewController presentViewController:feedbackViewController
-                                           animated:YES
-                                         completion:nil];
+
+  // TODO: Potentially move the check for the release version here.
+  if (feedbackViewController != nil) {
```
tagboola, line 233: I think at this point we don't want to show the feedback view controller if the release hasn't been identified. I _think_ this would mirror the behavior in Android.

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
```diff
@@ -294,9 +294,17 @@ - (void)screenshotDetected:(NSNotification *)notification {
...
-        [self startFeedbackWithAdditionalFormText:self.additionalFormText
-                                      releaseName:self.releaseName
-                                            image:screenshot];
+        if ([[FIRAppDistribution appDistribution] isTesterSignedIn]) {
+          [self startFeedbackWithAdditionalFormText:self.additionalFormText image:screenshot];
+        } else {
+          [[FIRAppDistribution appDistribution]
+              signInTesterWithCompletion:^(NSError *_Nullable error) {
+                if (error) {
+                  return;
```
tagboola, line 303: I see that you also try and identify the release here. Is there a scenario (other than a transient network call) where it would fail initially but succeed here?

tagboola (approved):

FirebaseAppDistribution/Sources/FIRAppDistributionUIService.m
tejasd: This only checks whether it's signed in - doesn't check for a release

PR #10977, opened 2023-03-21
Add new goog-listen-tags for BloomFilter

milaGGL:
#no-changelog
ported from web: https://github.com/firebase/firebase-js-sdk/pull/7107
android: https://github.com/firebase/firebase-android-sdk/pull/4777


Firestore/core/src/remote/remote_event.h
```diff
@@ -246,11 +248,11 @@ class TargetState {
 class RemoteEvent {
  public:
   using TargetChangeMap = std::unordered_map<model::TargetId, TargetChange>;
-  using TargetSet = std::unordered_set<model::TargetId>;
+  using TargetMismatchMap = std::map<model::TargetId, local::QueryPurpose>;
```
milaGGL, line 251: here, original TargetSet used unordered_set
in Web SDK, it used SortedSet:https://github.com/firebase/firebase-js-sdk/blob/eb049456555122e2a7350449b42ac07b8de175f7/packages/firestore/src/remote/remote_event.ts#L49
in android SDK, it used Set: https://github.com/firebase/firebase-android-sdk/blob/687e079d401db1a75d2559038b879726a476ca3e/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteEvent.java#L30

Should I use a SortedMap here?

dconeybe: No. We are moving away from SortedSet/SortedMap and the previous implementation used an unsorted set. Good question though.

Actually, please use `std::unordered_map` instead of `std::map` because `std::unordered_map` has much better performance because it is a "hash map", compared to `std::map` which is a "linked map".

Firestore/core/src/remote/remote_event.h
```diff
@@ -41,6 +42,8 @@ class TargetData;
 
 namespace remote {
 
+enum class BloomFilterApplicationStatus { Success, Skipped, FalsePositive };
```
dconeybe, line 45: Please check with the rest of the code, but I think the naming convention for enum elements begins with lowercase "k", such as `kSuccess`, `kSkipped`, and `kFalsePositive`.

Firestore/core/src/remote/remote_store.cc
```diff
@@ -351,8 +352,7 @@ void RemoteStore::RaiseWatchSnapshot(const SnapshotVersion& snapshot_version) {
     // that we flag the first re-listen this way without impacting future
     // listens of this target (that might happen e.g. on reconnect).
     TargetData request_target_data(target_data.target(), target_id,
-                                   target_data.sequence_number(),
-                                   QueryPurpose::ExistenceFilterMismatch);
+                                   target_data.sequence_number(), entry.second);
```
dconeybe, line 355: Please assign `entry.second` to a local variable to improve readability. As it is currently written, one would have to track down the definition of `target_mismatches()`, then the definition of `TargetMismatchMap` to know what the type of `entry.second` is.

dconeybe (changes requested):

dconeybe (approved):

PR #10981, opened 2023-03-21
Firestore: defer copying bloom filter bytes until the bloom filter is actually needed

dconeybe:
This is a temporary PR that will eventually be closed without merging. It shows one way that the bloom filter bytes could potentially avoid the expensive copy of the bloom filter bytes until the bloom filter is actually needed. See https://github.com/firebase/firebase-ios-sdk/pull/10953#discussion_r1143636462 for details.

#no-changelog

### DIFF
--- a/Firestore/core/src/nanopb/fields_array.h
+++ b/Firestore/core/src/nanopb/fields_array.h
@@ -227,6 +227,21 @@ inline const pb_field_t* FieldsArray<
   return google_firestore_v1_StructuredAggregationQuery_Aggregation_Count_fields;
 }
 
+template <>
+inline const pb_field_t* FieldsArray<google_firestore_v1_ExistenceFilter>() {
+  return google_firestore_v1_ExistenceFilter_fields;
+}
+
+template <>
+inline const pb_field_t* FieldsArray<google_firestore_v1_BloomFilter>() {
+  return google_firestore_v1_BloomFilter_fields;
+}
+
+template <>
+inline const pb_field_t* FieldsArray<google_firestore_v1_BitSequence>() {
+  return google_firestore_v1_BitSequence_fields;
+}
+
 }  // namespace nanopb
 }  // namespace firestore
 }  // namespace firebase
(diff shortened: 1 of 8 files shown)


Firestore/core/test/unit/remote/watch_change_test.cc
```diff
@@ -43,19 +43,31 @@ TEST(WatchChangeTest, CanCreateDocumentWatchChange) {
...
     EXPECT_EQ(change.target_id(), 5);
+    EXPECT_EQ(change.filter().count(), 7);
+
+    nanopb::Message<google_firestore_v1_BloomFilter> bloom_filter_copy;
+    bloom_filter_copy->hash_count = 33;
+    bloom_filter_copy->has_bits = true;
+    bloom_filter_copy->bits.padding = 7;
+    bloom_filter_copy->bits.bitmap =
+        nanopb::MakeBytesArray(std::vector<uint8_t>{0x42, 0xFE});
+    EXPECT_TRUE(change.filter().bloom_filter().value() == bloom_filter_copy);
```
milaGGL, line 70: `EXPECT_EQ(change.filter().bloom_filter().value(), bloom_filter_copy);` crashes the build.

dconeybe: The problem was that the `operator==` for `google_firestore_v1_BloomFilter` was defined in the wrong place. It was defined in the `firebase::firestore::remote` namespace in `existence_filter.h`, but should have been defined in the `firebase::firestore` namespace (the same namespace in which `google_firestore_v1_BloomFilter` is defined). Being in the wrong namespace prevented it from being found by ADL (address-dependent lookup) at compile time.

It seemed like a good idea to move the `operator==` functions out of the `existence_filter.h` file. So I did that by adding a new `operators.h` file in the `nanopb` directory. I also added unit tests for it (which need a bit more work in a follow-up PR).

I also added `operator==` for `nanopb::Message`, which was missing. This allows comparing `nanopb::Message` objects directly with each other, and eventually calling the `operator==` overloads for the underlying protos.

With these changes, the `EXPECT_EQ` now works, and the code is much better as well, with proper design and unit test coverage.

See https://github.com/firebase/firebase-ios-sdk/pull/10981/commits/7c7cb617af064db9648252333043ccffc441ed29

dconeybe:
I'm abandoning this work, for now.

PR #10991, opened 2023-03-22
[auth-swift] Style auth-swift with clang-format16

ncooke3:
#no-changelog


ncooke3:
I'll merge on green check. cc: @paulb777

ncooke3:
Err, actually, @paulb777, let me know you are actively working on these affected files and I can hold off.

paulb777:
I'm not sure we want to do this, since ideally the ObjC sample should stay as close as possible to the one in master.

ncooke3:
@paulb777, fair point, though I'm not sure why these style changes are failing in master. Closing.

paulb777:
See https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/style.sh#L162

PR #10993, opened 2023-03-23
AuthUser -> Swift

paulb777:
Eliminates all Objective C for the User module.  There are still some TODO's and Swift implementation gaps, but there is enough here to pass the unit tests and integration tests.

Next step is to rebase (and reimplement recent master stuff) to fix quickstart tests.
then to implement FIRUserTests in Swift.


FirebaseAuth/Sources/Swift/User/User.swift
```diff
@@ -0,0 +1,2076 @@
...
+    coder.encode(photoURL, forKey: kPhotoURLCodingKey)
+    coder.encode(displayName, forKey: kDisplayNameCodingKey)
+    coder.encode(metadata, forKey: kMetadataCodingKey)
+    coder.encode(tenantID, forKey: kTenantIDCodingKey)
+    if let auth {
+      coder.encode(auth.requestConfiguration.APIKey, forKey: kAPIKeyCodingKey)
+      coder.encode(auth.requestConfiguration.appID, forKey: kFirebaseAppIDCodingKey)
+    }
+    coder.encode(tokenService, forKey: kTokenServiceCodingKey)
+    #if os(iOS)
```
ncooke3, line 2043: thinking aloud for future improvements, maybe codable could help cut down this boilerplate? Auth serializes a couple pieces of data. And I bet we can handle the platform specific properties by #if'ing in their definitions on acceptable platforms

ncooke3 (commented):
great work Paul– that was a huge file to port 🙂

PR #10996, opened 2023-03-23
[auth-swift] Move AuthStorage protocol to own file; split up AuthKeychainServices impl.

ncooke3:
### Context
- AuthKeychainServices has two sets of API, both use to store different data in keychain. The implementation of both sets are not similar enough to share, so it makes more sense to split the API surface into two types. I will then re-shape the new type's API to confirm to AuthStorage protocol. The end result will be two types (both conforming to AuthStorage protocol) that can access the keychain. I'll then re-evaulate from there.
- Next up is working on passing a `id<FIRAuthStorage>` into the initializer of Auth

#no-changelog


FirebaseAuth/Sources/Swift/Storage/AuthSharedKeychainServices.swift
```diff
@@ -0,0 +1,89 @@
...
+import Foundation
+
+// TODO(ncooke3): Reshape API to conform to `AuthStorage`.
+@objc(FIRAuthSharedKeychainServices) public class AuthSharedKeychainServices: NSObject {
+  // MARK: - Private methods for shared keychain operations
+
+  /** @fn getItemWithQuery:error:
+   @brief Get the item from keychain by given query.
+   @param query The query to query the keychain.
+   @return The item of the given query. nil if not exsit.
```
paulb777, line 24: typo: exist

FirebaseAuth/Sources/Swift/Storage/AuthSharedKeychainServices.swift
```diff
@@ -0,0 +1,89 @@
...
+   */
+  @objc public func getItem(query: [String: Any]) throws -> DataWrapper {
+    var mutableQuery = query
+    mutableQuery[kSecReturnData as String] = true
+    mutableQuery[kSecReturnAttributes as String] = true
+    mutableQuery[kSecMatchLimit as String] = 2
```
paulb777, line 30: why 2?

paulb777 (approved):

FirebaseAuth/Sources/Swift/Storage/AuthSharedKeychainServices.swift
ncooke3: I'm assuming for the same reasoning here: https://github.com/firebase/firebase-ios-sdk/pull/10996/files#diff-44fc5c6e263aabae092d659983a63fc9a2f43db0468719b0dd7d7246e18a5d12L97

I'll add a similar comment.

paulb777:
The integration test failure might be a project data corruption resulting from my PR. I'm ok to disregard for now.  and quickstart is pending a master merge

ncooke3:
Thanks @paulb777, I was getting set up to repro.

PR #10998, opened 2023-03-23
[auth-swift] Move keychain intitialization code into intializer

ncooke3:
### Context
- Auth's keychain storage instances are initialized via a helper method called
  from the Auth designated initializer. Moving the keychain initialization code
  into the intializer should be safe as the initializing the keychain accessor
  types have no side effect.

#no-changelog


ncooke3:
Merging. Feel free to leave additional feedback.

PR #10068, opened 2022-08-04
Add Sign in with Apple Display Name API and unit test

aiwenisevan:
### Discussion

  * This resolves [Sign in with Apple display name not fulfilled [235848270] - Buganizer](http://b/235848270)

### Testing

  * Unit testing Passed for both `FIROAuthProviderTest` and `FIRAuthTest`

### API Changes

  * Internal API proposal: [Sign in with Apple Display Name API](http://doc/17BPHhMVfSfmKmELjt0m2F1x7YrM2TQUO13sgmFn5MSM#heading=h.x94y9zzc13xn)

PR #11005, opened 2023-03-24
Reimplement #10068 in auth-swift after rebase

paulb777:
Reimplement #10068 in auth-swift after rebase of master
Removes FIRAuthTests.m. It will need to be reimplemented from master after FIRAuth.m is eliminated.


ncooke3 (approved):

paulb777:
Going to merge here. To fix the quickstart, There is still an additional issue getting Swift computed properties visible to ObjC. I'll work on that next.

PR #11006, opened 2023-03-24
Objective C property functions should be Swift computed properties

paulb777:
#no-changelog


paulb777:
Quickstart is green again 😄

ncooke3 (approved):

PR #11009, opened 2023-03-24
[auth-swift] Use stored keychain service at app cleanup

ncooke3:
### Context
- Auth conforms to Core's library component protocol and implements the `appWillBeDeleted:` API.
- The implementation removes the Auth's instances stored info from the keychain.
- If we are injecting the keychain instance into Auth, we ideally are using the the injected version at cleanup time–– instead of creating this new one for cleanup purposes.
#no-changelog

### DIFF
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -2116,14 +2116,8 @@ typedef void (^FIRSignupNewUserCallback)(FIRSignUpNewUserResponse *_Nullable res
 
 - (void)appWillBeDeleted:(nonnull FIRApp *)app {
   dispatch_async(FIRAuthGlobalWorkQueue(), ^{
-    // This doesn't stop any request already issued, see b/27704535 .
-    NSString *keychainServiceName = [FIRAuth keychainServiceNameForAppID:app.options.googleAppID];
-    if (keychainServiceName) {
-      FIRAuthKeychainServices *keychain =
-          [[FIRAuthKeychainServices alloc] initWithService:keychainServiceName];
-      NSString *userKey = [NSString stringWithFormat:kUserKey, app.name];
-      [keychain removeDataForKey:userKey error:NULL];
-    }
+    NSString *userKey = [NSString stringWithFormat:kUserKey, app.name];
+    [_keychainServices removeDataForKey:userKey error:NULL];
     dispatch_async(dispatch_get_main_queue(), ^{
       // TODO: Move over to fire an event instead, once ready.
       [[NSNotificationCenter defaultCenter] postNotificationName:FIRAuthStateDidChangeNotification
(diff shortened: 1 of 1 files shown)


ncooke3:
@paulb777, could you give this one a careful review? I have a couple questions:
1. I don't think there is a retain cycle, do you see one? 
2. It makes sense for the block to be async but we need to retain `self`, since we don't weakify it, is self guaranteed to live long enough for the block to use it?
3. To retain self, is referencing the `_keychainServices` enough, or do I need to do `self->_keychainServices`.

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2116,14 +2116,8 @@ - (BOOL)getUser:(FIRUser *_Nullable *)outUser error:(NSError *_Nullable *_Nullab
...
-    // This doesn't stop any request already issued, see b/27704535 .
-    NSString *keychainServiceName = [FIRAuth keychainServiceNameForAppID:app.options.googleAppID];
-    if (keychainServiceName) {
-      FIRAuthKeychainServices *keychain =
-          [[FIRAuthKeychainServices alloc] initWithService:keychainServiceName];
-      NSString *userKey = [NSString stringWithFormat:kUserKey, app.name];
-      [keychain removeDataForKey:userKey error:NULL];
-    }
+    NSString *userKey = [NSString stringWithFormat:kUserKey, app.name];
+    [_keychainServices removeDataForKey:userKey error:NULL];
```
ncooke3, line 2120: @paulb777, could you give this one a careful review? I have a couple questions:
1. I don't think there is a retain cycle, do you see one? 
2. It makes sense for the block to be async but we need to retain `self`. Considering I don't  weakify it, is self guaranteed to live long enough for the block to use it?
3. To retain self, is referencing the `_keychainServices` enough, or do I need to do `self->_keychainServices`.

I think there are other ways to achieve what I need, but this one is the most direct.

ncooke3: Based on https://stackoverflow.com/a/10535411/9331576, https://stackoverflow.com/a/21215145/9331576, andhttps://stackoverflow.com/questions/12287696/clarification-on-apples-block-docs–– I think this is fine. And I think it would be equivalent to say `self->_keychainServices`, and I arguably should because it makes the self capture more explicit.

paulb777: Agreed that it's not necessary, but clearer to be explicit.

paulb777 (approved):

FirebaseAuth/Sources/Auth/FIRAuth.m
ncooke3: Cool, adjusted. Thank you for looking!

ncooke3: And followup @paulb777, if I did the weak/strong thing, would I run the risk of the non-capturing self so the block becomes a no-op? My understanding is yes and that is why I did not weakify. Do you agree?

paulb777: Agreed.

PR #11011, opened 2023-03-26
AuthWebView and AuthWebViewController

mortenbekditlevsen:
Another few files converted to Swift. In blind, as they are guarded by iOS platform checks and my swift-test runs on macOS.


paulb777:
Merged with #11014. Thanks!

PR #10995, opened 2023-03-23
Turn off flaky Crashlytics tests

samedson:
2 different flaky tests are disabled in this PR. A task has been filed for the Crashlytics team to fix this:

 - https://github.com/firebase/firebase-ios-sdk/actions/runs/4485746539/jobs/7889007887
 - https://github.com/firebase/firebase-ios-sdk/actions/runs/4319822264/jobs/7539418159

This should also fix https://github.com/firebase/firebase-ios-sdk/issues/10873

PR #11021, opened 2023-03-27
[Infra] Remove test artifact directory in test set up

ncooke3:
### Context
- A potential hint–– the last three times a test from this file failed in CI, it was the first test run in the file. This may mean another test is leaving some state behind. I'm proposing we copy some of the `tearDown` code into the `setUp` code to enforce a clean artifact directory.
  - https://github.com/firebase/firebase-ios-sdk/actions/runs/4485746539/jobs/7889007887#step:5:972
  - https://github.com/firebase/firebase-ios-sdk/actions/runs/4528004614/jobs/7974385037#step:5:454
  - https://github.com/firebase/firebase-ios-sdk/actions/runs/4528004614/jobs/7974385037#step:5:997

#no-changelog

### DIFF
--- a/Crashlytics/UnitTests/FIRCLSExistingReportManagerTests.m
+++ b/Crashlytics/UnitTests/FIRCLSExistingReportManagerTests.m
@@ -54,6 +54,11 @@
 
   self.fileManager = [[FIRCLSTempMockFileManager alloc] init];
 
+  // Cleanup potential artifacts from other test files.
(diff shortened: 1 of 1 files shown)


samedson (approved):
That's a good call. Thanks for digging into this 👍

paulb777:
Should we also revert #10995 to reenable the recently disabled tests?

paulb777 (approved):

PR #11024, opened 2023-03-27
[auth-swift] Make request components for deterministic to avoid test flakes

ncooke3:
### Context
- I was seeing non-deterministic flakes locally and tracked it down to the `testVerifyAssertionRequestOptionalFields` test method. They were more reproducible when I ran the mentioned test 100-1000 times. The issue is that the test specifies a hardcoded JSON string to compare the firstname/lastname request component, but the implementation dynamically creates the JSON string via the `JSONSerialization.data(withJSONObject:)` API. This API seems to create a JSON string without preserving the given dictionary's given order. While the order doesn't matter in practice, it eliminates the flakes as the hardcoded test string will always match. Interestingly, using `JSONEncoder().encode(_:)` seems to preserve the order as I wasn't able to reproduce when re-running 1k times.

#no-changelog

### DIFF
--- a/FirebaseAuth/Sources/Swift/Backend/RPC/VerifyAssertionRequest.swift
+++ b/FirebaseAuth/Sources/Swift/Backend/RPC/VerifyAssertionRequest.swift
@@ -247,7 +247,7 @@ private let kLastNameKey = "lastName"
       }
       let userDict = [kNameKey: nameDict]
       do {
-        let userJson = try JSONSerialization.data(withJSONObject: userDict)
+        let userJson = try JSONEncoder().encode(userDict)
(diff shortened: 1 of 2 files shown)


paulb777:
Interesting. This looks better than the hack fix I made at https://github.com/firebase/firebase-ios-sdk/blob/auth-swift/FirebaseAuth/Tests/Unit/VerifyAssertionTests.swift#L121.  Would you remove componentsAlt from there?

paulb777 (approved):
Thanks!

paulb777:
Were you able to find any documentation about `JSONEncoder().encode` being deterministic for encoding dictionaries?

PR #11035, opened 2023-03-29
Firestore: nanopb/message.h: Add ability to "steal" a proto using rvalue references

dconeybe:


dconeybe:
I'm abandoning this work.

PR #11038, opened 2023-03-29
Use ByteString instead of std::vector in the BloomFilter

dconeybe:


Firestore/core/src/remote/bloom_filter.cc
```diff
@@ -38,9 +39,9 @@ bool HasSameBits(const BloomFilter& lhs, const BloomFilter& rhs) {
     return true;
   }
 
-  const std::vector<uint8_t>& bitmap1 = lhs.bitmap();
-  const std::vector<uint8_t>& bitmap2 = rhs.bitmap();
-  const auto byte_count = static_cast<int32_t>(bitmap1.size());
+  const auto byte_count = static_cast<int32_t>(lhs.bitmap().size());
```
milaGGL, line 42: why auto since we know its type will be int32_t?

milaGGL (approved):
LGTM

Firestore/core/src/remote/bloom_filter.cc
dconeybe: Lint suggests (requires?) using `auto` when the type is explicitly mentioned on the right-hand side of the equal sign. By mentioning it twice it actually opens up the opportunity for an unintended implicit narrowing conversion if there is a typo on one side of the equal sign.

PR #11039, opened 2023-03-29
Scripts cleanup: remove python `six` usage and get `ruby` from `PATH`

dconeybe:
This PR makes the scripts easier to run.

The first change is to `scripts/sync_project.rb` to look up the `ruby` interpreter in the `PATH` rather than hardcoding `/usr/bin/ruby`. For me, personally, this fixed a bug where I wanted it to use `ruby` from homebrew rather than the system `ruby` interpreter.

The second change is to remove the usage of the `six` library from Python scripts. Since everyone should be using Python 3 by now, there is no need for `six`. Also, this removes the requirement that people install this dependency on their local machines and make the scripts usable from a "vanilla" Python distribution.

The third change is to upgrade the Python version used in the GitHub Actions workflows from 3.6/3.7 to the latest, 3.11.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,4 +15,3 @@ source 'https://rubygems.org'
 gem 'cocoapods', '1.12.0'
 gem 'cocoapods-generate', '2.0.1'
 gem 'danger', '8.4.5'
-
--- a/Gemfile
+++ b/Gemfile
@@ -15,3 +15,4 @@ source 'https://rubygems.org'
 gem 'cocoapods', '1.12.0'
 gem 'cocoapods-generate', '2.0.1'
 gem 'danger', '8.4.5'
+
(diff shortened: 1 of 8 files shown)


dconeybe:
@ncooke3

> No idea if this is relevant/helps but we do have a [`.ruby-version`](https://github.com/firebase/firebase-ios-sdk/blob/master/.ruby-version) file checked into the repo.

I _think_ that `.ruby-version` files are used by environment managers like [rbenv](https://github.com/rbenv/rbenv). In my case, I just have ruby installed via homebrew without any "version manager". I don't think `.ruby-version` is consulted when `/usr/bin/ruby` is hardcoded. But I've never written a line of ruby code in my life, so I could be wrong.

Gemfile
```diff
@@ -15,3 +15,4 @@ source 'https://rubygems.org'
 gem 'cocoapods', '1.12.0'
 gem 'cocoapods-generate', '2.0.1'
 gem 'danger', '8.4.5'
+
```
paulb777, line 18: revert

paulb777 (approved):
Looks good. Thanks!

PR #11047, opened 2023-03-30
Fix App Quality Sessions field for Bundle Short Version

samedson:
We had the fields mixed up:

Screenshot 2023-03-30 at 4 43 50 PM

### DIFF
--- a/FirebaseSessions/Sources/ApplicationInfo.swift
+++ b/FirebaseSessions/Sources/ApplicationInfo.swift
@@ -107,11 +107,11 @@ class ApplicationInfo: ApplicationInfoProtocol {
   }
 
   var appBuildVersion: String {
-    return Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? ""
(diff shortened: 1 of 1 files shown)


visumickey (approved):
Changes are fairly trivial. Can we bring in some unit tests to ensure that the display version is a semver?

ncooke3:
Does this have any implications on persistently stored SDK data?

samedson:
@ncooke3 this might not be changelog eligible because it isn't user visible yet. 

It will change how the data is stored on the backend, but that's the desired outcome.

ncooke3:
> @ncooke3 this might not be changelog eligible because it isn't user visible yet.
> 
> It will change how the data is stored on the backend, but that's the desired outcome.

Thanks Sam–– I mainly wanted to ensure this wouldn't cause a subtle schema change in local SDK storage.

samedson:
@ncooke3 Ah good question - this is a get-only field. We don't persist it, only write it into the proto in RAM.

PR #11057, opened 2023-04-01
New Cache Config API

wu-hui:


Firestore/Source/API/FIRLocalCacheSettings+Internal.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <memory>
+#include "Firestore/core/src/api/settings.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FIRPersistentCacheSettings (Internal)
+
+@property(nonatomic, assign)
+    const firebase::firestore::api::PersistentCacheSettings& internalSettings;
```
ehsannas, line 28: This is mainly for my benefit (pardon my ignorance. I'm not an obj-c expert).

I was at first a bit confused about having `internalSettings` here as public `@property` and then having `_internalSettings` as a private member in the .mm file.

this results in code like

return **_internalSettings** == otherSettings.**internalSettings**;

IIUC `otherSettings.internalSettings` invokes the getter method, and it's all good. But I wonder if a method is more idiomatic? e.g.

```
@interface FIRPersistentCacheSettings (Internal)
- (const firebase::firestore::api::PersistentCacheSettings&)internalSettings;
@end
```

ehsannas (commented):
Looks good... the swift code is somewhat new for me though

Firestore/core/test/unit/api/settings_test.cc
```diff
@@ -0,0 +1,163 @@
...
+  EXPECT_EQ(true, settings.persistence_enabled());
+  EXPECT_EQ(copy.cache_size_bytes(), settings.cache_size_bytes());
+  EXPECT_EQ(nullptr, settings.local_cache_settings());
+}
+
+TEST(Settings, CopyAssignmentOperator) {
+  Settings settings;
+  settings.set_host("host");
+  settings.set_ssl_enabled(true);
+  settings.set_local_cache_settings(PersistentCacheSettings{});
```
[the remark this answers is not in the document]
wu-hui: Done.

wu-hui (commented):
Thanks!

Firestore/core/src/api/settings.h
```diff
@@ -110,56 +109,50 @@ class LocalCacheSettings {
...
-                         const MemoryCacheSettings& rhs) {
-    return !(lhs == rhs);
-  }
   size_t Hash() const override;
-
- private:
-  friend class Settings;
 };
 
+bool operator!=(const Settings& lhs, const Settings& rhs);
```
ehsannas, line 144:
```
bool operator==(const Settings& lhs, const Settings& rhs);
```
is missing.

Firestore/core/src/api/settings.cc
```diff
@@ -28,15 +32,159 @@ constexpr bool Settings::DefaultPersistenceEnabled;
...
+        static_cast<const PersistentCacheSettings&>(settings));
+  }
+  UNREACHABLE();
+}
+
 size_t Settings::Hash() const {
+  if (cache_settings_ == nullptr) {
+    return util::Hash(host_, ssl_enabled_, persistence_enabled_,
+                      cache_size_bytes_);
+  }
```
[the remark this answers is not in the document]
ehsannas: I don't know about `util::Hash`, but `std::hash` handles nullptr correctly. ([example](https://godbolt.org/z/478j6xKE5)). I'd unit test it to make sure for `util::Hash`.

Firestore/core/src/api/settings.h
wu-hui: This is defined in class declaration at line 79.

Firestore/core/src/api/settings.cc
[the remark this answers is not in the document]
wu-hui: util::Hash does not handle unique_ptr (unlike std::hash), so we need to deference the pointer. In this case, we would have to dereference a null pointer.

I added a specialization for unique_ptr for this work.

Firestore/core/src/util/hashing.h
```diff
@@ -199,6 +201,12 @@ size_t RankedInvokeHash(K value, HashChoice<5>) {
   return InvokeHash(underlying);
 }
 
+template <typename K>
+auto RankedInvokeHash(const std::unique_ptr<K>& ptr, HashChoice<6>)
+    -> decltype(InvokeHash(*ptr)) {
+  return ptr ? InvokeHash(*ptr) : 23631;
+}
+
```
ehsannas, line 209: Will this code ever get executed?

Based on the comments above:

```
 * HashChoice is a recursive type, defined such that HashChoice<0> is the most
 * specific type with HashChoice<1> and beyond being progressively less
 * specific. This causes the compiler to prioritize the overloads with
 * lower-numbered HashChoice types, allowing compilation to succeed even if
 * multiple specializations match.
 ```

And `HachChoice<2>`: `Hashes the given value if it has a specialization of std::hash.`

And `std::hash` does have a specialization for `std::unique_ptr`:
https://en.cppreference.com/w/cpp/memory/unique_ptr/hash

wu-hui: Nope, because all these settings object do not support std::hash, we use our own Hash() functions.

ehsannas: Oh I see... Thanks for the clarification

ehsannas (approved):

PR #11066, opened 2023-04-03
[Firestore] Change Firestore's release distro from source to binary for SPM

ncooke3:
### Context
- Change Firestore target definition to be a binary target rather than a source target
- Change gRPC and Abseil dependencies to point to new binary dependency repos
- Corresponding gRPC PR: https://github.com/google/grpc-binary/pull/2
- Corresponding Abseil PR: https://github.com/google/abseil-cpp-binary/pull/2

#### Before Merging
- [x] Wait for rc0 zip to be created
- [x] Update and publish Abseil binary release using artifact from zip
  -  `abseil.xcframework`
- [x] Update and publish GRPC binary release using artifacts from zip
  -  `gRPC-C++.xcframework`
  -  `gRPC-Core.xcframework`
  -  `BoringSSL-GRPC.xcframework`
- [x] Update and publish Firestore binary release using artifact from zip
  -  `FirebaseFirestore.xcframework`
- [x] Update this branch with new Firestore checksum
- [x] Confirm this branch resolves
- [x] Confirm Firestore quickstart builds and functions
- [x] Update CHANGELOG entry's version
- [x] Update PR title and description
- [x] Merge this PR when:
   - [x] Approvals
   - [x] Green CI

#### After Merging
- [ ] Move release (CocoaPods) tag to include this new commit

### DIFF
--- a/FirebaseTestingSupport/Firestore/Sources/FIRQueryFake.mm
+++ b/FirebaseTestingSupport/Firestore/Sources/FIRQueryFake.mm
@@ -14,8 +14,6 @@
 
 #import "FirebaseTestingSupport/Firestore/Sources/Public/FirebaseFirestoreTestingSupport/FIRQueryFake.h"
 
-#import "Firestore/Source/API/FIRQuery+Internal.h"
-
 @implementation FIRQueryFake
 
 - (instancetype)init {
--- a/Package.swift
+++ b/Package.swift
@@ -638,73 +638,24 @@ let package = Package(
 
     .target(
       name: "FirebaseFirestoreTarget",
-      dependencies: [.target(name: "FirebaseFirestore",
-                             condition: .when(platforms: [.iOS, .tvOS, .macOS]))],
+      dependencies: [
+        .target(
+            name: "FirebaseFirestore",
+            condition: .when(platforms: [.iOS, .tvOS, .macOS])
+        ),
+        .product(name: "abseil", package: "abseil"),
+        .product(name: "gRPC-C++", package: "gRPC"),
+        .product(name: "nanopb", package: "nanopb"),
+        "FirebaseCore",
+        "leveldb"
+      ],
       path: "SwiftPM-PlatformExclude/FirebaseFirestoreWrap"
     ),
 
-    .target(
-      name: "FirebaseFirestore",
-      dependencies: [
-        "FirebaseCore",
-        "leveldb",
-        .product(name: "nanopb", package: "nanopb"),
-        .product(name: "abseil", package: "abseil-cpp-SwiftPM"),
-        .product(name: "gRPC-cpp", package: "grpc-ios"),
-      ],
-      path: "Firestore",
-      exclude: [
-        "CHANGELOG.md",
-        "CMakeLists.txt",
-        "Example/",
-        "LICENSE",
-        "Protos/CMakeLists.txt",
-        "Protos/Podfile",
-        "Protos/README.md",
-        "Protos/build_protos.py",
-        "Protos/cpp/",
-        "Protos/lib/",
-        "Protos/nanopb_cpp_generator.py",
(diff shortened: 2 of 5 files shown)


ncooke3:
Green CI: https://github.com/firebase/firebase-ios-sdk/pull/11066/checks?sha=d8bdd5a86a8f141397186704859175bf57b970da

Now, will remove the dummy file that was used to trigger the Firestore CI.

FirebaseTestingSupport/Firestore/Sources/FIRQueryFake.mm
```diff
@@ -14,8 +14,6 @@
 
 #import "FirebaseTestingSupport/Firestore/Sources/Public/FirebaseFirestoreTestingSupport/FIRQueryFake.h"
 
-#import "Firestore/Source/API/FIRQuery+Internal.h"
```
ncooke3, line 17: This is test code and the change was necessary to get FirestoreTestingSupport target building.

andrewheard:
Any luck with "Confirm Firestore quickstart builds and functions" ? Otherwise LGTM!

andrewheard (approved):

Package.swift
```diff
@@ -161,12 +161,12 @@ let package = Package(
       "2.30909.0" ..< "2.30910.0"
     ),
     .package(
-      url: "https://github.com/firebase/abseil-cpp-SwiftPM.git",
-      "0.20220203.1" ..< "0.20220204.0"
+      url: "https://github.com/google/abseil-cpp-binary.git",
+      "1.2021110200.0" ..< "1.2021110300.0"
```
paulb777, line 165: So we were previously using different versions of abseil between CocoaPods and SPM?  

Perhaps for the next release, we should update to 20220203?

Package.swift
```diff
@@ -637,73 +637,24 @@ let package = Package(
...
+        .product(name: "abseil", package: "abseil-cpp-binary"),
+        .product(name: "gRPC-C++", package: "grpc-binary"),
+        .product(name: "nanopb", package: "nanopb"),
         "FirebaseCore",
         "leveldb",
-        .product(name: "nanopb", package: "nanopb"),
-        .product(name: "abseil", package: "abseil-cpp-SwiftPM"),
-        .product(name: "gRPC-cpp", package: "grpc-ios"),
       ],
-      path: "Firestore",
```
paulb777, line 654: What is the strategy for re-enabling source SPM builds locally?

paulb777 (approved):

Package.swift
ncooke3: Yep, I noticed that. This version is coming from the zip README, which is coming from here: https://github.com/firebase/firebase-ios-sdk/blob/98e218b68643e8b02cbe31f5de53fe1e04c23d74/FirebaseFirestore.podspec#L93

Package.swift
ncooke3: In this PR's current state, there isn't a good strategy as you'd have to effectively revert this PR. One thing I could do is take a similar approach to how we conditionally configure GoogleAppMeasurement and introduce an env var that will conditionally build Firestore from source rather than binary.

Also worth noting that if we build Firestore from source, the Abseil/gRPC xcframework headers don't work in their current state so we'd also conditonally need to load in the source dependencies. (Alternatively, we could do some header processing on the existing xcframework headers to make them usable from source, but that would require some tooling).

ncooke3: Discussed offline. Decision is to hold off for now.

Package.swift
ncooke3: Discussed offline. I'll look at aligning the versions between CP and SPM. cc: @wu-hui

ncooke3:
Merging to re-tag for release while `master` is still clean. I can address any follow-up questions/concerns in a follow-up PR.

PR #11074, opened 2023-04-04
make style.sh error checking more tolerable

wu-hui:


wu-hui:
@paulb777 The fix i did yesterday was a bit too strict.

PR #11091, opened 2023-04-08
[Auth] Remove unreachable code

ncooke3:
_I recommend viewing the diff without whitespace changes. [See nice diff.](https://github.com/firebase/firebase-ios-sdk/pull/11091/files?diff=unified&w=1)_

### Context
The outer `else` clause is unreachable and should be removed. We pass a pointer to `- [FIRAuthStoredUserManager getStoredUserAccessGroupWithError:]` but there is no resulting code path for it be populated.
Screen Shot 2023-04-08 at 4 21 30 PM

The implementation of `- [FIRAuthStoredUserManager getStoredUserAccessGroupWithError:]` is pasted below. The only way for the `outError` to be populated is in the `- [FIRAuthUserDefaults dataForKey:error:]` method on line 53.
https://github.com/firebase/firebase-ios-sdk/blob/8badb28bf2727941c3e4b41fab87905de8595ca4/FirebaseAuth/Sources/SystemService/FIRAuthStoredUserManager.m#L52-L60

But, looking at the implementation of `- [FIRAuthUserDefaults dataForKey:error:]`, no code path will populate the error.
https://github.com/firebase/firebase-ios-sdk/blob/8badb28bf2727941c3e4b41fab87905de8595ca4/FirebaseAuth/Sources/Storage/FIRAuthUserDefaults.m#L44-L50

So given that, it's safe to remove the code in the outer `else` clause as it's unreachable. This simplifies the surrounding code.

I found this while porting `FIRAuthStoredUserManager` in the `auth-swift` branch.

#no-changelog

### DIFF
--- a/FirebaseAuth/Sources/Auth/FIRAuth.m
+++ b/FirebaseAuth/Sources/Auth/FIRAuth.m
@@ -526,54 +526,42 @@ static NSMutableDictionary *gKeychainServiceNameForAppName;
           [[FIRAuthStoredUserManager alloc] initWithServiceName:keychainServiceName];
     }
 
-    NSError *error;
     NSString *storedUserAccessGroup =
-        [strongSelf.storedUserManager getStoredUserAccessGroupWithError:&error];
-    if (!error) {
-      if (!storedUserAccessGroup) {
-        FIRUser *user;
-        if ([strongSelf getUser:&user error:&error]) {
-          strongSelf.tenantID = user.tenantID;
-          [strongSelf updateCurrentUser:user byForce:NO savingToDisk:NO error:&error];
-          self->_lastNotifiedUserToken = user.rawAccessToken;
-        } else {
-#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_MACCATALYST
-          if (error.code == FIRAuthErrorCodeKeychainError) {
-            // If there's a keychain error, assume it is due to the keychain being accessed
-            // before the device is unlocked as a result of prewarming, and listen for the
-            // UIApplicationProtectedDataDidBecomeAvailable notification.
-            [strongSelf addProtectedDataDidBecomeAvailableObserver];
-          }
-#endif  // TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_MACCATALYST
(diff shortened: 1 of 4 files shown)


ncooke3:
I also renamed two `FIRAuthStoredUserManager` APIs to prevent clients from doing unnecessary error handling when using them. These APIs are used internally only.
- `-[FIRAuthStoredUserManager getStoredUserAccessGroupWithError:]`
- `-[FIRAuthStoredUserManager setStoredUserAccessGroup:error:]`

FirebaseAuth/Sources/Auth/FIRAuth.m
```diff
@@ -2307,7 +2294,7 @@ - (nullable NSString *)getUserID {
 - (BOOL)internalUseUserAccessGroup:(NSString *_Nullable)accessGroup
                              error:(NSError *_Nullable *_Nullable)outError {
   BOOL success;
-  success = [self.storedUserManager setStoredUserAccessGroup:accessGroup error:outError];
```
paulb777, line 2310: Isn't the `outError` necessary to set the error code in the public api: `- (BOOL)useUserAccessGroup:(NSString *_Nullable)accessGroup
                     error:(NSError *_Nullable *_Nullable)outError`

ncooke3: If you look at the implementation of that API, no code paths set that error code.

ncooke3: The problem is that AuthUserDefaults conforms to AuthStorage– which defines throwing API for getting and setting values. But, the implementation of AuthUserDefaults does not actually throw any errors. So clients that call this code have error handling code paths that are effectively no-ops. We could change it so that instead of returning `nil` when, for example, nothing is stored, we return an error–– but that could get risky as there could be some subtle behavior changes (I'd have to think about it more).

paulb777: I see. thanks for the clarification.

paulb777 (commented):
LGTM

renkelvin (approved):
Thanks for cleaning this up!

PR #11094, opened 2023-04-10
[Infra] Avoid subshells in test script

ncooke3:
#no-changelog

### DIFF
--- a/scripts/check_firestore_symbols.sh
+++ b/scripts/check_firestore_symbols.sh
@@ -34,6 +34,8 @@
 #
 # USAGE: ./check_firestore_symbols.sh <PATH_TO_FIREBASE_REPO> <PATH_TO_FIRESTORE_XCFRAMEWORK>
 
+set -euo pipefail
+
 if [[ $# -ne 2 ]]; then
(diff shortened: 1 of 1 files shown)


scripts/check_firestore_symbols.sh
```diff
@@ -123,35 +125,38 @@ cd "$TEST_PKG_ROOT"
...
+#
+# Note: In the case where the diff is non-empty, the diff command will
+# return exit code 1, which will cause the set pipefail to terminate execution.
+# To avoid this, `|| true` ensures the exit code always indicates success.
+DIFF=$(
+    git diff --no-index \
+        objc_symbols_without_linker_flag.txt \
+        objc_symbols_with_linker_flag.txt \
+        || true
+)
```
ncooke3, line 159: cc: @morganchen12, in my case, adding the `|| true` ensure that the diff command always returns a successful error code. I figured this workaround was better than spawning subshells.

andrewheard (approved):

scripts/check_firestore_symbols.sh
ncooke3: that's a good point and interesting strategy. i think ill hold off only because it should be very unlikely for it to fail. future me could very well eat these words though.

PR #11095, opened 2023-04-10
Use fallback value when source == .static

peterfriese:
The @RemoteConfigProperty property wrapper should use the `fallback` value when the user selects “Use in-app default” in the Firebase console.

### DIFF
--- a/FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
+++ b/FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
@@ -37,19 +37,13 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
     self.key = key
     remoteConfig = RemoteConfig.remoteConfig()
     self.fallbackValue = fallbackValue
+
     // Initialize with fallback value
     configValue = fallbackValue
+
     // Check cached remote config value
-    do {
-      let configValue: RemoteConfigValue = remoteConfig[key]
-      if configValue.source == .remote || configValue.source == .default {
-        self.configValue = try remoteConfig[key].decoded()
-      } else {
-        self.configValue = fallbackValue
-      }
-    } catch {
-      configValue = fallbackValue
-    }
+    applyConfigValue()
+
     NotificationCenter.default.addObserver(
       self, selector: #selector(configDidActivate), name: .onRemoteConfigActivated, object: nil
     )
@@ -61,13 +55,19 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
     if FirebaseApp.app()?.name != appName {
       return
     }
+    applyConfigValue()
+  }
+
+  private func applyConfigValue() {
     do {
       let configValue: RemoteConfigValue = remoteConfig[key]
-      if configValue.source == .remote {
+      if configValue.source == .remote || configValue.source == .default {
         self.configValue = try remoteConfig[key].decoded()
+      } else {
+        self.configValue = fallbackValue
       }
     } catch {
-      // Suppresses a hard failure if decoding failed.
+      configValue = fallbackValue
     }
   }
 }


ncooke3 (commented):
LGTM. Also, reminder to add a CHANGELOG entry if applicable. I'll defer approval to @charlotteliang!

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -61,13 +55,19 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
...
+  }
+
+  private func applyConfigValue() {
     do {
       let configValue: RemoteConfigValue = remoteConfig[key]
-      if configValue.source == .remote {
+      if configValue.source == .remote || configValue.source == .default {
         self.configValue = try remoteConfig[key].decoded()
+      } else {
+        self.configValue = fallbackValue
```
ncooke3, line 67: Is this the codepath or when `source == .static`?

peterfriese: This is for `source == .static`, yes. Would you rather I make this  more explicit?

peterfriese: I think it might be a good idea to make it more explicit (and thus, easier to understand).

charlotteliang (changes requested):
I don't think fallback value is equivalent to "in app default" value.

The fallback value is a pure swift language default value that is not the "in app default" value that's the default config value.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -61,13 +55,19 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
...
       }
     } catch {
-      // Suppresses a hard failure if decoding failed.
```
charlotteliang, line 70: I don't think this is right, we should only get remote value since this method is only triggered when config has remote value needs to be activated.

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -61,13 +55,19 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
     if FirebaseApp.app()?.name != appName {
       return
     }
+    applyConfigValue()
```
charlotteliang, line 58: from the way you refactor the logic, I'm not sure it aligns with the description because for the applyConfigValue() fundamentally didn't change any logic.

charlotteliang (changes requested):

peterfriese:
> I don't think fallback value is equivalent to "in app default" value.

If the `fallbackValue` is *not* intended to act as the in-app default value, there should be a(nother) mechanism for specifying the default value. 

The way it currently works is:
1. In the console, you select _Use in-app default value_
2. *Nothing happens*

And this - at least to me - doesn't sound right either. Unless I am missing something.

> The fallback value is a pure swift language default value that is not the "in app default" value that's the default config value.

Not sure I follow. To specify a language-level default value, wouldn't you rather want to assign a value to the underlying variable at initialisation time? I.e., `@RemoteConfigProperty(key="foo") var foo = "Some default value"`?

peterfriese:
> > The way it currently works is:
> > 
> > In the console, you select Use in-app default value Nothing happens
> 
> Can you elaborate the "nothing happens" part? I think the expected case here should be "use default config value if it's being set by developers". So if developer didn't set a config value, nothing should happen.

Sure. Take a look at the following code, which uses the RC property wrapper to connect the `cardColor` to an RC variable.

```swift
@RemoteConfigProperty(key: "cardColor", fallback: "#ff2d55") var cardColor

var body: some View {
    VStack {
      Text("What's your favourite number?")
        .font(.title)
        .multilineTextAlignment(.center)
      Spacer()
      Stepper(value: $viewModel.favouriteNumber, in: minValue...maxValue) {
        Text("\(viewModel.favouriteNumber)")
      }
    }
    .foregroundColor(.white)
    .background(Color(hex: cardColor))
}
```

When defining the RC variable and setting it to `#0000FF` in the console, the UI will turn blue - as expected. When toggling the switch in the console to _Use in-app default_, the UI will stay red (i.e. _nothing happens_), instead of turning to "Sizzling red" (`#ff2d55`), as specified in the first line of the snippet. See https://github.com/FirebaseExtended/firebase-video-samples/blob/peterfriese/apple/remote-config/fundamentals/apple/remote-config/final/Favourites/Shared/FavouriteNumberView.swift#L69 for the complete code sample. 

Expected behaviour: toggling to _Use in-app default value_ uses the value defined by the PW.

charlotteliang:
Expected behaviour: toggling to Use in-app default value uses the value defined by the PW.

No that's not how we designed the API, the "in-app default" value should be defined by developers, not the fall back in PW as described in the API definition. https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift#L36

@karenyz

karenyz:
> No that's not how we designed the API, the "in-app default" value should be defined by developers, not the fall back in PW as described in the API definition. https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigProperty.swift#L36
> 

This is also how I understood it. In RC the 3 value sources work as follows:
- **remote**: a value was defined remotely (server-side) for this parameter
- **default**: a value was set client-side for this parameter using the [`setDefaults`](https://firebase.google.com/docs/reference/swift/firebaseremoteconfig/api/reference/Classes/RemoteConfig#setdefaults_:) method in the SDK. Toggling "use in-app default" in the Console indicates that a parameter should use this client-side value, the server "opts out" of providing a value.
- **static**: this essentially means "parameter not found", i think the initial intention was so RC wouldn't return `null` values, but in reality this often causes confusion. This could occur if "use in-app default" is toggled but no default was ever set in the client, or simply this parameter key doesn't exist.

Going back to this PR, the PW should use both **remote** and **default** values as-is, and only use the fallback if there happens to be a decoding issue with the value, or there was no (user-defined) RC value found (**static**). 

Hope this clarifies things, if not, happy to discuss more!

FirebaseRemoteConfigSwift/Sources/PropertyWrapper/RemoteConfigValueObservable.swift
```diff
@@ -61,13 +55,19 @@ internal class RemoteConfigValueObservable<T: Decodable>: ObservableObject {
...
+  }
+
+  private func applyConfigValue() {
     do {
       let configValue: RemoteConfigValue = remoteConfig[key]
-      if configValue.source == .remote {
```
karenyz, line 66: ahh i see, so previously the PW value would only change if there was a remote value, and does nothing otherwise

karenyz (approved):
this change makes sense to me, it follows the remote > default > fallback ordering we discussed today

sorry for the confusion!

paulb777:

PR #11097, opened 2023-04-10
[Tooling] Add FIR_CLANG_FORMAT_PATH env var to style.sh

andrewheard:
Added an environment variable to specify the path to the version of `clang-format` used by the `style.sh` script.

#no-changelog

### DIFF
--- a/scripts/style.sh
+++ b/scripts/style.sh
@@ -24,14 +24,28 @@
 # Commonly
 # ./scripts/style.sh master
 
+# Set the environment variable FIR_CLANG_FORMAT_PATH to use a specific version
+# of clang-format, regardless of its order in the shell PATH.
+#
+# Example (add to ~/.bash_profile, ~/.zshrc or run in interactive shell):
+#   FIR_CLANG_FORMAT_PATH="$(brew --prefix clang-format)/bin/clang-format"
+#   export FIR_CLANG_FORMAT_PATH
+if [[ -n "$FIR_CLANG_FORMAT_PATH" ]]; then
+  clang_format_bin="$FIR_CLANG_FORMAT_PATH"
+elif ! clang_format_bin=$(command -v clang-format); then
+  echo "clang-format not found, install with 'brew install clang-format'"
+  exit 1
+fi
+
 # Strip the clang-format version output down to the major version. Examples:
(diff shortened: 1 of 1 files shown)


scripts/style.sh
```diff
@@ -24,14 +24,28 @@
...
 
+# Set the environment variable FIR_CLANG_FORMAT_PATH to use a specific version
+# of clang-format, regardless of its order in the shell PATH.
+#
+# Example (add to ~/.bash_profile, ~/.zshrc or run in interactive shell):
+#   FIR_CLANG_FORMAT_PATH="$(brew --prefix clang-format)/bin/clang-format"
+#   export FIR_CLANG_FORMAT_PATH
+if [[ -n "$FIR_CLANG_FORMAT_PATH" ]]; then
+  clang_format_bin="$FIR_CLANG_FORMAT_PATH"
+elif ! clang_format_bin=$(command -v clang-format); then
```
ncooke3, line 35: Just so IUC, this line does the following:
1. Sets the value of `clang_format_bin` , which is available in the outermost scope.
2. If the value of `clang_format_bin` is nothing, then it enters the clause and exits.

Am I correct?

andrewheard: Yes, exactly that. Unlike `which`, [`command -v <command_name>`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html) doesn't output anything to the console if it can't find `command_name` and just sets a failure exit code. This Stack Overflow [post](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script) explains it nicely too.

After this part finishes, the existing code checks that `clang_format_bin` (from `FIR_CLANG_FORMAT_PATH` or `PATH`) satisfies the version requirements.

paulb777 (approved):

scripts/style.sh
ncooke3: thanks for clarifying and cool find about `command -v`. that's good to know!

ncooke3 (approved):
this is a really great improvement!

scripts/style.sh
andrewheard: Np! It was new to me too 😄

PR #11107, opened 2023-04-11
[Tooling] Add FIR_CLANG_FORMAT_PATH env var to style.sh (#11097)

andrewheard:
Added an environment variable to specify the path to the version of clang-format used by the style.sh script.


paulb777:
I just rebased and force pushed `auth-swift` with this change.

ncooke3:
thanks @paulb777 !

PR #11121, opened 2023-04-13
[Release] Enable macCatalyst support for Firestore targets

ncooke3:
### Context
- SPM only change that was missed in #11066
- We recently bumped to enforce swift-tools-version 5.6, which enforces being explicit about macCatalyst support. #11046

After this merges, I'll create the release branch and cherry-pick this into it.


paulb777 (commented):
Thanks! Is it feasible to add CI to catch this?

ncooke3:
@paulb777, I think so. I was going to investigate after getting the patch out but I could do that now.

ncooke3:
weirdly, testing the `swift-test` scheme on catalyst does not reproduce this.

PR #11144, opened 2023-04-18
Remove references to deprecated CTCarrier for Sessions

samedson:
https://developer.apple.com/forums/thread/714876
 - Deprecated with no replacement in iOS 16.4.

Currently FirebaseSessions doesn't use this data, as it's still in development, so no customer impact.

TODO
- [x] Run tests

### DIFF
--- a/FirebaseSessions/Sources/SessionStartEvent.swift
+++ b/FirebaseSessions/Sources/SessionStartEvent.swift
@@ -130,7 +130,7 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
     switch subscriber {
     case .Performance:
       let oldString = proto.application_info.apple_app_info.mcc_mnc
-      proto.application_info.apple_app_info.mcc_mnc = makeProtoString(appInfo.mccMNC)
+      proto.application_info.apple_app_info.mcc_mnc = makeProtoString("")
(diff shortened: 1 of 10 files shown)


mrober (approved):

FirebaseSessions/Sources/SessionStartEvent.swift
```diff
@@ -109,7 +109,7 @@ class SessionStartEvent: NSObject, GDTCOREventDataObject {
                                    appInfo: ApplicationInfoProtocol) {
     switch subscriber {
     case .Performance:
-      proto.application_info.apple_app_info.mcc_mnc = makeProtoString(appInfo.mccMNC)
+      proto.application_info.apple_app_info.mcc_mnc = makeProtoString("")
```
visumickey, line 112: Do we know what happens in the Backend when there is no radio information? Hopefully that does not drop the event due to validation failure?

visumickey (approved):

FirebaseSessions/Sources/SessionStartEvent.swift
samedson: We never send a radio when just Crashlytics is installed, so this should not impact

PR #11150, opened 2023-04-19
[Infra] Add sample app to test building SPM distro across supported platforms

ncooke3:
### Context
- Add "client" sample app that imports all applicable product in a Swift, Obj-C, and Obj-C++ context.
- Add corresponding multi-platform `spm / client-app` job to test that app builds from the branch's HEAD package state.
- Checked and verified that this would have caught the issue that required 10.8.1 release.
- This also surfaced some other warnings and discrepancies that can now be addressed in subsequent PRs.
- Theres's currently a bug in Xcode versions under 14.3 that will affect this job running on some platforms. The build error looks like:
  
  >  Showing All Messages
/Users/nickcooke/Library/Developer/Xcode/DerivedData/ClientApp-ftiilcvpyjwxxieryciqmiulqnea/SourcePackages/artifacts/googleappmeasurement/GoogleAppMeasurementOnDeviceConversion.xcframework:1:1: While building for macOS, no library for this platform was found in '/Users/nickcooke/Library/Developer/Xcode/DerivedData/ClientApp-ftiilcvpyjwxxieryciqmiulqnea/SourcePackages/artifacts/googleappmeasurement/GoogleAppMeasurementOnDeviceConversion.xcframework'.

  Any idea of a workaround we could do for `GoogleAppMeasurementOnDeviceConversion` that would let us run on earlier Xcodes? I tried editing the `GoogleAppMeasurementOnDeviceConversion` to use the `condition(when: [.iOS])` but that didn't work. 
 
  The relevant fix introduced in Xcode 14.3 was:

  > Fixed: Conditional target dependencies (SE-0273) in packages are now correctly applied to binary targets and leads to top-level targets being filtered out from builds of root packages. (85762201)"

  GitHub Actions doesn't support 14.3 yet so I'm expecting some platforms like Catalyst to fail to build. I will disable those platforms and add a comment to re-enable when GHA supports Xcode 14.3.

  The only other workaround I could think of to get all platforms running on our minimum supported Xcode is to add dummy xcframework slices for `mac` and `maccatalyst` to `(Analytics|GoogleAppMeasurement)OnDeviceConversion` products. And issue a [warning like Performance](https://github.com/firebase/firebase-ios-sdk/blob/767d7c28f4d3984226d3b6177e6a6c538926f89c/SwiftPM-PlatformExclude/FirebasePerformanceWrap/include/dummy.h#L20). Thoughts? @paulb777 @andrewheard 

---

There was also another issue where ObjC++ files could not import the headers/modules of binary targets. This seems to be an Xcode 14.2 issue that was resolved in Xcode 14.3 (however I can't find corresponding release note). 

---

_Decision: Recommend 14.3 for ObjC++ use and to workaround GoogleAppMeasurement issue._

### DIFF
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -86,7 +86,10 @@ jobs:
     runs-on: macos-12
     strategy:
       matrix:
-        platform: [iOS, tvOS, macOS, catalyst]
+        platform: [iOS, tvOS]
+        # TODO(Xcode 14.3): Remove above line and uncomment below array when GHA
+        # supports Ventura/Xcode 14.3.
+        # platform: [iOS, tvOS, macOS, catalyst]
     steps:
     - uses: actions/checkout@v3
     - uses: mikehardy/buildcache-action@c87cea0ccd718971d6cc39e672c4f26815b6c126
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -81,6 +81,8 @@ jobs:
   client-app:
     # Don't run on private repo unless it is a PR.
     if: (github.repository == 'Firebase/firebase-ios-sdk' && github.event_name == 'schedule') || github.event_name == 'pull_request'
+    env:
+      FIREBASECI_USE_LATEST_GOOGLEAPPMEASUREMENT: 1
     runs-on: macos-12
     strategy:
       matrix:
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -22,61 +22,61 @@ concurrency:
     cancel-in-progress: true
 
 jobs:
-  swift-build-run:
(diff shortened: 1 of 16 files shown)


ncooke3:
Manually verified everything works on Xcode 14.3 locally. Debugging why GHA is failing... xcpretty seems to be hiding the failure.

ncooke3:
> Manually verified everything works on Xcode 14.3 locally. Debugging why GHA is failing... xcpretty seems to be hiding the failure.

Failure was related to ObjC++ issue (see description for problem and resolution)

ncooke3:
This is PR good for merge/review.

ncooke3:
Rebased on `master` to get some testing in following #11165's merge

paulb777 (approved):

PR #11193, opened 2023-04-26
Extract GACAppCheckSettings protocol from FIRAppCheckSettingsProtocol

andrewheard:
`FIRAppCheckSettings` implements the extracted `GACAppCheckSettings` protocol for Firebase App Check and other integrators must provide their own implementations.


FirebaseAppCheck/Sources/Core/GACAppCheckSettings.h
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2023 Google LLC
```
paulb777, line 2: Should GACAppCheckSettings.h be in a new folder hierarchy to better correspond with its future migration - like FirebaseAppCheck/Sources/GAC/Sources/Core/ ?

PR #11200, opened 2023-04-27
Add equality operator on aggregate query.

tom-andersen:


Firestore/core/src/remote/grpc_root_certificates_generated.cc
dconeybe, line 1: Remove `grpc_root_certificates_generated.cc` from this PR.

Firestore/Example/Tests/API/FIRAggregateQueryUnitTests.mm
dconeybe, line 1: I think the unit tests should instead be written in `Firestore/core/test/unit/api/aggregate_query_test.cc` to test the class directly rather than indirectly via the objective-c classes. Can you look into that?

dconeybe (changes requested):

.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.7
+2.7.7
```
dconeybe, line 1: Please move this change to `.ruby-version` to a different PR, unless it is somehow related to aggregate queries or is required for this PR to pass CI.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,63 @@
...
+  {
+    auto firestore = std::make_shared<Firestore>();
+    AggregateQuery aggregate_query1 =
+        Query{core::Query{model::ResourcePath{"foo"}}, firestore}.Count();
+    AggregateQuery aggregate_query2 =
+        Query{core::Query{model::ResourcePath{"foo"}}, firestore}.Count();
+    AggregateQuery aggregate_query3 =
+        Query{core::Query{model::ResourcePath{"bar"}}, firestore}.Count();
+
+    EXPECT_EQ(aggregate_query1, aggregate_query1);
```
dconeybe, line 42: Since you are explicitly testing the `operator==` and `operator!=` operator overloads, IMO it's clearer to your intention to use those operators front-and-center in the test. Namely, my suggestion is to instead write

```cpp
EXPECT_TRUE(aggregate_query1 == aggregate_query1);
```

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,63 @@
...
+    EXPECT_EQ(aggregate_query1, aggregate_query2);
```
dconeybe, line 43: Augment the "Equality" test case to also test `operator!=`.

Firestore/Example/Tests/API/FIRAggregateQueryUnitTests.mm
```diff
@@ -0,0 +1,57 @@
...
+using firebase::firestore::testutil::Query;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FIRAggregateQueryUnitTests : XCTestCase
+@end
+
+@implementation FIRAggregateQueryUnitTests
+
+- (void)testEquals {
```
dconeybe, line 42: This test method is called "testEquals" but it appears to be testing the _hash_ method. This seems inconsistent to me. Am I missing something?

dconeybe (changes requested):

Firestore/core/test/unit/api/aggregate_query_test.cc
tom-andersen: Done

Firestore/Example/Tests/API/FIRAggregateQueryUnitTests.mm
dconeybe: Ahh, yes that other test also suffers from the not-quite-correct name. Your name suggestions LGTM.

.ruby-version
dconeybe: > Please move this change to `.ruby-version` to a different PR, unless it is somehow related to aggregate queries or is required for this PR to pass CI.

Bump

dconeybe (changes requested):

Firestore/Example/Tests/API/FIRAggregateQueryUnitTests.mm
tom-andersen: This seems more pervasive. Can we leave this as is for now.

image

.ruby-version
dconeybe: Hmm. It was showing up in my diff. I probably had the wrong diff selected. Sorry about that. It's clearly done now.

dconeybe (approved):

PR #11205, opened 2023-04-28
Add os_version and app_build_version to Sessions

samedson:
#no-changelog


mrober (approved):

visumickey (approved):
These changes look good to me. Let's wait for the sessions event proto changes to land before merging this. Also - I'm guessing there will be a followup PR to use these proto fields to pack the OS version and App Build Version?

samedson:
Yes going to follow up with filling in the fields

PR #11207, opened 2023-04-28
Fix link on AddNewProduct.md

paulb777:
#no-changelog


andrewheard (approved):
LGTM! Verified that the link works for me in the [preview](https://github.com/firebase/firebase-ios-sdk/blob/paulb777-patch-1/docs/AddNewProduct.md).

PR #11208, opened 2023-04-28
[Release] Update gRPC and Abseil deps

ncooke3:
### Context
- Corresponding Abseil PR: https://github.com/google/abseil-cpp-binary/pull/3
- Corresponding GRPC PR: https://github.com/google/grpc-binary/pull/3
- SPM version of #1065
- ~Manually verified that the Firestore QS builds and loads data~

Edit: I pinned the QS to `master` instead of `nc/bump-abseil-and-grpc`. Using `nc/bump-abseil-and-grpc`, the QS fails.

#no-changelog


ncooke3:
@paulb777, oh I didn't think of this... the `spm / ClientApp` workflow [fails](https://github.com/firebase/firebase-ios-sdk/actions/runs/4833445931/jobs/8613462903?pr=11208) because it is using these new dependencies with the old Firestore binary. So bumping these deps means also updating the Firestore binary in the `Package.swift` so SPM on `master` doesn't have a broken Firestore target.

paulb777:
Hmm, that shouldn't be the case if those are really minor version updates?

ncooke3:
> Hmm, that shouldn't be the case if those are really minor version updates?

Here's what I see.
Screenshot 2023-04-28 at 2 28 53 PM

paulb777:
Where are the undefined symbol references? They should show up if you click the top right hamburger

ncooke3:
Firestore QS builds and runs with the latest commit.

paulb777 (approved):

paulb777:
Merging so it's included started in tonight's nightlies.

PR #11210, opened 2023-04-28
Private APIs for sum/avg

MarkDuckworth:
Private APIs for sum/avg so that changes can land in master


Firestore/Source/API/FIRAggregateField.mm
```diff
@@ -16,6 +16,8 @@
 
 #import "FIRAggregateField.h"
 
+#include <string>
+
```
ehsannas, line 20: not sure why this was added?

ehsannas (approved):

Firestore/Source/API/FIRAggregateField.mm
MarkDuckworth: Removed. Thank you.

PR #11214, opened 2023-05-01
Extract GACAppCheckSettings from FIRAppCheckSettings

andrewheard:
Extracted the non-Firebase-specific implementation details out of `FIRAppCheckSettings` into `GACAppCheckSettings`.

Added a `GACAppCheckTokenAutoRefreshPolicy` enum to represent the tri-state (`nil`/`YES`/`NO`) of auto token refresh enablement in `GACAppCheckSettings`.

This is an alternative approach to PR #11193 that makes `FIRAppCheckSettings` a subclass of `GACAppCheckSettings` to share common code.

#no-changelog

### DIFF
--- a/FirebaseAppCheck/Sources/Core/FIRAppCheckSettings.h
+++ b/FirebaseAppCheck/Sources/Core/FIRAppCheckSettings.h
@@ -25,20 +25,20 @@ NS_ASSUME_NONNULL_BEGIN
 FOUNDATION_EXPORT NSString *const kFIRAppCheckTokenAutoRefreshEnabledUserDefaultsPrefix;
 FOUNDATION_EXPORT NSString *const kFIRAppCheckTokenAutoRefreshEnabledInfoPlistKey;
 
-/// A collection of Firebase app check wide settings and parameters.
-@protocol FIRAppCheckSettingsProtocol <NSObject>
-
-/// If Firebase app check token auto-refresh is allowed.
-@property(nonatomic, assign) BOOL isTokenAutoRefreshEnabled;
-
-@end
-
 /// Handles storing and updating the Firebase app check wide settings and parameters.
-@interface FIRAppCheckSettings : GACAppCheckSettings <FIRAppCheckSettingsProtocol>
+@interface FIRAppCheckSettings : GACAppCheckSettings
 
 - (instancetype)initWithApp:(FIRApp *)firebaseApp
                 userDefault:(NSUserDefaults *)userDefaults
-                 mainBundle:(NSBundle *)mainBundle;
+                 mainBundle:(NSBundle *)mainBundle NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)init NS_UNAVAILABLE;
+
+- (instancetype)initWithUserDefaults:(NSUserDefaults *)userDefaults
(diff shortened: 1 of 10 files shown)


FirebaseAppCheck/Sources/Core/FIRAppCheckSettings.h
```diff
@@ -32,7 +34,7 @@ FOUNDATION_EXPORT NSString *const kFIRAppCheckTokenAutoRefreshEnabledInfoPlistKe
 @end
 
 /// Handles storing and updating the Firebase app check wide settings and parameters.
-@interface FIRAppCheckSettings : NSObject <FIRAppCheckSettingsProtocol>
+@interface FIRAppCheckSettings : GACAppCheckSettings <FIRAppCheckSettingsProtocol>
```
andrewheard, line 37: Alternatively, `FIRAppCheckSettings` could conform to `GACAppCheckSettingsProtocol` and hold an instance of `GACAppCheckSettings`, instead of subclassing from `GACAppCheckSettings`, if we want to avoid inheritance.

This would add a small amount of boilerplate since the implementation of `@property(nonatomic, assign) GACAppCheckTokenAutoRefreshPolicy tokenAutoRefreshPolicy;` would need to call through to the held `GACAppCheckSettings` instance. E.g.:
```
- (GACAppCheckTokenAutoRefreshPolicy)tokenAutoRefreshPolicy {
  return self.appCheckSettings.tokenAutoRefreshPolicy;
}

- (void)setTokenAutoRefreshPolicy:(GACAppCheckTokenAutoRefreshPolicy)tokenAutoRefreshPolicy {
  self.appCheckSettings.tokenAutoRefreshPolicy = tokenAutoRefreshPolicy;
}
```

paulb777:
The merge target should be a long-lived branch.

Both approaches look reasonable to me. How do you recommend we decide?

FirebaseAppCheck/Sources/Core/FIRAppCheckSettings.h
andrewheard: For simplicity, I'll hold off on the alternative approach above unless we find a specific need.

andrewheard:
> The merge target should be a long-lived branch.
> 
> Both approaches look reasonable to me. How do you recommend we decide?

Thanks for catching the merge target, updated it to `appcheck-lite`. My preference would be to go with this approach since it allows more code to be moved into the shared "Lite" SDK.

paulb777 (commented):
I see code related to `AutoRefreshEnabledNumber` removed from Firebase without seeing the replacement in GAC.

I'm not sure what it's actually for, so I could easily be missing something.

andrewheard:
> I see code related to `AutoRefreshEnabledNumber` removed from Firebase without seeing the replacement in GAC.
> 
> I'm not sure what it's actually for, so I could easily be missing something.

That was to hold the three states (`nil`, `@(YES)` or `@(NO)`) for the token refresh. The replacement for that is `@property(nonatomic, assign) GACAppCheckTokenAutoRefreshPolicy tokenAutoRefreshPolicy` in `GACAppCheckSettings.h`.

paulb777 (approved):

PR #11216, opened 2023-05-02
Fixing threading-related hang during initialization in urgent mode

samedson:
This is a modification of https://github.com/firebase/firebase-ios-sdk/pull/10958

Paired on with @wowlocal 

Rotating the Installation ID when uploading in Urgent Mode is not necessary, so this modifies the previous PR to skip rotating the Installation ID when we have the conditions for deadlock.

One problem with this seems to be the isMainThread check. Dispatch Queues in the background are not guaranteed to run in a background thread, so I believe the isMainThread check is true even when run on a background dispatch_queue. This seems to over-count cases where this could deadlock.

### DIFF
--- a/Crashlytics/CHANGELOG.md
+++ b/Crashlytics/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [fixed] Fixed a threading-related hang during initialization in urgent mode (#11216)
+
 # 10.10.0
 - [changed] Removed references to deprecated CTCarrier API in FirebaseSessions. (#11144)
 - [fixed] Fix Xcode 14.3 Analyzer issue. (#11228)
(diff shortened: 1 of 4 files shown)


wowlocal:
I resolved the problem that led to the SPM test failure. https://github.com/firebase/firebase-ios-sdk/pull/10958#issuecomment-1532106435

samedson:
@wowlocal this should be a good way to solve the problem.

Thank you very much for your contribution! I was able to use that unit test to reproduce the issue. After removing the fix the test fails, and now it passes.

Going to merge this once my teammate approves.

themiswang (approved):
LGTM, thank you for simplified the logic!

wowlocal:
Great fix! I'm glad the test was helpful 😃

PR #11218, opened 2023-05-02
Delete invalid move tests

wu-hui:


wu-hui:
> 

Good idea, added.

PR #11223, opened 2023-05-03
Backward Compatible Count C++ API

MarkDuckworth:
Update the C++ API so the c++ SDK will build as-is using the old Count API.

### DIFF
--- a/Firestore/core/src/api/aggregate_query.cc
+++ b/Firestore/core/src/api/aggregate_query.cc
@@ -43,13 +43,7 @@ AggregateQuery::AggregateQuery(Query query,
     : query_{std::move(query)}, aggregates_{std::move(aggregates)} {
 }
 
-AggregateQuery::AggregateQuery(Query query) : query_{std::move(query)} {
-  AggregateField countAf(AggregateField::OpKind::Count,
-                         AggregateAlias("count"));
-  aggregates_ = std::vector<AggregateField>{countAf};
-}
-
-void AggregateQuery::Get(AggregateQueryCallback&& callback) {
+void AggregateQuery::GetAggregate(AggregateQueryCallback&& callback) {
   query_.firestore()->client()->RunAggregateQuery(query_.query(), aggregates_,
                                                   std::move(callback));
 }
--- a/Firestore/core/src/api/aggregate_query.cc
+++ b/Firestore/core/src/api/aggregate_query.cc
@@ -44,25 +44,24 @@ AggregateQuery::AggregateQuery(Query query,
 }
 
 void AggregateQuery::GetAggregate(AggregateQueryCallback&& callback) {
+  if (callback) throw 1;
   query_.firestore()->client()->RunAggregateQuery(query_.query(), aggregates_,
                                                   std::move(callback));
 }
 
 void AggregateQuery::Get(CountQueryCallback&& callback) {
-  query_.firestore()->client()->RunAggregateQuery(
-      query_.query(), aggregates_,
-      [callback](const StatusOr<ObjectValue>& result) {
-        if (result.ok()) {
-          auto countValue =
-              result.ValueOrDie().Get(AggregateAlias("count").StringValue());
-          HARD_ASSERT(countValue.has_value() &&
-                      countValue.value().which_value_type ==
-                          google_firestore_v1_Value_integer_value_tag);
-          callback(StatusOr<int64_t>(countValue->integer_value));
-        } else {
-          callback(StatusOr<int64_t>(std::move(result.status())));
-        }
-      });
+  this->GetAggregate([callback](const StatusOr<ObjectValue>& result) {
+    if (result.ok()) {
+      auto countValue =
+          result.ValueOrDie().Get(AggregateAlias("count").StringValue());
+      HARD_ASSERT(countValue.has_value() &&
+                  countValue.value().which_value_type ==
+                      google_firestore_v1_Value_integer_value_tag);
+      callback(StatusOr<int64_t>(countValue->integer_value));
+    } else {
+      callback(StatusOr<int64_t>(std::move(result.status())));
+    }
+  });
 }
 
 }  // namespace api
--- a/Firestore/core/src/api/aggregate_query.cc
+++ b/Firestore/core/src/api/aggregate_query.cc
@@ -23,6 +23,7 @@
 #include "Firestore/core/src/core/firestore_client.h"
 #include "Firestore/core/src/model/aggregate_field.h"
 
+using firebase::firestore::model::AggregateAlias;
 using firebase::firestore::model::AggregateField;
 
 namespace firebase {
@@ -42,9 +43,32 @@ AggregateQuery::AggregateQuery(Query query,
     : query_{std::move(query)}, aggregates_{std::move(aggregates)} {
 }
 
+AggregateQuery::AggregateQuery(Query query) : query_{std::move(query)} {
+  AggregateField countAf(AggregateField::OpKind::Count,
+                         AggregateAlias("count"));
+  aggregates_ = std::vector<AggregateField>{countAf};
+}
+
 void AggregateQuery::Get(AggregateQueryCallback&& callback) {
   query_.firestore()->client()->RunAggregateQuery(query_.query(), aggregates_,
                                                   std::move(callback));
+}
+
+void AggregateQuery::Get(CountQueryCallback&& callback) {
+  query_.firestore()->client()->RunAggregateQuery(
+      query_.query(), aggregates_,
+      [callback](const StatusOr<ObjectValue>& result) {
+        if (result.ok()) {
+          auto countValue =
+              result.ValueOrDie().Get(AggregateAlias("count").StringValue());
+          HARD_ASSERT(countValue.has_value() &&
+                      countValue.value().which_value_type ==
+                          google_firestore_v1_Value_integer_value_tag);
+          callback(StatusOr<int64_t>(countValue->integer_value));
+        } else {
+          callback(StatusOr<int64_t>(std::move(result.status())));
+        }
+      });
 }
 
 }  // namespace api
--- a/Firestore/core/test/unit/api/aggregate_query_test.cc
+++ b/Firestore/core/test/unit/api/aggregate_query_test.cc
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "Firestore/core/src/api/query_core.h"
+#include "Firestore/core/src/util/status.h"
+#include "gtest/gtest.h"
+
+#define private public
+#include "Firestore/core/src/api/aggregate_query.h"
+
+namespace firebase {
+namespace firestore {
+
+using model::AggregateAlias;
+using util::Status;
+
+namespace api {
+
+class MockAggregateQuery : public AggregateQuery {
+ public:
+  StatusOr<ObjectValue> mockResult;
+
+  MockAggregateQuery(Query query,
+                     std::vector<AggregateField>&& aggregates,
+                     StatusOr<ObjectValue>&& mockResult)
+      : AggregateQuery(query, std::move(aggregates)),
+        mockResult(std::move(mockResult)){};
+  void GetAggregate(AggregateQueryCallback&& callback) {
+    callback(mockResult);
+  }
+};
+
+namespace {
+
+// Assert that the Get method calls GetAggregate method
+// and that the result from GetAggregate is processed
+// appropriately
+TEST(AggregateQuery, GetCallsGetAggregateOk) {
+  // Params for Get: Query and AggregateField vector
+  Query query;
+  AggregateField countAggregateField(AggregateField::OpKind::Count,
+                                     AggregateAlias("count"));
+  std::vector<AggregateField> aggregates{countAggregateField};
+
+  // Mock aggregate field result
+  google_firestore_v1_AggregationResult_AggregateFieldsEntry
+      mockAggregatesField[1];
+  mockAggregatesField[0].key =
+      nanopb::ByteString(absl::string_view("count")).release();
+  mockAggregatesField[0].value.which_value_type =
+      google_firestore_v1_Value_integer_value_tag;
+  mockAggregatesField[0].value.integer_value = 10;
+
+  // Mock ObjectValue result
+  ObjectValue mockObjectValue =
+      ObjectValue::FromAggregateFieldsEntry(mockAggregatesField, 1);
+
+  // Set mocked result on our MockAggregateQuery object
+  MockAggregateQuery aggregateQuery(query, std::move(aggregates),
+                                    StatusOr<ObjectValue>(mockObjectValue));
+
+  // Call the Get method
+  size_t callbackCount = 0;
+  aggregateQuery.Get([&callbackCount](const StatusOr<int64_t>& result) mutable {
+    callbackCount++;
+    EXPECT_EQ(result.ok(), true);
+    EXPECT_EQ(result.ValueOrDie(), 10);
+  });
+
+  // Assert
+  EXPECT_EQ(callbackCount, 1);
+}
+
+// Assert that the Get method calls GetAggregate method
+// and that an error result from GetAggregate is processed
+// appropriately
+TEST(AggregateQuery, GetCallsGetAggregateError) {
+  // Params for Get: Query and AggregateField vector
+  Query query;
+  AggregateField countAggregateField(AggregateField::OpKind::Count,
+                                     AggregateAlias("count"));
+  std::vector<AggregateField> aggregates{countAggregateField};
+
+  // Mock error status
+  Status mockError = Status(Error::kErrorInternal, "foo");
+
+  // Set mocked result on our MockAggregateQuery object
+  MockAggregateQuery aggregateQuery(query, std::move(aggregates),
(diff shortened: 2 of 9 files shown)


Firestore/core/src/api/aggregate_query.cc
```diff
@@ -42,11 +43,27 @@ AggregateQuery::AggregateQuery(Query query,
     : query_{std::move(query)}, aggregates_{std::move(aggregates)} {
 }
 
-void AggregateQuery::Get(AggregateQueryCallback&& callback) {
+void AggregateQuery::GetAggregate(AggregateQueryCallback&& callback) {
+  if (callback) throw 1;
```
dconeybe, line 47: This "throw" line looks... wrong. Is this line here by accident?

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,141 @@
...
+
+namespace firebase {
+namespace firestore {
+
+using model::AggregateAlias;
+using util::Status;
+
+namespace api {
+
+class MockAggregateQuery : public AggregateQuery {
```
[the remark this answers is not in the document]
tom-andersen: We should have a broader team discussion on this. I think we are holding back our ability to unit test without virtual methods.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -37,16 +38,31 @@ class MockAggregateQuery : public AggregateQuery {
  public:
   StatusOr<ObjectValue> mockResult;
 
+  ~MockAggregateQuery() override = default;
```
dconeybe, line 41: You don't need to re-declare the destructor in subclasses. You can just omit it if its body is empty.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+namespace firestore {
+
+using model::AggregateAlias;
+using util::Status;
+
+namespace api {
+
+class MockAggregateQuery : public AggregateQuery {
+ public:
+  StatusOr<ObjectValue> mockResult;
```
dconeybe, line 39: nit: Use underscore_separated_names instead of camelCase, here and elsewhere in this file.

Firestore/core/test/unit/api/aggregate_query_test.cc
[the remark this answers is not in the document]
dconeybe: Fair enough. Can you just add a comment then at the declaration site explain why the method is marked "virtual"?

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+  Query query;
+  AggregateField countAggregateField(AggregateField::OpKind::Count,
+                                     AggregateAlias("count"));
+  std::vector<AggregateField> aggregates{countAggregateField};
+
+  // Mock aggregate field result
+  google_firestore_v1_AggregationResult_AggregateFieldsEntry
+      mockAggregatesField[1];
+  mockAggregatesField[0].key =
+      nanopb::ByteString(absl::string_view("count")).release();
```
dconeybe, line 82: Just pass `"count"` as the argument. absl::string_view has an implicit constructor that will create an instance from a `const char*`. In fact, that's how string_view is intended to be used.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+  // Mock ObjectValue result
+  ObjectValue mockObjectValue =
+      ObjectValue::FromAggregateFieldsEntry(mockAggregatesField, 1);
+
+  // Set mocked result on our MockAggregateQuery object
+  MockAggregateQuery aggregateQuery(query, std::move(aggregates),
+                                    StatusOr<ObjectValue>(mockObjectValue));
+
+  // Call the Get method
+  size_t callbackCount = 0;
```
dconeybe, line 96: Just use `int` instead of `size_t`. The fact that `size_t` is used for sizes in the standard library is a historical artifact. See go/unsigned

dconeybe, line 96: Is the callback given to `aggregateQuery.Get()` invoked on a different thread? If so, then use `std::atomic<int>` instead of `int` for `callbackCount` to eliminate any race conditions. https://en.cppreference.com/w/cpp/atomic/atomic

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+// appropriately
+TEST(AggregateQuery, GetCallsGetAggregateError) {
+  // Params for Get: Query and AggregateField vector
+  Query query;
+  AggregateField countAggregateField(AggregateField::OpKind::Count,
+                                     AggregateAlias("count"));
+  std::vector<AggregateField> aggregates{countAggregateField};
+
+  // Mock error status
+  Status mockError = Status(Error::kErrorInternal, "foo");
```
dconeybe, line 118: Please rename `mockError` to have a different prefix. The term "mock" is specifically referring to a mock object created by gmock. In this case, it's not a mock object, it's a real object. Consider calling it "sample_error" or "simulated_error" or even just "error".

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+    EXPECT_EQ(result.ok(), false);
+    EXPECT_EQ(result.status().code(), Error::kErrorInternal);
+    EXPECT_EQ(result.status().error_message(), "foo");
+  });
+
+  // Assert
+  EXPECT_EQ(callbackCount, 1);
+}
+
+// Assert that the Query.Count method creates an AggregateQuery
```
dconeybe, line 137: nit: `Query.Count` -> `Query::Count`, as is conventional in C++

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+// with the expected query and aggregates
+TEST(Query, Count) {
+  // Baseline Query
+  Query query;
+
+  // Testing the Count() method
```
dconeybe, line 143: ultranit: the term "method" is non-idiomatic in c++; instead, use the term "member function" (and "member variable" for an "instance variable").

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+  AggregateQuery aggregateQuery = query.Count();
+
+  auto internalQuery = AggregateQueryTest::GetQuery(aggregateQuery);
+  auto internalAggregates = AggregateQueryTest::GetAggregates(aggregateQuery);
+
+  // Assert
+  EXPECT_EQ(internalQuery, query);
+  EXPECT_EQ(internalAggregates.size(), 1);
```
dconeybe, line 151: Change `EXPECT_EQ(internalAggregates.size(), 1);` to `ASSERT_EQ`; otherwise, if the vector is empty then the lines below exhibit undefined behavior.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,160 @@
...
+// Assert that the Query.Count method creates an AggregateQuery
+// with the expected query and aggregates
+TEST(Query, Count) {
+  // Baseline Query
+  Query query;
+
+  // Testing the Count() method
+  AggregateQuery aggregateQuery = query.Count();
+
+  auto internalQuery = AggregateQueryTest::GetQuery(aggregateQuery);
```
dconeybe, line 146: Please don't use `auto` here because it detracts from readability. I have no idea what these types are without somehow finding the declaration. I typically recommend to only use `auto` when the type is immediately obvious from the context, such as the return value of `std::make_shared()`.

dconeybe (changes requested):

Firestore/core/test/unit/api/aggregate_query_test.cc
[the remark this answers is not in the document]
dconeybe: After looking at the code in more detail I am more sympathetic to your approach to making the one method `virtual`. However, I'd recommend to use gmock (which is built into googletest) instead of rolling your own hand-built mock object. See https://github.com/google/googletest/tree/main/googlemock for details.

Firestore/core/src/api/aggregate_query.cc
```diff
@@ -42,11 +43,28 @@ AggregateQuery::AggregateQuery(Query query,
...
-void AggregateQuery::Get(AggregateQueryCallback&& callback) {
+void AggregateQuery::GetAggregate(AggregateQueryCallback&& callback) {
   query_.firestore()->client()->RunAggregateQuery(query_.query(), aggregates_,
                                                   std::move(callback));
 }
 
+// TODO(b/280805906) Remove this count specific API after the c++ SDK migrates
+// to the new Aggregate API
+void AggregateQuery::Get(CountQueryCallback&& callback) {
+  this->GetAggregate([callback](const StatusOr<ObjectValue>& result) {
```
dconeybe, line 54: You should _move_ `callback` into the lambda. This is a cool new thing we can do now that we've "gotten with the times" and upgraded to c++14 ;)

Namely, something like this:

```cpp
this->GetAggregate([callback = std::move(callback)](const StatusOr<ObjectValue>& result) { ... }
```

Firestore/core/src/api/aggregate_query.cc
```diff
@@ -42,11 +43,28 @@ AggregateQuery::AggregateQuery(Query query,
...
+    if (result.ok()) {
+      auto count_value =
+          result.ValueOrDie().Get(AggregateAlias("count").StringValue());
+      HARD_ASSERT(count_value.has_value() &&
+                  count_value.value().which_value_type ==
+                      google_firestore_v1_Value_integer_value_tag);
+      callback(StatusOr<int64_t>(count_value->integer_value));
+    } else {
+      callback(StatusOr<int64_t>(std::move(result.status())));
```
dconeybe, line 63: Optional: Consider changing the structure of this method to reduce the amount of nesting, to improve readability. I'd recommend to change `if (result.ok())` to `if (!result.ok())` and in its block call the callback and return. The rest of the function's body can then be un-indented, since that is the "typical" path.

Firestore/core/src/api/aggregate_query.cc
```diff
@@ -42,11 +43,28 @@ AggregateQuery::AggregateQuery(Query query,
...
   query_.firestore()->client()->RunAggregateQuery(query_.query(), aggregates_,
                                                   std::move(callback));
 }
 
+// TODO(b/280805906) Remove this count specific API after the c++ SDK migrates
+// to the new Aggregate API
+void AggregateQuery::Get(CountQueryCallback&& callback) {
+  this->GetAggregate([callback](const StatusOr<ObjectValue>& result) {
+    if (result.ok()) {
+      auto count_value =
```
dconeybe, line 56: Also, here, please replace `auto` with the actual type, to improve readability.

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: The c++ style guide says to annotate overrides of virtual functions and destructors with "override" or "final". See Classes > Inheritance. Since the default destructor is an override of the parent virtual destructor, then depending on interpretation, it needs to be annotated with "override" in this re-declaration. Thoughts?

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: Because the response from Firestore is mocked, the callback is executed in the same thread. I appreciate the insight though.

Firestore/core/test/unit/api/aggregate_query_test.cc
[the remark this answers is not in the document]
MarkDuckworth: Comment added to the declaration site about use and future removal of `virtual`.

Thanks for pointing out gmock. It's worth using and learning, even if it make this code more complex.

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: I just removed this so it doesn't prolong this PR.

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: learned something new

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: Changed this even though I disagree. Mock is a concept that extends outside of gmock. In this case, the object is not mocking out an interface with fake behavior, the object does mock a status result from a lower layer.

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: Thanks

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: Thanks! much better.

Firestore/core/src/api/aggregate_query.cc
MarkDuckworth: new and improved! Thanks

Firestore/core/src/api/aggregate_query.cc
MarkDuckworth: I'm not a fan of that pattern when it requires an additional `return` statement because the code becomes more fragile.  A missed return statement could still show expected behavior but fall through and execute unexpected code. In a scenario not requiring the extra return, I agree.

Firestore/core/src/api/aggregate_query.cc
MarkDuckworth: updated this, but I do feel like using `auto` is easier to read. The type here is pretty cumbersome to read.

dconeybe (changes requested):

Firestore/core/src/api/aggregate_query.cc
```diff
@@ -51,10 +51,12 @@ void AggregateQuery::GetAggregate(AggregateQueryCallback&& callback) {
 // TODO(b/280805906) Remove this count specific API after the c++ SDK migrates
 // to the new Aggregate API
 void AggregateQuery::Get(CountQueryCallback&& callback) {
-  this->GetAggregate([callback](const StatusOr<ObjectValue>& result) {
+  this->GetAggregate([callback = std::move(callback)](
+                         const StatusOr<ObjectValue>& result) {
     if (result.ok()) {
-      auto count_value =
-          result.ValueOrDie().Get(AggregateAlias("count").StringValue());
+      absl::optional<firebase::firestore::_google_firestore_v1_Value>
```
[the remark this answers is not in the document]
MarkDuckworth: oh yeah, forgot about that convention

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: noted

Firestore/core/src/api/aggregate_query.cc
MarkDuckworth: modified this. After reading a few of the linked articles, I see the point of view.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,171 @@
...
+
+using model::AggregateAlias;
+using ::testing::Invoke;
+using util::Status;
+
+namespace api {
+
+class MockAggregateQuery : public AggregateQuery {
+ public:
+  MockAggregateQuery(Query query, std::vector<AggregateField>&& aggregates)
```
dconeybe, line 41: Since the `MockAggregateQuery` is merely delegating to the constructor of the superclass, you can just inherit the superclass' constructors by specifying

```
using AggregateQuery::AggregateQuery;
```

See https://abseil.io/tips/74.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  // Error result
+  Status error_result = Status(Error::kErrorInternal, "foo");
+
+  // Create an AggregateQuery with mocked GetAggregate function that
+  // invokes the callback with the error status from above
+  Query query;
+  AggregateField count_aggregate_field(AggregateField::OpKind::Count,
+                                       AggregateAlias("count"));
+  std::vector<AggregateField> aggregates{count_aggregate_field};
+  MockAggregateQuery mock_aggregate_query(query, std::move(aggregates));
```
dconeybe, line 126: nit: Consider replacing `query` here with `{}`, as was done on line 92 above.

Firestore/core/test/unit/api/aggregate_query_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  EXPECT_CALL(mock_aggregate_query, GetAggregate)
+      .Times(1)
+      .WillOnce(Invoke(
+          [error_result = std::move(error_result)](
```
dconeybe, line 130: Optional: Rather than capturing `error_result`, why not just create the object in the lambda's body? IMO it's more readable that way because the logic is simpler and you don't have to scroll up 22 lines to find the definition of `error_result`.

dconeybe (changes requested):
Looks good! Just a few more nitpicky comments after a final review.

Firestore/core/test/unit/api/aggregate_query_test.cc
MarkDuckworth: I understand your point of view but that would break structure of act-arrange-assert. go/unit-test-practices#structure

dconeybe: How so? Both the definition of `error_result` and `EXPECT_CALL` are part of the "arrange" block, no?

dconeybe: (for the record, I fully support the arrange-act-assert style of test structure)

MarkDuckworth: I guess that's true that expect is part of arrange, but the pattern of setting up the test doubles at the top of the test is used in this file. Changing that here would be inconsistent. Changing that everywhere would make the code hard to read.

dconeybe (approved):

PR #11184, opened 2023-04-25
Api Diff Report Tools

sunmou99:
Tools that will be used to generate API Diff Report
1. `icore_module.py`: tool that retrieve necessary modules information 
2. `theme` dir: `Jazzy` documentation theme. in case that Jazzy changes it's default theme.
3. `api_info.py`: tool that use `Jazzy` and generate module API information in JSON format
4. `api_diff_report`: tool that detects API diff between two versions, and generate plain text/markdown format report. 

Not include:
1. a tool that could create/modify(including hiding) commentson PR
2. a workflow that automate the API Diff Report generation.

PR #11234, opened 2023-05-05
Revert a PR causing CI breakage

paulb777:
Revert #11184.

It has whitespace and python errors that fail the check CI phase.


sunmou99 (approved):
LGTM

PR #11237, opened 2023-05-07
[auth-swift] Auth.swift

paulb777:
Replace FIRAuth.m and FIRAuth.h  with Auth.swift. 

All unit tests pass, but there are several TODOs:

- Protocols for Auth class
- UIApplicationDelegate methods
- notification for retry
- skipped and unimplemented unit tests

PR #11238, opened 2023-05-07
[auth-swift] AuthUseUserAccessGroupTests.swift

paulb777:
AuthUseUserAccessGroupTests.swift to Swift since the ObjC version depended on mocking


FirebaseAuth/Sources/Swift/SystemService/AuthStoredUserManager.swift
```diff
@@ -36,14 +36,17 @@ import Foundation
   /// The key to encode and decode the stored user.
   private static let storedUserCoderKey = "firebase_auth_stored_user_coder_key"
 
+  // TODO: Should keychainServices be AuthStorage
```
ncooke3, line 39: I'd say the answer depends on what kind of test coverage we want for this class. It'd be convenient and make things more consistent but it would require some extra work that might not be worth it. If you take a look at the API of `AuthSharedKeychainServices`, it doesn't easily conform to the protocol API of `AuthStorage`. We'd need to rework this class and the clients that call it. 

Do the class's unit tests reach codepaths that touch the keychain?

ncooke3 (approved):

FirebaseAuth/Sources/Swift/SystemService/AuthStoredUserManager.swift
ncooke3: got it, thanks. yeah, I think disabling for now sgtm.

paulb777:
OK. I'll merge to [pb-auth.swift](https://github.com/firebase/firebase-ios-sdk/tree/pb-auth.swift) which is being tracked in #11237

PR #11248, opened 2023-05-08
Add missing unowned specifiers

paulb777:
Add a `weak` specifier that was missed in the Swift transition. [Edit: `unowned` instead of `weak`. See discussion below.]

Testing seems to show that this causes a retain cycle leak for ObjC clients. The Swift runtime seems to detect the cycle and avoid the leak, even without this change.

### DIFF
--- a/FirebaseStorage/Sources/StorageComponent.swift
+++ b/FirebaseStorage/Sources/StorageComponent.swift
@@ -30,7 +30,7 @@ protocol StorageProvider {
 
   /// The app associated with all Storage instances in this container.
   /// This is `unowned` instead of `weak` so it can be used without unwrapping in `storage(...)`
-  private unowned var app: FirebaseApp
+  private unowned let app: FirebaseApp
 
   /// A map of active instances, grouped by app. Keys are FirebaseApp names and values are arrays
   /// containing all instances of Storage associated with the given app.
--- a/FirebaseFunctions/Sources/Internal/FunctionsComponent.swift
+++ b/FirebaseFunctions/Sources/Internal/FunctionsComponent.swift
@@ -37,7 +37,7 @@ protocol FunctionsProvider {
 
   /// The app associated with all functions instances in this container.
   /// This is `unowned` instead of `weak` so it can be used without unwrapping in `functions()`
-  private unowned var app: FirebaseApp
+  private unowned let app: FirebaseApp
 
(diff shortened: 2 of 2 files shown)


ncooke3 (approved):
nice! curious, how did this originally surface? via the analyzer?

paulb777:
I discovered it working on the Auth Swift implementation. See https://github.com/firebase/firebase-ios-sdk/pull/11246

FirebaseFunctions/Tests/Unit/SerializerTests.swift
```diff
@@ -198,7 +198,7 @@ class SerializerTests: XCTestCase {
       1 as Int32,
       "two",
       [3 as Int32, ["@type": "type.googleapis.com/google.protobuf.Int64Value",
-                    "value": "9876543210"]],
+                    "value": "9876543210"]] as [Any],
```
ryanwilson, line 201: Are this and the other similar changes related to the weak specifiers or separate?

FirebaseStorage/Sources/StorageComponent.swift
```diff
@@ -63,7 +63,7 @@ protocol StorageProvider {
 
   // MARK: - StorageProvider conformance
 
-  func storage(for bucket: String) -> Storage {
+  func storage(for bucket: String, app: FirebaseApp) -> Storage {
```
ryanwilson, line 66: There's a 1:1 relationship between a `StorageComponent` instance and `FirebaseApp` instance, so this shouldn't be necessary. as far as I can tell.

ryanwilson (changes requested):

FirebaseFunctions/Tests/Unit/SerializerTests.swift
paulb777: This is a new Xcode 14.3 warning I noticed.

FirebaseStorage/Sources/StorageComponent.swift
paulb777: This saves a force unwrap on the now optional `app`.

ryanwilson: Hmm, would `unowned` work in this case then?

paulb777: Good call! Looking promising in #11263

ryanwilson (approved):

PR #11259, opened 2023-05-09
Save activated ExperimentsDescriptions in DB

qdpham13:
- Save activated ExperimentsDescriptions in Experiments table
- Call updateActivatedExperiments method in experiments activate method
- Load activated ExperimentsDescriptions along with other experiment meta
- bug/263139421


FirebaseRemoteConfig/Tests/Unit/RCNConfigDBManagerTest.m
```diff
@@ -504,6 +504,45 @@ - (void)testWriteAndLoadExperiments {
...
+  XCTestExpectation *updateAndLoadExperimentExpectation =
+      [self expectationWithDescription:@"Update and load experiment in database successfully"];
+
+  NSError *error;
+  NSArray *payload2 = @[ @"ab", @"cd" ];
+  NSData *payloadData2 = [NSJSONSerialization dataWithJSONObject:payload2
+                                                         options:NSJSONWritingPrettyPrinted
+                                                           error:&error];
+  NSDictionary *payload3 =
+      @{@"experiment_ID" : @"35667", @"experiment_activate_name" : @"activate_game"};
```
amanda-xia, line 517: New to iOS code - Are the different `payload` variables here supposed to represent different experiments? I don't recognize the structure / what ab" "cd" and "experiment_activate_name" are.

qdpham13: Yeah it's to represent different ExperimentDescriptions. This is how other payloads are written too. I don't think the actual JSON contents is too important in this test b/c it just makes sure it can read and write to the DB.

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
```diff
@@ -32,7 +32,9 @@ @interface RCNConfigExperiment ()
     NSMutableArray<NSData *> *experimentPayloads;  ///< Experiment payloads.
 @property(nonatomic, strong)
     NSMutableDictionary<NSString *, id> *experimentMetadata;  ///< Experiment metadata
-@property(nonatomic, strong) RCNConfigDBManager *DBManager;   ///< Database Manager.
+@property(nonatomic, strong)
+    NSMutableArray<NSData *> *activatedExperimentPayloads;   ///< Activated experiment payloads.
```
danasilver, line 36: Nit: It'd be helpful to use either "active" and "activated" consistently

danasilver: And does this need to be initialized in `initWithDBManager:` similar to `_experimentPayloads = [[NSMutableArray alloc] init];`?

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
```diff
@@ -157,6 +179,18 @@ - (void)updateExperimentStartTime {
                          completionHandler:nil];
 }
 
+- (void)updateActiveExperiments {
```
danasilver, line 182: Nit: Might call this `updateActiveExperimentsInDb` or something to make it clear this is not related to updating active experiments in the Analytics SDK (I think those methods have similar names)

danasilver (commented):
Trying to grok the structure of experiments in the database...

```
experiments table
---
key: experiment_payload          value: experiment a
key: experiment_payload          value: experiment b
key: experiment_active_payload   value: experiment a
key: experiment_active_payload   value: experiment c
```

Does that match your understanding?

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
qdpham13: Switched to only "active". And I think `loadExperimentsFromTable` will always init the experiments variables even if it doesn't have that key in the table. But I added the explicit init to be in line with the other vars

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
```diff
@@ -131,6 +150,9 @@ - (void)updateExperimentsWithHandler:(void (^)(NSError *_Nullable))handler {
                            lastStartTime:lastStartTime
                                 payloads:_experimentPayloads
                        completionHandler:handler];
+
+  /// Update activated experiments payload and metadata in DB.
+  [self updateActiveExperiments];
```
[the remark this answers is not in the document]
qdpham13: Yeah the method that it's in is called with every activated. https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L339

FirebaseRemoteConfig/Sources/RCNConfigDBManager.m
```diff
@@ -827,12 +827,20 @@ - (void)loadExperimentWithCompletionHandler:(RCNDBCompletion)handler {
...
+      activatedExperimentPayloads = [[NSMutableArray alloc] init];
+    }
+
     if (handler) {
       dispatch_async(dispatch_get_main_queue(), ^{
         handler(
             YES, @{
               @RCNExperimentTableKeyPayload : [experimentPayloads copy],
-              @RCNExperimentTableKeyMetadata : [experimentMetadata copy]
+              @RCNExperimentTableKeyMetadata : [experimentMetadata copy],
```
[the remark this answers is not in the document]
qdpham13: Updated

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
qdpham13: Updated

qdpham13:
> Trying to grok the structure of experiments in the database...
> 
> ```
> experiments table
> ---
> key: experiment_payload          value: experiment a
> key: experiment_payload          value: experiment b
> key: experiment_active_payload   value: experiment a
> key: experiment_active_payload   value: experiment c
> ```
> 
> Does that match your understanding?

Yeah, that's how it works in iOS. One table and keys that differentiate the data

FirebaseRemoteConfig/Sources/RCNConfigExperiment.m
```diff
@@ -131,6 +151,9 @@ - (void)updateExperimentsWithHandler:(void (^)(NSError *_Nullable))handler {
                            lastStartTime:lastStartTime
                                 payloads:_experimentPayloads
```
karenyz, line 152: so alternatively we could call `updateActiveExperimentsInDB` before this block, then pass `_activeExperimentPayloads` here? 

do you think that would be better? right now `_activeExperimentPayloads` isn't used anywhere iiuc (though i guess we will use it once we add diff-ing for realtime

qdpham13: We could do that but I think it's fine the way it is. We plan on adding diff-ing for real-time right after this anyways and real-time will be the main user of `_activeExperimentPayloads`

karenyz: ok sg

karenyz (approved):

PR #11270, opened 2023-05-10
Firebase App Check One-time Use Tokens for Callable Functions SDK

aiwenisevan:
### Discussion

  * Implement one-time use token for Callable Functions SDK by following the [proposal](https://docs.google.com/document/d/1fZ8w6lIj5rODUQYqr4IUUUodCmcb_aWLS9w2LgLKgdU/edit?resourcekey=0-xnc7HhmtcKf8JPNdGKtCIg#heading=h.ptb4ebwum9vl)

### Testing

  - [x] Added unit tests

### API Changes

  * [proposal ](https://docs.google.com/document/d/1fZ8w6lIj5rODUQYqr4IUUUodCmcb_aWLS9w2LgLKgdU/edit?resourcekey=0-xnc7HhmtcKf8JPNdGKtCIg#heading=h.ptb4ebwum9vl) is waiting for approval

### DIFF
--- a/FirebaseAppCheck/Interop/FIRAppCheckInterop.h
+++ b/FirebaseAppCheck/Interop/FIRAppCheckInterop.h
@@ -31,10 +31,6 @@ NS_SWIFT_NAME(AppCheckInterop) @protocol FIRAppCheckInterop
                     completion:(FIRAppCheckTokenHandlerInterop)handler
     NS_SWIFT_NAME(getToken(forcingRefresh:completion:));
 
-/// Retrieve a new limited-use Firebase App Check token
-- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler
-    NS_SWIFT_NAME(getLimitedUseToken(completion:));
-
 /// A notification with the specified name is sent to the default notification center
 /// (`NotificationCenter.default`) each time a Firebase app check token is refreshed.
 /// The user info dictionary contains `-[self notificationTokenKey]` and
@@ -46,6 +42,14 @@ NS_SWIFT_NAME(AppCheckInterop) @protocol FIRAppCheckInterop
 /// `userInfo` key for the `FirebaseApp.name` in a notification for
 /// `tokenDidChangeNotificationName`.
 - (NSString *)notificationAppNameKey;
+
+// MARK: - Optional API
+
+@optional
+
+/// Retrieve a new limited-use Firebase App Check token
+- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler
+    NS_SWIFT_NAME(getLimitedUseToken(completion:));
 
 @end
 
--- a/FirebaseAppCheck/Interop/FIRAppCheckInterop.h
+++ b/FirebaseAppCheck/Interop/FIRAppCheckInterop.h
@@ -31,6 +31,10 @@ NS_SWIFT_NAME(AppCheckInterop) @protocol FIRAppCheckInterop
                     completion:(FIRAppCheckTokenHandlerInterop)handler
     NS_SWIFT_NAME(getToken(forcingRefresh:completion:));
 
+/// Retrieve a new limited-use Firebase App Check token
+- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler
+    NS_SWIFT_NAME(getLimitedUseToken(completion:));
+
 /// A notification with the specified name is sent to the default notification center
 /// (`NotificationCenter.default`) each time a Firebase app check token is refreshed.
 /// The user info dictionary contains `-[self notificationTokenKey]` and
--- a/FirebaseAppCheck/Sources/Core/FIRAppCheck.m
+++ b/FirebaseAppCheck/Sources/Core/FIRAppCheck.m
@@ -72,7 +72,6 @@ static NSString *const kDummyFACTokenValue = @"eyJlcnJvciI6IlVOS05PV05fRVJST1Iif
 @property(nonatomic, readonly, nullable) id<FIRAppCheckTokenRefresherProtocol> tokenRefresher;
 
 @property(nonatomic, nullable) FBLPromise<FIRAppCheckToken *> *ongoingRetrieveOrRefreshTokenPromise;
-@property(nonatomic, nullable) FBLPromise<FIRAppCheckToken *> *ongoingLimitedUseTokenPromise;
 @end
 
 @implementation FIRAppCheck
@@ -180,7 +179,7 @@ static NSString *const kDummyFACTokenValue = @"eyJlcnJvciI6IlVOS05PV05fRVJST1Iif
 
 - (void)limitedUseTokenWithCompletion:(void (^)(FIRAppCheckToken *_Nullable token,
                                                 NSError *_Nullable error))handler {
-  [self retrieveLimitedUseToken]
+  [self limitedUseToken]
       .then(^id _Nullable(FIRAppCheckToken *token) {
         handler(token, nil);
         return token;
@@ -235,7 +234,7 @@ static NSString *const kDummyFACTokenValue = @"eyJlcnJvciI6IlVOS05PV05fRVJST1Iif
 }
 
 - (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler {
-  [self retrieveLimitedUseToken]
+  [self limitedUseToken]
       .then(^id _Nullable(FIRAppCheckToken *token) {
         FIRAppCheckTokenResult *result = [[FIRAppCheckTokenResult alloc] initWithToken:token.token
                                                                                  error:nil];
@@ -312,26 +311,6 @@ static NSString *const kDummyFACTokenValue = @"eyJlcnJvciI6IlVOS05PV05fRVJST1Iif
 
     return token;
   });
-}
-
-- (FBLPromise<FIRAppCheckToken *> *)retrieveLimitedUseToken {
-  return [FBLPromise do:^id _Nullable {
-    if (self.ongoingLimitedUseTokenPromise == nil) {
-      // Kick off a new operation only when there is not an ongoing one.
-      self.ongoingLimitedUseTokenPromise =
-          [self limitedUseToken]
-              // Release the ongoing operation promise on completion.
-              .then(^FIRAppCheckToken *(FIRAppCheckToken *token) {
-                self.ongoingLimitedUseTokenPromise = nil;
-                return token;
-              })
-              .recover(^NSError *(NSError *error) {
-                self.ongoingLimitedUseTokenPromise = nil;
-                return error;
-              });
-    }
-    return self.ongoingLimitedUseTokenPromise;
-  }];
 }
 
 - (FBLPromise<FIRAppCheckToken *> *)refreshToken {
--- a/FirebaseAppCheck/Sources/Core/FIRAppCheck.m
+++ b/FirebaseAppCheck/Sources/Core/FIRAppCheck.m
@@ -234,6 +234,21 @@ static NSString *const kDummyFACTokenValue = @"eyJlcnJvciI6IlVOS05PV05fRVJST1Iif
       });
 }
 
+- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler {
+  [self retrieveLimitedUseToken]
+      .then(^id _Nullable(FIRAppCheckToken *token) {
+        FIRAppCheckTokenResult *result = [[FIRAppCheckTokenResult alloc] initWithToken:token.token
+                                                                                 error:nil];
+        handler(result);
+        return result;
+      })
+      .catch(^(NSError *_Nonnull error) {
+        FIRAppCheckTokenResult *result =
+            [[FIRAppCheckTokenResult alloc] initWithToken:kDummyFACTokenValue error:error];
+        handler(result);
+      });
+}
+
 - (nonnull NSString *)tokenDidChangeNotificationName {
   return FIRAppCheckAppCheckTokenDidChangeNotification;
 }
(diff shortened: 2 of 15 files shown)


ncooke3:
I've since updated the proposal with the latest revision. Introducing a new public `HTTPSCallableOptions` class is now part of the proposal.

FirebaseAppCheck/Interop/FIRAppCheckInterop.h
```diff
@@ -31,6 +31,10 @@ NS_SWIFT_NAME(AppCheckInterop) @protocol FIRAppCheckInterop
                     completion:(FIRAppCheckTokenHandlerInterop)handler
     NS_SWIFT_NAME(getToken(forcingRefresh:completion:));
 
+/// Retrieve a new limited-use Firebase App Check token
+- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler
+    NS_SWIFT_NAME(getLimitedUseToken(completion:));
+
```
ryanwilson, line 37: We haven't talked about this before but I think it's worth discussing: should this be marked as `optional` so if an older version of the SDK pulls this newer dependency in it doesn't fail? I'm not 100% convinced, but wanted to call it out (if someone had `FirebaseAppCheck` locked to a certain version, the interop library in CocoaPods could potentially be pulled in at a higher version?)

paulb777: I believe it must be optional. Otherwise it will be a breaking change.

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -139,11 +142,19 @@ internal enum FunctionsConstants {
    * Creates a reference to the Callable HTTPS trigger with the given name.
    * - Parameter name The name of the Callable HTTPS trigger.
    */
-  @objc(HTTPSCallableWithName:) open func httpsCallable(_ name: String) -> HTTPSCallable {
+  @objc(HTTPSCallableWithName:useLimitedUseAppCheckToken:) open func httpsCallable(_ name: String,
```
paulb777, line 145: This is an ObjC breaking change.  The old API must be kept for Objective C at least.

FirebaseFunctions/Tests/ObjCIntegration/ObjCAPITests.m
```diff
@@ -34,8 +34,8 @@ - (void)apis {
   func = [FIRFunctions functionsForApp:app customDomain:@"my-domain"];
 
   NSURL *url = [NSURL URLWithString:@"http://localhost:5050/project/location/name"];
-  FIRHTTPSCallable *callable = [func HTTPSCallableWithURL:url];
```
paulb777, line 37: These two lines should continue to compile.

FirebaseAppCheck/Interop/FIRAppCheckInterop.h
ncooke3: So to summarize:
- For backwards compatibility, this interop API needs to be optional 
- For forwards compatibility, the Functions podspec needs to set the lower bound of it's dependency on AppCheckInterop to be the next releasing version

Sound correct, @ryanwilson, @paulb777?

FirebaseFunctions/Tests/ObjCIntegration/ObjCAPITests.m
ncooke3: Fixed in latest batch of commits.

FirebaseAppCheck/Interop/FIRAppCheckInterop.h
ncooke3: Marked as `@optional` and also updated Function podspec to pin to AppCheckInterop `10.10`.

FirebaseFunctions/Sources/Functions.swift
ncooke3: Reverted this break and made the API change strictly additive.

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -139,11 +142,19 @@ internal enum FunctionsConstants {
...
+                                                                                   useLimitedUseAppCheckToken: Bool =
+                                                                                     false)
+    -> HTTPSCallable {
+    self.useLimitedUseAppCheckToken = useLimitedUseAppCheckToken
     return HTTPSCallable(functions: self, name: name)
   }
 
-  @objc(HTTPSCallableWithURL:) open func httpsCallable(_ url: URL) -> HTTPSCallable {
+  @objc(HTTPSCallableWithURL:useLimitedUseAppCheckToken:) open func httpsCallable(_ url: URL,
```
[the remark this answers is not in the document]
ncooke3: Reverted this break and made the API change strictly additive.

FirebaseAppCheck/Interop/FIRAppCheckInterop.h
```diff
@@ -43,6 +43,14 @@ NS_SWIFT_NAME(AppCheckInterop) @protocol FIRAppCheckInterop
 /// `tokenDidChangeNotificationName`.
 - (NSString *)notificationAppNameKey;
 
+// MARK: - Optional API
```
paulb777, line 46: Add TODO/issue to make required in 11.0?

ncooke3: great idea.

ncooke3: Opened #11284 to track.

FirebaseFunctions/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 10.10.0
 - [fixed] Fixed potential memory leak of Functions instances. (#11248)
+- [added] Callable functions can now opt in to using limited use App Check
```
andrewheard, line 3: super-nit: hyphenate "limited-use"
```suggestion
- [added] Callable functions can now opt in to using limited-use App Check
```

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -135,25 +135,50 @@ internal enum FunctionsConstants {
...
-  /// - Parameter requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
-  /// - Parameter responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
-  /// - Parameter encoder: The encoder instance to use to run the encoding.
-  /// - Parameter decoder: The decoder instance to use to run the decoding.
+  /// - Parameters:
+  ///   - name: The name of the Callable HTTPS trigger
+  ///   - requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  ///   - responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
+  ///   - encoder: The encoder instance to use to run the encoding.
+  ///   - decoder: The decoder instance to use to run the decoding.
```
andrewheard, line 181: super-nit/optional: "perform" instead of "run the"
```suggestion
  ///   - encoder: The encoder instance to use to perform encoding.
  ///   - decoder: The decoder instance to use to perform decoding.
```

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -164,16 +189,48 @@ internal enum FunctionsConstants {
...
-  /// - Parameter responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
-  /// - Parameter encoder: The encoder instance to use to run the encoding.
-  /// - Parameter decoder: The decoder instance to use to run the decoding.
+  /// - Parameters:
+  ///   - name: The name of the Callable HTTPS trigger
+  ///   - options: The options with which to customize the Callable HTTPS trigger.
+  ///   - requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  ///   - responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
+  ///   - encoder: The encoder instance to use to run the encoding.
+  ///   - decoder: The decoder instance to use to run the decoding.
```
andrewheard, line 207: super-nit/optional: "perform" instead of "run the"
```suggestion
  ///   - encoder: The encoder instance to use to perform encoding.
  ///   - decoder: The decoder instance to use to perform decoding.
```

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -164,16 +189,48 @@ internal enum FunctionsConstants {
...
+  }
+
+  /// Creates a reference to the Callable HTTPS trigger with the given name, the type of an `Encodable`
+  /// request and the type of a `Decodable` response.
+  /// - Parameters:
+  ///   - url: The url of the Callable HTTPS trigger
+  ///   - requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  ///   - responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
+  ///   - encoder: The encoder instance to use to run the encoding.
+  ///   - decoder: The decoder instance to use to run the decoding.
```
andrewheard, line 233:
```suggestion
  ///   - encoder: The encoder instance to use to perform encoding.
  ///   - decoder: The decoder instance to use to perform decoding.
```

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -184,7 +241,38 @@ internal enum FunctionsConstants {
...
+
+  /// Creates a reference to the Callable HTTPS trigger with the given name, the type of an `Encodable`
+  /// request and the type of a `Decodable` response.
+  /// - Parameters:
+  ///   - url: The url of the Callable HTTPS trigger
+  ///   - options: The options with which to customize the Callable HTTPS trigger.
+  ///   - requestAs: The type of the `Encodable` entity to use for requests to this `Callable`
+  ///   - responseAs: The type of the `Decodable` entity to use for responses from this `Callable`
+  ///   - encoder: The encoder instance to use to run the encoding.
+  ///   - decoder: The decoder instance to use to run the decoding.
```
andrewheard, line 259:
```suggestion
  ///   - encoder: The encoder instance to use to perform encoding.
  ///   - decoder: The decoder instance to use to perform decoding.
```

FirebaseFunctions/Sources/HTTPSCallableOptions.swift
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Configuration options for a ``HTTPSCallable`` instance.
+@objc(FIRHTTPSCallableOptions) public class HTTPSCallableOptions: NSObject {
+  /// Whether or not to protect the callable function with a limited use App Check token.
```
andrewheard, line 19:
```suggestion
  /// Whether or not to protect the callable function with a limited-use App Check token.
```

andrewheard (approved):
LGTM! My suggestions are all extremely nitpicky, feel free to ignore.

FirebaseAppCheck/Sources/Core/FIRAppCheck.m
```diff
@@ -234,6 +234,21 @@ - (void)getTokenForcingRefresh:(BOOL)forcingRefresh
       });
 }
 
+- (void)getLimitedUseTokenWithCompletion:(FIRAppCheckTokenHandlerInterop)handler {
```
weixifan, line 237: I may have discovered a bug in the existing `retrieveLimitedUseToken` method below. Right now it tries to see if there is an ongoing promise and waits for that if so. However, the contract for this method is that *every invocation of this method returns a fresh token*, but the existing behaviour makes it so that it's possible for multiple invocations to receive the same one-time-use token, which would be problematic. Instead, no state should be shared between invocations, and separate invocations should each have its own isolated waiting promises. I was wondering whether this can be fixed as part of this PR.

FirebaseFunctions/Sources/Functions.swift
```diff
@@ -353,8 +446,18 @@ internal enum FunctionsConstants {
       fetcher.setRequestValue(fcmToken, forHTTPHeaderField: Constants.fcmTokenHeader)
     }
 
-    if let appCheckToken = context.appCheckToken {
-      fetcher.setRequestValue(appCheckToken, forHTTPHeaderField: Constants.appCheckTokenHeader)
+    if options?.requireLimitedUseAppCheckTokens == true {
+      if let appCheckToken = context.limitedUseAppCheckToken {
```
weixifan, line 450: Does this syntax mean that state is somehow cached or stored in the Functions context? That's undesirable since we plan to make this token very short-lived in the future, and furthermore, every single Functions invocation (with the `requiredLimitedUseAppCheckTokens` options property set) must send a freshly obtained token (which will be one-time-use only) directly from the result of the network call.

FirebaseAppCheck/Sources/Core/FIRAppCheck.m
ncooke3: great catch Victor– thanks for the close review here.

FirebaseFunctions/Sources/Functions.swift
ncooke3: Good question the _context_ is not cached.

ncooke3: A new [context](https://github.com/firebase/firebase-ios-sdk/blob/7534914b6e9b0c8b709e3626b4a911e21c27eaa5/FirebaseFunctions/Sources/Internal/FunctionsContext.swift#L51) is used [per](https://github.com/firebase/firebase-ios-sdk/blob/7534914b6e9b0c8b709e3626b4a911e21c27eaa5/FirebaseFunctions/Sources/Functions.swift#L279) call.

FirebaseAppCheck/Sources/Core/FIRAppCheck.m
andrewheard: Thanks, Victor! PR #11298 has now been merged into this branch.

FirebaseFunctions/Sources/Functions.swift
ncooke3: Resolving to merge.

FirebaseAppCheck/Interop/FIRAppCheckInterop.h
ncooke3: Resolving to merge.

PR #11275, opened 2023-05-11
Create API Diff Report only when the first commit in PR has an API change

sunmou99:
Won't create Diff Report until the first commit has an API change.
If the API change got reverted in the following commit, the Diff Report will update with title "No API diff detected".


paulb777 (approved):
Thanks!

PR #11276, opened 2023-05-11
Patch LevelDB so Firestore can support international paths on Windows

jonsimantov:
LevelDB has a known issue on Windows - if the database path contains international characters (anything above 0xFF), the Windows calls to CreateFileA, DeleteFileA, etc. will fail, and the app will crash.

This PR contains a patch to leveldb that updates the calls to more modern CreateFileW, DeleteFileW, etc. This fixes a crash in the C++ and Unity SDKs on Windows when the user's home directory contains international characters: https://github.com/firebase/firebase-unity-sdk/issues/351

(This has been a known issue in Leveldb since 2019, but AFAIK they have not been interested in fixing it upstream.)

The PR contains two patches, one for leveldb 1.22 and one for leveldb 1.23, because the version of leveldb that the build uses [changes](https://github.com/firebase/firebase-cpp-sdk/blob/main/cmake/external/firestore.patch.txt) depending on whether the C++ SDK is doing the building.

A locally modified version of the C++ integration test, with international characters added to the path, was [tested and passed](https://github.com/firebase/firebase-cpp-sdk/actions/runs/4950581627).


dconeybe:
Holy smokes thanks for fixing this!! I wonder if we should be doing this patching in the firebase-ios-sdk repo instead of here in the firebase-cpp-sdk repo. I'm fine to merge it in here for now, but in the future we may want to move this patch upstream.

dconeybe (changes requested):
[deleted]

dconeybe (approved):
Please ignore my previous comments. I wasn't paying attention and thought we were in the firebase-cpp-sdk repo. This looks great. Thanks for fixing this long-standing issue!

dconeybe:
You'll need to add the patch file to https://github.com/firebase/firebase-ios-sdk/blob/eb9c7d6c65533c74ac3a3a0a81aea0ab16304f46/scripts/check_whitespace.sh#L29 to fix the whitespace checking.

jonsimantov:
> You'll need to add the patch file to
> 
> https://github.com/firebase/firebase-ios-sdk/blob/eb9c7d6c65533c74ac3a3a0a81aea0ab16304f46/scripts/check_whitespace.sh#L29
> 
> to fix the whitespace checking.

Done - let's see if the check passes. :)

cmake/external/leveldb_patch.py
```diff
@@ -50,6 +53,10 @@ def main() -> None:
   with cmakelists_txt_file.open("wt", encoding="utf8") as f:
     f.writelines(patched_lines)
 
+  if additional_patch_file:
```
dconeybe, line 56: @jonsimantov I just found a pretty annoying bug with this code: it needs to be re-entrant. That is, if you re-run "cmake" to update the build files this will attempt to re-apply the patch, then fail. The only recovery is to delete your cmake "build" directory and regenerate.

Typically these "re-entrant" semantics are achieved by creating a "stamp" file. After successfully applying the patch you'd create a "stamp" file to mark the patch as having been applied. Then, before applying the patch, you check the "stamp" file and only apply the patch if it has not previously been applied.

Here is an example from the python virtualenv setup code that uses a "stamp" file to avoid re-creating the virtualenv if it's already created: https://github.com/firebase/firebase-ios-sdk/blob/52ebbbf19d23baeb2b8c08b4b3d2b7195d5074cd/cmake/python_setup.cmake#L96-L117

PR #11277, opened 2023-05-11
Extract heartbeat logger from GACAppCheckAPIService

andrewheard:
Removed the direct dependency on `FIRHeartbeatLoggerProtocol` from `GACAppCheckAPIService`, moving it up into the App Check providers.

Added request hooks (blocks) to allow heartbeat logging headers to be added to requests before they are sent.

#no-changelog

### DIFF
--- a/AppCheck/Sources/AppAttestProvider/GACAppAttestProvider.m
+++ b/AppCheck/Sources/AppAttestProvider/GACAppAttestProvider.m
@@ -44,6 +44,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+// TODO(andrewheard): Remove from generic App Check SDK.
+// FIREBASE_APP_CHECK_ONLY_BEGIN
+static NSString *const kHeartbeatKey = @"X-firebase-client";
+// FIREBASE_APP_CHECK_ONLY_END
+
 /// A data object that contains all key attest data required for FAC token exchange.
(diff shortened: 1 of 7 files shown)


AppCheck/Sources/AppAttestProvider/GACAppAttestProvider.m
```diff
@@ -143,11 +148,17 @@ - (nullable instancetype)initWithApp:(FIRApp *)app {
   GACAppAttestKeyIDStorage *keyIDStorage =
       [[GACAppAttestKeyIDStorage alloc] initWithAppName:app.name appID:app.options.googleAppID];
 
+  GACAppCheckAPIRequestHook heartbeatLoggerHook = ^(NSMutableURLRequest *request) {
```
paulb777, line 151: This is temporary also, right?

andrewheard: Yep, thanks for catching! I just added some comments around these code blocks in each of the existing providers so they don't get missed.

Note: The request hooks concept will likely remain (or some alternative approach) but I'll need to pass them in from the Firebase App Check SDK since the generic App Check SDK shouldn't know about `FIRHeartbeatLoggerProtocol` once we're done refactoring (these code blocks themselves are temporary though).

PR #11279, opened 2023-05-11
[auth-swift] CocoaPods Swift-only sources

paulb777:
- No more Objective C `.m` files. 
- Public headers are kept around for typedef and global constant backwards compatibility
- Depend on Interop pods instead of headers in the repo

#no-changelog


ncooke3 (approved):

paulb777:
Will address the FirebasePod test failure in a subsequent PR

PR #11280, opened 2023-05-12
[auth-swift] Restore SPM

paulb777:
Now that the Auth implementation is all Swift, restore SPM build.  Note that Package.swift should be compared against the master version.

- Separate remaining ObjC unit tests to another directory so they're not built for SPM
- Cannot use an imported protocol via `@_implementationOnly`
- Disable heartbeat logging to restore build for SPM
- Remove GoogleUtilities imports from AuthDefaultUIDelegate.swift pending a solution for detecting running in an extension for Swift
- Continue disabling keychain dependent unit tests for SPM.
- Restore SPM CI
- The combine test failures and updated integration will be addressed in a subsequent PR


FirebaseAuth/Sources/Swift/Backend/AuthRequestConfiguration.swift
```diff
@@ -14,7 +14,6 @@
 
 import Foundation
 
-// TODO: Prefix with `@_implementationOnly` after port.
```
ncooke3, line 17: Don't we still want to do implementation only imports for Core Extension?

ncooke3 (approved):

paulb777:
Protocols don't work from  implementation only imports:

Screenshot 2023-05-12 at 6 48 13 AM

PR #11289, opened 2023-05-15
Progress closure for async/await APIs

### LINKED ISSUE #10574: [FR]: Async Await FirebaseStorage observers
### Description

https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorage/Sources/AsyncAwait.swift 

In this file we emit StorageUploadTask, and we can't add observers, so we don't know about the progress. (e.g. putFileAsync)

Uploading large files difficult to handle without knowing status

### API Proposal

Perhaps you can add delegate and handle all observers inside, then tell that something is changed to delegate?

### Firebase Product(s)

Storage

paulb777:
Fix #10574 

- Add an optional `onProgress` parameter to `putDataAsync`, `putFileAsync`, and `writeAsync`. 

Googlers, see API review at [go/firebase-storage-async-progress](http://goto.google.com/firebase-storage-async-progress)

### DIFF
--- a/FirebaseStorage/CHANGELOG.md
+++ b/FirebaseStorage/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+- [added] Add progress tracking capability for `putDataAsync`, `putFileAsync`, and
+  `writeAsync`. (#10574)
+
 # 10.10.0
 - [fixed] Fixed potential memory leak of Storage instances. (#11248)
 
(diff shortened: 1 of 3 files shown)


ncooke3 (approved):
Cool! I think a potential next step to be considered for future work would be to make a variation that accepts a SwiftUI binding that the  progress can be directly assigned to.

paulb777:
Now that the API review has been approved, I'll merge. If there are additional comment, I'll address in a follow up PR.

PR #11311, opened 2023-05-19
Fix error generation when status is .internal

tciuro:
[REQUIRED] Step 1: Describe your environment
Xcode version: Version 14.3 (14E222b)
Firebase SDK version: 10.9.0
Installation method: Swift Package Manager
Firebase Component: Functions
Target platform(s): iOS
[REQUIRED] Step 2: Describe the problem
When the error received is of type `.internal`, the current code rewrites the message with `descriptionForErrorCode`. This is pretty bad because any contextual information we might have received from Firebase is destroyed. This fix attempts to extract the message from the response payload, preserving the original value. However, if message is not found, the message is set to the value returned by `descriptionForErrorCode`, as was done before.

Steps to reproduce:
Simple. Just throw the following from a Cloud Function:

throw new functions.https.HttpsError('internal', 'Something really bad happened!')

The error generated in `func generatedError(userInfo: [String: Any]? = nil) -> NSError {...}` received a nil `userInfo`. This causes the message to be set to the error's `descriptionForErrorCode` value.

Relevant Code:
https://github.com/firebase/firebase-ios-sdk/commit/2aedc9a8c8321c0a8e2410c8ec610f07800a6b50

### DIFF
--- a/FirebaseFunctions/Sources/FunctionsError.swift
+++ b/FirebaseFunctions/Sources/FunctionsError.swift
@@ -225,18 +225,18 @@ internal func FunctionsErrorForResponse(status: NSInteger,
     if let status = errorDetails["status"] as? String {
       code = FunctionsErrorCode.errorCode(forName: status)
 
+      if let message = errorDetails["message"] as? String {
+        description = message
+      } else {
+        description = code.descriptionForErrorCode
+      }
+      
       // If the code in the body is invalid, treat the whole response as malformed.
       guard code != .internal else {
-        return code.generatedError(userInfo: nil)
+        return code.generatedError(userInfo: [NSLocalizedDescriptionKey: description])
       }
     }
-
-    if let message = errorDetails["message"] as? String {
-      description = message
-    } else {
-      description = code.descriptionForErrorCode
-    }
-
+    
     details = errorDetails["details"] as AnyObject?
     if let innerDetails = details {
       // Just ignore the details if there an error decoding them.


tciuro:
Hi @paulb777! I found a new issue and proposed the fix above

tciuro:
@paulb777 

OMG. The fact that this is by design floors me:
```swift
  func testUnknownError() {
    let byName = functions.httpsCallable(
      "unknownErrorTest",
      requestAs: [Int].self,
      responseAs: Int.self
    )
    let byURL = functions.httpsCallable(
      emulatorURL("unknownErrorTest"),
      requestAs: [Int].self,
      responseAs: Int.self
    )
    for function in [byName, byURL] {
      let expectation = expectation(description: #function)
      function.call([]) { result in
        do {
          _ = try result.get()
        } catch {
          let error = error as NSError
          XCTAssertEqual(FunctionsErrorCode.internal.rawValue, error.code)
          XCTAssertEqual("INTERNAL", error.localizedDescription)
          expectation.fulfill()
          return
        }
        XCTFail("Failed to throw error for missing result")
      }
    }
    waitForExpectations(timeout: 5)
  }
```
Why in the World would anyone benefit from this? I mean... an internal issue, which is already pretty bad and you remove all hope to determine what went wrong on the server. How could get past a review? 🤯

willbattel:
Respectfully, I think it makes sense. An internal error is just that: internal. The client should not be privy to the details of an internal error for a remote resource. For example, why should the server inform the client that it failed to reach a 3rd-party API or that it exceeded some quota in the GCP project- things that the client has no business knowing? If there was a problem with the request made by the client, then the other error codes would be more appropriate and allow the client an opportunity to resolve the problem. Internal error resolution, on the other hand, should be handled purely on the server. The client should simply retry (with backoff) and/or provide a general error message if appropriate.

That being said, it might make sense to break down the internal error code into subtypes (e.g., `internal-transient` and `internal-irreparable`) that give the client clues about  things like whether or not it should try again, etc. Just my 2 cents.

tciuro:
I get what you're saying but I, the API developer should be the one deciding whether the message/reason should reach the client. Why should Firebase decide for me? I own both client and server. Firebase is the conduit and should not have any business in deciding whether I should know or not the reason. It's my stack, after all. Shouldn't it be my decision?

Now, some 3rd party might want to conceal the reason. That's perfectly fine. Really. But I just happen to vend an API and I should have the last word. It's quite simple, really: why does Firebase care about it? Firebase API should be the messenger, not the judge.

This is straight from RFC 7231:
```
Server Error 5xx

   The 5xx (Server Error) class of status code indicates that the server
   is aware that it has erred or is incapable of performing the
   requested method.  Except when responding to a HEAD request, the
   server SHOULD send a representation containing an explanation of the
   error situation, and whether it is a temporary or permanent
```

Reference: end of page 62: https://datatracker.ietf.org/doc/html/rfc7231#page-62

paulb777:
IIRC, the logic was something along the lines that @willbattel describes. However, IMO there's no reason to hide information since it's really only being made harder to access. How about fixing the test as well?

@inlined any thoughts or suggestions?

tciuro:
I'm having issues with the Swift Package Manager (Xcode 14.3). No schemes are recognized, so I can't compile or test. Any ideas (I'm up-to-date with `main`).

image

tciuro:
I've reset the package cache, cleared derived data and it won't load.

paulb777:
@tciuro Sorry about the process problem here. We're in the middle of preparing the next release and the `master` branch is not buildable without setting an environment variable to use the next GoogleAppMeasurement release. 

Do `export FIREBASECI_USE_LATEST_GOOGLEAPPMEASUREMENT=1` in the command line before opening `Package.swift`. See https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift#LL1348C17-L1348C17

tciuro:
No worries, thanks @paulb777. I have been able to open the package and make the changes. I'm trying to verify them, but unfortunately I'm seeing other unit tests fail (e.g. `testExplicitError`). The ones I have updated fail here (not my actual change) due to:
```
testUnknownError(): XCTAssertEqual failed: ("13") is not equal to ("-1004")
```
Unit test:
```swift
XCTAssertEqual(FunctionsErrorCode.internal.rawValue, error.code) <<< fails here (?)
XCTAssertNotEqual("INTERNAL", error.localizedDescription) <<< my change
```
I suppose I could try to update the PR and see if all goes well, but I'd like to see them pass locally, of course.

inlined:
@tciuro I can see both sides of this argument. As far back as I can remember (e.g. 2001 ASP pages), major platforms hide internal error details in any production configuration because they tend to leak important information that can be used for a security exploit. Conversely, Firebase doesn't have an explicit development vs production mode (aside from the emulator) and these restrictions are server-side in those platforms and not client-side. Conversely, INTERNAL comes from [the Google-wide RPC status code](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L169), not the HTTP status. It is somewhat separate that status 13 is one of multiple codes that map to HTTP response 500.

Are you personally returning your own INTERNAL errors, or are you wishing that Firebase would return context in the INTERNAL errors that we return? I feel strongly against the latter because that includes stack traces, but would defer to @taeold on the former. Doing this right within the bureaucracy of Google may take a short while (e.g. a formal proposal reviewed by all SDK owners); in the meantime, is it an acceptable workaround to use the `precondition-failed` error?

inlined:
Also, regarding your comment about debugability, internal errors (with full details) are sent to Cloud Error Reporting.

paulb777:
Side note - test changes to make the PR green are at #11322

tciuro:
@inlined Thanks for the follow-up.

I'm not trying to pick a bone here, but I have some thoughts :-) I too have worked with older technologies (e.g. `CORBA`, `SOAP`, `RPC`), but I think we can agree that the standard nowadays tends to be `HTTP`. Heck, it's even in Firebase's API naming scheme! (`httpsCallable` and `HTTPSCallableResult`). Why won't the suite of `HTTP` error codes be available as part of the contract baffles me. But it is what it is.

Where I get a bit more uncomfortable is around APIs mentioning `https` and not quite following the standard. As I showed above, the `RFC` is pretty clear (I will give you my own reason in a bit) and the Firebase SDK isn't following it when it comes down to `INTERNAL` errors. Now, there might be an underlying reason why this is the case.. RPC parity? Similar core implementation with non-HTTP solutions? But then... why call it `httpsCallable`? As it is, it feels like it only goes half-way there. Again, there must be a reason or two for this, but from the consumer POV, it feels strange.

Now, why is this important to me? Say I have an API which itself utilizes other APIs to do the job: `Algolia` or a 3rd-party payment processor like Google Pay or Stripe. Our server is a client of these APIs. Being able to surface the fact that something in the transaction failed (and the reason why) would allow the client to display the proper UI to retry later, for example. Or at least mention that the payment didn't go through. Now, contrast this with `INTERNAL`. That's it. What does that even mean? Should we retry? Did it go through? Now what? You mentioned `precondition-failed` above. Sure it works... and so does `abort`. From our point of view... does it really matter? We're using `abort` for something else, not for the reason it probably was designed, but can you really blame me for that? According to https://www.rfc-editor.org/rfc/rfc7232#section-4.2:
```
The 412 (Precondition Failed) status code indicates that one or more
   conditions given in the request header fields evaluated to false when
   tested on the server
```
Not really what happened following the example I gave, right? I mean... this is not even close.

If I could simply send an `INTERNAL` with an associated message, then I'd probably say something like: "Unable to perform the payment" or something like that. This is something that my app could use... at least better than simply `INTERNAL`.

Having said that, I worked around this with a better solution. Skirting around this issue, I'm piggy-backing around this limitation by serializing an object to a string, which then becomes the message string. On the other side, I recreate the object and access the data. Of course, `INTERNAL` is the only code I cannot use, so I ended up using `abort` (because why not?) and encode everything I need in the object. I can encode things like:
```typescript
{
  "code" : "abort",
  "statusCode" : 500,
  "message" : "Unable to perform the payment",
  "details" : {
    "parameters" : {
      "foo" : 1,
      "bar" : "abc-def",
      "date" : "2023-05-23"
    },
    "service_details" : {
      "code" : "-10023"
    }
  }
}
```
This allows our app to react accordingly to some unforeseen issues We like it this way :-)

So all I'm saying is that in general, any web server/service should not have the final word. It should be impartial and follow the standard. That's all really. Let the service (us) and client (also us) decide how and what to communicate.

And lastly:

> Are you personally returning your own INTERNAL errors, or are you wishing that Firebase would return context in the INTERNAL errors that we return?

My own. I would not expect Firebase to return anything other than what I specify. I suppose I could return some stack trace if I was in debug mode. Though I don't, I'd like to have the freedom to encapsulate the response sent to the client. Of course other engineers won't agree, and that's fine. But who should really own the stack? :-) If I issue a `curl` or `Postman` request and decide to return anything, I should be able to debug it right there without having to resort to some logs somewhere. Just saying... it's a matter of preference.

If you've read the whole rant/explanation, thank you. I hope my POV is a bit more clear.

inlined:
This is helpful, thank you. I think the attitude of "Firebase should return the data I provided" makes sense, even if it's possible in this case that it risks conflict with our philosophy of "take footguns away from developers". I wonder if there is a middle ground between our perspectives, for example, where internal errors might print stack traces in the emulator and be opaque in production. That would at least follow state of the art, though it would still inhibit your case of providing structured details for why a request failed. Frankly, I think this stems from the fact that HTTP needs a new error code for a server side error that is safer to disclose (since, irrespective of the RFC, server-side libraries tend to treat INTERNAL as sensitive).

I think the best course of action for now is that you continue to use "abort" or "precondition-failed" while we continue to take your request seriously and escalate it amongst people with very detailed opinions about very arcane questions. I promise to escalate this issue internally, but very smart people who have seen many innocent mistakes may be of the opinion that explaining "payment failed" may indicate to an attacker whether they've successfully caused your backend to attempt an unauthorized payment or whether validation rejected the attempt. The general dogma is that a client shouldn't know these details.

vpusher:
100% with @tciuro here. Let's be honest, it took me a certain amount of time to understand what was going wrong with my iOS client implementation. To finally discover it was actually right.

Why Firebase, as an open development platform, would interfere in a such non-standard, arbitrary and hidden way between a client and its backend ? I can understand the security concerns, but in that case don't even send the response payload from the server in the first place. Not parsing it won't mitigate the data disclosure risk.

Side question to @tciuro, why did you not include the error `details` too in the **userInfo** in your fix ?

tciuro:
@vpusher because... I did not know about it 😉

I have forced a firebase `verifyIdToken` error (simply by appending extra chars to the valid token) and this is what I get back:

image

Questions:

- what would I get by including `FunctionsErrorDetailsKey`?
- how do I access that key?

vpusher:
@tciuro If you look at the `HttpsError` constructor:

```
constructor(code: FunctionsErrorCode, message: string, details?: unknown);
```

You can provide an `details` object to be sent to the client. Again, in the case of `internal`, this object is not parsed on the client side cause of the [early return](https://github.com/firebase/firebase-ios-sdk/blob/89adc2a82388070352219cc7d81e35087998050d/FirebaseFunctions/Sources/FunctionsError.swift#L236). Thanks to this last parameter, you don't need to _Stringify_ anything. Error details ends up directly in the **userInfo** dictionary :
```
throw new functions.https.HttpsError('internal', 'My error message', {'foo': 'bar'});
```

`FunctionsErrorDetailsKey` is just a global variable from the iOS SDK to access the details object in the **userInfo**:
```
error.userInfo[FunctionsErrorDetailsKey] // ["foo": "bar"]
```

PR #11322, opened 2023-05-23
Update tests for exposing internal error description

paulb777:
Fix tests for #11311


inlined (commented):
While I understand the intent behind this code, I'm curious where this came from and whether it would help. For example, [here's](https://github.com/firebase/firebase-functions/blob/master/src/common/providers/https.ts#LL187C1-L189C31) the docs for the internal error:

```
- `internal`: Internal errors. Means some invariants expected by
   underlying system has been broken. If you see one of these errors,
   something is very broken.
 ```

For this category of errors, we cannot know if they contain sensitive (esp security sensitive) information and so the server shouldn't be sending context, let alone the client exposing it. The only two places where we return an internal response are if the customer's version of the Admin SDK is too old to support AppCheck features they're trying to use or if the RPC has an unhandled exception.

If you want to make this change because you have user reports that customers are returning "internal" errors with message details, I'd suggest talking to @taeold (our TL) to agree this is the right course and then draft an API proposal so we can keep this in sync in our iOS, Android, and Web SDKs if not also our C++ and Unity SDKs.

paulb777:
@inlined or @taeold Would you follow up on the external PR discussion at https://github.com/firebase/firebase-ios-sdk/pull/11311#issuecomment-1555395942?

PR #11324, opened 2023-05-24
Remove Firebase-specific implementation from GACAppCheckAPIService

andrewheard:
- Removed the unused `appID` property from `GACAppCheckAPIService`.
- Removed heartbeat logging tests from `GACAppCheckAPIServiceTests`.
- Added request hook, additional headers and API key tests to `GACAppCheckAPIServiceTests`.

#no-changelog

### DIFF
--- a/AppCheck/Sources/Core/APIService/GACAppCheckAPIService.m
+++ b/AppCheck/Sources/Core/APIService/GACAppCheckAPIService.m
@@ -25,8 +25,6 @@
 #import "AppCheck/Sources/Core/Errors/GACAppCheckErrorUtil.h"
 #import "AppCheck/Sources/Core/GACAppCheckLogger.h"
 
-#import "FirebaseCore/Extension/FirebaseCoreInternal.h"
-
 #import <GoogleUtilities/GULURLSessionDataResponse.h>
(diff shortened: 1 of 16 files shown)


AppCheck/Sources/Core/APIService/GACAppCheckAPIService.m
```diff
@@ -40,8 +38,7 @@
 @interface GACAppCheckAPIService ()
 
 @property(nonatomic, readonly) NSURLSession *URLSession;
-@property(nonatomic, readonly) NSString *APIKey;
-@property(nonatomic, readonly) NSString *appID;
+@property(nonatomic, readonly, nullable) NSString *APIKey;
```
ncooke3, line 41: why is APIKey nullable?

ncooke3 (approved):

AppCheck/Sources/Core/APIService/GACAppCheckAPIService.m
andrewheard: The new endpoint that the GSI SDK will use won't require an API key (just the OAuth Client ID). The Firebase App Check SDK will still need to provide an API key though.

PR #11325, opened 2023-05-24
Fix API diff tool not detecting api changes

sunmou99:
There is an issue capturing the API change, the API change was not caught: https://github.com/firebase/firebase-ios-sdk/pull/11289

This is because we added "customized section headers". This is intentional.
Now, the tool will check all the apis under all the "customized section headers"

Test result: https://github.com/firebase/firebase-ios-sdk/pull/11326#issuecomment-1561687774


morganchen12 (approved):

paulb777 (approved):
Nice! Thanks @sunmou99

PR #11328, opened 2023-05-25
[auth-swift] Rewrite Auth RPC types using Swift types

ryanwilson:
This largely stems from the unsafe casting that occurs. Current state:
all types have been converted, and it builds. Further testing may be needed.


ryanwilson:
Rebased on auth-swift

paulb777:
Subsumed into #11529

